PR #585, opened 2011-10-28
Suggestion for single embedded document support

pmlopes:
Hi,

One of my challenges with mongoose is to have a single embedded doc in a doc, this patch allows this kind of schema declaration:

var mongoose = require('mongoose'),
    Schema = mongoose.Schema;

mongoose.connect('mongodb://localhost/mongoose');

var PersonSchema = new Schema({
  FirstName: {type: String, required: true},
  LastName: String
});

var BlogPostSchema = new Schema({
  creator: PersonSchema
});

/**
- Define model.
  */

var BlogPost = mongoose.model('BlogPost', BlogPostSchema);
var Person = mongoose.model('Person', PersonSchema);

var person = new Person({FirstName: 'Paulo', LastName: 'Lopes'});

var post = new BlogPost({title: 'test', creator: person});

post.save(function(err) {
  if (err) return console.log(err);
  BlogPost.find({}, function(err, docs) {
    if(err) return console.log(err);
    console.log(docs[0].creator);
    docs[0].creator = null;
    docs[0].save(function(err) {
      if(err) return console.log(err);
      mongoose.disconnect();
    });
  });
});

I don't know much about the internals of mongoose but this seems to work for me... could someone review and give some feedback? Thanks

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -96,18 +96,38 @@ Schema.prototype.tree;
 
 Schema.prototype.add = function (obj, prefix) {
   prefix = prefix || '';
+
   for (var i in obj) {
     // make sure set of keys are in `tree`
     if (!prefix && !this.tree[i])
       this.tree[i] = obj[i];
 
-    if (obj[i].constructor.name == 'Object' && (!obj[i].type || obj[i].type.type)) {
-      if (Object.keys(obj[i]).length)
-        this.add(obj[i], prefix + i + '.');
-      else
-        this.path(prefix + i, obj[i]); // mixed type
-    } else
-      this.path(prefix + i, obj[i]);
+    // one-to-one
+    var tree = null;
+
+    if(obj[i] instanceof Schema) tree = obj[i].tree;
+    if(obj[i].type !== undefined && obj[i].type instanceof Schema) tree = obj[i].type.tree;
+    if(obj[i].type !== undefined && obj[i].type.type !== undefined && obj[i].type.type instanceof Schema) tree = obj[i].type.type.tree;
+
+    if(tree !== null) {
+      var embeddedDoc = {};
+      for(var k in tree) {
+        if(k !== 'id' && k !== '_id') {
+          embeddedDoc[k] = tree[k];
+        }
+      }
+      this.add(embeddedDoc, prefix + i + '.');
+    }
+    // normal path
+    else {
+      if (obj[i].constructor.name == 'Object' && (!obj[i].type || obj[i].type.type)) {
+        if (Object.keys(obj[i]).length)
+          this.add(obj[i], prefix + i + '.');
+        else
+          this.path(prefix + i, obj[i]); // mixed type
+      } else
+        this.path(prefix + i, obj[i]);
+    }
   }
 };
 


seanrucker:
+1

dortzur:
+1
suppp

johan13:
+1

JakeBent:
@jesseclark @joshdickson40 

If i'm reading correctly, the concern is about handling the lifecycle of the subdocument. Ruby figured it out: http://mongoid.org/en/mongoid/docs/relations.html#embeds_one javascript should be able to as well.

joshdickson40:
@JakeBent It has nothing to do with the language when slapping it in an array makes it work, it's just an arbitrary point that for some reason we can't get a better definition on.

@aheckmann is there something that everybody +1ing this should be looking at instead? This is a huge sticking point to people who are new to Mongo/Mongoose

vkarpov15:
@joshdickson40 It's something I'm planning on addressing after 4.0 is released. Its definitely an additive feature that I'd like to put into 4.1. I don't know of any alternatives that support this.

ashaffer:
@vkarpov15 Is there a roadmap for 4.x somewhere? Or any sort of place for the community to discuss/suggest features/architecture for the next version?

vkarpov15:
Right now github is the best place, although I'm open to moving this discussion over to Google groups as well.

Jeff-Lewis:
May I suggest we keep the topic here for those that might join in order to keep its context? ( I +1'ed this [over a year ago](https://github.com/LearnBoost/mongoose/pull/585#issuecomment-26552251).)

Also, can we try to define the scope of the desired behavior if this were to be implemented? If we can agree on the behavior and scope, hopefully the implementation will be the easier part. :)

Although this issue was [closed in 2012](https://github.com/LearnBoost/mongoose/pull/585#event-11740912), @aheckmann mentioned [a pull request is welcome](https://github.com/LearnBoost/mongoose/pull/585#issuecomment-32186973) which somewhat keeps issue [open in the minds of us developers](http://i.imgur.com/WHGi867.gif).

@aheckmann & @vkarpov15 Can you help us outline what the pull request would require in order to capture the main responsibilities of a schema/model?

Everyone add/change as you see fit. Here are some starting points.

#### Requested Behavior: ####
* Calling .save() on sub-document actually saves changes to the sub-doc. [https://github.com/LearnBoost/mongoose/pull/585#issuecomment-2621715]
* Validation events fire for sub-documents as normal docs would.

#### Questions: ####
* Do validation events fire for parent hierarchies of objects? Can we keep it simple and have the validation events fire only on the sub-document(s) and not their parent doc(s)?
* @aheckmann Would supporting sub-documents play well with mquery? Does supporting sub-tasks have larger implications than most of us here are not aware of?

ashaffer:
@Jeff-Lewis @vkarpov15 

I agree that we should try to define the scope of this feature, but the reason I was requesting a place to discuss 4.x more generally was that I think the reason this feature hasn't been implemented yet, and the reason that it is so difficult to fit into the existing codebase is that mongoose has grown pretty byzantine over time.

As such, I think any approach to implementing a complex additional feature like this should probably begin with modularizing and simplifying the existing codebase.  I.e. pulling features like population and mquery which are relatively orthogonal to the core of mongoose out into official plugins, etc.

Jeff-Lewis:
@ashaffer Thanks. I suspected this should be considered a larger undertaking than possibly expressed in this thread. If implementing this carries the baggage of splitting mongoose core, those of us with +1's should not expect anything for a while, if ever.

AFAIK,  [4.0 Release notes are here](https://github.com/LearnBoost/mongoose/wiki/4.0-Release-Notes).

vkarpov15:
Don't worry too much about mquery - I'm inclined to stop using mquery in mongoose in the future, because IMO its not an effective abstraction. Does more to obfuscate the internal code than anything else.

Re: this issue, I don't see it as being dependent on splitting the core at all. The change in the API is relatively simple, I just have not experimented enough to see what roadblocks there are.

ashaffer:
Ya, my point was more that those things should probably be configurable by the end user, rather than included as core features of mongoose.  And that if that stuff is pulled out of the core, it'll be easier to extend the actual core features of the models like this in a way that makes sense now and in the future.

E.g.
```javascript
mongoose.plugin(require('mongoose-populate'))
mongoose.plugin(require('mongoose-mquery'));
```

Then people can build their own query engines or population strategies either replacing the official plugins or building on top of them, and the core codebase will be far simpler to understand.

Also, mquery and population are just two examples that came to mind, there are probably a number of other features that this could be done with.

cava23:
+1

suceava:
+1

bertramn:
+1

ramSeraph:
+.  This was a rude shock.

dozoisch:
:+1: !

alejandromagnorsky:
+1

sirhoe:
+1

josencv:
+1

jeremyml:
+1

julienboulay:
+1

netusco:
+1

molinto:
+1

v1k4s:
+1

adambuczynski:
+1

zaun:
@vkarpov15 - Is there any status on this? Its version 4.1.0 but when I set a property to a schema I get errors, it still needs to be in an array of length 1.

vkarpov15:
The reason this is closed is that its a pull request that will never get merged. If you look at the thread you'll see that this issue has been moved over to #2689

adambuczynski:
Thanks @vkarpov15 .
It seems this feature is scheduled for the 4.2 milestone. Good to hear.

zaun:
Thank for the info @vkarpov15 
I'm glad to see it'll be added but disappointed it'll be so far out. I can't complain though, I'm not the one working on the project.

willerup:
I ran into this issue and then found this thread. I updated my mongoose to 4.2 and "voila" it now works as expected. Thank you mongoosers!

willerup:
I ran into this issue and then found this thread. I updated my mongoose to 4.2 and "voila" it now works as expected. Thank you mongoosers!

vkarpov15:
@willerup glad you like it!

For posterity's sake, mongoose `>= 4.2.0` supports this behavior.

joniba:
I'm on 4.2.5 and still getting this MongooseError: "Did you try nesting Schemas? You can only nest using refs or arrays."

Is there an example or documentation somewhere of this functionality?

vkarpov15:
@joniba [single nested subdoc docs](http://mongoosejs.com/docs/subdocs.html#single-embedded)

jeremyml:
Great feature - thank you!

vkarpov15:
@jeremyml glad you find it useful :beers:

PR #1167, opened 2012-10-20
markNotModified

ghost:
Added markNotModified to document.js for certain cases where you want to prevent fields from being saved to MongoDB in pre-save middleware.


lib/document.js
```diff
@@ -698,6 +698,19 @@ Document.prototype.markModified = function (path) {
...
 /**
+ * Marks the path as not having changes to write to the db.
+ *
+ * _Helpful to prevent changes to certain fields based on logic in  middleware_
+ *
+ * @param {String} path the path to mark as not modified
+ * @api public
+ */
+
+Document.prototype.markNotModified = function (path) {
```
paglias: @vkarpov15 as far as you know this method still works?

vkarpov15: Nope. The general idea would work, but this particular implementation won't work with modern mongoose.

PR #1214, opened 2012-11-15
multiple validator errors

### LINKED ISSUE #1031: test and record multiple validation failures
In my Schema, I'm adding multiple validations for a user registration field. Here's an example: 

```javascript
	, email : {
				type: String
			, index: true
			, lowercase: true
			, validate: [
					{
							'validator' : validator.isRequired
						, 'msg' : 'This field is required.'
					}
				, {
							'validator' : validator.isEmail
						, 'msg' : 'This field must contain an Email Address.'

					}
			]
		}
```

I've noticed that it seems to execute the validations starting fromt the last. So if the field is both empty and not an email address upon submission, the validator records the 'Email' validation error and not the 'isRequired' error. 

```javascript
{ message: 'Validation failed',
  name: 'ValidationError',
  errors: 
   { password: 
      { message: 'Validator "This field is required" failed for path password',
        name: 'ValidatorError',
        path: 'password',
        type: 'This field is required' },
     email: 
      { message: 'Validator "This field must contain an Email Address." failed for path email',
        name: 'ValidatorError',
        path: 'email',
        type: 'This field must contain an Email Address.' },
     name: 
      { message: 'Validator "This field is required" failed for path name',
        name: 'ValidatorError',
        path: 'name',
        type: 'This field is required' } },
  body: { name: '', email: '', password: '' } }
```

I'm wondering if there would be any way to go through all of the validations, regardless if one fails, so that multiple validation error messages could be recorded and sent back to the user. I realize this probably isn't optimal default behaviour, but the option might be nice.

### LINKED ISSUE #1104: Trying to retrieve all validation errors
Hi,

Let's say I want to validate several conditions on a given path:
User.schema.path('password').validate(function (value) { ... }, 'error1');
User.schema.path('password').validate(function (value) { ... }, 'error2');

When I save a document that fails on both conditions like this:
user.save(function(err) {..})

I have two problems:
- err only contains 'error1'
- 'error2' is not caught by the user.save callback and then appears in the node process:
  "No listeners detected, throwing. Consider adding an error listener to your connection"

Is there a way to ask mongoose not to call User#save callback until all validations are tested?

Thanks,
Guillaume

RGBboy:
Fixes #1031
Fixes #1104

This fix changes the way validator errors are added to the validation error. When multiple errors occur for a property, an array is created and populated with all the validator errors.

A test has also been added to schema.test.js

Let me know if this fits the bill.


memelet:
This is closed because ???

vkarpov15:
Opened #2612 to track

PR #1337, opened 2013-02-07
Domain module support

DrewMartin:
Added domain support to mongoose.

Almost every query uses either a Promise or utils.tick so both of those will bind the current process's domain to the callback, if the process has a domain. As far as I could tell, the only other places that needed it were mapReduce and aggregate.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1676,6 +1676,9 @@ Model.mapReduce = function mapReduce (o, callback) {
     q = undefined;
   }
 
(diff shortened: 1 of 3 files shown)


coolgeng:
@aheckmann thanks.

vkarpov15:
FWIW domains are really buggy, I would recommend avoiding them. More modern alternatives include [zones](https://github.com/strongloop/zone).

PR #1647, opened 2013-08-15
added support for discriminator mapping (closes #1003)

j:
So I gave discriminator mapping a shot.  After coding this out, it feels pretty natural and seems to fit well with the current API of mongoose.

Anyway, a quick example of how it works:

``` javascript
function BaseSchema() {
    Schema.apply(this, arguments);

    this.add({
        name: String,
        createdAt: { type: Date, default: Date.now }
    });
}
util.inherits(BaseSchema, Schema);

var EventSchema = new BaseSchema();

var ImpressionEvent = new BaseSchema();

var ConversionEvent = new BaseSchema({
    revenue: Number
});

var Base = mongoose.model('Event', EventSchema),
var Impression = Base.discriminator('Impression', ImpressionEvent),
var Conversion = Base.discriminator('Conversion', ConversionEvent);
```

Then assuming we have inserted a single document for each event in the order of Base, Impression, and Conversion:

``` javascript
Base.find({}, function(err, events) {
    expect.ok(events[0] instanceof Base);       // true
    expect.ok(events[1] instanceof Impression); // true
    expect.ok(events[2] instanceof Conversion); // true
    expect.ok(events[2] instanceof Base);       // false

    // however
    expect.ok(events[2].schema instanceof BaseSchema); // true
});
```

Lastly, if you take a the specific discriminator type model, it will query for only the documents of that type:

``` javascript
// find all conversion events
Conversion.find({}, function(err, events) {
    expect.equal(events.length, 1);
});

// or a single document
Conversion.findOne({}, function(err, event) {
    expect.ok(event instanceof Conversion); // true
});

// or querying for a specific id which is an invalid type (produces query { _id: ..., __t: 'Conversion' })
Conversion.findOne({ _id: impression.id }}, function(err, event) {
    expect.equal(event, null); // true
});
```

Just one thing before moving too far forward:  I introduced a new method "discriminator" to be clear of what you're doing.  It creates a model in the same connection and returns it and does some other stuff behind the scenes.  We could also make the static method Model.model do the same (or alias / remove discriminator method) if the community would rather have the following instead:

``` javascript
var Base = mongoose.model('Event', EventSchema);
var Impression = Base.model('Impression', ImpressionEvent),
var Conversion = Base.model('Conversion', ConversionEvent);
```

Another note, in my example, I'm using util.inherit to simply show that you can have a base model.  This is completely optional.  You can technically store whatever schema's you want and they don't have to inherit from one another.  I'm also +1 for creating an extend method in Schema to shortcut schema inheritance.

Anyway, here you go! :) :surfer: 
## Todos
- [ ] Reference mapping validation (throw error when reference type doesn't match)
- [ ] More regression tests (reference mapping, ...)
- [x] Force discriminator schemas to inherit root (?)
- [x] Model inheritance (?)
- [ ] ...
##### Off topic:

After this, we can discuss adding "discriminator" style mapping to embedded documents.  They are sort of unrelated in the scope of this PR... Unless someone can think of a way to combine the two in a mongoose style fashion :P

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -653,8 +653,27 @@ Model.discriminator = function discriminator (name, schema) {
     schema.add(obj);
     schema.discriminatorMapping = { key: key, value: name, isRoot: false };
 
-    var toJSON = schema.options.toJSON,
-        toObject = schema.options.toObject;
+    if (baseSchema.options.collection) {
+      schema.options.collection = baseSchema.options.collection;
+    }
+
+      // throws error if options are invalid
+    (function validateOptions(a, b) {
+      a = utils.clone(a);
+      b = utils.clone(b);
+      delete a.toJSON;
+      delete a.toObject;
+      delete b.toJSON;
+      delete b.toObject;
+
+      if (!utils.deepEqual(a, b)) {
+        throw new Error("Discriminator options are not customizable (except toJSON & toObject)");
+      }
+    })(schema.options, baseSchema.options);
+
+    var toJSON = schema.options.toJSON
+      , toObject = schema.options.toObject;
+
     schema.options = utils.clone(baseSchema.options);
     if (toJSON)   schema.options.toJSON = toJSON;
     if (toObject) schema.options.toObject = toObject;
--- a/test/model.discriminator.test.js
+++ b/test/model.discriminator.test.js
@@ -38,7 +38,7 @@ PersonSchema.post('save', function (next) {
 PersonSchema.set('toObject', { getters: true, virtuals: true });
 PersonSchema.set('toJSON',   { getters: true, virtuals: true });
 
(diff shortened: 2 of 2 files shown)


ericsaboia:
Well done, @j! 

I just get an issue using discriminators: If you have some references in your child model, you can't populate it using the base model:

Child:

```javascript
new BaseSchema({
  user: { type: mongoose.Schema.Types.ObjectId, ref: 'User' }
});
```

This works as expected:

```javascript
Child.find().populate('user');
```
But this does not populate referenced users:

```javascript
Base.find().populate('user');
```

vkarpov15:
@ericsaboia can you open up a new issue for that? Makes it much easier to track :)

ericsaboia:
@vkarpov15 sure.

Issue #1269, opened 2013-01-02
Connection's use of for..in loops breaks when used with code that decorates Object.prototype


carpeliam:
Right now I'm working on a project that makes use of a library that (unfortunately) decorates Object.prototype (http://code.google.com/p/jsmeter/source/browse/trunk/jsmeter/tokens.js). Because mongoose uses for..in loops (e.g. https://github.com/LearnBoost/mongoose/blob/122ed8f4b1a8a0dd3fe7b2ca9ed61df2d3f020ba/lib/connection.js#L427), using this library with mongoose results in serious breakage. From Ch. 6 of The Good Parts,

> Since JavaScript's arrays are really objects, the for in statement can be used to iterate over all of the properties of an array. Unfortunately, for in makes no guarantee about the order of the properties, and most array applications expect the elements to be produced in numerical order. **Also, there is still the problem with unexpected properties being dredged up from the prototype chain.**

Refactoring for..in style for loops to be replaced by old school C style for loops (eg, `for (var i = 0; i < ary.length; i++) { ...`) will protect mongoose from other libraries that decorate the prototype.

Although I'd much rather use a library that doesn't decorate Object.prototype, it'd be nice (and not too hard, it seems) for mongoose to be safe to use in the event that someone does do this.

Let me know if you'd like me to add test code and fixes.

PR #1826, opened 2013-12-07
Take 2: Ensure use of .hasOwnProperty() check when using for/in loops

ahildoer:
This is a fix for #1269
Added style fix


aheckmann:
Closing. Nearly 3 years old.

Issue #1898, opened 2014-01-26
versionKey check broken for certain update queries


davidtsai:
Sometimes mongoose generates incorrect queries that doesn't check the version key, __v, properly.  Example query generated by mongoose 3.8.5:

query: { _id: ObjectId('52e49ed52996a6145200000c') } update: { $set: { itemIndex: 6, money: 700, rid: 9 }, $pushAll: { items: [ { itemId: 102, qty: 1, sockets: 0, id: 6, runes: {} } ] }, $inc: { __v: 1 } }

Notice the __v is missing in the query.  This query was generated from a save() call to a model.  Looking into the mongoose code a bit, it appears this bug is caused when the update operation contains a pushAll to a sub-document as well as a $set on the parent document's fields.

PR #1942, opened 2014-02-27
Fix model version not being checked properly on queries with push operands.  Issue #1898

davidtsai:
This bug affects anyone who is depending on mongoose for optimistic locking and has arrays in their documents.


paglias:
@vkarpov15 do you know if this is still happening? We've an issue that might be caused by the versioning key not being checked/increased correctly when saving a document with a plain array modified by using `doc.array.push()`.

vkarpov15:
@paglias can you open up a separate issue? This PR is super stale.

PR #1981, opened 2014-03-28
Clear error after document initialization

ngoyal:
An error may be set when doing the following (lines 287 - 289)
```
self.$__try(function(){
    doc[i] = schema.cast(obj[i], self, true);
});
```
The error lingers as `this.$__.saveError` and can cause an error if the document is updated with valid members and saved via `.save()`.


vkarpov15:
This was fixed for 4.0.9, there's no more `this.$__.saveError`, cast errors are instead stored on a per-path basis.

PR #2247, opened 2014-08-13
Conditionalrequired

nelisbijl:
Per request moved from 3.8.x to master (see  #2220)
(Git is great but extremely difficult when you lack the experience; hope I didn't cause any trouble)

Conditional required: allows the value for required to be a callback:
name : { type: String, required: function () { /\* this is doc */ return true/false; } }

### DIFF
--- a/test/schema.validation.test.js
+++ b/test/schema.validation.test.js
@@ -204,7 +204,7 @@ describe('schema', function(){
         done();
       });
 
-      it.only('string conditional required', function (done) {
+      it('string conditional required', function (done) {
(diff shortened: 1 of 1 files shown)


poohitan:
Is it possible to have a custom validation message while using the conditional required?

vkarpov15:
@poohitan unfortunately not at the moment. Please open up a separate issue if you want to make a formal feature request :+1:

poohitan:
@vkarpov15 thanks, I have already created #5123

PR #2428, opened 2014-10-31
Added globalConditions to schema options

JNK:
This allows you to apply a fixed condition to every mongoldb query.

Example:
var userSchema = new Schema({
    globalConditions: {
        _type: 'user',
    },
    collection: 'accounts',
});

var adminSchema = new Schema({
    globalConditions: {
        _type: 'admin',
    },
    collection: 'accounts',
});

var accountSchema = new Schema({
    collection: 'accounts',
});

This way you can easily manage single table inheritance.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -21,6 +21,7 @@ var Document = require('./document')
   , Promise = require('./promise')
   , assert = require('assert')
   , util = require('util')
+  , utils = require('./utils')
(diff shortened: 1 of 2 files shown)


vkarpov15:
Gonna close this. It's inactive, and also this is something that really should be done as a [schema plugin](http://mongoosejs.com/docs/plugins.html)

Issue #2378, opened 2014-10-14
Dot operator not fetching child properties of a Mongoose Document object


laggingreflex:
I have a mongoose schema less schema 

```
var User = new mongoose.Schema({}, {
    strict: false,
});
```

and any object I save with it, seems to contain the properties that I saved

```
console.log(user); /* =>
{ username: 'user1',
  salt: '3303187e50a64889b41a7a1c66d3d3c10b9dec638fdd033bee1221d30d01c5e1',
  hash: 'a174c206d88bee1594bb081dbd32d53420f6ef3d6322104f3d0722d58bc8dd8d',
  _id: 52d3177481daf59c11000001,
  __v: 0 } */
```

but when I try to access those properties individually with a dot operator, it outputs undefined!!

```
console.log(user.hash); // => undefined
```

![](http://i.imgur.com/Z2Ij5Bw.png)

What could be causing this?

This only happens with the properties that aren't declared in the schema. If I declare `hash` in the schema it works. Is this by design? (that you can't access schema-less properties)  And even if so, how is it even achieved - not allowing access properties with dot operator?

PR #2439, opened 2014-11-04
gh-2378 enable getters for properties defined with non-strict schemas; u...

alabid:
See #2378 for more details. The tests in `test/document.strict.test.js` have been appropriately updated as well. 
All tests (including new `assert`'s pass.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -61,6 +61,17 @@ function Document (obj, fields, skipId) {
     this.set(obj, undefined, true);
   }
 
+  if (!schema.options.strict && obj) {
+    var self = this
+      , keys = Object.keys(this._doc);
+
+    keys.forEach(function(key) {
+      if (!(key in schema.tree)) {
+        define(key, null, self);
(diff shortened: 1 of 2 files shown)


laggingreflex:
Is this solved, or pushed to npm? I'm using 3.8.23 and the issue is still present

laggingreflex:
Also just found out that this doesn't solve all cases. Only values that are defined work. Any new properties created are still ignored.

    var User = mongoose.model('users', new mongoose.Schema(
        {username: String} , {strict: false}
    ));
    
    var user = new User({a: 'a'}); //=> a is defined

    user.a = 'b';
    console.debug(user.a);        //=> 'b' (ok)
    console.debug(user.get('a')); //=> 'b' (ok)

    user.c = 'c';                 //=> c was not defined
    console.debug(user.c);        //=> 'c' (...)
    console.debug(user.get('c')); //=> undefined (not ok!)

:(

PR #2554, opened 2014-12-20
The performance improvement Model.prototype.$__handleSave() method

archangel-irk:
By removing unnecessary steps `utils.clone(this.schema.options)` can achieve the best performance for obvious reasons.


archangel-irk:
Yes, of course, I'll also stick with it.

PR #2576, opened 2015-01-05
Allow explicitly excluding paths from versioning

csabapalfi:
First of all versioning is super useful. Thanks!

I'd like the ability to exclude certain array paths though. If elements in an array are only ever soft-deleted or properties updated one-by-one there's no point in including it in versioning and can result in false positive ```VersionError```s on concurrent updates.

Suggested syntax: 
```js
schema.set('notVersionedPaths',['safeArrayPath']);
```

Happy to contribute this with whatever other syntax/naming you prefer. 

What do you think? Are you interested in something like this?

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -287,9 +287,9 @@ function operand (self, where, delta, data, val, op) {
   // disabled versioning?
   if (false === self.schema.options.versionKey) return;
 
-  // path excluded from versioning
-  var notVersionedPaths = self.schema.options.notVersionedPaths;
-  if (notVersionedPaths && notVersionedPaths.indexOf(data.path) >= 0) return;
+  // path excluded from versioning?
+  var skipVersioning = self.schema.options.skipVersioning;
+  if (skipVersioning && skipVersioning[data.path]) return;
 
   // already marked for versioning?
   if (VERSION_ALL === (VERSION_ALL & self.$__.version)) return;
--- a/lib/model.js
+++ b/lib/model.js
@@ -287,6 +287,10 @@ function operand (self, where, delta, data, val, op) {
   // disabled versioning?
   if (false === self.schema.options.versionKey) return;
 
+  // path excluded from versioning
+  var notVersionedPaths = self.schema.options.notVersionedPaths;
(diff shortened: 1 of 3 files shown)


csabapalfi:
Trying to add a test now :)

csabapalfi:
I just realized that what's causing our problems is that push is considered a dangerous operation, incrementing the version even if position wasn't specified. Another potenatial change is to update to code to check if push is actually changing Array position by looking at if the position argument was specified.

lib/model.js
```diff
@@ -287,6 +287,10 @@ function operand (self, where, delta, data, val, op) {
   // disabled versioning?
   if (false === self.schema.options.versionKey) return;
 
+  // path excluded from versioning
+  var notVersionedPaths = self.schema.options.notVersionedPaths;
+  if (notVersionedPaths && notVersionedPaths.indexOf(data.path) >= 0) return;
```
vkarpov15: IMO `notVersionedPaths` should be stored as a map, no sense in storing it as an array if you only use it for lookups

csabapalfi:
Makes sense, let me update my patch. I assume you mean something like this:
```schema.set('skipVersioning', { safeArrayPath: true });```

Also changed the config key name to skipVersioning as it seems bit more readable like that.

vkarpov15:
I like this idea, but you should add tests.

Also, I'm not entirely sure about whether push() should cause a version increment in all cases. It definitely should cause a version increment if $position is specified as you mentioned, as well as $slice and $sort, but I haven't been able to quite convince myself that its safe if none of those are specified. Do you have a good argument why it shouldn't increment the version if $position, $slice, and $sort aren't used?

csabapalfi:
The main argument I have is that we're confident it's not changing array positions in that case so we should allow concurrent updates (as they're safe). I think mongoose should not force you to handle VersionErrors when the updates would not cause problems.

csabapalfi:
Maybe I'm missing something but I assume safe === no array positions are changed.

Also skipVersioning should do the job for now for my use-case :)

csabapalfi:
Tests added and now the option called `skipVersioning` and it's a map.

Issue #2519, opened 2014-12-06
Model.findOneAndUpdate fails when using option upsert: true


benkroeger:
When using findOneAndUpdate in combination with upsert = true, the mongodb driver throws a MongoError:

```
MongoError: remove and upsert can't co-exist
```

This is what I'm trying to execute:

``` javascript
return Tag.findOneAndUpdate({
        $and: [{
          term: tag.term
        }, {
          category: tag.category
        }]
      }, {
        $setOnInsert: {
          term: tag.term,
          category: tag.category
        }
      }, {
        select: {
          _id: 1
        },
        new: true,
        upsert: true
      }).exec('findOneAndUpdate');
```

The reason this fails is, that the mongodb dirver sets the option "remove" to 1 when options.remove resolves to "true:

mongodb/lib/mongodb/collection/core.js: Line 747

``` javascript
queryObject.remove = options.remove ? 1 : 0;
```

Model.findOneAndUpdate results in an internal call to Query.prototype._findAndModify, where the opts object is created like this

mongoose/lib/query.js: Line 1502

``` javascript
opts = this._optionsForExec(model)
```

after which opts.remove is set to a function. I assume, this is some sort of getter function from the model?

I've added a conditional overwriting of opts.remove in case that opts.upsert resolves to "true"

mongoose/lib/query.js: Line 1508

``` javascript
if (!('upsert' in opts)) opts.upsert = false;
if (opts.upsert) {
  opts.remove = false;
}
```

which solved my problem.
However, I don't know if this will have any side-effects.

Verified, that the same issue exists on 3.8.20

vkarpov15:
@benkroeger this change went in to master branch, and so it's in the unstable 3.9.7 release. I won't merge master into 3.8.x. However, I can backport this fix in for 3.8.22, thanks for reminding me.

PR #2598, opened 2015-01-15
fixing conflict in findAndModify --> follow-up to #2519

benkroeger:
disabling opts.remove when opts.new resolves to true.

This resolves a similar issue to the one discussed in #2519 but related to the "new" property.


benkroeger:
Just a gentle reminder. Would be nice to get this fix brought into a release soon... since I have to patch that manually in every deployment right now.

vkarpov15:
Will do. There's a merge conflict but I'll make this change for rc2

PR #2615, opened 2015-01-26
unnecessary conditions

danielrohers:


lib/query.js
```diff
@@ -1127,9 +1123,7 @@ Query.prototype.findOne = function (conditions, fields, options, callback) {
 
   if (options) {
     this.setOptions(options);
-  }
-
-  if (fields) {
+  } else if (fields) {
```
vkarpov15: This isn't necessarily correct, both fields and options can be defined. You're right about the other two ifs being redundant though. If you can get rid of this change I can merge this PR

danielrohers: Really this case I'm wrong. But the other two ifs are redundant.

vkarpov15: Yeah the other two ifs are redundant, you're right. Just remove the L-1126 change and this is good :)

danielrohers:
@vkarpov15 ready!!

PR #2623, opened 2015-01-27
make model.hydrate use createModel so that it works with discriminators

ashaffer:
It seems like hydrate should work with discriminators.  Also i'm not sure why, but if I require queryhelpers outside of the function it breaks several tests, which is why i'm doing it in the awkward way that I am.  Not sure if there is a workaround for that.


vkarpov15:
Seems legit. Thanks for the contribution.

Issue #2617, opened 2015-01-27
mongoose runs validators more than once on save of a nested document


ashaffer:
``` javascript
var mongoose = require('mongoose');
var Schema = mongoose.Schema;

mongoose.connect('localhost');

var SubSchema = new Schema({str: String});
SubSchema.path('str').validate(function(val, next) {
  console.log('path.validate');
  next();
});

SubSchema.pre('validate', function(next) {
  console.log('pre validate');
  next();
});

var DocSchema = new Schema({arr: [SubSchema]});
var Doc = mongoose.model('b', DocSchema);

var p = new Doc();
p.arr.push({str: 'asdf'});
p.save();
```

Run this code.  You would expect each console.log to happen once, but they both happen twice in the latest version of mongoose and at least as far back as 3.8.8, the version i'm using.

ashaffer:
Nvm, this issue seems to be fixed in the unstable version in the repository.  I was only testing against the most recently released version.

ashaffer:
Reopening because I realized that this is not fixed in the latest version of 3.8, and that branch is still being updated with bugfixes.

vkarpov15:
Works fine in 4.9.4

troywweber7:
I'm still seeing this happen in `mongoose@5.3.13`.

PR #2624, opened 2015-01-27
Dont run validators more than once for nested subdocuments

ashaffer:
Fix for #2617.  I had to change the 'nested subdocuments' test a little bit, because it was asserting that mongoose not return all of the errors that were actually wrong with the document, which didn't seem like the right behavior to me, but it is possibly a breaking change for code that depends on that behavior.  Not sure if it was actually intended to work that way though.


ashaffer:
Closing, created pull request incorrectly.  Will reopen a new one.

PR #2625, opened 2015-01-27
dont run validators more than once on nested subdocs

ashaffer:
Fix for #2617.  I had to change the 'nested subdocuments' test a little bit, because it was asserting that mongoose not return all of the errors that were actually wrong with the document, which didn't seem like the right behavior to me, but it is possibly a breaking change for code that depends on that behavior.  Not sure if it was actually intended to work that way though.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1275,7 +1275,7 @@ Document.prototype.$__registerHooks = function () {
 
   // Don't duplicate calls to validate() on sub-documents
   // validate already descends the object tree on its own
-  if(! this.parent || ! this.parent()) {
+  if(! this.__parent) {
     this.pre('save', function validation (next) {
       return this.validate(next);
     }, emitErrorOnModel);
--- a/lib/document.js
+++ b/lib/document.js
@@ -639,7 +639,7 @@ Document.prototype.$__set = function (
     } else {
       if (obj[parts[i]] && 'Object' === obj[parts[i]].constructor.name) {
         obj = obj[parts[i]];
-      } else if (obj[parts[i]] && 'EmbeddedDocument' === obj[parts[i]].constructor.name) {  
+      } else if (obj[parts[i]] && 'EmbeddedDocument' === obj[parts[i]].constructor.name) {
         obj = obj[parts[i]];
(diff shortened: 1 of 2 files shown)


ashaffer:
Closing/reopening to force rebuild in travis, since it seems like the build failed for reasons unrelated to my changes.

ashaffer:
I'm not sure what to do about the remaining two tests.  The failures seem to be side-effects of the change in evaluation order, but haven't actually broken the behavior these tests are trying to test.  Anyone have any thoughts on this?

vkarpov15:
Thanks for opening this, I'll review a little later today, this is a sizable bit of work.

lib/document.js
```diff
@@ -1258,6 +1258,29 @@ Document.prototype.$__registerHooks = function () {
 
   DocumentArray || (DocumentArray = require('./types/documentarray'));
 
+  function emitErrorOnModel(err) {
```
vkarpov15: use `var emitErrorOnModel = function(err)` instead

lib/document.js
```diff
@@ -1258,6 +1258,29 @@ Document.prototype.$__registerHooks = function () {
...
+                  + 'Consider adding an error listener to your connection.\n'
+                  + err.stack
+      }
+      this.db.emit('error', err);
+    }
+  }
+
+  // Don't duplicate calls to validate() on sub-documents
+  // validate already descends the object tree on its own
+  if(! this.__parent) {
```
vkarpov15: This screws up order and make validation happen before checking for cast errors, hence why the travis build failed. Why not just add the `if (!this.__parent)` check on L-1337 on the existing code? Would make life much easier.

vkarpov15:
Couple comments. I think this can go into 3.8.x, but as long as the tests that are currently failing in travis succeed.

lib/document.js
ashaffer: I did that initially, but that causes the 'document: hooks: pre save hooks on sub-docs should not exec after validation errors' test to fail, because then validation errors do not prevent the execution of other pre-save hooks.

vkarpov15: Then I don't think this is going to work in 3.8.x unfortunately. However, master passes the test you specified :)

PR #2630, opened 2015-01-29
To use the transform function only for those documents to which it refers.

chetverikov:
```javascript

    var schema = new Schema({
      name: String,
      places: [{type: ObjectId, ref: 'Places'}]
    });

    var schemaPlaces = new Schema({
      identity: String
    });

    schemaPlaces.set('toObject', {
      transform: function (doc, ret) {

        // here should be only Places documents
        assert.equal(doc.constructor.modelName, 'Places');

        return ret;
      }
    });

    var db = start()
      , Test = db.model('Foo', schema)
      , Places = db.model('Places', schemaPlaces);

    Places.create({ identity: 'a' },{ identity: 'b' },{ identity: 'c' }, function(err, a, b, c){
      Test.create({ name: 'chetverikov', places: [a, b, c]}, function( err ){
        assert.ifError(err);
        Test.findOne({}).populate('places').exec(function( err, docs ){
          assert.ifError(err);

          docs.toObject({transform: true});

          done();
        });
      });
    });
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ docs/source/_docs
 docs/*.html
 tags
 test/triage/*.js
+.idea/
--- a/History.md
+++ b/History.md
@@ -1,3 +1,10 @@
+3.8.17 / 2014-09-29
+==================
+ * fixed; use schema options retainKeyOrder in save() #2274
+ * fixed; fix skip in populate when limit is set #2252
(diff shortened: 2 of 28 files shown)


vkarpov15:
Hmm as far as I can tell, your test passes on master @chetverikov

chetverikov:
Recreate #2639

PR #2668, opened 2015-02-11
Make defaulted paths as modified, on existing documents

### LINKED ISSUE #2558: default values are not saved when added to existing documents
If default values are added to your schema after a document has been created, updating your document without manually touching the property will result in the document being saved without the default value. 

Steps to reproduce
1. Create a test schema and save a few records
2. Update the test schema and add a default of Date.now
3. Write a script to pull a record. If you were to console.log this record, you'll actually see that the default is kicking in and the Date.now is populated
4. perform record.save( function(err,saved){ console.log(saved)}). Saved contains the default of Date.now. 
5. Check your db, and look for Date.now. You'll see that the property still does not exist. 

I think this is a pretty bad bug as it shows the property being populated before and after the record has been saved.

jondavidjohn:
This should fix #2558

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -209,6 +209,7 @@ Document.prototype.$__buildDoc = function (obj, fields, skipId) {
             if ('undefined' !== typeof def) {
               doc_[piece] = def;
               self.$__.activePaths.default(p);
+              self.$__.activePaths.modify(p);
             }
 
           } else if (p in fields) {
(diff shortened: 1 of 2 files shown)


jondavidjohn:
Well, this apparently breaks a lot of other stuff and I'm having trouble coming up with a fix at the moment, but I think this makes the point.  We probably want fields populated from the schema defaults to persist when saved right?

vkarpov15:
I see the general idea and thanks for the work thus far. I agree that those defaults should persist, but I unfortunately can't merge this PR with so many broken tests :(

jondavidjohn:
Of course, any general direction or idea as to why it's breaking the other tests?

vkarpov15:
I suspect there are better ways to do this - I'd imagine somewhere there's a check that says "if this is a new document then save defaults, otherwise don't". I'd try removing that logic and seeing what happens.

Issue #851, opened 2012-04-20
consider saving empty objects to the db


aheckmann:
one of the objectives of mongoose has been to minimize the amount of data stored in mongo. ignoring undefined values and empty objects etc.

however this approach is inconsistent. for example, while empty objects are currently ignored, empty arrays are not. we should probably fix this.

the current approach also does not take into consideration other processes querying / updating your dataset which may lead to problems if it doesn't expect this behavior.

in the master branch (which will be mongoose 3.x) we have doc#to{Object,JSON} options which are configurable at the schema level. with this PR we can now set the minimize option to false at the schema level as well

var options = { toJSON: { minimize: false }};
new Schema({ empty: {} }, options);
var o = doc.toJSON();
o.empty == {}

but that will still only impact explicit calls of toObject/toJSON by application logic; mongoose will still send minimized data (non-empty objects) to the db.

_relates to #848_

PR #2711, opened 2015-02-25
describe minimize option on schemas in guide

AndrewSouthpaw:
Schemas have a `minimize` option, default to true, which will remove empty objects, see #851. This feature is not highlighted in the guide. Because such behavior can be unexpected for a typical user, it would be valuable for it to be present in the guide.


vkarpov15:
This is awesome, thanks!

PR #2717, opened 2015-02-26
Support custom error type for validators.

rodriguise:
The mongoose-validator project has been updated to pass a custom error type name to the validator. This will check the old style msg and new type fields. The resulting errors will have the custom type in err.properties.type and err.kind fields.


rodriguise:
Same as https://github.com/LearnBoost/mongoose/pull/2712. But this is on master.

lib/.tern-port
```diff
@@ -0,0 +1 @@
+65242
```
vkarpov15: Delete this file

lib/schematype.js
```diff
@@ -451,7 +451,7 @@ SchemaType.prototype.validate = function (obj, message, type) {
 
       throw new Error(msg);
     }
-    this.validate(arg.validator, arg);
+    this.validate(arg.validator, arg.msg || arg, arg.type || undefined);
```
vkarpov15: This is causing an obvious test failure on Travis - don't think you can implement this in exactly the same way as you did in 3.8.x because we did [some internal refactoring of validators](http://thecodebarbarian.com/2014/12/19/mongoose-397).

rodriguise:
Resolved with an update to mongoose-validators.

PR #2727, opened 2015-03-02
Added "instance" property strings for Array, Boolean, Date, and Mixed schema types

csdco:
The `Array`, `Boolean`, `Date`, and `Mixed` schema types don't define an `instance` property in `SchemaType.call(...)`, thus resulting in `undefined` being set when `SchemaType` sets `this.instance` in lib/schematype.js:

    function SchemaType (path, options, instance) {
      this.path = path;
      this.instance = instance;

See previous conversations requesting standard behavior at:

 - https://github.com/LearnBoost/mongoose/issues/1430
 - https://github.com/LearnBoost/mongoose/issues/1938

This PR is very simple, it just provides these strings so that anyone loading a Mongoose model can check the instance's schema type. This is useful, for example, if one wishes to load Mongoose models and describe their schema programmatically. Even without this use case, though, I think that the `instance` being set should be normalized.

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -61,7 +61,7 @@ function SchemaArray (key, cast, options) {
     }
   }
 
-  SchemaType.call(this, key, options);
+  SchemaType.call(this, key, options, 'Array');
 
   var self = this
     , defaultArr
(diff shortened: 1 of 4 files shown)


vkarpov15:
Thanks!

formula1:
^__^

Issue #2729, opened 2015-03-03
CastError when populating schema with reference saved as {}


aaunel:
`Population: CastError: Cast to ObjectId failed for value “[object Object]” at path “_id”`

Running into a `CastError` in Mongoose when populating nested ObjectId references (value `{}`) that are apparently _valid_, to the extent that they're not blocked when saving to schema.

Interested in resolving this on the server-side to prevent malformed data in future, however, I understand not saving these values from the client would be a good idea. The fact that an empty object _can_ be saved to a Mongoose schema with `type: mongoose.Schema.Types.ObjectId` and later throws a `CastError` using population is my primary concern.

Originated from [StackOverflow](http://stackoverflow.com/questions/28705810/mongoose-population-casterror-cast-to-objectid-failed-for-value-object-objec)

vkarpov15:
Hmm actually @aaunel, as far as I can tell, mongoose is failing to save `{}`. The CastError is happening when you call `create()` in the test case, not when you call `findById`, which is what I'd expect: `{}` is not a valid ObjectId. In other words, I'm not sure how you managed to save `{}` in an ObjectId field, mongoose doesn't allow that. Can you verify that this is the behavior you're seeing?

aaunel:
I've replicated the issue with test cases (and now fixed) in the pull request referenced above PR #2732

PR #2730, opened 2015-03-03
Created test case to replicate Issue #2729

aaunel:
See Issue #2729 

Added two test cases to `model.populate`:

- Population does not fail on document with ref of `null`.
- Population **does** fail on document with ref of `{}` - CastError.

### DIFF
--- a/test/model.populate.test.js
+++ b/test/model.populate.test.js
@@ -125,6 +125,50 @@ describe('model: populate:', function(){
     });
   });
 
+  it('not failing on null as ref', function(done) {
+    var db = start()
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks for the test cases, they'll be very handy in reproducing and fixing :)

aaunel:
Cool, yes a test case for this was probably the best way to communicate / replicate the bug. If I get a chance this week, I'll start making an attempt to patch the issue, however, I'm sure someone more familiar with the working codebase would be able to resolve this faster.

PR #2732, opened 2015-03-04
Added fix to model.populate() for Issue #2729

aaunel:
See Issue #2729, PR #2730 

The test case in my previous PR was actually failing correctly, where the offending `{}` as ObjectId ref would be caught when validating on `model.create()`. To correctly replicate the error this raises, I've added a layer to negotiate the validator using `model.findByIdAndUpdate()` within the `model.create()` callback.

This possibly raises a general issue that SchemaTypes are not being validated effectively on `model.findByIdAnd..` methods, causing malformed data to be stored?


aaunel:
Investigated this further and found the method used in populate to parse ObjectIds from the ModelMap. Added a check here to ensure `{}` is not passed to the population query as an incomplete ObjectId reference.

This value being persisted by negotiating validation of the ObjectId SchemaType using `model.findByIdAnd..` methods is still outstanding, however, I understand these methods ignore validation rules set in Mongoose. At least this case now completes gracefully.

vkarpov15:
See my PR #2748. The issue is that `findByIdAndUpdate()` was not throwing a CastError when it should . Mongoose shouldn't allow you to save that. But, I'd argue that its not correct to ignore cast errors like that. If the server has something that mongoose can't cast to an objectid, it should error out.

aaunel:
@vkarpov15 yes, certainly agreed, I had otherwise assumed that this scenario using `findByIdAndUpdate()` was a flaw by design, the same way these methods ignore custom validation in schemas, but this is definitely the preferable solution - glad this helped to get that fixed

vkarpov15:
Query functions cast, but don't do custom validation unless you're using 4.x and specify the `runValidators` option. The problem was that this casting logic wasn't handling `$set` properly :)

aaunel:
Oh interesting, `runValidators`! I heard it here first, look forward to upgrading to 4.x once it's stable, then. Thanks for clarifying.

PR #2765, opened 2015-03-16
Update api.html

cdelauder:
Fixed a type on where distinct was spelled "distict".

doh


vkarpov15:
Thanks for pointing this out. api.html is generated code. The corresponding raw text is somewhere in the lib directory - if you can find it I'll be happy to merge the PR :)

cdelauder:
Thanks for the tip on where to find it!

PR #2773, opened 2015-03-19
Fix #2719

### LINKED ISSUE #2719: .populate() does not work from a base schema, using discriminators, when the related collection is defined in the polymorphic child.
**mongoose version**: "^3.8.19"
**related to**: https://github.com/LearnBoost/mongoose/pull/1647

When you use discriminators, and want to query from the base schema, populating some related collection, which was defined in the polymorphic child schema, the query runs, but the result goes without the related population objects.

Example: 

```javascript
var mongoose = require('mongoose')
  , Schema = mongoose.Schema
  , util = require('util')
;

function BaseSchema() {
  Schema.apply(this, arguments);
  this.add( { name: String } );
}

util.inherits(BaseSchema, Schema);

var EventSchema = new BaseSchema();

var ConversionEvent = new BaseSchema({
  owner: { type: Schema.Types.ObjectId, ref: 'User', required: true }
});

var Base = mongoose.model('Event', EventSchema),
var Conversion = Base.discriminator('Conversion', ConversionEvent);

Conversion.create({
  name: 'test',
  owner: user._id
}, function (err, conversion) {

  Base.find().populate('owner').exec(function (err, bases) {
    // Owner object isen't attached
    console.log(bases); 
  });

});
```

chetverikov:
Test include +)


vkarpov15:
You're a champ, thanks.

Issue #2782, opened 2015-03-23
document#set(obj) doesn't work if obj is a sub doc


artiifix:
My schema is similar to the following:
```js
let schema = new mongoose.schema({
  confirmation: {
    data: {
      email: String
    }
  }
});
```

I also have a User model, which is updated with data from the above schema:
```js
user.set(this.get('confirmation.data')); // `user.email` won't be changed
```
But this doesn't work (there is now change to the user document). Meanwhile, if make `confirmation.data` of type Mixed, it works:
```js
let schema = new mongoose.schema({
  confirmation: {
    data: mongoose.Schema.Types.Mixed
  }
});
// ...
console.log(this.get('confirmation.data')); // { email: 'some@example.com' }
user.set(this.get('confirmation.data')); // This works just fine
```
Another strange behaviour: Setting strict mode to false leads to none of the above examples working:
```js
let schema = new mongoose.schema({
  confirmation: {
    data: mongoose.Schema.Types.Mixed
  }
});
user.set(this.get('confirmation.data'), { strict: false }); // This won't work
```
By the way, I'm using version `4.0.0-rc4`.
Should I write a test for it? Maybe, I could also try to work out a fix if you want me to.

vkarpov15:
What does your user schema look like? A test would be helpful.

vkarpov15:
So I pushed a fix for the "won't set data from a sub doc" test, that was legitimately a bug.

"won't set data from an object of type Mixed when in non-strict mode" is not an issue with mongoose - the correct way to do that would be `doc2.set(doc3.data, undefined, { strict: false });`. Note the 'undefined' as the 2nd parameter. That's because technically that is not a publicly supported syntax for `doc2.set()`, it's used as an internal hack, but it should work for the foreseeable future.

PR #2785, opened 2015-03-24
Add a test for #2782

artiifix:
For me, only the second test passes at the moment.


vkarpov15:
Thanks! I'll investigate this later this week.

PR #2827, opened 2015-04-01
Update docs for Model.update to describe new signature

irnc:
As was noted in https://github.com/Automattic/mongoose/wiki/4.0-Release-Notes the signature of a callback was changed to return raw response from update command.


irnc:
Sorry, this pull request isn't ready for merge yet; added `@see` tag doesn't render correctly by `dox`.

irnc:
After latest commit, the link in `@see` tag would render properly in used version of `dox`.

Issue #2833, opened 2015-04-02
type Date not checking for undefined causing generic failure


winfinit:
if field is defined in schema with type of Date, and value is passed for that fields as undefined, failure doesn't propagate to CastError, instead failing with TypeError when toString method is attempted against undefined value, causing obscure error as such:

```
TypeError: Cannot read property 'toString' of undefined)
```

instead of expected:

```
CastError: Cast to date failed for value "undefined" at path "myFieldName"
```

winfinit:
pull request submitted:
https://github.com/Automattic/mongoose/pull/2834

winfinit:
snippet reproducing this issue:

```javascript
var mongoose = require('mongoose');
var mySchema = mongoose.Schema({date: { type: Date }});
var myModel = mongoose.model('mySchema', mySchema);

var myObject = new myModel();
myObject.update({date: undefined});
```

output:
```
TypeError: Cannot read property 'toString' of undefined
...
```

new output with provided patch
```
CastError: Cast to date failed for value "undefined" at path "date"
...
```

vkarpov15:
This only affects 3.8.x, 4.0 fixes this thanks to #2272, so #2834 is unnecessary. I'll backport it to 3.8 though

Issue #2272, opened 2014-08-31
findAndUpdate fails when a field is undefined


iliakan:
The code sample:
```js
yield this.findOneAndUpdate(
      { key: doc.key },
      // $set every field of the document (to fully replace, not update)
      // setting to undefined doesn't work here (mongoose bug?)
      {
        key: doc.key,
        tags: doc.tags || [],
        value: value,
        expireAt: undefined,
        generatingStartTimestamp: null,
        generatingTimeLimit: null
      }
    ).exec();
```

The problem is `expireAt: undefined`. I expect it to remove the field from the document.
But instead I get:
```
     TypeError: Cannot read property 'toString' of undefined
      at SchemaDate.cast (/js/javascript-nodejs/node_modules/mongoose/lib/schema/date.js:101:17)
      at SchemaDate.castForQuery (/js/javascript-nodejs/node_modules/mongoose/lib/schema/date.js:151:17)
      at Query._castUpdateVal (/js/javascript-nodejs/node_modules/mongoose/lib/query.js:2031:17)
      at Query._walkUpdatePath (/js/javascript-nodejs/node_modules/mongoose/lib/query.js:1980:25)
      at Query._castUpdate (/js/javascript-nodejs/node_modules/mongoose/lib/query.js:1873:23)
      at castDoc (/js/javascript-nodejs/node_modules/mongoose/lib/query.js:2054:18)
      at Query._findAndModify (/js/javascript-nodejs/node_modules/mongoose/lib/query.js:1510:17)
      at Query.findOneAndUpdate (/js/javascript-nodejs/node_modules/mongoose/node_modules/mquery/lib/mquery.js:2077:15)
      at Query.exec (/js/javascript-nodejs/node_modules/mongoose/node_modules/mquery/lib/mquery.js:2213:16)
      at Query.exec (/js/javascript-nodejs/node_modules/mongoose/lib/query.js:1778:19)
```

PR #2834, opened 2015-04-02
#2833 added condition to check for undefined value

winfinit:
wrapped timestamp support around IF to check for undefined values

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Mongoose
+# Mongoose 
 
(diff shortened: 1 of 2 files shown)


winfinit:
failure that was reported by travis before, was not attributed to this change.

vkarpov15:
This issue doesn't affect master, see L209 in schema/date.js and #2272. Also please don't do things like 6aee997, that's entirely unnecessary. Tests are somewhat flakey on travis, no need to retrigger with another unnecessary commit.

PR #2856, opened 2015-04-09
added a schema.construct(fn) method to create a 'constructor' function ...

derickbailey:
In version 3.8.x of Mongoose, I used the https://github.com/ilsken/mongoose-construct plugin to create a "construct" method on my models. I need this functionality due to some complexities in my models that require code to run after creating a my model instance, but prior to any other code running. This plugin worked great to solve my problem.

With v4 of mongoose, however, this plugin no longer works.

I wanted to open a dialog around creating this functionality inside of mongoose directly, and I thought it would be easier to express my thoughts and intent via a pull request with code changes (and tests) rather than just describing the API.

The changes I'm making here are to replace that plugin with functionality built in to Mongoose, directly. 

The general design of the API is:

```js
var MySchema = mongoose.Schema({
  foo: { type: String }
 // ...
});

MySchema.construct(function(){
  // "this" is the model instance
  console.log(this.foo);
});

var MyModel = mongoose.Model("my-model", MySchema);

var m = new MyModel({
  foo: "this is the foo"
});

// output of new MyModel is to log this.foo
// => this is the foo
```

There can be only one (insert Highlander gif here :D) construct function per schema, since this is intended to be a 'constructor' function for the model.

I'm open to any / all feedback and questions about this API design. I made some very specific design decisions in this, to avoid clobbering any existing code and ensure that this worked as smoothly as possible. I recognize that my design ideals may not be the same as the Mongoose team, though, and I'm certainly open to changing this further, to make sure the changes fit.

In the end, though, I need this functionality to be able to complete my upgrade to mongoose v4. I intend to back-port these changes in to the stable release of v4.x, in my own branch, so I can use this immediately.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2724,6 +2724,9 @@ Model.compile = function compile (name, schema, collectionName, connection, base
     if (!(this instanceof model))
       return new model(doc, fields, skipId);
     Model.call(this, doc, fields, skipId);
+    if (schema.construct && (typeof schema.construct === "function")){
+      schema.construct.call(this);
+    }
   };
 
   model.hooks = schema.s.hooks.clone();
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -305,6 +305,7 @@ Schema.prototype.add = function add (obj, prefix) {
 
 Schema.reserved = Object.create(null);
 var reserved = Schema.reserved;
+reserved.construct =
 reserved.emit =
 reserved.on =
 reserved.once = // EventEmitter
@@ -670,6 +671,31 @@ Schema.prototype.post = function(method, fn) {
 
 Schema.prototype.plugin = function (fn, opts) {
   fn(this, opts);
+  return this;
+};
+
+/**
+ * Adds a function that is called immediately after constructing a model from this schema. There can be only one 'construct' per schema definition.
+ *
+ * ####Example
+ *
+ *     var schema = kittySchema = new Schema(..);
+ *
+ *     schema.construct(function () {
+ *       console.log('meeeeeoooooooooooow');
+ *     })
+ *
+ *     var Kitty = mongoose.model('Kitty', schema);
+ *
+ *     var fizz = new Kitty(); // meeeeeooooooooooooow
+ *
+ *
+ * @param {Function} [fn]
+ * @api public
+ */
+
+Schema.prototype.construct = function (fn) {
+  this.construct = fn;
   return this;
 };
 
--- a/test/model.construct.test.js
+++ b/test/model.construct.test.js
@@ -18,7 +18,7 @@ var schemaB = Schema({
 });
 
 describe('model', function(){
-  describe.only('construct', function(){
+  describe('construct', function(){
     var db;
     var B;
     var calledConstruct, constructCtx;
--- a/test/model.construct.test.js
+++ b/test/model.construct.test.js
@@ -1,0 +1,54 @@
+/**
+ * Test dependencies.
+ */
+
+var start = require('./common')
+  , assert = require('assert')
+  , mongoose = start.mongoose
+  , random = require('../lib/utils').random
+  , Schema = mongoose.Schema
+  , DocumentObjectId = mongoose.Types.ObjectId
+
+/**
+ * Setup
+ */
+
+var schemaB = Schema({
+    title: String
+});
+
+describe('model', function(){
+  describe.only('construct', function(){
+    var db;
+    var B;
+    var calledConstruct, constructCtx;
+
+    before(function(){
+      db = start();
+
+      schemaB.construct(function(){
+        calledConstruct = true;
+        constructCtx = this;
+      });
+
+      B = db.model('model-create', schemaB, 'gh-2637-1');
+    })
+
+    after(function(done){
+      db.close(done);
+    })
+
+    it('calls the schema construct method', function(done) {
+      var b = new B();
+      assert.equal(calledConstruct, true);
+      done();
+    });
+
+    it('calls the schema construct with the model instance as the context', function(done) {
+      var b = new B();
+      assert.equal(constructCtx, b);
+      done();
+    });
+  });
+})
+


derickbailey:
FWIW, as an example of how I am using this now:

```js
ScheduleSchema.construct(function(next){
  this._jobRestarted = this._jobRestarted.bind(this)
  this._jobResolved = this._jobResolved.bind(this)
  this._jobComplete = this._jobComplete.bind(this)
  this._stepReady = this._stepReady.bind(this)
  this._releaseStep = this._releaseStep.bind(this)
  this._jobStatusUpdated = this._jobStatusUpdated.bind(this)

  this.limiter = new Limiter();

  this.nextJob = new qdot.Stack();
  this.nonConcurrentKeys = {};
});
```

There are several placed in my code where I use methods as named callbacks. Having the `.bind` calls ensures that these callbacks are always running with the model instance as `this` within those methods.

Additionally, I have code at the bottom of this constructor that sets up in-memory objects, arrays, etc. that I need within my object. These are temporal bits that should not be saved to the database, but get reconstructed and used when the document is pulled from the database. 

Hopefully this will shed some light on why i need this feature.

vkarpov15:
As far as I can tell, schema methods are always bound to the model instance, so there's no need to do that. "These are temporal bits that should not be saved to the database, but get reconstructed and used when the document is pulled from the database." sounds like a textbook definition of virtuals. Furthermore, this .construct() function sounds exactly like a `post('init')` hook, no?

Also, mongoose-construct is 8 SLOC, should be reasonably easy to make a fix, no?

Finally, since mongoose is now on semver, I can't merge this until at least v5.0.0 because it's backwards breaking.

derickbailey:
> this .construct() function sounds exactly like a post('init') hook, no?

no.

the post init hook only works when a model hydrates from a database. it does not run when creating a new model instance, in-memory using the `new MyModel` syntax.

> since mongoose is now on semver, I can't merge this until at least v5.0.0 because it's backwards breaking.

what does this break? it is net new code that doesn't affect anything, as far as I can tell. i'd like to know what it breaks.

> schema methods are always bound to the model instance, so there's no need to do that.

I'll write a test to see if this is true or not. perhaps v4 of mongoose changed this, but in v3, using a schema method as a callback would set "this" to undefined.

> sounds like a textbook definition of virtuals 

possibly... can i assign a virtual to a new object instance, for each instance of my model? i had not thought of trying this. i'll see what happens.

derickbailey:
FWIW, I wrote a test that shows methods are not implicitly bound to the model.

```js
  describe('method binding', function(){
    var db, B, ctx;

    before(function(){
      db = start();

      schemaB.method("foo", function(){
        ctx = this;
      });

      B = db.model('model-create', schemaB, 'gh-2637-1');
    })

    after(function(done){
      db.close(done);
    })

    it('calls the schema construct method', function(done) {
      var b = new B();

      // using b.foo as a callback method
      // the foo method will no longer be bound to the b instance
      b.on("doFoo", b.foo);

      b.emit("doFoo");
      assert.equal(b, ctx);
    });
  });
```

the output of the above test is:

```
 0 passing (479ms)
  1 failing

  1) model method binding calls the schema construct method:
     AssertionError: { _id: 55271fd434a320d611a23c4d } == { domain: null,
  _events: { doFoo: [Function] },
  _maxListeners: 0 }
```

In order to correct this, I have to `.bind` the foo method:

```js
    it('calls the schema construct method', function(done) {
      var b = new B();

      // using b.foo as a callback method, using .bind
      // the foo method is now bound to the b instance
      b.on("doFoo", b.foo.bind(b));

      b.emit("doFoo");
      assert.equal(b, ctx);
    });
```

but calling `.bind` creates a new function instance every time. If i wanted to unbind this event, i would not be able to do do:

```js
b.on("foo", b.foo.bind(b));

b.removeListener("doFoo", b.foo)
```

this does not unbind the listener, because `b.foo` is not the function that was passed in as the listener. that's why i do the `this.foo = this.foo.bind(this)` in my construct call... so that my events and other callbacks will work properly, and be removable from event handlers.

derickbailey:
I've also written a test that shows how `post('init')` does not get called when creating a new model in memory.

```js
  describe('model post init', function(){
    var db, B, postInitWasCalled = false;

    before(function(){
      db = start();

      schemaB.post("init", function(){
        postInitWasCalled = true;
      });

      B = db.model('model-create', schemaB, 'gh-2637-1');
    })

    after(function(done){
      db.close(done);
    })

    it('calls the schema construct method', function(done) {
      var b = new B();
      assert.equal(postInitWasCalled, true);
    });
  });
```

the output of this test:

```
 0 passing (359ms)
  1 failing

  1) model model post init calls the schema construct method:
     AssertionError: false == true
```

vkarpov15:
What it breaks is if somebody has a schema property named "construct". Not huge, but enough to avoid IMO. I see what you mean re: using methods as event handler callbacks though, that's pretty strange.

Still, my preferred solution would be to fix mongoose-construct and/or mongoose in order to make mongoose-construct work properly. As far as I can tell there shouldn't be any backwards breaking changes that make mongoose-construct no longer work, but I'm clearly wrong on that front. I'll see if I can make some time to take a look at mongoose-construct over the next few days.

Thanks for your detailed feedback, btw. Very useful for pointing out how the module can be better :)

derickbailey:
> What it breaks is if somebody has a schema property named "construct". 

ah, right. i see that now. i'll take a look at this in the morning and see if there's a way around that.

> Still, my preferred solution would be to fix mongoose-construct and/or mongoose in order to make mongoose-construct work properly. 

that would work, too! i was hoping to see something like this become part of mongoose directly, so that it wouldn't require a plugin... but i'm ok with the plugin, if it continues to work. 

> Thanks for your detailed feedback, btw.

i'm glad it was helpful!

I've run a few open source projects in my time, and i hope i've lived up to my own expectations of contributors and feedback. :)

vkarpov15:
Yeah I've seen migroose - some good stuff there.

This is something to investigate adding to core in the future, but right now best suited as a plugin IMO.

vkarpov15:
@derickbailey so I've done some investigating and there are a couple issues:

1) mongoose-construct is using internal mongoose functionality badly
2) some work to support promises with hooks broke the schema call queue

Thankfully, both of these problems have a simple solution. First, mongoose-construct is overly verbose, the below code will work in 3.8.x and >= 4.0.2 (with a little help from #2873):

```javascript
mongoose.set('debug', true);
var construct = require('mongoose-construct');

mongoose.connect('mongodb://localhost:27017/gh2856');

var MySchema = mongoose.Schema({
  foo: { type: String }
});

MySchema.methods.construct = function() {
  console.log(this.foo);
};

// Queue up the above `construct()` method to be called
MySchema.queue('construct');

var MyModel = mongoose.model("my-model", MySchema);

var m = new MyModel({
  foo: "this is the foo"
});

console.log('Done!');
```

derickbailey:
that looks great! :) i'll be watching for the 4.0.2 release, and putting this in my code once it's out.

thanks for all the feedback, R&D and solution on this @vkarpov15

Issue #2879, opened 2015-04-16
broken since 3.8.15: enumerating `virtual` property with nested objects results in crash from `undefined` object


toblerpwn:
When using a `virtual` extension on a schema that returns an object with nested objects (imagine an exact copy of a schema with some fields stripped, e.g. `user.public_user_document`), operations such as `JSON.stringify()` and `util.inspect()` are unable to enumerate the virtual property object.

This was working prior to Mongoose 3.8.15; it was broken by this PR #2211 - see the diff:

https://github.com/Automattic/mongoose/pull/2211/files

The specific line that breaks this functionality is line `lib/document.js:1154` in the PR - specifically setting `enumerable: false` (setting to `true` will resolve the crash), but I trust that the nested objects created in #2211 should be enumerable, so it is probably not appropriate to simply undo the change.

I encountered this while migrating to MongoDB 3.0.1, but I suspect this would exist on MongoDB 2.6.x as well.

Here is an example stack trace of the crash from release 3.8.26:

```
TypeError: Cannot read property 'adhocPaths' of undefined
    at Object.Document.$__path (/path/to/project/node_modules/mongoose/lib/document.js:733:24)
    at Object.Document.get (/path/to/project/node_modules/mongoose/lib/document.js:706:21)
    at Object.nested.toObject (/path/to/project/node_modules/mongoose/lib/document.js:1214:27)
    at Object.Document.inspect (/path/to/project/node_modules/mongoose/lib/document.js:1679:23)
    at model.userSchema.static.User (/path/to/project/lib/models/user.js:304:33) // our file
    at VirtualType.applyGetters (/path/to/project/node_modules/mongoose/lib/virtualtype.js:77:25)
    at model.Document.get (/path/to/project/node_modules/mongoose/lib/document.js:717:18)
    at model.Object.defineProperty.get [as public_user] (/path/to/project/node_modules/mongoose/lib/document.js:1234:45) // our `virtual` property
    at Promise.<anonymous> (/path/to/project/lib/api_versions/version1_actions.js:119:33)  // our file
    at Promise.<anonymous> (/path/to/project/node_modules/mongoose/node_modules/mpromise/lib/promise.js:177:8)
```

PR #2880, opened 2015-04-16
Fix for #2879

toblerpwn:
This PR fixes Issue #2879 by allowing self-owned nested properties of virtual objects (or any objects?) to be enumerable. (There is also a minor fix for a case where an undefined property crashes the process.)

Suggest also porting this fix to 3.8.x branch (which we still use in production; will file separate PR if this is approved/etc).

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -781,7 +781,8 @@ Document.prototype.get = function (path, type) {
  */
 
 Document.prototype.$__path = function (path) {
-  var adhocs = this.$__.adhocPaths
+  var base = this.$__
+    , adhocs = (base) ? base.adhocPaths : undefined
     , adhocType = adhocs && adhocs[path];
 
   if (adhocType) {
--- a/lib/document.js
+++ b/lib/document.js
@@ -781,8 +781,7 @@ Document.prototype.get = function (path, type) {
  */
 
 Document.prototype.$__path = function (path) {
-  var base = this.$__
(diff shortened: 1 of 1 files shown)


vkarpov15:
Hmm is a50a8c1 actually necessary? Seems like c575d20 should be sufficient.

toblerpwn:
@vkarpov15 - Yes, that is correct; https://github.com/Automattic/mongoose/commit/a50a8c10dd0774af4197d7c88c0aad56b39367aa was my first step to troubleshoot #2879 in my environment (and it was a partial fix), but, when it became clear  https://github.com/Automattic/mongoose/commit/c575d20f19a95cdc1d940510fa7d6aa461c58698 was also needed (or really ALL that was needed), but I just https://github.com/Automattic/mongoose/commit/a50a8c10dd0774af4197d7c88c0aad56b39367aa in place because it has zero functional impact but a positive stability impact (i.e. for future cases that might encounter this issue).

OR, you could argue that it 'covers up' an expected behavior; in either case I am OK to roll back https://github.com/Automattic/mongoose/commit/a50a8c10dd0774af4197d7c88c0aad56b39367aa if you prefer. Let me know!

vkarpov15:
IMO it's covering up an expected behavior, that case should not be happening because a Document **always** has `$__`.

toblerpwn:
@vkarpov15 - OK, reverted the change.

Issue #2809, opened 2015-03-29
performance issue of 4.0.0&4.0.1


losogithub:
After upgrading from 3.9.* to 4.0.0&4.0.1, my node process's CPU usage increased by more than 200%.
Each http request which uses mongoose costs cpu % 3 times than before.
And when I downgrade back to 3.8.25, everything is OK.

vkarpov15:
I've noticed some performance degradation, but it seems like you have some more in-depth info. I'll investigate but seeing some sample code would really help

lbeschastny:
@losogithub do you see the following warning message when you starting your app?

    Failed to load c++ bson extension, using pure JS version

If you do, then it's not a mongoose issue. And it would explain your performance issues.

AntouanK:
@lbeschastny I get that message https://github.com/Automattic/mongoose/issues/2744#issuecomment-91179107
Any ideas on how to fix that?

paambaati:
@AntouanK Make sure you have all the build tools installed. Which platform are you on? If you're on OS X, that will be `xcode`. If you're on a Linux-based distro, that will be `gcc`, `build` and `make-essential`. Once that is in place, install `node-gyp` and rebuild your modules. That should solve your issue.

losogithub:
@lbeschastny I had fixed the c++ bson extension issue before I encountered this performance issue. Thank u all the same.

AntouanK:
@paambaati I'm on CoreOS but I use docker containers ( it doesn't really matter what OS the container is in ).
I'll try to get `gcc`, `build` and `make-essential` in the container I'm running mongoose in, and see if I get the C++ bindings.

AntouanK:
Turns out I also need the `python` package...

You can check out the `node_modules/mongoose/node_modules/mongodb/node_modules/mongodb-core/node_modules/bson/builderror.log` file to see why build fails.

paambaati:
Oh yeah. I'd recommend 2.7.6 or higher.

AHDPuK:
I confirm that I experience big app slowdown - but without bson warning. 
Ubuntu 14
After downgrade also everything is OK,

vkarpov15:
Can you provide me an example of an operation or block of code that you're experiencing a big performance degradation for? Difficult to debug when I don't have anything to go on :)

[...]

PR #2895, opened 2015-04-17
Fixed performance degradation

lbeschastny:
Fixed performance degradation mentionned in #2809  and caused by @6ce5a6c.

According to by benchmarks, it's not the only performance degradation, though it's the most severe one.

### DIFF
--- a/lib/schema/documentarray.js
+++ b/lib/schema/documentarray.js
@@ -25,16 +25,16 @@ function DocumentArray (key, schema, options) {
 
   // compile an embedded document for this schema
   function EmbeddedDocument () {
-    // apply methods
-    for (var i in schema.methods) {
-      this[i] = schema.methods[i];
-    }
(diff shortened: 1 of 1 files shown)


lbeschastny:
Looks like it was already done in #2863 PR.

Well, at least the most important part.

lbeschastny:
Rebased it to master.

Though, now this PR doesn't make much of a sense, since all important things was already done in #2863.

You could merge it anyway, or you could just close it, doesn't make much of a difference.

lbeschastny:
And what's the problem with travis? It just randomly fails tests with `timeout of 2000ms exceeded` error.

Shouldn't you increase the timeout? Or is there some other problem causing tests to randomly hung?

vkarpov15:
@lbeschastny Travis is pretty flakey, tests time out frequently. Haven't really investigated why, but I'm ok with running tests manually before releasing to double check.

Issue #2911, opened 2015-04-25
Support deep population


timbur:
I realize this has been discussed (at least) a few times before, and it's considered a bit of an anti-pattern, but I think it's something that should be supported out of the box.  I like the simplicity of @buunguyen's [mongoose-deep-populate](https://github.com/buunguyen/mongoose-deep-populate), but the required approach (monkey-patching, etc.) makes further customization and integration with other plugins too difficult and hacky for my tastes.

For example, using the ideas behind Facebook's GraphQL (web apps should know and retrieve exactly the data it needs), a very contrived example of some query structure:
```
Comment {
  for,
  text,
  created,
  createdBy {
    name,
    avatar,
    followers.first(5) {
      name
    }
  }
}
```

And so the app would ideally be able to retrieve everything with a single request:
```
http://localhost/api/comments?conditions[for]=someId&select=for text created createdBy.name createdBy.avatar createdBy.followers.name&limit[createdBy.followers]=5
```

I think mongoose has the potential to shine here.  Ideally, the resulting query on the server might look like:
```js
Comment
  .find({for: someId})
  .select('for text created createdBy')
  .populate({
    path: 'createdBy',
    select: 'name avatar followers',
    populate: [{
      path: 'followers',
      select: 'name',
      options: { limit: 5 }
    }]
  })
  .exec();
```

All that's really needed is for mongoose to support something like this:
```js
query.populate({
  populate: [
    {
      path: ...,
      populate: [...]
    },
    {
      path: ...,
      populate: [...]
    },
    etc.
  ]
});
```

The object passed to `.populate()` should probably be constructed by something else, which would allow other plugins/middleware to alter it as necessary and would also keep things simple for mongoose.  I can't imagine it would be too difficult to implement and I'd be glad to do it myself if there are no objections.  It would certainly help me out a lot to have this in the core of mongoose rather than as some monkey-patched solution!

It's entirely possible I'm just being dumb and blind to a better way to do this.  If that's the case, please let me know.  :)

vkarpov15:
You're welcome to give it a shot, but I imagine it would be fairly tricky. Mongoose does a lot of magic to make populate() behave in a mostly sane way, I doubt it would be a trivial task to properly support deep population.

Also, I worry that enabling this will encourage people to use mongodb for graph-like applications, which isn't really mongodb's sweet spot. In an ideal mongodb setup, you don't use `populate()` at all.

chetverikov:
[...]

PR #2916, opened 2015-04-27
added subPopulate functionality

timbur:
This is for #2911.  It seems to work based on my limited tests.  I'm a git noob and a unit test noob so you may have to give me some time to figure all that out, or maybe nudge me in the right direction for writing proper tests.  :)

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2162,7 +2162,7 @@ Model.populate = function (docs, paths, cb) {
   while (i--) {
     path = paths[i];
     if ('function' === typeof path.model) model = path.model;
-    populate(model, docs, path, next);
+    populate(model, docs, path, subPopulate.call(model, docs, path, next));
   }
 
   return promise;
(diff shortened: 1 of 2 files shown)


timbur:
Forgot to add that the usage matches the example I outlined in the issue.

Basically, just add `populate` option(s) to any `populate()` call:
```js
Comment.populate({
  path: 'createdBy',
  model: 'User',
  select: 'name avatar followers',
  populate: [{
    path: 'followers',
    select: 'name',
    options: { limit: 5 }
  }]
});
```

Or you can go as deep as necessary (though I think it won't populate any circular references):
```js
Comment.populate({
  path: 'createdBy',
  model: 'User',
  select: 'name avatar followers',
  populate: [{
    path: 'followers',
    select: 'name followers',
    options: { limit: 5 },
    populate: {  // can also use a single object instead of array of objects
      path: 'followers',
      select: 'name',
      options: { limit: 2 }
    }
  }]
});
```

And at the application level you might have something that converts a string to the proper `populate` parameters.  So for a real-world use-case, converting `req.query.select` from a URL like `http://localhost/api/comment/123?select=createdBy.name createdBy.followers.name` to the proper `populate` parameters would yield:
```js
Comment.populate({
  path: 'createdBy',
  model: 'User',
  select: 'name followers',
  populate: [{
    path: 'followers',
    select: 'name'
  }]
});
```

`decorators/deep-select.js`:
```js
'use strict';

/**
 * Adds automatic deep populate + select functionality.
 *
 * @api public
 */
exports = module.exports = deepSelect;
function deepSelect () {
  var model = this.model();

  this.request(function (req, res, next) {
    var query = model.find(req.query.conditions);
    var select = req.query.select || '';
    var populate = [];
    var pointers = {};

    select.split(' ').forEach(function (deepPath) {
      gatherPopulateParams(populate, pointers, model, deepPath.split('.'));
    });

    select = Object.keys(pointers).join(' ');
    if (select) {
      query.select(select);
    }
    
    populate.forEach(function (params) {
      query.populate(params);
    });

    query.exec(function (err, docs) {
      res.send(err || docs);
    });
  });
}

/**
 * Gathers the params required for population of some deep path, if any.
 *
 * @param {Array} populate
 * @param {Object} pointers
 * @param {Model} model
 * @param {Array} deepPath
 * @return {Mixed}
 * @api private
 */
function gatherPopulateParams (populate, pointers, model, deepPath) {
  var modelName = model.modelName;
  var paths = model.schema.paths;
  var path = '';
  var options;
  var params;
  var ref;
  
  while (deepPath.length) {
    path += (path ? '.' : '')+deepPath.shift();
    options = paths[path] && (
      paths[path].caster && paths[path].caster.options
      || paths[path].options
    );
    if (!options) {
      continue;
    }

    params = pointers[path] || (pointers[path] = {});
    ref = options.ref || options.type[0] && options.type[0].ref;
    if (!ref) {
      if (params.pointers) {
        params.select = Object.keys(params.pointers).join(' ');
      }
      continue;
    }

    if (!params.path) {
      params.path = path;
      params.model = ref;
      params.populate = [];
      params.pointers = {};
      populate.push(params);
    }
    gatherPopulateParams(
      params.populate,
      params.pointers,
      model.model(ref),
      deepPath
    );
    params.select = Object.keys(params.pointers).join(' ');
  }
}
```

Issue #1323, opened 2013-01-31
Avoid setting 'type' property on errors


hurrymaplelad:
Heads up, CastError sets `error.type` which is apparently frowned upon in V8:
http://code.google.com/p/v8/issues/detail?id=2397

Doesn't cause immediate issues, but if I do something silly like

```
var e = new CastError('a', 'b', 'c');
delete e.message;
e.toString();
```

Things get ugly quick.  I get a cryptic `illegal access`.

PR #2928, opened 2015-04-27
Best way to mitigate invalid ID query breaking change of 4.0?

thanpolas:
The 4.x series brought an unexpected breaking change.

Now when you query based on the record `_id`, the string you are passing has to be a proper Mongo Object Id or look-alike. Anything else will throw a `CastError` vs a null result object that it used to return in the 3.x series.

Needless to say this change breaks an array of tests and existing codebases. What used to return a `null` result now throws an error.

Is there a way to mitigate this issue or are we locked to 3.x until we allocate resources to refactor all the codebase where needed?

### DIFF
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -1501,4 +1501,31 @@ describe('Query', function(){
       });
     });
   });
+
+  describe('Invalid IDs', function () {
+    it('should not throw an error when bogus id using find', function (done) {
+      var db = start();
+      var Product = db.model('Product', 'Product_setOptions_test');
+      Product.find({_id: 'bogus'}).exec(function(error) {
+        assert.ifError(error);
+        db.close(done);
+      });
+    });
+    it('should not throw an error when bogus id using findOne', function (done) {
+      var db = start();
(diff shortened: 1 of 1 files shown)


thanpolas:
I've attached a PR with test cases illustrating the problem @vkarpov15, check the Diff

vkarpov15:
Hmm @thanpolas so as far as I can tell, this behavior is the same in 3.8 - these tests fail for the same reason on 3.8.0, 3.8.15, and 3.8.28, so it doesn't qualify as a backwards-breaking change. I suspect there's another reason why your tests are failing after upgrade - can you tell me what version you're upgrading from and show me the schema that this is failing for?

vkarpov15:
No activity for 2 weeks, re-open if this is still an issue.

thanpolas:
@vkarpov15 i nailed down the issue, I was suppressing those errors based on the now apparently deprecated error attribute `type` which in mongoose 4.0.0 has been renamed to `kind`

vkarpov15:
Yeah that change has bitten a few people unfortunately, glad you figured out what was going on. There's a very good reason for that change, see discussion on #1323.

PR #2941, opened 2015-04-30
Update utils.js

rknell:
Cafe, when pluralised, was being changed to caves


vkarpov15:
Unfortunately, this is a very dangerous change that I can't make without it being backwards breaking. Gonna have to hold off on this until I bump 5.0. In the meantime, there are numerous workarounds:

```javascript
mongoose.model('Cafe', cafeSchema, 'cafes');
```

Issue #2777, opened 2015-03-20
Feature: Helper Method: Model.upsert


justsml:
I think it would make the API more complete if it had a `Model.upsert` helper method ( a simple wrapper for `Model.update(q, obj, {upsert: true}, fn)`

If I'm off-base please let me know, I just thought it was in line with the extended `find()` methods, `findById`, `findByIdAndUpdate`, etc.

justsml:
I've completed a commit for ```Model.upsert``` with additional tests:
https://github.com/justsml/mongoose/tree/feature/upsert
https://github.com/LearnBoost/mongoose/compare/master...justsml:feature/upsert

I think I followed the guide, but this is my first (hopeful) PR for mongoose...
So any feedback === appreciated^2

vkarpov15:
So general feedback - if this is to go in, there should be a function in `lib/query.js` as well so you can chain calls like `model.update().upsert().exec()`. I would also imagine that things like `model.update().upsert(false).exec()` should be valid, so the query function should be able to handle cases where it receives params "[query], [update], [options], [callback]" as well as "[query], [update], [shouldUpsert], [callback]".

Also, if we're going to support chaining with the `upsert()` option, that would also imply that there should be a helper for the `multi()` option as well. The tricky bit with multi is that upsert's meaning is clear, whereas `Model.multi()` or `Query.multi()` can be unclear because the option is 'multi' for update but 'justOne' for remove, and there's no corresponding option for findAndModify...

justsml:
Thanks @vkarpov15 

Just so I know I understand, are you saying it should support modifying the Query chain, switching the upsert flag on or off?

I'll write out the tests and see what I can come up with...

The ```{multi}``` option I've always thought was an awkward design...
If a ```findOne``` method was added, why isn't there a ```updateOne()```

justsml:
Ok, read #803 ... I see... I'll finish the code + tests for upsert

justsml:
I checked in some code, and I'm not sure why I'm having trouble with the ```Query.upsert```

I tried just overriding the ```options.upsert``` after the parameter shifting, and then returning something like ```this.update(criteria, obj, newOpts, callback)```

I also tried returning just ```this``` after calling the update method (with the callback)

I just tried copying the entire Query.update method and just tweaking it -- that got me close, but still returning (null, null) to the callback.

@vkarpov15 I'm sure I'm just mis-reading the functionality or missed a dot somewhere , any insight is really appreciated.

[...]

PR #2976, opened 2015-05-07
Feature: Model.upsert

justsml:
RE: Issue #2777

Please review my changes to the file test/model.update.test.js -- What other criteria should be reasonably tested?

Any feedback is greatly appreciated.


vkarpov15:
Is there still that bug with mquery getting stuck in an infinite loop that you mentioned in #2777?

Issue #2502, opened 2014-12-02
Regression in 3.8.20: methods on BaseClass of discriminator leads to TypeError when calling Object.defineProperty


RBLU:
When switching from 3.8.19 to 3.8.20 we experience a TypeError:

```
/Volumes/SourceDev2/yp/backendlib/node_modules/mongoose/lib/model.js:2581
        Object.defineProperty(model.prototype, _i, {
               ^
TypeError: Cannot redefine property: toJsonConfig
```

This is caused by this commit 0bcd28d39b20cc97152da693c6030994c816ffc8 of @alabid. I am not sure whether this commit introduces a bug or we are just using the Discriminator functionality incorrectly.

In difference to the example in #2167  we are using the "methods" functionality on the Base Class, and are not actually adding a method, just a plain data object. 

```
BaseSchema.methods.toJsonConfig = {
        include: ['attr1', 'attr2', 'attr3']
    };
```

This used to work nicely, but now the new method `model.applyMethods` is first called on the BaseModel and then on the SubModel, so the "methods.toJsonConfig" is defined twice, and according to the JS-Spec Object.defineProperty can only called once when configurable is not set explicitly to true for that property.

RBLU:
I am seeing the same error again now, when I am trying to startup my app with mongoose 4.0.2. Is it possible that this fix ist not in the 4.x branches?

vkarpov15:
It's possible that this fix got lost. Can you provide me some code that reproduces this so I can put it in tests?

PR #2978, opened 2015-05-07
reproduction #2502 in v4.0.x

RBLU:
This reproduces #2502 in v4.0.x

I am still not sure whether this is actually allowed (assigning objects to documents.methods) . We just have been doing this for ages. There is an easy workaround: just define methods that return the objects, so if you @vkarpov15 say, that this is not a valid usage, just close the issue and PR.


vkarpov15:
@RBLU fix will be in next release. Thanks for pointing this issue out :beers:

PR #2993, opened 2015-05-17
Add method to remove path from schema

JohnnyEstilles:
Not sure if you guys might find this useful, but I've already encountered several use-cases, so I thought it would make a good addition to Mongoose.

Unit tests included.


vkarpov15:
Thanks for the PR! This is definitely a useful feature, but will have to wait until 4.1 unfortunately because it adds a new feature.

JohnnyEstilles:
That's exactly what I expected. :-)

PR #2999, opened 2015-05-19
ClarifyDocs

anonmily:
Clarified that the model name is singular, but the collection it specifies is plural. Also added a note about the mongodb connection, since sometimes if the hostname has been changed, there might be problems connecting to localhost.

### DIFF
--- a/docs/connections.jade
+++ b/docs/connections.jade
@@ -9,7 +9,7 @@ block content
     mongoose.connect('mongodb://localhost/myapp');
 
   :markdown
-    This is the minimum needed to connect the `myapp` database running locally on the default port (27017). If the connection fails then try using 127.0.0.1 instead of localhost. Sometimes issues may arise when the local hostname has been changed.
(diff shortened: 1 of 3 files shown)


docs/connections.jade
```diff
@@ -9,7 +9,9 @@ block content
     mongoose.connect('mongodb://localhost/myapp');
 
   :markdown
-    This is the minimum needed to connect the `myapp` database running locally on the default port (27017). We may also specify several more parameters in the `uri` depending on your environment:
+    This is the minimum needed to connect the `myapp` database running locally on the default port (27017). If the connection fails or if attempting to connect via a Mongo GUI only shows a blank screen despite successful connection, then try using 127.0.0.1 instead of localhost. Sometimes issues may arise when the local hostname has been changed.
```
vkarpov15: I'd argue that the GUI stuff shouldn't be in here. IMO those docs really belong in the GUI's docs, especially since you can't be entirely sure that's how all mongodb GUI's work on all OS.

vkarpov15:
Thanks for the clarifications! Just one detail - I don't think that mongoose docs are the right place for clarifying potential issues with mongodb guis. Change that and I will merge :beers:

anonmily:
Ah, good point! I've removed mention of the MongoDB GUI problem now.

PR #3063, opened 2015-06-10
Add a property to validators from certains SchemaType to help build m…

KLicheR:
Hi!,
I work on a module to do advance localization of the Mongoose's ValidatorErrors. To have the value of the specific validator would really help to build custom error message with all the infos needed without having to dig in the schema.

## Actual ValidationError
Here's an example of an error for the `maxlength` validator. I have everything that I need in the `properties` object to build a localized error message but the actual `maxlength`.

```json
{
  "properties": {
    "type": " maxlength",
    "message": "Path `{PATH}` (`{VALUE}`) is longer than the maximum allowed length (3).",
    "path": "description",
    "value": "foobar"
  },
  "message": "Path `description` (`foobar`) is longer than the maximum allowed length (3).",
  "name": "ValidatorError",
  "kind": "maxlength",
  "path": "description",
  "value": "foobar"
}
```

## ValidationError with an additional property
I propose to add a property, keyed with the value of the property `kind` / `properties/type` that contain the value entered in the schema for this validator.

```json
{
  "properties": {
    "type": " maxlength",
    "message": "Path `{PATH}` (`{VALUE}`) is longer than the maximum allowed length (3).",
    "path": "description",
    "value": "foobar",
    "maxlength": 3
  },
  "message": "Path `description` (`foobar`) is longer than the maximum allowed length (3).",
  "name": "ValidatorError",
  "kind": "maxlength",
  "path": "description",
  "value": "foobar"
}
```

Thank you!

### DIFF
--- a/lib/schema/date.js
+++ b/lib/schema/date.js
@@ -135,7 +135,8 @@ SchemaDate.prototype.min = function (value, message) {
         return val === null || val.valueOf() >= min.valueOf();
       },
       message: msg,
-      type: 'min'
+      type: 'min',
+      min: value
     });
   }
 
@@ -190,7 +191,8 @@ SchemaDate.prototype.max = function (value, message) {
         return val === null || val.valueOf() <= max.valueOf();
       },
       message: msg,
-      type: 'max'
+      type: 'max',
+      max: value
     });
   }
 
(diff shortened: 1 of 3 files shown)


KLicheR:
Oh! And I just thought that as `SchemaString` and `SchemaDate` has both a `min` and `max` validator, it would be a good thing to know from which schema the error come from to be able to build a correct error message and handle the value of the additional property accordingly.

Is this a good approach? Do you have any suggestions?

Thanks!

vkarpov15:
Thanks, this feature is awesome. Unfortunately, it'll have to be held off until 4.1 is released, because it's a new feature.

Issue #3062, opened 2015-06-10
Document is saved after asynchronous post hooks


gregthegeek:
I have a schema, let's say `ParentSchema`. I have views that stored cached versions of information on `Parent` models. When the `Parent` is saved, I need to update related models based on this new parent. I do this in a post save hook. There are a lot of updates, so some code within this hook needs this new `Parent` to be in the database. However, it's not saved yet. Here's a simple example to replicate the issue. It seems a bit stupid in the example, since `doc` is already passed to the hook, but it succinctly demonstrates the problem.
````javascript
ParentSchema.post('save', function(doc, next) {
    Parent.findById(doc._id, function(err, result) {
        // result is null!
        next();
    });
});
````
Note that this issue only occurs when the document is new, otherwise the `findById` query can find what I assume is the old document.

Edit: It appears that when you already have an item in the database and are updating it, the update query is successfully run before the post hook. This only seems to occur when you are creating a new entity and inserting it into the db.
````javascript
// method 1
var par = new Parent({
    /* data goes here */
});
par.save(); // post('save') will be called before doc is actually saved

// method 2
Parent.findById(/* id */, function(err, par) {
    par.name = 'something else';
    par.save(); // post('save') is called after doc is actually updated (as it should be)
});
````

Edit 2: It apparently happens with synchronous hooks too:
````javascript
ParentSchema.post('save', function(doc) {
    Parent.findById(doc._id, function(err, result) {
        // result is still null!
    });
});
````
But this could probably be worked around using a call to `setImmediate` or something similar.

vkarpov15:
Which version of mongoose are you using? There was a fix for I think 4.0.4 that addressed the issue with synchronous post hooks.

gregthegeek:
I think #3066 fixes the issue. I was using 4.0.5.

PR #3066, opened 2015-06-11
Proper support for asynchronous document post hooks

gregthegeek:
Asynchronous document post hooks were actually being treated as asynchronous document pre hooks and were being registered as such. This fixes mongoose so it can properly handle these asynchronous post hooks. See issue #3062.


vkarpov15:
Great work! Thanks. :beers:

PR #3067, opened 2015-06-11
Fix for setters not being applied for push, unshift, addToSet, and set

victorkohl:
This fixes setters not being applied for push, unshift, addToSet, and set
Also added tests for MongooseArray#push

Let me know if anything could be improved/changed.


vkarpov15:
Good work, thanks.

Issue #3055, opened 2015-06-05
Mongoose v3.8.x filling populated fileds with null


eloytoro:
I really hate `null` for the obvious reasons everyone else does and also because it's bringing inconsistencies to my API

I have a Schema that goes as follows

```javascript
var UserSchema = new Schema({
  client: { type: ObjectId, ref: 'Client' }
});

var ClientSchema = new Schema({
  name: String
})
```

For some reason when I use Model#populate on an array of `users` that may or **may not** have a client it correctly populates the users that have clients, but the ones that don't **get a `null` on the client field**

Repro
```javascript
User.find({}, function (err, users) {
  Client.populate(users, { path: 'client' }, function (err) {
    console.log(users);
    // [{ client: { name: 'John' } }, { client: null }]
  });
});
```

vkarpov15:
That's by design. If it can't find a client whose id matches `user.client`, `user.client` will be null. If you can think of an alternative behavior that would be better, I'd love to hear it :)

eloytoro:
Well whats odd about it is that when _none_ of the documents have the reference set then none of them will have `null` set.

Extending the example previously used
```javascript
[{ client: 'foreign_key', age: 22 }] // populates to [{ client: {/* object */}, age: 22 }]
[{ age: 22 }] // populates to [{ age: 22 }]
```

vkarpov15:
I don't quite understand that last sentence, can you please clarify?

eloytoro:
Basically what i'm getting at is that its no use for me that the populated (but undefined) values are set to null and mongoose doesn't provide any tools to prevent this

paton:
So it looks like the bug is that when the field being populated doesn't exist on the original document, it also shouldn't exist on the populated document.

So 

{ age: 22 }

should never be populated to 

{ age: 22, client: null }

Mongoose adds the null "client" that didn't exist on the original document.

vkarpov15:
Ah ok I see, thanks for the clarification @paton

Tyrannogina:
Hi! I have mongoose version 4.11.11.
I am experiencing the exact same issue @paton is describing about adding null "client".

I created a Schema with a field that is an array filled with two types of objectId's according to a discriminator. I don't know if I am using something wrong, but when I populate the document I get the field that does not belong to the discriminator as null.

Better I put an example, I am not sure how to explain.
When I save the document, the field looks like this (got this from Robo3T):

```
"dataInputs" : [
        {
            "flag" : "potato",
[...]

PR #3074, opened 2015-06-15
prevent null assignment to populated values that didn't match any doc

eloytoro:
see #3055

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2465,6 +2465,8 @@ function assignVals (o) {
   var options = o.options;
 
   for (var i = 0; i < docs.length; ++i) {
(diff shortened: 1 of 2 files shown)


vkarpov15:
There's a broken test, see https://travis-ci.org/Automattic/mongoose/jobs/66950183. Won't merge to master if there's a real test failure :) Also, can you please add a test that covers the issue you're trying to fix?

eloytoro:
Added the test.
The first build passed, don't why it fails after the merge (passes locally for me using node v0.12.0), I'll look into it later

vkarpov15:
Don't worry about it, tests flake on travis occasionally. Removing hooks is a priority :)

Issue #2688, opened 2015-02-17
Ability to specify which ES6 promises library mongoose uses


vkarpov15:
See discussion on #1699

bigblind:
Really looking forward to being able to use Promise.all(), to do something after all database work has been done.

miguelcobain:
:+1: 

One thing that wasn't very clear on https://github.com/LearnBoost/mongoose/issues/1699 is what implementation will be the default.

vkarpov15:
Well given that it's not a backwards-breaking release, mpromise will have to be the default, but you'll be able to override.

tamtakoe:
+1

albertorestifo:
:+1:

joshperry:
Is there a branch where this is being hacked on? I can not stand the error handling flow with mpromise and I'm about to crack open the source to see what this would take.

mpromise:
```
  query.exec()
  .then(function(ou) {
    if(!ou) {
      return next(new errors.http.NotFound('The specified OU was either not found, or your credentials lack the required permissions to view it.'));
    }

    res.send(ou);
  }, next)
  .end(next);
```
Have to handle the rejection and also put `end` there. Without `end`, exceptions (for example, I had NotFound class misspelled) are silently swallowed and express just stalls.

bluebird:
```
  query.exec()
    .then(function(ou) {
      if(!ou) {
        return next(new errors.http.NotFound('The specified OU was either not found, or your credentials lack the required permissions to view it.'));
      }

      res.send(ou);
    })
    .catch(next);
```

Doing `promisifyAll(require('mongoose'))` seems to actually be still be working with Mongoose 4 so far. Would regression tests that cover this be too targeted?

vkarpov15:
Not at the moment. Most of the work will take place in the [kareem](https://github.com/vkarpov15/kareem) module as per #2754, because that will allow us to kill two birds with one stone and remove the really messy point cut nonsense that was written to make hooks and promises work together. Feel free to take a crack at it though, I'm open to PRs.

albertorestifo:
But why should we keep support for other promises library? The ES6 Promises spec is now rock-solid and here to stay. Can't we just use pure ES6 Promises, with a polyfill loaded when they are not available in older versions of node?

If so, I can have a try at it.

vkarpov15:
[...]

PR #3081, opened 2015-06-17
Implement all() and unittest it.

kristianbenoit:


vkarpov15:
Thanks for your work but I really don't want to merge this. Better solution would be #2688 IMO - implementing `.all()` in mongoose opens the door for my worst nightmare, which is having `lib/promise` look like [q](https://github.com/kriskowal/q/blob/v1/q.js)

vkarpov15:
No need for this now that #2688 is in for 4.1.0.

PR #3082, opened 2015-06-17
Feature/gh3034 this error

taxilian:
I have made one change that may be controversial, and I'm willing to change it if you like.  I dislike checking if doc === global just because in principle there could be cases where global isn't what is expected; we don't actually care specifically if it is global, we care if it isn't the object we expect.  

Therefore, instead of checking against global I'm checking to see if the features / objects we need are present.

I see this somewhat akin to the difference between checking for features in a web browser vs checking which web browser we're in -- feature checking is much more flexible =]

### DIFF
--- a/lib/types/array.js
+++ b/lib/types/array.js
@@ -35,7 +35,9 @@ function MongooseArray (values, path, doc) {
 
   // Because doc comes from the context of another function, doc === global
   // can happen if there was a null somewhere up the chain (see #3020)
-  if (doc && doc !== global) {
+  // RB Jun 17, 2015 updated to check for presence of expected paths instead
+  // to make more proof against unusual node environments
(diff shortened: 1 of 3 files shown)


vkarpov15:
You're right that checking for global is gnarly in the extreme. Admittedly it's a bit of a hack because of the special case when you do `Function.call(null)`. The approach right now of checking for certain properties is pretty brittle. How about, instead, we use `doc instanceof Document`?

taxilian:
I like it -- I'll update when I get home tonight.

taxilian:
Okay, I've updated both of those spots and they still pass the tests.  I feel like this is actually the correct fix -- good call!

vkarpov15:
Great job @taxilian, thanks

Issue #3088, opened 2015-06-19
pre update hook, update another field


danpe:
in the documentation you have this example:
> For instance, if you wanted to add an updatedAt timestamp to every update() call, you would use the following pre hook.

```
schema.pre('update', function() {
  this.update({ $set: { updatedAt: Date.now() } });
});
```

This throws: ```"unknown top level operator: $set"```

Issue #3038, opened 2015-05-31
Middleware pre and post not working when using schema.update


AlfonsoPantoja:
//version used: mongoose 4.0.4
```
//Model Report.js
var ReportSchema = new Schema({
	date_start:{ type: Date },
    date_end :{ type: Date},
    analyzed_day:{ type: Date},
    space_code: { type: String },
    metrics:[],
    updated : { type: Date },
    created: {type: Date, default: moment.utc()}
});

// Middleware pre function is executed but field "updated" is not being modified.
// Function from http://mongoosejs.com/docs/middleware.html#notes
ReportSchema.pre('update', function () {
  // change last update date.
  this.update({ $set: { updated: new Date() } });
});

mongoose.model('Report', ReportSchema);
module.exports = mongoose.model('Report');
```
----

```
var	mongoose = require('mongoose'),
Report =  mongoose.model('Report');

// More code here. Removed for convenience.

// Search for an existing document and then update it.
Report.findOne(query_condition, field_selection, query_options, function (err, report) {
          if (err) return cb(err, null); 
          // Get  data from another mongoose model. This is the data we want to update.
          var obj = dailyReportDataObj.metrics.toObject(); 
          // Update report
           Report.update({_id: report._id}, { $set: { metrics: obj }}, function(err, data){
	        		if (err) return cb(err, null);
	        		return cb(null, data);
	   });

});

```

**Using the "pre" middleware function I get this error:**
```
 { [MongoError: unknown top level operator: $set]
  name: 'MongoError',
  message: 'unknown top level operator: $set',
  index: 0,
  code: 2,
  errmsg: 'unknown top level operator: $set' } { ok: 0, n: 0, nModified: 0 }
```
**Also if I change "pre" to "post", there are no errors but the "updated" field never gets modified.**

How can I update the "updated" field? I'm missing something?

antolovich:
If you us `this.update({},{ $set: { updated: new Date() } });` it will update properly.  The problem is you are not passing a conditional object, and mongoose is adding the first parameter to the conditions.  

>_conditions:
>   { _id: 5578685dbd59f7d305c76596,
>     '$set': { updated: Wed Jun 10 2015 12:39:57 GMT-0400 (EDT) } }

Since $set is not a valid conditional operator, mongo throws the error. 

The documentation does say to do it the way you are, and it works in a pre findOneAndUpdate, but it does not work in just a pre update.

danpe:
:+1:

[...]

PR #3094, opened 2015-06-22
Corrected pre update hook

danpe:
Corrected pre update hook, confused a lot of users.
#3088 #3038


vkarpov15:
That's a bad one. Thanks for catching this! :beers:

PR #3123, opened 2015-06-30
Fix incorrect examples of some String validators

rhmeeuwisse:
On String validators minlength and maxlength:
- the special tokens {MINLENGTH} and {MAXLENGTH} were said to be replaced by the 'invalid value'. Changed that to 'minimum allowed length' and 'maximum allowed length'.
- validate() was called on variable s instead of address.
- The minlength example validated min 10, but shows error msg for min 5. Changed to validate 5
- The maxlength example validated max 10, but validates a max 10 string. Changed to validate 9 and show error msg for 9.

### DIFF
--- a/lib/schema/string.js
+++ b/lib/schema/string.js
@@ -200,12 +200,12 @@ SchemaString.prototype.trim = function () {
  *     })
  *
  *     // custom error messages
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks!

PR #3128, opened 2015-07-01
Fixes syntax errors in jsdoc; fixes #3073

### LINKED ISSUE #3073: API docs for Model#save not up to date/wrong
The API docs for Model#save are not up to date/wrong. Compare http://mongoosejs.com/docs/api.html#model_Model-save and https://github.com/Automattic/mongoose/blob/4.0.5/lib/model.js. For example, look at the parameters:

**API docs:**

    Model#save([options], product,)
    
    Parameters:
    
        [options] <Object> options set `options.safe` to override [schema's safe option](/docs/guide.html#safe)
        product, <function(err, > Number)} [fn] optional callback

**sourcecode:**

     * @param {Object} [options] options set `options.safe` to override [schema's safe option](/docs/guide.html#safe)
     * @param {function(err, product, Number)} [fn] optional callback

rhmeeuwisse:
The docs for functions model.save() and model.remove():
- used unsupported jsdoc formatting instead of markdown. This made it so that most of the function description was ignored and didn't compile into the html docs.
- the callback function params use unsupported jsdoc and got mangled in the output. I've fixed it so it's supported, but it's still illegal jsdoc. (Can't do "@param {function(err,product,Number)}", it should be "@param {Function}")
- The link to options.safe was relative and didn't render as a link; made it into a proper url but it still doesn't render as a link. (Not sure if the markdown syntax is supported inside a @param element)

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -149,41 +149,41 @@ Model.prototype.$__handleSave = function $__handleSave(options) {
 };
 
 /**
- * @description Saves this document.
+ * Saves this document.
  *
- * @example:
+ * ####Example:
  *
  *     product.sold = Date.now();
  *     product.save(function (err, product, numberAffected) {
  *       if (err) ..
  *     })
  *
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks!

PR #3131, opened 2015-07-02
Correct text on index.html

ikroeber:
Replaced misspelled word 'modeling' with 'model'.


vkarpov15:
Can you make this change in `index.jade` instead? `index.html` is compiled code.

PR #3132, opened 2015-07-02
Test added for empty strings as dates case

eloytoro:
I've encountered an issue with the Date defined in mongoose.
It seems as if the `Date` type allows the user to save empty strings `''` into the database, passing validation.
Doesn't make much sense to me.
Shouldnt it turn the value to `null` or throw a validation error?
I've added the test, but I'm not sure which would the correct behaviour be.


eloytoro:
#3134

eloytoro:
#3134

Issue #3135, opened 2015-07-03
findOneAndUpdate when upsert: true crashes if query contains a null value


odeke-em:
```javascript
targetModel.findOneAndUpdate({siteId: null}, {$set: {lastIndexingDate: Date.now()}, {upsert: true}, function(err, updatedItem) {
...
});
```
gives
```shell
/Users/emmanuelodeke/[redacted]/node_modules/newrelic/lib/transaction/tracer/index.js:168
      throw err
            ^
TypeError: Object.keys called on non-object
    at Function.keys (native)
    at module.exports (/Users/emmanuelodeke/[redacted]/node_modules/mongoose/lib/services/updateValidators.js:56:36)
    at Query._findAndModify (/Users/emmanuelodeke/[redacted]/node_modules/mongoose/lib/query.js:1651:18)
    at Query.findOneAndUpdate (/Users/emmanuelodeke/[redacted]/node_modules/mongoose/node_modules/mquery/lib/mquery.js:2118:15)
    at /Users/emmanuelodeke/[redacted]/node_modules/mongoose/node_modules/kareem/index.js:156:8
    at /Users/emmanuelodeke/[redacted]/node_modules/mongoose/node_modules/kareem/index.js:18:7
    at wrapped (/Users/emmanuelodeke/[redacted]/node_modules/newrelic/lib/transaction/tracer/index.js:157:28)
    at wrapped (/Users/emmanuelodeke/[redacted]/node_modules/newrelic/lib/transaction/tracer/index.js:157:28)
    at process._tickDomainCallback (node.js:486:13)
    at process.wrappedFunction (/Users/emmanuelodeke/[redacted]/node_modules/newrelic/lib/transaction/tracer/index.js:262:51)
```

I have been able to trace the cause of the crash and it is that Object.keys is being invoked on `null` because the previous check of `typeof null === 'object'` passes.
I'll be submitting a patch to fix it.

vkarpov15:
Thanks for the detailed report, will fix.

odeke-em:
Fixed by PR https://github.com/Automattic/mongoose/pull/3136. Did you want me to close this issue or would you prefer to close it yourself as part of your process?

vkarpov15:
Closed it. Thanks for your help!

PR #3136, opened 2015-07-03
fixed updateValidator object checker

odeke-em:
Fixed up updateValidator object checker on encountering null during upsert.
This is because it attempts a `Object.keys(v)` after checking
that a condition is an `object` but this fails for null which.
is an object.

This PR addresses https://github.com/Automattic/mongoose/issues/3135


vkarpov15:
Great job, thanks!

odeke-em:
Thank you for the merge!

PR #3154, opened 2015-07-10
Changing misspelled word "modeling" to "model".

ikroeber:


vkarpov15:
Actually I think 'for modeling' would be more correct here than 'to model'. Sorry to nitpick, but this is one of the first things people see when they go to mongoosejs.com and I want it to read nicely :)

PR #3157, opened 2015-07-11
Fix documentarray id method when using object _id

siboulet:
When trying to use an object _id field ("compound" or "composite" _id) in documentarray (subschema), I have found that the documentarray id() method is improperly casting the id to string (which effectively turn the object into "[object Object]" string).

I've read object _id fields are [not officially supported in Mongoose](https://groups.google.com/forum/#!msg/mongoose-orm/yp_hHanUruo/ffM3WQPD-i8J). However, in my use case they seems to work rather well, apart from that id() issue.


vkarpov15:
Thanks!

PR #3178, opened 2015-07-20
Fix documentarray id method when _id is null-ish.

rosston:
Fixes a bug introduced by a190d41. The tests in this PR pass in 4.0.6 but fail in 4.0.7.

When assigning a new array to a MongooseDocumentArray, with both arrays containing multiple objects/subdocuments (respectively) without _id fields, the resulting array would consist entirely of the first element. This code example explains it better:

    var Foo = mongoose.model('Foo', new mongoose.Schema({
        things: [{
            _id: false,
            name: String
        }]
    }));
    var fooInstance = new Foo();

    var thingsArray = [
        {name: 'qux'},
        {name: 'baz'},
        {name: 'bar'}
    ];

    fooInstance.things = thingsArray;
    console.log(fooInstance.things);
    // Works!
    // => [
    //   {name: 'qux'},
    //   {name: 'baz'},
    //   {name: 'bar'}
    // ]

    fooInstance.things = thingsArray;
    console.log(fooInstance.things);
    // Broken :(
    // => [
    //   {name: 'qux'},
    //   {name: 'qux'},
    //   {name: 'qux'}
    // ]

### DIFF
--- a/lib/types/documentarray.js
+++ b/lib/types/documentarray.js
@@ -117,7 +117,9 @@ MongooseDocumentArray.mixin.id = function (id) {
   for (var i = 0, l = this.length; i < l; i++) {
     _id = this[i].get('_id');
 
-    if (_id instanceof Document) {
+    if (_id === null || typeof _id === 'undefined') {
+      continue;
+    } else if (_id instanceof Document) {
       sid || (sid = String(id));
       if (sid == _id._id) return this[i];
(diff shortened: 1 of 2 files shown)


rosston:
I'd be happy to put the PR code example into a test somewhere, but I'm not sure where it should go.

vkarpov15:
Thanks!

PR #3195, opened 2015-07-24
Bug fix: Model#save should pass numberAffected as 0 when nothing changed

burtonjc:
Currently, calling `Model#save` on an unchanged document results in `numberAffected` in the callback being set to the document itself. According to the docs, `numberAffected` should always be `0` or `1` (never a document):
> ... `numberAffected` which will be 1 when the document was found and updated in the database, otherwise 0.

In mongoose v3.x this used to work as expected, it was only after upgrading to v4 that we noticed this.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -161,7 +161,7 @@ Model.prototype.$__handleSave = function $__handleSave(options) {
         });
       } else {
         _this.$__reset();
-        resolve(_this);
+        resolve();
         return;
       }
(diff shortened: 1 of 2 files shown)


vkarpov15:
Thanks!

burtonjc:
@vkarpov15 I was fixing merge conflicts for this PR and somehow closed it, sorry about that. Reopening.

PR #3205, opened 2015-07-27
Bug fix: default fields should not dirty a document across saves

burtonjc:
Currently, `document#save` does not clear the 'default' dirty cache. This means when you have a schema field with a `default` specified, sequential `document#save` calls will each execute a query even if no document changes occurred. This PR clears the 'default' dirty cache when a model is saved so the dirty check works appropriately across saves.


vkarpov15:
Great work, thanks

PR #3206, opened 2015-07-27
fix; updateValidators should not flatten dates (Fix #3194)

### LINKED ISSUE #3194: Validator does not run for Date on Model.update and $set
In my schema I have a field called `dob` which has a `max` date with the current date. So, if I try to update the document using a value like `2016` the validator should spit out a message.
```
var testSchema = new mongoose.Schema({
    firstName: String,
    dob: { type: Date, max: new Date()}
});

var Test = mongoose.model('Test', testSchema);
```
So, I've been using the validator for `update()` using `$set` like so. 
```
// Adding date to the document
var d = new Date('1/1/2016');

// Update document
Test.update({firstName : 'Alex'}, { $set: { dob:  d} }, { runValidators:true }, function (err, doc) {
    if(err){
        // Validator should run because 2016 > 2015, which it doesnt in this case
        console.log('The validator worked???????');
        done();
        return err;
    }
    done();
});
```

But the validator doesn't work for `Date`. This works completely fine with `String` and `Number` which I tried so far. 

Of course, when I try this with `findOne()` and `save()` it works
```
Test.findOne({firstName : 'Alex'}, function (err, doc) {
    // Adding date to the document
    var d = new Date('1/1/2016');
    doc.dob = d;

    // Save the updated document
    doc.save(function( err, doc ){
        // Check for error
        if(err){
              // Validator should run because 2016 > 2015, , which it does in this case
              console.log('The validator worked!!!!'); 
        }
        done();
    })
});
```
But it would be great if it validated for `Date` with update.

Here's the actually mocha tests I did: https://gist.github.com/alexhgian/97f94f84f0fbe4a033dc

victorkohl:
This fixes updateValidators for Date fields. It was trying to flatten the Date object and thus not validating the value properly.

### DIFF
--- a/lib/services/updateValidators.js
+++ b/lib/services/updateValidators.js
@@ -7,7 +7,7 @@ var ValidationError = require('../error/validation.js');
 var ObjectId = require('../types/objectid');
 
 /**
- * Applies validators and defaults to update and fineOneAndUpdate operations,
+ * Applies validators and defaults to update and findOneAndUpdate operations,
  * specifically passing a null doc as `this` to validators and defaults
  *
  * @param {Query} query
@@ -171,5 +171,6 @@ function flatten(update, path) {
 function shouldFlatten(val) {
   return val &&
     typeof val === 'object' &&
+    !(val instanceof Date) &&
     !(val instanceof ObjectId);
 }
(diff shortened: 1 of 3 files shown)


vkarpov15:
Thanks :beers:

alexhgian:
Thanks!

Issue #1918, opened 2014-02-05
Using hydrated embedded documents as query criteria leads to RangeError: Maximum call stack size exceeded


fluidsonic:
When you load a document with Mongoose and use one of it's embedded documents as a query criteria then a `RangeError: Maximum call stack size exceeded` is thrown.

**Test case:** https://gist.github.com/fluidsonic/8820684

I tracked down the error to `define` in `lib/document.js`. For whatever reason embedded documents are derived from the document which contains it in [document.js, line 1161](https://github.com/LearnBoost/mongoose/blob/3.8.x/lib/document.js#L1161). In [document.js, line 1173](https://github.com/LearnBoost/mongoose/blob/3.8.x/lib/document.js#L1173) the function tries to shadow the inherited keys by defining new non-enumerable ones on the embedded document itself.

The problem is that the [bson](https://github.com/mongodb/js-bson) library uses `for...in` to examine the size of the document recursively in [bson.js, line 215](https://github.com/mongodb/js-bson/blob/master/lib/bson/bson.js#L215). `for...in` correctly ignores the non-enumerable properties which `define` added for shadowing. But `for...in` still iterates over these properties because they are defined as enumerable in the prototype chain.

This leads to an endless recursion and the call stack overflow.

I suggest that embedded documents should no longer be derived from the document which contain it. This saves memory, makes the object structure simpler and the bson library and other code just see the keys they should.

vkarpov15:
Should be fixed by https://github.com/Automattic/mongoose/commit/978732b12187976606cd590e60b653aa0cbd7a7d

PR #3238, opened 2015-08-06
handleArray should ensure val is an Array.

jloveridge:
It seems reasonable that the `handleArray` helper method should validate that the passed value is in fact an array or coerce it into such rather than try and call `map` on a non-array object. The lack of this protection caused me some grief recently when a single value was accidentally passed in for an `$in` condition.

### DIFF
--- a/lib/schema/boolean.js
+++ b/lib/schema/boolean.js
@@ -63,6 +63,9 @@ SchemaBoolean.prototype.cast = function (value) {
 
 function handleArray (val) {
   var self = this;
+  if (!Array.isArray(val)) {
+    return [self.cast(val)];
+  }
   return val.map(function (m) {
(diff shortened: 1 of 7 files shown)


jloveridge:
I honestly don't see how these code changes would affect test #1918. Nothing in these changes should affect `$and` handling.

vkarpov15:
Hmm can you add a test for this? Good work otherwise :)

jloveridge:
Sure, I will add a test for the issue later today and update the pull request.

jloveridge:
@vkarpov15 did you want me to add more test cases or is the one that I added sufficient for this to be able to be merged?

vkarpov15:
Nope this is sufficient. Thanks.

PR #3242, opened 2015-08-07
Add UUIDv1 and UUIDv4 as Schema Types

niahmiah:
I added UUIDv1 and UUIDv4 as Schema Types.

Currently these are stored as strings, but are validated as the proper type.

I wanted to support storing them as Binary data in mongodb, because BSON supports storing UUIDs in a more optimal format (see http://mongodb.github.io/node-mongodb-native/2.0/api/Binary.html). However, I wanted to work with them as strings in the application. I was not able to determine how to implement that transformation. Any advice appreciated.


niahmiah:
I updated the branch to use the Binary data type. The 'cast' function always converts valid strings to Binary. I created a 'getter' that converts Binary back to string. This means they get stored in the most optimal way, but you can also easily work with them as Strings.

iamkhalidbashir:
why this got closed without being merged?
what is the alternative?

PR #3250, opened 2015-08-10
Fixed getUpdate() in pre update hooks

gregthegeek:
The problem:
````javascript
// whatever schema
schema.pre('update', function(next) {
    console.log('update', this.getUpdate()); // used to be null!
    next();
});

var Model = mongoose.model('model', schema);
Model.update({_id: 1}, {hello: 'world'}, {}, function(err, result) {
    // whatever
});
````


vkarpov15:
Thanks!

PR #3260, opened 2015-08-12
Deleting with _id undefined will remove all records

thanpolas:
Using `Model.remove({_id: undefined}, cb);` will remove all records from the collection.

I noticed this issue by chance from one of our [broken] tests when i upgraded from 4.1.0 to 4.1.2, my test was fixed and works ok now (no longer sending `undefined`) but this should be addressed here...

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -1915,6 +1915,24 @@ describe('Model', function(){
         });
       });
     });
+
+    it('should not remove any records when deleting by id undefined', function(done) {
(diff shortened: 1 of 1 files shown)


vkarpov15:
Worth considering making `.remove()` safer in the future. However, right now this is just how mongoose casting works - by default, `{ _id: undefined }` will get casted to `{}`, which in turn deletes all your documents.

PR #3331, opened 2015-09-02
fix: assertion test

vedmalex:
it some times fails..


vkarpov15:
The fact that this test is somewhat flakey is an indicator of a potential bug, and I'd rather not sweep that under the rug. Opened up a separate issue to track.

PR #3342, opened 2015-09-04
Adding new option skipValidation to save function

victorkohl:
Adding option to disable validation at save level.


vkarpov15:
See https://github.com/Automattic/mongoose/commit/3c50fec59cf0e2aca2c2fd806d3e2692fd08e6cd, added a way that's more consistent with the already existing `validateBeforeSave` schema option. Will be in 4.2.0.

PR #3367, opened 2015-09-12
docs; fix typos in SchemaTypes documentation

jasson15:
There are syntax errors making the examples fail in the runtime,
including missing commas and wrong attribute names.

Also, this commit appends semicolons to the ends of some codes to keep
the coding style in the same style.


vkarpov15:
Thanks! In the future, don't change the HTML files though, they're just compiled from jade.

PR #3375, opened 2015-09-15
Linting

ChristianMurphy:
Clears lint warnings from existing ESLint rules

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -270,18 +270,6 @@ Model.prototype.save = function (options, fn) {
       resolve(_this, numAffected);
     });
   });
-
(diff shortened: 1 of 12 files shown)


lib/model.js
```diff
@@ -270,18 +270,6 @@ Model.prototype.save = function (options, fn) {
       resolve(_this, numAffected);
     });
   });
-
-  return _this.$__handleSave(options).
```
ChristianMurphy: This return is made unreachable by the return on line [223](https://github.com/ChristianMurphy/mongoose/blob/linting/lib/model.js#L223)

PR #3376, opened 2015-09-15
Lint for Spacing

ChristianMurphy:
used `eslint . --fix` to resolve the warnings automatically :smile:


vkarpov15:
Can we make it so that `function() {}` is the accepted convention instead of `function () {}`, no space between function and `()`?

ChristianMurphy:
@vkarpov15 absolutely!
Updated the configuration and the code.

PR #3398, opened 2015-09-23
Replace internal pluralize with pluralize package

### LINKED ISSUE #3382: extract pluralize from utils.js
`lib/utils.js` that there is a section for pluralizing words.
and a deprecation notice that refers back to #1350.

Is this something that could now be completely deprecated and removed?
Or if the functionality needs to stay, could a library like https://github.com/blakeembrey/pluralize be used rather than maintaining an internal implementation?

Thoughts? :thought_balloon:

ChristianMurphy:
resolves #3382

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -7,6 +7,8 @@ var cloneRegExp = require('regexp-clone');
 var sliced = require('sliced');
 var mpath = require('mpath');
 var ms = require('ms');
+var pluralize = require('pluralize');
+pluralize.addUncountableRule('/_\w+_/');
 var MongooseBuffer;
 var MongooseArray;
 var Document;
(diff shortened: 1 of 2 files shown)


ChristianMurphy:
resolves #2941

vkarpov15:
Should be ok to merge this if we can ensure that all the pluralization rules stay the same. Can we add some tests to make sure that all the old pluralization rules still work? Right now the tests (utils.test.js) don't really cover the end results.

Also, let's hold off on this until 4.2 in the interest of being defensive because a mistakenly changed pluralization rule can really make somebody's life miserable.

ChristianMurphy:
@vkarpov15 This does have some breaking changes.
I manually tested the example from #2941 and 'cafe' pluralized to 'cafes' instead of the current 'caves'.
There are likely more names/words that will be affected.

vkarpov15:
It would be great to minimize the breaking changes, or better yet get rid of them entirely. Pluralizing cafes properly is something that's gonna have to wait for a major.

ChristianMurphy:
> It would be great to minimize the breaking changes

Breaking changes should already be minimal.
Pluralize appears to have started with very similar rule list, to the list used currently in Mongoose.

> better yet get rid of them entirely

I don't think this would be possible.
pluralize has the ability to add rules, but no ability to remove rules, and their rules are different.

vkarpov15:
Hmm then it may make sense to postpone this change to 5.0, unless we fork pluralize, or better yet contribute the ability to remove pluralization rules :)

ChristianMurphy:
> better yet contribute the ability to remove pluralization rules :)

This could work,
my worry would be that removing `pluralize` rules and adding `mongoose pluralize` rules, could end up being longer than the code that the existing code.

vkarpov15:
It's definitely a step in the right direction though - in a future release we could easily switch over to using pluralize's default behavior, which is undoubtedly much better than mongoose's.

PR #3437, opened 2015-10-06
Ensuring properties are casted. This fixes #3435

### LINKED ISSUE #3435: $elemMatch combined with $or not working properly
I have the following query below that worked with version 3.8.15.  After upgrading to 4.1.10 the query no longer returns any items.  I'm not sure if it has to do with the $or operator as it does work if it is commented out.

```js 
var query = {
  'teams': {
    $elemMatch: {
      team_id: '5451147e8513dd5c03e2c6e8',
      $or: [
        { 'active': { $exists: false } },
        { 'active': true }
      ]
    }
  }
};
```

Any Ideas?

**EDIT**

I tracked it down to version 3.8.25.  Looking at History.md there was #2199 that may have affected $elemMatch

DefinitelyCarter:
I found that when 'hasDollarSign' is set to true the value is returned [see here](https://github.com/Automattic/mongoose/blob/389c8d01576ced32f5cbe5e3001b008c83946ab1/lib/schema/array.js#L342-L344).

This causes properties to not get cased to ObjectId and other types.

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -326,7 +326,6 @@ function cast$all(val) {
 }
 
 function cast$elemMatch(val) {
-  var hasDollarKey = false;
   var keys = Object.keys(val);
   var numKeys = keys.length;
   var key;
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks! :beers:

PR #3448, opened 2015-10-09
docs; add doc for options.timestamps when create new Schema

isayme:
`timestamps` option is convenient but not documented, I add some simple doc for this option.


vkarpov15:
Thanks!

PR #3449, opened 2015-10-10
tests; fix unit test error in node v0.10

isayme:
var a = undefined;
    // v0.10: TypeError: Cannot call method 'b' of undefined
    // others: TypeError: Cannot read property 'b' of undefined
    a.b();


vkarpov15:
Thanks!

PR #3450, opened 2015-10-10
update updatedAt in findOneAndUpdate/update middleware if timestamps on

isayme:
mongoose has support `findOneAndUpdate` and `update` middlewares.

 when create a schema with  `options.timestamps` on, we'd better update `updatedAt` field in `findOneAndUpdate` and `update` middlewares.

Besides, I add some tests for `options.timestamps` feature.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -138,6 +138,25 @@ function Schema(obj, options) {
 
       next();
     });
+
+    function genUpdates() {
+      var now = new Date();
+      var updates = {$set: {}, $setOnInsert: {}};
+      updates.$set[updatedAt] = now;
+      updates.$setOnInsert[createdAt] = now;
+
+      return updates;
+    }
+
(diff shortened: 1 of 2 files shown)


lib/schema.js
```diff
@@ -138,6 +138,25 @@ function Schema(obj, options) {
 
       next();
     });
+
+    function genUpdates() {
```
vkarpov15: This'll break in strict mode. Use `var genUpdates = function() {};` instead please :)

alexkavon:
This was committed before the change to var genUpdates = function() {}; as stated by vkarpov15. Why?

vkarpov15:
@alexkavon because it was easier to just make that change in master after merging, I've seen first-hand how excessive nitpicking on community contributions leads to perfectly good improvements getting mothballed.

alexkavon:
Thanks for the clarification!

PR #3454, opened 2015-10-11
Authentication will now still occur when auth method is MONGODB-X509 and no password is supplied.

NoxHarmonium:
I use MONGODB-X509 authentication with my mongo DB and I have found a problem with the logic around whether mongoose should authenticate or not.

MONGODB-X509 authentication uses certificates to authenticate clients instead of a password. This means that you can leave the password field blank. The problem is that currently if you leave the password as blank than authentication doesn't happen at all and you do not have permissions to execute anything on the database.

I have added a method that checks if the current authentication method needs a password and, if it does not, still calls the mongo driver's authenticate method.

Have I missed any other places where authenticate would be called?

Thanks!

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -480,7 +480,7 @@ Connection.prototype.onOpen = function(callback) {
   }
 
   // re-authenticate
-  if (self.user && (self.pass || self.authMechanismDoesNotRequirePassword())) {
+  if (this.shouldAuthenticate()) {
     self.db.authenticate(self.user, self.pass, self.options.auth, function(err) {
       open(err, true);
     });
@@ -680,6 +680,25 @@ Connection.prototype.modelNames = function() {
   return Object.keys(this.models);
 };
 
+/**
+ * @brief Returns if the connection requires authentication after it is opened. Generally if a
+ * username and password are both provided than authentication is needed, but in some cases a
+ * password is not required.
+ * @api private
+ * @return {Boolean} true if the connection should be authenticated after it is opened, otherwise false.
+ */
+Connection.prototype.shouldAuthenticate = function() {
+  return self.user && (self.pass || self.authMechanismDoesNotRequirePassword());
+}
+
+/**
+ * @brief Returns if the current authentication mechanism needs a password to authenticate according
+ * to the passed in options object. The options object is the one passed into the open and openSet
+ * methods.
+ * @api private
+ * @return {Boolean} true if the authentication mechanism specified in the options object requires
+ *  a password, otherwise false.
+ */
 Connection.prototype.authMechanismDoesNotRequirePassword = function() {
   if (this.options.auth) {
     return this.options.auth.authMechanism === 'MONGODB-X509';
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -480,7 +480,7 @@ Connection.prototype.onOpen = function(callback) {
   }
 
   // re-authenticate
-  if (self.user && self.pass) {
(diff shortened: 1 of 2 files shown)


NoxHarmonium:
Actually I'll have to do some further testing. The authentication method is now firing for MONGODB-X509 when there is not password but the authentication fails with a very unhelpful "MongoError: auth failed". There must be a check for a password further down the line.

vkarpov15:
Can you show me how you're authenticating?

NoxHarmonium:
At the moment I have something like the following.

    var mongo_username = "CN=xxx,OU=xxx,O=xxx,L=xxx,ST=xxx,C=xxx";
    var mongo_host = "db1.xxx.com";
    var mongo_port = 12345;
    var mongo_db = "yyy";

    var mongo_url = util.format('mongodb://%s@%s:%s/%s', encodeURIComponent(mongo_username), mongo_host, mongo_port, mongo_db);

    var mongo_opts = {
        user: mongo_username,
        pass: "password to make mongoose authenticate",
        auth: {
            authMechanism: 'MONGODB-X509',
            authSource: '$external'
        },
        server: {
            ssl: true,
            sslCert: this.sslCert,
            sslKey: this.sslCert,
            sslCA: this.sslCA,
            sslPass: process.env.CLIENT_CERT_PASSPHRASE,
            logger: logger,
            auto_reconnect: true
        }
    };

I can connect fine using the mongo CLI tool.

    mongo --ssl --sslPEMKeyFile ./client-cert.pem --sslCAFile ./ca-chain.crt --host db1.xxx.com --port 29019

and authenticate with

    db.getSiblingDB("$external").auth(
      {
        mechanism: "MONGODB-X509",
        user: "CN=xxx,OU=xxx,O=xxx,L=xxx,ST=xxx,C=xxx"
      }
    )

I thought I had the authentication working at one stage when I made the changes in the pull request but I can't seem to get it to authenticate again. I'll have to keep investigating.

NoxHarmonium:
Sorry I have got back to the state I was before my config files were all mixed up.

The current behaviour is as follows:

- Using the latest version of mongoose
    - With any password string
        - It works!
    - Without password 
        - It never calls the authenticate method and the user is not authenticated
- Using my branch of mongoose
    - With any password string
        - It works!
    - Without password 
        - The authenticate method is called but the connection fails with the message "MongoError: auth failed"

NoxHarmonium:
Strangely enough, when the password is not provided, the username seems to be sent in its escaped form. The server then cannot find the user.

    [conn354] Failed to authenticate CN%3Dxxx%2COU%3Dxxx%2CO%3Dxxx%2CL%3Dxxx%2CST%3Dxxx%2CC%3Dxxx@$external with mechanism MONGODB-X509: AuthenticationFailed There is no x.509 client certificate matching the user.

but when a password is provided it is sent in the correct unescaped form.

NoxHarmonium:
I worked out what was happening. There was another place that was checking for the existence of both a username and password. If the password is was not provided, the username would not be taken from the options object that is passed in. It would then fallback to getting the username from the connection URL which didn't work because the username is not unescaped when it is parsed.

vkarpov15:
Does aff75de resolve your issue?

NoxHarmonium:
Yes, it is working well now in my project. Do you see any problems with fixing it this way?

vkarpov15:
Nope it works pretty well :) I'll merge this with 4.1.12 this week, thanks for the excellent work :beers:

PR #3461, opened 2015-10-13
Make array props non-enumerable

boblauer:
Libraries like `should` and `chai` do an array comparison by checking the length of `Object.keys(array)` (among other things).  Because a `MongooseArray` has a few properties that are enumerable, `Object.keys` will return them, and so the equality test fails.  To get around this, you have to call `toObject` on the `MongooseArray`.  This works, but it prevents you from testing a full object at once, and instead forces you to test each property separately.

This pull marks those `MongooseArray` properties as non-enumerable, so that they are not returned by `Object.keys`.  I added a new test to verify this works, and all existing tests pass.

### DIFF
--- a/lib/types/array.js
+++ b/lib/types/array.js
@@ -28,6 +28,12 @@ function MongooseArray (values, path, doc) {
   var arr = [].concat(values);;
   arr.__proto__ = MongooseArray.prototype;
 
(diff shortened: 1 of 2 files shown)


vkarpov15:
Thanks! :beers:

Issue #3470, opened 2015-10-16
Questions about read preference for query population


buunguyen:
I have a query that explicitly specifies a read option:

```js
const users = await this.Users
        .find(...)
        .read('secondaryPreferred')
        .populate('stuff')
```

Will the population query reuse the read preference of the main query? If not, what is the right way to specify the read preference for the population query? 

As far as the second question goes, I know I can specify populate options with `.populate{path: 'stuff', options: {...}}`, but the documentation doesn't specify how to declare read option.

Any help is much appreciated.

buunguyen:
The code comment of [Query.prototype.setOptions](https://github.com/Automattic/mongoose/blob/master/lib%2Fquery.js#L855) mentions about the `slaveOk` option. But AFAIK, this was deprecated long ago. So I suppose that comment is out of date.

vkarpov15:
Should be `.populate({ path: 'stuff', options: { options: { readPreference: 'secondaryPreferred' } } });`

buunguyen:
Awesome, thanks for the quick response!

PR #3471, opened 2015-10-16
Fix comment of Query.setOptions to specify correct read preference option

buunguyen:
Reported in https://github.com/Automattic/mongoose/issues/3470. The old comment referred to `slaveOk`, which is deprecated. This pull request updated the comment to use `readPreference` instead.


vkarpov15:
Thanks!

PR #3489, opened 2015-10-22
[Breaking] replace mongoose internal pluralize with pluralize package

### LINKED ISSUE #3382: extract pluralize from utils.js

ChristianMurphy:
Resolves #3382
Same as #3398


niftylettuce:
@vkarpov15 can we merge this yet?

vkarpov15:
Going to close this in favor of #5877.

PR #3492, opened 2015-10-22
Pluralize Mongoose 0.2.0

### LINKED ISSUE #3490: 4.2 breaks pluralization
When 4.2 was released a couple hours ago, it broke anyone who relied on NPM's "^4.y.z" syntax for installing "compatible" versions because pluralization rules changed.

Previously, a model named "category" would have been pluralized **categories**, but in 4.2.0 is pluralized **categorys**.

This indicates that #3415 was actually not a non-breaking version of #3398

ChristianMurphy:
Differences in filtering between Mongoose and Pluralize resulted in #3490
This update resolves #3490 by using the old filtering system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "mquery": "1.6.3",
     "ms": "0.1.0",
     "muri": "1.0.0",
-    "pluralize-mongoose": "0.1.0",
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks @ChristianMurphy

Issue #3484, opened 2015-10-20
Connect options no longer accept user & pass as empty strings


mleanos:
Previously, in version `4.1.11` I was able to pass the `user` and `pass` options into the `connect` method as empty strings. Like so..
```javascript
var db = mongoose.connect('mongodb://localhost/dev', { user: '', pass: '' }, function (err) {
  if (err) {
    console.log(err);
  } else {
    console.log('connected');
  }
});
```
However, in version `4.1.12`, this no longer works. I get a `auth failed` error. I have verified that there's a difference in the way the authentication is handling the `user` and `pass` between the two versions of this library.

```javascript
// from 4.1.11
// test passes
it('including user/pass as empty strings', function(done) {
      var db = mongoose.createConnection('localhost', 'fake', 27000, {user: '', pass: ''});
      db.on('error', function() {});
      assert.equal('object', typeof db.options);
      assert.equal('object', typeof db.options.server);
      assert.equal(true, db.options.server.auto_reconnect);
      assert.equal('object', typeof db.options.db);
      assert.equal(false, db.options.db.forceServerObjectId);
      assert.equal('fake', db.name);
      assert.equal('localhost', db.host);
      assert.equal(27000, db.port);
      assert.equal(undefined, db.pass);
      assert.equal(undefined, db.user);
      db.close();
      done();
    });

// from 4.1.12
// test fails on the db.shouldAuthenticate() assertion. 
describe('when username and password are empty strings', function() {
        it('should return false', function(done) {
          var db = mongoose.createConnection('localhost', 'fake', 27000, { user: '', pass: ''});
          db.on('error', function() {});
          assert.equal('object', typeof db.options);
          assert.equal('object', typeof db.options.server);
          assert.equal(true, db.options.server.auto_reconnect);
          assert.equal('object', typeof db.options.db);
          assert.equal(false, db.options.db.forceServerObjectId);
          assert.equal('fake', db.name);
          assert.equal('localhost', db.host);
          assert.equal(27000, db.port);
          assert.equal('', db.pass);
          assert.equal('', db.user);

          assert.equal(false, db.shouldAuthenticate()); // db.shouldAuthenticate() returns true

          db.close();
          done();
        });
      });
```

If you notice that in `4.1.11` the `user` and `pass` options are passed back as undefined. This tells me that they are being deliberately handled & bypassed by the authentication. `4.1.12` doesn't handle them the same. They are passed back as their original empty strings.

[...]

PR #3498, opened 2015-10-24
optionsProvideAuthenticationData logic refactor

mleanos:
As described in #3484 

Removes the null check for the user/pass options
in the optionsProvideAuthenticationData method of Connection. This reverts the logic back to it's previous state before https://github.com/Automattic/mongoose/commit/9af0d5fb0c5ffd4f8382715a1837e3c110b4ed3b

Adds a test to verify empty strings provided for the user/pass options
properly bypass authentication.

connections:
  shouldAuthenticate()
    when using standard authentication
      when username and password are empty strings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,10 @@ test/triage/*.js
 /inspections
 bin/mongoose.min.js
 coverage
+
+# Visual Studio
+# =========
+*.suo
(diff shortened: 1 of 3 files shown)


vkarpov15:
Thanks! :beers:

Issue #3478, opened 2015-10-18
Mongoose mocha asynch test


MMMalik:
The following code works fine with mongoose@4.0.0 (i.e. throws an error):

```
var	mongoose = require("mongoose"),
  User = require("../models").User;

mongoose.connect("mongodb://localhost/test");

describe("#create()", function () {
  it("should create a new user", function (done) {
    var u = {
      name: {
        firstName: "foo",
        lastName: "bar"
      },
      email: "email@foo.com",
      password: "hashed12345",
      credentials: "baz"
    };
    User.create(u, function (err, created) {
      throw new Error("foooooooooo");
      done();
    });
  });
});

```
The same code with mongoose >=4.1.0 hangs mocha:
```
Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```
I was running Mocha 2.3.3.

vkarpov15:
Should have been fixed re: #3222. I'll double-check. In the meantime, can you clarify which version of mongoose?

MMMalik:
I have noticed this issue with the latest 4.1.11, so I kept downgrading. With mongoose 4.1.0 still no success, but 4.0.0 works fine. I am currently working with mongoose 3.8.6 and mocha works as I expect.

Ks89:
It's still happening with mongoose 4.6.5 and mocha 3.1.2.
So frustrating, because I cannot run my tests on both MacOS and travis ci.
I'm using mongo db 3.2.10 and I chose 'bluebird' as recommended.

```
var mongoose = require( 'mongoose' );
mongoose.Promise = require('bluebird');
var dbURI = 'mongodb://localhost/test-db';
mongoose.connect(dbURI);
```

This problem it's happening with gulp-mocha, but also simply with mocha.


this is my packages.json:
```
"engines": {
    "node": "~6.7.0",
    "npm": "~3.10.3"
  },
  "dependencies": {
    "async": "^2.1.2",
    "bcrypt-nodejs": "0.0.3",
    "bluebird": "^3.4.6",
    "body-parser": "^1.15.2",
    "compression": "^1.6.2",
    "connect-redis": "^3.1.0",
    "csurf": "^1.9.0",
    "dotenv": "^2.0.0",
    "express": "^4.14.0",
    "express-content-length-validator": "^1.0.0",
    "express-session": "^1.14.0",
    "gulp": "github:gulpjs/gulp#4.0",
    "helmet": "^2.3.0",
[...]

Issue #3399, opened 2015-09-23
Model#save hangs forever and does not save


sitegui:
Hi all,

After v4.1.5 (and up to, at least, v4.1.8), the `Model#save` does not execute the callback.
Looking closer, it doesn't even send the `update` command to MongoDB.

I can repro this with the following code:
```js
'use strict'

var mongoose = require('mongoose'),
	assert = require('assert')
mongoose.set('debug', true)
mongoose.connect('mongodb://localhost:27017/test')

// Define schema
var Order = mongoose.model('Order', new mongoose.Schema({
	status: {
		// This is the important bit:
		// the embedded field `status.code`
		code: Number
	}
}), 'Order')

// Clean all data in the collection
Order.remove({}, function (err) {
	assert.ifError(err)
	
	// Insert a document using the driver
	Order.collection.insertOne({
		// Note here the field is not `status.code`
		status: 7
	}, function (err) {
		assert.ifError(err)
	
		// With findOne() we'll get the inserted doc above
		Order.findOne(function (err, order) {
			assert.ifError(err)
			console.log(order) // { _id: 56020ad372cc1fd419cb8f30, status: 7 }
			
			// And now we try to save the doc, but this fails somewhere
			// internally (even when the document is actually changed)
			// No update command is sent to MongoDB
			order.save(function (err) {
				// We never get called back...
				assert.ifError(err)
				console.log('Saved!!')
				mongoose.connection.close()
			})
			
			// Exit after a while
			setTimeout(function () {
				console.log('Failed ;(')
				process.exit(1)
			}, 3e3).unref()
		})
	})
})
```

This seems to be caused by the fact the schema is `{status: {code: Number}}` and the document from the DB is `{status: 7}`. Probably mongoose can't handle this internally, but why doesn't it throw some exception or something? Maybe it is, but the exception is sent somewhere else?

FWIW: I'm using node v0.12.7 and mongo v3.0.6, but this does not seem related

Can someone confirm this?

zladuric:
Confirming, I have the same issue with mongoose 4.1.18, with MongoDB 2.6.11 and Node 0.12.7 or 0.10.40.

FWIW I'm compiling my code with babel, from ES6. Additionally, when installing mongoose, I get compile errors on the kerberos stuff:
`../lib/kerberos.h:5:27: fatal error: gssapi/gssapi.h: No such file or directory`

Since npm continued, I assumed it's something optional. There even isn't a npm-debug.log. Was I wrong?

[...]

Issue #3222, opened 2015-08-02
mongoose 4.1.0 interferes with exception handling


dkashkin:
I just spent several hours and verified plenty of times that upgrading mongoose from 4.0.7 to 4.1.0 messes up exception handling in my node app. In other words, if I run "throw new Error('test')" after a few interactions with mongoose,  this exception never reaches my process.on('uncaughtException') handler. The same happens if I handle exceptions via node.js Domains. Everything works before I start hitting Mongoose, and even after the first few database queries. But then after running a several queries through mongoose 4.1.0 node stops calling my uncaughtException. I verified that process._events.uncaughtException still points at my function but node does not call it. I wish I could localize it better, but running out of time and decided to share because somebody might be struggling with the same thing... I am rolling back to 4.0.7 for now...
P.S. One more detail - my app never swallows exceptions. The uncaughtException handler always calls process.exit. I know that swallowing exceptions cause a lot of weird issues for V8, but we're not talking about this situation here.

sitegui:
@dkashkin 

I got the same problem here and I can reproduce this with:
```js
'use strict'

var mongoose = require('mongoose')
mongoose.connect('mongodb://localhost:27017/mongoose')
var User = mongoose.model('user', new mongoose.Schema({}))

User.create({}, function () {
	console.log('Throwing exception, see it crashing...')
	setTimeout(function () {
		console.log('Ops! The exception went somewhere else')
		process.exit()
	}, 100)
	throw new Error('This should crash the process')
})
```

In `mongoose@4.0.8` the exception crashes the process as expected.
In `mongoose@4.1.0` the exception vanishes

This does not seem to affect `find` and `update`.

sitegui:
Using `git bisect` here, it seems this was introduced by 9de7aa3c86

gergoerdosi:
I'm having an issue with this too, see https://github.com/hapijs/lab/issues/422 for details.

ksullivan:
I'm having the same issue. It has the potential to cause a debugging nightmare.

This code works as expected:

    var mongoose = require('mongoose');

    mongoose.connect('mongodb://localhost/mongoose_bug', function (err) {
        var Test = mongoose.model('Test', {field: String});

        Test.find(function (err, tests) {
            // This error is written to stdout and terminates the process.
            throw new Error('test error');
        });
    });

This code does not work as expected:

    var mongoose = require('mongoose');

    mongoose.connect('mongodb://localhost/mongoose_bug', function (err) {
[...]

PR #3499, opened 2015-10-24
repro; #3478

sitegui:
A test that reproduces #3478 and #3222 and confirms the fix introduced by 6b24976a and 5e78a568 actually works.

A related discussion about how to test these kind of behaviors came up in https://github.com/Automattic/mongoose/issues/3399#issuecomment-142713133:
> And testing this is really hard, unless you're willing to go the process.on('unchaughtException'); route. I'm open to alternative suggestions though :)

This PR uses [domain](https://nodejs.org/api/domain.html) to route exceptions through the test, instead of messing with `process.on('unchaughtException')`. Do you think this is a valid approach?

### DIFF
--- a/test/model.create.test.js
+++ b/test/model.create.test.js
@@ -4,6 +4,7 @@
 
 var start = require('./common')
   , assert = require('assert')
+  , domain = require('domain')
   , mongoose = start.mongoose
   , random = require('../lib/utils').random
   , Schema = mongoose.Schema
@@ -121,6 +122,20 @@ describe('model', function() {
       });
     });
 
+    it('does not swallow exceptions (gh-3222) (gh-3478)', function(done) {
+      var d = domain.create();
+      B.create({title: 'hi'}, function (err) {
+        assert.ifError(err);
+        d.run(function () {
+          throw new Error('This exception should be caught by the domain');
+        });
+      });
+
+      d.once('error', function (err) {
+        assert.equal(err.message, 'This exception should be caught by the domain');
+        done();
+      });
+    });
 
     describe('callback is optional', function() {
       it('with one doc', function(done) {


sitegui:
Travis CI marked the build as failing but it seems to be unrelated, since only one test in one job failed

vkarpov15:
Unfortunately this isn't a valid approach, because domains aren't really a substitute for a `try/catch`, if I remember correctly you're supposed to `process.exit()` after a domain catches an error. Of course, that's assuming that domains actually work, which is a bit of a stretch - they're really buggy last I checked.

sitegui:
@vkarpov15 I get your concerns about `domain` and `uncaughtException` being bad approaches in the general case.

But for testing purposes, at least, it seems to be more acceptable. Take for example mocha. It [does exactly that](https://github.com/mochajs/mocha/blob/master/lib/runner.js#L749). Otherwise, how could it catch asynchronous assertions?

But my real concern here is that fixing issues and not adding tests to check and prevent regressions does not seem sustainable on the long run. I'm not great fan of `domains` as a solution either, but I couldn't figure out another way out.

Would you reconsider this approach them? Until we find a less intrusive solution.

Aside: AFAIK, io.js (and consequently Node.js) deprecated `domains` because it should not be in the core. As the code is scattered and rooted in so many places, that brings to much pain to core maintainers. That decision was documented here: [TC Meeting 2014-12-10](https://github.com/rvagg/io.js/blob/dcf73cafddad745bc7833446ac9075f52dca194d/doc/tc-meetings/2014-12-10.md#deprecate-domains) and https://github.com/nodejs/node/issues/66#issuecomment-66512617.

vkarpov15:
You're 100% right about fixing issues and not adding tests, I don't do that unless there's a good reason not to. The reason why I don't like using `uncaughtException` is 1) global state, 2) may make subsequent tests inconsistent. Domains are better, but they're just so quirky and buggy that I don't want to rely on them for any reason.

Here's an alternative - re: crashing in callback, how about we just wrap callbacks in a try/catch and emit the error on the model? Would probably incur only minimal perf overhead (https://github.com/nodejs/node-v0.x-archive/wiki/Best-practices-and-gotchas-with-v8) and give us a much cleaner way to handle callback errors.

sitegui:
re: the alternative to add a try/catch, I liked that. It seems that the mongodb driver does [something like that](https://github.com/mongodb/node-mongodb-native/blob/9293941e75f1135002240f61c6526cbdc8fe9abc/lib/utils.js#L92), but instead of emitting an `'error'` event, it throws it in the next tick.

Not sure which would be preferable for mongoose. But for testing purposes, the event approach it's testable without recurring to uncaughtException/domain. :wink:

It's important to emit the event in the next tick, otherwise Node.js would throw the error [in the same tick](https://github.com/nodejs/node/blob/master/lib/events.js#L141) and all the try/catch/emit machinery would be rendered useless :/

vkarpov15:
Hmm why is it important to emit in the next tick? I think throwing the error in the same tick as the callback would be the correct behavior - if there isn't a listener on the model for the 'error' event, it'll just throw the error as if there was no try/catch in the first place.

sitegui:
> it'll just throw the error as if there was no try/catch in the first place

Yes, my point was this exception could be swallowed inside mongoose by a bug we would not be able to test for.

I did not elaborate my thoughts on this last time: if we throw in the next tick and #3222 comes back to life again, it would not harm mongoose nor users, since the exception would be throw directly to Node.js' root. No chance of that being ignored somehow.

vkarpov15:
That's a reasonable point, it is a good defensive measure. Will re-open this for 4.3.

sitegui:
Thanks @vkarpov15 

Regarding the try/catch, I wonder where in the source code it should be added:
1. [Model.create](https://github.com/Automattic/mongoose/blob/master/lib/model.js#L1755-L1771)
2. [Model#save](https://github.com/Automattic/mongoose/blob/master/lib/model.js#L288-L297)
3. [Model#$__save](https://github.com/Automattic/mongoose/blob/master/lib/model.js#L190-L230)
4. [Model#$__handleSave](https://github.com/Automattic/mongoose/blob/master/lib/model.js#L94)
5. Somewhere else?

What's your view on this?

vkarpov15:
Anywhere there's a callback would be a good criteria :) lots and lots of refactoring.

image

Issue #3502, opened 2015-10-26
Date casting using MomentJS does not work properly


alyyousuf7:
Whenever I use moment() to set a Date field, it misses out the milliseconds and set them to zero which causes issues when dealing with milliseconds or even when dealing with the ```endOf('day')``` function of momentjs.

PR #3503, opened 2015-10-26
fix; issue #3502; MomentJS date is now supported

alyyousuf7:


vkarpov15:
This really doesn't belong in mongoose core. I'd love to have a way to make this work in the general case, but relying on the constructor name is too brittle - could have unintentional behavior on classes named 'Moment' that have nothing to do with MomentJS or with different versions of moment. Would be nice to make this pluggable in a more general way in the future, but till then you can either monkey-patch the date casting logic in your code or just do `.toDate()` before passing a moment into mongoose.

alyyousuf7:
Instead of the constructor name, maybe we can check for the existence of ```.toDate()``` function, just like its done for ```.toString()```

PR #3508, opened 2015-10-27
fix; issue #3502; MomentJS date is now supported - 2nd attempt

alyyousuf7:
Maybe we can check if ```.toDate()``` function exists, just like we are doing it for ```.toString()```

### DIFF
--- a/lib/schema/date.js
+++ b/lib/schema/date.js
@@ -221,8 +221,8 @@ SchemaDate.prototype.cast = function(value) {
     if (value instanceof Number || 'number' == typeof value
         || String(value) == Number(value)) {
       date = new Date(Number(value));
-    } else if (value.constructor.name == 'Moment') {
-      date = new Date(value);
+    } else if (value.toDate) {
+      date = value.toDate();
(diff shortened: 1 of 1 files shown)


vkarpov15:
`toDate()` is still very moment-specific, and it's easy to imagine a case where a user has a `.toDate()` function on an object that returns a string or something that isn't a date...

Here's a better alternative; replace [the check for `.toString` with a check for `.valueOf`](https://github.com/Automattic/mongoose/blob/8db9f3fe7691f1ef194b49eca68d1901670ec7ee/lib/schema/date.js#L224-L231). For dates and moments, `.valueOf()` returns the numeric timestamp, which is perfect for casting dates, and [`.valueOf()` is a well-specified function](http://www.ecma-international.org/ecma-262/6.0/#sec-getvalue) so there isn't much risk of user-defined methods conflicting with it.

TrejGun:
@vkarpov15 this commit is superseded by my commit, and can be closed

vkarpov15:
Yep thanks @TrejGun

Issue #3496, opened 2015-10-23
Crash because of double casting Buffers before update


6uliver:
If I update one of my schemas:
```
Schema.update({
    _id: .id
},
{
    myBufferField: myBuffer
}, callback);
```
it throws a TypeError: data.copy is not a function at serializeBinary.
I bisected it, and the problem appears first in pull request #3250, in 4.1.2 everything works fine, since 4.1.3 it throws the error. The fault is not in the serialzation, rather in the update doc calculation. 

First, the 
```
castedDoc = this._castUpdate(utils.clone(doc))
```
wraps the doc in a Binary object:
```
{ '$set': 
   { myBufferField: 
      Binary {
        _bsontype: 'Binary',
        sub_type: 0,
        position: 379354,
        buffer: [Uint8Array] 
      } 
   } 
}
```

Later, because this._update is set before mquery's update method, it will call this._mergeUpdate(doc) and wraps the buffer again:

```
{ '$set': 
   { myBufferField: 
      Binary {
        _bsontype: 'Binary',
        sub_type: 0,
        position: 379354,
        buffer: 
          Binary {
            _bsontype: 'Binary',
            sub_type: 0,
            position: 379354,
            buffer: [Uint8Array] 
          } 
      } 
   } 
}
```

After that the serializer craches, because _updateForExec transforms the query to:
```
{ '$set': 
   { myBufferField: 
      Binary {
        _bsontype: 'Binary',
        sub_type: 0,
        position: 379354,
        buffer: [Array] 
      } 
   } 
}
```
where buffer array contains a lots of undefined element.

I haven't got any idea how this should be fixed, but I hope my detailed backtrace helps enough, because it's a critical problem for those who use Buffers heavily. 
Should you need any further information, please do not hesitate to ask.

tgckpg:
I have the same issue, thank for the detailed info. Now I just need to fallback to 4.1.2 :)

foster:
Have been tracking this myself. In the code below, `exports.isObject(from[key])` returns `true` when it encounters a node 0.12 `Buffer`. In node 4, it returns `false`.


    exports.mergeClone = function(to, from) {
      var keys = Object.keys(from)
        , i = keys.length
        , key;

      while (i--) {
        key = keys[i];
        if ('undefined' === typeof to[key]) {
          // make sure to retain key order here because of a bug handling the $each
          // operator in mongodb 2.4.4
          to[key] = exports.clone(from[key], { retainKeyOrder : 1});
[...]

PR #3510, opened 2015-10-27
Failing test for using update() to set a buffer.

foster:
Failing test only. This test passes in Node 0.12 but fails in Node 4.1.2.

As I mentioned in #3496, I think this is because `utils.mergeClone` treats a buffer as an object in Node 0.12 but not in Node 4. The determination is made by `utils.isObject`.

In node 0.12:

    Object.prototype.toString.call(new Buffer([])) // [object Object]

In node 4.1.2

    Object.prototype.toString.call(new Buffer([])) // [object Uint8Array]

### DIFF
--- a/test/types.buffer.test.js
+++ b/test/types.buffer.test.js
@@ -380,6 +380,25 @@ describe('types.buffer', function() {
 
   });
 
+  it('can be directly updated', function(done) {
+    var db = start()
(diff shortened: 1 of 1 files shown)


vkarpov15:
Fixed in #3496.

foster:
:+1: When will you be cutting `4.2.4`?

vkarpov15:
This afternoon.

PR #3513, opened 2015-10-27
Fix documentation error for Aggregate

rafakato:
In the current api docs, is said to use `mongoose.Schema.Types.ObjectId(String)` to convert a string to an ObjectId, but the correct is `mongoose.Types.ObjectId(String)`

Regards


vkarpov15:
Thanks! :beers:

PR #3517, opened 2015-10-29
Comma Last

ChristianMurphy:
All folders except `test` have been changed to the preferred comma last style.
The `test` folder will be a separate PR as it will be more changes that all the other folders combined.


vkarpov15:
Thanks :beers:

Issue #3521, opened 2015-10-30
Enum validation issue in subdocuments


tusbar:
Hey,

```js
var mongoose = require('mongoose'); // mongoose@4.2.3

var s = new mongoose.Schema({
    array: [
        {
            string: {
                type: String,
                enum: [
                    'foo',
                    'bar',
                    'baz'
                ]
            }
        }
    ]
});

var Cool = mongoose.model('Cool', s);

it('should fail', function (done) {
    mongoose.connect('localhost/cool', function () {
        var cool = new Cool();
        var entry = cool.array.create({
            string: 'hello?'
        });
        cool.array.push(entry);

        cool.save(function (err) {
            expect(err).to.be.ok;
            done();
        });
    });
});
```

The enum validation is not performed, and the `string` property is saved with garbage value.
Not sure if related to #3511, since validation works if I use a `required` instead of the `enum`.

vkarpov15:
Fix will be in 4.2.5, thanks for the repro, that was very helpful :beers:.

In the future, I'd recommend avoiding using `.create()`, it's unnecessary, you can just pass a POJO to `.push()` and get an equivalent result.

PR #3533, opened 2015-11-03
repro; #3521

tusbar:
@vkarpov15 so here’s a test scenario for #3521.

It happens in very specific cases (the test may need to be rewritten a little to highlight those specificities).

The subdocument has to be created with `subdoc.create()` and pushed into the mongoose array.
If you also add a `required` attribute to the `votes.vote` field, the validation works fine, for some reason.

I am not sure where to fix that in the code, so here’s a test, that may help you pinpoint where it is failing exactly.

### DIFF
--- a/test/schema.validation.test.js
+++ b/test/schema.validation.test.js
@@ -757,6 +757,36 @@ describe('schema', function() {
       });
     });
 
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks, will investigate this week.

PR #3565, opened 2015-11-11
Fix hang encountered with custom id type that contains a buffer

johnpeb:
modifiedPaths() walks the object structure, encounters the Buffer, then
encounters the parent of the buffer, which hangs the server for up to 30
seconds.


vkarpov15:
If you can fix the merge conflicts I can merge :) Thanks :beers:

johnpeb:
Thanks! I was planning to fix the conflict but you beat me to it.

Issue #2262, opened 2014-08-21
findOneAndUpdate option "new" has different default value from MongoDB findAndModify


mocheng:
In Mongoose doc http://mongoosejs.com/docs/api.html#model_Model.findOneAndUpdate
> new: bool - true to return the modified document rather than the original. defaults to true

In MongoDB doc http://docs.mongodb.org/manual/reference/method/db.collection.findAndModify/
>new	Boolean	Optional. When true, returns the modified document rather than the original. The findAndModify() method ignores the new option for remove operations. The default is false.

The mongoose default value of `options.new` is true while MongoDB is `false`.
This is really counter-intuitive.

Any reason for this API design?

Shuliyey:
can you return both?

Shuliyey:
can you return both original and new document for comparison purpose?

vkarpov15:
@Shuliyey there's no way to do so that I'm aware of. Would be a useful feature though :)

mnort9:
This change makes the api less expressive IMO and will likely trip up newcomers. I believe most people will expect 'findOneAndUpdate' to return the updated doc. I can understand the desire to mirror the MongoDB driver, but creating an expressive high level api should take precedence.

Our dev team recently upgraded mongoose and we've had to go back to add what seems to be a boilerplate condition ({new: true}) to all of our 'findOneAndUpdate' queries. This seems like a step backward.

I could be missing something, but what scenario would you not want the updated doc to be returned?

vkarpov15:
@mnort9 one common use case for `findOneAndUpdate` is as a distributed locking / user-implemented transaction mechanism. For instance, suppose you want to do something like "switch job's state to 'running', but I also want to know if this job was in the 'pending' state before it was changed" in order to ensure that only one process runs a given job. In that case, you want `new: false`. Admittedly we don't have good information on whether `new: true` or `new: false` is used more frequently, so that's why we changed the default for 4.0: we don't have a solid basis for diverging from the mongodb API in such a big way.

albs-br:
One common use for the original object is the "Undo edit" action so common these days. It's the same case as findOneAndDelete, which returns the deleted object that can be used to implemente the "Undo Remode" action.

dreadrocksean:
There are better ways to implement an optimistic query than this.  It is very counter intuitive and time wasting for new comers.

PR #3585, opened 2015-11-18
Test cases that illustrate incr fail

thanpolas:
I noticed that the very first `$inc` command using `findOneAndUpdate()` method will fail to perform the expected action. 

It will not increase the attribute.

The second try of the same will work.


vkarpov15:
Use the `new: true` option. See [4.0 release notes on #2262](https://github.com/Automattic/mongoose/wiki/4.0-Release-Notes#backwards-breaking-changes) and [API  docs for `findOneAndUpdate()`](http://mongoosejs.com/docs/api.html#query_Query-findOneAndUpdate)

PR #3593, opened 2015-11-19
docs; clarify param projection in findOne

gunar:
1. Short link removed because it's a duplicate of `@see findOne`.
2. `@see Query.select` added because of its relation to the `projection` argument.

Thank you!


vkarpov15:
Thanks! :beers:

Issue #2363, opened 2014-10-10
Post init event does not fire after document creation


ioncreature:
Here is CardType.js

``` js
var CardType = new Schema({
    name: {type: String, required: true},
    //...
});

CardType.post( 'init', function(){
    this._original = this.toObject();
});
```

I expect that all documents will have `_original` field

``` js
CardType.create( [{name: 'test'}, {name: 'test2'}], function( error ){
    var cardTypes = Array.prototype.splice.call( arguments, 1 );

    cardTypes.forEach( function( doc ){
        console.log( doc._original );
    });
});
```

`doc._original` always return `undefined`

Is it expected behavior or bug?

bshamblen:
Similar to the code in the original issue above I was storing a copy of the existing document in the post `init` hook, then in the pre `save` hook I was comparing the original document with the updated doc (if not new), then in the post `save` hook I had logic to send a notification of the changes. 

As @ioncreature mentioned in his original post, this all used to work. Once I upgraded to 4.0 I received the same error message `Object #<EventEmitter> has no method 'toObject'`. I've been trying to come up with a workable solution, but I can't see a way to persist the original document through to the post `save` event.

**My original code**

	schema.post('init', function() {
		this._original = this.toObject();
		this._changes = []; //array of objects with the path, new and old values. Populated by `observe` function in pre `save` hook
	});

    schema.pre('save', function(next) {
		//other logic here. Cut out for this example..

		this.observe(); //compares the current doc with the original and stores the new and old values in `this._changes` array.
		next();
	});

    schema.post('save', function(doc) {
		if (this._original && this._changes.length > 0) {
			//send web hook notification, with changes
		} else {
			//send web hook notification, with new document
		}

        //reset the values, in case the document is saved again.
		this._original = null;
		this._original = this;
		this._changes = [];
	});

The problem with the new post `init` architecture is that if I now add a parameter to the hook function I can get the `doc` and add a copy to itself, but that copy will no longer be available in the pre or post `save` hooks. 

**Changed to**

    schema.post('init', function(doc) {
		doc._original = doc.toObject();
		doc._changes = [];
	});

But in the pre `save` hook `this._original` is undefined, even for documents that we returned from the database. Any suggestions on how to reliably save a copy of the original document and have is survive the document lifecycle.

vkarpov15:
Please open up a separate issue for that, that is an unrelated issue.

azhang:
[...]

PR #3599, opened 2015-11-19
adds "instantiate" hook for document creation @ #2363

sparkida:
If this is still a potential feature, maybe something like this would work?

https://github.com/Automattic/mongoose/issues/2363

```
var Person = new Schema({ ... });
Person.pre('instantiate', function (next) {
    var doc = this;
    ...
    next();
});
Person.post('instantiate', function () {
    ...
});
var person = new Person();
```

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -74,6 +74,9 @@ function Document(obj, fields, skipId) {
   }
 
   this.$__registerHooksFromSchema();
+  if (typeof this.instantiate === 'function') {
+    this.instantiate();
+  }
 }
 
 /*!
--- a/lib/model.js
+++ b/lib/model.js
@@ -179,6 +179,13 @@ Model.prototype.$__handleSave = function(options, callback) {
   }
 };
 
+
+Model.prototype.$__instantiate = function (options, callback) {
+  var _this = this;
+  _this.emit('instantiate', _this);
+  callback(null, _this);
+};
+
 /*!
  * ignore
  */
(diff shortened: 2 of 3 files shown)


vkarpov15:
Unfortunately this isn't really that useful - because the `instantiate()` function doesn't really do anything, `pre()` hooks are essentially the same thing as `post()` hooks, and that would be a source of confusion because `pre('instantiate');` implies you get the document before any fields are set. This would require considerably more refactor. Also, adding an extra promise instantiation to creating a document would increase our memory overhead for limited benefit.

If you want to [execute some function every time a new document is created, try `Person.queue()`](http://mongoosejs.com/docs/api.html#schema_Schema-queue)..

ajsharp:
@vkarpov15 The `post('init')` hook behavior is problematic because it fails the test of reasonable expectations. This functionality is called "post-init", which leads most people to the conclusion that this hook will be fired when a document is initialized, regardless of whether this happens when a document is returned from the DB or a new document.

My recommendation is either:

1. The existing `post('init')` functionality should be expanded to include when a local object is initialized, or
1. The existing `post('init')` functionality should be changed to `post('fetch')`.

The current behavior and recommended solutions are inconsistent: use post-init when you want to run logic after a document is _fetched_, and queue `schema.queue()` when you want to run logic after a document is _initialized_. I think this library should be on a path to fix this inconsistency in the API in either a minor version bump or major, depending on if the old API is broken or not.

vkarpov15:
You're right, I opened up a separate issue to track this for the next backwards breaking release. For now, `schema.queue()` works, but I agree pre/post init hooks are quite counter-intuitive. I personally never use them...

PR #3633, opened 2015-12-01
Update populate.html

raq929:
When we tried to replicate the code in the "Population" section of the document, we couldn't replicate the results until we treated the results of populate as an array, even though it was linked to only one Id.  This update reflects what we found to work.  

@LaurenGrant, @bbarbersox


vkarpov15:
Thanks for pointing this out! Can you make this change in `populate.jade` instead? `populate.html` is compiled code and it's only in github for the sake of gh-pages.

PR #3644, opened 2015-12-03
Update virtuals.html

Psarna94:
I think the grammar here is not entirely correct. made suitable changes

### DIFF
--- a/docs/2.7.x/docs/virtuals.html
+++ b/docs/2.7.x/docs/virtuals.html
@@ -224,7 +224,7 @@
     }
 });
 </code></pre>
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks! `virtuals.html` is compiled code though, can you update `docs/virtuals.jade` with these changes instead?

Psarna94:
On it

Psarna94:
Cant find any file by the name virtuals.jade. 
screen shot 2015-12-04 at 11 10 46 am

vkarpov15:
Ah you're right, these are the old legacy docs...

PR #3653, opened 2015-12-08
Update Marked

ChristianMurphy:
Resolves two Node Security Warnings

https://nodesecurity.io/advisories/marked_redos
https://nodesecurity.io/advisories/marked_vbscript_injection

Its in devDependencies so there shouldn't be a problem.
Still better safe than sorry :smile: 

Reopens Accidentally closed #3651 :sweat_smile:


vkarpov15:
Thanks! :beers:

PR #3667, opened 2015-12-12
Docs: Update keepAlive info

ChrisZieba:
The current docs make it seem like this is a Boolean. According to the [native driver](https://mongodb.github.io/node-mongodb-native/driver-articles/mongoclient.html), the keepAlive is a number of seconds.


vkarpov15:
It's actually milliseconds: http://mongodb.github.io/node-mongodb-native/2.0/api/Server.html. I'll fix that in master after merge though.

Issue #3680, opened 2015-12-16
init and save hooks not firing for EmbeddedDocuments in a single nested schema


aliatsis:
hooks The .pre('save') hook works fine, but the .post('save') hook doesn't fire for me.

```
var childSchema = Schema({ count: Number });

childSchema.pre('init', function(next) {
// never called
});

childSchema.pre('save', function(next) {
  // never called
});

childSchema.post('save', function(next) {
  // never called
});

var SingleNestedSchema = new Schema({
  children: [childSchema]
});

var ParentSchema = new Schema({
  singleNested: SingleNestedSchema
});

mongoose.model('Parent', ParentSchema);
```

vkarpov15:
Is this the same issue as #3679 or is there something different?

aliatsis:
@vkarpov15 they are different in that pre('init') seems to works for a single nested schema but not for embedded documents within a single nested schema.

aliatsis:
also pre('save') works on a single nested schema as well but not for embedded documents within a single nested schema.

vkarpov15:
Ah ok now I understand. Thanks for reporting and clarifying, will fix today.

aliatsis:
awesome! thanks for the quick fix!

aliatsis:
@vkarpov15  not sure if this is an issue or not so I didn't want to create a new ticket.

What is the expected order of the hooks in this case?
I put logs in my hooks and received the following:
```
pre save: single nested
pre save: embedded
pre save: parent

post save: parent
post save: single nested
post save: embedded
post save: embedded
post save: single nested
```

Shouldn't pre-save be bottom-up (e.g. embedded > single nested > parent) and post-save be top-down (e.g. parent > single nested > embedded)?
post-save does seem to do this with the exception of the last 2 logs (not sure what's up with that).

Thanks again!

PR #3687, opened 2015-12-17
add subdocs of single nested first for proper hook order

aliatsis:
This is in reference to my last comment on https://github.com/Automattic/mongoose/issues/3680
By adding the sub-docs of the single nested first, pre-save can be bottom-up (e.g. embedded > single nested > parent) and post-save can be top-down (e.g. parent > single nested > embedded).

This is a critical flow for encryption tools like https://github.com/joegoldbeck/mongoose-encryption


vkarpov15:
Got it, good catch. Will push a 4.3.3 release tomorrow :beers:

aliatsis:
@vkarpov15 awesome thanks!

PR #3688, opened 2015-12-18
Feature/replset caps issue

### LINKED ISSUE #3685: Support 'replSet' as well as 'replset' for connect options

taxilian:
Completely fixes #3685 where the previous fix was incomplete

### DIFF
--- a/lib/drivers/node-mongodb-native/connection.js
+++ b/lib/drivers/node-mongodb-native/connection.js
@@ -232,7 +232,7 @@ NativeConnection.prototype.parseOptions = function(passed, connStrOpts) {
   o.db || (o.db = {});
   o.auth || (o.auth = {});
   o.server || (o.server = {});
-  o.replset || (o.replset = {});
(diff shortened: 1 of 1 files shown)


taxilian:
Note that this change does have the side effect of creating a replSet member in the config object passed in if replset exists and the other doesn't; I see this as actually being effectively a feature since it will fix the issue I just spent so much time dealing with, namely that I was using the old syntax and node-native-mongodb doesn't support the old syntax, but it's up to you.

the bit that is actually important (critical) is in lib/drivers/node-mongodb-native/connection.js line 235 which you missed when you fixed this before.

I'll take a quick look and see if I can fix the mock for that unit test

taxilian:
*sigh*.  I didn't realize that the old replset was ingrained into so many things.  I guess I'll just monkey patch it; a fixed commit will shortly follow

vkarpov15:
Thanks for the fix, I'll push it early tomorrow.

PR #3711, opened 2015-12-23
Bump mongodb dep to avoid error 'MODULE_NOT_FOUND'

jonmaim:
```
{ [Error: Cannot find module '../build/Release/bson'] code: 'MODULE_NOT_FOUND' }
js-bson: Failed to load c++ bson extension, using pure JS version
```


vkarpov15:
Not gonna happen on 3.8.x, 3.8 will always use mongodb driver 1.4, bumping to 2.0 breaks semver bad. Would recommend upgrading to 4.x.

PR #3712, opened 2015-12-23
Update mongo driver to 2.1.2

sansmischevia:
This fixes the "No primary server available" errors that occur under high load.

https://jira.mongodb.org/browse/NODE-622


vkarpov15:
Thanks :beers:

PR #3718, opened 2015-12-28
Fix support for array pipeline argument when using Aggregate#append

dbkup:
I have an issue where it's not possible to append an array of pipeline operators to an existing aggregate query object. I searched for a related issue, and couldn't find an explanation as to why this feature is ommited.

The docs state it is possible to use `Aggregate#append` with an array of pipeline operators, but the output is an error: **Error: Arguments must be aggregate pipeline operators**. Support for an array of pipeline operators exists for the `Aggregate` constructor, so I added similar functionality to `Aggregate#append` with the corresponding tests.

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -83,7 +83,9 @@ Aggregate.prototype.model = function(model) {
  */
 
 Aggregate.prototype.append = function() {
-  var args = utils.args(arguments);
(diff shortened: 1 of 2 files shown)


vkarpov15:
Thanks :beers:

dbkup:
I see now this is also present in 3.8, I'll make a separate PR for the 3.8 branch.

PR #3730, opened 2016-01-04
backport Aggregate#append fix for arrays from master

dbkup:
Backporting changes from #3718


vkarpov15:
thanks :beers:

dbkup:
For some reason the code is not showing in `aggregate.js` even though I see that lines are changed. Still there is a difference in what's shown in the changes on line **268**, instead of:

    var args = utils.args(arguments)
        , arg;

it shows only:

    var args = utils.args(arguments);

Am I missing some mechanic on GH with PR's ?

PR #3733, opened 2016-01-04
fix erroneous change in #3730

dbkup:
Fixes changes in #3730, messed up the line for changes. Now correctly placed in `Aggregate#append`.


vkarpov15:
Good catch, thanks :+1:

PR #3744, opened 2016-01-10
colorized output

TrejGun:
this PR fixes colorized output

have no idea why you did like so but `Document.inspect` same as `MongooseDocumentArray.mixin.inspect` returns a string instead of object so util.inspect can't colorize it properly

you can check this by running 
```bash
node examples/population/population-across-three-collections.js
```

this fix makes 
`Document.inspect` return `Document.toObject()` with filtered private props
`Document.toString` return string representation of `Document.inspect()`
`MongooseDocumentArray.mixin.inspect` returns array of `Document`


vkarpov15:
thanks :beers:

PR #3757, opened 2016-01-12
Update uglify-js

ChristianMurphy:
Resolves https://nodesecurity.io/advisories/uglify-js_regular-expression-denial-of-service

This should have no affect on production Mongoose.
The unpatched flaw could *theoretically* slow the build down.


vkarpov15:
Not really a problem for us since we don't export our version of uglify-js or use it for anything user-facing, but best to be safe anyway.

PR #3761, opened 2016-01-14
Update README.md

marcusmellis89:
fixing spelling error


vkarpov15:
Thanks for pointing this out, but unfortunately I think that it really means "atomically". It would probably be best to just remove that sentence entirely because it doesn't fit with the rest of the README. Do you want to take care of that or should I?

marcusmellis89:
I sent a new PR. Thank you!

PR #3766, opened 2016-01-15
Once connection opens, function should have the parameters callback

arnaud-zg:


vkarpov15:
Hmm why should this function have a callback parameter? AFAIK connection events don't emit any values.

PR #3790, opened 2016-01-24
make connect/open/openSet/disconnect return a promise, fixes #3622

### LINKED ISSUE #3622: why connect() does not return a promise ?
Since most of the api is promised, i was wondering why connect() function is not.

It looks bad (example: https://gist.github.com/davibe/8986a865a96005fa2467)

TrejGun:


vkarpov15:
Thanks :beers:

Btw how's the discussion going over at strongloop/express#2844 ? Looks like you stirred up a hornet's nest over there :)

TrejGun:
hehe, you are welcome

I'm not really participating in there. As for me Doug should give a f**k to IBM and its managers whos can bring to the express only one thing - bureaucracy. Fork the project and maintain it as before, instead he removed expressjs/express repo :( so may be my next project will be on koa.

vkarpov15:
Tried using koa but I'm not really a fan yet. I'm sticking with express for now, but also considering forking because of fear of bureaucracy.

Starefossen:
mongoose@4.4.0 completely broke my production systems because of this change. Why was this not considered a major version change? If not, what is the versioning of Mongoose releases?

vkarpov15:
@Starefossen because AFAIK this shouldn't be backwards breaking - the old callback approach should still work. How did this break your system?

Starefossen:
[This line](https://github.com/Starefossen/commit-message-ci/blob/master/lib/db.js#L17) results in `can not call function "on" on undefined` error when upgrading to v4.4.0.

vkarpov15:
You're right, sorry for the oversight. Will push a 4.4.1 fix for this today.

PR #3813, opened 2016-01-28
use valueOf to cast date

### LINKED ISSUE #3502: Date casting using MomentJS does not work properly

TrejGun:
fixes #3502 #3508


vkarpov15:
Thanks

PR #3828, opened 2016-02-01
power assert added

TrejGun:
not sure if you like this or not but want to try.

power assert: drop-in replacement of native assert library
mocha: reporter changed from `dot` matrix to `spec`

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "mocha": "2.3.4",
     "node-static": "0.7.7",
     "open": "0.0.5",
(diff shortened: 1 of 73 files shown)


vkarpov15:
Why switch the reporter? I use dot because it makes it easier to see all the results at a glance and there's a lot of tests.

TrejGun:
@vkarpov15 
ok that was a mistake push both in one commit
i have rolled back part about reporter and not this only adds `power-assert`

vkarpov15:
Thanks!

PR #3830, opened 2016-02-02
eslint airbnb config

TrejGun:
@vkarpov15, @ChristianMurphy 

sorry guys, i don't want to argue about if it should be space after function and before brace or something like this. just want code smells good

so i want to offer not to pull blanket to someones side but add `eslint-config-airbnb` package with linting rules. its much more strict than what its now and allows us to concentrate on real problems like performance

unfortunately i have to downgrade eslint to 1.10.3 for a while because it breaks tests on environments with npm 2.x which works different with peer dependencies than npm 3.x but i will put it back once `eslint-config-airbnb` switches to eslint 2.0

cheers


ChristianMurphy:
I'm ambivalent to the `space-before-function-paren` settings, either way works for me.

IMO any style guide (including the existing one) that can be automatically checked allows the project to concentrate on real problems.

vkarpov15:
Let's not bike-shed this too much. The airbnb style guide used to recommend no space before function parens, and they've broken their "API" so many times that I don't want to invest in learning what their stylistic whim of the week is. We're gonna go with `function() {}` vs `function () {}`. Thanks for the suggestion though, I wholeheartedly agree with the point about focusing on perf :palm_tree:

PR #3832, opened 2016-02-02
Use Number.isNaN() instead of global isNaN() w/ polyfill

alexkavon:
The new ES6 Number.isNaN() is much more accurate. I've updated the function to use it, as well as a polyfill for older versions of Node or IE/older browsers for the browser component. I've also added a special use case for 'undefined' (a string) and undefined (undefined) to return undefined, so that fields can still be unset.

Let me know what else I can do to improve this (are there other pieces of code that need updating? do I need to write a test? I'm not sure as this is my first time contributing to mongoose).

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -7,7 +7,7 @@ env:
 rules:
   comma-style: 2
   consistent-this:
-    - 1
+    - 2
     - _this
   indent:
     - 2
--- a/History.md
+++ b/History.md
@@ -1,3 +1,21 @@
+4.4.0 / 2016-02-02
+==================
+ * docs: fix expireAfterSeconds index option name #3831 [Flash-](https://github.com/Flash-)
+ * chore: run lint after test #3829 [ChristianMurphy](https://github.com/ChristianMurphy)
+ * chore: use power-assert instead of assert #3828 [TrejGun](https://github.com/TrejGun)
+ * chore: stricter lint #3827 [TrejGun](https://github.com/TrejGun)
+ * feat(types): casting moment to date #3813 [TrejGun](https://github.com/TrejGun)
+ * chore: comma-last lint for test folder #3810 [ChristianMurphy](https://github.com/ChristianMurphy)
+ * fix: upgrade async mpath, mpromise, muri, and sliced #3801 [TrejGun](https://github.com/TrejGun)
+ * fix(query): geo queries now return proper ES2015 promises #3800 [TrejGun](https://github.com/TrejGun)
+ * perf(types): use `Object.defineProperties()` for array #3799 [TrejGun](https://github.com/TrejGun)
+ * fix(model): mapReduce, ensureIndexes, remove, and save properly return ES2015 promises #3795 #3628 #3595 [TrejGun](https://github.com/TrejGun)
+ * docs: fixed dates in History.md #3791 [Jokero](https://github.com/Jokero)
+ * feat: connect, open, openSet, and disconnect return ES2015 promises #3790 #3622 [TrejGun](https://github.com/TrejGun)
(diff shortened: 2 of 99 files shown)


vkarpov15:
This is great work! Couple suggestions to make this better though:

1) Let's not add a property to the global `Number` function, that's an unintentional side effect. Maybe something like:

```javascript
var _isNaN = Number.isNaN || function() { /** polyfill here */ };
```

2) There's a couple other places where mongoose uses `isNaN()`: https://github.com/Automattic/mongoose/blob/9890109c12b3ba80e9e7824fa988857f7788ee2c/lib/schema/operators/bitwise.js#L30, https://github.com/Automattic/mongoose/blob/9890109c12b3ba80e9e7824fa988857f7788ee2c/lib/schema/date.js#L230 that need to be fixed. In the latter case we use `Number.isNaN()` without polyfill because of a mistake in https://github.com/Automattic/mongoose/commit/cc6c2d19e55de3787298ed36174e505f3e1909af, can you fix that too please?

alexkavon:
Of course! I'll get to work.

alexkavon:
@vkarpov15 The places I'm seeing isNaN used are the following:

* lib/schema/date.js
* lib/schema/number.js
* lib/schema/operators/bitwise.js
* test/docs/schematypes.test.js
* bin/mongoose.debug.js
* bin/mongoose.js
* docs/api.html

Do I need to update all of these? Also, when I render the docs in my browser, I'm not able to Ctrl+F the isNaN it talks about, so I'm not sure what it's referring to in docs/api.html (which I assume is the most current docs)

vkarpov15:
Don't worry about bin, that's compiled code. Also don't worry about docs or `test/docs`

alexkavon:
One more thing. Should I just move the new function/polyfill to lib/utils.js? Rather than adding the code to 3 files.

vkarpov15:
@alexkavon yep that would be great

alexkavon:
Okay, should be good to go. Not sure why the build is failing.

alexkavon:
Actually looking more at this, it needs more work. The L#207 if statement needs to be reconfigured since different values are returned differently with Number.isNaN() as per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN:
~~~
Number.isNaN(NaN);        // true
Number.isNaN(Number.NaN); // true
Number.isNaN(0 / 0)       // true

// e.g. these would have been true with global isNaN()
Number.isNaN("NaN");      // false
Number.isNaN(undefined);  // false
Number.isNaN({});         // false
Number.isNaN("blabla");   // false

// These all return false
Number.isNaN(true);
Number.isNaN(null);
Number.isNaN(37);
Number.isNaN("37");
Number.isNaN("37.37");
Number.isNaN("");
Number.isNaN(" ");
~~~

alexkavon:
My latest push tackles this issue a bit, but I'm not sure if I like it: https://github.com/alexkavon/mongoose/blob/master/lib/schema/number.js#L217-L219

vkarpov15:
@alexkavon can you fix your branch / remake this PR? Looks like something got very wonky with the commit history. I'll review this more later but let's make sure to reduce accidental backwards breaking changes, I already have my hands full with minimizing the impact of #3856 :)

alexkavon:
I can do that, but I think we should first discuss how to apporach things with Number.isNaN before creating a new PR, just so things don't get lost or taxing to refer to later.

vkarpov15:
Yeah in hindsight, this may be a backwards breaking change in a bunch of cases... I opened up a separate issue above to track, I'll close this PR for now because its unlikely to get merged in the near future re: edge cases.

PR #3840, opened 2016-02-03
Babel (WIP)

TrejGun:
Don't merge!

I was wondering if its possible to compile mongoose with babel 
so here is a result of my experiment

its totally possible to make it works and pass tests
benchmark's difference is about +1%

tell me if you are interested in porting mongoose to es6 and i can continue with this branch


TrejGun:
any comments on this?

vkarpov15:
Porting mongoose to ES2015 is an idea that's been floating around for some time. Here's my thoughts:

1) I don't want to rely on transpiled server code. Debugging mongoose issues is tricky enough without source maps.
2) If we want to do ES2015 without a server compile step, we need to drop support for node 0.10 and 0.12, which is a sizable backwards breaking change, but definitely something to consider for 5.0.
3) That being said, we'll need a transpiler for the client-side component, and babel seems like the best choice.

IMO we should keep this idea on hold for now, because there's only [a couple relatively minor changes that will definitely be in 5.0](https://github.com/Automattic/mongoose/wiki/5.0-Deprecation-Warnings) and I'm not really set on whether we'll actually drop 0.10 + 0.12 for 5.0.

Issue #3843, opened 2016-02-03
schematypes.html page is screwed


codemorpher:
http://mongoosejs.com/docs/schematypes.html is rendered  broken.
Screenshot:
http://pasteboard.co/1hlwoDLj.jpg

My OS: Linux Mint 17
Browsers: Last Chrome and FF

mynameiscoffey:
It looks like schematypes is the only page in the docs that uses both a .jade file and acquit

mynameiscoffey:
Sorry I don't know how to remove the bad PR (a3cccd7).  Not sure if this is the direction you would like to take this issue or not as well as if I am supposed to include the built pages since they are checked into the repo.

PR #3848, opened 2016-02-03
Fix for #3843

mynameiscoffey:
The /docs/schematypes.html was the only page being generated from both a .jade file and acquit.  Moved acquit output into new page and added reference in appropriate area on schematypes.jade.


vkarpov15:
Thanks! :beers:

PR #3897, opened 2016-02-20
Include the nested error when throwing CastError from arrays

kotarou3:
Helps pinpoint what caused the problem, rather than a generic-ish error from the array

Not sure of how to appropriately add tests for this


vkarpov15:
Thanks! :beers: will add a test later

PR #3899, opened 2016-02-21
Make sure Document Exists, After we check to see is obj is an Object. 

neoadventist:
If `obj` is an Object , let it first check that. Check to see if its a Document later, but only if `Document` is defined.

This is because in some instances `Document` may not be defined.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -416,10 +416,6 @@ exports.toObject = function toObject(obj) {
     return obj;
   }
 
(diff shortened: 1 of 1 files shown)


vkarpov15:
Can you provide a case where `Document` might not be defined? Might be better to instead make sure Document is always defined than just check.

vkarpov15:
Been stale for a while, closing

PR #3923, opened 2016-02-29
addition at line 10737

drewbiewankenobi:
Added pluralization case for "Hero" & "Zero" - currently pluarlizes to "heros" and "zeros."


vkarpov15:
Can't quite merge this, because it's very scary to change the pluralization code at the moment in a point release. Could very easily break somebody's prod setup if they rely on the current pluralization behavior and are using `"mongoose": "4.x"`.

Also, this change is in a compiled file - `bin/mongoose.js` is our browserify build.

In the future, re: #3489, we'll make pluralization more pluggable, so you can opt in to this sort of behavior if you want. I'm gonna close this. For now, if you want to tell mongoose to use the `heroes` collection, just pass the collection name as the third param to the `model()` function: `mongoose.model('Hero', heroSchema, 'heroes');`

Issue #3925, opened 2016-03-01
Make ValidatorError.properties not enumerable


Fonger:
The `properties` of ValidatorError is a internal property hence not necessary to be displayed on output.

EDIT: I just create a pull requrest.

```json
  {
    "message": "User validation failed",
    "name": "ValidationError",
    "errors": {
      "email": {
        "properties": {
          "message": "{VALUE} is not a valid email",
          "type": "Format",
          "path": "email",
          "value": "qqqq#wer.asdfsdf"
        },
        "message": "qqqq#wer.asdfsdf is not a valid email",
        "name": "ValidatorError",
        "kind": "Format",
        "path": "email",
        "value": "qqqq#wer.asdfsdf"
      }
  }
```

Currently, I use the following code to hide `properties` automatically.
I wish this can be implemented in Mongoose internally.
I have submitted a pull request.

```js
/**
 * Hide Mongoose ValidatorError internal properties
 */
var MongooseError = require('mongoose').Error;
Object.defineProperties(MongooseError.ValidatorError.prototype, {
    properties: { enumerable: false },
});
```

And it will display cleaner error like this:
(although properties are still accessible, but when it's converted to JSON, it will be hidden)

```json
  {
    "message": "User validation failed",
    "name": "ValidationError",
    "errors": {
      "email": {
        "message": "qqqq#wer.asdfsdf is not a valid email",
        "name": "ValidatorError",
        "kind": "Format",
        "path": "email",
        "value": "qqqq#wer.asdfsdf"
      }
    }
  }
```

Note: It's **Validator**Error not **Validation**Error

PR #3926, opened 2016-03-01
make ValidatorError.properties not enumerable

Fonger:
Resolve issue #3925

### DIFF
--- a/lib/error/validator.js
+++ b/lib/error/validator.js
@@ -19,7 +19,11 @@ function ValidatorError(properties) {
     msg = errorMessages.general.default;
   }
 
-  this.properties = properties;
+  Object.defineProperty(this, 'properties', {
+    enumerable: false,
+    value: properties
+  });
+
   var message = this.formatMessage(msg, properties);
   MongooseError.call(this, message);
   if (Error.captureStackTrace) {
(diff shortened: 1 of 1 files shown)


vkarpov15:
Can you define this on the prototype like you showed in https://github.com/Automattic/mongoose/issues/3925 ? Better performance that way.

Fonger:
@vkarpov15 I can't use that way. It will fail the test because it will erase the value of properties in constructor.

```js
          var M = mongoose.model('gh-2132', schema, 'gh-2132');

          var m = new M({x: 'a'});
          m.validate(function(err) {
            assert.equal('Error code 25', err.errors.x.toString());
            assert.equal(25, err.errors.x.properties.errorCode);
            done();
          });
```

```
  1462 passing (17s)
  1 failing

  1) schema validation messages are customizable supports custom properties (gh-2132):
     Uncaught TypeError: Cannot read property 'errorCode' of undefined
      at test/schema.validation.test.js:649:53
      at lib/document.js:1090:19
      at complete (lib/document.js:1164:7)
      at lib/document.js:1195:20
      at ObjectId.SchemaType.doValidate (lib/schematype.js:682:12)
      at lib/document.js:1191:9
```

vkarpov15:
Will review more closely in a bit. I'd rather not do `defineProperty` transactionally, but if there's no other way this should work too.

vkarpov15:
@Fonger found a way to do this by defining the property on the prototype. `defineProperty()` is very slow and I'd rather not do that for every ValidatorError.

Fonger:
@vkarpov15 thanks :beers:

Issue #3768, opened 2016-01-16
Throw error if doing `import { model } from mongoose;` or `const { model } = require('mongoose')`


Elergy:
It's common when user wants to import variables from module without saving module into a variable, e.g.

```
let {Schema} = require('mongoose');
```

or 

```
var Schema = require('mongoose').Schema;
```

but when I do this with model 

```
var model = require('mongoose').model;
```

and try to create a model with imported function, I have an error

```
TypeError: Cannot read property 'modelSchemas' of undefined
```

Yes, model is a property of Mongoose.prototype, but Schema also is.
If documentation contains an instruction like 

```
var Schema = mongoose.Schema,
...
```

it's possible to assume that 'model' will have similar behavior

vkarpov15:
Supporting this behavior would involve a massive refactor of mongoose connections. Long story short, `require('mongoose').model();` is different than `var model = require('mongoose').model; model();` because the value of `this` in the former function call is the `require('mongoose');` singleton, whereas it is the global object in the latter call. If you want a workaround, `model.call(require('mongoose'), 'MyModelName', schema);` should work fine. I'm not a huge fan of the fact that mongoose's top level functions are laden with side effects, but to get rid of that we'd also have to get rid of the `mongoose.model()` getter syntax, which would break a lot of people's code.

Elergy:
@vkarpov15 thanks for your answer!

There are 2 different cases:
1. We know it's a bug, but fixing it takes a time for refactoring. Then we should keep the issue open
2. We assert it is not a bug. Then we should close the issue

When we have the issue open, it can be fixed with breaking change (in new major release) or without breaking change (you know, everything can be refactored with backward compatibility, it just depends on effort). Someone (e.g. me) probably has time for it. 

But when we mark the issue closed, we give contributors to know that pull requests with fixing this bug will be rejected, and someone doesn't even start to fix it. 

What do you think about it?

vkarpov15:
Fair point, I'll keep this issue open. The workaround is easy, just ensure the functions have proper context. The eventual solution is no side-effects.

SummerWish:
+1

Currently this doesn't work:
```js
import { model } from 'mongoose'
```

vkarpov15:
[...]

PR #3943, opened 2016-03-04
Update index.js

jacob-israel-turner:
See #3768

To allow invocation of `model` in any context, I do a quick check to see if `this` within the invocation refers to the `mongoose` instance.  If it does not, `Mongoose.prototype.model` is called again with the proper context.


jacob-israel-turner:
All the tests are failing.  I should have worked with this more extensively locally first.  Closing for now.

PR #3952, opened 2016-03-06
Allow setup timestamps by Schema.prototype.set

### LINKED ISSUE #3951: Timestamps doesn't work with schema.set('timestamps', ...)
The following code won't work.

```js
var userSchema = new Schema({
    name: String
});
userSchema.set('timestamps', true); //true or any other config
```

The following code works. According to [docs](http://mongoosejs.com/docs/guide.html#timestamps), timestamps is one of the `options` so I think it should be set by the first way.

```js
var userSchema = new Schema({
    name: String
}, { timestamps: true });
```

Fonger:
This pull request resolves #3951

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -119,51 +119,8 @@ function Schema(obj, options) {
     this[m.kind](m.hook, !!m.isAsync, m.fn);
   }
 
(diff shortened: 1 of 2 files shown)


vkarpov15:
Thanks :beers:

PR #3966, opened 2016-03-10
Fixed example code to print 'meow' on success

janoskk:
The code was misleading, please consider to merge.


vkarpov15:
index.html is generated code, please instead modify index.jade and put in a PR against master.

PR #3985, opened 2016-03-16
DOC: add remark about MoongooseArray.pull(...)

seriousManual:
It took me a while to figure out why my pull operations resulted in actual `$set` operations until I figured out that this is only the case when more then one `pull`s are happening.


vkarpov15:
Thanks :beers:

Issue #3946, opened 2016-03-05
Model.aggregate.then is not a function


adambuczynski:
It seems the `aggregate` method on models is not thenable, is this intentional or an accidental omission?

vkarpov15:
Intentional. Use the .exec() function. We'll add a .then() that does this for you in a future release

adambuczynski:
Ok that'd be great. I use the promises interface for all other Mongoose operations, so would be good if it can be consistent with aggregate as well. Thanks!

TrejGun:
+1

PR #3990, opened 2016-03-16
Wrong doc.

megagon:
I am going to make a promise for #3946, but meanwhile i found that it has wrong doc.


vkarpov15:
Thanks :beers:

PR #3993, opened 2016-03-17
Hack for backward compatibility, see #3946

megagon:
Made a hack to make promise and keep backward compatibility. #3946 
All tests locally passed, added new test to test aggregate promise.


megagon:
need to test properly

PR #3994, opened 2016-03-17
Args order passing to resolve changed

megagon:
Args order passing to resolve changed, see #3946.


megagon:
you can make an issue and assign it to me, we can re-work this and make promises by default on next major release. Meanwhile i can add a warning here.

megagon:
need to change string wrapper

PR #3995, opened 2016-03-17
aggregate promise #3946

megagon:
aggregate promise for #3946.
tests passed locally, new test added.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2263,12 +2263,11 @@ Model.aggregate = function aggregate() {
 
   aggregate.model(this);
 
-  //TODO find proper way to check that we will need call done()
-  if (typeof callback === 'undefined' && arguments.callee.caller.toString().indexOf(").then") == -1) {
+  if (typeof callback === 'undefined') {
     return aggregate;
   }
 
-  return aggregate.exec(callback);
+  aggregate.exec(callback);
 };
 
 /**
--- a/lib/model.js
+++ b/lib/model.js
@@ -2210,7 +2210,7 @@ Model.geoNear = function(near, options, callback) {
 /**
  * Performs [aggregations](http://docs.mongodb.org/manual/applications/aggregation/) on the models collection.
  *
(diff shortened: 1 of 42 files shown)


lib/model.js
```diff
@@ -2263,7 +2263,8 @@ Model.aggregate = function aggregate() {
 
   aggregate.model(this);
 
-  if (typeof callback === 'undefined') {
+  //TODO find proper way to check that we will need call done()
+  if (typeof callback === 'undefined' && arguments.callee.caller.toString().indexOf(").then") == -1) {
```
vkarpov15: Really not a big fan of this solution. Would be much better to add a `.then()` function to aggregate.prototype.

vkarpov15:
Would be better to do something like this: https://github.com/aheckmann/mquery/blob/d82d97f7a81be6c5288d7092482b3b7c81980fdf/lib/mquery.js#L2452-L2462

megagon:
we can't do, because we have situations, when we don't have nor callback, nor promise.
And we need to return `aggregate` object if we don't have it.
If we will make it like in the link you provided - we will return promise every time, but we can't return it every time. For example, [this one](https://github.com/Automattic/mongoose/blob/master/test/aggregate.test.js#L276-L283) will not work
am i missing something?

vkarpov15:
You would only return a promise when somebody calls .then, which would be fine IMO. That's what we do for queries, what superagent does for requests, etc. and this pattern seems to work well

megagon:
@vkarpov15 
Updated, please take a look

vkarpov15:
Lgtm. Gonna hold off on merging this until 4.5 because it's a reasonably big new feature

TrejGun:
is it possible to squash all this into one commit so i can patch my fork?

vkarpov15:
@TrejGun I'll merge this into the 4.5 branch tomorrow so you can use the branch.

vkarpov15:
@TrejGun merged into 4.5: https://github.com/Automattic/mongoose/commit/b21de207e29e5d5464e4dd1a87695ad4b3ef5655

Thanks @megagon , gonna close this since it was merged above

PR #4017, opened 2016-03-23
Fix typo in queries doc

MichaelArnoldOwens:


vkarpov15:
I think these docs are correct, they provide an example of querying with regulator expressions. Do these not work for you?

vkarpov15:
Issue's gone stale, re-open if you still think these docs are incorrect.

PR #4026, opened 2016-03-29
Add autoIndex config example at connection level

tilleps:
Adding more visibility to the `autoIndex` option by adding examples at the connection level


vkarpov15:
Thanks

PR #4032, opened 2016-03-31
fix(mongos connection): Allow connecting with {mongos: true} to handle connection query params

burtonjc:
Currently, if `ssl` and `sslValidate` are specified in the connect string query, and you pass `{mongos: true}` to the `mongoose.connect`, `ssl` and `sslValidate` are not set on the `mongos` opts. You can manually specify `{mongos: {ssl: true, sslValidate: false}}` and it works, but it is [documented](http://mongoosejs.com/docs/connections.html) that `{mongos: true}` is enough. This adds that capability.


vkarpov15:
Thanks

Issue #1334, opened 2013-02-05
Saving document with projected sub document modifies the wrong sub document


mseegers:
Hi,

when retrieving a document using findOne while limiting the result of an array of sub documents via an $elemMatch projection, the wrong array element will get modified in the database when saving the document back.

Please see the following test case:

https://gist.github.com/4711139

Which produced the following debug output:

> Mongoose: docs.findOne({ someId: 1 }) { fields: { subDocs: { '$elemMatch': { someValue: 3 } } }, safe: undefined }  
> Mongoose: docs.update({ _id: ObjectId("511054f985a8270000000002") }) { '$set': { 'subDocs.0.someValue': 4 } } {} 

Notice up the update command sets 'subDocs.0.someValue' instead of 'subDocs.2.someValue'. I assume this is because an array with only a single element was returned due to the projection.

We have been relying on mongoose's ability to selectively update parts of large documents including many subdocuments depending on what data had been modified. Without the ability to use projection, the document retrieval kills our performance as node will spend a bulk of its time parsing the bson data. See also the following issue for a similar issue that we've encountered:

https://github.com/LearnBoost/mongoose/issues/1303

Thanks!

This is using mongoose 3.5.4

geloescht:
This might actually be doable using $(update): https://docs.mongodb.org/v2.6/reference/operator/update/positional/
We'd have to pull out the projection clause (for $elemMatch) or query criterion (for $) from the original query and copy it to our update filter. Then, instead of using "array.0.element" to update the array we need to use "array.$.element". This can only with projection operators that select a single array element (thus, not with $slice) and if no new elements were added to the array.
I admit, it sounds quite complicated and fragile, but do you see any chance that this is ever going to be implemented @vkarpov15 ? I had a look at the code generating queries for Document.save() myself, but do not feel confident enough to touch it.

vkarpov15:
@geloescht can you provide a code sample that demonstrates what you're looking to do?

geloescht:
I am developing a solution for content localization. Basically, instead of modifying a translatable document property directly, I use a virtual that modifies an entry inside an array with localized data. To save memory and computing power for documents with a lot of translations I use projection to only retrieve localized data for a single language. This is all wired up using mongoose middleware.
Basically it boils down to:
```javascript
apps.findOne(
{
  _id: ObjectId("567452bae5b25d6e6c1a0f7e")
},
{
  localization: { '$elemMatch': { language: 'de' } }, dummy_: 0
[...]

PR #4053, opened 2016-04-07
Allow saving of array fields selected via $elemMatch

geloescht:
See #1334 for discussion.
Test "model field selection with $elemMatch projection disallows saving modified elemMatch paths (gh-1334)" will fail, because saving now succeeds.
I do not know exactly, which additional checks are necessary. For example, what does Mongoose do, when an array field is removed or overwritten with a completely new array? Do I have to check for that?

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -482,17 +482,14 @@ Model.prototype.$__delta = function() {
       var pathSplit = data.path.split('.');
       var top = pathSplit[0];
       if (this.$__.selected[top] && this.$__.selected[top].$elemMatch) {
-        //If the selected array entry was modified
-        if (pathSplit.length > 1 && pathSplit[1] == 0) {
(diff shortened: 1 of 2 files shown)


vkarpov15:
This looks like a pretty good solution :+1: Can you change the gh-1334 test so it succeeds? Would be helpful for me if we could see all the travis tests pass :)

geloescht:
What should happen when someone empties the array? This happens in the test. Currently, without further special casing, an empty array is written to the database. This seems very dangerous to me. Would it be more logical to either $pull the selected array entry or disallow its removal altogether?

vkarpov15:
Honestly I'd just recommend throwing a [divergent array error](https://github.com/Automattic/mongoose/blob/master/lib/error/divergentArray.js). That's a pretty dangerous case.

Issue #3379, opened 2015-09-16
[Discussion] Node 4/ES6


ChristianMurphy:
This issue is to start a **discussion** about the potential for a future release of Mongoose, that would drop support for Node JS versions less than Node  4.
This would allow Mongoose to take advantage of new ES6 features natively supported in Node 4+.

Some ideas on changes that could come with better ES6 support.
### replace `var` with `let` and `const`

`let` and `const` are block scoped and are not hoisted, this can make the code more explicit in its intent and has been shown in **_some**_ benchmarks to be as fast or faster than `var`.

ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let
ref http://davidwalsh.name/for-and-against-let
ref https://jsperf.com/let-vs-var-definition/21
### replace `mpromise` package with native `promise`

This would remove a dependency and offers a more spec compliant version.

ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise
### replace `var _this = this` with arrow functions

Arrow functions pull the `this` keyword from the lexical parent, which is what most of the uses of `var _this = this` in the code are doing just that.

ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions
### replace `Object` based models with `class` based models

Replace the existing system for creating a new model

``` js
var Cat = mongoose.model('Cat', { name: String });

var kitty = new Cat({ name: 'Zildjian' });
kitty.save(function (err) {
  if (err) // ...
  console.log('meow');
});
```

With a class based system

``` js
class Cat extends MongooseModel {
   constructor (name) {
      super({name: String});
      this.name = name;
   }   
}

var kitty = new Cat('Zildjian');
kitty.save(function (err) {
  if (err) // ...
  console.log('meow');
});
```

This could allow mongoose to be more extensible, and offer a system familiar to users of ORMs in other languages.

ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes
ref https://github.com/getify/You-Dont-Know-JS/blob/master/es6%20&%20beyond/ch3.md#classes

ChristianMurphy:
@vkarpov15 Would a release that drops pre Node 4 support even be feasible?
and if so do any of these ideas sound interesting?

vkarpov15:
[...]

PR #4057, opened 2016-04-10
converted ALL your `var` to `const`\`let`! have a nice from boost(js)

astman:


vkarpov15:
Thanks, but we currently support node 0.10 and 0.12, so no ES2015 for mongoose yet. I'm unfortunately going to have to close this, because we're not going to do this for a while, but follow #3379 for the discussion on this issue.

Issue #1627, opened 2013-08-07
Add Favicon for documentation website


djrenren:
Having a favicon on the mongoose site would make it look more professional as well as allow users to recognize the website in their browser tabs more easily.

PR #4082, opened 2016-04-17
Add docs favicons.

robertjustjones:
It's been a while since #1627, but here's the PR. Not fancy, but helps me find the tab. I've based it on the last tagged commit because "make docs" is failing on the latest.


vkarpov15:
Great work, thanks :beers:

PR #4097, opened 2016-04-26
Fix override of pre-set options when passing custom opts

amit3vr:
when passing custom opts object to toJSON() or toObject(), the default/pre-set (or as you may to call it) schema options are mistakenly (?) overriden.

**meaning**, that if i have a transformed toJSON/toObject (using the `transform` option), and i want to pass my own custom opts, i'll accidentally override the schema options object which i pre-set to act as defaults, and ignore it -- causing inconvenience and discrepancy.

this commit fixes it. nothing too fancy or complicated.

### '/schemas/animal.schema.js':
```
AnimalSchema.set('toJSON', { virtuals: true, getters: true });

AnimalSchema.options.toJSON.transform = function(animal, animalJSON, opts)
{
    console.log(opts);
    return animalJSON;
}
```

### 'controllers/animal.controller.js'
```
animalDoc.toJSON(); // outputs "{ transform: [Function], getters: true,virtuals: true, json: true,_useSchemaOptions: true, minimize: false }"

animalDoc.toJSON({ customOpt: true }); // outputs "{ customOpt: true, transform: true, json: true, minimize: false }"
```

==

this PR will **correct** toJSON opts object, extending it with the custom passed custom opts param:
(after merge)
```
animalDoc.toJSON({ customOpt: true }); // outputs "{ customOpts: true, transform: [Function], getters: true,virtuals: true, json: true,_useSchemaOptions: true, minimize: false }"
```

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1880,6 +1880,15 @@ Document.prototype.$__handleReject = function handleReject(err) {
 
 Document.prototype.$toObject = function(options, json) {
   var defaultOptions = {transform: true, json: json};
+  var inheritedMethodOptions = json ?
+      clone(this.schema.options.toJSON) :
+      clone(this.schema.options.toObject);
+
+  if(!(inheritedMethodOptions && utils.getFunctionName(inheritedMethodOptions.constructor) === 'Object')) {
+    inheritedMethodOptions = {};
+  }
+
+  defaultOptions = Object.assign(defaultOptions, inheritedMethodOptions);
 
   if (options && options.depopulate && !options._skipDepopulateTopLevel && this.$__.wasPopulated) {
     // populated paths that we set to a document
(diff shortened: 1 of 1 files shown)


amit3vr:
Okayyy this is kinda awkward.. :sweat_smile: but i cannot understand this assertion error!

vkarpov15:
The test failure is by design unfortunately - right now if you pass any options to `toObject()` or `toJSON()`, those options will overwrite **all** the default options. You an workaround this by doing something like `doc.toJSON(Object.assign({}, schema.options.toJSON, { customOpt: true });` or by writing a plugin. Unfortunately can't merge this because it's backwards breaking :(

amit3vr:
This is super nasty, but I get the whole backwards compatability notion.

To work this around better, maybe add an options to pass 'extendDefaults' options?
If not, I will appreciate it if you'll tag this as a milestone for the next version. :)

vkarpov15:
TBH I think this is a better api but I'm loathe to add new options unless really necessary. I'll open up a separate issue for 5.0.

Issue #4106, opened 2016-04-28
Plugins are applied to schemas more than once.


RobertWHurst:
I noticed that plugins were running applied to my schema tens of times within some tests I wrote for a server my team has created. I noticed the issue was caused by the way we have our server structured. We create our schema instances and export them. Then have a modules containing a server constructor that requires them. When the server constructor is executed it constructs a new instance of Mongoose then registers global plugins with it. Lastly it registers the schemas with the mongoose instance. All is fine up to this point.

The problem arises when a new instance of the server constructor is created. The plugins are registered with the new mongoose instance, but then when the schemas are registered, which remember, are the same schema instances passed the the previous mongoose instance, the plugins are applied the the schemas yet again. Now the schemas have had all of the global plugins applied to them twice.

I'm aware that this can be fixed by exporting a factory function instead of the schemas directly, but I would like to suggest that mongoose should check to see if a plugin has already been registered with a schema before adding it. I'll make a PR to demonstrate this. Thoughts?

vkarpov15:
Useful in some cases, but the tricky part is how do you check if a plugin has already been registered? Plugins are just functions - you can do an exact equality check, but that will miss cases where plugins use the "function that returns a function" pattern

RobertWHurst:
Going to close this as my fix was merged. Thanks to @vkarpov15 for merging it. For anyone who's curious, my fix keeps a reference to the passed function. If the function was encountered prior, then the plugin function simply returns this, instead of executing the plugin.

jorgecuesta:
Hey guys,

The functionality described here by @RobertWHurst  we implemented it in a npm package
[mongoose-plugins-recurrency](https://github.com/jorgecuesta/mongoose-plugins-recurrency).

Since we (@jeffreysoriano5) basically copied the functionality we would like to put you as a contributor to the package.

We are open to suggestions, ideas and unit-test enhancements.

**Regards...**

vkarpov15:
Thanks! :beers:

PR #4107, opened 2016-04-28
Apply plugins to schemas only once

RobertWHurst:
Prevents the same plugin from being applied to the same schema more than once. See issue #4106 for details.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -81,6 +81,7 @@ function Schema(obj, options) {
   this._requiredpaths = undefined;
   this.discriminatorMapping = undefined;
   this._indexedpaths = undefined;
+  this._appliedPlugins = [];
 
   this.s = {
     hooks: new Kareem(),
@@ -905,6 +906,8 @@ Schema.prototype.post = function(method, fn) {
  */
 
 Schema.prototype.plugin = function(fn, opts) {
+  if (this._appliedPlugins.indexOf(fn) > -1) { return this; }
+  this._appliedPlugins.push(fn);
   fn(this, opts);
   return this;
 };
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -861,6 +861,22 @@ describe('schema', function() {
       assert.equal(true, called);
       done();
     });
+
+    it('does not apply more than once', function(done) {
+      var Tobi = new Schema,
+          called = 0;
+
+      var plugin = function(schema) {
+        assert.equal(schema, Tobi);
+        called += 1;
+      };
+
+      Tobi.plugin(plugin);
+      Tobi.plugin(plugin);
+
+      assert.equal(1, called);
+      done();
+    });
   });
 
   describe('options', function() {


vkarpov15:
Thanks

RobertWHurst:
@vkarpov15 🍻

Jeff-Lewis:
I like this addition but I wonder if use cases exist where people apply the same plugin multiple times but with different arguments/options? I know we don't like more config options but should we make this configurable?

vkarpov15:
@Jeff-Lewis perhaps we could do a deepEquals check on `opts` too as well as checking if the function is the same?

Jeff-Lewis:
Imagine a validation plugin that is processed multiple times for different fields. 

This PR silently ignores the plugin the second time which I feel alters mongoose's behavior without the consumer knowing and would be hard to track down.

IMHO, we should keep the default behavior of processing a plugin as many times as it's requested. 

We can either:
1) Expose the schema's _appliedPlugins for consumers and let them interrogate it themselves to determine if they want to process the plugin again 

or 

2) Add a schema configuration option of 'allowDuplicatePluginAssignments' or something that is true by default.

I think the cleanest for Mongoose is option 1.  @RobertWHurst Would option 1 work for you?

vkarpov15:
I still think checking if opts is equal would be a good alternative, no?

1 could work well actually - you could easily write a plugin that prevents plugins from getting applied multiple times :)

RobertWHurst:
If we must, I think I prefer the idea of checking the opts personally. I can't think of a use case as a plugin author where I would actually need my plugin function executed multiple times. Once your plugin function is called and you are passed the schema you can continue to work on the schema anytime after that by simply keeping a reference. In the case of a validation plugin, would one not just add a method to the schema to add validation rules rather than needing to be loaded over and over? Executing the same plugin more than once seems like a recipe for nasty bugs for the majority of plugins. In our case we passed our schema to a number of mongoose instances for testing. Each instance would apply the same plugins over and over causing a number of issues. Plugins that need that behavior must be extremely rare, and to be honest it seems like a code smell.

Jeff-Lewis:
@RobertWHurst As I said before, I like the PR and I would agree that applying a plugin multiple times with different options could be considered code smell. While Mongoose being a very popular library, I am more concerned about changing the behavior and silently ignoring subsequence .plugin() calls. Consumers of Mongoose do all sorts of crazy things that might not be best practice and we can't expect everyone to do so. It's more about retaining a reliable API in Mongoose.

How about outputting a debug message that it is being ignored of MONGOOSE_DEBUG=true? 

If you guys feel documentation for the change is enough, I'm fine with it. Let's just make sure we add something to the docs and the changelog. I was just trying to save someone a few hours of debugging when they update their mongoose package...

RobertWHurst:
@Jeff-Lewis Yeah, I see where your coming from on this, it does change API behavior. I do agree that documentation is needed at the very least. Let me mull over this a bit ;)

vkarpov15:
Yeah Jeff is right on this front as usual. I think I'm gonna revert this and instead put this code in a plugin that you can just require() to get this behavior

PR #4111, opened 2016-05-02
Add test for broken subdoc subprop enum validation

tusbar:
This is very similar to #3533.

The validation of subdocuments’ subproperties is broken when the subdocument is created using `MongooseDocumentArray.create` and no hash is passed to that method.

Creating `child` using 

```js
      var child = model.children.create({
        prop: {
          val: 'valid'
        }
      });
```

will make this test pass.

### DIFF
--- a/test/schema.validation.test.js
+++ b/test/schema.validation.test.js
@@ -793,6 +793,41 @@ describe('schema', function() {
       });
     });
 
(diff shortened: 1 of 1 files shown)


tusbar:
Same behavior happens when using 

```js
      model.children.push({});
      var child = model.children[0];
```

instead of creating `child` using `model.children.create()`.

PR #4126, opened 2016-05-05
Updated QueryStream#destroy documentation.

AnthonyCC:
Currently the doc explicitly says no events will be emitted after calling the `destory()` method on stream. However, that is not true and misleading because it does emit the close event.


vkarpov15:
Thanks :beer:

PR #4133, opened 2016-05-08
chore(document): use function reference

dciccale:
`isMongooseObject` was referenced in the top of the file to use `utils.isMongooseObject`, so no need to use `utils.` here


vkarpov15:
Thanks :beers:

Issue #2284, opened 2014-09-07
beautiful mongodb native errors (unique - code 11000)


xpepermint:
I really miss a mechanism that would create validator-like `unique index` error message. Setting the index of an attribute to `unique: true` will tell mongodb to return ugly error message like this:

```
{ [MongoError: insertDocument :: caused by :: 11000 E11000 duplicate key error index: mydb-api.users.$email_1  dup key: { : "my@duplicate.com" }]
  name: 'MongoError',
  code: 11000,
  err: 'insertDocument :: caused by :: 11000 E11000 duplicate key error index: mydb-api.users.$email_1  dup key: { : "my@duplicate.com" }' }
```

This message should be treated as validator error message thus we could easily display error messages on a REST service.

```js
// koa example
index: function*() {
    try {
      var email = new User(body);
      this.body = yield email.save();
    } catch(e) {
      this.body = e; // nice json errors to display
    }
}
```

andriesfilmer:
mongoose-unique-validator is a plugin which adds pre-save validation for unique fields within a Mongoose schema. https://github.com/blakehaswell/mongoose-unique-validator

iliakan:
@vkarpov15 as of now I have a wrapper over mongoose, adding this.

Mongoose-unique-validator is doing it wrong. Pre-save validation is not a real solution here, because only unique index guarantees uniqueness.

user A:   run pre-save, is the value unique? yes it is.
user B:   run pre-save, is the value unique? yes it is.
user B:   insert value
user A:   insert value (BUG!)

vkarpov15:
Yep mongoose-unique-validator is not at all guaranteed to be unique, at least with multiple processes. The "plugin" tag is a reference to OP's issue about beautifying MongoErrors, which would be a valid plugin use case.

xpepermint:
any progress?

vkarpov15:
Nope, but I think you might be able to do this as a plugin...

iliakan:
@xpepermint  I'm doing that by a wrapper around save. No idea what @vkarpov15 means about using a plugin here (?)

vkarpov15:
I think the way to implement this would be a small separate module that does something like this:

```javascript
module.exports = function(schema) {
  schema.methods.trySave = function(callback) {
    this.save(function(error, doc) {
      // Handle error logic here and callback();
    });
  };
};
```

[...]

PR #4139, opened 2016-05-11
Add support for express-style error handling middleware

vkarpov15:
Re: #2284

### DIFF
--- a/lib/drivers/node-mongodb-native/connection.js
+++ b/lib/drivers/node-mongodb-native/connection.js
@@ -2,13 +2,14 @@
  * Module dependencies.
  */
 
-var MongooseConnection = require('../../connection'),
-    mongo = require('mongodb'),
-    Db = mongo.Db,
-    Server = mongo.Server,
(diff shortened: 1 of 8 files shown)


Miyou:
Is this in the latest version? It's not working for me:
```javascript
TestSchema.post('save', function (error, doc, next) {
	console.log(error); //Is actually the document
	next(); //Error "next is not a function"
});
```

vkarpov15:
@Miyou what does TestSchema look like?

Miyou:
Here's an example that reproduces the problem:
```javascript

var TestSchema = new mongoose.Schema({
	name: String,
});

TestSchema.post('save', function (error, doc, next) {
	console.log(error); //Is actually the document
	next(); //Error "next is not a function"
});

var Test = mongoose.model('Test', TestSchema);

var test = new Test({
	name: 'test',
});

test.save();
```

vkarpov15:
Try upgrading to 4.5.3, there was a related issue in 4.5.2

PR #4142, opened 2016-05-11
Path validation is called with the incorrect scope

JedWatson:
Unless I'm mistaken, calling `p.originalRequiredValue` with `this` instead of `_this` doesn't pass the correct scope to custom validation functions.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1271,7 +1271,7 @@ Document.prototype.validateSync = function(pathsToValidate) {
     }
     var p = _this.schema.path(path);
     if (typeof p.originalRequiredValue === 'function') {
-      return p.originalRequiredValue.call(this);
+      return p.originalRequiredValue.call(_this);
     }
     return true;
   });
--- a/lib/document.js
+++ b/lib/document.js
@@ -1144,7 +1144,7 @@ Document.prototype.$__validate = function(callback) {
     }
     var p = _this.schema.path(path);
     if (typeof p.originalRequiredValue === 'function') {
-      return p.originalRequiredValue.call(this);
+      return p.originalRequiredValue.call(_this);
     }
     return true;
   });


JedWatson:
For more context, this is currently breaking Keystone's test suite, as the required method is being called with the global scope: https://travis-ci.org/keystonejs/keystone/builds/129432041

To see how we're using the feature:

https://github.com/keystonejs/keystone/blob/master/fields/types/Type.js#L74-L77

JedWatson:
btw. the error in the CI is a timeout on node.js 5.

Also worth mentioning that making this change locally gets the Keystone test suite to pass again :)

JedWatson:
... and this is the commit where it was broken, 15 days ago: https://github.com/Automattic/mongoose/commit/423dc8f567e21e50e14a16d2f77a7f5387b543cc

airatshigapov:
I can confirm that this PR fixes our tests suite which broke after `4.4.15` upgrade. There definitely should be `_this` instead of `this`.

airatshigapov:
@JedWatson btw there is same issue [few lines below](https://github.com/JedWatson/mongoose/blob/88b10f594639db57fa10dab6f27353e45c8b5f4b/lib/document.js#L1274) (in `validateSync` definition).

JedWatson:
Thanks for that @airatshigapov, I've just updated the PR to address the issue in `validateSync` as well!

varunjayaraman:
Can confirm, after switching `this` to `_this`, the correct scope is used

utvara:
> // or make a path conditionally required based on a function
> var isOver18 = function() { return this.age &gt;= 18; };
> Schema.path('voterRegistrationId').required(isOver18);

This is broken ATM (this within isOver18 is referencing process) and will be fixed by merging this PR.

vkarpov15:
Thanks @JedWatson and everyone else :beers:

JedWatson:
No worries @vkarpov15! Thanks for merging 😄

Issue #4173, opened 2016-05-22
Model.update(...).catch is undefined


adambuczynski:
Consider the following:

```js
    Subscription.update({
      _id: { $not: subscription._id },
      startDate: { $lt: today },
      endDate: { $gt: newEndDate }
    }, {
      $set: { endDate: newEndDate }
    }).catch(errorHandler);
```

This complains that `Subscription.update(...).catch` is undefined.

Whereas calling it after a `save()` operation on an item works fine:

```js
user.save().catch(errorHandler);
```

Also note that I am using Bluebird promises:

```js
mongoose.Promise = require('bluebird');
```

Mongoose version is 4.4.19.

Jokero:
http://mongoosejs.com/docs/promises.html

adambuczynski:
I dislike redundancy :(

Why the need for exec? If there is support for `.then` without exec, why
not `.catch` as well?

Jokero:
As I see `then` implementation is:
```js
Query.prototype.then = function(resolve, reject) {
  return this.exec().then(resolve, reject);
};
```
So we can do the following for `catch`:
```js
Query.prototype.catch = function(reject) {
  return this.exec().then(null, reject);
};
```
Maybe there are some reasons why it can't be implemented like this

adambuczynski:
Yes that would work for me. @vkarpov15 what are your thoughts?

vkarpov15:
We can in theory, hasn't been a big priority because `.catch()` is such a trivial function and such a (relatively) new development.

adambuczynski:
I think that's maybe more of a personal opinion :)

I don't find it trivial at all, and there are many use cases for me to only use `.catch` and not have a `.then`, for example when doing some background updating of data which doesn't need to send back a response, but does need error handling in case it goes wrong.

I find it much cleaner to write `.catch(doSomething)` than something like `.then(null, doSomething)`, as it is much easier to distinguish and separate success handlers and failure handlers that way, and it makes the code easier to read.

It has been around for quite a while now and most promise libraries implement the method, as well as `.finally()`, which in my opinion is also a must have to avoid code duplication in certain circumstances.

If you'd want me to make a PR for this let me know, I'll be happy to look into it.

vkarpov15:
[...]

PR #4215, opened 2016-06-09
Add Query prototype `.catch()` method

adambuczynski:
See #4173

I haven't added a test as I couldn't find one for the `.then()` method either, and it's a pretty trivial implementation. Let me know if anything else needs adding.


vkarpov15:
Much appreciated, thanks for fighting the good fight and getting us better pseudo-promises support :beers:

PR #4222, opened 2016-06-20
Passing undefined to MongooseArray#pull pulls all values

Sebmaster:
Passing undefined (and only undefined; doesn't occur with either no args, nor other args + undefined, but does crash with undefined + other args) results in an empty MongooseArray on the mongo side (but not locally until a refresh).

See this example:
```js

var Cat = mongoose.model('Cat', {
    name: String,
    colors: [
        { hex: String }
    ]
});

var kitty = new Cat({ name: 'Zildjian', colors: [{hex:'#FFF'}, {hex:'#FFF'}] });
kitty.save(function (err) {
  kitty.colors.pull(undefined);
  console.log(kitty) // all fine here
  kitty.save(() => {

    Cat.find({ _id: kitty.id }).then((res) => {
        console.log(res[0]) // array empty
    })
  })
});
```

Output:
```
{ colors: 
   [ { _id: 575b3e792d0799125dea917c, hex: '#FFF' },
     { _id: 575b3e792d0799125dea917b, hex: '#FFF' } ],
  _id: 575b3e792d0799125dea917a,
  name: 'Zildjian',
  __v: 0 }
{ colors: [],
  __v: 1,
  name: 'Zildjian',
  _id: 575b3e792d0799125dea917a }
```

This is not the case in mongoose 3.8.x, not sure when that regressed.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2259,6 +2259,10 @@ Document.prototype.toString = function() {
  */
 
 Document.prototype.equals = function(doc) {
+  if (!doc) {
+    return false;
+  }
+
   var tid = this.get('_id');
   var docid = doc.get ? doc.get('_id') : doc;
   if (!tid && !docid) {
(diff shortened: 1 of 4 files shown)


Sebmaster:
Attached a PR which fixes this issue.

vkarpov15:
Thanks :beers:

PR #4255, opened 2016-06-21
lean() for migrating from stream() to cursor()

pyramation:
this should close https://github.com/Automattic/mongoose/issues/4253

PR #4276, opened 2016-06-27
consider lean option in populate

gchudnov:
It looks #4255 missed a case where populate() was invoked.
Added the missing code.

Please, review.

Thank you,
Grigoriy.


vkarpov15:
Good catch, thanks :beers:

gchudnov:
Thank you!

PR #4294, opened 2016-07-04
Small doc fixes

simonxca:
Connection#global as shown in the docs is undefined
Should be Connection#config


simonxca:
I'm currently writing Typescript definitions for mongoose.  If you want to wait a bit, I'll add other small issues I find as I go through the api docs.

PR #4295, opened 2016-07-04
increase mongodb native client version

RoCat:
increase mongodb version to be able to use  gridFS bucket openUploadStreamWithId new function.


vkarpov15:
Thanks!

vkarpov15:
Next Monday. Do you need it immediately?

PR #4302, opened 2016-07-06
More small doc fixes

simonxca:


lib/query.js
```diff
@@ -3257,15 +3257,6 @@ Query.prototype.centerSphere = function() {
...
 
 /**
- * Executes this query and returns a promise
- *
- * @method then
- * @memberOf Query
- * @return {Promise}
- * @api public
- */
-
```
simonxca: See: http://mongoosejs.com/docs/api.html#query_Query-then

lib/model.js
```diff
@@ -776,14 +776,14 @@ Model.prototype.model = function model(name) {
...
-
 Model.discriminator = function discriminator(name, schema) {
+  var CUSTOMIZABLE_DISCRIMINATOR_OPTIONS = {
+    toJSON: true,
+    toObject: true,
+    _id: true,
+    id: true
+  };
+
   if (!(schema && schema.instanceOfSchema)) {
```
simonxca: see: http://mongoosejs.com/docs/api.html#model_
Moved CUSTOMIZABLE_DISCRIMINATOR_OPTIONS into discriminator() since it's the only place those options are used.

lib/types/buffer.js
```diff
@@ -131,7 +131,7 @@ MongooseBuffer.mixin = {
    *
    * `Buffer#copy` does not mark `target` as modified so you must copy from a `MongooseBuffer` for it to work as expected. This is a work around since `copy` modifies the target, not this.
    *
-   * @return {MongooseBuffer}
+   * @return {Number} The number of bytes copied.
```
simonxca: Buffer.prototype.copy() returns number of bytes copied.

vkarpov15:
Thanks :beers:

PR #4309, opened 2016-07-08
fix embededDocument removal

RoCat:
fix documentation for new 'options' parameter, and set it as optional one


vkarpov15:
Thanks :beers:

PR #4327, opened 2016-07-13
Revert "Merge pull request #4295 from RoCat/master"

Kelvyne:
This reverts commit 194cddaeeed4bffc54d87cf868c98ec4d0580b50, reversing
changes made to d7ca0dc1e150faf18c6b47ebdef5dc8c07c6a238.

https://github.com/mongodb/node-mongodb-native/blob/2.2/HISTORY.md#2119-2016-05-17

It seems to break things in Mongoose.
I think the right thing to do here is to revert the commit, and then investigate the problem

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "bson": "~0.4.23",
     "hooks-fixed": "1.1.0",
     "kareem": "1.1.3",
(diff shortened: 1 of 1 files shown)


vkarpov15:
Merged an equivalent PR, #4334

PR #4344, opened 2016-07-19
ownerDocument() now works when nesting arrays and single embedded doc…

### LINKED ISSUE #3589: Reference model from within validator on nested schema
I'm trying to find a way to get a reference to the parent model of a nested schema from inside the `validator`.

I maintain a plugin which queries a collection for unique records before mongo returns a duplicate key error, so these validators are built dynamically.

I have a feeling it's not possible, and I need to check `schema.nested` and use some custom logic for these nested documents.

When the validator is called on a nested schema, the validator's `this` is `Document`:

    // this.constructor.name = "Document"
    // typeof this.model = "undefined"

Yet for a validator directly on the parent, it's a `model`:

    // this.constructor.name = "model"
    // typeof this.model = "function"

An example schema:

    var ContactSchema = new mongoose.Schema({
        email: {
            type: String,
            index: true,
            unique: true
        }
    });
    ContactSchema.plugin(uniqueValidator);

    var CompanySchema = new mongoose.Schema({
        name: String,
        contact: ContactSchema
    });
    // CompanySchema.plugin(uniqueValidator);

vichle:
…s (Fix #3589)

Fixes the issues that I brought up yesterday in #3589

### DIFF
--- a/lib/types/embedded.js
+++ b/lib/types/embedded.js
@@ -254,8 +254,8 @@ EmbeddedDocument.prototype.ownerDocument = function() {
     return this;
   }
 
-  while (parent.__parent) {
-    parent = parent.__parent;
+  while (parent.__parent || parent.$parent) {
+    parent = parent.__parent || parent.$parent;
   }
 
   this.$__.ownerDocument = parent;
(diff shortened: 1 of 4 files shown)


vichle:
BTW, a question: What's the reason for EmbeddedDocument having a __parent property while Subdocument has a $parent property?

vkarpov15:
@vichle no particular reason, inconsistency in the code. Thanks for fixing :+1:

PR #4351, opened 2016-07-20
Make sure that createdAt does not get overriden

### LINKED ISSUE #4340: Timestamps: createdAt is overridden when it is not selected
Hey,

Currently, if I `.save()` a model where `createdAt` was not selected, the creation date is overridden (https://github.com/Automattic/mongoose/blob/master/lib/schema.js#L756-L758)

```js
return Model
    .findById('xxx')
    .select({
        _id: true
    })
    .exec()
    .then(function (model) {
        // here model.createdAt is undefined
        return model.save();
    })
    .then(function (model) {
        // createdAt is now overridden :(
    });
```

For reference: https://github.com/drudge/mongoose-timestamp/pull/34 (I only discovered recently the existence of the { timestamps } option :))

@vkarpov15 what do you think about checking if `createdAt` was selected before overriding the prop?

tusbar:
`createdAt` should always represent the document’s creation date. Whenever
`createdAt` is not selected (so `undefined`) and we update the document – it
should not be set to a new value.

Fix #4340

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -753,7 +753,7 @@ Schema.prototype.setupTimestamp = function(timestamps) {
       var defaultTimestamp = new Date();
       var auto_id = this._id && this._id.auto;
 
-      if (!this[createdAt]) {
+      if (!this[createdAt] && this.isSelected(createdAt)) {
         this[createdAt] = auto_id ? this._id.getTimestamp() : defaultTimestamp;
       }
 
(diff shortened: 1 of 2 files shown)


tusbar:
(Timeout fail on node 0.12)

vkarpov15:
Thanks :beers:

tusbar:
🍻

PR #4366, opened 2016-07-26
Update subdocs.html

gulatichayan:
Old:
var doc = parent.children.id(id)
New:
var doc = parent.children.id(_id)
At line number: 19 & 33


vkarpov15:
Thanks but please update subdoccs.jade, the subdocs.html file is compiled code so your changes will just get overwritten

gulatichayan:
Yeah sure

PR #4363, opened 2016-07-23
4.6

vkarpov15:
Fix issues from #4341

PR #4439, opened 2016-08-20
should not cause unhandle reject promise when create fail

isayme:
When create a instance with a callback, an unhandled reject promise event will triggered if create failed.

```
var TaskSchema = new Schema({
  name: {
    type: String,
    maxlength: 4
  }
})

var Task = mongoose.model('Task', TaskSchema)

Task.create({
  name: 'too long name'
}, function (err, task) {
  // here will have an err.
  console.log(err)
})

process.on('unhandledRejection', function (reason, promise) {
  // and there will a event.
  console.log(reason)
})
````

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1858,8 +1858,11 @@ Model.create = function create(doc, callback) {
 
     async.parallel(toExecute, function(error, savedDocs) {
       if (error) {
-        cb && cb(error);
-        reject(error);
+        if (cb) {
+          cb(error);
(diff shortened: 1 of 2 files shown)


vkarpov15:
Thanks :beers:, will merge tomorrow

vkarpov15:
Sorry for the delay, was waiting on #4363 because I wanted to release this as part of 4.6

isayme:
🍻

PR #4444, opened 2016-08-22
doc: fix typo on documents.jade

Gabri3l:


Gabri3l:
I'm sorry I am not sure why the checks fail, not very familiar with Travis CI. I did try to simply fix the typo from the jade file and the checks failed, although the html is compiled I tried editing that too but it still fails the checks. Is there something I'm missing ? Sorry for this nonsense.

vkarpov15:
Thanks :+1: don't worry about travis, it's pretty flakey, their infrastructure doesn't work too well with mongodb.

PR #4471, opened 2016-08-30
docs(document): clarify the findById and findByIdAndUpdate examples

mdcanham:
The current example given for the 'findById' method places the updated document in the callback function while the 'findByIdAndUpdate' method places the original document in the callback. To keep things consistent they should both return the updated document which is returned from the db.

Wasn't sure which branch to put this on, happy to move it.


vkarpov15:
Thanks :beers:

PR #4512, opened 2016-09-12
Fix webpack build

### LINKED ISSUE #3578: Using isomorphic mongoose with webpack
I am getting the following warning and then a bunch of "Module not found errors" when packing my project:
```
WARNING in ./~/mongoose/lib/drivers/index.js
Critical dependencies:
9:11-74 the request of a dependency is an expression
 @ ./~/mongoose/lib/drivers/index.js 9:11-74

WARNING in ./~/mongoose/lib/drivers/SPEC.md
Module parse failed: C:\development\banquetninja\node_modules\mongoose\lib\drivers\SPEC.md Line 2: Unexpected token ILLEGAL
You may need an appropriate loader to handle this file type.
| 
| # Driver Spec
| 
| TODO
 @ ./~/mongoose/lib/drivers ^\.\/.*$
```
Webpack seems to failing on the following code in ./~/mongoose/lib/drivers/index.js:

```
var driver;

if (typeof window === 'undefined') {
  console.log('global: ' + global.MONGOOSE_DRIVER_PATH);
  driver = require(global.MONGOOSE_DRIVER_PATH || './node-mongodb-native');
} else {
  driver = require('./browser');
}
```

If I force driver = require('./browser') and skip the window conditional everything works but this is a horrible hack.  I believe the answer lies in webpack.config but have been unable to find a solution.

Thanks for any help you can provide.

### LINKED ISSUE #3713: Cannot find mongoose module with webpack
Getting the following error with webpack and mongoose - `Uncaught Error: Cannot find module '././binary_parser'`

Below is my webpack.config.js and the mongoose version I am using is 4.3.3
```
var path = require('path');
var webpack = require('webpack');
var config = require('./tasks/config');

module.exports = {
  context: path.join(__dirname, 'src'),
  devtool: 'eval',
  historyApiFallback: true,
  entry: {
    app: [
      'webpack-hot-middleware/client',
      './' + config.src.main
    ]
  },
  target: 'node',
  output: {
    path: path.join(__dirname, config.dest),
    filename: 'bundle.js',
    publicPath: '/static/'
  },
  node: {
    fs: 'empty',
    net: 'empty',
    tls: 'empty'
  },
  plugins: [
    new webpack.DefinePlugin({
      'process.env': {
        NODE_ENV: JSON.stringify(process.env.NODE_ENV || 'development')
      }
    }),
    new webpack.HotModuleReplacementPlugin(),
    new webpack.IgnorePlugin(/vertx/),
    new webpack.NoErrorsPlugin()
  ],
  resolve: {
    alias: {
      'redux': path.join(__dirname, 'node_modules/redux')
    },
    extensions: ['', '.js', '.json', '.node'],
    fallback: path.join(__dirname, 'node_modules')
  },
  resolveLoader: { fallback: path.join(__dirname, 'node_modules') },
  module: {
    loaders: [{
      test: /\.js$/,
      loaders: ['react-hot', 'babel'],
      exclude: /node_modules/
    }, {
      test: /\.scss$/,
      loaders: ['style', 'css', 'sass'],
      include: __dirname
    }, {
      test: /\.html$/,
      exclude: /node_modules/,
      loader: 'file?name=[path][name].[ext]'
    }, {
      test: /\.json$/,
      loader: 'json'
    }, {
      test: /\.node$/,
      loader: 'node'
    }]
  }
};
```

The error is being thrown here 
```
function webpackContextResolve(req) {
	return map[req] || (function() { throw new Error("Cannot find module '" + req + "'.") }());
};
```

Can someone please point me in the right direction????

saiichihashimoto:
With this fix, we can have the browser implementation be compatible with webpack.

Solution derived from @patterncoder's ["hack"](https://github.com/Automattic/mongoose/issues/3578#issuecomment-166437240).

The "solutions" in related issues haven't been satisfactory. The current `lib/drivers/index.js`'s `typeof window === 'undefined'` check doesn't satisfy webpack without extraneous (and irrelevant to mongoose in the browser) loaders. Requiring a separate `<script>` tag also breaks the point of webpack. While webpack isn't a specified case mongoose supports, this one file would handle the problem entirely.

Resolves webpack/webpack#1040
Resolves webpack/webpack#2566
Resolves #3713
Resolves #3578

### DIFF
--- a/lib/document_provider.web.js
+++ b/lib/document_provider.web.js
@@ -1,0 +1,17 @@
+'use strict';
+
+/* eslint-env browser */
+
+/*!
+ * Module dependencies.
+ */
+var BrowserDocument = require('./browserDocument.js');
+
+/**
+ * Returns the Document constructor for the current context
+ *
+ * @api private
+ */
+module.exports = function() {
+  return BrowserDocument;
+};
--- a/lib/drivers/index.web.js
+++ b/lib/drivers/index.web.js
@@ -1,0 +1,1 @@
+module.exports = require('./browser');


saiichihashimoto:
The build failure is the same failure that is currently on `master`, so deciding to ignore it.

vkarpov15:
Thanks. TBH have not tried mongoose with webpack at all, I don't really expect it to work particularly well.

varunjayaraman:
I'm not sure if this fixed the issue? The problem still occurs if you `require('mongoose')` because webpack can't statically analyze `if (typeof window === 'undefined')` in `mongoose/lib/drivers/index.js`:

Here's my take on the issue (and correct me if I've got this wrong @saiichihashimoto):

Webpack can't statically analyze binary files so in `mongoose/lib/drivers/index.js`, the fact that `node-mongodb-native` is getting required in, which in turn imports the `mongodb` bson binary file, causes an issue in webpack's bundling:

```
WARNING in ./~/mongoose/lib/drivers/index.js
8:11 Critical dependency: the request of a dependency is an expression
```

Looking at the file:

```
/*!
 * ignore
 */

var driver;

if (typeof window === 'undefined') {
  driver = require(global.MONGOOSE_DRIVER_PATH || './node-mongodb-native');
} else {
  driver = require('./browser');
}

/*!
 * ignore
 */

module.exports = driver;
```

Webpack doesn't actually run your code so it can't properly go down the right dependency path here. 

How exactly did this solution prevent that problem @saiichihashimoto ?

vkarpov15:
I'd be curious myself. @varunjayaraman are y'all switching off of browserify and onto webpack?

varunjayaraman:
@vkarpov15 yep. Combination of npm scripts and webpack. Was initially turned off by webpack's steep learning curve and seemingly excessive complexity but now I'm a proponent. 

Also, it seems a simple solution (albeit a hack) to get webpack working with mongoose is to add a plugin to webpack's array of plugins:

```
new webpack.DefinePlugin({
   'typeof window': "\"object\""
})
```

This enables webpack to sort of understand the dynamic require in `mongoose/lib/drivers/index.js`.

vkarpov15:
Please do, that would be super helpful

PR #4531, opened 2016-09-16
fix #3719

### LINKED ISSUE #3719: "Cast to ObjectId failed" for custom ID and $elemMatch/$not
Here is an example schema

```
var child = new mongoose.Schema({
	_id: {type: String}
});

var parent = new mongoose.Schema({
	children: [child]
});

```
The following query has been abbreviated to exclude the non-relevant info (my query is quite large). When $elemMatch is used without $not, it works fine and can cast the _id. However, when placed within a $not{} tag, it tries to cast as ObjectID and fails.

```
model.find({
	$and: [{
		$or: [{
			$and: [
				{
					myField: {
						$not: {
							$elemMatch: {
								_id: 'customString'
							}
						}
					}
				}
			]
		}]
	}]
})
```

timbowhite:
I whipped this PR together, seems to pass all tests.  Note that I have only given the change and its code context a cursory review.

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -203,7 +203,9 @@ module.exports = function cast(schema, obj) {
             }
 
             if ($cond === '$not') {
-              cast(schema, nested);
+              //cast(schema.paths[path], nested);
+              //cast(schematype.caster.schema, nested); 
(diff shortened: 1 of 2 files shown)


vkarpov15:
Thanks :beers:

PR #4541, opened 2016-09-19
Read Model from this

Nainterceptor:
Hi,

I'm looking for a workaround on an issue (on compound primary key). I think I have a Quick & Dirty solution, so I want to override Model.

The following PR allow that by using the Mongoose.prototype.Model = Model; (line 674) and not the require (line 13)

Regards,

Gaël - HungryUp


vkarpov15:
thanks :beers:

PR #4579, opened 2016-09-29
upgrade async to 2.0.1

billouboq:
Async 2 gives more modularity and performance.
[async changelog](https://github.com/caolan/async/blob/master/CHANGELOG.md)

Was unable to test properly with npm Test, with or without the change.


vkarpov15:
Thanks :beers:

Issue #4592, opened 2016-10-03
Can't use $not with String


mdlavin:
Starting with 4.6.2, the following query stopped being allowed:

`{ from: {$not: /test/}}` where `from` is a string field.

The error is:

    Error: Can't use $not with String.
      at SchemaString.castForQuery (node_modules/mongoose/lib/schema/string.js:494:13)
      at cast (node_modules/mongoose/lib/cast.js:214:43)
      at Query.cast (node_modules/mongoose/lib/query.js:2685:10)
      at Query.find (node_modules/mongoose/lib/query.js:1139:10)
      at node_modules/mongoose/lib/query.js:2244:21
      at Query.exec (node_modules/mongoose/lib/query.js:2238:17)

It looks like it might be releated to #4495.

KATT:
+1

PR #4616, opened 2016-10-10
Fixed issue #4592

prssn:
Added $not to SchemaString.prototype.$conditionalHandlers.


vkarpov15:
Fixed in master but this is actually a better fix, thanks :beers:

PR #4619, opened 2016-10-11
Fix for array field false default values

EricMcRay:
When you set default value any false value like 0, false or empty string, it was return empty array. And you need to give that default value in function to hack that. It was because weird undefined check. I hope this PR won’t break anything. It passes tests.


EricMcRay:
Wrong commit, sorry

PR #4635, opened 2016-10-19
Punctuation added

silvermanj7:
I added an apostrophe to denote MongoDB's ownership of "rich querying syntax." Originally just a small punctuation oversight. Fixed.
One more error: I see that Line 28 of docs/index.html's first paragraph tag is not displaying the same on the website as it is coded here. The way it is coded here is correct, but http://mongoosejs.com/docs/index.html shows, 
"This performs a search for all documents with a name property that begins with 'Fluff' and returns the result as an array of kittens to the callback" 
instead of, 
"This performs a case-insensitive search for all documents with a name property containing "fluff" and returns the result as an array of kittens to the callback." I'd suggest using the latter sentence if you can.


vkarpov15:
Thanks but please change the `.jade` file, the `.html` file is compiled code, so it'll just get overwritten

Issue #4629, opened 2016-10-16
[BUG] Array of references returned empty on discriminated schema when querying child schema.


danpe:
This might be related to #4073

``` javascript
    var entityDefinition = {
      team: { type: mongoose.Schema.ObjectId, ref: 'Team', required: true },
      teams: [{ type: Schema.ObjectId, ref: 'Team', required: true }],
      updated: { type: Date, default: Date.now },
      created: { type: Date, default: Date.now }
    };

    // define the schema for our entity model
    var entitySchema = new mongoose.Schema(entityDefinition, { discriminatorKey: 'type' });
    var Entity = mongoose.model('Entity', entitySchema);

    var userDefinition = {
      name: { type: String, required: true },
    };

    // define the schema for our user model
    var userSchema = new mongoose.Schema(userDefinition, { discriminatorKey: 'type' });
    var User = Entity.discriminator('User', userSchema);
```

And when I'm trying to select a User and populate it's teams array (from **Entity Schema**) it fails...

``` javascript
    User.findOne({ 'entityid': userid }, 'name team teams').populate('team teams').exec(function (err, user) {
        console.log(user.name); // prints: "danpe"
        console.log(user.team.name); // prints: "General"
        console.log(user.teams); // prints: undefined (even though it is not empty when I'm inspecting the mongo model directly using Robomongo)
        console.log(user.$__.populated.teams.value) // prints an array of ObjectIds, which is correct, but not populated into the main object..
    }
```

If I'm printing the `user` on debug I get the following:

``` javascript
    { '$__':
       { strictMode: true,
         selected: { team: 1, teams: 1, type: 1, name: 1 },
         shardval: undefined,
         saveError: undefined,
         validationError: undefined,
         adhocPaths: undefined,
         removing: undefined,
         inserting: undefined,
         version: undefined,
         getters: {},
         _id: undefined,
         populate: undefined,
         populated: { teams: [Object], team: [Object] },
         wasPopulated: false,
         scope: undefined,
         activePaths: { paths: [Object], states: [Object], stateNames: [Object] },
         ownerDocument: undefined,
         fullPath: undefined,
         emitter: { domain: null, _events: {}, _eventsCount: 0, _maxListeners: 0 } }
    ,
      isNew: false,
      errors: undefined,
      _doc:
       { profile: {},
         teams: undefined,
         type: 'User',
         name: 'danpe',
         team:
          { '$__': [Object],
            isNew: false,
            errors: undefined,
[...]

PR #4643, opened 2016-10-20
#4629 Added test to reproduce

danpe:
#4629 - This test reproduces the issue of array of references returns empty when querying 'child' discriminated schema.


vkarpov15:
Upon further investigation, your issue is that you're creating a new vehicle with a reference to the wheel, and then a completely separate document with no reference to the wheel, and expecting that populating the separate document gives you a reference to the wheel. See ca46e04 for a working example :beers:

PR #4653, opened 2016-10-25
reconnectFailed error from driver not being caught

loris:
It looks like the error message emitted by the mongo driver when auto reconnect mechanism fails has been changed recently (https://github.com/christkv/mongodb-core/blob/319f26dcbdf71a13f234ca33aac2a49642984b5b/lib/connection/pool.js#L307) from `...after x retries...` to `...after x attempts...`, thus the regex used to caught the error and emit a mongoose `error` is not matching anymore

### DIFF
--- a/lib/drivers/node-mongodb-native/connection.js
+++ b/lib/drivers/node-mongodb-native/connection.js
@@ -60,7 +60,7 @@ NativeConnection.prototype.doOpen = function(fn) {
 
     if (!mongos) {
       server.s.server.on('error', function(error) {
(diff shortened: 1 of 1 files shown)


vkarpov15:
Good catch, thanks. This regexp is a bit of a hack unfortunately

amit777:
curious if there is a simple or elegant way to add this check to a test suite?

vkarpov15:
Not with mongoose's current testing paradigm unfortunately, because the mongoose test suite doesn't have control over the mongod process that it's testing against. There are ways, just too tricky to justify at the moment.

PR #4668, opened 2016-10-30
Integration for ES6 classes

rockmacaca:
The reason for this PR is that I noticed that mongoose under the hood is essentially acting as a mixin to a class, and I saw room for improvement with the syntax for defining models. Generally static and instance methods have to be inconveniently defined through helpers on a `Schema` before they are copied onto the model, and `Model.compile` is also being inefficient by manually setting the prototype. So essentially, mongoose is doing work that ES6 already natively supports.

With these changes `mongoose.model` can support taking an ES6 class where `name` is normally provided:

``` javascript
const { Model, Schema } = mongoose;

const schema = new Schema({
  _id: Schema.ObjectId,
  field1: Number,
  field2: String
});
class SomeThing extends Model {
  static explode() {
    return SomeThing.explode();
  }

  get numString() {
    return this.field1 + this.field2;
  }

  increment() {
    return SomeThing.update({ _id: this._id }, {
      $inc: { 'field1': 1 }
    }, { new: true }).exec();
  }
}

module.exports = mongoose.model(SomeThing, schema, 'some_things');
```

I've also included a `loadClass` instance method for schemas that accomplishes some of what [`mongoose-class-wrapper`](https://github.com/aksyonov/mongoose-class-wrapper) can do. It allows similar functionality as the above, but it's useful for sub-schema definitions.

This should be fully backward compatible and shouldn't introduce any breaking changes.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -309,14 +309,23 @@ Mongoose.prototype.disconnect.$hasSideEffects = true;
  *     var collectionName = 'actor'
  *     var M = mongoose.model('Actor', schema, collectionName)
  *
- * @param {String} name model name
+ * @param {String|Function} name model name or class extending Model
  * @param {Schema} [schema]
- * @param {String} [collection] name (optional, induced from model name)
+ * @param {String} [collection] name (optional, inferred from model name)
  * @param {Boolean} [skipInit] whether to skip initialization (defaults to false)
  * @api public
  */
 
 Mongoose.prototype.model = function(name, schema, collection, skipInit) {
+  var model;
+  if (typeof name === 'function') {
+    model = name;
+    name = model.name;
+    if (!(model.prototype instanceof Model)) {
+      throw new mongoose.Error('The provided class ' + name + ' must extend Model');
+    }
+  }
+
   if (typeof schema === 'string') {
     collection = schema;
     schema = false;
@@ -354,7 +363,6 @@ Mongoose.prototype.model = function(name, schema, collection, skipInit) {
     this._applyPlugins(schema);
   }
 
-  var model;
   var sub;
 
   // connection.model() may be passing a different schema for
@@ -393,7 +401,7 @@ Mongoose.prototype.model = function(name, schema, collection, skipInit) {
   }
 
   var connection = options.connection || this.connection;
-  model = this.Model.compile(name, schema, collection, connection, this);
+  model = this.Model.compile(model || name, schema, collection, connection, this);
 
   if (!skipInit) {
     model.init();
--- a/lib/model.js
+++ b/lib/model.js
@@ -3190,7 +3190,7 @@ Model._getSchema = function _getSchema(path) {
 /*!
  * Compiler utility.
  *
- * @param {String} name model name
+ * @param {String|Function} name model name or class extending Model
  * @param {Schema} schema
  * @param {String} collectionName
  * @param {Connection} connection
@@ -3207,19 +3207,28 @@ Model.compile = function compile(name, schema, collectionName, connection, base)
     schema.add(o);
   }
 
-  // generate new class
-  function model(doc, fields, skipId) {
-    if (!(this instanceof model)) {
-      return new model(doc, fields, skipId);
-    }
-    Model.call(this, doc, fields, skipId);
+  var model;
+  if (typeof name === 'function' && name.prototype instanceof Model) {
+    model = name;
+    name = model.name;
+    schema.loadClass(model, true);
+  } else {
+    // generate new class
+    model = function model(doc, fields, skipId) {
+      if (!(this instanceof model)) {
+        return new model(doc, fields, skipId);
+      }
+      Model.call(this, doc, fields, skipId);
+    };
   }
 
   model.hooks = schema.s.hooks.clone();
   model.base = base;
   model.modelName = name;
-  model.__proto__ = Model;
-  model.prototype.__proto__ = Model.prototype;
+  if (!(model.prototype instanceof Model)) {
+    model.__proto__ = Model;
+    model.prototype.__proto__ = Model.prototype;
+  }
   model.model = Model.prototype.model;
   model.db = model.prototype.db = connection;
   model.discriminators = model.prototype.discriminators = undefined;
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1504,6 +1504,45 @@ Schema.prototype.remove = function(path) {
   }
 };
 
+/**
+ * Loads an ES6 class into a schema. Maps setters + getters, static methods, and instance methods to schema virtuals, statics, and methods.
+ *
+ * @param {Function} model
+ */
+Schema.prototype.loadClass = function(model) {
+  if (model === Object.prototype || model === Function.prototype) {
+    return this;
+  }
+
+  // Add static methods
+  Object.getOwnPropertyNames(model).forEach(function(name) {
+    if (name.match(/^(length|name|prototype)$/)) {
+      return;
+    }
+    var method = Object.getOwnPropertyDescriptor(model, name);
+    if (typeof method.value === 'function') this.static(name, method.value);
+  }, this);
+
+  // Add methods and virtuals
+  Object.getOwnPropertyNames(model.prototype).forEach(function(name) {
+    if (name.match(/^(constructor)$/)) {
+      return;
+    }
+    var method = Object.getOwnPropertyDescriptor(model.prototype, name);
+    if (typeof method.value === 'function') {
+      this.method(name, method.value);
+    }
+    if (typeof method.get === 'function') {
+      this.virtual(name).get(method.get);
+    }
+    if (typeof method.set === 'function') {
+      this.virtual(name).set(method.set);
+    }
+  }, this);
+
+  return (this.loadClass(Object.getPrototypeOf(model)));
+};
+
 /*!
  * ignore
  */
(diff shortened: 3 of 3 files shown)


rockmacaca:
Just looking ahead even further, eventually the Schema constructor could be an `@Schema` class decorator if we opt to support Babel. This would mean we don't have to manually call mongoose.model at the end and we could have

```
@Schema({
  _id: Schema.ObjectId,
  field1: Number,
  field2: String
}, { collection: 'some_things' })
export class SomeThing extends Model { ... }
```

which IMO looks very nice.

lib/schema.js
```diff
@@ -1504,6 +1504,49 @@ Schema.prototype.remove = function(path) {
...
+ *
+ * @param {Function} model
+ */
+Schema.prototype.loadClass = function(model, virtualsOnly) {
+  if (model === Object.prototype || model === Function.prototype) {
+    return this;
+  }
+
+  // Add static methods
+  if (!virtualsOnly) {
```
vkarpov15: Hmm what's the point of this virtualsOnly option? Looks like it's always true anyway...

rockmacaca: I suppose it's probably unnecessary. Its only use is [mainly internal by `Model.compile`](https://github.com/Automattic/mongoose/pull/4668/files/27d2342456bf366706842e2510db9f1148f52322#diff-a140b60b6a99a8c69f4b5d0d1f1f2bfaR3214). That's also why I didn't add JSDoc for the parameter.

It seems like Schemas *need* the virtuals defined on them for them to function properly, e.g. in `toObject` and `toJSON`. It just felt wasteful copying all of the statics and instance methods from the class onto the schema when `applyMethods(model, schema)` and `applyStatics(model, schema)` are being called immediately afterward.

lib/index.js
```diff
@@ -309,14 +309,23 @@ Mongoose.prototype.disconnect.$hasSideEffects = true;
...
  * @param {Boolean} [skipInit] whether to skip initialization (defaults to false)
  * @api public
  */
 
 Mongoose.prototype.model = function(name, schema, collection, skipInit) {
+  var model;
+  if (typeof name === 'function') {
+    model = name;
+    name = model.name;
+    if (!(model.prototype instanceof Model)) {
```
mleanos: Correct me if I'm wrong.. I don't think this would work in the case when using a Base class on an individual class.

```javascript
export class Foo extends MyBase {...}
```

mleanos:
I'm very interested in having this feature. I would like to decouple my application's classes from my data access layer. If I can provide an es6 class in lieu of the Mongoose Schema definition, then I can effectively use the same classes with a different database technology.

Without this PR's implementation, is there a recommend way of achieving this?

lib/index.js
rockmacaca: No, `instanceof` looks through the whole prototype chain. So as long as `MyBase` extends `Model`, it's fine. I'm actually currently using this use of `extends` for some of the models in my app.

lib/index.js
```diff
@@ -393,7 +401,7 @@ Mongoose.prototype.model = function(name, schema, collection, skipInit) {
   }
 
   var connection = options.connection || this.connection;
-  model = this.Model.compile(name, schema, collection, connection, this);
+  model = this.Model.compile(model || name, schema, collection, connection, this);
```
rockmacaca: Not sure where the problem you see is, can you elaborate? `model` is unset if the `name` passed in is a string, which means that `Model.compile` receives a string name rather than an ES6 class, and everything proceeds normally.

rockmacaca:
@mleanos Yeah, that is the direction I'd like to see mongoose go.

This is mostly attainable using the [`mongoose-class-wrapper`](https://github.com/aksyonov/mongoose-class-wrapper) plugin, but obviously it's less convenient and uglier, and IDEs that support ES6 don't aren't aware of the methods provided by `Model`.

lib/index.js
mleanos: You're right. I just saw the diff and thought it wasn't necessary based on my misunderstanding of what the `compile` method was doing.

lib/index.js
mleanos: Thank you for clarifying that. I didn't know that about `instanceof`; obviously :)

lib/index.js
mleanos: When I first looked at this, I thought this bit was still inside the `if (typeof name === 'function')..` scope.

malixsys:
@mleanos this is now merged and released in 4.7

malixsys:
@rockmacaca 👏 👏 👏 Thank you for this!

TrejGun:
🎉

vkarpov15:
:tada:

klimashkin:
Is there docs for this feature?
How to deal with `path`,  `pre`, etc?

rockmacaca:
@klimashkin I suppose there probably should be an example like the one I initially gave in the API docs. This shouldn't affect anything with `path` or `pre`, since those are for schemas.

However, depending on how you're set up, you might be able to natively add pre and post hooks like this:

```javascript
class MyClass extends Model {
  save() {
    doSomethingPreSave();
    return super.save()
      .then((newDoc) => {
        doSomethingPostSave();
      });
    }
  }
```

aksyonov:
Awesome :tada:

alexis-y:
This is great, I've been looking how to do this for ages!

However I can't get it to work with discriminators/schema extension.

Here's what I have (roughly):
```javascript
mongoose.model(
  class LoginBase extends Model {}, 
  new Schema({ enabled: Boolean }, { collection: 'logins', discriminatorKey: '_type' })
);

LoginBase.discriminator(
  class LocalLogin extends LoginBase {},
  new Schema({ email: String, password: String })
);
```

This throws `Virtual path "__v" conflicts with a real path in the schema`. If I disable versioning, then it's "_id" that fails. At that point it's clear either I've got something backwards, or this use case isn't supported (in which case I'd be glad to help out)

vkarpov15:
@alexis-y can you please open up a separate issue? Hard to track issues when new issues are nested underneath old ones :)

malcommac:
Is there a doc to see how it should work?

vkarpov15:
http://mongoosejs.com/docs/advanced_schemas.html

NickBolles:
I assume this will work with typescript classes, but I don't see any mention of it. Will this work? Also as a side question, can you define the constructor in the class to extend the base Model constructor? Again it would seem so.

Emiya0306:
> Just looking ahead even further, eventually the Schema constructor could be an @Schema class decorator if we opt to support Babel. This would mean we don't have to manually call mongoose.model at the end and we could have something like

@vkarpov15 I'm very like the code style of using schema decorator which @specialkk metioned. Not only more graceful syntax, but also more friendly to IDE. Is this feature released in mongoose?

everythingspirals:
Why I use the syntax from the original post, i.e:

`mongoose.model(UserSession, schema, 'user.session');`

 I get the following error:

MissingSchemaError: Schema hasn't been registered for model "user.session".

However, if I revert to:

`mongoose.model('user.session', schema, 'user.session');`

I no longer get this error.

lib/model.js
```diff
@@ -3207,19 +3207,28 @@ Model.compile = function compile(name, schema, collectionName, connection, base)
...
-  // generate new class
-  function model(doc, fields, skipId) {
-    if (!(this instanceof model)) {
-      return new model(doc, fields, skipId);
-    }
-    Model.call(this, doc, fields, skipId);
+  var model;
+  if (typeof name === 'function' && name.prototype instanceof Model) {
+    model = name;
+    name = model.name;
```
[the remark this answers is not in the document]
NickBolles: Because `type of name=== 'function'` and `model = name` model is a function. This function.prototype.name is what it's calling. So it would be "myFunction" in the below function `function my function() {}`

With a class this is the name of the class.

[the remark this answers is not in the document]
everythingspirals: What if we want our model to have a name different than the class name, i.e if our model is name is 'my.function'?

We have to override it? i.e
```
class myFunction {
     static get name(){
          return 'my.function';
     }
}
```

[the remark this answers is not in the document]
NickBolles: name is a read-only property, so you can't change it
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name

I suggest you look at the link that vkarpov15 commented to use .LoadClass into the schema you want

benjamin-halter:
@everythingspirals the fix is on line 3767 of model.js

Change to `schema.loadClass(model);`

Hesitant to submit a pull request as not 100% sure on the inner workings. Perhaps a contributor eg. @vkarpov15 could reopen and review this?

benjamin-halter:
@everythingspirals problem lies in index.js:390. Specifically:

```
if (options.cache === false) {
    return model;
}
```

Removing this allow it to work as expected. Not 100% positive on submitting a pull request as I'm sure there's a deeper issue but I myself do not have time to look into it. Perhaps a contributor like @vkarpov15  could look into this?

vkarpov15:
@benjamin-halter can you provide a script that demonstrates the issue you're seeing? That'll make it much easier for me to look into this.

abettke:
> http://mongoosejs.com/docs/advanced_schemas.html

In regards to this in case anyone else runs into this. If you are using a model class to define document methods, use the `schema.loadClass()` method to correctly integrate it into the schema. Not doing so creates some weird behavior with schemas hooks being registered multiple times.

```
class Test extends mongoose.Model {
    ....
}
const schema = new mongoose.Schema('testschema', {
	...
})
  /** Middleware */
  .pre('save', async function () {
     ... some stuff
  })
  .loadClass(Test); // GOOD

-----------

mongoose.model(Test, schema); // BAD, pre-save middleware calls twice
```

Issue #4671, opened 2016-11-02
setDefaultsOnInsert causes high CPU utilization unnecessarily


CamHenlin:
There is unnecessary code in setDefaultsOnInserts (most of lines 25 - 49, in fact!) that can be extremely expensive in terms of node process utilization when updating documents with large numbers of properties. That code should be cut down to only the essentials and moved inside of the conditional currently located on line 51 for big performance gains. I have a PR fixing this issue here: https://github.com/Automattic/mongoose/pull/4672

vkarpov15:
Merged the PR, thanks for catching this :+1:

PR #4672, opened 2016-11-02
Fix high process utilization issue #4671

CamHenlin:
Removes expensive code used to set unused variables at the top of setDefaultsOnInsert, and moves all of the remaining code in to the conditional for options.upsert, which is the only case it would apply in.


CamHenlin:
by the way, the travis CI tests do not appear to work consistently

vkarpov15:
Good catch, thanks :+1: Re: tests, yeah, they're very flakey on Travis, not sure why but we're gonna move off of travis and onto something more reliable.

Issue #4705, opened 2016-11-11
transform option for cursor()


cblanc:
I understand mongoose is on course to deprecate `.stream()` in favour of `.cursor()`

Is there a option of transforming documents returned by `cursor()` which is analogous to the way  we could transform documents returned by `stream()` with something like `Thing.find().stream({ transform: JSON.stringify });`

I'm just wondering if this omission was a deliberate design decision because this should be approached some other way (presumably transform streams?). If not, would you accept a PR which implements a `transform` option for `cursor()`?

vkarpov15:
Yeah it was an omission because it didn't seem too important at the time, especially since, as you mentioned, transform streams make this doable. Unfortunately, transform streams are confusing because the node stream API is not very clean or well documented. What I'd love to see would be a `.map()` function for cursors, which would go nicely with a transform option. Looking forward to the PR :)

PR #4714, opened 2016-11-15
Document transform options for QueryCursor

cblanc:
This PR adds 2 (related) features discussed in #4705

- `.cursor(opts)` accepts a `transform` function which is applied when the cursor returns a document

- `QueryCursor#map` which allows additional transform functions to be applied to documents pulled with the cursor (whether by the streams interface or `next()`)

The `transform` option and `.map()` can be combined and happens to be chainable e.g. `.cursor({ transform: function () {} }).map(...).map(...).on('data')`


vkarpov15:
You sir are a champion among champions. Great work, thanks :beers:

Issue #4682, opened 2016-11-04
upgrade 3.8.25 to 4.6.6: big performance drop


analyseDev:
We have a production application with mongoose 3.8.25. We have updated to version 4.6.6 and have noted a huge performance drop. Requests become 3 times slower than with version 3.8.5. We have tested several versions 4.x and is exactly the same. If we analyze the requests we see that all requests (find, insert, count) are increased their response time. Is there any reason why this happens?

Thanks

loris:
Probably a duplicate of https://github.com/Automattic/mongoose/issues/4690

vkarpov15:
Hard to determine without examples, can you provide examples of some operations that are slower?

Sebmaster:
We're currently debugging a similiar issue. From diffing a few CPU profiles between 3.8 and 4.6 it seems like [`wrappedPointCut`](https://github.com/Automattic/mongoose/blob/3e8b54b9e90bd889cc19d3275e328f35cda293ac/lib/document.js#L1956) is new. In one of our routes it takes like 2s (total, not self time) of 10s program time, while in other routes it's not there at all (which seems to be because hooks aren't used in that model?).

I'm trying to dig more into this and see if I can reduce that to a testcase somehow.

Sebmaster:
Found out the `wrappedPointCut` path was actually misleading. It's slowness was caused by a slower hook we have. However releasing a build with the new mongoose increases time spent in mongo queries by about 20-30ms per request and it seems much more spikey (although that might just be expensive requests).

vkarpov15:
Yeah that function is used for hooks+promises. How are you measuring "time spent in mongo queries"?

Sebmaster:
We have a class we dispatch most of our queries through, which basically wraps the actual querying and saves the timestamps before + in the callback and then sums them up for logging.

vkarpov15:
That's smart. Couple of suggestions:

1) Use `.lean()` in your queries where possible. If you don't need the extra mongoose functionality and are just executing a query to send the data out of your API, it will make your queries much faster.
2) I'd recommend minimizing your usage of `populate()` by embedding data into necessary docs. The biggest performance wins often come from simply executing fewer queries.

PR #4726, opened 2016-11-17
Delay stack generation until it's required

Sebmaster:
(continued from #4682)

If some kind of stack beautification library (like with coffeescript) is used, generating the stack for every query is really expensive (in the order of 1,5s per 1000 queries).

This should still give the same stacktrace, but only generate the beautifyed stack if it's actually used, increasing performance for the common non-error case.

Even without stack beautification, accessing Error().stack is [10x slower](https://jsperf.com/error-vs-error-stack) than just storing Error(), making this a generally applicable optimization.


vkarpov15:
Good catch, thanks :beers:

PR #4753, opened 2016-11-29
fix: upgrade mongodb -> 2.2.12

mdlavin:
Mostly excited about a memory leak that was fixed, but there are other fixes in there too.


vkarpov15:
Thanks :beers:

PR #4756, opened 2016-11-30
fixed two tests

japod:
test/types.array.test.js
  comparator should now work with Timsort algorithm

test/document.modified.test.js:
   made sure a different date is used for update


vkarpov15:
Thanks :beers:

PR #4763, opened 2016-12-02
Adds additional pluralization irregularities.

huckbee:


vkarpov15:
This is a very dangerous change that I really don't want to risk. Very much backwards breaking. I appreciate the PR and apologize for rejecting it, but the plan long term is to make pluralization more customizable, these hard-coded rules have proven to be a very bad idea.

PR #4769, opened 2016-12-03
Fix: Read benchmark dep/id/output

billouboq:
This PR fix 3 issues with read (benchmark/bench/read.js) benchmark : 

  - missing benchmark dev dependency
  - error in insert function (res[0] is undefined so res[0]._id is throwing an error)
  - the output of this benchmark wasn't pretty and so hard to read


vkarpov15:
Thanks for your work! Haven't used benchmarks for a while but your help is much appreciated :+1:

PR #4772, opened 2016-12-05
Fix justOne test

cblanc:
`npm test` will occasionally fail because the justOne test doesn't cover the case where the alternate populated document is returned

Also added in an assertion that `justOne` virtuals should not be returned as an array for good measure


vkarpov15:
Thanks :beers:

Issue #4656, opened 2016-10-25
4.6.4 regression - MongooseArray now simply Array


mathieumg:
This change https://github.com/Automattic/mongoose/commit/687eecde0c1564b873b8e3c0651f1066c47c3379 now always returns instances of `Array` whereas before it could sometimes return instances of `MongooseArray` when populating, depending on the value of `val`. This introduces problems/crashes wherever a `MongooseArray` is expected in a result, such as when doing `.toObject()` on fields of a returned Document.

I don't know what is the clean way of cloning `MongooseArray`s, but if told I wouldn't mind putting together the appropriate PR. Let me know if you need additional information.

Thanks!

vkarpov15:
Whoa that's a bad bug, will fix ASAP

vkarpov15:
Actually @mathieumg can you clarify which situations it would return a MongooseArray? I don't see any instances like that in our test cases...

mathieumg:
```js
const demoSchema = new Schema({
  demo: [{
    type: String,
    ref: "Demo"
  }],
});
```

Then populated this way:

```js
Demo
.findOne({_id: "someid"})
.exec(doc => doc.populate({
  path: "demo",
  select: "somefield",
  model: "Demo",
}));
```

vkarpov15:
Fixed in master, will be released with 4.6.6 :+1:

mathieumg:
So that's how you copy a `MongooseArray`! :) Thanks for the quick fix. :tada:

PR #4795, opened 2016-12-10
Restored ability to override populated array caster from 4.6.2

pdspicer:
This is related to #4656, however the original fix did not address some "edge" cases that were functioning well prior to 4.6.2. Essentially I've found it convenient to extend the casters and defaults for fields that are MongooseArrays, and the simple cloning of a MongooseArray during population does not persist these extensions.

I'm pretty sure the ability to do this was never really documented or specified anywhere, but it's absence in versions > 4.6.2 might constitute a breaking change, and this pull request seems to resolve the issue without breaking any tests (those related to population, as of this PR it seems that there are a couple of unrelated failing tests).

An example of the situation I am referring to is provided in [this gist](https://gist.github.com/pdspicer/82d96e4e96f569f6e8f381a378baafb8). If additional test cases for this PR are requested I can add them.


vkarpov15:
Tests are only failing because of lint issues. LGTM, thanks for catching this :beers:

pdspicer:
@vkarpov15 Happy to help, noticed this a while back but hadn't found the time to put together the requisite materials. Thanks for getting it in.

Issue #4799, opened 2016-12-13
safe to upgrade mongo 3.4?


bonesoul:
title says it, are we safe to upgrade 3.4 and use mongoose?

vkarpov15:
Yep. 4.7.2 should be safe but I'm about to push 4.7.3, which uses mongodb driver 2.2.16 and thus supports 3.4. We're still missing support for some 3.4 features, most notably the decimal type. See #4759 for updates.

PR #4813, opened 2016-12-16
Use MongoDB 3.4 in CI

cblanc:
Use latest MongoDB in CI

Also had to amend a test which uses an invalid parameter to build a collection (I'm not able to find a `safe` option documented for `createCollection` in mongodb 2.6 and above)

Helpful for #4799

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ node_js:
   - "0.10"
   - "iojs"
 before_script:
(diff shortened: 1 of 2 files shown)


vkarpov15:
`safe` is a not very commonly used mechanism for setting write concern in mongoose.

Re: using 3.4, I'd rather not upgrade for now because using 2.6 helps us ensure backwards compat. Thanks for the suggestion though :beers:

Issue #4823, opened 2016-12-20
Subdocuments and strict mode. bug?


dkrosso:
So i'm testing the follow code, to try a way to have a optional subdoc, but if it exists, the fields need to be validated. it works. But, in the 'vendor4', i tried to add a not mapped field with extra information, and set the strict to false, but it is not working. Is it a possible bug? or i'm doing something wrong?

```
let vendorSchema = {
            "name" : {"type" : "String", "required": true},
            "contacts" : {
                "type" : [{
                    "name" : {"type" : "String", required : true},
                    "email" : {"type" : "String", required : true},
                }],
                "required" : false
            }
        };

        let vendor1 = {
            "name" : "v1",
            "contacts" : []
        }; //inserted. 

        let vendor2 = {
            "name" : "v2",
            "contacts" : [{"name" : "v2_c1", "email" : "v2_c1@v2_c1.com"}]
        }; //inserted.


        let vendor3 = {
            "name" : "v3",
            "contacts" : [{"name" : "v3_c1"}]
        }; //missing email.

        let vendor4 = {
            "name" : "v4",
            "contacts" : [{"name" : "v4_c1", "email" : "v4_c1@v4_c1.com", "extra" : "extraInfo"}]
        }; //inserted but without 'extra' field

     let Thing = mongoose.model('vendorssample', new mongoose.Schema(vendorSchema, {strict : false}));
        let thing = new Thing(vendor4);
        thing.save().then((val) => {
            console.log(val);
        }).catch(err => {
            console.log("err",err);
        });

```
mongoose 4.6.8
mongodb 3.0.5
node.js 6.0.0

vkarpov15:
Mongoose implicitly creates a nested schema when you create a document array, you need to make sure to set `strict` on the array schema, not the parent schema.

```javascript
let vendorSchema = new Schema({
            "name" : {"type" : "String", "required": true},
            "contacts" : {
                "type" : [new Schema({
                    "name" : {"type" : "String", required : true},
                    "email" : {"type" : "String", required : true},
                }, { strict: false })],
                "required" : false
            }
        });
```

dkrosso:
thanks, i've been trying the 'implicit' schema constructor but this should work.

PR #4830, opened 2016-12-22
Update schema.js

dkrosso:
propagate the 'strict' option to a child schema.

solves the issue [#4823](https://github.com/Automattic/mongoose/issues/4823)


vkarpov15:
Reasonable suggestion but probably backwards breaking. Mind if I keep this open for a future release?

PR #4831, opened 2016-12-22
propagate strict option to a child schema

dkrosso:
propagate the 'strict' option to a child schema.

solve this issue [#4823](https://github.com/Automattic/mongoose/issues/4823)


vkarpov15:
I'll put this into 4.8, probably too big of a change for a point release but this is a pretty good idea. Just one concern: if somebody specifies a child schema with a different strict option, would this overwrite it?

dkrosso:
this sounds good. thanks,  I Believed the expected behavior should respect the previous opt definition on the child schema (if is that set, of course), maybe a new option 'propagateOpts' to help on this feature.

Issue #4841, opened 2016-12-27
Document.validate is behaving inconsistenly


arciisine:
**Do you want to request a *feature* or report a *bug*?**

**What is the current behavior?**
There is a bug that was introduced in 4.7.5 or earlier (not 100% certain), that when calling `validate` on a document is not returning a promise, and it has in the past.  We are also experiencing a bug in which the validation code is complaining about callback not being a function, when it is optional per the jsdocs.

**If the current behavior is a bug, please provide the steps to reproduce.**

```js
let mongoose = require('mongoose/lib/browser');
mongoose.Document = require('mongoose/lib/browserDocument.js');
let d = new mongoose.Document(
   {name : 5}, 
   mongoose.Schema({ name : { type : String, required : true }}));
let res = d.validate();
console.log(res);
```

**What is the expected behavior?**
Res should be a promise

**Please mention your node.js, mongoose and MongoDB version.**
```
node.js: 7.3
mongoose: 4.7.5
mongodb: 3.2.10
```

vkarpov15:
Thanks for reporting :beers:

PR #4842, opened 2016-12-27
Allowing callback to be optional

arciisine:
This allows callback to be optional as `$__validate` assumes callback is defined and not optional.  

**Summary**
This is a small fix to handle the issue with callback not being defined, breaking the validation process.  This is tied to #4841

**Test plan**


vkarpov15:
Thanks!

Issue #4860, opened 2017-01-03
bugs during update from 4.7.2 to 4.7.3


worotyns:
Hello,

Yesterday after update the mongoose to 4.7.3 i hava some trouble with "all" tests in my application
All tests returns me: Validation Errors, and Cast errors, TypeError: Cannot read property 'remove' of undefined;

Sample error:
```
 CastError: Cast to ObjectId failed for value "586b672be11a4670c4eb3f66" at path "_id" for model "User"
      at MongooseError.CastError (/common/node_modules/mongoose/lib/error/cast.js:26:11)
      at ObjectId.cast (/common/node_modules/mongoose/lib/schema/objectid.js:147:13)
      at ObjectId.castForQuery (/common/node_modules/mongoose/lib/schema/objectid.js:187:15)
      at cast (/common/node_modules/mongoose/lib/cast.js:229:32)
      at Query.cast (/common/node_modules/mongoose/lib/query.js:2752:12)
      at Query.findOne (/common/node_modules/mongoose/lib/query.js:1353:10)
      at /common/node_modules/mongoose/lib/query.js:2310:21
      at Query.exec (/common/node_modules/mongoose/lib/query.js:2304:17)
      at Query.then (/common/node_modules/mongoose/lib/query.js:2351:15)
      at next (/common/node_modules/co/index.js:100:51)
      at onFulfilled (/common/node_modules/co/index.js:69:7)
      at /common/node_modules/co/index.js:54:5
      at co (/common/node_modules/co/index.js:50:10)
      at Test.<anonymous> (src/routes/me.spec.js:169:11)
      at Test.assert (/common/node_modules/supertest/lib/test.js:156:6)
      at Server.assert (/common/node_modules/supertest/lib/test.js:127:12)
      at emitCloseNT (net.js:1537:8)

```

Let me know what more informations you know.

karliky:
We are having the same problem.

stackcats:
We are having the same problem.

wanghui2016:
We are having the same problem.

chinaguo:
We are having the same problem.

wangpeiyong:
We are having the same problem.

karliky:
We also have this problem:

```
"TypeError: Unknown encoding: 2",
        " at Buffer.slowToString (buffer.js:363:17)",
        " at Buffer.toString (buffer.js:376:27)",
[...]

PR #4862, opened 2017-01-03
Fixes the bson dependency version

IvanGuardado:
**Summary**
Recently we found an issue related with mongoose in our production application. After some investigation that @jvrbaena summarized [here](https://github.com/Automattic/mongoose/issues/4860#issuecomment-270089296) we saw that the problem is an update to the bson dependency. With this change we fix the dependency version to avoid this kind of unforeseen problems.

**Test plan**
I tried to run the tests in local and I got 123 errors with the current version.
With this change I get 26 errors but they all are related with my local enviroment (sharding, hostnames, etc...). I'm pretty sure they'll pass in the CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "license": "MIT",
   "dependencies": {
     "async": "2.1.4",
-    "bson": "~1.0.1",
+    "bson": "1.0.1",
     "hooks-fixed": "1.2.0",
     "kareem": "1.2.0",
     "mongodb": "2.2.16",


kyrylkov:
@IvanGuardado mongoose, depends on mongodb, which in turn depends on mongodb-core, which in turn has tilde dependency on bson, so AFAIK your PR won't solve the problem you're trying to solve.  

In fact bson bug caused mongodb-core bugs.

vkarpov15:
Yeah @kyrylkov is right, we'll just use the same bson version that mongodb-core uses

Issue #4446, opened 2016-08-22
populate array+single


TrejGun:
new bug in populate function :(
```js
var q = require("q");
var mongoose = require("mongoose");
mongoose.set("debug", true);
mongoose.Promise = q.Promise;

var db = mongoose.createConnection();
db.open("mongodb://localhost/issue-123");

var UserSchema = new mongoose.Schema({
	name: {
		type: String
	}
}, {
	timestamps: true,
	versionKey: false
});

var GroupSchema = new mongoose.Schema({
	name: {
		type: String
	},
	students: [{
		type: mongoose.Schema.Types.ObjectId,
		ref: "User"
	}],
	teacher: {
		type: mongoose.Schema.Types.ObjectId,
		ref: "User"
	}
}, {
	timestamps: true,
	versionKey: false
});


var Group = db.model("Group", GroupSchema);
var User = db.model("User", UserSchema);


q.all([
	User.create({name: "Student Name 1"}),
	User.create({name: "Student Name 2"}),
	User.create({name: "Teacher Name 1"})
])
	.spread(function (student1, student2, teacher1) {
		return Group.create({name: "Group Name", students: [student1._id, student2._id], teacher: teacher1._id})
			.then(function () {
				return Group.findOneAndUpdate({
					teacher: teacher1._id
				}, {
					name: "New Group Name"
				}, {
					new: true,
					populate: [{path: "students"}, {path: "teacher"}]
				})
					.then(function (result) {
						console.log("result", result);
					});
			});
	});
```
throws
```
TypeError: utils.populate: invalid path. Expected string. Got typeof `object`
```

doesn't throw if `populate: [{path: "students"}]` or `populate: [{path: "teacher"}]`

TrejGun:
thanks

PR #4893, opened 2017-01-11
fixed populate can not use more model

PromeYang:
as before use this code:

populate: {
path: 'data.user data.group',
model: 'User Group',
select: '_id'
}

you will get an err like this:
MissingSchemaError: Schema hasn't been registered for model "User Group".

after fixed this code, all thing is going well, it can use more model at same time

**Summary**

**Test plan**

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -624,9 +624,15 @@ exports.populate = function populate(path, select, model, match, options, subPop
 
   var ret = [];
   var paths = path.split(' ');
+  var models = undefined;
+  if (typeof model === 'string') {
+    models = model.split(' ');
+  }
   options = exports.clone(options, { retainKeyOrder: true });
   for (var i = 0; i < paths.length; ++i) {
-    ret.push(new PopulateOptions(paths[i], select, match, options, model, subPopulate));
+    var _model = model;
+    if (models && i < models.length) _model = models[i];
+    ret.push(new PopulateOptions(paths[i], select, match, options, _model, subPopulate));
   }
 
   return ret;
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -632,6 +632,7 @@ exports.populate = function populate(path, select, model, match, options, subPop
   for (var i = 0; i < paths.length; ++i) {
     var _model = model;
     if (models && i < models.length) _model = models[i];
+    
     ret.push(new PopulateOptions(paths[i], select, match, options, _model, subPopulate));
   }
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -632,7 +632,6 @@ exports.populate = function populate(path, select, model, match, options, subPop
   for (var i = 0; i < paths.length; ++i) {
     var _model = model;
     if (models && i < models.length) _model = models[i];
-    
     ret.push(new PopulateOptions(paths[i], select, match, options, _model, subPopulate));
   }
 


PromeYang:
the check system is unuseful ,  the same code can be different check result

lib/utils.js
```diff
@@ -624,9 +624,15 @@ exports.populate = function populate(path, select, model, match, options, subPop
...
   var paths = path.split(' ');
+  var models = undefined;
+  if (typeof model === 'string') {
+    models = model.split(' ');
+  }
   options = exports.clone(options, { retainKeyOrder: true });
   for (var i = 0; i < paths.length; ++i) {
-    ret.push(new PopulateOptions(paths[i], select, match, options, model, subPopulate));
+    var _model = model;
+    if (models && i < models.length) _model = models[i];
```
vkarpov15: Seems kinda dangerous - what if models length is different from paths length?

TrejGun:
what if ?
```
populate: {
path: 'data.user data.group',
model: 'Group User',
select: '_id'
}
```

lib/utils.js
PromeYang: if the paths length > models length, only models exist and the index in the paths repeat is little than models length , will do "_model = models[i]" , if not , the "_model" as default "model", there is nothing change as before!

if the paths length < models length, only repeat paths length times to do "_model = models[i]", other models is ignore!

I only change the models rule , the other code is as before to do ,
if someone input the illegal models, will get the error as before.

PromeYang:
@TrejGun 
you can see some code like this,
```
var matchOrder = {
    addUser: user._id,
    type: 1,
    delFlag: 2
};
var optionOrder = {
    select: '',
    sort: '-addDate',
    limit: pageSize,
    skip: pageSize * (searchPage - 1),
    populate: {
        path: 'storehouse package country transport',
        select: 'name',
        model: 'Storehouse Package Country Transport'
    }
};
Order.find(matchOrder, '', optionOrder);
```
and you can get the result like this,
```
[ { _id: 'R010001485918471',
    storehouse: { _id: 1001, name: '仓储服务22' },
    package: { _id: 1000, name: '包裹11' },
    country: { _id: 1001, name: '国家2' },
    transport: { _id: 1000, name: '运输方式11' },
    ...
    ...
} ]
```
as you see , I have 4 model "Storehouse Package Country Transport", I can find the result one time!

vkarpov15:
I'm afraid this API's a little too implicit for my tastes - having to put both path and model and have them line up exactly seems like a not very clear API. You can do something like this to get a similar result without this PR

```javascript
var models = ['Storehouse', 'Package', 'Country', 'Transport'];
var optionOrder = {
    select: '',
    sort: '-addDate',
    limit: pageSize,
    skip: pageSize * (searchPage - 1),
    populate: models.map(model => ({
        path: model.toLowerCase(),
        select: 'name',
        model: model
    }))
};
```

PromeYang:
@vkarpov15 
I want to ask you , which version mongo and mongoose as you use?
I use mongo for mongodb-linux-x86_64-3.2.0, mongoose for "mongoose": "^4.1.12",
If I set the populate as a array, I will get an error like this:

```
[TypeError: utils.populate: invalid path. Expected string. Got typeof `object`]
```

Please help and give me some tips , why I got this error? 
thanks a lot!

vkarpov15:
You're correct that fails on mongoose 4.5.4, so does the below script:

```javascript
'use strict';

var assert = require('assert');
var Schema = mongoose.Schema;

mongoose.connect('mongodb://localhost/gh4893');
mongoose.set('debug', true);

var schema = new Schema({
  a: { type: mongoose.Schema.Types.ObjectId, ref: 'A' },
  b: { type: mongoose.Schema.Types.ObjectId, ref: 'B' }
});

var childSchema = new Schema({
  name: String
});

var A = mongoose.model('A', childSchema);
var B = mongoose.model('B', childSchema);
var M = mongoose.model('M', schema);

A.create({ name: 'a' }).
  then(a => B.create({ name: 'b' }).then(b => [a, b])).
  then(([a, b]) => M.create({ a, b })).
  then(m => M.findById(m, null, { populate: [{ path: 'a', model: A }, { path: 'b', model: B }] })).
  then(doc => console.log(doc)).
  catch(error => console.error(error));
```

However, works just fine in >= 4.6.0 because of #4446. Try upgrading to 4.6.0, should work.

PromeYang:
@vkarpov15 
ok, thanks a lot

Issue #4873, opened 2017-01-04
Feature Request: Smarter queries


QuotableWater7:
I would like to chat about possibly removing some boilerplate around queries, by accepting either an array of values or a single value for a given query key.

**What is the current behavior?**
```js
someModel.find({ someKey: oneValue })

// OR

someModel.find({
  someKey: { $in: arrayOfValues },
})
```

**What is the desired behavior?**
```js
someModel.find({  someKey: oneValue })

// OR

someModel.find({ someKey: arrayOfValues })
```

**Why?**
I worked previously with ActiveRecord, and they had abstracted the difference between single value and an array of values into the ORM.  It makes for _slightly_ cleaner code, as the nested braces can be difficult to read on the same line.  We often take a simple query and spread it across 3 lines just to avoid that.

I wanted to see if the `mongoose` maintainers would be open to this sort of feature -- if so, I am more than happy to write the code that can make this happen.  I might have a question or two about the likely spots I'd need to hit, though.

Also, to clarify, this would be 100% backwards compatible.

Thanks for your time!

TrejGun:
good point for 5.0

QuotableWater7:
@TrejGun Just wanted to check in, are there any next-steps from my side (getting more decision makers to take a look, submitting a PR, etc)?  Thanks!

TrejGun:
actually, looking at the code more close it seems you can use array operations on single element
https://github.com/Automattic/mongoose/blob/master/lib/schematype.js#L881

howeven it will create a query like
```
{ _id: { '$in': [ ObjectId("5875a365589f0c2d29f0c2e6") ] } }
```
but not
```
{ _id: { '$eq': [ ObjectId("5875a365589f0c2d29f0c2e6") ] } }
```

I'm pretty sure @vkarpov15 won't merge this in mongoose 4, so if you want to switch operator at runtime you probably should try to write a plugin

QuotableWater7:
@TrejGun Correct me if I'm wrong, but I think your example is slightly different than I'm going for.  I'm basically wanting to do:
```js
{ _id: someVariableContainingArrayOfIds }
```

Whereas right now, I'm forced to expand out some boilerplate:
```js
{ _id: { $in: someVariableContainingArrayOfIds } } 
```

Does that make sense?

vkarpov15:
[...]

PR #4912, opened 2017-01-19
Abstract away $in boilerplate when querying values

QuotableWater7:
**Summary**
Implementation of feature described in https://github.com/Automattic/mongoose/issues/4873

Normally, a query to find a key that matches multiple values would look like:
```js
SomeModel.find({ 
  _id: { 
    $in: [id1, id2, id3],
  },
})
```

With this change, the nested `$in` part would no longer be necessary.

```js
SomeModel.find({ _id: [id1, id2, id3] })
```

Motivation:  Cleaner code when writing queries.

**Test plan**
See the added tests in `cast.test.js`


lib/cast.js
```diff
@@ -225,6 +225,13 @@ module.exports = function cast(schema, obj, options) {
             }
           }
         }
+      } else if (val.constructor.name === 'Array' && ['Buffer', 'Array'].indexOf(schematype.instance) === -1) {
+        var casted = [];
```
QuotableWater7: My preference would have been to put this logic into the individual `cast` methods for each type, but the problem is that in those `cast` methods, I have no idea if the value is actually supposed to be an array of values.  e.g. I could have a schema with:
```js
{
  someKey: { type: [Number] }
}
```
OR, I could have a schema with:
```js
{
  someKey: { type: Number } 
}
```

However, in `lib/schema/number.js`'s `cast` method, I don't know if the value is _supposed to be an array_, or if the value can simply be one of any number of values and thus can be converted to `$in` query.  

I would have passed another variable into the cast method at line `236` to contain the `schematype`, but the problem is having more than 1 argument will change how the method behaves entirely.

Long story short, I'm happy to hear possible improvements or a better location for this logic that can still pass the tests.

vkarpov15:
This looks great :+1: I'll merge it when we release 4.8, which should be next week

lib/cast.js
```diff
@@ -226,12 +226,9 @@ module.exports = function cast(schema, obj, options) {
           }
         }
       } else if (val.constructor.name === 'Array' && ['Buffer', 'Array'].indexOf(schematype.instance) === -1) {
-        var casted = [];
-        for (var valIndex = 0; valIndex < val.length; valIndex++) {
-          casted.push(schematype.castForQuery(val[valIndex]));
-        }
+        var castedValues = val.map(schematype.castForQuery.bind(schematype));
```
vkarpov15: Can you revert this change? `map()` is very neat and clean but much slower than for loop, and `.bind()` is extra slow.

QuotableWater7: @vkarpov15 I just reverted 👍

PR #4919, opened 2017-01-20
mongodb as a peerDep

amongouser:
**Summary**
This allows users of mongoose to ensure they only have one version of the mongodb library installed.

**Test plan**


vkarpov15:
Worth considering peer dep for a future release but unlikely since we always pin exact versions anyway. Mongoose will never use up-carat, `^` is a gross anti-pattern that needs to die.

Issue #4960, opened 2017-02-07
Array of object with nested schema doesn't work with runValidators


valterkraemer:
```
var mongoose = require('mongoose');
mongoose.Promise = require('bluebird');
var Schema = mongoose.Schema;

mongoose.connect('mongodb://localhost/test');

var ProductSchema = new Schema({
  name: String
});

var UserSchema = new Schema({
  sell: [{
    product: ProductSchema
  }]
});

var User = mongoose.model('User', UserSchema);

return User.create({})
.then(function(user) {
  return User.update({
    _id: user._id
  }, {
    sell: [{
      product: {
        name: 'Product 1'
      }
    }]
  }, {
    runValidators: true
  });
});
```

Gives the error `TypeError: value.validate is not a function`. Update validation doesn't seem to like an array of objects with nested schema.

mongoose version: 4.8.1
mongodb version: 3.2.6

dkrosso:
is it a good solution for your problem?

`var ProductSchema = new Schema({
    product : {
        name: String
    }
});

var UserSchema = new Schema({
    sell: [ProductSchema]
});`

valterkraemer:
No, my real use case is much more complex. This is a slimmed down version

dkrosso:
```
var UserSchema = new Schema({
    sell: { type : [ProductSchema]}
});
```

valterkraemer:
Something like this is the actual implementation

```
var ProductSchema = new Schema({
  _id: false,
  a: String,
  b: String,
  c: String
});

var AreaSchema = new Schema({
  _id: false,
  a: String,
  b: String
});

var UserSchema = new Schema({
  sell: {
    type: [{
      product: {
        type: ProductSchema,
        required: true,
        validate: function...
      },
      areas: {
        type: [AreaSchema],
        validate: function...
      }
    }],
    select: false
  }
})
```

PR #4966, opened 2017-02-08
verifying if the value.validate() exists as a function

dkrosso:
fix the issue [#4960](https://github.com/Automattic/mongoose/issues/4960)
avoiding the Exception `TypeError: value.validate is not a function` by checking if the typeof value.validate is a function in a embed subdoc.


vkarpov15:
Thanks for the PR but the underlying issue is more complex. See #4960

PR #4977, opened 2017-02-12
rawResult option

### LINKED ISSUE #4925: post-findOneAndUpdate middleware hangs with passRawResult = true
bug

Post-findOneAndUpdate middleware function defined with two arguments hangs when passRawResult=true. Apparently, it is passing a 3rd `next()` argument that it is expecting to be called. However, when I define the middleware function with 3 arguments, the middleware is skipped altogether (I'm guessing because it assumes this is 'error middleware'). I'm very confused about what is supposed to happen here, or how such middleware should be written for a plugin that is capable of handling both passRawResult=true and passRawResult=false scenarios...

Running 
`Model.findOneAndUpdate({ prop: 'xyz' }, { prop: 'abc' }, { upsert: true, passRawResult: true })`
produces the following behavior for each of the different middleware setups:

```
  // This is called, and is passed 2 arguments (the doc, and the raw result)
  schema.post('findOneAndUpdate', function (doc) {
    console.log(arguments)
  })

  // output:
  //
  // { '0': null,
  //  '1': 
  //   { lastErrorObject: 
  //      { updatedExisting: false,
  //        n: 1,
  //     upserted: 5883edc039241549e8322aa0 },
  //     value: null,
  //     ok: 1,
  //     _kareemIgnore: true } }
```

```
  // This hangs, logging arguments shows that a 3rd hidden parameter (function) is passed
  schema.post('findOneAndUpdate', function (doc, result) {
    console.log(arguments)
  })

  // output:
  //
  // { '0': null,
  //  '1': 
  //   { lastErrorObject: 
  //      { updatedExisting: false,
  //        n: 1,
  //     upserted: 5883edc039241549e8322aa0 },
  //     value: null,
  //     ok: 1,
  //     _kareemIgnore: true },
  //  '2': [Function]  }
```

```
  // This is never called, presumably because it's assumed to be an error handler?
  schema.post('findOneAndUpdate', function (doc, result, next) {
    console.log(arguments)
  })
```

```
  // This is actually called, but next and _ are undefined
  schema.post('findOneAndUpdate', function (doc, result, next, _) {
    console.log(arguments)
  })

  // output:
  //
  // { '0': null,
  //  '1': 
  //   { lastErrorObject: 
  //      { updatedExisting: false,
  //        n: 1,
  //     upserted: 5883edc039241549e8322aa0 },
  //     value: null,
  //     ok: 1,
  //     _kareemIgnore: true } }
```

I'd like to be able to access the raw result in my post middleware without causing it to hang. I'd also like to know what the appropriate signature is for a post middleware that might not know if the update operation set passRawResult to true or false

Node 7.2.1, mongoose 4.7.7, mongodb 3.2

vkarpov15:
Fix #4925

The `passRawResult` option has proven to be exceptionally cumbersome for 2 reasons:

1. We had to hack around it to make it not break error handler middleware. Because an error handler middleware is defined by the number of arguments it takes, conditionally adding a third param makes it impossible to write error handler middleware that can handle both `passRawResult: true` and `passRawResult: false`. The current hack makes it so that raw result is _never_ passed to error handler middleware.
2. You need to use the callback API or mpromise (deprecated). With ES6 promises, you can never get the raw result.

This PR introduces the `rawResult` option, meant as a replacement for `passRawResult`, which makes the 2nd param to the callback the raw result (as opposed to the 3rd). This works better with promises and error handler middleware. @bdentino @evilive3000 @varunjayaraman y'all have reported related issues, do you have any thoughts on this approach?

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -1307,7 +1307,7 @@ Query.prototype._findOne = function(callback) {
     if (!options.populate) {
       return options.lean === true
           ? callback(null, doc)
-          : completeOne(_this.model, doc, null, projection, null, callback);
+          : completeOne(_this.model, doc, null, {}, projection, null, callback);
     }
 
     var pop = helpers.preparePopulationOptionsMQ(_this, options);
@@ -1318,7 +1318,7 @@ Query.prototype._findOne = function(callback) {
       }
       return options.lean === true
           ? callback(null, doc)
-          : completeOne(_this.model, doc, null, projection, pop, callback);
+          : completeOne(_this.model, doc, null, {}, projection, pop, callback);
     });
   });
 };
@@ -1628,7 +1628,7 @@ Query.prototype.remove = function(cond, callback) {
  * @param {Function} callback
  */
 
-function completeOne(model, doc, res, fields, pop, callback) {
+function completeOne(model, doc, res, options, fields, pop, callback) {
   var opts = pop ?
   {populated: pop}
       : undefined;
@@ -1638,7 +1638,12 @@ function completeOne(model, doc, res, fields, pop, callback) {
     if (err) {
       return callback(err);
     }
-    if (res) {
+
+    if (options.rawResult) {
+      res.value = casted;
+      return callback(null, res);
+    }
+    if (options.passRawResult) {
       return callback(null, casted, decorateResult(res));
     }
     callback(null, casted);
@@ -1963,20 +1968,21 @@ Query.prototype._findAndModify = function(type, callback) {
     }
 
     if (!doc || (utils.isObject(doc) && Object.keys(doc).length === 0)) {
+      if (opts.rawResult) {
+        return callback(null, res);
+      }
+      // opts.passRawResult will be deprecated soon
       if (opts.passRawResult) {
         return callback(null, null, decorateResult(res));
       }
       return callback(null, null);
     }
 
-    if (!opts.passRawResult) {
-      res = null;
-    }
-
     if (!options.populate) {
-      return options.lean === true
-          ? (opts.passRawResult ? callback(null, doc, decorateResult(res)) : callback(null, doc))
-          : completeOne(_this.model, doc, res, fields, null, callback);
+      if (options.lean === true) {
+        return _completeOneLean(doc, res, opts, callback);
+      }
+      return completeOne(_this.model, doc, res, opts, fields, null, callback);
     }
 
     var pop = helpers.preparePopulationOptionsMQ(_this, options);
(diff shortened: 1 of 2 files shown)


bdentino:
How feasible would it be to support a middleware signature that gets called no matter what and accepts all possible arguments? For example, something like:

```
schema.post('findOneAndUpdate', function (err, doc, raw, next) {
  if (err) {
    return handleErr(err, next)
  }
  if (raw) {
    doSomethingWithRawResponse(raw)
  }
  next()
})
```

The use case that I'm currently stuck on is emitting events when a new object is created via an upsert with the new object's ID. To do this, I have a plugin that force-sets the `passRawResult` option to true in pre-middleware if `upsert = true` so that I can access the ID of the upserted document in the post-middleware and emit an event with that information. Correct me if I'm wrong, but it seems that doing so would potentially break other plugins that use post middleware because that option changes the semantics of how arguments are passed to those functions and/or whether that function is considered an error-handler or not. Either way, I personally find that the whole dynamic arguments concept makes it a lot more difficult for me to reason about the behavior of the schemas/plugins I write. I would gladly do the work of splitting the error-handling case and raw-result case in my own middleware functions if it meant I didn't have to think about whether or not my plugins will all work together depending on what options were passed or what errors may or may not have been thrown.

varunjayaraman:
I'm in support of this, makes getting the raw result promise-friendly. I'm not sure what other solution would exist that would be reasonably backwards compatible, but also work with promises.

@bdentino are you suggesting that the raw result always gets passed to a middleware with an arity of 4?

bdentino:
Yeah, it would be nice to be able to count on it being there regardless of options, although even just passing it in the cases where the query options include `rawResult=true` (and passing null otherwise) would be an improvement IMO. Otherwise it would seem that this option fundamentally undermines many useful applications of post-findOneAndUpdate middleware. Unless you can guarantee that every such query will use the same value for `rawResult`, you have no idea whether your function will have access to a mongoose document or will be passed a raw mongo response. 

BTW, I'm also suggesting that such middleware get called in both the error case and in the success case (albeit with `null` in the err param).

One thing I'm not clear on though: will setting this option to `true` result in the final resolved value being the mongo raw response? Or will you still get a mongoose document back? i.e. 
```
Model.findOneAndUpdate({}, {update}, {rawResult:true}).then((/* raw result or mongoose doc? */) => {})
```

vkarpov15:
I like the idea of `function (err, doc, raw, next)` but the problem is that would be a backwards breaking change. In fact, any resolution I can think of to this issue is backwards breaking except encouraging people to switch to the `rawResult` flag.

Your middleware should be able to access the `rawResult` option and switch on it if necessary, `this` is the query in query middleware. It's up to the middleware to account for that flag, but accounting for it should be pretty simple: `if (this.options.rawResult) {} else {}`

Re: promises, setting this option to `true` will make the promise resolve to the raw result, with one caveat: `result.value` will be the full-fledged mongoose doc, so `Model.findOneAndUpdate({}, update, { rawResult: true }).then(res => res.value)` resolves to the same thing as `Model.findOneAndUpdate({}, update, { rawResult: false })`

bdentino:
Awesome, I did not know about the `result.value` caveat. I think I can cover every use case I need with that. Thanks @vkarpov15!

artofspeed:
@vkarpov15 with `rawResult: true`, how to get `model.findOneAndUpdate` actually return the document? Otherwise it'll be the same as `model.update()` and loses its purpose....

PR #5018, opened 2017-02-24
Add a clone() function for schemas

### LINKED ISSUE #4983: Bug: Discriminator models mutate the underlying schema
Bug

Discriminator models mutate the underlying schema which is passed by reference, rather than by copy. This makes discriminator models significantly harder to test because they retain state from schema singletons across test blocks (and might also cause nasty bugs in applications that statefully use the schemas underlying any given discriminator model). 

If you try to create the same discriminator model twice within a test file, it'll complain: `Error: Discriminator "{ModelName}" cannot have field with name "{DiscriminatorKey}"`

1) `yarn / npm install mongoose co assert mocha`
2) `$(npm bin)/mocha {filename}`

```javascript
const co = require('co');

// Globals: place schemas here as a global to mimick the singleton pattern
const globals = {
  baseSchema: new mongoose.Schema({
    profile: {
      name: String,
      email: String
    }
  }, { timestamps: true }),

  discriminatorSchema: new mongoose.Schema({
    profile: {
      address: { city: String, line1: String }
    }
  })
};

// Constants
const GITHUB_ISSUE = `overwrite-discriminators-bug`;

describe('before each bug with discriminators', function() {
  let Base;
  let Discriminator;
  let db;

  function init() {
    return co(function* () {
      db = yield connectToDb();
      const models = createModels(db);
      Base = models.Base;
      Discriminator = models.Discriminator;
    });
  }

  describe('Block 1', function() {
    before(function(done) {
      co(function* () {
        yield init();
        done();
      }).catch(done);
    });

    after(function(done) {
      db.close();
      done();
    });

    it('test 1', function(done) {
      assert.ok(true);
      done();
    });

    it('test 2', function(done) {
      assert.ok(true);
      done();
    });
  });

  describe('Block 2', function() {
    beforeEach(function(done) {
      co(function* () {
        yield init();
        done();
      }).catch(done);
    });

    after(function(done) {
      db.close();
      done();
    });

    it('test 3', function(done) {
      assert.ok(true);
      done();
    });

    it('test 4', function(done) {
      assert.ok(true);
      done();
    });
  });
});

function createModels(db) {

  const Base = db.model('Base', globals.baseSchema);
  const Discriminator = Base.discriminator('Discriminator', globals.discriminatorSchema);
  return { Base, Discriminator };
}

function connectToDb() {
  return co(function* () {
    return mongoose.createConnection(`mongodb://localhost:27017/${GITHUB_ISSUE}`);
  });
}


```

Mongoose should create a copy of the schema so that if you try to recreate the discriminator model, it does not already have the discriminator key

Node v6.9.2, mongodb v3.4.1, mongoose v4.8.2

vkarpov15:
Fix #4983.

@varunjayaraman can't change the behavior described in #4983 in 4.9 because it's documented and tested behavior. However, this commit adds a nice `clone()` function to schemas that you can use to clone a schema before passing it to `discriminator()`


varunjayaraman:
Looks good, this should hopefully eliminate the state in tests that reinstantiate models in a `before` hook

PR #5024, opened 2017-02-27
Support isAsync option for validators

### LINKED ISSUE #4290: Add explicit `validateAsync` function for schemas, deprecate implicit async validators
Re: #4084, #4064

vkarpov15:
Fix #4290

### DIFF
--- a/lib/schematype.js
+++ b/lib/schematype.js
@@ -727,19 +727,11 @@ SchemaType.prototype.doValidate = function(value, fn, scope) {
         validate(true, validatorProperties);
         return;
       }
-      if (validator.length === 2) {
-        var returnVal = validator.call(scope, value, function(ok, customMsg) {
-          if (typeof returnVal === 'boolean') {
-            return;
-          }
-          if (customMsg) {
-            validatorProperties.message = customMsg;
-          }
-          validate(ok, validatorProperties);
-        });
-        if (typeof returnVal === 'boolean') {
-          validate(returnVal, validatorProperties);
-        }
+      if (validatorProperties.isAsync) {
+        asyncValidate(validator, scope, value, validatorProperties, validate);
+      } else if (validator.length === 2 && !('isAsync' in validatorProperties)) {
+        legacyAsyncValidate(validator, scope, value, validatorProperties,
+          validate);
       } else {
         try {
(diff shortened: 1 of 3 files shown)


varunjayaraman:
Looks good to me. Are there tests for this bad boy yet?

vkarpov15:
Nope other than the docs test. Will add a couple more before I merge this one :+1:

crossman:
@vkarpov15 It looks like this option isn't fully accessible when adding a validator from a plugin.

http://mongoosejs.com/docs/api.html#schematype_SchemaType-validate

The documentation for that still says that synchronicity is assumed based on the function signature. And after looking at the [code](https://github.com/Automattic/mongoose/blob/91866a8cff78b92d1f60e973e99910c6e8fdb612/lib/schematype.js#L452) it seems like `isAsync` can be set by passing in an object for the `errorMsg` param, but only if the `type` param is omitted.

I'm just wondering if `SchemaType#validate` needs to be updated to make `isAsync` easier to set or if it is intended that you should only be able to set it when passing an object to the `errorMsg` param without setting the `type`.

vkarpov15:
Thanks for the suggestion @crossman , will fix asap.

crossman:
Thank! I greatly appreciate your work.

vkarpov15:
@crossman actually it's pretty easy, just do this:

```javascript
schema.path('name').validate({
  isAsync: true,
  validator: function(val, cb) { /* ... */ },
  message: 'blah' // custom message is optional
});
```

IMO much cleaner than relying on positional args anyway

crossman:
I actually ran into this because a plugin I was using triggered the deprecation message and didn't expose enough of its usage of `validate` for me to do something like what you suggested. Which led me into a deep dive of figuring out it if it was possible for me to solve the deprecation without waiting for the plugin to release an update. The plugin has since been updated anyway, so I guess everything is solved. Just lead me to some questions on intended use.

vkarpov15:
Glad you figured it out and sorry about the trouble. There have just been an awful lot of hard-to-debug issues with people plugging in outside functions into `validator` and not checking the function arity carefully enough, so we want to discourage this behavior going forward. Let me know if you run into any other plugins that need to upgrade their usage :+1:

PR #5072, opened 2017-03-14
InsertMany behaves as expected for validation errors while ordered op…

sjorssnoeren:
### Summary
The native mongodb driver documents functionality for an option called: 'ordered'. It's documentation is as follows: If true, when an insert fails, don't execute the remaining writes. If false, continue with remaining inserts when one fails. I expected this to work with mongoose as well, as it is documented. However this is only true for failing unique constraint errors. I've made little modifications to make it work for all validation errors.

### Test plan
Added an extra unit test for making sure the behavior of validation errors is identical to the one of constraint errors

### Notice
This is one of my very first pull-requests, if I went too far without discussion or if there are other concerns that need taken care of please let me know.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1976,6 +1976,12 @@ Model.insertMany = function(arr, options, callback) {
       doc = new _this(doc);
       doc.validate({ __noPromise: true }, function(error) {
         if (error) {
+          // Option `ordered` signals that insert should be continued after reaching
+          // a failing insert. Therefore we delegate "null", meaning the validation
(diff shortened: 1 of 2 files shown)


sjorssnoeren:
One error occurred at CI for a function I haven't touched, or could have modified. It also only rejects for one environment. Is it possible to run that CI container again?

vkarpov15:
This is great work, thank you very much for your contribution, will be released with 4.9.1, ETA Sunday March 19 :+1: :rocket:

PR #5076, opened 2017-03-15
Update description said "documents" instead of "one document"

bertolo1988:
People will notice that they can update several documents at once as soon as they learn about the  `multi` attribute a few lines below.


vkarpov15:
Thanks :beers:

PR #5080, opened 2017-03-15
fix(query): handle $type for arrays

### LINKED ISSUE #5079: Error: Can't use $type with Array.
There seems to be a bug when using $type queries on fields that are an array although that is supported by mongodb. See https://docs.mongodb.com/manual/reference/operator/query/type/#arrays for details.

using
node.js v6.9.5
mongoose v4.9.0
mongodb v3.4.1

Script to reproduce:

```
'use strict';

mongoose.connect('mongodb://localhost/github-mongoose-type-array-error', {});

const TestSchema = new mongoose.Schema({user: {type: [String], index: true}}, {});

const Test = mongoose.model('Test', TestSchema);

const query = { user: { $type: 'string' } };

Test.create({user: ['test']}, function(err) {
  if (err) {
    console.log('failed create');
    console.error(err);
  }
  Test.collection.find(query).next(function(err, doc) { //native find command works fine
    if (err) {
      console.log('failed native find');
      console.error(err);
    }
    console.log(doc);

    Test.find(query).exec(function(err, docs) { //mongoose find command doesn't work
      if (err) {
        console.log('failed mongoose find');
        console.error(err);
      }
      console.log(docs);
      process.exit();
    });
  });
});

```

could be related/similar to #4937

zoellner:
**Summary**

Fix #5079
allow $type for array fields

**Test plan**
Test script from bug report works as expected with this version

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -3,6 +3,7 @@
  */
 
 var $exists = require('./operators/exists');
+var $type = require('./operators/type');
 var SchemaType = require('../schematype');
 var CastError = SchemaType.CastError;
 var Types = {
@@ -319,6 +320,7 @@ handle.$minDistance =
 handle.$maxDistance = castToNumber;
 
 handle.$exists = $exists;
+handle.$type = $type;
 
 handle.$eq =
 handle.$gt =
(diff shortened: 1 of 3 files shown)


vkarpov15:
Great work, thanks :+1:

PR #5091, opened 2017-03-20
Handled passwords with @ in the connection URI

GavinDmello:
If there's a `@` symbol in the password, the host would not be correctly picked up.


vkarpov15:
Thanks for the PR, I'd love to accept it (as well as aheckmann/muri#12) but unfortunately this behavior [is explicitly forbidden in the mongodb spec](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#userinfo-optional), you [should use `encodeURIComponent()` on your password if it contains illegal chars](http://mongodb.github.io/node-mongodb-native/2.0/tutorials/connecting/#authentication:d67a243cdf521fafcb45cf3f560d3e12)

PR #5093, opened 2017-03-21
Update guide.jade

Jiang-Xuan:
I guess `unpredictable` is right.Because i can't find it when i use bing online dictionary.

<https://pageshot.net/m4637zhoHDeXviM1/cn.bing.com>

If i am wrong. Please don't keep in mind😁.


vkarpov15:
Thanks :beers:

PR #5129, opened 2017-03-28
Avoid exception on constructor check

monkbroc:
I was pushing object created by `Object.create(null);` into an array schema type. Since the prototype is `undefined` is causes `Cannot read property 'baseCasterConstructor' of undefined`.

This PR checks that the prototype is defined first.


vkarpov15:
Thanks! :beers:

PR #5131, opened 2017-03-29
document.js fixes for functions prepended with `$`

krmannix:
Small fix in the documentation for the errors that occur [here](http://mongoosejs.com/docs/api.html#document-js) and [here](http://mongoosejs.com/docs/api.html#document_function%20Object()%20%7B%20%5Bnative%20code%5D%20%7D-%24isDefault)

The issue was that `dox` wasn't correctly generating the function names for functions that start with `$` without explicitly specifying `@method` and `@memberOf`.

Current:
`$isDefault()`
screen shot 2017-03-29 at 10 09 45 am

`$ignore()`
screen shot 2017-03-29 at 10 09 06 am

Fixes:
`$isDefault()`
screen shot 2017-03-29 at 10 03 41 am

`$ignore()`
screen shot 2017-03-29 at 10 03 34 am

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1048,6 +1048,8 @@ Document.prototype.unmarkModified = function(path) {
  *     doc.$ignore('foo');
  *     doc.save() // changes to foo will not be persisted and validators won't be run
  *
+ * @memberOf Document
+ * @method $ignore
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks! :beers:

Issue #4500, opened 2016-09-08
Query.prototype.exec can swallow errors


jeanbaptiste-brasselet-planorama:
Hi,

We just upgraded recently to mongoose 4.6.0  and this is what I found : 

```
someModel.findById(someId).lean().exec(function(err, data) {
  if (err) return callback(err);
  throw new Error('This error will disappear');
  //do something
  return callback(null, data);
});
```

I can provide a more detailed example if it's not enough but basically the error will be swallowed and the callback will never be called.

This happens here : https://github.com/Automattic/mongoose/blob/master/lib/query.js#L2255 because the pattern

```
somePromise.then(function() {
  doSomething();
}, function(err) {
  //will catch if somePromise is rejected but will not 
  //check that something has happened during the doSomething
})
```

so maybe something like 

```
    promise.then(
      function() {
        callback.apply(null, _results);
      },
      function(error) {
        callback(error);
      }).catch(function(err) {
        throw new Error(err);
      });
```

Not sure about it.

What do you think ?

Issue #4379, opened 2016-07-30
Check unhandled rejections with queries


vkarpov15:
Re: #4364 but for queries

PR #5151, opened 2017-04-08
fix blue bird "a promise was created in a handler but was not returne…

maa105:
…d from it"

if u use the default promise library it wont show. if u use bluebird it is smart enough to catch this warning namely: "Warning: a promise was created in a handler but was not returned from it, see http://goo.gl/rRqMUw" this will resolve this issue here.

**Summary**

if u use bluebird promise library with mongoose. using simple queries will emit a warning:
    (node:8148) Warning: a promise was created in a handler but was not returned from it, see http://goo.gl/rRqMUw
        at Promise.then (xxx\bluebird\js\release\promise.js:124:17)

bluebird is intelligent and it looks for such discrepancies. this is also mentioned in bluebird issue https://github.com/petkaantonov/bluebird/issues/508 but the source i guess is mongoose not bluebird

**Test plan**

first set promise to bluebird:
mongoose.Promise = require('bluebird');

then try running any get query like:
User.findOne({'username':'maa'}).exec(function(err,user){...});

this is because at line 2743 when u do "promise.then" this will create a new promise. this new promise you are not returning. I see why u dont want to return it and instead u return the old promise (which is to keep the original state of the promise i.e. the resolved status of the promise). what I did is return this new promise and in its then and catch functions returned the original values/rejections to maintain its state.

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -2740,13 +2740,14 @@ Query.prototype.exec = function exec(op, callback) {
   });
 
   if (callback) {
-    promise.then(
+    return promise.then(
       function() {
         callback.apply(null, _results);
-        return null;
+        return Promise.ES6.resolve(_results);
       },
       function(error) {
         callback(error);
+        return Promise.ES6.reject(error);
       }).
       catch(function(error) {
         // If we made it here, we must have an error in the callback re:
--- a/lib/query.js
+++ b/lib/query.js
@@ -2742,19 +2742,30 @@ Query.prototype.exec = function exec(op, callback) {
   if (callback) {
     return promise.then(
       function() {
-        callback.apply(null, _results);
-        return Promise.ES6.resolve(_results);
+        var err;
+        try {
+          callback.apply(null, _results);
+        } catch(err) {
+          // If we made it here, we must have an error in the callback re:
+          // gh-4500, so we need to emit.
+          setImmediate(function() {
+            _this.model.emit('error', err);
+          });
+        }
+        return _results;
       },
       function(error) {
-        callback(error);
-        return Promise.ES6.reject(error);
-      }).
-      catch(function(error) {
-        // If we made it here, we must have an error in the callback re:
-        // gh-4500, so we need to emit.
-        setImmediate(function() {
-          _this.model.emit('error', error);
-        });
+        var err;
+        try {
+          callback(error);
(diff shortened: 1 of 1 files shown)


maa105:
in the mean time you can patch if by putting this code on ur application startup just after ur first mongoose require and preferably after the setting the mongoose promise

`
var bluebird = require('bluebird');
mongoose.Promise = bluebird;

(function() {
  var mongooseProto = Object.getPrototypeOf(mongoose);
  var PromiseProvider = mongooseProto.PromiseProvider;
  mongooseProto.Query.prototype.exec = function exec(op, callback) {
    var Promise = PromiseProvider.get();
    var _this = this;

    if (typeof op === 'function') {
      callback = op;
      op = null;
    } else if (typeof op === 'string') {
      this.op = op;
    }

    var _results;
    var promise = new Promise.ES6(function(resolve, reject) {
      if (!_this.op) {
        resolve();
        return;
      }

      _this[_this.op].call(_this, function(error, res) {
        if (error) {
          reject(error);
          return;
        }
        _results = arguments;
        resolve(res);
      });
    });

    if (callback) {
      return promise.then(
        function() {
          try {
            callback.apply(null, _results);
          } catch (err) {
            // If we made it here, we must have an error in the callback re:
            // gh-4500, so we need to emit.
            setImmediate(function() {
              _this.model.emit('error', err);
            });
          }
          return _results;
        },
        function(error) {
          try {
            callback(error);
          } catch (err) {
            // If we made it here, we must have an error in the callback re:
            // gh-4500, so we need to emit.
            setImmediate(function() {
              _this.model.emit('error', err);
            });
          }
          throw error;
        });
    }

    return promise;
  };
  mongooseProto = undefined;
})();
`

vkarpov15:
What version of bluebird? The below script emits no warnings for me:

```javascript

mongoose.Promise = require('bluebird');

var testSchema = new mongoose.Schema({ name: String });
var Test = mongoose.model('Test', testSchema);

Test.find({}).exec(err => { console.error(err) }).then(() => {});
```

But otherwise, lgtm, thanks for digging out this issue :+1:

vkarpov15:
Actually, I'm going to have to revert this one. It breaks our fix for #4379. Gonna have to find an alternative solution.

Issue #5141, opened 2017-04-02
Regression in pre update hook


clozanosanchez:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
In version 4.9.2 & 4.9.3, my pre update hooks are not being executed anymore. They run perfectly in version 4.9.1, if I install it again.

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Add a pre update hook to a schema:
```
schema.pre('update', myPreHook);
       
function myPreHook(next){
      console.log('PRE UPDATED')
      const query = this;
      ...
      next();
}
```
2. Execute an update operation:
```
const model = mongoose.model('test', schema);

model.update(updateQuery, {$set: updateData}, {runValidators: true}).exec()
   .then((updateResult) => {
        console.log('UPDATED')
        ...
    }).catch((error) => {
        ...
    });
```
3. The console writes 'UPDATED', but never 'PRE UPDATED'.

**What is the expected behavior?**
Pre update hook should be executed before executing the update operation.

**Please mention your node.js, mongoose and MongoDB version.**
Node.js: 6.10.1
Mongoose: 4.9.2 & 4.9.3
MongoDB: 2.2.25

varunjayaraman:
do you define your hooks before calling `mongoose.model()`?

clozanosanchez:
Yes, we define our hooks before calling mongoose.model()

clozanosanchez:
We think it is related to #5098 (that was fixed in version 4.9.2). 
We are cloning the schema and then we execute the model method. Maybe the prehooks are not being cloned properly.

```
....

initialSchema.pre('update', myPreHook);
       
function myPreHook(next){
      console.log('PRE UPDATED')
      const query = this;
      ...
      next();
}

....

const clonedSchema = initialSchema.clone();
const model = mongoose.model('test', clonedSchema);

model.update(updateQuery, {$set: updateData}, {runValidators: true}).exec()
   .then((updateResult) => {
        console.log('UPDATED')
        ...
    }).catch((error) => {
        ...
    });
```

Issue #5098, opened 2017-03-22
Schema.clone() still retains state


varunjayaraman:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Calling `Schema.clone()` still maintains a reference to the underlying schema such that calling `Parent.discriminator()` more than once will still throw an error relating to the discriminator schema because mongoose will complain the discriminator schema can't have `__t` (or whatever the discrim key is) as a field.

**If the current behavior is a bug, please provide the steps to reproduce.**

```javascript
const mongoose = require('mongoose');
const co = require('co');
const assert = require('assert');
const _ = require('lodash');

mongoose.Promise = global.Promise;

// Globals: place schemas here as a global to mimick the singleton pattern
const globals = {
  baseSchema: new mongoose.Schema({
    profile: {
      name: String,
      email: String
    }
  }, { timestamps: true }),

  discriminatorSchema: new mongoose.Schema({
    profile: {
      address: { city: String, line1: String }
    }
  })
};

// Constants
const GITHUB_ISSUE = `overwrite-discriminators-bug`;

describe('before each bug with discriminators', function() {
  let Base;
  let Discriminator;
  let db;

  function init() {
    return co(function* () {
      db = yield connectToDb();
      const models = createModels(db);
      Base = models.Base;
      Discriminator = models.Discriminator;
    });
  }

  describe('Block 1', function() {
    before(function(done) {
      co(function* () {
        yield init();
        done();
      }).catch(done);
    });

    after(function(done) {
      db.close();
      done();
    });

    it('test 1', function(done) {
      assert.ok(true);
      done();
    });

    it('test 2', function(done) {
      assert.ok(true);
      done();
    });
  });

  describe('Block 2', function() {
    beforeEach(function(done) {
      co(function* () {
        yield init();
        done();
      }).catch(done);
    });

    after(function(done) {
      db.close();
      done();
    });

    it('test 3', function(done) {
      assert.ok(true);
      done();
    });

    it('test 4', function(done) {
      assert.ok(true);
      done();
    });
  });
});

function createModels(db) {

  const Base = db.model('Base', globals.baseSchema);
  
  /**
   * This line is still necessary to prevent the shared state from the underlying schema
   */
  // _.unset(globals.discriminatorSchema, 'paths.__t');
  console.log(globals.discriminatorSchema.paths['__t']); // this should be undefined every time
[...]

PR #5153, opened 2017-04-08
Solves issue #5141, regression caused by #5098

clozanosanchez:
**Summary**
Solve issue #5141 caused by a regression in #5098 

**Test plan**
Cloned schemas are executing hooks again


lib/schema.js
```diff
@@ -348,6 +348,7 @@ Schema.prototype.clone = function() {
   s.callQueue = this.callQueue.map(function(f) { return f; });
   s.methods = utils.clone(this.methods);
   s.statics = utils.clone(this.statics);
+  s.s.hooks = utils.clone(this.s.hooks);
```
vkarpov15: The more correct way of doing this is `this.s.hooks.clone()` but either way works. Thanks for finding this :+1:

PR #5164, opened 2017-04-13
mongoose 4.8.3 broke $pullAll: null on subdocs

Sebmaster:

bug

```js
var mongoose = require('.');
var Promise = require('bluebird');

var Cat = mongoose.model('Cat', { name: String, props: [{
  name: String
}]});

(async function () {
  var kitty = new Cat({ name: 'Zildjian', props: [null, {name: 'abc'}] });
  await kitty.save();

  kitty = await Cat.findById(kitty._id);
  console.log(kitty.props)

  await Cat.update({ _id: kitty._id }, { $pullAll: { props: [null] }})

  kitty = await Cat.findById(kitty._id);
  console.log(kitty.props)
}());
```

Returns `[ null, { _id: 58efc1d703b69d6d1f1d2e72, name: 'abc' } ]`.

Should return `[ { _id: 58efc1d703b69d6d1f1d2e72, name: 'abc' } ]`.

node v7.8.0
mongoose 4.8.3, 4.9.4
mongodb 3.2

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -241,7 +241,7 @@ SchemaArray.prototype.castForQuery = function($conditional, value) {
           v = method.call(caster, v);
           return v;
         }
-        if (val != null) {
+        if (v != null) {
           v = new Constructor(v);
           return v;
         }


Sebmaster:
Aaand attached a PR to fix it. Seems to be just a wrong variable reference.

vkarpov15:
Good catch, thanks :beers:

PR #5172, opened 2017-04-17
Fix the bulkWrite::updateMany case

tzellman:
**Summary**

The `update` option was inadvertently being wrapped by the passed-in `filter` option. Most likely a copy-paste error in the code.

I ran into this issue when trying to do a bulkWrite->writeMany operation.

**Test plan**

No additional tests were added, as it should be easy to eyeball the error.

Diff

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2115,7 +2115,7 @@ Model.bulkWrite = function(ops, options, callback) {
         try {
           op['updateMany']['filter'] = cast(_this.schema,
             op['updateMany']['filter']);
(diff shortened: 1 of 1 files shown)


tzellman:
It looks like the test that failed is unrelated to the one-line change in this PR. Let me know if you need me to do anything else in order to get this merged. Thanks!

vkarpov15:
Good catch, sorry for the mistake. Thanks! :beers:

tzellman:
No worries, thanks!

PR #5178, opened 2017-04-18
Update schematype.js

milesbarr:
Fixed documentation for returning a promise from a validator to match documentation at http://mongoosejs.com/docs/validation.html.


vkarpov15:
Thanks :beers:

Issue #5189, opened 2017-04-20
Bug: Assigning modified document arrays causes validation errors


jhob:
This bug looks much like issues #1638 and #3511 but is more general. We're seeing it accidentally introduced in various forms throughout our usage of mongoose.

The issue arises when an array of embedded documents is rearranged in some way and assigned back to its parent, and then embedded documents are modified. A subsequent `validate` or `save` will throw errors in some cases. It appears the `modify` paths created for each modification of an embedded document are incorrect, as the indices of the element paths are not in sync with modified indices.

**Steps to reproduce:**
The following code reproduces the issue.

```
const mongoose = require('mongoose');

mongoose.connect('mongodb://localhost/test');
mongoose.Promise = global.Promise;

const Schema = new mongoose.Schema({
  items: [{foo: {type: String, required: true}}]
});
const Model = mongoose.model('TestModel', Schema);

Model.create({
  items: [{foo: 'hi'}, {foo: 'there'}],
}).then(function (model) {
  // directly assign a modified copy of the document array
  model.set({items: model.items.slice(1)})
  // modify the elements
  model.items.forEach((item, index) => {
    item.set({foo: 'bar'});
  });

  return model.save();
}).then(function () {
  console.log('success');
}).catch(function (err) {
  console.log(err);
}).then(function () {
  mongoose.disconnect();
});
```

Here I'm using `Array#slice`, but the issue exists with `Array#filter` and various other native array methods that return a copy rather than mutate the original document array.

Executing this code throws this exception:

```
{ MongooseError: TestModel validation failed
    at ValidationError (/tmp/mongoose/lib/error/validation.js:23:11)
    at model.Document.invalidate (/tmp/mongoose/lib/document.js:1599:32)
    at /tmp/mongoose/lib/document.js:1474:17
    at validate (/tmp/mongoose/lib/schematype.js:734:7)
    at /tmp/mongoose/lib/schematype.js:779:11
    at Array.forEach (native)
    at SchemaString.SchemaType.doValidate (/tmp/mongoose/lib/schematype.js:739:19)
    at /tmp/mongoose/lib/document.js:1472:9
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickCallback (internal/process/next_tick.js:104:9)
  errors:
   { 'items.1.foo':
      { MongooseError: Path `foo` is required.
          at ValidatorError (/tmp/mongoose/lib/error/validator.js:24:11)
[...]

PR #5192, opened 2017-04-20
Update `parentArray` references when directly assigning document arrays

jhob:
**Summary**

This is a proposal to fix the primary issue described in #5189.

**Test plan**

A test has been added that demonstrates the failure case without the included fix. See [`test/document.modified.test.js#468`](https://github.com/Automattic/mongoose/pull/5192/files#diff-2242966c7ef28bf9e498506e1bcff490R468). Steps to manually reproduce can be found in #5189.

**Notes**

I must admit it feels wrong reaching into `val.__parentArray` within `Document.prototype.$__set`. A new method to encapsulate this on `DocumentArray` might be preferable, and I'd be happy to propose one if this direction feels right.

In any case, modifying the value passed to `Document.prototype.set` might itself be quite undesirable, and other more involved solutions may avoid this, for example:
- cloning assigned document arrays and their embedded documents, with consistent internals like `__parentArray` and `__index`.
- override `Array.prototype.filter`, `Array.prototype.slice` etc. on `MongooseArray` to update parent references on returned arrays.
- do not keep a reference to `__parentArray` altogether, and instead store its path to retrieve the actual array when we need to generate element paths.

Feedback very welcome! This issue is raising its head for us quite often so any path forward would be much appreciated.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -878,6 +878,13 @@ Document.prototype.$__set = function(pathToMark, path, constructing, parts, sche
     if (val && val.isMongooseArray) {
       val._registerAtomic('$set', val);
 
+      // Update embedded document parent references (gh-5189)
+      if (val.isMongooseDocumentArray) {
+        val.forEach(function(item) {
+          item && item.__parentArray && (item.__parentArray = val);
+        });
+      }
+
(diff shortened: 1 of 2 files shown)


vkarpov15:
Thanks for the fix, LGTM :+1:

Issue #5190, opened 2017-04-20
Bug: insertMany fails on Decimal Types


alaa-ismail:
When using insertMany on a model whose schema contains a Decimal type, the operation fails with the error:
`Error: key $numberDecimal must not start with '$'`

**Steps to reproduce:**
The following code reproduces the issue: 
```javascript
const mongoose  = require('mongoose');
mongoose.Promise = global.Promise;

mongoose.connect('mongodb://localhost:27017/test');

var schema = new mongoose.Schema({
    amount : mongoose.Schema.Types.Decimal
});
var Money = mongoose.model('Money', schema);

Money.create({amount : '123.45'});  // Works fine

// This does not work
Money.insertMany([{amount : '123.45'},{amount : '321.54'}]).then((docs) => { 
    console.log(docs.length);
}).catch((err) => {
    console.log(err);
});
```
**Current Behavior:**
The insert operation fails with the error: 
`Error: key $numberDecimal must not start with '$'`

**Expectation:**
I expect the documents passed to insertMany to be inserted into MongoDB.

**Versions:**
Mongoose **v4.9.5**
MongoDB **v3.4.3**
Node **v6.9.1**

varunjayaraman:
thanks for the full repro script, this looks like a bug

varunjayaraman:
it looks like the mongoose `insertMany()` debug output might shed some light on this:

```
Mongoose: money.insertMany([ { __v: 0, amount: { '$numberDecimal': '132.45' }, _id: 58fea3fe3e
4166463e444c43 }, { __v: 0, amount: { '$numberDecimal': '321.54' }, _id: 58fea3fe3e4166463e444
c44 } ])
```

i believe in mongo it should be setting it as Decimal128 or `NumberDecimal` and the `create()` function seems to do just that:

```
Mongoose: money.insert({ amount: Decimal128 { _bsontype: 'Decimal128', bytes: <Buffer 39 30 00
 00 00 00 00 00 00 00 00 00 00 00 3c 30> }, _id: ObjectId("58fea4ef8a63684695772839"), __v: 0
})
```

nicroto:
I am getting this same error, when I install my node dependencies with `npm install --production`.
In development, the same create query works properly. Here is (a simplified ver of) the document I try to insert:

```javascript
    MyModel.create({
      ...
      "price": {
        "$numberDecimal": "9.99"
      }
      ...
    });
```

And this is the model descriptor:

```json
{
  ...
[...]

Issue #3931, opened 2016-03-02
Don't return promise if callback specified


artzstudio:
Hello, I'm the author of [meanify](https://github.com/artzstudio/meanify), a library that uses Mongoose to cleverly generate API routes.  

One of my users is bumping into an issue where he is seeing console.error output because I use callbacks to deal with Mongoose errors.  

Bluebird appears to expect an explicit catch statement, forcing one to prefer one style to another i.e. promises vs callbacks upon installation.

Consider the following code:

``` javascript
var mongoose = require('mongoose');
mongoose.Promise = require('bluebird');

var User = mongoose.model('User', new mongoose.Schema({
    name: { type: String, required: true },
}));

var user = new User();
user.save(function (err, doc) {
  if (err) {
    // error handled in the callback
    console.error('Error:', err.message);
  } else {
    console.log('Success:', doc);
  }
});
```

Results in this console output:

``` bash
Error: User validation failed
Unhandled rejection ValidationError: User validation failed
    at MongooseError.ValidationError (/Users/artz/Sites/meanify/node_modules/mongoose/lib/error/validation.js:23:11)
    at model.Document.invalidate (/Users/artz/Sites/meanify/node_modules/mongoose/lib/document.js:1298:32)
    at /Users/artz/Sites/meanify/node_modules/mongoose/lib/document.js:1173:16
    at validate (/Users/artz/Sites/meanify/node_modules/mongoose/lib/schematype.js:662:7)
    at /Users/artz/Sites/meanify/node_modules/mongoose/lib/schematype.js:693:9
    at Array.forEach (native)
    at SchemaString.SchemaType.doValidate (/Users/artz/Sites/meanify/node_modules/mongoose/lib/schematype.js:667:19)
    at /Users/artz/Sites/meanify/node_modules/mongoose/lib/document.js:1171:9
    at nextTickCallbackWith0Args (node.js:452:9)
    at process._tickCallback (node.js:381:13)
    at Function.Module.runMain (module.js:449:11)
    at startup (node.js:139:18)
    at node.js:999:3
```

My only solution to avoid this error for my users using Bluebird appears to be refactoring `meanify` to use promises so there's no error output.  Any advice would be much appreciated.  

Thanks!

vkarpov15:
Right now mongoose always returns a promise, even if you specify a callback. I'd love to change this for a future release. However, not happening anytime soon because we've got more pressing issues. Here's a couple alternatives:

[...]

PR #5212, opened 2017-04-30
Adding return to promise callback to suppress bluebird promise warnings

James9074:
**Summary**

When using .exec() callbacks, bluebird outputs the following error:

`(node:16) Warning: a promise was created in a handler at lib/query.js:2557:9 but was not returned from it, see http://goo.gl/rRqMUw
    at new Promise (node_modules/bluebird/js/release/promise.js:79:10)
`

Looks like bluebird wants to see the promise handler return _something_, I just added a standard promise resolve object and the warnings are now gone. 

**Note**: If you don't want to return a promise ([#3931](https://github.com/Automattic/mongoose/issues/3931)) , I was able to use `return null` just as currently is done on line 2554 in that same function, and bluebird warnings will still not appear anymore.

**Edit:** The travis build failed (`Promise.ES6.resolve is not a function`), so I reverted to returning null, which still works. The build still failed for an unrelated issue though:
 ```
1) Model bug fixes insertMany with Decimal (gh-5190):
     Uncaught MongoError: connection 863 to localhost:27017 closed)
```

**Test plan**

Here's an example of normal code that causes this issue:
`User.findOne({'_id': req.params.id}).populate('role','_id').exec(aCallback);`

After adding the return statement to the error callback, bluebird is happy and the warnings stop appearing.

Same issue as [d869ea5d1...](https://github.com/Automattic/mongoose/commit/d869ea5d1a173ba2a679703adb2995fd8cb912c5), but this solution is fairly simple, doesn't change any logic, and cleans up the test output files (which are currently cluttered with bluebird warnings)

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -2747,7 +2747,7 @@ Query.prototype.exec = function exec(op, callback) {
       },
       function(error) {
         callback(error);
-	return Promise.ES6.resolve(error);
+	return null;
       }).
       catch(function(error) {
         // If we made it here, we must have an error in the callback re:
--- a/lib/query.js
+++ b/lib/query.js
@@ -2747,7 +2747,7 @@ Query.prototype.exec = function exec(op, callback) {
       },
       function(error) {
         callback(error);
-	return null;
+	return Promise.ES6.resolve(error);
       }).
       catch(function(error) {
         // If we made it here, we must have an error in the callback re:
--- a/lib/query.js
+++ b/lib/query.js
@@ -2747,6 +2747,7 @@ Query.prototype.exec = function exec(op, callback) {
       },
       function(error) {
         callback(error);
+	return null;
       }).
       catch(function(error) {
         // If we made it here, we must have an error in the callback re:
--- a/lib/query.js
+++ b/lib/query.js
@@ -2747,7 +2747,7 @@ Query.prototype.exec = function exec(op, callback) {
       },
       function(error) {
         callback(error);
-	return null;
+        return null;
       }).
       catch(function(error) {
         // If we made it here, we must have an error in the callback re:

### CODE REVIEW

**Comment by vkarpov15** [comment]
I've seen this issue get reported before, but I've never been able to actually repro it. Below script doesn't print the warning with bluebird 3.5.0

**Comment by vkarpov15** [comment]
```javascript
const co = require('co');
mongoose.Promise = require('bluebird');
const GITHUB_ISSUE = `gh-5212`;


exec()
  .then(() => {
    console.log('successfully ran program');
    process.exit(0);
  })
  .catch(error => {
    console.error(`Error: ${ error }\n${ error.stack }`);
  });


function exec() {
  return co(function*() {
    mongoose.connect(`mongodb://localhost:27017/${ GITHUB_ISSUE }`)
    
    const User = mongoose.model('Test', new mongoose.Schema({
      role: { type: mongoose.Schema.Types.ObjectId, ref: 'Role' }
    }));
    const Role = mongoose.model('Role', new mongoose.Schema({}));

    yield User.findOne({}).populate('role','_id').exec(err => console.log(err.stack));
  });
}
```

James9074:
Odd... Perhaps you've got bluebird warnings disabled?

Also, note that this only applies to the error callback, calling .findOne({}) on my model doesn't produce an error, it just returns nothing.

For example: If I query `Role.findOne({'_id': 'blah'})` into model Role where _id is a `mongoose.Schema.Types.ObjectId`, then mongoose prints:
```
CastError: Cast to ObjectId failed for value "blah" at path "_id" for model "Role"`
```

 ...and Bluebird follows up with the warning mentioned in the initial PR referencing ther lack of a return statement at lib/query.js:2557

 
Try using this modified version of your script and see if it does anything (warnings should be on by default, but I'm throwing the `warnings:true` in there just to rule out that possibility:

```
const co = require('co');
mongoose.Promise = require('bluebird');
mongoose.Promise.config({ warnings: true });
const GITHUB_ISSUE = `gh-5212`;

exec()
  .then(() => {
    console.log('successfully ran program');
    process.exit(0);
  })
  .catch(error => {
    console.error(`Error: ${ error }\n${ error.stack }`);
  });


function exec() {
  return co(function*() {
    mongoose.connect(`mongodb://localhost:27017/${ GITHUB_ISSUE }`)
    
    const User = mongoose.model('Test', new mongoose.Schema({
      role: { type: mongoose.Schema.Types.ObjectId, ref: 'Role' }
    }));
    const Role = mongoose.model('Role', new mongoose.Schema({}));

    yield User.findOne({'_id':'blah'}).populate('role','_id').exec(err => console.log(err.stack));
  });
}
```

vkarpov15:
The above script yells at me because `mongoose.Promise.config({ warnings: true });` <-- `mongoose.Promise.config` is not a function. Even if I remove the yield statement and the process.exit() I still don't get the warning... Also tried running `env BLUEBIRD_W_FORGOTTEN_RETURN=true node gh-5212.js` and no warning. @petkaantonov any insights you might be able to offer on how to repro this warning?

James9074:
Sorry for the delay, I'm still looking into this, I promise. Hopefully will have something soon.

James9074:
Closing due to lack of time. Will re-open if I get around to a real fix + demo.

vkarpov15:
Thanks @James9074 . Any help you or anyone else can provide in reproducing this issue would be much appreciated :+1:

Issue #5211, opened 2017-04-30
SchemaType Boolean tries to convert input


c0d0g3n:
I encountered a bug with schemaType Boolean where its cast method converts the input to `true` (or `false`?) as a last resort. I think it should throw or spit a validation error if the input is not a boolean, a string representation of a boolean or 0 or 1. Right now it returns `!!value` (https://github.com/Automattic/mongoose/blob/master/lib/schema/boolean.js#L70), which seems  unexpected to me and may introduce subtle bugs. In addition to that, writing a custom validator (to override this behavior) is pointless, because the passed value is `true` or `false` anyway.

I suspect that mongoose itself is tricked by current behavior: https://github.com/Automattic/mongoose/blob/master/lib/schema/boolean.js#L47

[Here is a gist](https://gist.github.com/c0d0g3n/40b4b62041969abfafe4199c4101c047) showing the current behavior, as well as providing a temporary workaround.

I'm using mongoose 4.9.6.

c0d0g3n:
As this issue is labeled as discussion, let me give a clearer example of what I mean:

Say you want to store in the database whether a visitor likes pancakes or not, represented by a boolean. You receive the info in JSON, like this:

    {
        "likePancakes": true
    }

You apply the new value like this:

    user.likePancakes = input.likePancakes
    user.save()

So far, so good. Note that input validation is left to Mongoose as validating here would be redundant and inconsistent relating error messages etc. (The issue could be due to my reasoning right here.)

Now, assume the visitor passes this:

    {
        likePancakes: "nope"
    }

Currently, Mongoose parses this as `true`, even though this is clearly not the visitor's intent, which creates confusion. Better would be to just let the validation fail.

Hope this makes my point more clear :)

P.s. I've never made a pull request before, but if we can agree on the wanted behavior of this schemaType, I am willing to try.

c0d0g3n:
@varunjayaraman I've made an [alternative implementation](https://github.com/c0d0g3n/mongoose/commit/d57351e57077047b2deafaa5e786f7ae6e033728) as a matter of practice. What do you think about it and where would be the best place to document this new option? Without enabling the `strict` option, behavior will remain intact, so this is fully backwards-compatible.

varunjayaraman:
this looks good to me. looping in @vkarpov15 to sign off on it. And if you want to write a test, that would be A+, otherwise i can knock 1 or 2 out pretty easily

c0d0g3n:
[...]

Issue #4245, opened 2016-06-16
Mongoose Boolean validation not working if a String is provided


andresgariglio:
Given the following schema with a Boolean field:

```
var UserSchema = new Schema({
  _id: {
    type: Number
  },
  enabled: {
    type: Boolean,
    default: false
  }
});
```

When trying to save a document, if I supply a random string instead of a Boolean value or "Boolean string" (e.g. "false" or "true" instead of just false/true) for enabled field, the document is still being saved properly, with the Boolean field set to true. 

Expected behavior: save operation should throw a ValidationError.

vkarpov15:
Yeah unfortunately this is backwards breaking, mongoose will cast anything into a boolean and this is expected behavior. Will make this change for the next major release.

PR #5288, opened 2017-05-25
Boolean strict mode #5211

c0d0g3n:
See #5211 and #4245. Strict mode (`strict: true`) throws cast error if input is not a boolean representation. Not backwards breaking as new behavior is opt-in via option. Unit test for accepted values included. Could not add documentation because `make docs` errors (should go into separate issue?).

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,11 @@
+4.9.8 / 2017-05-07
+==================
+ * docs(subdocs): rewrite subdocs guide #5217
+ * fix(document): avoid circular JSON if error in doc array under single nested subdoc #5208
+ * fix(document): set intermediate empty objects for deeply nested undefined paths before path itself #5206
+ * fix(schema): throw error if first param to schema.plugin() is not a function #5201
+ * perf(document): major speedup in validating subdocs (50x in some cases) #5191
+
 4.9.7 / 2017-04-30
 ==================
  * docs: fix typo #5204 [phutchins](https://github.com/phutchins)
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -143,41 +143,14 @@ Connection.prototype.db;
 
 Connection.prototype.config;
 
-/**
- * Opens the connection to MongoDB.
- *
- * `options` is a hash with the following possible properties:
- *
- *     config  - passed to the connection config instance
- *     db      - passed to the connection db instance
- *     server  - passed to the connection server instance(s)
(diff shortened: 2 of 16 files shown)


vkarpov15:
So first your branch is very confusing, hard to tell what actually changed. But looking at d57351e, I don't think I can merge this. `strict: true` is on by default, and already has its own semantics. I don't want to conflate new behavior with strict mode. I think a [custom schema type](http://mongoosejs.com/docs/customschematypes.html) or a plugin that replaces mongoose's `Boolean.prototype.cast` function would be more appropriate in this case.

c0d0g3n:
Apologies, I indeed made a bit of a mess. For clarification, d57351e57077047b2deafaa5e786f7ae6e033728 adds the strict option, ede9c64d0d06e22312656bdd34d612ba823a400a adds a unit test. 8212b49bb19d14c5ae2bd5587a63a5a07431f508 (getting up-to-date) was a pull/merge (I don't know anymore) from upstream, which in hindsight was probably not the way things go (Sorry, it's my first pr ever :)). I can create a new, proper pull request if you want.

You're probably referring to this: http://mongoosejs.com/docs/guide.html#strict? I thought it wouldn't be a problem because that option only applies to schemas, whereas the pr operates only on paths of type `Boolean`. How about changing `strict` in this pr to `strictBool` or `strictCast` to avoid confusion?

vkarpov15:
I'm not very excited to add yet another option to schemas, that's fast becoming a dumping ground for 4.x behaviors that we want to get rid of. However, I can stomach it, `strictBool` it is :)

c0d0g3n:
I understand that it would be more elegant to update behavior in 5.0, if that is what you mean. I've a workaround published [here](https://gist.github.com/c0d0g3n/40b4b62041969abfafe4199c4101c047) for now. I was also thinking if it would be useful that users can customize the values that are accepted as true and false, because in extreme cases users may not want to validate f.i. `'true'` as `true` either. (Maybe we should even abolish converting altogether and introduce something like Wordpress filter hooks.) This may be a little far fetched though.

Anyway, just give me a shout if you still want a clean pr with `strictBool` and docs.

vkarpov15:
Users can customize the values by overwriting the boolean type's `cast()` function, I don't see a need to add any more abstractions there currently. However, if you're still willing to do a clean `strictBool` PR I'd be happy to merge it :+1:

PR #5292, opened 2017-05-25
reveal subtopics in sidebar on first click

### LINKED ISSUE #5291: 2 clicks to reveal sub-topics in sidebar & broken glyphs [mongoose docs]
It is currently very difficult to navigate the mongoose documentation using the sidebar. And it is only through trial and error that I discovered that if you click *twice* on an item it will reveal a dropdown of available methods (a lovely feature!).

**Behavior**
- When I visit http://mongoosejs.com/docs/api.html
- ...and click on a menu item in the sidebar (1)
- ...I see the unicode "□" (not found) appear (2)
- And should I click the menu item again
- ..I would see the desired dropdown list of sub-topics (3), but the caret fails to toggle from right to down.

Screenshots reproducing steps 1, 2, 3:
1-2-3-mongoose

- Clicking on an item should immediately reveal a list of sub-topics
- I should not see a "□"

Screenshot/suggestion of how it should look after 1 click (and only 1 click):
screen shot 2017-05-25 at 11 37 23 am

nathanallen:
fixes #5291

**Summary**

* Missing font-awesome "fa" class caused incorrect font to be used, leading to unicode □ appearing instead of the desired caret.
* Desired "toggling" behavior on first click was added by modifying/simplifying existing javascript.

**Test plan**

Running `make docs` locally leads to correct toggling behavior.

screenshot

### DIFF
--- a/docs/api.jade
+++ b/docs/api.jade
@@ -279,14 +279,15 @@ html(lang='en')
     script(src="/docs/js/cookies.min.js")
     script.
       $(".toggle-item").click(function() {
-        if ($(this).hasClass('fa-caret-right')) {
-          $(this).removeClass('fa-caret-right');
-          $(this).addClass('fa-caret-down');
-          $(this).parent().addClass('displayed');
+        var $parent = $(this).parent(),
+            $icon   = $(this).find("i");
+
+        if($parent.hasClass('displayed')){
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks :beers:

PR #5297, opened 2017-05-26
Tweak updateOne and replaceOne documentation notes

dhritzkiv:
It seems that the note about the hooks that are called after `updateMany` was copy+pasted into the `updateOne` and `replaceOne` documentation.

It either does not apply, or the reference to `updateMany` needs to be updated to say `updateOne`/`replaceOne`

**Summary**

**Test plan**


vkarpov15:
Good catch, thanks :+1:

PR #5298, opened 2017-05-27
eachAsync `parallel` option

### LINKED ISSUE #4244: Add a pool size to QueryCursor#eachAsync
I wrote a little wrapper that pretty much acts like `QueryCursor#eachAsync(fn, [callback])` but with a small difference which basically allows it to have a pool of items available, enabling to exec faster and to make stuff in parallel.
### Use case:

Each `fn` given to `eachAsync` is a promise that makes some IO and takes 5s to run. `eachAsync` would wait it to be finished before executing a new one but it doesn't mean you cannot execute multiple at once and to always have a pool of X `fn` executing at the same time, which would enable a higher throughput.

Hence, if `eachAsync` could buffer the next Y items to not have to wait for a new IO between mongo and the Node app it would enable a higher throughput too.
### API suggestion:

`QueryCursor#eachAsync(fn, [options], [callback])`
where `options` is something like:

``` js
{
    poolSize: >=1, // How many [fn] will be executed in parallel
    bufferSize: >=0 // How many items shall be retrieved in advance
}
```

If it has any interest I could think about making a PR for it.

Thanks for the good job guys!

vkarpov15:
Fix #4244


alpfactory-alex:
Why has this been merged? According to mongo-db driver docs this is not supposed to work:

https://jira.mongodb.org/browse/NODE-926

And this is exactly the error I am getting when trying to use parallel > 1

Or did I miss something?

vkarpov15:
Quite possible, we might have to remove this. Do you have a mongoose issue open re: this? Perhaps #5376 ?

vkarpov15:
Also, if you can provide a script that reproduces this issue that would be much appreciated

alpfactory-alex:
@vkarpov15 Actually it happens with any code. As soon as you use parallel promises on cursors mongo is going to break as this is unsupported. See also the referenced JIRA Ticket in Mongo stating its not supported.

PR #5301, opened 2017-05-27
Add data dirs in tools/ to .npmignore

Starfox64:
**Summary**

75052328bf5f5544194f15137f1bf749e8125eae added the data directories in tools/ to `.gitignore` but not `.npmignore` resulting in the latest npm release to be 900MiB large. This PR will add these data directories to .npmignore.


vkarpov15:
Good catch, sorry about that :+1:

Issue #5176, opened 2017-04-18
autoIndex doesn't work for `mongoose.connect()`


vkarpov15:
See https://github.com/Automattic/mongoose/issues/1875#issuecomment-292855301

vkarpov15:
Upon closer inspection it looks like autoIndex is working as expected, the below script prints nothing, but if you change `autoIndex` to true it prints that it called `ensureIndex`. @yasharma can you please modify the below script to reproduce your issue? Couple questions:

1) Are you sure you don't have a plugin that's setting `autoIndex` on your schema?
2) Are you expecting that mongoose would remove the index for you if `autoIndex` is false? If `autoIndex` is false, mongoose will not remove existing indexes, so any indexes in the db will stay there.

mhueter:
Hi @vkarpov15 , I think I'm experiencing this issue. I was experiencing large spikes in my response time in production code for some queries, and through the debug setting, I think I have tracked it down to `ensureIndex` being called on the slow queries, which brought me to this issue.

Here is my `server.js` snippet:

`mongoose.connect(uri, { user: '', pass: '', replset: '', config: { autoIndex: false } })`

Then, with debugging enabled, it prints immediately after that:

`Mongoose: sites.ensureIndex({ subdomain: 1 }, { background: true })`

This happens when index is set on the prop (i.e. `subdomain: { type: string, index: true }` or when `sites.index({ subdomain: 1})` is set in the schema definition file.

vkarpov15:
Below script reproduces the issue. As a workaround, make sure you call `mongoose.connect()` **before** `mongoose.model()`

PR #5324, opened 2017-06-05
Fix(ensureIndexes) manual call with !schema.options.autoIndex

whyhankee:
Sometimes the test would fail (based on the order the database would return the results)

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -956,11 +956,6 @@ function _ensureIndexes(model, options, callback) {
   };
 
   var create = function() {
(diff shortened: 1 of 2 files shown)


whyhankee:
The failing test was already in master when I made the pull-request :(

vkarpov15:
So just to be clear this is re: #5328. The code you're removing from `model.js` was added to fix #5176, so the fix isn't quite that simple, that code was added by design. Will add an alternative fix.

Issue #5321, opened 2017-06-03
Querying aliases feature


rocketspacer:
Following #5184, I now want to implement the ability to query (and update) based on aliased fields. I've been looking at the code for a while. At first, I was looking for ways apply conditions aliasing on each and every model query. But then, I came up with another solution, which looks something like:  
```
var Foo = mongoose.Model('Foo', FooSchema);

Foo.findOneAndUpdate(
    Foo.translateAliases({aliasedField: ...}),
    Foo.translateAliases({aliasedField: ...})
)
```
Which I think is neater, clearer, and obviously optional. Plus, this won't change our codebase too much, avoiding cases when we forgot to do conditions aliasing in some queries.

I want to know if you feel okay going down this approach, so I put it here for feedback

vkarpov15:
`translateAliases()` would definitely be a great step towards making aliases work seamlessly. If that function works, we can easily add it as a hook and then have it work transparently. Good idea :+1:

wong2:
any progress?

vkarpov15:
Nope not yet

rafaelcorreiapoli:
Anu updates on this?

vkarpov15:
@rafaelcorreiapoli no, not yet. Will bump this up for the next minor version.

vkarpov15:
I didn't quite read this thread carefully, it looks like this was fixed by #5338. [Here's the docs for `Model.translateAlias()`](https://mongoosejs.com/docs/api/model.html#model_Model.translateAliases) @rafaelcorreiapoli

PR #5338, opened 2017-06-08
Query and construct document with aliases with Model.translateAliases

rocketspacer:
Follow up from #5321   
Usage:
```
var Character = mongoose.model('Character', new mongoose.Schema({
  name: { type: String, alias: '名' },
  bio: {
    age: { type: Number, alias: '年齢' }
  }
}));

// The condition for the 'find' query will become { name: 'Stark', bio: { age: 40 } }
Character
  .find(Charater.translateAliases({
    '名': 'Stark',
    '年齢': 40
  })
  .exec((err, characters) => {})

// Can be use similarily when constructing new document
var char = new Character(Character.translateAliases({
  '名': 'Stark',
  '年齢': 40
});
```

### DIFF
--- a/test/model.translateAliases.test.js
+++ b/test/model.translateAliases.test.js
@@ -24,9 +24,7 @@ describe('model translate aliases', function() {
       // How translated aliases suppose to look like
       {
         name: 'Stark',
-        bio: {
-          age: 30
-        }
+        'bio.age': 30
       }
     );
   });
--- a/test/model.translateAliases.test.js
+++ b/test/model.translateAliases.test.js
@@ -1,0 +1,33 @@
+/**
+ * Test dependencies.
+ */
+
+var start = require('./common'),
+    assert = require('power-assert'),
+    mongoose = start.mongoose;
+
+describe('model translate aliases', function() {
+  it('should translate correctly', function() {
+    var Character = mongoose.model('Character', new mongoose.Schema({
+      name: { type: String, alias: '名' },
+      bio: {
+        age: { type: Number, alias: '年齢' }
+      }
+    }));
+
+    assert.deepEqual(
+      // Translate aliases
+      Character.translateAliases({
+        '名': 'Stark',
+        '年齢': 30
+      }),
+      // How translated aliases suppose to look like
+      {
+        name: 'Stark',
+        bio: {
+          age: 30
+        }
+      }
+    );
+  });
+});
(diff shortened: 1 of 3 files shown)


rocketspacer:
I don't think the CI check fails because of me. Can you try rebuilding the PR?

rocketspacer:
Maybe I should put it under Schema instead of Model?

test/model.translateAliases.test.js
```diff
@@ -0,0 +1,33 @@
...
+    }));
+
+    assert.deepEqual(
+      // Translate aliases
+      Character.translateAliases({
+        '名': 'Stark',
+        '年齢': 30
+      }),
+      // How translated aliases suppose to look like
+      {
```
vkarpov15: I'm not sure separating out dotted paths is the right behavior here. If you're querying by '年齢', you'd want to query `'bio.age': 30` rather than `bio: { age: 30 }`, because the latter will only match if `bio` is an object that contains exactly 1 key, `age`. On the other hand, `'bio.age': 30` would match even if `bio` has other keys, for example `birthday`.

rocketspacer: My bad, I was over thinking it  
Do you prefer that I put it under schema instead of model?

vkarpov15: Hmm. I think `schema` is actually the better place to put the logic, but there should still be a helper function on model. But this code looks good, I'm gonna merge it into the 4.11 branch

bencmbrook:
Any plans to have querying with aliases without additional query code?
e.g. 
```js
Person.findOne({ name: 'Ben' }, ...)
```
where
```js
var personSchema = new Schema({
  n: {
    type: String,
    alias: 'name'
  }
});
```
Would be a great way to change schema without breaking existing query code

rocketspacer:
Pre-hooks seems like a good way to implement transparent aliased queries. I'll look on to it next weekend

test/model.translateAliases.test.js
mozi: dotted paths needs more attention as it could be queried with and without dots within $elemMatch as per the example below

- `"usuario.comentarios.gustos": { $gt: 1 }` translates into `"user.comments.likes": { $gt: 1}` 
- `"usuario.comentarios": { "$elemMatch": { "gustos": { $gt: 1 } } }` translates into `"user.comments":  { "$elemMatch": { "likes": { $gt: 1 } } }`

vkarpov15:
Should be easy to implement as a plugin, but haven't tried. I appreciate the suggestion, but please open up a new issue instead of commenting on a closed one :+1:

PR #5344, opened 2017-06-10
`strictBool` option for schema type boolean (clean)

c0d0g3n:
**Summary**
See #5288, #5211 and #4245. Adds an option for schema type Boolean, called `strictBool` (disabled by default to ensure compatibility). When `strictBool:true` is passed, path input is restricted to `true`, `false`, `'true'`, `'false'`, `1`, `0`, `'1'` and `'0'`. Other values throw a `castError`. `null` is also accepted (as before) so the field can be marked as unspecified.

**Test plan**
0fb09d9 (second one) includes a test which verifies that the values we expect to be accepted are indeed accepted. (Need to test that other values are rejected?)

`npm test -- -g 'strictBool'`: test passed

**TODO**
- Document new option.

### DIFF
--- a/test/schema.boolean.test.js
+++ b/test/schema.boolean.test.js
@@ -29,5 +29,21 @@ describe('schematype', function() {
       assert.strictEqual(true, m3.b);
       done();
     });
+    it('strictBool option (gh-5211)', function() {
+      console.log('chekc');
+      var db = start(),
+          s1 = new Schema({b: {type: Boolean, strictBool: true}}),
+          M1 = db.model('StrictBoolTrue', s1);
+      db.close();
+
+      var m1 = new M1;
+      var strictValues = [true, false, 'true', 'false', 0, 1, '0', '1'];
+      var validatePromises = strictValues.map(function(value) {
+        m1.b = value;
+        return m1.validate();
+      });
+
+      return global.Promise.all(validatePromises);
+    });
   });
 });
--- a/test/schema.boolean.test.js
+++ b/test/schema.boolean.test.js
@@ -29,21 +29,29 @@ describe('schematype', function() {
       assert.strictEqual(true, m3.b);
       done();
     });
-    it('strictBool option (gh-5211)', function() {
-      console.log('chekc');
+    it('strictBool option (gh-5211)', function(done) {
       var db = start(),
           s1 = new Schema({b: {type: Boolean, strictBool: true}}),
           M1 = db.model('StrictBoolTrue', s1);
       db.close();
 
-      var m1 = new M1;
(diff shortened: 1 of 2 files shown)


test/schema.boolean.test.js
```diff
@@ -29,5 +29,21 @@ describe('schematype', function() {
...
+      db.close();
+
+      var m1 = new M1;
+      var strictValues = [true, false, 'true', 'false', 0, 1, '0', '1'];
+      var validatePromises = strictValues.map(function(value) {
+        m1.b = value;
+        return m1.validate();
+      });
+
+      return global.Promise.all(validatePromises);
```
vkarpov15: This will fail on node < 4 because no promises :p . Just use a for loop to chain all the promises together.

test/schema.boolean.test.js
```diff
@@ -29,5 +29,21 @@ describe('schematype', function() {
       assert.strictEqual(true, m3.b);
       done();
     });
+    it('strictBool option (gh-5211)', function() {
+      console.log('chekc');
```
vkarpov15: Remove this please

ekulabuhov:
@vkarpov15 @c0d0g3n is there anything else preventing this from being merged except for failing style tests? I have a need for this feature and can help to complete it if you're ok with this?

c0d0g3n:
The issues mentioned by @vkarpov15 should have been adressed in last commit. (But now I think of it, the `Array.forEach` I used may not be compatible with node <4 as well.) You're more than welcome to add this feature to the docs though.🙂 (Last time I tried, I couldn't get the build system to work.)

ekulabuhov:
Hi! I just used this feature and it works like a charm. The only problem is that I have 10 or so Booleans on a single Schema. Would it make sense to move this variable to the Schema level? Or even higher? Seems like a reasonable default to enforce.

ancmikesmullin:
@vkarpov15 why was this feature removed? 
https://github.com/c0d0g3n/mongoose/blob/master/lib/schema/boolean.js#L62-L73
no longer in master
https://github.com/Automattic/mongoose/commit/6227232ad324d05b8e1cdb21633602e9c6f00e61#diff-17c0b6da40c0e37e45bc929cbd5b7056

i would like to disable the strictness. OR
i would like the string value `'on'` to be in the list of truthy values, as this is the default value provided by browsers like Chrome for `<input type="checkbox"/>` elements that have no `value=` attribute set.

vkarpov15:
@ancmikesmullin we removed this for 5.0 because converting any non falsy value to true was a bit too permissive. We'll open up an alternative way to support this behavior, but for now you can just do a custom setter `set: v => v === 'on' ? true : v`

PR #5357, opened 2017-06-13
Fix `parseOption` mutates user passed option map

igwejk:
**Summary**

Consider the following code snippet:

```javascript
var mongoose = require( 'mongoose' );

var options = Object.create( null, {
    auth: { value: {} }
} );

// internally calls `parseOption( options, connStrOpts /* options that were passed in the connection string, if any */ )`
mongoose.connect( 'mongodb://localhost/my_database', options );
```

The issue is that `parseOption( options )` currently mutates the passed `options` parameter. Should that really happen?
- Library users may not expect that their data will possibly get mutated by mongosse
- If a library user passes a secured object composed with `Object.create( ... )`, the connection attempt will fail
- Mutating users references is not playing-nice, imho

The change in this PR resolves the described issue.

**Test plan**

A unit test for this change exists in _test/parse.options.test.js_. Run the unit test specific to this change with

    $ mocha test/parse.options.test.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
   "dependencies": {
     "async": "2.1.4",
     "bson": "~1.0.4",
-    "clone": "^2.1.1",
     "hooks-fixed": "2.0.0",
     "kareem": "1.4.1",
     "mongodb": "2.2.27",
@@ -31,6 +30,7 @@
     "ms": "2.0.0",
     "muri": "1.2.1",
     "regexp-clone": "0.0.1",
+    "sinon": "~2.3.4",
     "sliced": "1.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


igwejk:
Please note that build failure is not due to this change

package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "async": "2.1.4",
     "bson": "~1.0.4",
+    "clone": "^2.1.1",
```
vkarpov15: Not going to add an external dependency, especially one with a very dangerous upcarat. Use `require('../../util').clone()` instead please in :point_up:

igwejk: Totally agree, I will update right away, thanks :-)

igwejk:
@vkarpov15 it may be a good idea to implement a simple test-double library in the `utils.js` module, what do you think?

Issue #1860, opened 2014-01-04
Required fields on non-required nested documents


keithminder:
/*
When a schema contains a field that is a nested document, then the nested document is by default non-required.  Adding a required field to a nested document makes the nested document required field validation fail in cases in which property's value is null.   This should not be the case and the document should pass validation in this case.
 
Consider the following:
*/
 var applicationSchema = mongoose.Schema( {
 firstName: { type: String, required: true }
 , lastName: { type: String, required: true }
 , father: {
  firstName: { type: String, required: true }, middleName: String, lastName: { type: String, required: true }, address1: { type: String, required: true }, address2: String, city: { type: String, required: true }, state: { type: String, required: true }, zip: { type: String, required: true }
}
);
var Application=  mongoose.model('application',applicationSchema);
/*
     the following document should pass validation but does not:
     {firstName:'shaheed', lastName:'isthatdude'}
*/
var application = {firstName:'shaheed', lastName:'isthatdude'};  //pojo
var applicationToSave = new Application(application);
applicationToSave.save(function(err,saved){if(err){console.log('error: '+ err);}});
 
/*
    Using custom validation works as expected:
*/
function requiredOnOptional(val)
{

  return (val instanceof String || typeof val == 'string') && val.length;

}
 var applicationSchema = mongoose.Schema( {
 firstName: { type: String, required: true }
 , lastName: { type: String, required: true }
 , father: {
  firstName: { type: String, validate:requiredOnOptional }, middleName: String, lastName: { type: String, , validate:requiredOnOptional }, address1: { type: String, , validate:requiredOnOptional }, address2: String, city: { type: String, , validate:requiredOnOptional }, state: { type: String, , validate:requiredOnOptional }, zip: { type: String, , validate:requiredOnOptional }
}
);
var Application=  mongoose.model('application',applicationSchema);
var application = {firstName:'shaheed', lastName:'isthatdude'};  //pojo
var applicationToSave = new Application(application);
applicationToSave.save(function(err,saved){if(err){console.log('error: '+ err);}});

//the above example bypasses the validation of the null nested field's properties.  This is inconsistent and I think that it should work as it does with custom validation.

mr-e-:
Why is this closed? Was it fixed?

vkarpov15:
[...]

PR #5387, opened 2017-06-21
Makes Indexed Paths for Optional Nested Documents Sparse

c1moore:
**Summary**

This PR addresses my comments made on issue #1860.  Specifically, if a document has optional nested documents which themselves have fields that are indexed, an index will be created on the main document for that indexed field.  When 2 documents that omit the optional nested document are saved, an error is thrown for duplicate keys.  This leads to behavior similar to the analogy I made where "I don't have to have a kid, but I better know his birthday."

It is also inappropriate to require the sub-schema that creates the index for this field to declare the index sparse in some situations where the sub-schema may be used in a variety of different situations, each with differing needs.  Some of these situations may not require the sub-schema while others do.  This makes it difficult or impossible to make the sub-schema accurately make an index sparse.

This PR solves that problem by making the indexed fields of nested documents sparse if and only if the nested document (or one of its ancestors) are optional.

**Test plan**

I added 3 new test cases to confirm that indexed fields within subdocuments are made sparse only when expected.  All other tests are passing.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1362,14 +1362,14 @@ Schema.prototype.indexes = function() {
         subIndex = 0,
         keyIndex;
 
-    for(subIndex = 0; subIndex < len; ++subIndex) {
+    for (subIndex = 0; subIndex < len; ++subIndex) {
       indexObj = subindexes[subIndex][0];
       keys = Object.keys(indexObj);
       klen = keys.length;
       newindex = {};
 
       // use forward iteration, order matters
-      for(keyIndex = 0; keyIndex < klen; ++keyIndex) {
+      for (keyIndex = 0; keyIndex < klen; ++keyIndex) {
         key = keys[keyIndex];
         newindex[prefix + key] = indexObj[key];
       }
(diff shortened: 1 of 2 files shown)


lib/schema.js
```diff
@@ -1345,45 +1345,99 @@ Schema.prototype.indexes = function() {
...
-    for (var i = 0; i < keys.length; ++i) {
-      key = keys[i];
+    for (keyIndex = 0; keyIndex < keys.length; ++keyIndex) {
+      key = keys[keyIndex];
       path = schema.paths[key];
 
-      if ((path instanceof MongooseTypes.DocumentArray) || path.$isSingleNested) {
-        collectIndexes(path.schema, prefix + key + '.');
+      if (path instanceof MongooseTypes.DocumentArray || path.$isSingleNested) {
+        collectIndexes(path.schema, prefix + key + '.', !path.isRequired || optional);
```
vkarpov15: Problem: `isRequired` is not necessarily known until you create a document, `required: function() { return false; }` is also supported.

c1moore:
But if the behavior is moved to a plugin that means the index has to be created by mongoose then removed later by the plugin, which isn't very efficient and might lead to race conditions and unexpected results.  Plus it's not really "magic", it's expected behavior that optional nested documents should not throw errors when they are not specified.  The current behavior is unexpected.

Also, if a function is defined that usually means that at least part of the time the result can be false, which suggests that the index should be made sparse anyways.  If the user wanted `required` to always be false, they should just make `required` false.

vkarpov15:
If the behavior is created by a plugin, that means the plugin modifies the index before it is created by mongoose, so you can get this same behavior in a plugin. The optional nested docs throwing errors is surprising from a certain perspective, but mongoose inferring index properties is similarly surprising, especially since index management is a non-goal for the project.

c1moore:
Damned if you do, damned if you don't I guess.  I understand the decision, but I'm not sure I agree with it since both may result in unexpected behavior but the latter is more intuitive and reflective of what we see in real life.  Also while the former (current) situation is remedied by a plugin, the latter situation could be remedied just as easily internally using some sort of flag (similar to `strict`) or alternatively making the current behavior default and using a flag to turn on the suggested behavior.

What would a plugin that accomplished this same goal look like?  More specifically, how could I make an index sparse after the Schema is created?  I haven't come across this before and I don't see any documentation for doing that.  Would I need to use a schema-level index and redefine the original index?  Or would the plugin have to go into mongoose's internals and take advantage of otherwise secret/private interfaces and/or data?  I need a solution for this so I'll work on implementing it tomorrow.

vkarpov15:
In general I'd argue that mongoose creating indexes is a useful tool for testing/dev but should not be used in production, and adding sophisticated index management starts making devs more dependent on mongoose for building indexes.

Here's a quick example that turns all indexes into sparse indexes

```javascript
module.exports = function sparseIndexPlugin(schema) {
  schema.indexes().forEach(function(indexArgs) {
    indexArgs[1].sparse = true;
  });
};
```

Here's a sample script:

```javascript
const co = require('co');
const GITHUB_ISSUE = `gh-5387`;

const ObjectId = mongoose.Schema.Types.ObjectId;

mongoose.set('debug', true);

exec()
  .then(() => {
    console.log('successfully ran program');
    process.exit(0);
  })
  .catch(error => {
    console.error(`Error: ${error}\n${error.stack}`);
  });


function exec() {
  return co(function* () {
    const db = mongoose.createConnection(`mongodb://localhost:27017/${GITHUB_ISSUE}`)

      name: { type: String, unique: true }
    });
    schema.plugin(sparseIndexPlugin);

    const Model = db.model('ModelA', schema);
    Model.on('index', err => console.log(err));
    yield cb => setTimeout(() => cb(), 1000);
  });
}

function sparseIndexPlugin(schema) {
  schema.indexes().forEach(function(indexArgs) {
    indexArgs[1].sparse = true;
  });
}
```

PR #5397, opened 2017-06-24
bug(query): justOne is actually single, and it default to false

### LINKED ISSUE #5396: Docs for default behavior of Model/Query.remove and justOne option are wrong

Error in docs.

http://mongoosejs.com/docs/api.html#model_Model.remove says it *"Removes the first document that matches conditions from the collection. To remove all documents that match conditions, set the justOne option to false."*

As far as I've tested and read the source code, `Model.remove`/`Query.remove` actually removes *all* matching documents by default, and the `justOne` option does nothing.

The corresponding option in the mongodb driver is called `single` and it defaults to `false` ([docs](http://mongodb.github.io/node-mongodb-native/2.2/api/Collection.html#remove)).

Passing a `single` option to `Model.remove`/`Query.remove` *does* work.

I'm about to submit a PR to change the option name and correct the docs. It's sort of a breaking change to rename the option to `single`, except that it doesn't work currently.

zbjornson:
**Summary**

Fixes #5396

Renames the `justOne` option to `single`, and correct the docs. `justOne` doesn't work.

**Test plan**

All three cases (default, true, false) are covered.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1058,9 +1058,9 @@ Model.base;
 Model.discriminators;
 
 /**
- * Removes the first document that matches `conditions` from the collection.
- * To remove all documents that match `conditions`, set the `justOne` option
- * to false.
+ * Removes all documents that matche `conditions` from the collection.
+ * To remove just the first document that matches `conditions`, set the `single`
+ * option to true.
  *
  * ####Example:
  *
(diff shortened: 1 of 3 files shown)


zbjornson:
I rebased to resolve conflicts but github isn't updating this PR to reflect the latest commit in the branch. Closing to try making a new PR.

zbjornson:
(Apparently GitHub's webhooks are having issues: https://twitter.com/githubstatus/status/879397164673368066)

Issue #5365, opened 2017-06-15
Won't build with rollup.js


krstffr:
Using rollup with rollup-plugin-commonjs will throw the error `Error: Dynamic requires are not currently supported by rollup-plugin-commonjs` when trying to run a bundled app depending on mongoose.

This is the part throwing the error, I guess there's no obvious way around this? https://github.com/Automattic/mongoose/blob/67c465aac5c864c3004d11d49934605037c8f520/lib/drivers/index.js#L8

node v8.1.0
mongoose 4.10.5
rollup 0.42.0
rollup-plugin-commonjs 8.0.2

vkarpov15:
No obvious away around this right now. Removing this dynamic import would be very backwards breaking. Are you trying to use mongoose in the browser?

varunjayaraman:
yeah unfortunately the same problem occurs with webpack, although there's a hack that can be implemented to bypass this in webpack. I wonder if it's worth avoiding the dynamic import in 5.0 @vkarpov15 ?

vkarpov15:
Might be fixed by #5443 . Re-open if there's a similar issue.

Issue #5304, opened 2017-05-28
Use MongoClient.connect() instead of authenticate() if user provides URI to avoid Db.prototype.authenticate deprecated


glecoz:
I'm using a connection method like :

```
var Mongoose = require('mongoose');
connectionUrl = mongodb://some_login:some_pwd@some_host:27017/some_db?authSource=admin
Mongoose.connect(connectionUrl); 
```
... as indicated in the [docs](http://mongoosejs.com/docs/connections.html), but receiving the following **warning** on NodeJS console :

```
Db.prototype.authenticate method will no longer be available 
in the next major release 3.x as MongoDB 3.6 will only allow auth 
against users in the admin db and will no longer allow multiple credentials on a socket. 
Please authenticate using MongoClient.connect with auth credentials.

Connection with database succeeded.
```

(I get an authentication failure if I omit `?authSource=admin`)

**Is this normal at this stage (do we need to wait for a next mongoose release ?), 
or what should I  change from now ?**

Versions :

Node : 7.10.0
Mongoose : 4.10.3
MongoDB : 3.4

Thank you.

wellcaffeinated:
+1

monkeymon:
+1

AussieFlem:
+1
Mac OS X 10.12.5
Node: v7.10.0
Mongoose: 4.10.3
MongoDB v3.4.4

cbratschi:
+1

varunjayaraman:
yup looks like it only happens once you actually touch the db to do some operation (like a write):

```javascript
const mongoose = require('mongoose');
const co = require('co');
mongoose.Promise = global.Promise;
const GITHUB_ISSUE = `gh-5304`;


exec()
  .then(() => {
    console.log('successfully ran program');
    process.exit(0);
  })
  .catch(error => {
    console.error(`Error: ${ error }\n${ error.stack }`);
  });


function exec() {
  return co(function*() {
    const db = mongoose.connect(`mongodb://gh_5304:gh_5304@localhost:27017/${ GITHUB_ISSUE }?authSource=admin`);

    const schema = new mongoose.Schema({ name: String });

    const Model = db.model('Model', schema);

    const doc = yield Model.create({ name: 'Test' });
  });
}
```

Log output:

```
Db.prototype.authenticate method will no longer be available in the next major release 3.x
as MongoDB 3.6 will only allow auth against users in the admin db and will no longer allow
[...]

Issue #5229, opened 2017-05-07
Connection#open() promise rejection is broken


daltones:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
When `Connection#open()` fails, it is emitting an error event that turns into an uncaught exception, breaking the promise rejection handling.

**If the current behavior is a bug, please provide the steps to reproduce.**
```js
import mongoose from 'mongoose';

const conn = mongoose.createConnection(); // don't open now

async function start() {
  try {
    await conn.open('some wrong connection string');
    // this promise will throw
    // but before that an error event is emitted, leading to an uncaught exception
  } catch (error) {
    console.log('error on open conn');
  }
}

start();
```

I'm guessing the problem is here: https://github.com/Automattic/mongoose/blob/2dd76ab803b809ab7be2b6881256a6527fd3e11c/lib/connection.js#L316  
That `promise.$hasHandler` is not valid here, once `$hasHandler` is a property of `MongooseThenable` instances (which is not the case there).

This affects `Connection#openSet()` too.

**What is the expected behavior?**
While open returns a promise, I expect that any error appears only as the rejection reason and not thrown somewhere else.

**Please mention your node.js, mongoose and MongoDB version.**
node.js: v7.10.0
mongoose: v4.9.7
MongoDB: v3.4.3

vkarpov15:
As a workaround, add an `on('error')` handler. 

```javascript
const mongoose = require('mongoose');

const conn = mongoose.createConnection(); // don't open now

conn.on('error', () => {}); // Add an error handler to prevent the exception

async function start() {
  try {
    await conn.open('some wrong connection string');
    // this promise will throw
    // but before that an error event is emitted, leading to an uncaught exception
  } catch (error) {
    console.log('error on open conn');
  }
}

start();
```

In general, I recommend not using `open` and `openSet` directly unless you're certain you want to connect to a standalone. Use `createConnection()` with a URI param instead because that'll figure out replica set connections etc.

```javascript
const mongoose = require('mongoose');

async function start() {
  try {
    const conn = await mongoose.createConnection('some wrong connection string');
    // this promise will throw
    // but before that an error event is emitted, leading to an uncaught exception
  } catch (error) {
    console.log('error on open conn');
  }
}

start();
```

vkarpov15:
Fix will be in 4.9.9. Monkey-patching `.then()` hurts but is necessary

Issue #1365, opened 2013-02-28
Index creation fails after creating the first background index


hvrauhal:
Index creation fails with error `MongoError: cannot add index with a background operation in progress` for all but the first index if there are many indexes defined for a schema and the indexes take some time to create.

Run https://github.com/hvrauhal/mongoosebug to reproduce.

This is related to issue #1059

PR #5443, opened 2017-07-05
Load bundled driver first to avoid packaging problem

prototypeme:
**Summary**
See [5365](https://github.com/Automattic/mongoose/issues/5365)
Using expression in require doesn't work while packaging using Webpack. For those who doesn't intend to override the 'node-mongodb-native' that is included with mogoose, need not have to solve this problem of packaging if the default native driver is loaded first and then override a mongodb driver using the path provided in global.MONGOOSE_DRIVER_PATH

**Test plan**
Ran the full test, below is the result. I don't think the failure are related to the change in this PR.

  1814 passing (3m)
  5 failing

  1) connections: useMongoClient/openUri (gh-5304) with mongoose.connect():
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

  2) connections: errors .catch() means error does not get thrown (gh-5229):
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

  3) model create() creates in parallel:
     Error: Timeout of 1000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

  4) model indexes do not trigger "MongoError: cannot add index with a background operation in progress" (gh-1365) LONG:
     Uncaught MongoError: connection 331 to localhost:27017 timed out
      at Function.MongoError.create (node_modules\mongodb-core\lib\error.js:29:11)
      at Socket.<anonymous> (node_modules\mongodb-core\lib\connection\connection.js:188:20)
      at Socket._onTimeout (net.js:342:8)

  5) model indexes auto creation can be disabled:
     Uncaught MongoError: connection 331 to localhost:27017 timed out
      at Function.MongoError.create (node_modules\mongodb-core\lib\error.js:29:11)
      at Socket.<anonymous> (node_modules\mongodb-core\lib\connection\connection.js:188:20)
      at Socket._onTimeout (net.js:342:8)

### DIFF
--- a/lib/drivers/index.js
+++ b/lib/drivers/index.js
@@ -5,7 +5,10 @@
 var driver;
 
 if (typeof window === 'undefined') {
-  driver = require(global.MONGOOSE_DRIVER_PATH || './node-mongodb-native');
+  driver = require('./node-mongodb-native');
+  if (global.MONGOOSE_DRIVER_PATH) {
+    driver = require(global.MONGOOSE_DRIVER_PATH || './node-mongodb-native');
+  }
 } else {
   driver = require('./browser');
 }
--- a/lib/drivers/index.js
+++ b/lib/drivers/index.js
@@ -7,7 +7,7 @@ var driver;
 if (typeof window === 'undefined') {
   driver = require('./node-mongodb-native');
   if (global.MONGOOSE_DRIVER_PATH) {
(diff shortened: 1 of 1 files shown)


vkarpov15:
LGTM, thanks :beers:

PR #5479, opened 2017-07-17
propagate runSettersOnQuery option to child schema

ValYouW:
**Summary**
Currently `runSettersOnQuery` will not run on array with subdocs as this option is not propagated to the child schema, this PR just propagate this option (like it is currently done for the `strict` option)
for example:
```js
var UserSchema = new mongoose.Schema({
	friends: [
		{
			name: {
				type: String,
				set: function(val) {
					var res = '_' + val;
					return res;
				}
			}
		}
	]
}, {collection: 'users', runSettersOnQuery: true});

var UserModel = mongoose.model('User', UserSchema);
// The setter of friends.name will NOT run
UserModel.find({"friends.name": "aaa"}).exec()
```

**Test plan**
Not sure where to add tests, didnt fins something similar

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -596,7 +596,7 @@ Schema.interpretAsType = function(path, obj, options) {
         }
         //propagate 'runSettersOnQuery' option to child schema
         if (options.hasOwnProperty('runSettersOnQuery')) {
-            childSchemaOptions.runSettersOnQuery = options.runSettersOnQuery;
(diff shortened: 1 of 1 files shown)


ValYouW:
I have no idea why it fails on Node v5.. Ideas?

vkarpov15:
Good catch! Don't worry about the test failure on node v5, we have some tests that are flakey on travis. Re-ran and they succeeded.

PR #5502, opened 2017-07-25
Ensure a real parallel execution in asyncEach...

lchenay:
keeping the fact that cursor.next() must be call once at time.

CF: https://github.com/Automattic/mongoose/commit/f6e4dcec3252b2147261a329d94bd6b2154e167a


vkarpov15:
Thanks :beers:

PR #5503, opened 2017-07-25
Fix openuri connection props

cdchase:
**Summary**

After noticing missing connection information while using `openUri()`, I investigated and discovered it did not handle its arguments consistent with `open()` or `openSet()`. Updating adds back the information from the base connection object to make this information available at connection time. This also adds back the ability to have the user & pass information in the options object when using `openUri()`.

I added `Connection.prototype._handleOpenUriArgs` in a similar fashion as `_handleOpenArgs` & `_handleOpenSetArgs`.

**Test plan**

Updated related tests to add additional assertions based on the availability of the additional connection parameters.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -695,6 +695,60 @@ Connection.prototype.onOpen = function(callback) {
   }
 };
 
+/*!
+ * ignore
+ */
+
+Connection.prototype._handleOpenUriArgs = function(uri, options, callback) {
+  if (!rgxProtocol.test(uri)) {
+    uri = 'mongodb://' + uri;
+  }
+
+  var parsed;
+  try {
+    parsed = muri(uri);
+  } catch (err) {
+    this.error(err, callback);
+    throw err;
+  }
+
+  if (!this.name) {
+    this.name = parsed.db;
+  }
+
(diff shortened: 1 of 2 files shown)


cdchase:
Failing builds for v6 & older seem unrelated. But, if you can give me insight to the conflict, I'll look into fixing.

cdchase:
In trying to fix the issues with v6, it seems for some reason a different test is getting executed for middleware, and that is failing.

`at Query.<anonymous> (test/query.middleware.test.js:287:14)`

Based on the [*new* Error handling](http://mongoosejs.com/docs/middleware.html) I believe the assertion for the second argument should be `true`, not `false`.

Can someone with more experience with this code check this?

vkarpov15:
Sorry, but this will absolutely not be supported. `openUri()` is designed to delegate as much as possible to the underlying driver's connection logic, which is much more robust. If there are individual features you need we can consider adding them, but `openUri()` will not support passing in hostname, port, user, etc. as positional arguments.

cdchase:
This was not intended to allow setting these elements, only to *report* them after a connection is made. (I realize setting user/pw via options came along for the ride with this, but was not my goal.)

Is their another way to have connection report its hostname, port & user? This was the previous function I was really looking to enable.

PR #5516, opened 2017-07-29
fix wrong reference link to $lookup by removing brackets

TalhaAwan:
screenshot from 2017-07-29 16-21-14

The closing bracket becomes part of the url and takes to "File not found" page. Removed them, as they look unnecessary in the statement anyway.


vkarpov15:
Thanks :beers:

PR #5529, opened 2017-08-04
fix: limiting number of async operations per time in insertMany

andresattler:
**Summary**
For huge numbers of documents to insert using Model.insertMany produces an ‘CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory’ error.

This PR solves this issue by using parallelLimit() to limit the number of operations per time to 1000. This fits well to the underlying  db.collections.insertMany which itself limit the number of DB operations to 1000 per time.


vkarpov15:
Thanks :beers:

PR #5547, opened 2017-08-13
fix: added type boolean to options documentation

ndabAP:
**Summary**

When setting `mongoose.set('debug', true)`, the IDE indicates wrong argument type.

**Test plan**

Doesn't change any code.


vkarpov15:
Thanks :beers:

PR #5578, opened 2017-08-24
Use story schema defined key author instead of creator.

dmric:
**Summary**

Found it confusing to read creator instead of author as defined in story schema. This pull request will improve docs.
**Test plan**


vkarpov15:
Good catch, we updated these docs recently and missed a spot. Thanks :+1: :beers:

PR #5642, opened 2017-09-13
fixed plugins link

renanrboliveira:
Correct the plugins links on some pages.


vkarpov15:
I think mongoosejs.io is the correct link, just that link is currently down :(

renanol:
What's the problem mongoosejs.io link is down? I thought the link was removed.

PR #5644, opened 2017-09-15
Asterisk should not render as markdown bullet

timkinnane:
Minor issue with readability in the docs, there's a follow up clarificaiton to the "almost*" clause on `findOne`, but the leading asterisk was being converted to a dot point. Took my a while to find what that "almost" refered to.

**Summary**

**Test plan**


vkarpov15:
Nice, good catch, thanks! :beers:

PR #5659, opened 2017-09-22
Broken link for 'plugin-search'. Changed .io -> .com

awareness481:
Changed http://plugins.mongoosejs.io/ to http://plugins.mongoosejs.com/

The link with the **.io** is unreachable while the **.com** link works as intended


vkarpov15:
The .io link is the correct one, just currently broken. The issue is fixing the .io link, not changing to the .com link.

PR #5665, opened 2017-09-24
fix replaceOne with validators

sime1:
**Summary**
calling model.replaceOne() with the option runValidator = true generates an error ("multi update only works with $ operators"), as when the validation is successful query.updateMany()  is called instead of query.replaceOne().


vkarpov15:
Good catch, thanks :+1:

PR #5675, opened 2017-09-28
Update mquery dependency

jonathanprl:
mquery 2.3.1 uses debug 2.6.8 which has the following security advisory: https://nodesecurity.io/advisories/534

mquery 2.3.2 uses debug 2.6.9, which will resolve the security issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "mongodb": "2.2.31",
     "mpath": "0.3.0",
     "mpromise": "0.5.5",
-    "mquery": "2.3.1",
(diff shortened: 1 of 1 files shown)


hairyhenderson:
@jonathanprl the `package-lock.json` will probably need to be updated as well

Looks like the CI failures are for Node 0.10 and 4

vkarpov15:
Thanks :beers: I'll update the package-lock.json before deploying

jonathanprl:
Thank you!

PR #5676, opened 2017-09-28
Fix plugins link

strathausen:
**Summary**

**Test plan**


vkarpov15:
Current link is correct, just site has been down for a while

Issue #5677, opened 2017-09-28
mquery@2.3.1 depends on the vulnerable package debug@2.6.8


subesokun:
As `mquery@2.3.1` explicitly depends on the vulnerable package `debug@2.6.8` the NSP security check fails, see https://nodesecurity.io/advisories/534

Please upgrade at least to `mquery@^2.3.2` and if possible avoid pinning dependencies.

vkarpov15:
Fixed in #5679 :+1:

PR #5679, opened 2017-09-29
Adding nsp check to the CI build

hairyhenderson:
**Summary**

Currently there are no checks being done during CI to ensure dependencies are patched for known vulnerabilities. Since mongoose is such a widely-used project, it'd be useful for known vulnerabilities (as reported by tools like nsp or Snyk) to be highlighted quickly.

See #5675 and #5677 for one example of where it would have been useful to have a check like this. 

_An alternative_ (perhaps a better one) would be to integrate nsp's service (see https://nodesecurity.io/#pricing) as well as Snyk - https://snyk.io/

**Test plan**

The CI should pass (though currently it does not due to #5675 / #5677)

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "mongoose-long": "0.1.1",
     "mongodb-topology-manager": "1.0.11",
     "node-static": "0.7.7",
-    "nsp": "^2.8.1",
+    "nsp": "~2.8.1",
     "power-assert": "1.4.1",
     "q": "1.4.1",
     "tbd": "0.6.4",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "mongoose-long": "0.1.1",
     "mongodb-topology-manager": "1.0.11",
     "node-static": "0.7.7",
+    "nsp": "^2.8.1",
     "power-assert": "1.4.1",
     "q": "1.4.1",
     "tbd": "0.6.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -50,6 +50,7 @@
     "mongoose-long": "0.1.1",
     "mongodb-topology-manager": "1.0.11",
     "node-static": "0.7.7",
+    "nsp": "^2.8.1",
```
vkarpov15: Nit: we generally don't allow ranges in package.json, and definitely don't allow `^`. `~` is ok in this particular case.

hairyhenderson: sorry - this was the default from `npm install`...

IMO `^` should be fine for a tool like this, especially since it's a `devDependency`, but it's your call! 😉

vkarpov15:
Thanks for adding this, this is excellent.

hairyhenderson:
you're welcome! 😁

vkarpov15:
@hairyhenderson can you explain this failure? https://travis-ci.org/Automattic/mongoose/jobs/281422013 looks like nsp still thinks mongoose is using the old version of mquery even though we bumped it?

hairyhenderson:
```
│ Path          │ mongoose@4.11.14-pre > mquery@2.3.1 > debug@2.6.8    
```

@vkarpov15 that's really odd - I can see from the `npm install` log that `mquery@2.3.2` and `debug@2.6.9` are getting installed... I would have thought maybe the `package-lock.json` is to blame, but maybe some other caching?

What would probably be even better would be to use https://nodesecurity.io's GitHub integration - it's free for public repos, and runs the same tests as the `nsp` tool. And https://snyk.io is also a good idea - mostly the two tools overlap, but one may catch a vulnerability that the other doesn't...

vkarpov15:
Yeah it was a `package-lock.json` issue looks like. I added the check back to travis.yml :+1:

paton:
Awesome stuff. We use the https://nodesecurity.io Github integration (probably the same as manually running it). NSP is great

Issue #5337, opened 2017-06-08
insertMany multiple validation failures


Mattchewone:
**Issue/Comment**
Not sure if this is a bug or expected behaviour. 

I am using insertMany and when I have a validation error(s) only the first error is returned and this throws an error.

For example:
```
User.insertMany(users)
.then(users => {
	console.log(`Results: ${users.length}`);
})
.catch(err => {
	console.log(`Error: ${err}`);
});
```
If I have 5 users, and 2 of them have failed due to validations, then the `catch` is hit, but it will only show the first validation error, rather than a list of validations. 

**Expected behavior?**
A list of validations to be returned if there are more than one error. 

I would also think that it would be better to return the results and the errors much like Mongodb does natively, where it lists the inserted ids and any failures. Not sure if that would be a breaking change to the current API, but it seems more of how the mongodb driver works natively. 

**Versions**
Node = v8.0.0
Mongoose = 4.10.5

varunjayaraman:
ill label this an enhancement for now.

```javascript
const mongoose = require('mongoose');
const co = require('co');
mongoose.Promise = global.Promise;
const GITHUB_ISSUE = `gh-5337`;


exec()
  .then(() => {
    console.log('successfully ran program');
    process.exit(0);
  })
  .catch(error => {
    console.error(`Error: ${error}\n${error.stack}`);
  });


function exec() {
  return co(function* () {
    const db = mongoose.createConnection(`mongodb://localhost:27017/${GITHUB_ISSUE}`)
    const schema = new mongoose.Schema({
      name: { type: String, required: true },
      age: Number
    });


    const Model = db.model('Model', schema);

    yield Model.remove({});

    try {
      const op = yield Model.insertMany([
        { name: 'test1', age: 1 },
        { age: 2 },
        { name: 'test3', age: 3 },
        { age: 4 }
      ]);

    } catch (error) {
      console.log(error);
    }

  });
}
```

NineToeNerd:
I am having a related issue. I expect to get the inserted objects back in the result even when there are validation errors, but I only get the first validation error - _no result array_. In my case, my app may send duplicate objects to the server and there will be a validation error with `"code":11000`. When this happens I'm not getting an array with the objects that were successfully saved (I get `undefined`).

[...]

PR #5698, opened 2017-10-07
insertMany() multiple validation errors (#5337)

vichle:
**Summary**
See #5337.
````javascript
let Movie = new Schema({name: {type: String, required: true}})
Movie.insertMany([{ foo: 'bar' }, {foo: 'baz'}])
.catch((err) => {
  console.log(Object.keys(err.errors).length) // 2
})
````
**Test plan**

Added test 'insertMany() multiple errors (gh-5337)'.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -13,6 +13,7 @@ var PromiseProvider = require('./promise_provider');
 var Query = require('./query');
 var Schema = require('./schema');
 var VersionError = require('./error').VersionError;
+var ValidationError = require('./error').ValidationError;
 var applyHooks = require('./services/model/applyHooks');
 var applyMethods = require('./services/model/applyMethods');
 var applyStatics = require('./services/model/applyStatics');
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/model.js [hunk 1]
```diff
@@ -2108,29 +2109,33 @@ Model.insertMany = function(arr, options, callback) {
...
 
   parallelLimit(toExecute, limit, function(error, docs) {
-    if (error) {
-      callback && callback(error);
+    // `error` should never be anything but null.
+
+    // Option `ordered: false` signals that insert should be continued after reaching
+    // a failing insert. Return grouped ValidationError if this option doesn't exist
+    // or is true (if there are any).
+    if ((options == null || typeof options !== 'object' || options['ordered'] === true) && Object.keys(validationError.errors).length > 0) {
```
vkarpov15: `ordered` is true by default: https://docs.mongodb.com/manual/reference/method/db.collection.insertMany/#db.collection.insertMany so we shouldn't return multiple validation errors if `ordered` isn't set

lib/model.js
```diff
@@ -2108,29 +2109,33 @@ Model.insertMany = function(arr, options, callback) {
...
-          // a failing insert. Therefore we delegate "null", meaning the validation
-          // failed. It's up to the next function to filter out all failed models
-          if (options != null && typeof options === 'object' && options['ordered'] === false) {
-            return callback(null, null);
-          }
-          return callback(error);
+          validationError.errors[index] = error;
+          // Never return the error, as it will cause parallelLimit to immediately
+          // abort. Return null doc for `ordered: false` option.
+          return callback(null, null);
```
vkarpov15: If `ordered` isn't set to `true` we still want it to abort and only give a single validation error.

PR #5707, opened 2017-10-12
Fix schema methods didn't add into properties

jhen0409:
**Summary**

Our builds are broken since upgrade to `^4.12.0`, it looks like schema methods didn't add into properties and nested properties.

**Test plan**

In the following test case, it works fine in `<= 4.11.14`:

```js
var PropMethodSchema = new Schema({
  nested: {
    a: { type: String }
  }
});
PropMethodSchema.methods.func = function() {};
var PropMethod = db.model('propertiesMethod', PropMethodSchema);
var doc = new PropMethod;
doc.init({nested: {a: 'test'}});
assert.strictEqual(doc.nested.func, doc.func);
```

### DIFF
--- a/lib/services/document/compile.js
+++ b/lib/services/document/compile.js
@@ -77,7 +77,7 @@ function defineKey(prop, subprops, prototype, prefix, keys, options) {
         }
 
         if (!this.$__.getters[path]) {
-          var nested = Object.create(Document.prototype, getOwnPropertyDescriptors(this));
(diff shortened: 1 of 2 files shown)


jhen0409:
Okay I missed #5546, sorry for noise.

vkarpov15:
Sorry for the trouble. The whole `doc.nested.func` bit was always a grey area with mongoose, it worked but was an unintentional side effect that was never really documented. Are you ok with changing your code to work around this change?

PR #5773, opened 2017-11-01
Add Pokémon to list of uncountables

hugomd:
This PR adds Pokémon to the list of uncountables.

The plural of Pokémon is [not Pokémons](https://en.wikipedia.org/wiki/Portal:Pok%C3%A9mon). 

I've personally encountered this issue when using Mongoose on private projects, although I am aware of the `pluralize` option, I think Pokémon is used widely enough that it should be added as a default exception.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -102,7 +102,8 @@ exports.uncountables = [
   'news',
   'expertise',
   'status',
-  'media'
+  'media',
+  'pokemon'
 ];
 var uncountables = exports.uncountables;
 


vkarpov15:
This is a great PR and I really appreciate it, but unfortunately this would be a very frustrating backwards breaking change for anyone that currently has a 'Pokemon' model and is relying on the current pluralization. We're strongly considering just removing pluralization entirely in the next backwards breaking release, so hopefully this change won't be necessary.

I am curious though, do you use 'Pokemon' or 'Pokémon' as your model name?

hugomd:
@vkarpov15 Thanks for taking the time to respond 🙏 I hadn't thought about the fact that it might be a breaking change, that is a very good point! 

I use 'Pokemon' as my model name, I'd rather not play with Unicode diacritics 😂

✌️

vkarpov15:
That's smart. We've been discussing just removing pluralization entirely and making the collection name match the model name by default, so `mongoose.model('Pokemon', pokemonSchema)` will use the 'Pokemon' collection rather than the 'pokemon' or 'pokemons' collection by default. The distinction between model name and collection name is confusing and not terribly useful, so we've considered switching. Any thoughts on that?

hugomd:
@vkarpov15 I agree that there shouldn't be any distinction. The model name should be used as the collection name. Conceptually, it might make sense to pluralise, e.g. a collection is really a _group_ of documents, so a `Users` collection makes sense, despite the model being `User`, however, pluralisation of more complex words becomes a hassle to maintain (as I'm sure you know), and adds unnecessary complexity. 

I agree with everything listed in this [SO thread](https://stackoverflow.com/a/5841297) (even though it's for SQL, the same reasoning applies). 

If someone really wants pluralised names they can probably write middleware for it, or just manually define the collection name anyway! 🤷‍♂️

vkarpov15:
Thanks for the insight, I appreciate it :+1: :rocket:

PR #5777, opened 2017-11-01
update links in docs to correct Plugins URL

staylor:


vkarpov15:
The links are correct, its just the plugins site is down and we haven't fixed it.

PR #5818, opened 2017-11-14
Travis Add a check for node js 8

superheri:
**Summary**
Adding a check in Travis for Node Js 8 since it is now the recommended version on Node JS download page.

**Test plan**


vkarpov15:
Thanks :beers:

superheri:
@vkarpov15 Should we add a not required test (```allow_failures```) for Node 9?

Issue #5706, opened 2017-10-11
Middleware is not invoked on embedded discriminators


asabhaney:
I would like to report what I believe to be a bug in mongoose `4.12.1`.

It seems that when using embedded discriminators (either in arrays or a subdoc), middleware functions that are defined on the discriminator schema are not invoked. I have provided a test to illustrate this:

```js
const assert = require('chai').assert
const sinon = require('sinon')
const mongoose = require('mongoose')

const Schema = mongoose.Schema

describe('Mongoose Discriminators', () => {
   it('Middleware should succesfully be invoked on base and discriminators', async () => {
      const eventSchema = new Schema(
         { message: String },
         { discriminatorKey: 'kind', _id: false }
      )

      const trackSchema = new Schema({
         event: eventSchema
      })

      const clickedSchema = new Schema({
         element: String
      }, { _id: false })

      // This is the discriminator which we will use to test middleware

      const purchasedSchema = new Schema({
         product: String
      }, { _id: false })

      // Setup some spies to test whether the middleware functions execute

      const eventPreValidateSpy = sinon.spy()
      const eventPostValidateSpy = sinon.spy()
      const eventPreSaveSpy = sinon.spy()
      const eventPostSaveSpy = sinon.spy()

      eventSchema.pre('validate', function (next) {
         eventPreValidateSpy()
         next()
      })

      eventSchema.post('validate', function (doc) {
         eventPostValidateSpy()
      })

      eventSchema.pre('save', function (next) {
         eventPreSaveSpy()
         next()
      })

      eventSchema.post('save', function (doc) {
         eventPostSaveSpy()
      })

      const purchasedPreValidateSpy = sinon.spy()
      const purchasedPostValidateSpy = sinon.spy()
      const purchasedPreSaveSpy = sinon.spy()
      const purchasedPostSaveSpy = sinon.spy()

      purchasedSchema.pre('validate', function (next) {
         purchasedPreValidateSpy()
         next()
      })

      purchasedSchema.post('validate', function (doc) {
         purchasedPostValidateSpy()
      })

      purchasedSchema.pre('save', function (next) {
         purchasedPreSaveSpy()
         next()
      })

      purchasedSchema.post('save', function (doc) {
         purchasedPostSaveSpy()
      })

      // Register the discriminators

      trackSchema.path('event').discriminator('Clicked', clickedSchema)
      trackSchema.path('event').discriminator('Purchased', purchasedSchema)

      // Test

[...]

PR #5821, opened 2017-11-15
Fixed call stack overflow when using recursive embedded discriminators

Faibk:
After the recent changes to applyHooks() in pursue of  #5706  embedded discriminators recursively referencing themselves led to a call stack overflow. The problem was, that the childModel.discriminators tree is descended into before the hooks are applied on the actual discriminator model. That leads to the $appliedHooks attribute, which is to prevent these circular calls, never being set.

I swapped the order of applying the hooks to model first, discriminators second. To be sure there are no implications (because I am new to mongoose), I adapted the embedded discriminators test and wrote a test for these self-referencing/recursive discriminators.

### DIFF
--- a/lib/services/model/applyHooks.js
+++ b/lib/services/model/applyHooks.js
@@ -28,6 +28,7 @@ function applyHooks(model, schema) {
     if (childModel.$appliedHooks) {
       continue;
     }
+    applyHooks(childModel, schema.childSchemas[i].schema);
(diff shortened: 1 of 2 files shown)


vkarpov15:
Thanks for taking the time to dig into this and write a test, this is great work. I really appreciate your help :+1: :rocket: :guitar:

PR #5822, opened 2017-11-16
Add node 9 in travis allowing failures

superheri:
**Summary**
Adding a check in Travis for Node Js 8 since it is now downloadable on Node JS download page.

I allowed it to fail because this version of node is still in developpement. It can help to see if there are possible bugs with the next version of Node.

**Test plan**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 sudo: false
 node_js:
+  - "9"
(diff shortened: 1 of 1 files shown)


vkarpov15:
Works for me, although I'll probably remove the 'allow failures' part. No reason why mongoose shouldn't pass on node 9.

PR #5834, opened 2017-11-18
fix(model): allow virtual ref function to return arrays

brunohcastro:
Allow the virtual ref function to return an array of model names.

When referencing fields inside an array of subdocuments, the ref function was expecting a single string as the model name. This was very limiting when using dynamic references on an array of subdocuments.

To solve this I just did a simple type check. If the return value of the ref function is already an array, there's no need to change it. If the result is anything else the code resume with the previous behaviour. This enabled me to use something as:

```Javascript
const appointmentServiceSchema = new Schema({
  serviceId:    Schema.Types.ObjectId,
  serviceType:  String,
  providerId:   Schema.Types.ObjectId,
  providerType: String,
  notes:        String,
  cost:         Number
}, { toJSON: { virtuals: true } });

const appointmentSchema = new Schema({
  services:         [appointmentServiceSchema]
}, { toJSON: { virtuals: true } });

appointmentServiceSchema.virtual('service', {
  ref() {
    return this.services.map((item) => item.serviceType);
  },
  localField:   'serviceId',
  foreignField: '_id',
  justOne:      true
});

appointmentServiceSchema.virtual('provider', {
  ref() {
    return this.services.map((item) => item.providerType);
  },
  localField:   'providerId',
  foreignField: '_id',
  justOne:      true
});
```

Sorry but I didn't have time to write the tests yet. Just did the quick fix and used right away.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3581,7 +3581,7 @@ function getModelsMapForPopulate(model, docs, options) {
             ref = ref.call(doc, doc);
           }
 
-          // When referencing nested arrays, the ref should be an Array 
+          // When referencing nested arrays, the ref should be an Array
           // of modelNames.
           if (Array.isArray(ref)) {
             modelNames = ref;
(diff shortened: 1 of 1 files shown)


vkarpov15:
Nice, thanks :+1: :beers:

vkarpov15:
Oh woops why did this go into the 4.13 branch?

PR #5853, opened 2017-11-25
Travis update mongodb version

superheri:
**Summary**
Updating the mongoDB version that Travis uses to do his tests with mongoose. Using 3.4.10, the actual MongoDB version.

**Test plan**

**Note**
The current commits can be squashed into one if needed.


vkarpov15:
Thanks but we're sticking with 2.6 for testing on mongoose 4.x. I'd be open to adding testing 3.4 in addition to 2.6 but not as a replacement right now.

Issue #5737, opened 2017-10-23
.slice() before .populate() causes data loss (with test)


aforty:
**What is the current behavior?**
If .slice() is called before .populate() then data of the original object is lost.

**Reproduce**
Add this test to `test/model.populate.test.js` and run mocha. I took the test "populating an array of refs and fetching many" and added the line `assert.equal(blogposts[0].title, 'Woot');`

```
  it('populating an array of refs, slicing, and fetching many', function(done) {
    var db = start(),
        BlogPost = db.model('RefBlogPost', posts),
        User = db.model('RefUser', users);

    User.create({
      name: 'Fan 1',
      email: 'fan1@learnboost.com'
    }, function(err, fan1) {
      assert.ifError(err);

      User.create({
        name: 'Fan 2',
        email: 'fan2@learnboost.com'
      }, function(err, fan2) {
        assert.ifError(err);

        BlogPost.create({
          title: 'Woot',
          fans: [fan1, fan2]
        }, function(err, post1) {
          assert.ifError(err);

          BlogPost.create({
            title: 'Woot',
            fans: [fan2, fan1]
          }, function(err, post2) {
            assert.ifError(err);

            BlogPost
            .find({_id: {$in: [post1._id, post2._id]}})
            .slice('fans', [0, 5])
            .populate('fans')
            .exec(function(err, blogposts) {
              db.close();
              assert.ifError(err);

              assert.equal(blogposts[0].title, 'Woot');

              assert.equal(blogposts[0].fans[0].name, 'Fan 1');
              assert.equal(blogposts[0].fans[0].email, 'fan1@learnboost.com');
              assert.equal(blogposts[0].fans[1].name, 'Fan 2');
              assert.equal(blogposts[0].fans[1].email, 'fan2@learnboost.com');

              assert.equal(blogposts[1].fans[0].name, 'Fan 2');
              assert.equal(blogposts[1].fans[0].email, 'fan2@learnboost.com');
              assert.equal(blogposts[1].fans[1].name, 'Fan 1');
              assert.equal(blogposts[1].fans[1].email, 'fan1@learnboost.com');
              done();
            });
          });
        });
      });
    });
  });
```

**What is the expected behavior?**
The above test should pass.

**Please mention your node.js, mongoose and MongoDB version.**
node 8.3.0
mongod 3.2.11
mongoose 4.12.4

vkarpov15:
Thanks for the detailed repro instructions, will investigate ASAP :+1:

[...]

PR #5856, opened 2017-11-29
Added strictBool option to schema

ekulabuhov:
**Summary**
Same as #5344 but on Schema level.

**Test plan**
Include a test.
npm test -- -g 'strictBool': test passed

### DIFF
--- a/lib/schema/boolean.js
+++ b/lib/schema/boolean.js
@@ -52,6 +52,7 @@ SchemaBoolean.prototype.checkRequired = function(value) {
  * Casts to boolean
  *
  * @param {Object} value
+ * @param {Object} model - this value is optional
  * @api private
  */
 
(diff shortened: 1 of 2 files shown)


lib/schema/boolean.js
```diff
@@ -55,12 +55,21 @@ SchemaBoolean.prototype.checkRequired = function(value) {
...
  */
 
-SchemaBoolean.prototype.cast = function(value) {
+SchemaBoolean.prototype.cast = function(value, model) {
   if (value === null) {
     return value;
   }
 
-  if (!this.options.strictBool) {
+  if (this.options.strictBool || (model.schema.options.strictBool && this.options.strictBool !== false)) {
```
vkarpov15: `model` might not be defined if we're casting for a query, which is why the tests are failing on travis. Good work thus far but to get this merged we need to figure out what to do with that case so the tests succeed.

ekulabuhov:
I've looked at the failing tests and it looks like they are unrelated to the change. Tried running first one locally (gh-5737) and it fails randomly. Maybe a race condition?

PR #5865, opened 2017-12-01
support circular json

lopugit:
**Summary**
This pull request implements support for circular json in the cloning functions and some util and testing functions

I was getting stack overflow errors when trying to save a circular json document

**Test plan**
This is my first pull request ever... I'm not too good at testing, but I've been testing it with circular json documents and it's been working great.

You can run json such as
```javascript
parent = {
    hey: 'hey'
}
child = {
    woo: 'woo',
    parent
}
parent.child = child

child.save() // will work
```

### DIFF
--- a/lib/querystream.js
+++ b/lib/querystream.js
@@ -4,6 +4,7 @@
  * Module dependencies.
  */
 
+var CircularJSON = require('circular-json');
 var Stream = require('stream').Stream;
 var utils = require('./utils');
 var helpers = require('./queryhelpers');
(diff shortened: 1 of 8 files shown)


lib/querystream.js
```diff
@@ -4,6 +4,7 @@
  * Module dependencies.
  */
 
+var CircularJSON = require('circular-json');
```
markstos: Why is CircularJSON imported here? It is never used in this file.

markstos:
It's hard to to recommend merging this pull request. 

 * It imports CircularJSON into many files where it is never used.
 * There is no test coverage
 * It would seem to break backward compatibility, because CircularJSON is not normal JSON.
 * No documentation updates were included to explain how the behavior would change.

lib/querystream.js
lopugit: Not sure actually, there's a few files with the addition that don't need it.. not sure why

mhombach:
My point of view is, that the idea of "circular JSON" is wrong from start on.
JSON is a way to represent data in a Text-Format and not with internal references.
If you need to handle circular objects, one should maybe handlete/delete the circular attributes in another way and then stringify it.
Also, if you ( @lopugit  ) are not supporting this idea/commit/pullrequest anymore (since your last action is months ago), please close this issue so the list of open issues stays up to date :)

vkarpov15:
Yeah I'm going to close this PR. I left it open to think about it more but it has just gone stale. Storing a circular object in MongoDB just won't work in general. Thanks for your patience with this one.

PR #5871, opened 2017-12-05
fix(Mongoose): fix connection setter

jinasonlin:
**Summary**
This pr will fix and optimize default connection setter, ref connection models with default models (Mongoose.modelNames).

**Test plan**
```javascript
mongoose.connection = 'not connection instance'; // it is not suitable 
```


vkarpov15:
Thanks :beers:

Issue #5874, opened 2017-12-06
Promise.reject() without an arg should be treated as a Promise rejection


ZacharyRSmith:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

eachAsync does not recognize `fn` calling `reject()` without an arg as a Promise rejection.

**If the current behavior is a bug, please provide the steps to reproduce.**

```
const cursor = Model.find({}).cursor();
const fn = new Promise((resolve, reject) => reject());
cursor.eachAsync(fn, err => {
  console.log('err', err); // undefined
});
```

**What is the expected behavior?**

Calling `reject()` should be treated as a Promise rejection.

**Please mention your node.js, mongoose and MongoDB version.**

node.js: >=6.11.4
mongoose: 4.13.7-pre
MongoDB: 3.2.7

vkarpov15:
Fixed by #5875

PR #5875, opened 2017-12-06
fix: eachAsync - Treat reject() as a Promise rejection (gh-5874)

ZacharyRSmith:
https://github.com/Automattic/mongoose/issues/5874


vkarpov15:
Good catch, thanks :beers:

PR #5878, opened 2017-12-06
BREAKING CHANGE: remove pluralization, instead add `mongoose.pluralize()` so plugins can attach their own

### LINKED ISSUE #5877: Remove pluralization, make model name = collection name by default
Re: #3489, etc. Should also add a `mongoose.pluralize()` function that enables plugins to pluralize the collection name so we can add a plugin for backwards compat.

vkarpov15:
Fix #5877
Re: 3489

@wlingke @niftylettuce @ChristianMurphy would love to hear your thoughts. The plan right now is to create a plugin like `mongoose-legacy-pluralize` that bakes in the old pluralization logic so there's a one-line solution for applications that can't switch over in the short term.


wlingke:
I would support this idea. Reduces confusion for new users + removes another feature. We just need to make sure this is very well documented

lib/index.js
```diff
@@ -245,7 +261,7 @@ Mongoose.prototype.disconnect = function(fn) {
  *     // retrieve the Actor model
  *     var Actor = conn.model('Actor');
  *
- * _When no `collection` argument is passed, Mongoose produces a collection name by passing the model `name` to the [utils.toCollectionName](#utils_exports.toCollectionName) method. This method pluralizes the name. If you don't like this behavior, either pass a collection name or set your schemas collection name option._
+ * _When no `collection` argument is passed, Mongoose uses the model name. If you don't like this behavior, either pass a collection name, use `mongoose.pluralize()`, or set your schemas collection name option._
```
varunjayaraman: nit: looks like the description is wrapped by unnecessary `_`s

lib/index.js
```diff
@@ -227,6 +227,22 @@ Mongoose.prototype.disconnect = function(fn) {
...
+ * @return {Function|null} the current function used to pluralize connection names, `undefined` by default.
+ * @api public
+ */
+
+Mongoose.prototype.pluralize = function(fn) {
+  if (arguments.length > 0) {
+    this._pluralize = fn;
+    return fn;
+  }
+  return this._pluralize;
```
varunjayaraman: the `return fn` is unnecessary because you've already set `this._pluralize` to `fn` so you can just use 1 return at the end.

vkarpov15:
Upon more careful thought, we're going to err on the side of stability and include the mongoose-legacy-pluralize package by default, so no backwards breaking changes. This is something that we just haven't given devs sufficient notice about - we've had deprecation warnings about promises, `useMongoClient`, etc. for quite a while, so affected users have been warned about these changes even if they don't check slack, GitHub, or twitter. If we're going to drop legacy pluralization, I'd like to have a loud deprecation warning for at least a few months so we minimize the likelihood someone will be surprised by this change.

PR #5889, opened 2017-12-10
Update schematypes.jade

gokaygurcan:
**Summary**
There's a syntactical error in one of the examples, not a big deal but better be fixed. 

**Test plan**
Just open [this](http://mongoosejs.com/docs/schematypes.html) page and see if the first example's schema looks fine.

```javascript
var schema = new Schema({
  // ...
  ofObjectId: [Schema.Types.ObjectId],
  ofArrays:   [[]], // <--- here
  ofArrayOfNumbers: [[Number]], // <--- and here
  nested: {
    stuff: { type: String, lowercase: true, trim: true }
  }
})
```


vkarpov15:
Thanks! 🍻

PR #5814, opened 2017-11-13
typo fix on old mongoosejs.io link

steffes:
**Summary**
typo fix

**Test plan**

PR #5901, opened 2017-12-12
fix: link to plugins

cklanac:
http://plugins.mongoosejs.io/ is non-responsive
http://plugins.mongoosejs.com/ works fine

**Summary**
Broken link to the plugins site on mongoosejs.com homepage

**Test plan**
Simply clicking the link seems to do the job


vkarpov15:
.io is the correct link, this is a duplicate of #5814, #5777, #5907, etc. etc.

PR #5907, opened 2017-12-14
fix references to old domain (mongoosejs.io) with new domain.

naterkane:
the current website contains references to the old domain, mongoosejs.io, leaving users with broken links.

links on the website to the plugins section were broken and pointed to the wrong URL.

n/a, no code, just changes to body copy/links where necessary.

PR #5926, opened 2017-12-20
Fix plugin search site links

remarks:
**Summary**

The plugin search site currently links to http://plugins.mongoosejs.io, however the site is down. Changed links to http://plugins.mongoosejs.com.


vkarpov15:
Next time please look for duplicate PRs, there probably have been at least 10 thus far: #5907, #5901, #5814, etc. The mongoosejs.io link is correct, its just been down for a while and we haven't had the time to fix it because we've been busy working on 5.0. Thanks for taking the time to put in a PR though :+1:

PR #5952, opened 2018-01-01
ci: add AppVeyor as Windows continuous integration

ChristianMurphy:
**Summary**

Automatically test changes on Windows (AppVeyor) as well as Linux (Travis CI).

https://www.appveyor.com

This may or may not be worth while for Mongoose. :man_shrugging: 

:notebook: AppVeyor would need to be enabled in the Automatic organization and for Mongoose.

**Test plan**

example build, testing this branch https://ci.appveyor.com/project/ChristianMurphy/mongoose/build/1.0.1

**Caveats**

* Only one AppVeyor build can run at a time on the free plan, so having a large matrix will be slow.
* AppVeyor can be a bit flaky at times, sometimes build will fail for no reason.
* Currently many of the tests are timing out of AppVeyor, and individual builds are taking 15 minutes each.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,22 @@
+environment:
+  matrix:
+    - nodejs_version: "9"
+    - nodejs_version: "8"
+    - nodejs_version: "7"
+    - nodejs_version: "6"
+    - nodejs_version: "5"
+    - nodejs_version: "4"
+
+services:
+  - mongodb
+
+install:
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks but I'm not sure how useful this is for us, I haven't seen a windows-specific mongoose bug ever and we have enough flakiness with travis.

PR #5985, opened 2018-01-11
re: refactoring tests to start fewer connections

fenanquin:
**Summary**

The main change was in reducing the number of `db.start()` calls on tests to start fewer connections. It was considered that each test file is a high-level test suite and one connection can be opened and shared through each test case that makes part of it. There are exceptions, depending on the semantics of each test case. But with this change the number of calls to `db.start()` dropped from 668 to 96.

**Test plan**

Executing `npm test` should open fewer connections.

### DIFF
--- a/test/aggregate.test.js
+++ b/test/aggregate.test.js
@@ -26,7 +26,7 @@ var EmployeeSchema = new Schema({
 
 mongoose.model('Employee', EmployeeSchema);
 
-function setupData(callback) {
+function setupData(db, callback) {
   var saved = 0;
   var emps = [
     { name: 'Alice', sal: 18000, dept: 'sales', customers: ['Eve', 'Fred'] },
@@ -34,7 +34,6 @@ function setupData(callback) {
     { name: 'Carol', sal: 14000, dept: 'r&d', reportsTo: 'Bob' },
     { name: 'Dave', sal: 14500, dept: 'r&d', reportsTo: 'Carol' }
   ];
-  var db = start();
   var Employee = db.model('Employee');
 
   emps.forEach(function(data) {
@@ -42,7 +41,7 @@ function setupData(callback) {
 
     emp.save(function() {
       if (++saved === emps.length) {
-        callback(db);
+        callback();
       }
     });
   });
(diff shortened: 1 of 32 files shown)


vkarpov15:
Excellent work, thank you so much! Let me know if you're ever in the SF area, I'll buy you a beer :beers:

fenanquin:
Thank you for letting me do that! I really enjoy refactoring and testing. Ok then :+1: :+1: :+1: , and I'll buy you a great drink of my country: the caipirinha!

Even though the number of connections was reduced, the duration of the tests remains the same on both storage engines :confused:. So, I think that each test with long running time has to be deep analyzed to understand this particular behavior, what do you think about it?

vkarpov15:
We have some tests that are very slow because they stop and start mongodb. Perhaps we should have an option to skip those? They only test reconnection logic

fenanquin:
The option to skip them is a nice approach, specially during development.

Also, I've been studying the [connection events](https://github.com/Automattic/mongoose/blob/1cd36a822cc07416752af7bfdb03e27722cecfe5/test/connection.test.js#L78-L191) tests and the issues related to them, and I believe that mocking the objects from `node-mongodb-native` in [`openUri` method](https://github.com/Automattic/mongoose/blob/master/lib/connection.js#L280) with [sinon](http://sinonjs.org/) would provide the ability to emit any event at any time on tests, then there would be no need to start and stop mongod programatically. I could work on a gist to show you how this approach would be. (I didn't do that at first to avoid working on something that conceptually had a chance of not being accepted)

fenanquin:
Understood.

Then, I think a flag would be the best option (maybe `--skip-connection-event-test` ?). Also, I think that [connection events](https://github.com/Automattic/mongoose/blob/master/test/connection.test.js#L75-L271) are so important and intrinsic that they should be isolated in a separate file (maybe `connection.event.test.js`), because it would be easier to track and maintain them. And, maybe these pieces of code: [hooks](https://github.com/Automattic/mongoose/blob/master/test/common.js#L146-L154) and [Server](https://github.com/Automattic/mongoose/blob/master/test/common.js#L161-L165) (without exporting it, because it is only used once) could be moved to that new file (`connection.event.test.js`) to group them all in one single file. What do you think about the described approach?

PR #6024, opened 2018-01-21
Do not overwrite default values in schema when nested timestamps are provided.

cdeveas:
**Summary**

Sometime after mongoose ~4.6, schemas created with the `timestamps` began overwriting schema defaults in nested properties. For example, schema `{ name: String, misc: { isHappy: { type: Boolean, default: true } }` defined with timestamp `misc.updatedAt` would not instantiate new models with `misc.isHappy` defined.

**Test plan**

Unit test "Declaring defaults in your schema with timestamps defined" defined in `defaults.test.js` captures the logic executing successfully.

Other unit tests pass with this change in place (at least locally for me).

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -730,22 +730,18 @@ Schema.prototype.hasMixedParent = function(path) {
  */
 Schema.prototype.setupTimestamp = function(timestamps) {
   if (timestamps) {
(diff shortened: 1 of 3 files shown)


vkarpov15:
Thanks :beers: will backport this to 4.x as well.

Issue #3302, opened 2015-08-24
useDb() connection caching


vkarpov15:
Make it so that calling `conn.useDb('abc');` twice will give you back the same connection rather than creating a new connection each time.

xpepermint:
+10 :), this means I can clean up my code (deleting my own caching function)

kheide:
Whats the status on this? I am fiddling with my own multi-tenant implementation and I am not doing any caching and my node instance seems to go fubar after 10-15 requests suddenly stopping to respond to requests until I restart node. I am thinking it might be a connection issue.

kheide:
Btw @xpepermint would it be possible for your to share your caching function ?

xpepermint:
@kheide after spending 2 days just reading about mongodb benchmarks, I decided not to use multiple databases (I have 10k users, each user has projects ...). The majority of speed tests suggest carefully setting indexes and splitting processing among multiple servers. But yeah... I had a working uncommited code which looked something like this:
```js
class Mongo {
  constructor() {
    this.conn = mongoose.createConnection(...);
    this.dbs = {};
  }

  use(name) {
    if (this.dbs[name]) {
      let db = this.conn.useDb(name);
      db.model('User', userSchema);
      ... register other models
      this.dbs[name] = db;
    }
    return this.dbs[name];
  }
  
  unuse(name) {
    if (this.dbs[name]) {
       this.dbs[name].close();
       delete this.dbs[name];
    }
  }
}
```

rocketspacer:
@vkarpov15 How about I make the otherDbs a POJO and just share them between all connections?
https://github.com/Automattic/mongoose/blob/70b74f95d4439a5a14c757d36cdef660f65f257d/lib/drivers/node-mongodb-native/connection.js#L88-L89


Something like this:
```
this.otherDbs[newConn.name] = newConn;
newConn.otherDbs = this.otherDbs;
```

rocketspacer:
Could add an option to maintain the default behaviour for backward compatibility

vkarpov15:
@rocketspacer sounds reasonable, just add the option to maintain backwards compat

vkarpov15:
There's now an option for this, see [`Connection#useDb()` and the `useCache` option](https://mongoosejs.com/docs/api/connection.html#connection_Connection-useDb)

tomerzcod7:
Hey guys, a question on this subject (I might have a misunderstanding):

[...]

PR #6036, opened 2018-01-24
Cache connection in useDb, replace connection.otherDbs array with a hashmap

rocketspacer:
**Summary**

- Replace ```connection.otherDbs``` array with a hashmap
- Make the ```useDb``` API use it as a cache

See #3302

**Test Plan**

```
it('cache connections to the same db', function() {
    var db = start();
    var db2 = db.useDb('mongoose-test-2');
    var db3 = db.useDb('mongoose-test-2');

    assert.strictEqual(db2, db3);
    db.close();
});
```

### DIFF
--- a/lib/drivers/node-mongodb-native/connection.js
+++ b/lib/drivers/node-mongodb-native/connection.js
@@ -40,7 +40,10 @@ NativeConnection.prototype.__proto__ = MongooseConnection.prototype;
  * @api public
  */
 
-NativeConnection.prototype.useDb = function(name) {
+NativeConnection.prototype.useDb = function(name, options) {
+  // Return immediately if cached
+  if (options && options.useCache && this.relatedDbs[name]) return this.relatedDbs[name];
+
   // we have to manually copy all of the attributes...
   var newConn = new this.constructor();
   newConn.name = name;
@@ -87,6 +90,12 @@ NativeConnection.prototype.useDb = function(name) {
   // push onto the otherDbs stack, this is used when state changes
   this.otherDbs.push(newConn);
   newConn.otherDbs.push(this);
+
+  // push onto the relatedDbs cache, this is used when state changes
+  if (options && options.useCache) {
+    this.relatedDbs[newConn.name] = newConn;
+    newConn.relatedDbs = this.relatedDbs;
+  }
 
   return newConn;
 };
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -59,7 +59,8 @@ function Connection(base) {
   this.pass = null;
   this.name = null;
   this.options = null;
-  this.otherDbs = [];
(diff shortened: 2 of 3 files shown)


rocketspacer:
@vkarpov15 A lot of failed tests on un-related code. Is there anything happening with the tests?

vkarpov15:
@rocketspacer  thanks for the PR but can we use a separate property name for this? I'm aware that otherDbs is not a documented property but I don't want to mess up anyone that might be relying on it because that code hasn't changed in a while

lib/drivers/node-mongodb-native/connection.js
```diff
@@ -84,9 +87,9 @@ NativeConnection.prototype.useDb = function(name) {
 
   newConn.name = name;
 
-  // push onto the otherDbs stack, this is used when state changes
-  this.otherDbs.push(newConn);
-  newConn.otherDbs.push(this);
```
rocketspacer: @vkarpov15 Do you want to keep the original behaviour of the ```otherDbs``` prop

vkarpov15: Yes please, lets do that

lib/drivers/node-mongodb-native/connection.js
```diff
@@ -41,6 +41,9 @@ NativeConnection.prototype.__proto__ = MongooseConnection.prototype;
  */
 
 NativeConnection.prototype.useDb = function(name) {
+  // Return immediately if existed
+  if (this.relatedDbs[name]) return this.relatedDbs[name];
```
vkarpov15: Would you be open to putting this behind a flag? I'm wary because this could very likely break peoples code

vkarpov15: @rocketspacer any updates?

rocketspacer:
Sorry for the delay, I've been quite busy lately.

lib/drivers/node-mongodb-native/connection.js
```diff
@@ -88,6 +91,12 @@ NativeConnection.prototype.useDb = function(name) {
   this.otherDbs.push(newConn);
   newConn.otherDbs.push(this);
 
+  // push onto the relatedDbs stack, this is used when state changes
+  if (options.useCache) {
```
vkarpov15: One more detail that we missed, need to check if `options` is undefined here, that's why [the tests are failing](https://travis-ci.org/Automattic/mongoose/jobs/344515056)

vkarpov15: @rocketspacer you also need to apply your changes from 5560bd0 to this line :+1:

rocketspacer:
@vkarpov15 It's the test I added. I forgot to put the useCache flag in the test. Btw, I squashed the whole PR to 1 commit.

vkarpov15:
Thanks!

PR #6062, opened 2018-01-28
fix url in plugins link

### LINKED ISSUE #6057: Dead Link to plugins page
When clicking the "Discover Plugins" it leads to plugins.mongoose.io which returns a HTTP 408 error. If I navigate to plugins.mongoose.com on my own, however, it works fine. It seems there may have been a typo in the link and it lead to plugins.mongoose.com and not plugins.mongoose.io?

lineus:
this fixes #6057

**Summary**

tested with
make docclean
make gendocs

### DIFF
--- a/docs/plugins.jade
+++ b/docs/plugins.jade
@@ -54,4 +54,4 @@ block content
 
     ### Community!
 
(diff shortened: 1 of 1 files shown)


vkarpov15:
Duplicate of more issues than I can count. plugins.mongoosejs.io is down, we haven't gotten around to fixing it because we've been working on the 5.0.0 release. Thanks for your PR but it isn't necessary.

PR #6072, opened 2018-01-30
Fix typo

adursun:
I fixed typo.

Before: `Appepnds new custom $sample operator(s) to this aggregate pipeline.`
After: `Appends new custom $sample operator(s) to this aggregate pipeline.`


vkarpov15:
Please make this change in `lib/aggregate.js`, `api.html` is compiled code so your changes will be automatically overwritten :+1:

PR #6092, opened 2018-02-05
[optimization]re-use the modifiedPaths list

tarun1793:
In `_getPathsToValidate` the `doc.modifedPaths()` function is called for every subDoc. the result of the function is same, thereby wasting time and computation to calculate it for every subDoc. This becomes a time hog in documents with a lot of `EmbeddedDocuments`

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1103,12 +1103,12 @@ Document.prototype.modifiedPaths = function() {
  * @api public
  */
 
-Document.prototype.isModified = function(paths) {
+Document.prototype.isModified = function(paths, modifiedPaths) {
   if (paths) {
     if (!Array.isArray(paths)) {
       paths = paths.split(' ');
     }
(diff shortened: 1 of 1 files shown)


vkarpov15:
Hmm what makes you think they're all the same? You can definitely modify one subdoc and have it have a different `modifiedPaths()` than another...

tarun1793:
Of course they can, the code change doesn't assumes that the subDocs will _never_ be altered.  The change is very specific to a *single* synchronous execution  of `_getPathsToValidate` method, during which the subDocs don't change. 
Whenever the `_getPathsToValidate` is called again, the `modifiedPaths` list is rebuilt.

vkarpov15:
That's a very good point, I didn't read sufficiently carefully. I understand the PR now, good catch, thanks for your work :+1:

Issue #2596, opened 2015-01-14
Discriminator Key is tied to Model Name


arciisine:
Within mongoose you can only have one Model with a specific name.  This is generally acceptable as you have one collection per Model.  With the discriminator structure, it would be nice to allow multiple subtypes (of different discriminators) to have the same name.

An example of this would be:

Vehicle
  |-- Car
  |-- Van
  |-- Plane

Shape
  |-- Sphere
  |-- Plane
  |-- Cone

If Vehicle and Shape are both the base discriminator type, I would have two collections in mongodb of 'shapes' and 'vehicles'

If I where to run in one file:
  var PlaneModel = VehicleModel.discriminiator('Plane', {...});

And in another file:
  var PlaneModel = ShapeModel.discriminiator('Plane', {...});

I would get an error, because I am unable to have two Models with the same name.  

Is there anyway to solve this with the current codebase or would it require a change?

vkarpov15:
Nope, it'll require a change. The assumption that model names are unique is very core to mongoose: lots of people write code that references models by name, that is, `mongoose.model('Breakfast').find()`. I don't think we'll be supporting discriminator name conflicts any time soon...

zoellner:
@vkarpov15 digging out this older issue as I have a related question:
Would it be possible to implement a backward compatible change to switch to an alternate model naming convention for discriminated models?

I would propose a way to set the model name for discriminators to include the base name with dot notation. So in the example above one would access them by `mongoose.model('Vehicle.Plane')` and `mongoose.model('Shape.Plane')`. That way the model names are still unique but the discriminator keys can be shorter. 

My use case is a graph-like structure where an edge type such as `has` would be useful for many edges. Book has Author, Comment has Creator, etc. 
It would allow me to use shorter discriminator keys which makes a difference for the storage requirements when having millions of edges. 

Seems that a change to the naming convention might also be needed for #2851 
Issues like #4148 could be avoided as well

vkarpov15:
I like the 'Vehicle.Plane' idea at first glance, always found it confusing that discriminators could be accessed via top-level `.model()`. I'm going to open this one up for discussion and think about it some more.

vkarpov15:
Further comments, thoughts, arguments, and haikus most welcome from anyone :+1:

CKarper:
I like this idea, it makes more sense that way for discriminators.

wentout:
Hello everyone.
[...]

PR #6100, opened 2018-02-06
For https://github.com/Automattic/mongoose/issues/2596

wentout:
Initial commit.

Allows to make model.discriminator call
using 3rd argument, tied to discriminatorKey Value
Population also works

Changes made from->to using parts of, but no model naming changes!
https://gist.github.com/wentout/2171022fe597e5b2564435d6b44eac49

**Summary**

It probably solves parts of this issue, and I able to add naming composition, check the gist above
https://github.com/Automattic/mongoose/issues/2596

**Test plan**

I made no hard changes, only interface addition is **value**, which means what to be tied to the discriminator key. And, as this is a last parameter, I thought of no additional tests. However I'm not very sure. It works in my environment for my current task. If you can give me a way, I will try to understand what I was able to break with this change.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -22,6 +22,7 @@ var applyStatics = require('./services/model/applyStatics');
 var cast = require('./cast');
 var castUpdate = require('./services/query/castUpdate');
 var discriminator = require('./services/model/discriminator');
+var getDiscriminatorByValue = require('./queryhelpers').getDiscriminatorByValue;
 var internalToObjectOptions = require('./options').internalToObjectOptions;
 var isPathSelectedInclusive = require('./services/projection/isPathSelectedInclusive');
 var get = require('lodash.get');
@@ -829,10 +830,11 @@ Model.prototype.model = function model(name) {
  *
  * @param {String} name   discriminator model name
  * @param {Schema} schema discriminator model schema
+ * @param {String} value  discriminator model typeKey value
  * @api public
  */
 
-Model.discriminator = function(name, schema) {
+Model.discriminator = function(name, schema, value) {
   var model;
   if (typeof name === 'function') {
     model = name;
@@ -842,7 +844,7 @@ Model.discriminator = function(name, schema) {
     }
   }
 
-  schema = discriminator(this, name, schema);
+  schema = discriminator(this, name, schema, value);
   if (this.db.models[name]) {
     throw new OverwriteModelError(name);
   }
@@ -2034,7 +2036,7 @@ Model.create = function create(doc, callback) {
     args.forEach(doc => {
       toExecute.push(callback => {
         var Model = this.discriminators && doc[discriminatorKey] ?
-          this.discriminators[doc[discriminatorKey]] :
+          this.discriminators[doc[discriminatorKey]] || getDiscriminatorByValue(this, doc[discriminatorKey]) :
           this;
         var toSave = doc;
         var callbackWrapper = (error, doc) => {
--- a/lib/queryhelpers.js
+++ b/lib/queryhelpers.js
@@ -46,6 +46,34 @@ exports.preparePopulationOptionsMQ = function preparePopulationOptionsMQ(query, 
   return pop;
 };
 
+
+/*!
+ * returns discriminator by discriminatorMapping.value
+ *
+ * @param {Model} model
+ * @param {string} value
+ */
+function getDiscriminatorByValue(model, value) {
+  var discriminator = null;
+  if (!model.discriminators) {
+    return discriminator;
+  }
+  for (var name in model.discriminators) {
+    var it = model.discriminators[name];
+    if (
+      it.schema &&
+      it.schema.discriminatorMapping &&
+      it.schema.discriminatorMapping.value == value
+    ) {
+      discriminator = it;
+      break;
+    }
+  }
+  return discriminator;
+}
+
+exports.getDiscriminatorByValue = getDiscriminatorByValue;
+
 /*!
  * If the document is a mapped discriminator type, it returns a model instance for that type, otherwise,
  * it returns an instance of the given model.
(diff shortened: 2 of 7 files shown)


wentout:
And what am I need to do?
I removed that trailing space, but it still finds it.

vkarpov15:
This change looks good, just a couple details:

1) What is the difference between this and #6101 ? They seem to be very similar, not sure what the difference is.
2) Can you add a test to `test/model.discriminator.test.js`? I know it won't break any existing code, but will be helpful for me to review and write relevant docs.

vkarpov15:
Also, it goes without saying, but thank you very much for your work, your help is much appreciated :+1: :medal_sports:

wentout:
Glad to See You!
Thank You for the answer!

I'm going with the tests, I will notify you when ready.
I suppose there can be stability issue with this change, I need to check.
The work is still in progress, cause I was not very familiar with how to run on my environment. But last evening I have had conversation with our test team lead, and he gave me the way how to make it.

And yes, there is difference between 6100 and 6101, you will need to choose which is better.

6101 allows very hidden discriminator model naming support: if you will send discriminarorTiedValue to true it will change discriminator model name to parentModelName (.baseModelName) plus '.' plus name argument, passed to discriminator method.
6100 leaved untouched to this cheaty behaviour, simply doing discriminator value is not tied anymore.

Thank You Very Very Much :)

vkarpov15:
I prefer #6100 to #6101, #6101 enforces the discriminator model name to be 'BaseModel.ChildModel'. I like that approach but right now I prefer to err on the side of flexibility. I'll close that one so we can focus on this one. Let me know if I can help you out with tests :+1:

wentout:
That is great. Ok.
Unfortunately, I'll be able to return to work only after February 15th.
Got to ski-ride, planned before starting this issue.
Very sorry for delay. I will try to do the tests earlier, but can't trust in this, sorry.
Kind Regards

wentout:
I got back.
I passed discriminators.test with no errors.
However, I'm still getting errors while running just "npm test"/
But I suppose it may be far related to my changes, cause I'm unable to imagine how can it be.
Also, I reviewed the following list:
./lib/schema/embedded.js
./lib/schema/documentarray.js
./lib/schema/array.js
./lib/schema.js
./lib/document.js
./lib/model.js
./lib/types/documentarray.js
./lib/types/array.js
./lib/query.js
./lib/queryhelpers.js
./lib/services/model/discriminator.js
./lib/services/model/applyHooks.js
./lib/services/populate/getSchemaTypes.js
./lib/aggregate.js

And didn't found anything able to catch my eye while seeking /discriminator/gi 
Therefore I think everything must be going well...

Thank you Very Much for awaiting!

wentout:
Hello again :)

Think this version of code is more reliable.
Some tests are failed, but again, I can't understand where there can be trouble.
However, this version has much more changes, related to other code parts, mostly type casting for discriminators in following: embeddeds, documentarrays, arrays.
This may be not so necessary, but it looks more general way for all of that.

Kind Regards!

wentout:
Hello @vkarpov15 :)

I think there will be NO additional changes.
Can't find anything else.
What do you think of current state, is it ready enough to merge or not?

If I must do something else -- I need the additional information.

Kind Regards!

vkarpov15:
LGTM :+1:

RubyRubenstahl:
@wentout, any chance you can provide a minimal usage example for this? 

I'm fairly certain that this will resolve an issue I'm having, but I'm not quite understanding how to use the new feature.

wentout:
Yes, sure, take a look.
First of all you need to read prepositions from:
https://github.com/Automattic/mongoose/issues/2596

So, our next level, for example, you have to make the following complexity:

Apple
|-- green
|-- red
|-- yellow

Tomato
|-- green
|-- red
|-- yellow

So, the "Apple" and "Tomato" are base models with related discriminators, tied to colours names:

```JS

var colours = [ 'red', 'green', 'yellow' ];

var AppleSchema = new mongoose.Schema({
   ...
   colour : {
      type : String,
      enum : colours
   }
}, {
   discriminatorKey : 'colour'
});

var AppleModel = mongoose.model('AppleModel', AppleSchema);

var TomatoSchema = new mongoose.Schema({
   ...
   colour : {
      type : String,
      enum : colours
   }
}, {
   discriminatorKey : 'colour'
});

var TomatoModel = mongoose.model('TomatoModel', TomatoSchema);
```
From now your mongoose models are:

```JS
console.log(Object.keys(mongoose.models));
// [ 'AppleModel', 'TomatoModel' ]
```
And now you are trying to add discriminators to AppleModel:


```JS
colours.forEach(colour => {
  AppleModel.discriminator(colour, { ...ThisColorDiscriminatorSchema... });
});
```
From here your models are:

```JS
console.log(Object.keys(mongoose.models));
// [ 'AppleModel', 'TomatoModel', 'red', 'green', 'yellow' ]
```

As we know, this done internally, inside of mongoose, discriminator model name is tied to discriminator key value by default!

And now, when you will try to add, TomatoModel discriminators, with, as you see, absolutely the same values for **discriminatorKey** field...

```JS
colours.forEach(colour => {
  TomatoModel.discriminator(colour, { ...ThisColorDiscriminatorSchema... });
});
```

You will reach the error, because model name is strictly tied to discriminator name. And you just can't simply re-use the same names. So... the solution is the following; 

```JS
colours.forEach(colour => {
  AppleModel.discriminator(`Apples.${colour}`, 
    { ...ThisColorDiscriminatorSchema... }, colour);
  TomatoModel.discriminator(`Tomatos.${colour}`,
    { ...ThisColorDiscriminatorSchema... }, colour);
});

console.log(Object.keys(mongoose.models));
/* [ 'AppleModel', 'TomatoModel', 
     'Apples.red', 'Apples.green', 'Apples.yellow', 
     'Tomatos.red', 'Tomatos.green', 'Tomatos.yellow' ]
*/
```

From now, as you see, **discriminatorKey** is no more tied to discriminator model name, and you are allowed to use the same string values, but you need to take care of discriminators Model Names by yourself, to reach the necessary requirement of model naming.

And, for sure you are able to make the following Hack, if you need:

```JS
colours.forEach(colour => {
  const appleColorDiscriminator = 
    AppleModel.discriminator(`Apples.${colour}`, 
       { ...ThisColorDiscriminatorSchema... }, colour);
  AppleModel.discriminators[colour] = appleColorDiscriminator;

  const tomatoColorDiscriminator = 
     TomatoModel.discriminator(`Tomatos.${colour}`, 
         { ...ThisColorDiscriminatorSchema... }, colour);
  TomatoModel.discriminators[colour] = tomatoColorDiscriminator; 
});
```

With this hack model names are new, but your models discriminators has the same names as they have had before this technique usage. And all you code parts related to collection items creation via **new model.discriminator[discriminatorName]** will work the same way as it was before. And for sure internal Mongoose mechanic will not treat your cheating with discriminators hashes, because mongoose.models has no additional model names.

Hope this will help, feel free to ask more :)

Kind Regards

RubyRubenstahl:
That all seems to make sense, thanks for the very thorough response! 

I just want to clarify one thing: the `...`'s throughout your code examples are intended as "insert your code here", not ES6 spread operators, correct?

wentout:
Yes! Sure, it all ...'s are not spread operators.
You need to put your Schema Object instead of { ...ThisColorDiscriminatorSchema... }
And that part 
```JS
   ... // is about some your code of other parts of schema
   colour : {
```

PR #6101, opened 2018-02-07
For https://github.com/Automattic/mongoose/issues/2596

wentout:
Allows to make model.discriminator call
using 3rd argument, tied to discriminatorKey Value
Population also works

Changes made from->to using parts of, but no model naming changes!
https://gist.github.com/wentout/2171022fe597e5b2564435d6b44eac49

**Summary**

It probably solves parts of this issue, and I able to add naming composition, check the gist above
https://github.com/Automattic/mongoose/issues/2596

Re Fix 4 CI about previous branch.

**Test plan**

I made no hard changes, only interface addition is **value**, which means what to be tied to the discriminator key. And, as this is a last parameter, I thought of no additional tests. However I'm not very sure. It works in my environment for my current task. If you can give me a way, I will try to understand what I was able to break with this change.


vkarpov15:
Closing in favor of #6100

Issue #6086, opened 2018-02-03
Array update operator $addToSet doesn’t work as described in the official documentation of MongoDB native driver


melyourhero:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When to I try to update a document using the $addToSet operator with strings, mongoose simply concatenates it to a single element, when I try it with numbers mongoose simply cleans the whole array.

**If the current behavior is a bug, please provide the steps to reproduce.**
Minimum steps to reproduce this behavior: https://github.com/melyourhero/mongoose-bug

**What is the expected behavior?**
The expected behavior described in official MongoDB native driver documentation here: https://docs.mongodb.com/manual/reference/operator/update/addToSet/#behavior

**Please mention your node.js, mongoose, and MongoDB version.**
Node: v8.5.0
MongoDB: v3.4.10
mongoose: v5.0.3

varunjayaraman:
youre right, it seems to be clearing the array:

```javascript
const mongoose = require('mongoose');
mongoose.Promise = global.Promise;

const GITHUB_ISSUE = `gh-6086`;

run().catch(error => console.error(error.stack));

async function run() {
  await mongoose.connect(`mongodb://localhost:27017/${GITHUB_ISSUE}`);

  const schema = new mongoose.Schema({
    letters: [String],
    numbers: [Number]
  });

  const Model = mongoose.model('Model', schema);

  // with letters
  const doc = await Model.create({ letters: ['a', 'b'] });

  const update = await Model.findByIdAndUpdate(doc._id, { $addToSet: { letters: ['c', 'd'] } }, { new: true });

  console.log(update);

  const docNumbers = await Model.create({ numbers: [1, 2] });
  const updateNumbers = await Model.findByIdAndUpdate(
    docNumbers._id,
    { $addToSet: { numbers: [3, 4] } },
    { new: true }
  );

  console.log(updateNumbers);
}
```

That final log output shows:

```
{ letters: [ 'a', 'b', 'c,d' ],  numbers: [],
  _id: 5a791b58293ba7ab0c48a349,  __v: 0 }
{ letters: [],  numbers: [],  _id: 5a791b58293ba7ab0c48a34a,  __v: 0 }
```

where `numbers` should be `[1, 2, 3, 4]`

evanhenke:
[...]

PR #6128, opened 2018-02-13
Potential fix for issue retrieving documents containing arrays within arrays

evanhenke:
Potential fix for #6086 

**Summary**
It looked like mongoose was not updating a document correctly based on whether the $addToSet modifier was adding an array of numbers vs an array of strings.  Turns out the update was working fine and the document was being saved as intended, but there were issues with its retrieval.

The reason that an array of strings "worked" was because casting an array to string would cast to a literal string in the shape of the array, e.g. "['foo','bar']" the problem here was it would be of type string instead of truly being an array within another array.

In the case of an array of numbers, it would try to cast the array to a number and fail and throw a CastError and eventually send back an empty array.

The fix was to recursively call SchemaArray.cast until the value to be cast was no longer an array.

**Test plan**

Reran scripts given with issue statement as well as existing unit tests

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -185,7 +185,8 @@ SchemaArray.prototype.cast = function(value, doc, init) {
     if (this.caster) {
       try {
         for (i = 0, l = value.length; i < l; i++) {
-          value[i] = this.caster.cast(value[i], doc, init);
+          value[i] = Array.isArray(value[i]) ?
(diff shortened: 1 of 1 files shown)


vkarpov15:
@evanhenke it looks like the tests all fail. Can you please fix the tests and add your own test case that reproduces the issue that you're seeing?

evanhenke:
I see that you had a better fix so I closed this

Issue #6145, opened 2018-02-16
Dates saved to non-strict sub properties not stored in Date format (5.0.5+)


s4rbagamble:
Using mongoose@5.0.6 and mongoDB v3.4 and node@9.5

There is a bug where saving a date that is not explicitly specified in the model's schema saves the date as a unix timestamp (Number) rather than as a Date.

This appears to have been introduced in mongoose@5.0.5

### Replication

Run the following script with mongoose@5.0.4 and 5.0.6

```javascript
const mongoose = require('mongoose');

mongoose.connect('mongodb://localhost/test').then(async db => {
	const ProductSchema = new mongoose.Schema({
		_id: String,
		name: String,
		startDate: Date,
		other: {}
	});

	db.model('Product', ProductSchema);

	await db.model('Product').remove(); // clean up past runs of the script

	await db.model('Product').update({
		_id: 'product1',
	}, {
		name: 'Test product',
		startDate: new Date(),
		other: {
			endDate: new Date(),
		}
	}, { upsert: 1 });

	console.log(await db.model('Product').findOne().lean());

	process.exit();
});
```

Output mongoose@5.0.6 and 5.0.5
```
{ _id: 'product1',
  __v: 0,
  specific: { thisDate: 2018-02-16T14:23:36.436Z },
  other: { endDate: 1518791016436 }, 
  startDate: 2018-02-16T14:23:36.436Z,
  name: 'Test product' }
```

Output mongoose@5.0.4
```
{ _id: 'product1',
  __v: 0,
  specific: { thisDate: 2018-02-16T14:24:34.834Z },
  other: { endDate: 2018-02-16T14:24:34.834Z },
  startDate: 2018-02-16T14:24:34.834Z,
  name: 'Test product' }
```

Key difference is `other: { endDate: 1518791016436 }, ` vs `other: { endDate: 2018-02-16T14:24:34.834Z },`

The property should be saved as as a date.

vkarpov15:
Good catch, thanks for finding this :+1:

PR #6146, opened 2018-02-16
Prevented converting of dates to timestamps when respecting valueOf()

s4rbagamble:
**Summary**

Add check when respecting object's `valueOf` to ignore dates. This prevents dates being cast as timestamps when saved to the database as reported in #6145, where `Date.valueOf` returns a timestamp.

**Test plan**
 The script described in #6145 functions as expected


vkarpov15:
Good catch, thanks for finding this! :+1:

PR #6160, opened 2018-02-21
migration doc fix: correct syntax is eachAsync

simllll:
this is also wrong in the documentation http://mongoosejs.com/docs/api.html#Aggregate (or at least not documentated that in mongoose 5 / useMongooseAggCursor flag you have to use eachAsync instead of each)


vkarpov15:
Thanks :beers:

PR #6179, opened 2018-02-27
Add dbName parameter to documentation

petrometro:
Without obscure dbName parameter, mongoose.connect to Atlas only connects to admin database, even if I put my desired database name after the slash in the uri. The only way to make mongoose work in this context is to pass it to dbName.

**Summary**

**Test plan**


vkarpov15:
Thanks but `connections.html` is compiled code, please modify `docs/connections.jade` instead.

vkarpov15:
Ended up fixing it in 98c6cb9, thanks for your PR :+1:

Issue #6176, opened 2018-02-27
Array required validation is not checked


natanavra:
EDIT:
Found this: https://github.com/Automattic/mongoose/issues/1335#issuecomment-252129243
With the following schema I get the expected behaviour:
```js
{
  testArray: {
    type: [String],
    default: () => void 0,
    required: true
  }
}
```
But this should definitely show in `required` documentation

------------------------------------
**Do you want to request a feature or report a bug?**
Bug

**What is the current behavior?**
An empty array is always saved in the database even if not provided upon creation.

**If the current behavior is a bug, please provide the steps to reproduce.**
Create a model with the following schema: 
```js
{testArray: {type: [String], required: true}}
```
or even this schema: 
```js
{
  testArray: {
        type: [String],
	required: function() {
		return Array.isArray(this.testArray) && this.testArray.length > 0;
  	}
  }
}
```
Save a model of this type:
```js
let obj = new Model({});
obj.save();
```
You will have a document in the database with an empty `testArray`

**What is the expected behavior?**
- the existence of the array in the creation `new Model({})` should be checked.
- Additionally, Mongoose implicitly casts array items and implicitly creates arrays when passing an primitives. This is not documented and leads to unexpected behaviour which I just found out about.

**Please mention your node.js, mongoose and MongoDB version.**
- Mongoose v5.0.7
- Node v6.9.2
- Mongo v3.4.0

vkarpov15:
Thanks for reporting, will investigate ASAP

varunjayaraman:
yup, this looks like a confirmed bug:

```javascript
const mongoose = require('mongoose');
mongoose.Promise = global.Promise;

const GITHUB_ISSUE = `gh-6176`;

run().catch(error => console.error(error.stack));

async function run() {
  await mongoose.connect(`mongodb://localhost:27017/${GITHUB_ISSUE}`);

  const schema = new mongoose.Schema({
    testArray: {
      type: [String],
      required: function() {
        return Array.isArray(this.testArray) && this.testArray.length > 0;
      }
    }
  });

  const Model = mongoose.model('Model', schema);

  console.log(await Model.create({}));
}
```

log output:
```
{ testArray: [], _id: 5aa063141b34a8df566fab98, __v: 0 }
```

vkarpov15:
[...]

Issue #6170, opened 2018-02-24
FAQ about casting aggregation pipeline stages


peterpeterparker:
**Do you want to request a *feature* or report a *bug*?**

Not sure if it's a bug or not and neither if it's related to mongoose or mongo

**What is the current behavior?**
I could `find` elements `greater` as a `Date.now()` but if I do the same with `aggregate` it doesn't find any elements

    let query = {
         something: : {$gte: Date.now()}
    };
    
    Collection.find(query).exec(function (err, results) {
         console.log(results.length); // --> All good, x elements found
    }
    
    Collection.aggregate([{$match: query}], function (err, items) {
         console.log(results.length); // --> 0 elements found
    }

if I modify the query with `new Date()` instead of `Date.now()`, problem is solved

      let query = {
         something: : {$gte: new Date()}
    };
    
    Collection.aggregate([{$match: query}], function (err, items) {
         console.log(results.length); // --> All good, x elements found
    }
 
**If the current behavior is a bug, please provide the steps to reproduce.**

See above code

**What is the expected behavior?**

Would be cool to have the same behavior between `find` and `aggregate` regarding this comparison.

**Please mention your node.js, mongoose and MongoDB version.**

Node v9.4.0
mongoose v5.0.6
mongo v3.2.14

lineus:
it looks like this is a difference in the way find and aggregate work in mongodb. I created a test collection with this populate script:
```
#!/usr/bin/env node
'use strict'

const mongoose = require('../lib/test_db')
const Schema = mongoose.Schema
const faker = require('faker')

const testSchema = new Schema({
  what: String,
  when: Date
})

const Test = mongoose.model('test', testSchema)

const docs = []

for (let i = 0; i < 100; i++) {
  docs.push(new Test({
    what: faker.company.bsAdjective(),
    when: faker.date.past()
  }))
  docs.push(new Test({
    what: faker.hacker.adjective(),
    when: faker.date.future()
  }))
}

Test.create(docs)
  .then(
    (res) => console.log(res),
    (rej) => console.error(rej)
  )
```
my find script:
```
#!/usr/bin/env node
'use strict'

const mongoose = require('../lib/test_db')
const Schema = mongoose.Schema

const testSchema = new Schema({
  what: String,
  when: Date
})

const Test = mongoose.model('test', testSchema)

const query = {
  when: { $lte: Date.now() }
}

Test.find(query).exec((err, res) => {
  if (err) { return console.error(err) }
[...]

PR #6182, opened 2018-02-28
adds faqs for #6170 and #6176

lineus:
adding faqs for  #6170 and #6176

tested with make docclean && make gendocs && node static.js


lineus:
killing due to style.

PR #6184, opened 2018-02-28
adding faqs for 6170 and 6176

lineus:
adding a couple faqs per #6170 and #6176

### DIFF
--- a/docs/faq.jade
+++ b/docs/faq.jade
@@ -324,9 +324,7 @@ block content
     const CollectionSchema = new Schema({
       field1: { 
         type: [String], 
-        default: function() { 
-          return null 
-        }
+        default: void 0
       }
     });
     ```
--- a/docs/faq.jade
+++ b/docs/faq.jade
@@ -316,6 +316,29 @@ block content
     var Kitten = connection.model('Kitten', kittySchema);
     ```
 
+    **Q**. How can I change mongoose's default behavior of initializing an array 
+    path to an empty array so that I can require real data on document creation?
+
+    **A**. You can set the default of the array to a function that returns null.
+    ```javascript
+    const CollectionSchema = new Schema({
+      field1: { 
+        type: [String], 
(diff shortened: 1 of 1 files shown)


docs/faq.jade
```diff
@@ -316,6 +316,29 @@ block content
...
 
+    **Q**. How can I change mongoose's default behavior of initializing an array 
+    path to an empty array so that I can require real data on document creation?
+
+    **A**. You can set the default of the array to a function that returns null.
+    ```javascript
+    const CollectionSchema = new Schema({
+      field1: { 
+        type: [String], 
+        default: function() {
```
vkarpov15: I think the more correct way here is `default: void 0`?

lineus: done :) Thanks Val!

lineus:
it's worth noting that I tested this on 5.0.7:
```
#!/usr/bin/env node
'use strict'

const mongoose = require('./lib/test_db')
const Schema = mongoose.Schema

const voidSchema = new Schema({
  name: {
    type: [String],
    default: void 0
  }
})

const noVoidSchema = new Schema({
  name: {
    type: [String]
  }
})

const Test1 = mongoose.model('test1', voidSchema)
const Test2 = mongoose.model('test2', noVoidSchema)

const voidTest = new Test1({})
const noVoidTest = new Test2({})

voidTest.save((err, vdoc) => {
  if (err) { return console.error(err) }
  console.log(vdoc)
  noVoidTest.save((err, ndoc) => {
    if (err) { return console.error(err) }
    mongoose.connection.close()
    return console.log(ndoc)
  })
})
```
output:
```
InspiredMacPro:mongoose5 lineus$ ./voidTest.js 
{ _id: 5a9a786c124008afd971815e, __v: 0 }
{ name: [], _id: 5a9a786c124008afd971815f, __v: 0 }
InspiredMacPro:mongoose5 lineus$ 
```

PR #6192, opened 2018-03-02
Update promises.html

sansSpoon:
remove typo, line 5 open bracket before "if you&#39;re using"

**Summary**

**Test plan**


vkarpov15:
Thanks but `index.html` is compiled code, so these changes would get overwritten by our build system. Please modify `index.jade` instead.

sansSpoon:
@vkarpov15, I updated the `promises.html` not `index.html`. There is *no* `promise.jade` file in the build system. Not sure if this is a bug in itself or there never was a `promises.jade` file, tried `git log --follow 'docs/promises.jade'`

vkarpov15:
Oh my mistake. The file you need to modify is `test/docs/promises.test.js`, we compile that file to generate the `promises.html` file

Issue #6217, opened 2018-03-08
Deprecation warning when connecting with 4.13.11


bgSosh:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**What is the current behavior?**

I'm getting this message when I connect with latest release of v4:

```
`open()` is deprecated in mongoose >= 4.11.0, use `openUri()` instead, or set the `useMongoClient` option if using `connect()` or `createConnection()`. See http://mongoosejs.com/docs/connections.html#use-mongo-client
```

I found a previous report, but the issue was closed and the comments were locked.  

I checked the document link in the warning, but all I can see about this issue (under heading 'Option Changes in v5.x') relates to v5.  Please could you clarify what the situation is for latest v4?  Should I be adding useMongoClient in options?

**Please mention your node.js, mongoose and MongoDB version.**

Node 4.8.2, Mongoose 4.13.11, Mongodb 3.2,

lineus:
Hi @bgSosh. 
to get around this issue you can add an options object as the 2nd arg to .connect() with useMongoClient: true.

here is an example:
`mongoose.connect('mongodb://localhost:27017/test', { useMongoClient: true })`

Thanks for pointing out the issue with the error message. I'll submit a pull request to update it to point to the 4.x docs [here](http://mongoosejs.com/docs/4.x/docs/connections.html#use-mongo-client).

bgSosh:
Thanks @lineus - all clear now!

bgSosh:
@lineus Hmm, ran into an issue with this - is there perhaps no createConnection that accepts both an option and a callback?

This works:
```
 dbConnection = mongoose.createConnection(connectionString_Forum, function(err) { 
       //Called back
 });
```
But this doesn't:
```
 dbConnection = mongoose.createConnection(connectionString_Forum, { useMongoClient: true }, function(err) { 
       //Not called back
 });
```

lineus:
createConnection() doesn't take a callback in mongoose version 4. it only takes a uri and options and returns an instance of the native driver connection. Are you connecting to multiple dbs or is there some other compelling reason to use mongoose.createConnection() instead of mongoose.connect()?

bgSosh:
Exactly, we need to connect to multiple dbs, and need a reference to the non-default connections.

Thanks for the clarification on createConnection().  Though, I can confirm that the callback is definitely called (when cb is the second parameter, but perhaps that's some unintended effect?  

Is there any reason why connect() and createConnection() should have these differences?

[...]

PR #6225, opened 2018-03-09
add index.js to api docs

lineus:
While looking at createConnection for #6217, I noticed that createConnection is missing from the 5.x api docs. I added index.js to the files array in docs/source/api.js at the top to mimic it's previous placement in the 4.x docs. 

tested with make docclean && make gendocs && node static.js


vkarpov15:
Thanks :+1:

PR #6267, opened 2018-03-22
adding hint method to aggregate

lineus:
MongoDB 3.6 added the hint option to aggregations. you can currently add a hint with the .option method, but having a convenience method like the one for allowDiskUse makes sense to me.

I also updated the helper method formerly known as onlyTestMongo34. I made the version it looks for variable and changed the name to reflect the new behavior.

Test expects that the resulting array of docs is sorted according to the hint.

```
mongoose>: npm test
....
  2 failing

  1) Model
       bug fixes
         3.6 features
           watch() (gh-5964):
     Uncaught MongoError: Majority read concern requested, but it is not supported by the storage engine.
      at queryCallback (node_modules/mongodb-core/lib/cursor.js:223:25)
      at node_modules/mongodb-core/lib/connection/pool.js:541:18
      at _combinedTickCallback (internal/process/next_tick.js:131:7)
      at process._tickCallback (internal/process/next_tick.js:180:9)

  2) Model
       bug fixes
         3.6 features
           watch() before connecting (gh-5964):
     Uncaught MongoError: Majority read concern requested, but it is not supported by the storage engine.
      at queryCallback (node_modules/mongodb-core/lib/cursor.js:223:25)
      at node_modules/mongodb-core/lib/connection/pool.js:541:18
      at _combinedTickCallback (internal/process/next_tick.js:131:7)
      at process._tickCallback (internal/process/next_tick.js:180:9)



```
I'm assuming it's my env that's causing these failures.

### DIFF
--- a/test/aggregate.test.js
+++ b/test/aggregate.test.js
@@ -62,11 +62,11 @@ function onlyTestAtOrAbove(semver, ctx, done) {
       return;
     }
 
-    var [major, minor] = semver.split('.').map(function(s) {
+    var desired = semver.split('.').map(function(s) {
       return parseInt(s);
     });
 
-    var meetsMinimum = version[0] > major || (version[0] === major && version[1] >= minor);
+    var meetsMinimum = version[0] > desired[0] || (version[0] === desired[0] && version[1] >= desired[1]);
 
     if (!meetsMinimum) {
       ctx.skip();
--- a/test/aggregate.test.js
+++ b/test/aggregate.test.js
@@ -48,20 +48,27 @@ function setupData(db, callback) {
 }
 
 /**
- * Helper function to test operators that only work in MongoDB 3.4 and above (such as some aggregation pipeline operators)
+ * Helper function to test operators that only work in a specific versoin of MongoDB and above (such as some aggregation pipeline operators)
  *
+ * @param {String} semver, `3.4`, specify minimum compatible mongod version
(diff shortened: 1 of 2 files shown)


test/aggregate.test.js
```diff
@@ -48,20 +48,27 @@ function setupData(db, callback) {
...
+function onlyTestAtOrAbove(semver, ctx, done) {
   start.mongodVersion(function(err, version) {
     if (err) {
       done(err);
       return;
     }
-    var mongo34 = version[0] > 3 || (version[0] === 3 && version[1] >= 4);
-    if (!mongo34) {
+
+    var [major, minor] = semver.split('.').map(function(s) {
```
vkarpov15: Array destructuring doesn't work in node v4 and v5, which is one of the reason why tests are failing

lineus: ok, I'll fix it.

lineus:
The destructuring error is gone, but now on node 4, 6, and 7 there are different errors. Not just different from before, but different from each other, and in places that I can't readily explain by the changes I made. I'll keep digging into it. I know I shouldn't *want* to re-run the tests on the same commit in travis, it feels insane, but I do :)

lineus:
I created a new branch and a new PR #6272 with the same changes to test this code again, all tests passed. Is there a better strategy than this when it seems something went wrong on travis CI?

Issue #4353, opened 2016-07-21
ES6 Set type


ericandrewlewis:
Is it possible to use the `Map` and `Set` data types in a Mongoose schema?

vkarpov15:
Not at the moment. Worth supporting in a future release

Issue #681, opened 2012-01-12
docs embedded in object w/ arbitrary keys?


ghost:
Is there a way to use embedded docs within a object that doesn't have predetermined keys? i.e. Rather than having a situation like this:

``` javascript
var myschema = new mongoose.Schema({
   field : { foo: String, bar: String }
});
```

Being able to say something like:

``` javascript
var myschema = new mongoose.Schema({
   field : { *: String }
});
```

Makes it a little more array like in that way - but the idea is that I don't know the keys in advance, but I'd still like to be able to have an embedded doc.

coty-crg:
+1

tieblumer:
+1

rubenstolk:
+1

rohithn:
+1

Elements-:
+1

smarom25:
+1

aivel:
+1

voodee:
+1

raub:
+1

gabrielcsapo:
+1

Kanwer-Singh:
+1

Michael77:
I have needed this a well

mosane:
+1

qqilihq:
Stop those meaningless +1 , which end up in watching peoples' inboxes. 

There are "Subscribe" and "Reaction" buttons for a reason.

coty-crg:
+1

raub:
@qqilihq, do implement plx, then we stop +1, kthxbye

Flicksie:
"no one has needed it before"
5 years later, a ton of +1s later, still no updates on this?

vkarpov15:
Since this is the oldest issue I can find related to this one, I'll re-open this and close #1522 so as many people as possible can track for updates.

doneumark:
+1

adamreisnz:
How did this feature get released so quietly that I wasn't aware of it?! 💯

vkarpov15:
@adamreisnz lol thanks. Where do you look for Mongoose feature updates? Changelog? Twitter?

adamreisnz:
Changelog usually, but must have missed this one!

objective-cwan67:
> +1





https://admiral-tranquil.github.io/ram58/

PR #6287, opened 2018-03-27
WIP: Map type

vkarpov15:
Re: #681, #4353, and several other issues, we're planning on adding a Map type that builds on [ES6 maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map). There's some more work for this PR already in the 5.1 branch in f828de3e6b844eaa364dc38225a9a2f68792c924.

The primary motivation for maps is to support arbitrary keys while retaining change tracking and validation. Proxies would be an alternative implementation, but I'm not confident we could implement a pure proxy solution in a reasonable time frame. As a consequence, users will have to use `Map.prototype.get()` to get keys and `Map.prototype.set()` to set keys, which makes change tracking much easier but adds some overhead. However, devs that use maps should already be used to this syntax, so mongoose maps will feel familiar.

There's a noteworthy caveat that, under the hood, maps become objects when they're stored in MongoDB because there's no notion of a [map type in BSON](http://bsonspec.org/).

Still lots of work to be done to make sure maps work right with queries. @varunjayaraman @lineus I would love to hear your thoughts.

Here's a sample usage from the test case:

```javascript
    let nestedValidateCalls = [];
    let validateCalls = [];
    const TestSchema = new mongoose.Schema({
      v: {
        type: Map,
        of: {
          type: Number,
          validate: function(v) {
            nestedValidateCalls.push(v);
            return v < 4;
          }
        },
        validate: function(v) {
          validateCalls.push(v);
          return true;
        }
      }
    });

    const Test = db.model('MapTest', TestSchema);

    return co(function*() {
      const doc = yield Test.create({ v: { x: 1 } });
      assert.deepEqual(nestedValidateCalls, [1]);
      assert.equal(validateCalls.length, 1);
      assert.equal(validateCalls[0].get('x'), 1);

      assert.ok(doc.v instanceof Map);

      let threw = false;

      try {
        yield Test.create({ v: { notA: 'number' } });
      } catch (error) {
        threw = true;
        assert.ok(!error.errors['v']);
        assert.ok(error.errors['v.notA']);
      }
      assert.ok(threw);

      doc.v.set('y', 5);

      threw = false;
      try {
        yield doc.save();
      } catch (error) {
        threw = true;
        assert.ok(!error.errors['v']);
        assert.ok(error.errors['v.y']);
      }
      assert.ok(threw);
    });
```


lineus:
@vkarpov15 I haven't used Maps yet, but I'm going to see if I can wrap my head around them (specifically with respect to mongoose) in the next day or so. I've been reading about them but I have a ways to go before I can add much value to the conversation. I'll be watching for sure, and if there's anything you think I can do to help just say the word.

vkarpov15:
@lineus the particular details of maps are interesting but I think the biggest win here is that we can support objects with arbitrary keys. So:

```javascript
const TestSchema = new mongoose.Schema({
  myMap: {
    type: Map,
    of: Number
  }
});
```

Would mean that `doc.myMap.set('key1', 1)`, `doc.myMap.set('key2', 2)`, etc. are valid. Even if you aren't an expert in maps, I'd appreciate any thoughts you have on that and any testing you could do.

simonrichard:
This feature was implemented in version 5.1.0 but still appears to be unavailable as of 5.2.3. Am I missing something?

Issue #6284, opened 2018-03-27
Nested populate array and spaces in path don't work together


AbdelrahmanHafez:
**Do you want to request a *feature* or report a *bug*?** Report a bug.
The following code would populate the cityId & districtId fields succesfully, _with only using the space between the two fields_.

```
Post.find({})
  .populate({
    path: 'userId',
    populate: {
      path: 'cityId districtId',
      select: 'name'
    }
  });
```

The following would populate the `houseId` succesfully, and would *only* populate the `cityId` field, would not populate both `cityId` & `districtId` as expected.

```
Post.find({})
  .populate({
    path: 'userId',
    populate: [{
      path: 'houseId',
      select: 'location'
    }, {
      path: 'cityId districtId',
      select: 'name'
    }]
  });
```

Node v8.9.3
Mongoose v5.0.1
MongoDB v3.2.7

lineus:
confirmed with the following repro script:

### 6284.js
```
#!/usr/bin/env node
'use strict'

const assert = require('assert')
const mongoose = require('mongoose')
mongoose.connect('mongodb://localhost/test')
const Schema = mongoose.Schema

const houseSchema = new Schema({
  location: String
})

const userSchema = new Schema({
  name: String,
  houseId: {
    type: Schema.Types.ObjectId,
    ref: 'house'
  },
  cityId: {
    type: Schema.Types.ObjectId,
    ref: 'city'
  },
  districtId: {
    type: Schema.Types.ObjectId,
    ref: 'district'
  }
})

const citySchema = new Schema({
  name: String
})

const districtSchema = new Schema({
  name: String
})

const postSchema = new Schema({
  content: String,
  userId: {
    type: Schema.Types.ObjectId,
    ref: 'user'
  }
})

const House = mongoose.model('house', houseSchema)
const User = mongoose.model('user', userSchema)
const City = mongoose.model('city', citySchema)
const District = mongoose.model('district', districtSchema)
const Post = mongoose.model('post', postSchema)

const house = new House({ location: '123 abc st.' })
const city = new City({ name: 'Somecity' })
const district = new District({ name: 'That District' })

const user = new User({
  name: 'Billy',
  houseId: house._id,
  districtId: district._id,
  cityId: city._id
})

const post = new Post({
  content: 'Some meaningful insight.',
  userId: user._id
})

function cleanUp () {
  let colls = [ City, District, Post, User, House ]
  colls.forEach(async (Model) => {
    await Model.remove({})
  })
[...]

PR #6296, opened 2018-03-30
Feature #6284 process arrays in sub-populate option.

lineus:
Currently you can pass in an array as a sub-populate option as demonstrated in #6284, but only if the `path` in all of the sub-populate objects are singles ( not space separated ). The lack of support for space separated paths and the fact that currently, the only conditional on subPopulate  tests for `typeof ===  'object'` leads me to believe that this behavior of partially working with arrays is a *mostly* happy accident. This just formalizes the behavior.

all tests including the one I added pass ( except the same 2 that fail for me on master). my next goal is to sort out my environment, not sure why I keep seeing these:
```
  1) Model
       bug fixes
         3.6 features
           watch() (gh-5964):
     Uncaught MongoError: Majority read concern requested, but it is not supported by the storageengine.
```
I'm using mongodb server 3.6.2 with mmapv1 ~~so it's a bit perplexing :)~~ looks like I need wiredTiger for this, so that explains that.


lineus:
Are transient errors on travis ci common? Is there a way in the travis ui to re-run tests?

I resubmitted this same code in a second PR, and it failed again in different versions of node for different reasons. 

I'm not sure what to do. Either I leave a PR here with failing checks on travis, or spam you with newly opened pr requests until the tests pass.

lib/utils.js
```diff
@@ -538,6 +538,23 @@ exports.populate = function populate(path, select, model, match, options, subPop
...
+          ret.push(exports.populate(copy)[0]);
+        });
+      } else {
+        ret.push(exports.populate(obj)[0]);
+      }
+    });
+    subPopulate = exports.populate(ret);
+  }
+
   if (typeof subPopulate === 'object') {
```
vkarpov15: Make this an `else`, because `typeof [] === 'object'` so this `if` will also execute.

lib/utils.js
```diff
@@ -538,6 +538,23 @@ exports.populate = function populate(path, select, model, match, options, subPop
     throw new TypeError('utils.populate: invalid path. Expected string. Got typeof `' + typeof path + '`');
   }
 
+  if (Array.isArray(subPopulate)) {
+    let ret = [];
+    subPopulate.forEach(function(obj) {
+      if (/[\s]/.test(obj.path)) {
+        let copy = exports.object.shallowCopy(obj);
```
vkarpov15: I'd just use `Object.assign({}, obj)`, more standard way to do shallow cloning.

lib/utils.js
```diff
@@ -538,6 +538,23 @@ exports.populate = function populate(path, select, model, match, options, subPop
...
+        let paths = copy.path.split(' ');
+        paths.forEach(function(p) {
+          copy.path = p;
```
vkarpov15: Do you need to clone this?

test/model.populate.test.js
```diff
@@ -6073,5 +6073,87 @@ describe('model: populate:', function() {
...
+          districtId: district._id,
+          cityId: city._id
+        });
+
+        var post = new Post({
+          content: 'Some meaningful insight.',
+          userId: user._id
+        });
+
+        House.create(house, function() {
```
vkarpov15: Can you just use `co` here? We're trying to avoid callbacks in test code because co makes things much more readable.

lib/utils.js
lineus: because the select, populate, and match options are all going to be the same for each one, I can just duplicate the obj once ( line 545 ) and reuse it to call populate on the same object  with a new path each time.

lib/utils.js
lineus: fixed.

lib/utils.js
lineus: fixed.

test/model.populate.test.js
lineus: I'll checkout the existing examples of co and read the docs for it if I need to. I see a lot of tests use promises as well, is that also an acceptable option or should I stick with co?

lineus: co is way better. it's like async/await but from 2 years before it existed :). done.

PR #6298, opened 2018-03-31
Update schematypes.html

rayattack:
Corrected typo Sumbers = Numbers

**Summary**

**Test plan**


lineus:
@rayattack the schematypes.html file is generated dynamically from schematypes.jade. You need to update  the jade file instead:
```
docs: grep Sumbers schematypes.jade 
    and other general characteristics for Strings and Sumbers.
docs:
```

Issue #6299, opened 2018-03-31
Not playing along nicely with Decimal.js library


adamreisnz:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
I am using [decimal.js](https://github.com/MikeMcl/decimal.js/) to work with numbers, and it's supposed to have a cast to `Number` support, so you can just substitute Decimal objects whenever you'd use `Number`'s

However, when I pass these values to a Mongoose model, it seems to incorrectly cast it to a Number. The value becomes an object instead.

**If the current behavior is a bug, please provide the steps to reproduce.**
```js
'use strict';

const mongoose = require('mongoose');
const Decimal = require('decimal.js');
const {Schema, Types: {ObjectId}} = mongoose;

const TestSchema = new Schema({
  amount1: {
    type: Number,
  },
  amount2: {
    type: Number,
  },
});

const Test = mongoose.model('Test', TestSchema);

//Create test number
const num = new Decimal(10);

//For reference, use toNumber and casted values
const toNum = num.toNumber();
const casted = Number(num);

//Create instance
const test = new Test({
  amount1: 5,
  amount2: num,
});

//Test output
console.log(toNum, typeof toNum); //Expected 10, 'number'
console.log(casted, typeof casted); //Expected 10, 'number'
console.log(test.amount1, typeof test.amount1); //Expected 5, 'number'
console.log(test.amount2, typeof test.amount2); //Expected 10, 'number'
```

**What is the expected behavior?**
Expected output of console logs is:

```
10 'number'
10 'number'
5 'number'
10 'number'
```

Actual output however is:
```
10 'number'
10 'number'
5 'number'
[Number: 10] 'object'
```

That last entry is boggling me. What is Mongoose doing to produce this kind of strange object? It seems to neither cast the value to a `Number`, nor call the `toNumber()` method.

Worth noting that after actual persisting to database, the value is correctly cast and saved.

**Possible solutions**
I don't know where in the code this casting happens, but it may be that a `typeof value === 'object'` check is done and then it incorrectly converts. Maybe it needs a check for a `toNumber` method and then call it instead, or simply coerce with `Number(value)`?

**Please mention your node.js, mongoose and MongoDB version.**
Node 8.10.0
Mongoose 5.0.12

rwmia:
Decimal js creates objects from values. Use demicaljs to convert to number before inserting to db. I think it’s toDP(num).

adamreisnz:
[...]

PR #6306, opened 2018-04-03
adding toNumber to SchemaNumber cast method (gh-6299)

lineus:
#6299 when casting a decimals.js object with new Number(obj), the return is not a converted to a number as expected. This PR, adds a conditional call to obj.toNumber if it exists and is a function that returns the object's representation of itself as a number. 

This could have also been solved by just returning `Number(obj)` all of the time instead of `new Number(obj)` but this PR seems less backwards breaking on the off chance folks are relying explicitly on instanceof Number vs typeof Number.

The output changes to the expected output in #6299 and the test I added to types.number.test.js passes:
```
  it('prefers toNumber if one exists (gh-6299)', function(done) {
    var n = new SchemaNumber();
    var obj = {
      str: '10',
      toNumber: function() {
        return Number(this.str);
      },
      toString: function() {
        return '11';
      }
    };
    assert.strictEqual(n.cast(obj), 10);
    done();
  });
```
mongoose>: npm test -- -g 'prefers toNumber if one exists'

> mongoose@5.0.13-pre test /Users/lineus/dev/opc/mongoose
> mocha --exit test/*.test.js test/**/*.test.js "-g" "prefers toNumber if one exists"

 You're not testing shards!

  ․

  1 passing (5s)

mongoose>:
```

<!-- Demonstrate the code is solid. Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI. -->

### DIFF
--- a/lib/schema/number.js
+++ b/lib/schema/number.js
@@ -223,6 +223,9 @@ SchemaNumber.prototype.cast = function(value, doc, init) {
     if (typeof val === 'number') {
       return val;
     }
+    if (typeof val.toNumber === 'function') {
+      return val.toNumber();
+    }
     if (val.toString && !Array.isArray(val) && val.toString() == Number(val)) {
       return new Number(val);
     }
--- a/test/types.number.test.js
+++ b/test/types.number.test.js
@@ -4,8 +4,8 @@
  */
 
 var mongoose = require('./common').mongoose,
-    SchemaNumber = mongoose.Schema.Types.Number,
-    assert = require('power-assert');
+  SchemaNumber = mongoose.Schema.Types.Number,
+  assert = require('power-assert');
 
 /**
  * Test.
@@ -91,6 +91,21 @@ describe('types.number', function() {
     var n = new SchemaNumber();
     assert.strictEqual(n.cast(true), 1);
     assert.strictEqual(n.cast(false), 0);
+    done();
+  });
+
+  it('prefers toNumber if one exists (gh-6299)', function(done) {
+    var n = new SchemaNumber();
+    var obj = {
+      str: '10',
+      toNumber: function() {
+        return Number(this.str);
+      },
+      toString: function() {
+        return '11';
+      }
+    };
+    assert.strictEqual(n.cast(obj), 10);
     done();
   });
 });

### CODE REVIEW

**Comment by adamreisnz** [comment]
So yeah, I was a bit worried about `valueOf` and how it doesn't in fact return a `Number`, which is kind of strange. There's probably a reason for it on their end, so not sure how best to approach this.

Can you make it work on your end with using the output of `valueOf`?

**Comment by lineus** [comment]
I made the following changes:

```
mongoose>: git diff
diff --git a/lib/schema/number.js b/lib/schema/number.js
index b813025..1f207e9 100644
--- a/lib/schema/number.js
+++ b/lib/schema/number.js
@@ -223,8 +223,8 @@ SchemaNumber.prototype.cast = function(value, doc, init) {
     if (typeof val === 'number') {
       return val;
     }
-    if (typeof val.toNumber === 'function') {
-      return val.toNumber();
+    if (typeof val.valueOf === 'function') {
+      val = val.valueOf();
     }
     if (val.toString && !Array.isArray(val) && val.toString() == Number(val)) {
       return new Number(val);
diff --git a/test/types.number.test.js b/test/types.number.test.js
index 0eb7e4b..1980f2d 100644
--- a/test/types.number.test.js
+++ b/test/types.number.test.js
@@ -94,12 +94,12 @@ describe('types.number', function() {
     done();
   });

-  it('prefers toNumber if one exists (gh-6299)', function(done) {
+  it('prefers valueOf function if one exists (gh-6299)', function(done) {
     var n = new SchemaNumber();
     var obj = {
       str: '10',
-      toNumber: function() {
-        return Number(this.str);
+      valueOf: function() {
+        return this.str;
       },
       toString: function() {
         return '11';
mongoose>:
```

the test fails with:
```
mongoose>: npm test -- -g 'prefers valueOf function if one exists'

> mongoose@5.0.13-pre test /Users/lineus/dev/opc/mongoose
> mocha --exit test/*.test.js test/**/*.test.js "-g" "prefers valueOf function if oneexists"


 You're not testing shards!


  !

  0 passing (111ms)
  1 failing

  1) types.number
       prefers valueOf function if one exists (gh-6299):

      AssertionError [ERR_ASSERTION]: [Number: 10] === 10
      + expected - actual


      at Function.decoratedAssert [as strictEqual] (node_modules/empower-core/lib/decorate.js:51:30)
      at Context.<anonymous> (test/types.number.test.js:108:12)



mongoose>:
```

the `new Number` bit is still returning a Number object instead of a primitive even with a string as the input.

**Comment by lineus** [comment]
since we've already just returned strings converted to primitives with:
```
    if (typeof val === 'string' || typeof val === 'boolean') {
      val = Number(val);
    }
```
can we do something like:
```
    if (typeof val.valueOf === 'function' && (typeof val.valueOf() === 'string')) {
      return Number(val.valueOf());
    }
    if (val.toString && !Array.isArray(val) && val.toString() == Number(val)) {
      return new Number(val);
    }
  }
```
instead?

**Comment by lineus** [comment]
Passing a string into `new Number` in that last if statement still returns a Number object instead of a primitive. 

@vkarpov15 would it be too dangerous to change `return new Number(val)` to `return Number(val)` ?

or do:

```
    if (typeof val.valueOf === 'function') {
      return Number(val.valueOf());
    }
```

**Comment by vkarpov15** [comment]
@lineus this should be fine:

```javascript
if (typeof val.valueOf === 'function') {
      return Number(val.valueOf());
    }
```

adamreisnz:
Awesome, thanks for the swift action on this one 👍

vkarpov15:
Happy to help @adamreisnz , thanks for your insights :+1:

So weird that `Number([]) == 0`...

PR #6309, opened 2018-04-03
adding middleware to faq arrow function warning

lineus:
adding middleware to list of places on the FAQ not to use arrow functions.

### DIFF
--- a/docs/faq.jade
+++ b/docs/faq.jade
@@ -125,7 +125,7 @@ block content
 
     <hr id="arrow-functions" />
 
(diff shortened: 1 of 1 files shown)


docs/faq.jade
```diff
@@ -125,7 +125,7 @@ block content
 
     <hr id="arrow-functions" />
 
-    **Q**. I'm using an arrow function for a [virtual](./guide.html#virtuals), getter/setter, or [method](./guide.html#methods) and the value of `this` is wrong.
+    **Q**. I'm using an arrow function for a [virtual](./guide.html#virtuals), middleware (hooks), getter/setter, or [method](./guide.html#methods) and the value of `this` is wrong.
```
vkarpov15: Would be good to link to the [middleware docs](http://mongoosejs.com/docs/middleware.html) here

PR #6346, opened 2018-04-14
docs: fix Model.aggregate()

closingin:
Wrong method was used for projection with the aggregation pipeline builder.

**Summary**

I assume that the documentation was wrongly updated after the new 5.x version and that an old way to do projection with the aggregation pipeline builder by using .select() was changed. I didn't check the changelog.

**Test plan**

It fixes this error : `TypeError: Model.aggregate(...).select is not a function` that you have if you just follow the documentation.


vkarpov15:
Good catch, thanks, but please make these changes in `lib/aggregate.js` next time. `api.html` is compiled code. I made the changes this time. `.select()` never worked for aggregation as far as I know, so that syntax has been wrong for a while.

closingin:
My bad, I will do it correctly next time.

Issue #6362, opened 2018-04-17
First-class support for sessions


vkarpov15:

thaoula:
Hi @vkarpov15,

Thanks for your work on this!

When are you aiming to release a test version. Mongo DB 4 beta has just been released and we would love to start playing around with transactions.

Regards,
Tarek

vkarpov15:
@thaoula we normally don't do formal beta programs with mongoose, but we may for this release if there's sufficient demand for it. Is this blocking you from trying out transactions?

thaoula:
Hi @vkarpov15,  yeah, it is a little blocking.

thaoula:
Hi @vkarpov15,

Is this now available on 5.1?

Many thanks for this.

Regards,
Tarek

vkarpov15:
@thaoula yes it is in 5.1. We need to add better docs, but the general idea is that `findOne({}, null, { session })` means that future `save()` calls on that document will use the same session, and `doc.$session()` lets you get/set the session associated with a document

vkarpov15:
See:

http://mongoosejs.com/docs/api.html#document_Document-$session

http://mongoosejs.com/docs/api.html#query_Query-setOptions

PR #6363, opened 2018-04-18
Session support for docs

vkarpov15:
Re: #6362 

There's two primary changes here:

1) Models now have a `startSession()` function that returns a promise that resolves to the underlying `ClientSession`. You can pass the `session` option to `Model.find()`, `Model.updateOne()`, etc.

2) Mongoose documents now have a `$session()` function that tracks the session associated with the document. If you load a document using `Model.findOne({}, null, { session })`, that document will track `session` and use that same session when you `save()` the document by default.

@mbroadst @daprahamian any thoughts you have would be much appreciated.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -141,7 +141,8 @@ Model.prototype.$__handleSave = function(options, callback) {
     options.safe = null;
   }
   let safe = options.safe ? utils.clone(options.safe) : options.safe;
-  const session = this.$session();
+
+  const session = 'session' in options ? options.session : this.$session();
   if (session != null) {
     safe = safe || {};
     safe.session = session;
--- a/lib/model.js
+++ b/lib/model.js
@@ -130,17 +130,22 @@ Model.prototype.baseModelName;
  */
 
 Model.prototype.$__handleSave = function(options, callback) {
-  var _this = this;
-  var i;
-  var keys;
-  var len;
+  const _this = this;
+  let i;
+  let keys;
+  let len;
   if (!options.safe && this.schema.options.safe) {
     options.safe = this.schema.options.safe;
   }
   if (typeof options.safe === 'boolean') {
     options.safe = null;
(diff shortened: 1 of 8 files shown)


lib/model.js
```diff
@@ -2122,6 +2128,41 @@ Model.watch = function(pipeline, options) {
   return new ChangeStream(this, pipeline, options);
 };
 
+/**
+ * _Requires MongoDB >= 3.6.0._ Starts a [MongoDB session](https://docs.mongodb.com/manual/release-notes/3.6/#client-sessions)
+ * for causal consistency in reads.
```
mbroadst: I think probably better to leave it up to users to read up on why session are beneficial. It's not just for causal, but also retryable writes and eventually transactions (likely other features).

lib/model.js
```diff
@@ -2122,6 +2128,41 @@ Model.watch = function(pipeline, options) {
...
+ *
+ * This function does not trigger any middleware.
+ *
+ * ####Example:
+ *
+ *     // Need to set `causalConsistency: true` to turn on causal consistency
```
mbroadst: `causalConsistency` is `true` by default in sessions

test/model.test.js
```diff
@@ -4821,7 +4821,7 @@ describe('Model', function() {
         });
       });
 
-      it('watch() (gh-5964)', function() {
+      it.skip('watch() (gh-5964)', function() {
```
mbroadst: intentionally skipped?

vkarpov15: Yes for now. We don't test against replica sets on travis currently, that's a separate concern, so I added this skip in https://github.com/Automattic/mongoose/pull/6363/commits/b44367885fa4168db58b09f4d76b4a08dc0abdbd

mbroadst: :+1:

PR #6374, opened 2018-04-20
sort api methods based on their string property

lineus:
The list of links for each sub-section of the api docs is easier to scan through visually when they are in 'alphabetical order'. *where alphabetical in this case is really unicode point order*.  

This change adds a call to `sort()` on the props array based on the value of the string property. 

**This not only reorders the list of links, but also the order that the documentation appears within each sub-section.** 

### <- Before, After ->
screen shot 2018-04-20 at 1 12 28 pm

### DIFF
--- a/docs/source/api.js
+++ b/docs/source/api.js
@@ -112,6 +112,14 @@ function parse() {
       data.props.push(ctx);
     }
 
+    data.props.sort(function(a, b) {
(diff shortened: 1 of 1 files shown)


vkarpov15:
Brilliant, this should make it much easier to find functions, right now the order is pretty random.

PR #6199, opened 2018-03-03
Updating from vulnerable version of bson

Jaace:
**Summary**
Update to latest version of `bson` to avoid vulnerabilities, highlighted by Snyk.

```
Regular Expression Denial of Service (ReDoS)
Vulnerable module: bson
Introduced through: bson@1.0.5 and mongodb@3.0.3
Detailed paths
Introduced through: mongoose@Automattic/mongoose#74d33e19f914ed6585f3c318d983866ba6a5df2e › bson@1.0.5
Remediation: Upgrade to bson@2.0.0.
```
https://snyk.io/test/github/Automattic/mongoose?severity=high&severity=medium&severity=low

PR #6423, opened 2018-05-05
chore: update bson

ChristianMurphy:
**Summary**

Ensures bson before version 1.0.5 is out of version range (https://snyk.io/vuln/npm:bson:20180225).
Closes #6199

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,19 +11,19 @@
       "dev": true
     },
     "accepts": {
-      "version": "1.3.4",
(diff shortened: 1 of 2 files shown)


vkarpov15:
Unfortunately we can't upgrade nsp because it doesn't work on nodejs 4.x or 5.x:

image

https://travis-ci.org/Automattic/mongoose/jobs/375328983

Can we make this change without upgrading nsp?

ChristianMurphy:
@vkarpov15 can do, updated.

PR #6444, opened 2018-05-09
fix gh-6443

### LINKED ISSUE #6443: cast string representation of unix timestamp into date object
I am working on an express app that has an endpoint which takes a query, and passes it directly to the mongoose model for find, and such.

The way we use it is this:

```js
let query = req.query;

let cars = await Car.find(query).limit(30);
```

So, it's been all good until we tried to pass timestamps into that query, which is converted to a string by default. It would be awesome if mongoose supports casting string representation of unix timestamps.

reproduction script:
```js
const expect = require('chai').expect;
const { Schema } = mongoose;

mongoose.connect('mongodb://localhost:27017/test');

const userSchema = new Schema({ name: String }, { timestamps: true });

const john = new User({ name: 'John' });

let unixTimestamp = Date.now();
let stringUnixTimestamp = unixTimestamp.toString();

const cleanUp = async () => await User.remove();

const run = async () => {
  await cleanUp();
  await john.save();

  let firstFind = await User.findOne({ createdAt: { $gte: unixTimestamp } });
  expect(firstFind).to.be.ok;

  let secondFind = await User.findOne({ createdAt: { $gte: stringUnixTimestamp } }); // throws an error
  expect(secondFind).to.be.ok;
};

run().catch(console.error);
```

Perhaps one way to achieve this is:

```js
if (date === Number(date)) date = Number(date);
```

I'd love to contribute on this if possible, maybe someone point me in the right direction?

MongoDB v3.2.7
mongoose v5.0.17
node v8.9.3

AbdelrahmanHafez:
**Summary**

fix #6443

### DIFF
--- a/lib/schema/date.js
+++ b/lib/schema/date.js
@@ -232,7 +232,7 @@ SchemaDate.prototype.cast = function(value) {
 
   if (value instanceof Number || typeof value === 'number') {
     date = new Date(value);
-  } else if (typeof value === 'string' && !isNaN(Number(value)) && (Number(value) >= 275761 || Number(value) <= -271821)) {
+  } else if (typeof value === 'string' && !isNaN(Number(value)) && (Number(value) >= 275761 || Number(value) < 0)) {
     // string representation of milliseconds take this path
     date = new Date(Number(value));
   } else if (typeof value.valueOf === 'function') {
--- a/lib/schema/date.js
+++ b/lib/schema/date.js
@@ -232,12 +232,12 @@ SchemaDate.prototype.cast = function(value) {
 
   if (value instanceof Number || typeof value === 'number') {
     date = new Date(value);
-  } else if (typeof value === 'string' && !isNaN(Number(value)) && (Number(value) >= 275761 || Number(value) < 0)) {
+  } else if (typeof value === 'string' && !isNaN(Number(value)) && (Number(value) >= 275761 || Number(value) < -271820)) {
     // string representation of milliseconds take this path
     date = new Date(Number(value));
   } else if (typeof value.valueOf === 'function') {
-    // support for moment.js. This is also the path string representation of years
-    // will take because strings have a `valueOf()`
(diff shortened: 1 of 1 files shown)


lib/schema/date.js
```diff
@@ -232,9 +232,12 @@ SchemaDate.prototype.cast = function(value) {
 
   if (value instanceof Number || typeof value === 'number') {
     date = new Date(value);
+  } else if (typeof value === 'string' && !isNaN(Number(value)) && (Number(value) >= 275761 || Number(value) < 0)) {
```
vkarpov15: I'm not 100% sold on this idea but I like it in general. Nice compromise between using the 'Date' constructor where possible and falling back to the pre #5880 behavior when it makes sense. However, instead of `Number(value) < 0`, let's do `Number(value) < MIN_YEAR` because `new Date('-2017')` is perfectly valid in JS and I don't want to break the date behavior.

lib/schema/date.js
```diff
@@ -232,9 +232,12 @@ SchemaDate.prototype.cast = function(value) {
...
+    // string representation of milliseconds take this path
+    date = new Date(Number(value));
   } else if (typeof value.valueOf === 'function') {
-    // support for moment.js. This is also the path strings will take because strings
-    // have a `valueOf()`
+    // support for moment.js. This is also the path string representation of years
```
vkarpov15: Undo your changes to this comment please, because this is also the path that strings like `2017-06-01` take

lib/schema/date.js
AbdelrahmanHafez: I agree, it could cause confusion to someone who's trying to do something with the first 5 minutes of 1970.

lib/schema/date.js
AbdelrahmanHafez: done

lib/schema/date.js
AbdelrahmanHafez: done

Issue #6439, opened 2018-05-09
[BUG] $pull + $in remove all embedded documents in an array type field.


FrostyLeaf:
**Node Version**: 8.9.4

**Mongoose Version**: 5.0.17; 4.13.12

**Mongodb Version**: 3.2.9

This is how to reproduce this bug:

```js
const mongoose = require('mongoose');

mongoose.connect('mongodb://localhost/test', {debug: true});
mongoose.set('debug', true);

const Cat = mongoose.model('Cat', {
  name: String,
  props: [
    new mongoose.Schema(
      {
        name: String
      },
      {_id: false}
    )
  ]
});

(async function () {
  let kitty = new Cat({name: 'Zildjian',
    props: [
      {name: 'invalid'},
      {name: 'abc'},
      {name: 'def'}
    ]
  });
  await kitty.save();

  kitty = await Cat.findById(kitty._id);
  console.log(kitty.props);

  // The setter is casted as `{ '$pull': { props: {} } }`, which leads the all embedded documents of `props` field to be removed.
  await Cat.update({_id: kitty._id}, {$pull: {props: {$in: [{name: 'invalid'}, {name: 'def'}]}}});

  kitty = await Cat.findById(kitty._id);
  console.log('result:', kitty.props); // Expect an array with 1 embedded document inside, but get 0 here.
}());

```

**Expected Behavior**: `$pull` + `$in` should only remove the embedded documents match the assigned conditions, not all of them.

FrostyLeaf:
Any one help? @vkarpov15

lineus:
@FrostyLeaf thanks for the concise report!

FrostyLeaf:
So, when v5.0.18 will be published, please? And will you guys fix this on v4?

vkarpov15:
Monday is when we are planning on releasing this for 5.x. We will backport this to 4.x as well

lineus:
My bad, I meant to reopen this when I saw the request for v4, I must have archived the email unintentionally. PR submitted.

vkarpov15:
Fixed by #6467, will release 4.13.13 tomorrow

PR #6467, opened 2018-05-13
add fix for #6439 to 4.x

lineus:
fix for #6439.

added co to query.test.js and copied the same test from v5. No existing tests broke with this change and made new test fail, then pass.


vkarpov15:
Thanks but we can't use co in the mongoose 4.x tests, because co doesn't work on nodejs 0.10 or 0.12.

Issue #6461, opened 2018-05-11
insertMany with array of models not array of objects


d1manson:
I'm on `4.13.12`, but I think this applies forwards and backwards.

It seems you can call `insertMany` with either an array of plain objects or an array of mongoose models, however, if you provide mongoose models they are cloned inside the function with [the line](https://github.com/Automattic/mongoose/blob/master/lib/model.js#L2361):

```javascript
doc = new _this(doc)
```

This means that setting `isNew=false` further down inside the function gets applied to the clone rather than the original, and thus future `.save()`s of the original models are inserts not updates - this relates to #4099.

I propose to fix it by using something like the following (I don't know if there is a mongoose-ish way of doing the instance of check):

```javascript
 if(!(doc instanceof _this)){
     doc = new _this(doc);        
 }
```

vkarpov15:
That should work, would you like to put in a PR?

vkarpov15:
Closed by #6479

PR #6479, opened 2018-05-17
fix(model): skip redundant clone in insertMany

d1manson:
**Summary**
As discussed in [issue #6461](https://github.com/Automattic/mongoose/issues/6461#issuecomment-389696817).


lib/model.js
```diff
@@ -2365,7 +2365,7 @@ Model.$__insertMany = function(arr, options, callback) {
   var validationErrors = [];
   arr.forEach(function(doc) {
     toExecute.push(function(callback) {
-      if(!doc instanceof _this){
+      if (!(doc instanceof _this)) {
```
d1manson: thanks...that was pretty dumb.

PR #6516, opened 2018-05-24
Fix using of @memberOf tag in JSDoc

treble-snake:
**Summary**
There are some usages of `@memberOf` tag in project JSDocs.
According to http://usejsdoc.org/tags-memberof.html `@memberOf` tag by default tags symbols as static members - and default way is used everywhere, which is wrong.
To tag instance members one of three ways should be used. I used adding '#' after class name as one of the most reasonable and supported one.

**Motivation**: first of all, JSDocs should be correct for general reasons; then there are autocomplete issues in code editors due to wrong JSDocs.

**Test plan**
None required.


vkarpov15:
My hesitation with merging this is that we use this JSDoc tag to generate our docs, so this will break our docs build. Could we do the `@instance` approach instead?

treble-snake:
@vkarpov15 Yeah, I see now. 
I sure can change the PR using `@instane` tag (and will do).

But it appears that doc generating logic works in a bit incorrect way - it takes what is described as a static member and puts it in the docs as an instance member.
On the other hand, maybe this doesn't worth the effort to fix.

vkarpov15:
Yeah let's go with the `@instance` approach. The doc generation logic isn't strictly correct, but it works because mongoose's API doesn't really have static functions.

treble-snake:
@vkarpov15 I've update the PR. Docs generation seems working fine.

PR #6563, opened 2018-06-06
docs - add string option to projections that call query select

lineus:
updating projection parameter on find, findOne, and findById to reflect the usage of query select.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1364,7 +1364,7 @@ Model.deleteMany = function deleteMany(conditions, callback) {
  *     promise.addBack(function (err, docs) {});
  *
  * @param {Object} conditions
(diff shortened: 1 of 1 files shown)


lib/model.js
```diff
@@ -3028,6 +3028,9 @@ Model.aggregate = function aggregate(pipeline, callback) {
   return aggregate;
 };
 
+Model.popregate = function popugrate() {
```
vkarpov15: Is this here intentionally or is this a typo?

lineus: In my spare time I have working on trying to make a version of populate that uses aggregate. That lame attempt at a name for what I've been working on should have been in another branch. My bad. I'll remove it and re-commit.

PR #6577, opened 2018-06-11
Minimize docs about $unset update validation

joeytwiddle:
The documentation about the caveats when using update validators (`runValidators`) is excellent.

But I found the section on `required` fields to be overkill, to the point of confusion.

It seems obvious to me that validation will not run on a field which we are not updating. In fact, that is covered by the next section, "Update Validators Only Run On Specified Paths".

So for the sake of brevity, I suggest removing that section. But I have preserved a paragraph that lets developers know that `$unset` does perform validation on required fields.

### DIFF
--- a/test/docs/validation.test.js
+++ b/test/docs/validation.test.js
@@ -499,7 +499,7 @@ describe('validation docs', function() {
    * you try to explicitly `$unset` the key.
    */
 
-  it('Update Validator Paths', function(done) {
+  it('Update Validators Only Run On Updated Paths', function(done) {
     // acquit:ignore:start
     var outstanding = 2;
     // acquit:ignore:end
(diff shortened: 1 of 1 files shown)


joeytwiddle:
Ah now I see why the examples are so long, because they are actually full tests!

I have dropped my earlier commits, and just reworked the titles, which will hopefully make the content easier to absorb.

I also split up two concepts into two paragraphs. Cheers VK!

joeytwiddle:
The length of the examples still puts me off a bit, but changing that would be a bigger task.

I do have a couple of suggestions though:

- Convert the tests to use async-await. This will slightly reduce the complexity of the code.
- Use a stronger and more comprehensive syntax highlighter.
   Maybe it's just me, but I find more colours and stronger colours make the code easier to read.

Here is an example I knocked up using CodeMirror.net and RailsCast theme:

screenshot 2018-06-25 17 01 04

vkarpov15:
Re: [async/await](http://thecodebarbarian.com/common-async-await-design-patterns-in-node.js.html), we won't use that for now because we still intend to support nodejs v4 and v6. We can use [co+yield](http://npmjs.com/package/co) though to make our examples more concise, would that work?

joeytwiddle:
It doesn't look like `// acquit:ignore:start` and `end` are being stripped today: http://mongoosejs.com/docs/validation.html

I accept your point about `async-await`. Co-routines are cool, but I don't think they are very widely used. So probably clearest just to stick with callbacks.

It looks like mongoose docs are using highlight.js, which does have [a number of themes](https://highlightjs.org/static/demo/) available. Personally I would recommend one of these (in order of preference) because they are quite widely recognised:
- Atom One Dark (the default for the Atom editor, clear primary colours)
- Monokai Sublime (familiar to users of TextMate and Sublime Text editors)
- Railscasts (the theme in the screenshot above, but doesn't do anything special with `undefined` or `true`)

vkarpov15:
Thanks, I opened up an issue. We can switch to using async/await in examples if we make it so that certain tests only run on certain versions of node.

PR #6600, opened 2018-06-16
5.2.0-rc0: Use mongodb driver 3.1.0-beta4

vkarpov15:


vkarpov15:
@lineus thanks but not at the moment. This is mostly ready to go.

felipe-augusto:
@vkarpov15 Do you have an ETA for this mongoose version? I'm deciding between MongoDB and PostgreSQL and really need support for transactions. And I'm almost on the deadline for deciding.

Also, I'm using this branch and you are probably aware of this, but the `Model.create` function is not working with transactions.

jesster2k10:
+1 @felipe-augusto Any ETA on when this would be finalized?

felipe-augusto:
Also, I'm having a problem testing transactions because I need a container with replicaSet enabled. And I have no idea how to do this on Gitlab, since I'm using their services keyword

```yml
services:
  - docker:dind
  - mongo:latest
```

How did you managed to avoid this problem while testing? Thanks!!

vkarpov15:
@felipe-augusto if you're using linux or OSX check out [run-rs](https://www.npmjs.com/package/run-rs), it's a command line tool that'll run a replica set for you. That's what I use for local dev.

Issue #6635, opened 2018-06-29
Unhandled Exception: `TypeError: doc.toObject is not a function`


dbellavista:
**BUG**

When editing an existing document with an array. If I add a new element inside the array, but the `.save()` fails for another reason, then any consecutive fail will throw an unhandled exception:

```
/project/node_modules/mongoose/lib/types/documentarray.js:235
      return doc && doc.toObject(options) || null;
                        ^

TypeError: doc.toObject is not a function
    at /project/node_modules/mongoose/lib/types/documentarray.js:235:25
    at Array.map (<anonymous>)
    at Array.toObject (/project/node_modules/mongoose/lib/types/documentarray.js:234:17)
    at CoreMongooseArray.$__getAtomics (/project/node_modules/mongoose/lib/types/array.js:261:35)
    at handleAtomics (/project/node_modules/mongoose/lib/model.js:480:11)
    at model.Model.$__delta (/project/node_modules/mongoose/lib/model.js:601:7)
    at model.Model.$__handleSave (/project/node_modules/mongoose/lib/model.js:197:24)
    at model.Model.$__save (/project/node_modules/mongoose/lib/model.js:243:8)
    at /project/node_modules/kareem/index.js:278:20
    at _next (/project/node_modules/kareem/index.js:102:16)
    at process.nextTick (/project/node_modules/kareem/index.js:452:38)
    at process._tickCallback (internal/process/next_tick.js:61:11)
```

This gist will reproduce the error: https://gist.github.com/dbellavista/46df766acc09695f448be339ca96632e

**Please mention your node.js, mongoose and MongoDB version.**

Node 10 and 8.11, mongoose 5.1.6, mongoDB 3.4 and 3.6.

evanhenke:
@dbellavista this should be fixed in 5.1.8, can you update and try again?

dbellavista:
@evanhenke I tried with 5.2.0 and it works. However I had to modify the connection string due to #6647

vkarpov15:
Glad it works :+1:

evanhenke:
Sounds good!  Can you close the issue when you get the chance?

bornagainraj7:
I'm getting the same error even in 5.5.2

vkarpov15:
@bornagainraj7 please open a separate issue with code samples and follow the issue template.

PR #6638, opened 2018-06-29
fix for TypeError being thrown when calling save() more than once if the object being saved has atomics

evanhenke:
**Summary**

fix for issue #6635 , if an object had atomics and calling save() on that object failed for some reason, the atomics array of EmbeddedDocuments would already have been converted to POJOs.  So if you tried to save it again it would try to call .toObject() on the POJO which would throw a TypeError.

**Test plan**

Reran @dbellavista gist located here: https://gist.github.com/dbellavista/46df766acc09695f448be339ca96632e

Wrote a new test: 'no TypeError when attempting to save more than once after using atomics':
```
it('no TypeError when attempting to save more than once after using atomics', function(done) {
      const M = db.model('M', new Schema({
        test: {type: 'string', unique: true},
        elements: [{
          el: {type: 'string', required: true}
        }]
      }));
      const a = new M({
        test: 'a',
        elements: [{el: 'b'}]
      });
      const b = new M({
        test: 'b',
        elements: [{el: 'c'}]
      });
      a.save(function() {
        b.save(function() {
          b.elements.push({el: 'd'});
          b.test = 'a';
          b.save(function(error,res) {
            assert.strictEqual(!error,false);
            assert.strictEqual(res,undefined);
            b.save(function(error,res) {
              assert.strictEqual(!error,false);
              assert.strictEqual(res,undefined);
              done();
            });
          });
        });
      });
    });
```
Built it with Travis CI on my fork. Had a couple intermittent failures, mostly for timeouts, but they passed when run locally and on a rebuild. Link to build status: https://travis-ci.org/evanhenke/mongoose/

### DIFF
--- a/lib/types/documentarray.js
+++ b/lib/types/documentarray.js
@@ -248,7 +248,11 @@ MongooseDocumentArray.mixin = {
 
   toObject: function(options) {
     return this.map(function(doc) {
-      return doc && doc.toObject(options) || null;
+      try {
+        return doc.toObject(options);
+      } catch (e) {
+        return doc || null;
+      }
     });
   },
 
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -3729,6 +3729,38 @@ describe('Model', function() {
         done();
       });
     });
+
+    it('no TypeError when attempting to save more than once after using atomics', function(done) {
+      const M = db.model('M', new Schema({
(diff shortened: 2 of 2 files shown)


evanhenke:
Hmm seems like a couple are failing due to more than just a timeout, I'll take a peak.

PR #6719, opened 2018-07-12
update doc about keepalive. closes #6718

### LINKED ISSUE #6718: KeepAlive mongodb-core 3.1 changes
This is a wrong/missing documentation "bug":

In mongodb-native they have changed the "keepalive" paramtere. it is now a boolean, and keepAliveInitialDelay is the time to wait for a second.

This means the documentation here: http://mongoosejs.com/docs/connections.html is wrong. And should be cleaned up anyway regarding keepalive. There is a mix up of booleans and numbers.

See also http://mongodb.github.io/node-mongodb-native/3.1/api/MongoClient.html
and change  notes:
### `keepAlive`

Wherever it occurs, the option `keepAlive` has been changed. `keepAlive` is now a boolean that enables/disables `keepAlive`, while `keepAliveInitialDelay` specifies how long to wait before initiating keepAlive. This brings the API in line with [NodeJS's socket api](https://nodejs.org/dist/latest-v9.x/docs/api/all.html#net_socket_setkeepalive_enable_initialdelay)

keepAlive | boolean | true | optionalTCP Connection keep alive enabled
-- | -- | -- | --
keepAliveInitialDelay | number | 30000 | optionalThe number of milliseconds to wait before initiating keepAlive on the TCP socket

simllll:


vkarpov15:
Thanks!

PR #6744, opened 2018-07-19
fixes #6742 add 12 char objectid to FAQ

### LINKED ISSUE #6742: Mongoose "converts" string to ObjectId for properties of type ObjectId
Bug

When sending a string value that is 12 characters long, the "ObjectId" validation seems to be bypassed. We have tried with "Select Type:" and "Select Typ18" (not that they should be relevant I think), and these values are "saved| as ObjectId's inside the schema. The ObjectId's are actually random, even though that actual property refers to an external id from another schema.

I think the above description is explanatory.

An error of "Cast to ObjectID failed for value ..." should be thrown.

Node: 6.9.1
Mongoose: 4.13.11
MongoDB: 3.4

lineus:
I've seen this question come up a couple of times recently in the slack channel, and issues like #6742.

### DIFF
--- a/docs/faq.jade
+++ b/docs/faq.jade
@@ -362,7 +362,7 @@ block content
     `save()` multiple times in parallel on the same doc could result in conflicts. For example,
     validating, and then subsequently invalidating the same path.
 
-    <hr id="ObjectId Validation" />
+    <hr id="objectid-validation" />
 
     **Q**. Why is **any** 12 character string successfully cast to an ObjectId?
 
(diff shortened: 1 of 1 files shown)


docs/faq.jade
```diff
@@ -362,6 +362,13 @@ block content
     `save()` multiple times in parallel on the same doc could result in conflicts. For example,
     validating, and then subsequently invalidating the same path.
 
+    <hr id="ObjectId Validation" />
```
vkarpov15: Make this id include a `-`, spaces in urls are confusing

PR #6751, opened 2018-07-22
Prefix mongoose.Schema call with new operator

sajadtorkamani:
**Summary**

Whilst the existing code works, I thought maybe since `Schema` is a constructor, it makes sense to use the `new` operator. It also seems more consistent with other examples in the documentation :)

**Test Plan**
I don't think this requires any tests ?


sajadtorkamani:
Yay! thank you :)

Issue #6756, opened 2018-07-22
TypeError: Cannot use 'in' operator to search for 'promiseLibrary' in myproj


denizdogan:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

```
/Users/deniz/Code/myproj/node_modules/mongoose/lib/connection.js:463
  if (!('promiseLibrary' in options)) {
                         ^

TypeError: Cannot use 'in' operator to search for 'promiseLibrary' in myproj
    at NativeConnection.Connection.openUri (/Users/deniz/Code/myproj/node_modules/mongoose/lib/connection.js:463:26)
    at Mongoose.createConnection (/Users/deniz/Code/myproj/node_modules/mongoose/lib/index.js:189:17)
    at Object.<anonymous> (/Users/deniz/Code/myproj/dist/index.js:19:36)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:236:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:560:3)
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```
import mongoose from 'mongoose'

const connection = mongoose.createConnection(
  process.env.MONGODB_HOST,
  process.env.MONGODB_DATABASE,
  process.env.MONGODB_PORT
)

export default connection
```

...or transpiled...

```
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.default = void 0;

var _mongoose = _interopRequireDefault(require("mongoose"));

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

var connection = _mongoose.default.createConnection(process.env.MONGODB_HOST, process.env.MONGODB_DATABASE, process.env.MONGODB_PORT);

var _default = connection;
exports.default = _default;
```

I've verified that the values in `process.env` are valid.

**What is the expected behavior?**

Expected is no error or better error message

**Please mention your node.js, mongoose and MongoDB version.**

- Node 10.6.0
- mongoose 5.2.4
[...]

PR #6766, opened 2018-07-24
update jsdocs for createConnection

lineus:
per my comment on #6756 this change removes the old parameter signature from the createConnection jsdoc example. 

I'm also thinking we could change line 420 in connection.js from
```
  if (options) {
```
to
```
  if (options && typeof options === 'object') {
```
but I'm not sure if we should throw an error instead if options isn't an object? @vkarpov15, what do you think?

all tests pass.


vkarpov15:
We should throw an error if `options` isn't an object. Cleaner developer experience that way because people might stumble across the old syntax like the poster of #6756 did. Feel free to put in a PR for that.

Issue #6705, opened 2018-07-10
Webpack warnings when running Mongoose for browser 


ianpaschal:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

When using the browser version of Mongoose as described [here](http://mongoosejs.com/docs/browser.html), Webpack generates several warnings:

```bash
WARNING in ./node_modules/mongoose/lib/drivers/index.js
10:13-49 Critical dependency: the request of a dependency is an expression
 @ ./node_modules/mongoose/lib/drivers/index.js
 @ ./node_modules/mongoose/lib/schema.js
 @ ./node_modules/mongoose/lib/browser.js
 @ ./node_modules/mongoose/browser.js
 @ ./src/store/index.js
 @ ./src/client.js
 @ multi (webpack)-dev-server/client?http://localhost:8080 webpack/hot/dev-server ./src/client.js

WARNING in ./node_modules/require_optional/index.js
82:18-42 Critical dependency: the request of a dependency is an expression
 @ ./node_modules/require_optional/index.js
 @ ./node_modules/mongodb-core/index.js
 @ ./node_modules/mongodb/index.js
 @ ./node_modules/mongoose/lib/drivers/node-mongodb-native/binary.js
 @ ./node_modules/mongoose/lib/drivers/node-mongodb-native/index.js
 @ ./node_modules/mongoose/lib/drivers/index.js
 @ ./node_modules/mongoose/lib/schema.js
 @ ./node_modules/mongoose/lib/browser.js
 @ ./node_modules/mongoose/browser.js
 @ ./src/store/index.js
 @ ./src/client.js
 @ multi (webpack)-dev-server/client?http://localhost:8080 webpack/hot/dev-server ./src/client.js

WARNING in ./node_modules/require_optional/index.js
90:20-44 Critical dependency: the request of a dependency is an expression
 @ ./node_modules/require_optional/index.js
 @ ./node_modules/mongodb-core/index.js
 @ ./node_modules/mongodb/index.js
 @ ./node_modules/mongoose/lib/drivers/node-mongodb-native/binary.js
 @ ./node_modules/mongoose/lib/drivers/node-mongodb-native/index.js
 @ ./node_modules/mongoose/lib/drivers/index.js
 @ ./node_modules/mongoose/lib/schema.js
 @ ./node_modules/mongoose/lib/browser.js
 @ ./node_modules/mongoose/browser.js
 @ ./src/store/index.js
 @ ./src/client.js
[...]

PR #6771, opened 2018-07-25
refactor: use `global.$MongooseDriver` to store current driver to remove webpack warnings

vkarpov15:
This refactor is just a bit too meaty for me to merge into master directly, I'd like to review it a couple more times.

The problem with #6705 is that we use `global.MONGOOSE_DRIVER_PATH` all over the place in `require()` statements to support the existing custom driver paradigm, which is what makes the browser library possible. This PR changes it so that we look at `global.MONGOOSE_DRIVER_PATH` once when mongoose is first `require()`-ed in node and store it in a `global.$MongooseDriver` variable. The `lib/browser.js` entry point won't look at `global.MONGOOSE_DRIVER_PATH` at all.

The hard part is that this added the potential for messy cycling dependencies, because `require('./driver')` -> `require('./collection')` -> `require('./utils')` -> `require('./types/Decimal128')` -> `require('./driver')` and the alternative is to use `global.$MongooseDriver` in transactional code, which sounds gross.


vkarpov15:
My mistake, this PR is supposed to reference #6705

PR #6810, opened 2018-08-04
feat(model): support options in `deleteMany`

### LINKED ISSUE #6805: Why deleteMany doesn't support session
Assuming this function :

```
  const session = await mongoose.startSession()
  session.startTransaction();
  try{
    let deleted;
    if((deleted = await node.findByIdAndRemove(id, { session })) == null) throw Error("L'objet node ne peut pas être supprimé")
    await Capteur.deleteMany({_id : {$in : deleted.children}}, { session })
    await node.update({}, {$pull : {children : id}}, { session, multi : true})
    await session.commitTransaction();
    session.endSession();
  }catch(err){
      await session.abortTransaction();
      session.endSession()
      throw Error("Erreur lors de l'ajout d'un node")
  }
  return
```

`await Capteur.deleteMany({_id : {$in : deleted.children}}, { session })` return this MongooseError "Callback must be a function, got [object Object]"

Why can't we use session there as we can use it for Model.update() or Model.findByIdAndRemove() ?

Fonger:
**Summary**

1. `insertMany` and `updateMany` support options parameter, so I think `deleteMany` should support too to avoid confusion
2. This will also fix #6805

**Test plan**

Tests are included in the `test/model.test.js` and `test/docs/transaction.test.js`

```sh
╭─Fonger@Fongers-Macbook  ~/opensource/mongoose  ‹delete-many-options›
╰─$ mocha test/docs/transactions.test.js                                    1 ↵

(node:31830) DeprecationWarning: current URL string parser is deprecated, and will be removed in a future version. To use the new parser, pass option { useNewUrlParser: true } to MongoClient.connect.

  ․․․․․․․․

  8 passing (1s)
```

Locally passed

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1480,19 +1480,26 @@ Model.deleteOne = function deleteOne(conditions, callback) {
  * Like `Model.remove()`, this function does **not** trigger `pre('remove')` or `post('remove')` hooks.
  *
  * @param {Object} conditions
+ * @param {Object} [options] optional see [`Query.prototype.setOptions()`](http://mongoosejs.com/docs/api.html#query_Query-setOptions)
  * @param {Function} [callback]
  * @return {Query}
  * @api public
  */
 
-Model.deleteMany = function deleteMany(conditions, callback) {
+Model.deleteMany = function deleteMany(conditions, options, callback) {
   if (typeof conditions === 'function') {
     callback = conditions;
     conditions = {};
+    options = null;
+  }
+  else if (typeof options === 'function') {
+    callback = options;
+    options = null;
   }
 
   // get the mongodb collection object
   var mq = new this.Query(conditions, {}, this, this.collection);
+  mq.setOptions(options);
 
   if (callback) {
     callback = this.$wrapCallback(callback);
--- a/test/docs/transactions.test.js
+++ b/test/docs/transactions.test.js
@@ -226,4 +226,28 @@ describe('transactions', function() {
         then(article => assert.ok(!article.author));
     });
   });
+
+  it('deleteMany (gh-6805)', function() {
+    const Character = db.model('Character', new Schema({ name: String }), 'Character');
+
+    let session = null;
+    return db.createCollection('Character').
+      then(() => db.startSession()).
+      then(_session => {
+        session = _session;
+        session.startTransaction();
+        return Character.insertMany([
+          { name: 'Tyrion Lannister' },
+          { name: 'Cersei Lannister' },
+          { name: 'Jon Snow' },
+          { name: 'Daenerys Targaryen' }
+        ], { session: session });
+      }).
+      then(() => Character.deleteMany({ name: /Lannister/ }, { session: session })).
+      then(() => Character.find({}).session(session))
+      .then(res => {
+        assert.equal(res.length, 2);
+        session.commitTransaction();
+      });
+  });
 });
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -4823,7 +4823,32 @@ describe('Model', function() {
         assert.equal(postCalled, 0);
         assert.equal(postErrorCalled, 1);
       });
+    });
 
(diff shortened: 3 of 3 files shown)


krlozadan:
Hello I know this has been closed for some time but I'm trying to run a session with Model.deleteMany and it is still giving me the error of `"MongooseError: Callback must be a function, got [object Object]"`
I get the error passing the session in the second and third argument.
I'm using mongoose version: 5.5.6

Fonger:
@krlozadan Hi 

I can't reproduce here. Can you provide a full reproduce code?

krlozadan:
Yeah sure, here it is @Fonger  
I'm using babel
```javascript
import ModelA from "./models/modelA"; 
import ModelB from "./models/modelB";

export default async (req, res) => {

    let clientSession;
    try {
        clientSession = await mongoose.startSession();
        clientSession.startTransaction();
        
        // This works
        const modelAToDelete = await ModelA.findOneAndRemove({ app: appId }, {}, { session: clientSession });
        if (modelAToDelete == null) throw new Error("...");

        // This doesn't work
        const modelBToDelete = await ModelB.deleteMany({ app: appId, modelA: modelAToDelete.id }, { session: clientSession });

        await clientSession.commitTransaction();
        clientSession.endSession();
    } catch (e) {
        await clientSession.abortTransaction();
        clientSession.endSession();
        return res.status(500).json({ message: e.toString() });
    }
    res.status(200).json({ message: `...` });
};
```

Fonger:
@krlozadan Sorry for the delay. I've submitted a PR to fix the issue. Can you post a basic code of your `models/modelB` ? I guess they are not Model but Query instance.

krlozadan:
Hi @Fonger , thanks for the help.
Sure one of those models look like this:
```javascript
const modelASchemaOptions = {
    timestamps: {
        createdAt: 'created_at',
        updatedAt: 'updated_at'
    },
    collection: "models_a"
};

const modelASchema = new mongoose.Schema({
    field_one: {
        type: String,
        unique: true,
        required: true
    },
    field_two: {
        type: Number,
        required: true,
        min: 0
    }
}, modelASchemaOptions);

module.exports = mongoose.models.ModelA || mongoose.model("ModelA", modelASchema);
```

krlozadan:
@Fonger I just updated to version 5.6.6 which is supposed to be fixing it but I'm still getting the same error I've gone through the code and I saw this was added for the query object but as far as I can tell in [model.js file line 1908](https://github.com/Automattic/mongoose/blob/f9c3b7ceecff57d27d4a9804168bd685ae373fe8/lib/model.js#L1908), is going in to the **$wrapCallback** method and then is throwing an error inside that function. 
The weird thing is **callback** should be undefined there right if I only provide 2 arguments.

Fonger:
@krlozadan It's really weird but I can't reproduce it. Can you post the full error stack? Also, just in case npm somehow failed to update, what is the output of `console.log(mongoose.version)` ?

krlozadan:
The version i get is **5.6.6 ** and the error stack: 
```
{ MongooseError: Callback must be a function, got [object Object] 
    at new MongooseError (D:\dev\my-api\node_modules\mongoose\lib\error\mongooseError.js:10:11)
    at Function.Model.$wrapCallback (D:\dev\my-api\node_modules\mongoose\lib\model.js:4562:11)
    at Function.Model.findOneAndRemove (D:\dev\my-api\node_modules\mongoose\lib\model.js:2858:21)
    at findOneAndRemove (D:\dev\my-api\src\lib\modelA/deleteModelA.js:12:45)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  message: 'Callback must be a function, got [object Object]',
  name: 'MongooseError' }
```
You know what... by analizyng the stack trace I realized the function that is getting the error now, instead of the `deleteMany` is the `findOneAndRemove`, before the last merged change you would need to pass the projection object to the function first and then the object with the session. But now is not the case.

Thanks a lot for your help @Fonger

PR #6821, opened 2018-08-06
Fix DEP0079 warning on Node 10

### LINKED ISSUE #6420: DEP0079 warning on Node 10
I get following error: `(node:30514) [DEP0079] DeprecationWarning: Custom inspection function on Objects via .inspect() is deprecated` when I use:

```
Games.find({ $text: { $search: T } },
    (err, docs) => {
      if (docs) {
        console.log("Games:", docs);
        deferred.resolve({
          games: docs
        });
      } else {
        deferred.reject(new Error(err.message));
      }
    }
  );
```
`T` is just a text string like "I search for some food". The error disappears when I use a $or operator with some $elemMatch cases. I really need to use full-text search, my script breaks the hole function when this error occurs.

Node: 10.0.0
mongoose: 5.0.17
Entry from games collection: https://hastebin.com/nonutobeva.json

yelworc:
**Summary**

This PR fixes #6420 by defining a `util.inspect.custom` property wherever a custom `inspect` helper is used (which seems to be the canonical solution proposed in the [Node docs](https://nodejs.org/api/util.html#util_custom_inspection_functions_on_objects)).

It does so in a backwards compatible fashion by only adding the property if `util.inspect.custom` is actually defined (Node >= 6). If you would rather make this a semver-major change (dropping Node 4 and 5 support) and avoid the conditionals everywhere, just the first commit would be sufficient.

I tried to add the property in the least intrusive way for the respective location, but am happy to change things around if necessary.

**Test plan**

Existing unit tests are green for me, no new tests were added since the PR does not add or change any functionality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ node_js:
   - "8"
   - "7"
   - "6"
-  - "5"
-  - "4"
 before_script:
   - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-3.6.4.tgz
   - tar -zxvf mongodb-linux-x86_64-3.6.4.tgz
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ node_js:
   - "8"
   - "7"
   - "6"
+  - "5"
+  - "4"
 before_script:
   - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-3.6.4.tgz
   - tar -zxvf mongodb-linux-x86_64-3.6.4.tgz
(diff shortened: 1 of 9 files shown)


yelworc:
Those CI errors don't seem to be related to my changes, as far as I can tell 🤔

yelworc:
Cool, thanks for the quick merge!

I suppose the old `inspect` methods could be dropped in a future major release, which (together with removing the Node <6 conditionals) would make the codebase a tiny bit cleaner. Not sure if it makes sense to open a PR or task for that right now, though.

Issue #6271, opened 2018-03-24
Want to $inc 1, but it's actually increasing by 2


iTonyYo:
- mongodb v3.6.3
- mongoose v5.0.11
- node v8.9.4

The strange problem is described in the comment below.

```javascript
async function newVisit(ctx) {
  let {
    shortid,
  } = ctx.params;

  let { stringify } = JSON;
  let { openSync } = maxmind;

  let cityLookup = openSync(process.env.GEODATA);

  // test ip: 219.76.153.227
  let city = cityLookup.get('219.76.153.227');

  let TblShortUrlVisiting = ctx.model('Short Url Visiting');

  let shortUrlVisitingRecord = await TblShortUrlVisiting.findOne({
    'stId': shortid,
  });

  // no visiting record
  if (isEmpty(shortUrlVisitingRecord)) {
    let toBeSave = {
      stId: shortid,
      birth: now(),
      visitDetails: [
        {
          date: now(),
          ua: stringify(ctx.userAgent),
          geo: stringify(city),
        }
      ],
    };

    let tblShortUrlVisiting = new TblShortUrlVisiting(toBeSave);

    await tblShortUrlVisiting.save();

  } else {

    // has visiting record

    await TblShortUrlVisiting.update({
      'stId': shortid,
    }, {

      /**
       * ATTENTION:
       * 
       * I have to set 0.5 to increment 1.
       * If I set 1, it will going to increment by 2.
       * OMG!
       */
      $inc: {
        visitTimes: 0.5,
      },

      // This has been pushed twice...
      $push: {
        visitDetails: {
          date: now(),
          ua: stringify(ctx.userAgent),
          geo: stringify(city),
        },
      },
    }, {
      safe: true,
      upsert: false,
    }, (err, rawResponse) => {
      if (err) {
        throw err;
      }
    });

  }
}
```

lineus:
Hi @iTonyYo have you tried setting `mongoose.set('debug', true)` ?  If that doesn't help (ie. seeing multiple .update() calls when you only expect one, etc) then please comment here with a minimal but complete example of how to reproduce the behavior you are seeing.

Here is an example of a reproducible example:
### 6271.js
```
#!/usr/bin/env node
'use strict'

const assert = require('assert')
const mongoose = require('mongoose')
mongoose.connect('mongodb://localhost/test')
const Schema = mongoose.Schema

const schema = new Schema({
  name: String,
  age: Number
})

const Person = mongoose.model('person', schema)

[...]

Issue #6016, opened 2018-01-19
$unset nested not work 


mxyue:
mongoose 4.9.5 remove embed filed not work 

`
schema = new Schema({ nested: { field1: String, field2: String }})
model.update({}, {$unset: {nested: 1}) // not work
model.update({}, {$unset: {'nested.field1': 1}) // not work
model.update({}, {$set: {nested: {} } }) // set empty object not work
model.update({}, {$set: {nested: null } }) // work

`

vkarpov15:
Below script works as intended for me in mongoose 4.9.5, can you modify it to reproduce your issue?

```javascript
const mongoose = require('mongoose');
const Schema = mongoose.Schema;

run().catch(error => console.error(error.stack));

async function run() {
  await mongoose.connect('mongodb://localhost:27017/test');

  await mongoose.connection.dropDatabase();

  const schema = new Schema({ nested: { field1: String, field2: String }});

  const Test = mongoose.model('Test', schema);

  const doc = await Test.create({ nested: { field1: 'Test', field2: 'Test' } });

  await Test.updateOne({ _id: doc._id }, { $unset: { 'nested.field1': 1 } });

  console.log(await Test.findById(doc._id));
} 
```

```
$ npm install mongoose@4.9.5

+ mongoose@4.9.5
added 15 packages in 1.319s
$ node gh-6016.js 
(node:18635) DeprecationWarning: Mongoose: mpromise (mongoose's default promise library) is deprecated, plug in your own promise library instead: http://mongoosejs.com/docs/promises.html
{ _id: 5a62aa7b13172248cb518cdc,
  __v: 0,
  nested: { field2: 'Test' } }
^C
$ 
```

mxyue:
I set default value, when I unset filed mongoose return default value. so I don't know the field removed. use lean() can return origin data

Issue #4027, opened 2016-03-30
Mongoose driver doesn't send notification when it gives up on the reconnection with mongo DB


tokunal80:
There is no good way to know that driver has given up on reconnect and the application should try opening a fresh new connection and keep retrying till it succeeds.

Please refer to thread on the issue:
https://github.com/Automattic/mongoose/issues/3969

TomerAmirV:
Any news on this issue?

PR #6824, opened 2018-08-07
Fix several flaky tests (updated)

Fonger:
# Fix test flakes to increase CI pass rate
## 1. Fix flaky test case for #4014

- Ensure `2dsphere` index is created before the query is executed
- Specify `spherical: true` to use `2dsphere` index instead of `2d` index

   I don't know why it passes the test without this patch sometimes. This appears to be a bug and should fail every time without this fix.

   <details><summary>Mocha Output</summary><p>

   ```
   1) document
        bug fixes
          single embedded docs with $near (gh-4014):
          Uncaught MongoError: error processing query:
            ns=mongoose_test.gh4014
            Tree: GEONEAR  field=geo maxdist=1.79769e+308 isNearSphere=0
      Sort: {}
      Proj: {}
      planner returned error: unable to find index for $geoNear query
        at queryCallback (node_modules/mongodb-core/lib/cursor.js:247:25)
        at node_modules/mongodb-core/lib/connection/pool.js:531:18
        at process._tickCallback (internal/process/next_tick.js:176:11)
   ```
   See build failure on travis ci - [[1]](https://travis-ci.org/Automattic/mongoose/jobs/412698211) [[2]](https://travis-ci.org/Automattic/mongoose/jobs/410873839) [[3]](https://travis-ci.org/Automattic/mongoose/jobs/410407704)

   </p></details>

## 2. Fix flaky test for #6271 (EDIT: refactored)
- By ensuring the update request with callback is executed

   <details><summary>Mocha Output</summary><p>

   ```
   1) Query
          bug fixes
            consistently return query when callback specified (gh-6271):
         AssertionError [ERR_ASSERTION]: 1 == 2
         + expected - actual
         -1
         +2
         
         at Function.decoratedAssert [as equal] (node_modules/empower-core/lib/decorate.js:51:30)
         at test/query.test.js:2457:16
         at Generator.next (<anonymous>)
         at onFulfilled (node_modules/co/index.js:65:19)
         at <anonymous>
         at process._tickCallback (internal/process/next_tick.js:188:7)
   ```

   [See build failure on travis-ci](https://travis-ci.org/Fonger/mongoose/jobs/413010578)

   </p></details>

## 3. Add delay for test case in #4027
- Make sure driver has enough time to give up reconnection and emit `reconnectFailed` event.

   <details><summary>Mocha Output</summary><p>

   ```
     1) connections:
          openUri (gh-5304)
            connection events
              reconnectFailed (gh-4027):

         AssertionError [ERR_ASSERTION]: 0 == 1
         + expected - actual

         -0
         +1

         at Function.decoratedAssert [as equal] (node_modules/empower-core/lib/decorate.js:51:30)
         at test/connection.test.js:217:20
   ```

   </p></details>

# Miscellaneous

## 1. Fix duplicate test queries introduced in #6016
- Call [Query#exec()](http://mongoosejs.com/docs/api.html#query_Query-exec) to get real Promise to prevent duplicate queries because [queries are not promises](http://mongoosejs.com/docs/queries.html#queries-are-not-promises) and we shouldn't mix callback with `then()`

## 2. Update MongoDB configuration on Travis
- Test mongodb `3.6.6` on travis-ci (original: `3.6.4`)
- Increase mocha timeout when running on travis
- Remove `--nopreallocj` parameter

   It has typo `j` and [`--noprealloc` has been deprecated since v2.6 and it's default now](https://docs.mongodb.com/manual/reference/program/mongod/#cmdoption-mongod-noprealloc).


Fonger:
I don't know why travis-ci still pull from `master` instead of this pull request.

@vkarpov15 I've updated based on your review

I've tested several times on my local development environment and my travis-ci.

**It seems that all of the flaky tests are fixed (hopefully) after this pull request except for timeout error.**

I have several ideas regarding this, what do you think?

1. increase timeout (default: 2000)?
   - increase mocha `--timeout` value (default: `2000` ms) by adding it to `test` script in `package.json`
   - increase timeout only on travis-ci by adding `echo --timeout 2500 > tests/mocha.opts` to `before_script` section of `.travis.yml` 
   - find tests that take longer, and specify `it.timeout(xxx)` individually. This doesn't work well on travis-ci because it really depends on the server status.
2. write script to make travis-ci automatically retry the test if the failure reason is timeout?

test/document.test.js
```diff
@@ -2816,12 +2820,16 @@ describe('document', function() {
 
       var MyModel = db.model('gh4014', schema);
 
-      MyModel.
-        where('geo').near({ center: [50, 50] }).
-        exec(function(error) {
-          assert.ifError(error);
-          done();
-        });
+      MyModel.on('index', function(err) {
```
vkarpov15: In general its better to use `MyModel.init().then(() => {})` instead of `MyModel.on('index')`, because with the latter if you register the event listener after the index build is finished, you'll never get a callback.

Fonger: Yes it really makes sense! There are similar test codes like this elsewhere and I will try to update it in future fix to make the test more robust.

Although index takes time to build, if we use some sort of index cache in mongoose in the future, or `MyModel.on('index')` is not placed right after `MyModel.model()` in the same event loop, that's a potential issue.

PR #6853, opened 2018-08-13
4.x ci improve

Fonger:
**Summary**

cherry-pick some fix of flaky tests from `master` while maintaining compatibility with legacy `Node.js 0.10`

also manually fix 2 additional flaky tests specific to `4.x`


test/document.test.js
```diff
@@ -2860,12 +2866,16 @@ describe('document', function() {
 
       var MyModel = db.model('gh4014', schema);
 
-      MyModel.
-        where('geo').near({ center: [50, 50] }).
-        exec(function(error) {
-          assert.ifError(error);
-          done();
-        });
+      MyModel.on('index', function(err) {
```
Fonger: should have used `init()` here but this is cherry-pick from `master` branch

However there're many other test cases using `on('index')` so this should be done in a future fix in `master`

vkarpov15: Yeah that's fair, we'll do that later. For now, I don't want to put more effort into 4.x than is necessary. We'll make the changes in master when it is convenient.

PR #6868, opened 2018-08-15
Fix `this` scope of default functions for DocumentArray and Array

### LINKED ISSUE #6840: No access to "this" in default function for Array type
**bug**

When trying to set a default on an Array field, the default function doesn't give me access to the current object being created. I want to fill the array differently depending on the other values.
I found a similar issue https://github.com/Automattic/mongoose/issues/6263
But it is recommended to access pre-hooks, which I can't do. I need to have access to the *default* field.
Im using the latest version of mongoose (5.2.7).

Im using the same example as in the mongoose doc http://mongoosejs.com/docs/defaults.html#default-functions-and-this
I just added a field "actors" which is an array.
In releaseDate, I have access to **this** , but not in actors

```

const schema = new Schema({
	title: String,
	released: Boolean,
	releaseDate: {
		type: Date,
		default: function() {
			console.log("releaseDate: ", this.released);
			if (this.released) {
				return Date.now();
			}
			return null;
		}
	},
	actors: {
		type: [
			{
				name: String,
				age: Number
			}
		],
		default: function() {
			console.log("actors: ", this.released);
			return [
				{
					name: "Test",
					age: 20
				}
			];
		}
	}
});

const Movie = mongoose.model("Movie", schema);

const movie1 = new Movie({ title: "The Terminator", released: true });
const movie2 = new Movie({ title: "The Legend of Conan", released: false });

console.log(movie1);
```

Fonger:
Fix #6840 

**Summary**

DocumentArray ---(this=doc)---> Array ---(no this bind)---> Custom Default Function

**Test plan**

included in this pull request

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -101,7 +101,7 @@ function SchemaArray(key, cast, options, schemaOptions) {
     var defaultFn = function() {
       var arr = [];
       if (fn) {
-        arr = defaultArr();
+        arr = defaultArr.call(this);
       } else if (defaultArr != null) {
         arr = arr.concat(defaultArr);
       }
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -4852,6 +4852,29 @@ describe('Model', function() {
       });
     });
 
+    it('run default function with correct this scope in DocumentArray (gh-6840)', function() {
+      var schema = new Schema({
+        title: String,
+        actors: {
+          type: [{ name: String, character: String }],
+          default: function() {
+            // `this` should be root document and has initial data
+            if (this.title === 'Passenger') {
+              return [
+                { name: 'Jennifer Lawrence', character: 'Aurora Lane' },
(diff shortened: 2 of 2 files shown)


Fonger:
@vkarpov15 

Or maybe document this default behavior in website?
https://mongoosejs.com/docs/defaults.html#default-functions-and-this

Currently, it only shows
```
Unless it is running on a query with setDefaultsOnInsert,
a default function's this refers to the document.
```

vkarpov15:
@Fonger you mean document `$runBeforeSetters`? That sounds an awful lot making it part of the public API

Fonger:
@vkarpov15 Just document this behavior by mentioning that in primitive Array, `this` has no initial data unlike DocumentArray and other types have.

vkarpov15:
That sounds like it might be a bug. Can you open up an issue with a repro script @Fonger ?

PR #6877, opened 2018-08-17
Correctly propagate write concern options in save()

### LINKED ISSUE #6862: write concern timeout(wtimeout) not working
Mongoose version: 5.2.7
Node version: 8.9.4

This is timing out immediately if there are no enough number of nodes and not waiting for the wtimeout period. Previously I was using Mongoose v5.0.4 and it wasn't supporting write concern. After i upgraded to the latest version(5.2.7) w option works but not wtimeout. I tried passing these options even in the database connection URI but of no use.

```
 const options = {
    w: 3,
    wtimeout: 3000
 };
 await mongoose.connect(dburl, options)
   .then(() => console.log("Successfully established connection"))
   .catch((err) => {
       console.log("Failed to establish connection");
       throw err;
   });
```

Fonger:
**Summary**

`save()` doesn't work with write concern. Only legacy `safe` options are supported.
(IMO, I think `safe` should be deprecated in the next major release of mongoose) 

**This bug is critical because it makes mongoose lack of durability guarantee.**

Fixes #6862 and now write concern options are correctly propagated to mongodb.

**Test plan**

Add tests for both `writeConcern` defined in schema level and in save(`options`).

1 test to check if `w > 1` trigger error when topology of server is not a replica set
2 tests are skipped by default because they require replica sets and special setup to trigger write concern timeout. These 2 tests passed on my computer.

See the comments in code for details.

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -3,7 +3,7 @@
 /**
  * Test dependencies.
  */
-
+require('../lib/driver').set(require('../lib/drivers/node-mongodb-native'));
 const assert = require('power-assert');
 const co = require('co');
 const random = require('../lib/utils').random;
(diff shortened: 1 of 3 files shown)


test/model.test.js
```diff
@@ -5356,6 +5356,88 @@ describe('Model', function() {
...
+          assert.ok(error);
+          assert.equal(error.name, 'MongoError');
+          assert.equal(error.codeName, 'BadValue');
+          done();
+        });
+      });
+    });
+
+    it.skip('save() with wtimeout defined in schema (gh-6862)', function(done) {
+      // If you want to test this, setup replica set with 1 primary up and 1 slave down
```
vkarpov15: The correct term is "secondary", not "slave". Minor detail but we should use the correct term.

Fonger: Yes we should definitely use the correct term. Please fix this in a future commit.

PR #6888, opened 2018-08-18
WIP: fix: findAndModify and options [fields] deprecation

### LINKED ISSUE #6885: Aggregate.prototype.session() doesn't support chaining correctly
The recently added `session()` helper for aggregation queries fails to chain if given `null` as first argument.

```js
Aggregate.prototype.session = function(session) {
  if (session == null) {
    delete this.options.session;
    return;  // <------ should be `return this`!
  }
  this.options.session = session;
  return this;
};
```

### LINKED ISSUE #6887: If options.useFindAndModify is false, overwrite will not work
We should use findOneAndReplace to handle legacy `overwrite` options.

If `useFindAndModify` is set to false, the following CI test fail:
https://github.com/Automattic/mongoose/blob/7875d53d5e822acee852d23197d23466471a8467/test/model.findOneAndUpdate.test.js#L865-L880

IMO, `overwrite` option should be deprecated because we have `findOneAndReplace`.
`useFindAndModify` option should be deprecated too and default to false because the underlying library `mongodb` has been marked as deprecated since 2.2

Fonger:
**Summary**

Only change internal call and options.
If a user calls the deprecated API or use deprecated options on purpose, mongoose should still show the deprecation message. That's why CI tests have so many deprecation warning.

IMO, `useFindAndModify` and `overwrite` options should be deprecated in the next major release

- [ ] ~~`useFindAndModify` default to be `false`~~ (
EDIT after review from vkarpov15:
For your own good, please set to false explicitly to suppress deprecation warning because we can't guarantee if this is a breaking change at this moment. Please test if everything still works

- [x] find: options `fields` -> `projection`
- [x] findOndAndUpdate: options `fields` -> `projection`
- [x] remove: call deleteOne
- [ ] ensureIndex(es): call createIndex(es)
- [ ] check compatibility

**Test plan**

`useFindAndModify` default to be `false` may be a breaking change.
I fix #6887 and all current CI test pass. However I'm afraid the test coverage is not enough.

I check the following option so maintainers have access to add commit on this pull request
**Allow edits from maintainers**

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -86,7 +86,7 @@ Mongoose.prototype.STATES = STATES;
  * Currently supported options are:
  * - 'debug': prints the operations mongoose sends to MongoDB to the console
  * - 'bufferCommands': enable/disable mongoose's buffering mechanism for all connections and models
- * - 'useFindAndModify': false by default. Set to `true` to make `findOneAndUpdate()` and `findOneAndRemove()` use legacy `findAndModify()` rather than native methods`.
+ * - 'useFindAndModify': true by default. Set to `false` to make `findOneAndUpdate()` and `findOneAndRemove()` use native `findOneAndUpdate()` rather than `findAndModify()`.
  * - 'cloneSchemas': false by default. Set to `true` to `clone()` all schemas before compiling into a model.
  * - 'applyPluginsToDiscriminators': false by default. Set to true to apply global plugins to discriminator schemas. This typically isn't necessary because plugins are applied to the base schema and discriminators copy all middleware, methods, statics, and properties from the base schema.
  * - 'objectIdGetter': true by default. Mongoose adds a getter to MongoDB ObjectId's called `_id` that returns `this` for convenience with populate. Set this to false to remove the getter.
--- a/lib/index.js
+++ b/lib/index.js
@@ -86,7 +86,7 @@ Mongoose.prototype.STATES = STATES;
  * Currently supported options are:
  * - 'debug': prints the operations mongoose sends to MongoDB to the console
  * - 'bufferCommands': enable/disable mongoose's buffering mechanism for all connections and models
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -86,7 +86,7 @@ Mongoose.prototype.STATES = STATES;
  * Currently supported options are:
  * - 'debug': prints the operations mongoose sends to MongoDB to the console
  * - 'bufferCommands': enable/disable mongoose's buffering mechanism for all connections and models
- * - 'useFindAndModify': true by default. Set to `false` to make `findOneAndUpdate()` and `findOneAndRemove()` use native `findOneAndUpdate()` rather than `findAndModify()`.
+ * - 'useFindAndModify': false by default. Set to `true` to make `findOneAndUpdate()` and `findOneAndRemove()` use legacy `findAndModify()` rather than native methods`.
```
vkarpov15: I'm wary of changing `useFindAndModify` to false by default in a minor release, especially in light of the issue with `overwrite: true` that you found. I was not aware of that difference, which makes me wonder if there are other compatibility concerns that we're missing.

Fonger: It makes sense so reverting this change. Developers should opt in to suppress deprecation warning and test if it breaks anything at this moment.

lib/query.js
```diff
@@ -3018,6 +3014,9 @@ Query.prototype._findAndModify = function(type, callback) {
       return this;
     }
 
+    // honors legacy overwrite option for backward compatibility
+    const updateMethod = isOverwriting ? 'findOneAndReplace' : 'findOneAndUpdate';
```
vkarpov15: Can you add a test for this case?

Fonger: @vkarpov15 OK

vkarpov15:
Thanks. I'll merge this one for now. I added a `useCreateIndex` option in https://github.com/Automattic/mongoose/commit/7ecce58d90a0641565ba1c9a2ee28b09bcc9811b

PR #6893, opened 2018-08-18
style: use eslint to automatically migrate code to using let and const

ChristianMurphy:
**Summary**

Part of #3379

Helps move mongoose to the new block scoped variable declarations.

**Test plan**

CI should pass


ChristianMurphy:
Not entirely sure on the best approach to handle node 4 and 5 unhappiness:

1. We could set `--harmony` which should (?) allow `let` and `const` out strict mode for node 4.
2. We could add `'use strict';` to all the files
3. We could wait until Node 4 and 5 support are dropped
4. We could keep using `var` and not migrate to `let` and `const`

I'd personally lean toward 2 or 3.

thoughts? :thought_balloon:

ChristianMurphy:
Added strict mode

ChristianMurphy:
I'd appreciate some assistance with the `castToNumber` errors.
Could you look into them? Thanks! :bow:

PR #6895, opened 2018-08-19
Add redact and readConcern helper to Aggregate

Fonger:
### Add redact helper

```js
Model.aggregate().redact({
  $cond: {
    if: { $eq: [ '$level', 5 ] },
    then: '$$PRUNE',
    else: '$$DESCEND'
  }
})
```

`$redact` often comes with `$cond` operator, If 3 arguments are supplied,
Mongoose will wrap them with `if`-`then`-`else` of `$cond` operator respectively

```js
Model.aggregate().redact({ $eq: ['$level', 5] }, '$$PRUNE', '$$DESCEND')
```

### Add missing readConcern from Query
```js
Model.aggregate().readConcern('m'); // specify majority read concern
```

### Test Plan

Included in CI test

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -7,6 +7,7 @@ var Query = require('./query');
 var util = require('util');
 var utils = require('./utils');
 var read = Query.prototype.read;
+var readConcern = Query.prototype.readConcern;
 
 /**
  * Aggregate constructor used for building aggregation pipelines. Do not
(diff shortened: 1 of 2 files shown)


lib/aggregate.js
```diff
@@ -569,6 +572,79 @@ Aggregate.prototype.read = function(pref, tags) {
...
+ * @param {String|Object} [thenExpr] true case for the condition
+ * @param {String|Object} [elseExpr] false case for the condition
+ * @return {Aggregate} this
+ * @see $redact https://docs.mongodb.com/manual/reference/operator/aggregation/redact/
+ * @api public
+ */
+
+Aggregate.prototype.redact = function(expression, thenExpr, elseExpr) {
+  if (arguments.length === 3) {
+    if ((typeof thenExpr === 'string' && !thenExpr.startsWith('$$')) ||
```
vkarpov15: We can't use `startsWith()` right now because it isn't in Node.js 6 or earlier. Use `thenExpr.indexOf('$$') !== -1` instead.

Fonger: @vkarpov15 I tested it on node.js 4.0.0 and it works. node 0.12 does not work though.

```sh
$ nvm use 4.0.0
Now using node v4.0.0 (npm v2.14.2)
$ node --version
v4.0.0
$ node
> 'abc'.startsWith('ab')
true
> 'abc'.startsWith('aG')
false

$ nvm use 0.12.18
Now using node v0.12.18 (npm v2.15.11)
$ node --version
v0.12.18
$ node
> 'abc'.startsWith('ab')
TypeError: undefined is not a function
    at repl:1:7
```

Issue #6880, opened 2018-08-17
Deprecated collection methods from mongodb


icellusedkars:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Model.findByIdAndUpdate() is issuing a deprecation warning for mongodb - which has deprecated  collection.findAndModify. Mongodb suggests using findOneAndUpdate, findOneAndReplace or findOneAndDelete instead.

collection.ensureIndex is also deprecated - mongodb suggests using createIndexes instead.

**If the current behavior is a bug, please provide the steps to reproduce.**

Using Model.findByIdAndUpdate / findOneAndUpdate to alter a document returns these deprecation warnings.

**What is the expected behavior?**

No deprecation warnings!

**Please mention your node.js, mongoose and MongoDB version.**

Node v10.8.0
mongodb  v3.6.2
mongoose v5.2.9

ulrichb:
Same here. I get "find option [fields] is deprecated and will be removed in a later version."

Agamnentzar:
```
(node:13076) DeprecationWarning: collection.ensureIndex is deprecated. Use createIndexes instead.
(node:13076) DeprecationWarning: collection.remove is deprecated. Use deleteOne, deleteMany, or bulkWrite instead.
(node:13076) DeprecationWarning: collection.find option [fields] is deprecated and will be removed in a later version.
(node:13076) DeprecationWarning: collection.update is deprecated. Use updateOne, updateMany, or bulkWrite instead.
```
Here's some other ones I got

schmidtvl:
@(node:13076) DeprecationWarning: collection.ensureIndex is deprecated. Use createIndexes instead.

I noticed that if I add "autoIndex: false" to my mongoose.connect options, then this deprecation warning goes away.

Fonger:
Related to #6165 

You can [specify `useFindAndModify` to false](https://mongoosejs.com/docs/api.html#mongoose_Mongoose-set) to prevent partial deprecation message. IMO this should be default to false now.

However, other deprecated methods like ensureIndex(es) and `update` still need to be removed either.
Maybe we need to update all deprecated method and remove `useFindAndModify` options in the next major release?

Note:
[...]

PR #6917, opened 2018-08-24
Clean up remaining deprecation warnings

vkarpov15:
Re: #6880


test/schema.timestamps.test.js
```diff
@@ -326,7 +326,7 @@ describe('schema options.timestamps', function() {
     });
 
     after(function(done) {
```
lineus: we can pull done from the after function here.

Fonger:
Great!

Issue #6711, opened 2018-07-12
`Model.createCollection()` helper


vkarpov15:
Re: #6699, helpful so you don't need to pluralize the collection name to call `conn.createCollection()`. Plus, we should consider calling this in `Model.init()` to minimize headache re: having to create a collection before starting a transaction

PR #6948, opened 2018-08-31
Add Model.createCollection() helper

Fonger:
**Summary**

Re: #6711 (**Note: this PR doesn't call createCollection in Model.init()**).

`Model.createCollection()` will create the collection for the model with the pluralized or custom collection name in MongoDB. Unlike [Connection.createCollection()](https://mongoosejs.com/docs/api.html#connection_Connection-createCollection) you have to specify the right pluralized collection name for the model.

Note that if you create an index for the collection, there's no need to call this function. Just use Model.init().then(...) because MongoDB will create the collection for you during creating the indexes.

Maybe we should call this function in Model.init() by default if there's no index defined in schema in 5.3?

**Test plan**

call `db.collection('collection_names').stats()` to check if the collection of `Collection_Name` model is created.


vkarpov15:
Thanks, merged this into the 5.3 branch.

PR #6950, opened 2018-08-31
Fixing variable not defined - out of scope

cacothi:
**Summary**

The variable fixed was being used out of context. Simple fix.

### DIFF
--- a/lib/schematype.js
+++ b/lib/schematype.js
@@ -900,7 +900,7 @@ function asyncValidate(validator, scope, value, props, cb) {
       return;
     }
     called = true;
-    if (typeof returnVal === 'boolean') {
+    if (typeof customMsg === 'boolean') {
       return;
     }
     if (customMsg) {
--- a/lib/schematype.js
+++ b/lib/schematype.js
@@ -900,9 +900,6 @@ function asyncValidate(validator, scope, value, props, cb) {
       return;
     }
     called = true;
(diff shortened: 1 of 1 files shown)


lib/schematype.js
```diff
@@ -900,7 +900,7 @@ function asyncValidate(validator, scope, value, props, cb) {
       return;
     }
     called = true;
-    if (typeof returnVal === 'boolean') {
+    if (typeof customMsg === 'boolean') {
       return;
     }
```
Fonger: Good spot! After review the whole code, I find this code redundant. This if block should be totally removed. Can you do that?

It worked before because we used `var returnVal` and now we use `const returnVal` so it fails.

When we used `var returnVal` in the same line of code, it' s always get undefined so that's why this if check is unnecessary. Can you delete these 3 lines?

vkarpov15:
Good find on both @cacothi and @Fonger 's parts. I just removed this if statement, Fonger is right that it isn't necessary.

cacothi:
@vkarpov15 when this fix will be available? It's blocking my deploy on the latest version

vkarpov15:
@cacothi 5.2.13 will be released in a few hours. Follow on [slack](http://slack.mongoosejs.io/) or [twitter](https://twitter.com/mongoosejs) for release updates. We also have estimated release dates on the [milestone page](https://github.com/Automattic/mongoose/milestones)

PR #6983, opened 2018-09-05
Fixes #6982 fix wording on promise docs to not imply queries only return promises

### LINKED ISSUE #6982: Docs are unclear about whether queries are promises
The first paragraph sentence of https://mongoosejs.com/docs/promises.html is (my emphasis):
> Mongoose async operations, like .save() **and queries,** return ES6 promises.

However the first sentence of the very next paragraph is (original emphasis):
> Mongoose queries are **not** promises.

While it's not strictly contradictory to say "X is not a promise, but does return a promise", this is still pretty confusing. Perhaps someone with more expertise could re-write it to be a little less ambiguous about the intended meaning.

lineus:
As pointed out in #6982 The first sentence is incorrect. 

### DIFF
--- a/test/docs/promises.test.js
+++ b/test/docs/promises.test.js
@@ -24,15 +24,8 @@ describe('promises docs', function () {
   });
 
   /**
-<<<<<<< HEAD
    * Mongoose async operations, like `.save()` and queries, return thenables.
    * This means that you can do things like `MyModel.findOne({}).then()` and
-=======
-   * Mongoose async operations, like `.save()` and queries, return either an
-   * [ES6 promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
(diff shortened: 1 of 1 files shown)


test/docs/promises.test.js
```diff
@@ -24,11 +24,14 @@ describe('promises docs', function () {
   });
 
   /**
-   * Mongoose async operations, like `.save()` and queries, return
-   * [ES6 promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).
-   * This means that you can do things like `MyModel.findOne({}).then()` and
+   * Mongoose async operations, like `.save()` and queries, return either an
```
vkarpov15: Instead of listing out all possible return values, lets instead make it say we return a [thenable](https://promisesaplus.com/#terminology). There are other async operations that return non-promises, like `mongoose.createConnection()`.

PR #7012, opened 2018-09-15
add setUpdate to Query.prototype

lineus:
Query.prototype.set will iterate over the keys of an input object and add them to the current update doc, but at the moment the clearest way I can tell how to completely replace query._update is to use the private property directly. This change adds a setUpdate command to overwrite the current value of _update in a query object. 

I'm not super in love with the name `setUpdate`, maybe `replaceUpdate` would be better? The only reason I stuck with `setUpdate` is because it matches `setQuery` and `setOptions`. Regardless of the name, I think having a method to replace the existing update doc entirely in a pre query hook would be nice.

I Added a failing test, made it pass, all current tests pass.
```
mongoose>: npm test -- -g 'replaces existing update doc with new value'

> mongoose@5.2.15-pre test /Users/lineus/dev/node/src/mongoose
> mocha --exit test/*.test.js test/**/*.test.js "-g" "replaces existing update doc with new value"


 You're not testing shards!


  !

  0 passing (266ms)
  1 failing

  1) Query
       setUpdate
         replaces existing update doc with new value:
     TypeError: q.setUpdate is not a function
      at Context.<anonymous> (test/query.test.js:2739:9)



mongoose>: npm test -- -g 'replaces existing update doc with new value'

> mongoose@5.2.15-pre test /Users/lineus/dev/node/src/mongoose
> mocha --exit test/*.test.js test/**/*.test.js "-g" "replaces existing update doc with new value"


 You're not testing shards!


  ․

  1 passing (122ms)

mongoose>: npm test

> mongoose@5.2.15-pre test /Users/lineus/dev/node/src/mongoose
> mocha --exit test/*.test.js test/**/*.test.js


 You're not testing shards!


  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․,,․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․,,,․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․,,,․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․,,,,,,,,․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․,․․,․․,․,,․․,․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․,,․․․․․․․․․․․․,
  ,․․․․․․,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
  ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․,․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․,,․․․
  ․․․․․․․․․․․․․․․․․․․․,,․․․․․․․․․․․․․․․․․․․․․․․,․․․․․․․․,․,․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․,․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․
  ․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․․,,,,,,,,,․․․․․․
  ․․․․․․․․․․․․․․․․

  1899 passing (1m)
  163 pending

mongoose>: git status
On branch feature-setUpdate
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   lib/query.js
        modified:   test/query.test.js

no changes added to commit (use "git add" and/or "git commit -a")
mongoose>:
```

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -1393,6 +1393,25 @@ Query.prototype.getUpdate = function() {
 };
 
 /**
+ * Sets the current update operation to new value.
+ *
+ * ####Example:
+ *
+ *     var query = new Query();
+ *     query.update({}, { $set: { a: 5 } });
+ *     query.setUpdate({ $set: { b: 6 } });
+ *     query.getUpdate(); // { $set: { b: 6 } }
+ *
+ * @param {Object} new update operation
+ * @return {undefined}
+ * @api public
+ */
+
+Query.prototype.setUpdate = function(val) {
+  this._update = val;
+};
+
+/**
  * Returns fields selection for this query.
  *
  * @method _fieldsForExec
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -2731,4 +2731,14 @@ describe('Query', function() {
       assert.deepStrictEqual(q._conditions, { a: 1 });
     });
   });
+
+  describe('setUpdate', function() {
+    it('replaces existing update doc with new value', function() {
+      const q = new Query({}, {}, null, p1.collection);
+      q.set('testing', '123');
+      q.setUpdate({ $set: { newPath: 'newValue' } });
+      assert.strictEqual(q._update.$set.testing, undefined);
+      assert.strictEqual(q._update.$set.newPath, 'newValue');
+    });
+  });
 });


vkarpov15:
How about instead of `setUpdate()`, we do `set(obj, { replace: true })` ? Probably better to have an option than a separate function in this case.

lineus:
That works for me :+1

lineus:
@vkarpov15 After looking more closely at the current behavior of `Query.prototype.set()`, I think this would be better as a separate method. 

To me, the actions of mutating just the `$set` property of the update object **or** replacing the entire update object are far enough apart to warrant the separate method.

I did however, implement it, but I don't like how I did it. The main issue being that only I saw two choices for the parameter signature:

**The first** way was to add a third option to `.set()` representing the options object. This parameter however, wouldn't be used in the case of passing strings for path and value, and in the case of passing in an object, would always be null. 

I see this resulting in people being confused as to whether they need to pass in the options object to do `.set('path', 'value')`.

**The second** way was to make the second parameter serve as either the value to set when passing in strings, or the options object when passing in an object.  This was how I implemented it.

This feels kinda gross, and I was unable to find precedent for it, so I think it's a bad idea. There are obviously many function parameters that accept multiple types of values, and many that take optional parameters, but none that I could find that have one parameter that represents disparate variables in this way.

What do you think? Can we revisit the idea of a separate method?

vkarpov15:
Yeah because of support for both `set('path', val)` and `set({ path, val })`, the `options` approach is not gonna work too well. You're right.

Naming is still a problem though. I think `setUpdate()` is the best option though, to line up with `setQuery()` and `setOptions()` as you originally suggested.

lineus:
closing this because I mistakenly had my git origin set incorrectly. will open a new PR.

Issue #4412, opened 2016-08-14
Increment `updatedAt` for nested schemas when setting a nested path in update


vkarpov15:
``` javascript
var childSchema = new Schema({ name: String }, { timestamps: true });
var parentSchema = new Schema({ child: childSchema });
var Parent = mongoose.model('Parent', parentSchema);

// Should set `child.updatedAt` as well. Currently does not do so
Parent.update({}, { $set: { 'child.name': 'Luke' } });
```

Re: #4049

joaoreynolds:
+1 for this! In the meantime I can set it explicitly with my update, but it does seem like something mongoose should handle.

Thanks for your hard work, I wish I could help!

vkarpov15:
Thanks for reporting the $push issue, will fix as well.

akashdeepsinghal:
Hi! Any updates on this?

vkarpov15:
@akashdeepsinghal nope, not yet. We'll add this to our next minor release :+1:

MBayoumi:
AssignmentSchema = new mongoose.Schema({
  agentId: {type: String},
}, { strict: false, timestamps: true });


this.conversationModel.findByIdAndUpdate(conversationID, { $push: { assignments: assignment } }, {
      new: true,
      upsert: true,
    });

when i use $set timestamps fields were created and works fine but when i'm using $push timestamps fields don't

vkarpov15:
@MBayoumi please open up a separate issue with detailed code samples and what version of Mongoose you're using.

PR #7033, opened 2018-09-20
feat(timestamps): update timestamps when doing `$set` on a child path

vkarpov15:
The crux of this issue (#4412) is that if you have a child schema with timestamps:

```javascript
const childSchema = new Schema({ name: String }, { timestamps: true });
const parentSchema = new Schema({ child: childSchema }, { timestamps: true });
const Parent = db.model('gh4412', parentSchema);
```

Doing `Parent.updateOne({}, { $set: { 'child.name': 'Luke' } });` will **not** currently update `child.updatedAt`. In order to enable this, I added a couple properties: a schematype will now track `$parentSchema`, and a schema may have an associated `$schemaType` if it is the schema for an array or single nested doc. This enables L90-L104 of `lib/helpers/update/applyTimestampsToChildren.js` to do the magic of figuring out that `$set: { 'child.name': 'Luke' }` should bump `child.updatedAt` because of `childSchema.options.timestamps`.

There's still some work to be done:

[ ] Make `$schemaType` and `$parentSchema` non-enumerable
[ ] Test that this works with arrays

But at least with this PR, all the timestamp tests pass:

```
$ npm test -- -g "imestamp"

> mongoose@5.2.17-pre test /XYZ
> mocha --exit test/*.test.js test/**/*.test.js "-g" "imestamp"


 You're not testing shards! 

(node:23633) DeprecationWarning: Mongoose: the `safe` option for `save()` is deprecated. Use the `w` option instead: http://bit.ly/mongoose-save

  ․․․․․(node:23633) DeprecationWarning: collection.update is deprecated. Use updateOne, updateMany, or bulkWrite instead.
․․․․․․․․․․․․․․․․․․․․․․․․․․,․․․․․․․․․․․

  42 passing (2s)
  1 pending

$ 
```

@fonger @lineus


Fonger:
I find that `Map` subdocs doesn't get the updated timestamps.

The following test fail:

```js
  it('increments timestamps for map subdocs (gh-4412)', function() {
    const childSchema = new Schema({ name: String }, {
      timestamps: true,
      versionKey: false
    });
      map: {
        type: Map,
        of: childSchema
      }
    }, {
      // timestamps: true,
      versionKey: false
    });
    const Parent = db.model('gh4412-map', parentSchema);

    return co(function*() {
      let doc = yield Parent.create({ map: { mapProp: { name: 'foo' } } });
      assert.ok(doc.map.get('mapProp').updatedAt);
      assert.ok(doc.map.get('mapProp').createdAt);

      const start = Date.now();
      yield cb => setTimeout(cb, 10);

      yield Parent.updateOne({}, { $set: { 'map.mapProp.name': 'Luke' } });

      doc = yield Parent.findOne();

      const updatedAt = doc.map.get('mapProp').updatedAt.valueOf();

      assert.ok(updatedAt > start, `Expected ${updatedAt} > ${start}`); // FAIL TEST
    });
  });
```

vkarpov15:
@gregksurveillus  we missed that. Will fix.

PR #7059, opened 2018-09-27
fix(connection): add useFindAndModify option in connect

NormanPerrin:
**Summary**
Add `useFindAndModify` option to config in the `Mongoose.prototype.connect`.
Also update the relationated documentation.

**Motivation**
To be consistant with the `Mongoose.prototype.set` and `Mongoose.prototype.connect` options.
It would be nice to be able to set the config as the `options` param.

**Example**
```javascript
mongoose.connect(MONGODB_URI, {
    useNewUrlParser: true,
    useFindAndModify: false,
    useCreateIndex: true
});
```

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -449,6 +449,11 @@ Connection.prototype.openUri = function(uri, options, callback) {
       delete options.useCreateIndex;
     }
 
+    if ('useFindAndModify' in options) {
+      this.config.useFindAndModify = !!options.useFindAndModify;
+      delete options.useFindAndModify;
+    }
+
     // Backwards compat
     if (options.user || options.pass) {
       options.auth = options.auth || {};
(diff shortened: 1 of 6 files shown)


lib/connection.js
```diff
@@ -449,6 +449,11 @@ Connection.prototype.openUri = function(uri, options, callback) {
       delete options.useCreateIndex;
     }
 
+    if ('useFindAndModify' in options) {
+      this.base.options.useFindAndModify = !!options.useFindAndModify;
```
vkarpov15: I like the idea of these options, but this is a big no-no. Connections should not modify the mongoose global's options, because you're in for a nasty surprise if you create multiple connections with different useFindAndModify options:

```javascript
mongoose.createConnection(uri1, { useFindAndModify: true });
mongoose.createConnection(uri2, {}); // Forgot to set useFindAndModify

mongoose.options.useFindAndModify; // false!
```

Instead, please use the same pattern as `useCreateIndex`: if schema option is set, use that. If schema option is not set, use the connection option if it is set. If not, fall back to the mongoose global's option.

NormanPerrin: Now it differences the query, the connection level and the global.

PR #7068, opened 2018-09-29
fixes #7022 add close() to ChangeStream

### LINKED ISSUE #7022: model.watch.close not is not a function
i'm using version: `"mongoose": "^5.2.5"`
my model: 
```js
const WatchSchema = new mongoose.Schema({
    channel: {type: String, required: true},
    createdAt: Date,
    updatedAt: {type: Date, default: Date.now},
}, {
    toObject: {
        versionKey: false,
        virtuals: true,
        transform: function (doc, ret) {
            delete ret._id;
        }
    }
});

const Watch = mongoose.model('watch', WatchSchema);
module.exports = Watch;
```
test controller:

```js
const Watch = require("../models/Watch");
Watch.remove({});

let resultWatch = Watch.watch([{$match: {operationType: 'insert'}}]);

resultWatch.on('change', (data) => {
    console.log(data);
});

resultWatch.on('close', (data) => {
    console.log(data);
});

console.log(resultWatch);

let i = 0;
setInterval(() => {
    new Watch({name: 'test' + i, channel: 'test'}).save().catch();
    i++;
}, 500);

setTimeout(()=>{
    resultWatch.close(); //<<---resultWatch.close is not a function
},10000);
```

i was tested with `mongodb@3.1.1` working well.

I try to research mongoose code && search issue on repo but can not resolve that.

Please help.

lineus:
re #7022. This change exposes the underlying ChangeStream `close()` method on mongoose ChangeStreams while honoring mongoose's ability to buffer commands. If a `watch()` operation is buffered when `close()` is called on the ChangeStream, the watch operation will never run on the underlying driver.  

added a couple of failing tests, made them pass locally. Lots of tests fail for me in master at the moment, digging into why. Creating this as is to see if they pass on Travis.

With this code change, the repro script in #7022 can be changed to call `inserts.close()` and it works as expected.

### DIFF
--- a/lib/cursor/ChangeStream.js
+++ b/lib/cursor/ChangeStream.js
@@ -15,10 +15,13 @@ class ChangeStream extends EventEmitter {
     super();
 
     this.driverChangeStream = null;
-
+    this.closed = false;
     // This wrapper is necessary because of buffering.
     if (model.collection.buffer) {
       model.collection.addQueue(() => {
+        if (this.closed) {
+          return;
+        }
         this.driverChangeStream = model.collection.watch(pipeline, options);
         this._bindEvents();
         this.emit('ready');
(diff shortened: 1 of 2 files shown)


vkarpov15:
Any idea why the tests are failing?

PR #7080, opened 2018-10-02
fixes #7076 add schema types to api docs and update links on schematypes page

### LINKED ISSUE #7076: Fix broken links in schematypes.html
In https://mongoosejs.com/docs/schematypes.html, the 'Number' and 'Buffer' links are still broken. I did some work in #7040 to add the missing sections to `schematypes.html` but didn't quite have time to finish.

lineus:
re #7076 This change adds the files from lib/schema to the api docs and updates the links to them on the schematypes page. 

**note re the changes to lib/schema/{objectid,decimal128}.js where I removed the eslint rule at the top:**

This was necessary because that line caused the doc build to fail. ( I could toggle the failure by adding/removing that line ) not sure why :smile:

make docclean && make gendocs && node static.js - verified the links work.


docs/source/api.js
```diff
@@ -22,6 +22,18 @@ const files = [
   'lib/aggregate.js',
   'lib/cursor/AggregationCursor.js',
   'lib/schematype.js',
+  'lib/schema/array.js',
```
vkarpov15: IMO we shouldn't add `schema/boolean`, etc. to the API docs. These docs expose a lot of functions that people just don't use, and the distinction between `schema/boolean` and 'boolean' is pretty confusing.

docs/schematypes.jade
```diff
@@ -20,14 +20,14 @@ block content
     The following are all the valid SchemaTypes in mongoose.
 
     - [String](#strings)
-    - [Number](./api.html#schema-number-js)
+    - [Number](./api.html#Number)
```
vkarpov15: I'd recommend this link to a `#numbers` section in this file that describes the particulars of the Mongoose number type, like the `#strings` section. The intent with this issue is to add a section that describes what gets casted to a number, what custom validators are available for numbers, and what passes the `required` validator for numbers. Maybe in the future we add notes on how `$inc` works with numbers. IMO the particulars of numbers are best explained in prose rather than API docs, because the end user generally doesn't call the functions in the `schema/number` API. Does that make sense?

docs/source/api.js
lineus: That makes sense. I just figured that the @public jsdocs meant they should be in the api docs. Fixing this now.

docs/schematypes.jade
lineus: Yeah, I'll do this instead :)

lineus:
I'm working on buffer now. 

*I'm also seriously considering creating a keyboard snippet that turns `bugger` into `buffer`. I always misspell buffer (even just now) the **same-wrong-way** .*

lineus:
@vkarpov15 can you take a look and see if it looks ok? Thanks!

docs/schematypes.jade
```diff
@@ -213,6 +213,37 @@ block content
...
+    new Car({ age: true }).age; // 1 as a Number
+    new Car({ age: false }).age; // 0 as a Number
+    new Car({ age: { valueOf: () => 83 } }).age; // 83 as a Number
+    ```
+
+    If you pass an object with a `valueOf()` function that returns a Number, Mongoose will
+    call it and assign the returned value to the path.
+
+    There are a few noteworthy values that will not be cast to a Number:
+    The values `null` and `undefined` are not cast, they are instead assigned to the path.
```
vkarpov15: Fine for now, but in general 'X is not cast' and 'Y are assigned' are tech writing antipatterns because they do not make it clear what does the casting and assigning. Whenever I read these type of sentencnes in docs, I'm always scratching my head as to whether I, the consumer of the software, am responsible for casting or whether Mongoose does it for me. Whenever you find yourself tempted to write something like this, just say "Mongoose does not cast the values `null` and `undefined`".

docs/schematypes.jade
```diff
@@ -228,6 +259,26 @@ block content
...
+    constructor or the string `'Buffer'`.
+
+    ```javascript
+    const schema1 = new Schema({ binData: Buffer }); // binData will be cast to a Buffer
+    const schema2 = new Schema({ binData: 'Buffer' }); // Equivalent
+
+    const Data = mongoose.model('Data', schema2);
+    ```
+
+    There are several types of values that will be successfully cast to a Buffer.
```
vkarpov15: DItto above, "Mongoose will successfully cast the below values to buffers"

PR #7101, opened 2018-10-07
fixes #7099 orFail() returns undefined when a match IS found

### LINKED ISSUE #7099: orFail() returns undefined when a match IS found
Mongodb: **3.6.5**, Mongoose: **5.3.1**, Node Driver: **3.1.6**

As per the docs `onFail()` helper function should throw when nothing is found (or done in the case of update/delete) but if it matches on a document should return it right? It seems that just adding it to a working query with a valid `_id` it would make the query output `undefined`:

Schema:
```

var AuthorSchema = new Schema({
  email: {
    type: String,
    lowercase: true,
    required: true
  })
module.exports = mongoose.model('Author', AuthorSchema)
```

Running this then:
```
var A = require('./models/author');

var r = A.findOne({ _id: mongoose.Types.ObjectId("5bb5806083512515c9dd88af")})
   .then(x => console.log('result', x))
```

Would return the **expected model as x**. However:

```
var r = A.findOne({ _id: mongoose.Types.ObjectId("5bb5806083512515c9dd88af")})
   .orFail()
   .then(x => console.log('result', x))
```

Would output **undefined**. Same happens if we use `.exec()`.

gsandorx:
Fixes issue #7099: "orFail() returns undefined when a match IS found"

**Summary**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -3707,13 +3707,16 @@ Query.prototype.orFail = function(err) {
           err = typeof err === 'function' ? err.call(this) : err;
           throw err;
         }
-        break;
+        return res;
+        // there's no need for a break below the return, as this code is unreachable
+        // break
       case 'findOne':
         if (res == null) {
           err = typeof err === 'function' ? err.call(this) : err;
           throw err;
         }
-        break;
+        return res;
+        //break;
       case 'update':
       case 'updateMany':
       case 'updateOne':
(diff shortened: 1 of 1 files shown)


lineus:
We should update the tests to catch this problem in the future. Currently the tests only ensure that an Error isn't returned for a good match as opposed to **actually** testing that queries with `.orFail()` still return a valid response. The relevant tests are in `test/query.test.js`.

gsandorx:
Yes, please, give me a hand with the tests if you can now :)

vkarpov15:
Well that's embarrassing. Thanks for catching this. I'll add a test in master asap.

Issue #7108, opened 2018-10-08
DeprecationWarning: collection.findAndModify is deprecated


Dev-Wito:
https://github.com/Automattic/mongoose/blob/5c0e444ae8c50e352de5c4f7a763419fbb4dd316/lib/query.js#L2727

DeprecationWarning: collection.findAndModify is deprecated. Use findOneAndUpdate, findOneAndReplace or findOneAndDelete instead.

"mongoose": "5.3.1",

lineus:
@Dev-Wito  In mongoose@5.3.1 you can use `mongoose.set('useFindAndModify', false)`. There's [a guide here](https://mongoosejs.com/docs/deprecations.html) with more details. Here's an example script and it's output:

### 7108.js
```js
#!/usr/bin/env node
'use strict';

const assert = require('assert');
const mongoose = require('mongoose');
mongoose.set('useFindAndModify', false);
const { Schema, connection} = mongoose;
const DB = '7108';
const URI = `mongodb://localhost:27017/${DB}`;
const OPTS = { family: 4, useNewUrlParser: true };

const schema = new Schema({
  name: String
});

const Test = mongoose.model('test', schema);

async function run() {
  assert.strictEqual(mongoose.version, '5.3.1');
  await mongoose.connect(URI, OPTS);
  await connection.dropDatabase();
  const cond = { name: 'test' };
  const update = cond;
  const opts = { new: true, upsert: true };
  await Test.findOneAndUpdate(cond, update, opts);
  const created = await Test.findOne({});
  assert.strictEqual(created.name, 'test');
  console.log('All Assertions Pass.');
  await connection.close();
}

run();
```
### Output:
```
issues: ./7108.js
All Assertions Pass.
issues:
```

lineus:
**It seems like there may be an issue with setting useFindAndModify in the connection options in mongoose@5.3.2 though**

Repro script:

### 7059.js
```js
#!/usr/bin/env node
'use strict';

const assert = require('assert');
const mongoose = require('mongoose');
const { Schema, connection} = mongoose;
const DB = '7059';
const URI = `mongodb://localhost:27017/${DB}`;
const OPTS = { family: 4, useNewUrlParser: true, useFindAndModify: false };

const schema = new Schema({
  name: String
});

async function run() {
  assert.strictEqual(mongoose.version, '5.3.2');
  await mongoose.connect(URI, OPTS);
  await connection.dropDatabase();
  const Test = mongoose.model('test', schema);
[...]

PR #7110, opened 2018-10-08
add connection option useFindAndModify to findOneAndUpdate

lineus:
As I noticed in #7108, `Model.findOneAndUpdate()` in 5.3.2 doesn't honor the option useFindAndModify in connection options.

Repro script:

### 7059.js
```js
#!/usr/bin/env node
'use strict';

const { Schema, connection} = mongoose;
const DB = '7059';
const URI = `mongodb://localhost:27017/${DB}`;
const OPTS = { family: 4, useNewUrlParser: true, useFindAndModify: false };

const schema = new Schema({
  name: String
});

async function run() {
  assert.strictEqual(mongoose.version, '5.3.2');
  await mongoose.connect(URI, OPTS);
  await connection.dropDatabase();
  const Test = mongoose.model('test', schema);
  const cond = { name: 'test' };
  const update = cond;
  const opts = { new: true, upsert: true };
  await Test.findOneAndUpdate(cond, update, opts);
  const created = await Test.findOne({});
  assert.strictEqual(created.name, 'test');
  await connection.close();
}

run();
```
### Output:
```
issues: ./7059.js
(node:54429) DeprecationWarning: collection.findAndModify is deprecated. Use findOneAndUpdate, findOneAndReplace or findOneAndDelete instead.
All Assertions Pass.
issues:
```

### Output with this change: 
```
issues: ./7059.js
All Assertions Pass.
issues:
```


lineus:
@vkarpov15 

do we need the following in lib/model.js find/findOne and the corresponding tests? This doesn't make sense to me, can you help me out :smile:


```
  const connectionConfig = this.collection.conn.config;
  if (connectionConfig && typeof connectionConfig.useFindAndModify !== 'undefined') {
    mq._mongooseOptions.useFindAndModify = !!connectionConfig.useFindAndModify;
  }
```

lib/model.js
```diff
@@ -2153,6 +2153,11 @@ Model.findOneAndUpdate = function(conditions, update, options, callback) {
   const mq = new this.Query({}, {}, this, this.collection);
   mq.select(fields);
 
+  const connectionConfig = this.collection.conn.config;
+  if (connectionConfig && typeof connectionConfig.useFindAndModify !== 'undefined') {
```
vkarpov15: We shouldn't rely on setting options in `Model` like this, because it is perfectly valid to do something like `await Model.findOne({ foo: 'bar' }).updateOne({ bar: 'baz' }).findOneAndUpdate()`. Chaining means we need to be careful and make sure we don't check global options until the last minute. Check how we handle `strictQuery` for an example.

vkarpov15:
@lineus I didn't read your followup comments before commenting. You're right, we can't just set `useFindAndModify` in `Model.findOneAndUpdate()` because of chaining. Add this logic to `Query.prototype._findAndModify()` and it should work

lineus:
@vkarpov15 I'll have time to get this done this afternoon. **Can you check my intentions below for sanity?**

   1. remove the following from `Model.find()`, `Model.findOne()`, and `Model.findOneAndUpdate()`

    const connectionConfig = this.collection.conn.config;
    if (connectionConfig && typeof connectionConfig.useFindAndModify !== 'undefined') {
      mq._mongooseOptions.useFindAndModify = !!connectionConfig.useFindAndModify;
    }

   2. add the logic for grabbing the `useFindAndModify` setting from the connection options to `Query.prototype._findAndModify()`

   3. change the tests in `connection.test.js to call methods that are relevant to the option.

vkarpov15:
@lineus that sounds generally correct.

lineus:
@vkarpov15 The current tests for this rely on the connection `useFindAndModify` setting being saved to `query._mongooseOptions.useFindAndModify` before the query is executed. The changes I made removed that logic, the end result being that the winning option ( `base` vs. `conn` vs. `query` ) is only available internally at query execution time. 

I'm not quite sure how to test this as it stands now. *I suppose we could set a property on the returned doc that identifies the option that won and it's value, but my instincts tell me this might be overkill*. 

**What do you think I should do about the tests?**

vkarpov15:
@lineus can you change these tests to line up with https://github.com/Automattic/mongoose/blob/9af233a2884976a475af131974af0bad6df905b7/test/model.findOneAndUpdate.test.js#L1899-L1915 ? That's how you test that `findOneAndUpdate()` won out over `findAndModify()`.

PR #7122, opened 2018-10-10
Model.init() fails when an index exists on schema + autoCreate == true

jesstelford:
**Summary**

Failure:
```
❯ node index.js
~/dev/mongoose-collection-bug/node_modules/mongoose/lib/model.js:1400
      model.collection.addQueue(create, []);
                       ^
TypeError: model.collection.addQueue is not a function
    at ~/dev/mongoose-collection-bug/node_modules/mongoose/lib/model.js:1400:24
    at process._tickCallback (internal/process/next_tick.js:61:11)
```

[`Model.init()`](https://github.com/Automattic/mongoose/blob/763ec8cee75406df172faa3cab798cd9235d87a0/lib/model.js#L1005), when the model schema contains an index and the `autoCreate` option is `true`, [will call `Model.listIndexes()`](https://github.com/Automattic/mongoose/blob/763ec8cee75406df172faa3cab798cd9235d87a0/lib/model.js#L1032-L1037), erroring as above.

This is because [within `Model.listIndexes()`](https://github.com/Automattic/mongoose/blob/763ec8cee75406df172faa3cab798cd9235d87a0/lib/model.js#L1241), it [expects `this.collection.addQueue` to exist](https://github.com/Automattic/mongoose/blob/763ec8cee75406df172faa3cab798cd9235d87a0/lib/model.js#L1251), however it does not.

Tracking this one step further back; `Model.init()` [will call `Model.createCollection()`](https://github.com/Automattic/mongoose/blob/763ec8cee75406df172faa3cab798cd9235d87a0/lib/model.js#L1032-L1037), which incorrectly uses the `db.createCollection()` method, resolving `this.collection` to a `NativeDb Collection`. Code elsewhere in `model.js` does not expect a `NativeDb Collection`, but rather an instance of `Collection`, which is obtained via `db.collection()` instead.

And so this PR proposes that change.

**Test plan**

Based on this reproduction:

`package.json`
```json
{
  "name": "mongoose-collection-bug",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "mongoose": "5.3.2"
  }
}
```

`index.js`
```javascript

  // Setup an index
  { name: {type: String, index: { unique: true }} },
  // Toggle creation on
  { autoCreate: true },
);

mongoose.connect('mongodb://localhost:27017', { useNewUrlParser: true }).then(() => {
  // Create a model and wait for initialisation to occur
  return mongoose.model('Actor', schema).init();
}).then(() => {
  console.log('Success');
}).catch(error => {
  console.error(error);
});
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1112,10 +1112,10 @@ Model.createCollection = function createCollection(options, callback) {
   }
 
   return utils.promiseOrCallback(callback, cb => {
-    this.db.createCollection(this.collection.collectionName, options, utils.tick((error, collection) => {
+    this.db.createCollection(this.collection.collectionName, options, utils.tick((error) => {
       if (error) return cb(error);
-      this.collection = collection;
-      cb(null, collection);
+      this.collection = this.db.collection(this.collection.collectionName, options);
+      cb(null, this.collection);
     }));
   });
 };


jesstelford:
🤔  I'm not sure how to go about fixing those test failures. Any pointers would be appreciated!

jesstelford:
Temporarily published as `@jesstelford/mongoose@5.3.3-pre-gh7122` for anyone who might be blocked by this 👍

Issue #7144, opened 2018-10-18
Error: key {key} must not contain '.' (working on 5.1.5, not on later versions)


srokatonie:
**BUG**

We were running Mongoose on version 5.1.5 and all was good. Then cloned repo, npm installed everything (Mongoose updated to 5.1.7, but not working on most recent release 5.3.4).

Started getting error:
````
Error: key {key} must not contain '.'
````

It happens when you try to create a key that contains "." (dot), for example:
````
{
  "settings.appName": "Funky giraffes"
}
````

I believe that `"settings.appName"` is a valid JSON string and should be supported?

Node version: v9.3.0
Mongoose: 5.1.7 - 5.3.4 (working on 5.1.5)
MongoDB shell version v3.6.2

**Comment by lineus** [comment]
@srokatonie can you please include your schema and the operation that leads to the error?

**Comment by srokatonie** [comment]
@lineus I tried to reproduce, but it worked fine for some reason! possibly creating schema/collection using one version and then updating may cause the error, but need to look deeper into that. Let's close for now and reopen if I come up with something.

**Comment by sibelius** [comment]
this is happening to me with a `Mixed` field

here is a failing model

```jsx
raw: {
          type: Mixed,
          description: 'raw data of this calendar event',
        },
```

```jsx
{
      '@odata.etag': 'blah',
}
```

vkarpov15:
So I repro-ed this with the below script, it starts failing with Mongoose 5.2.0:

```javascript
const assert = require('assert');
const mongoose = require('mongoose');
mongoose.set('debug', true);

const GITHUB_ISSUE = `gh7144`;
const connectionString = `mongodb://localhost:27017/${ GITHUB_ISSUE }`;
const { Schema } = mongoose;

run().then(() => console.log('done')).catch(error => console.error(error.stack));

async function run() {
  await mongoose.connect(connectionString, { checkKeys: false });
  await mongoose.connection.dropDatabase();

  const s = new Schema({ raw: { type: Schema.Types.Mixed } });
  const M = mongoose.model('gh7144', s);

  const raw = {
    '@odata.etag': 'blah'
  };
  
  const data = await M.create({ raw: raw });

  assert.deepEqual(data.raw, raw);
}
```

[...]

PR #7228, opened 2018-11-08
test(mixed): add failing test to see #7144

sibelius:
**Summary**

I've upgraded mongoose from 5.1.4 to 5.3.10 and some tests in our codebase started failing.
This is my attempt to provide a failing test to help debug this issue.
The test is not working, is it possible to use async await on tests? it would make writing tests much easier.
I'm not sure either if the test is in the right file

I've got this error when running the tests:

```
  1) schematype mixed
       mixed types with dot
         should enable key with dot(.) on mixed types:
     OverwriteModelError: Cannot overwrite `M1` model once compiled.
      at new OverwriteModelError (lib/error/overwriteModel.js:20:11)
      at Mongoose.model (lib/index.js:427:13)
      at Context.<anonymous> (test/schema.mixed.test.js:41:26)
```

**Test plan**
run to get the error
`npm test -- -g 'schema'`

### DIFF
--- a/test/schema.mixed.test.js
+++ b/test/schema.mixed.test.js
@@ -34,4 +34,25 @@ describe('schematype mixed', function() {
       done();
     });
   });
+
+  describe('mixed types with dot', function() {
+    it('should enable key with dot(.) on mixed types', function(done) {
+      const s = new Schema({ raw: { type: Schema.Types.Mixed } });
+      const M = mongoose.model('M1', s);
+
+      const raw = {
+        '@odata.etag': 'blah',
+        id: 'id',
+      };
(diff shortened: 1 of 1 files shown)


vkarpov15:
I'm going to close this because the test is incorrect because mongoose tests don't open up the default `mongoose` connection. I'll make more comments in #7144 . In general, if you want to repro an issue, please create a script that does `require('mongoose')` instead of adding a test, unless you intend to add code as well. It is much easier for us to test your code against different versions of mongoose and much easier for you to not have to figure out the quirks of Mongoose's test setup.

PR #7238, opened 2018-11-12
Removing ^4.11.0 as compatible with MongoDB 3.6

a-harrison:
Mongoose `^4.11.0` use the `2.2.X` version of the MongoDB Node.js driver (https://github.com/Automattic/mongoose/blob/4.11.0/package.json#L26) whereas only versions `>= 3.0` of the driver is recommended for MongoDB 3.6 (https://docs.mongodb.com/ecosystem/drivers/driver-compatibility-reference/#reference-compatibility-mongodb-node).

**Summary**

The compatibility documentation currently lists Mongoose `^4.11.0` as compatible with MongoDB 3.6. However, this version of Mongoose does not use the MongoDB Node.js driver version that is recommended for this version of MongoDB 

Removing this range will prevent users from using a not-recommended version of the underlying Node.js driver on MongoDB 3.6.

**Test plan**

Documentation change only; no testing should be needed.


vkarpov15:
@a-harrison I didn't notice when I was reviewing this, but you need to make these changes to `compatibility.jade`, `compatibility.html` is compiled code so your changes got overwritten: https://mongoosejs.com/docs/compatibility.html

a-harrison:
Sorry, fixing.

Issue #7045, opened 2018-09-24
Support mongoose.SchemaType.cast() to overwrite Mongoose's built-in casting for a given SchemaType


sebastian-nowak:
I find it very hard to write reliable validation code when values are silently converted to another type.

```js
#!/usr/bin/env node
(async function () {
  const mongoose = require('mongoose');
  mongoose.connect('mongodb://localhost:27017/test', {useNewUrlParser: true});

  const schema = new mongoose.Schema({
    text: String
  });
  const Model = mongoose.model('model', schema);
  const doc = new Model();

  doc.text = 123123;
  console.log(doc.text);
  console.log(typeof doc.text);
  console.log(doc.validateSync());
})();
```

I think there should be a toggle to enable strict mode which would cause a validation error when there is a type mismatch.

lineus:
This is a great idea. We'll give some thought to the implementation and add this. Thanks @sebastian-nowak !

sebastian-nowak:
Thanks @lineus, that's great news :)

I think one edge case to consider is what to do about ObjectIds. I rarely see code that creates them as objects, they're often used as strings. I don't know what's the best option.
* Maybe it's fine to require creating proper ObjectId objects by hand? After all, it's called 'strict mode'.
* Or maybe ObjectIds should be still automatically converted, while all other types would throw on type mismatch
* Or maybe strict mode should be flexible and it would be possible to set the behavior manually for each type

Cheers!

vkarpov15:
@sebastian-nowak what do you think about making it possible to edit this on a per-schema-type basis? So instead of toggling casting on/off at the global level, you can do stuff like:

```javascript
mongoose.Number.cast(false); // Disable casting for numbers
mongoose.ObjectId.cast(v => typeof v === 'string' ? new ObjectId(v) : v); // Custom casting for ObjectIds
```

What do you think of that idea?

sebastian-nowak:
I think it would be okay for majority of use cases, assuming that:
* Cast callback would be called only on type mismatch, otherwise there would be a lot of boilerplate code
* It would be somehow possible to invalidate the field from within the callback. Perhaps by throwing an exception? The value would be assigned as is, but the field would be invalidated.

A nice extra thing to have would be an ability to override the `cast` method on a schema level. It would add a lot of flexibility and definitely cover all use cases:

```js
mongoose.String.cast(false);

const schema = new mongoose.Schema({
  userId: {
    type: String,
[...]

PR #7256, opened 2018-11-17
feat(schematype): allow custom casters for numbers

vkarpov15:
This is a proof of concept for `mongoose.Number.cast()`, a getter/setter around the function Mongoose uses to cast numbers. This will enable users to plug in their own casting logic for all numbers, including geo queries, etc., including disabling casting for numbers using `mongoose.Number.cast(false)`.

Still todo:

- [x] inheritance: http://thecodebarbarian.com/static-properties-in-javascript-with-inheritance.html
- [x] ObjectId
- [x] Boolean
- [x] String
- [x] Date
- [x] Decimal128


vkarpov15:
Re: #7045

PR #7296, opened 2018-12-03
fix: added changes to pass session in model create to rollback

jaiyashree:
**Summary**

Added session param in create method. This is made optional to work on all existing cases and passed as a session instead of a document-spread only if it contains a 'session' key.

**Test plan**

The 'options' param consisting the session to rollback or commit the transaction is appended.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2600,13 +2600,18 @@ Model.create = function create(doc, options, callback) {
     options = options != null && typeof options === 'object' ? options : {};
   } else {
     const last = arguments[arguments.length - 1];
-    options = {};
+    options = options.session ? { session: options.session } : {};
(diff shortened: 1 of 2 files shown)


jaiyashree:
Hi @vkarpov15 ,
Please let me know if it is fine that i add a check to ensure inner properties of options.session matches with the properties of db-session, bcoz by passing an array of single object, one needs to handle the result too (like result[0] instead of result).

vkarpov15:
@jaiyashree I'm not particularly thrilled with that idea either. Trying to guess whether 2nd arg is `options` or document is very error prone and we'll have a lot of edge cases. If you want to create one doc without doing `.then(res => res[0])`, you can just do `new User({ name: 'foo' }).save({ session })`

PR #7335, opened 2018-12-14
Fix mistake in populate documentation

zcfan:
In the section "What If There's No Foreign Document?", I found the sample code confusing, I believe it's a mistake.

After `Story.deleteMany` the story "Casino Ryale" not exists anymore. so we can't `findOne`. the `story` it returned should be `null` and `story.author` will throw `TypeError: Cannot read property 'author' of null`.

**Summary**

**Test plan**


vkarpov15:
You're 100% right, I made a mistake when writing these docs. Thanks for fixing it!

Issue #7342, opened 2018-12-15
TypeError: Cannot read property 'listeners' of undefined


abhishekjain2604:
**Do you want to request a *feature* or report a *bug*?** 
I want to report a bug

**What is the current behavior?** 
I am getting an error ''TypeError: Cannot read property 'listeners' of undefined" while updating an object through custom method defined using schema.methods

**If the current behavior is a bug, please provide the steps to reproduce.**
```js
const mongoose = require('mongoose');
mongoose.Promise = global.Promise;

mongoose.connect('mongodb://localhost:27017', { dbName: 'test-db', useNewUrlParser: true }, function (error) {
  if (error) {
    console.log('Error in mongo connection', error);
  } else {
    console.log('Mongo connection established');
  }
});
mongoose.set('debug', true);

let catalogSchema = new mongoose.Schema({
  name: {
    type: String,
    required: true,
  },
  image: {
    type: new mongoose.Schema({
      mediaUrl: String,
      mediaType: String,
      thumbURL: String
    }),
    default: null  //TODO : Make this 'required'
  }

}, { runSettersOnQuery: true });

catalogSchema.methods.update = function (data) {
  for (let key in data) {
    this[key] = data[key];
  }
  return this.save();
};

let modelCatalog = mongoose.model('catalogue', catalogSchema);

let item = new modelCatalog({
  'name': 'Bikes',
});
item.save()
  .then((result) => {
    result.update({ name: 'Cars' });
  }).catch((error) => {
    console.log(error);
  });
```

**What is the expected behavior?**
It should have successfully updated the object.

Here is the stack trace that I am getting:

>/home/abhishek/Desktop/Test-project/node_modules/mongoose/lib/utils.js:264
        if (ee.listeners('error').length > 0 && !error[emittedSymbol]) {
               ^
TypeError: Cannot read property 'listeners' of undefined
    at /home/abhishek/Desktop/Test-project/node_modules/mongoose/lib/utils.js:264:16
    at /home/abhishek/Desktop/Test-project/node_modules/kareem/index.js:246:48
    at /home/abhishek/Desktop/Test-project/node_modules/kareem/index.js:135:16
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)

That's not all. If I go back a version(5.3.16), I am getting a different error. Here is the stack trace for that:

[...]

PR #7346, opened 2018-12-17
Check if EventEmitter is not null

MiloCasagrande:
* Fix for #7342.

Signed-off-by: Milo Casagrande <milo@foundries.io>


MiloCasagrande:
I tried running the tests locally, but I have 40 failing ones: not sure if this change needs a new test though.

The change I did is just mimicking the behaviour for the case where the passed callback is a function.

PR #7353, opened 2018-12-18
feat(objectid): add valueOf to ObjectId prototype

### LINKED ISSUE #7299: ObjectId.prototype.valueOf()
Might be handy:

```javascript
$ node
> const mongoose = require('mongoose')
undefined
> mongoose.Types.ObjectId.prototype.valueOf = function() { return this.toString() }
[Function]
> let id = new mongoose.Types.ObjectId()
undefined
> let id2 = new mongoose.Types.ObjectId(id.toString())
undefined
> id.valueOf()
'5c06e90e9f9d0a288edccbc6'
> id2.valueOf()
'5c06e90e9f9d0a288edccbc6'
> id === id2.toString()
false
> id == id2.toString()
true
```

Re: https://github.com/graphql/graphql-js/issues/1518#issuecomment-438362934

sibelius:
fix #7299
fix graphql/graphql-js#1518

**Summary**

this make interop of ObjectId and GraphQL easier

**Test plan**

validate valueOf works as expected
check tests here https://github.com/Automattic/mongoose/issues/7299

### DIFF
--- a/lib/types/objectid.js
+++ b/lib/types/objectid.js
@@ -24,4 +24,8 @@ Object.defineProperty(ObjectId.prototype, '_id', {
   }
 });
 
+ObjectId.prototype.valueOf = function() {
+  return this.toString();
+};
+
 module.exports = ObjectId;


sibelius:
not sure if tests failing because of my changes

vkarpov15:
Yeah the tests are failing because of your changes, I suspect because somewhere we check for `.valueOf()`. 

In 5.4.0 we introduced global getters for schema types, so you can do something like this:

```javascript

// Do this instead
mongoose.ObjectId.get(v => v.toString());

mongoose.set('debug', true);
run().catch(error => console.error(error.stack));

async function run() {
  await mongoose.connect('mongodb://localhost:27017/test', { useNewUrlParser: true });
  const Model = mongoose.model('Test', new mongoose.Schema({ name: String }));
  const docs = await Model.create([{ name: 'test1' }, { name: 'test2' }]);

  console.log(typeof docs[0]._id);

  const doc = new Model(docs[0].toObject());
  console.log(docs[0]._id === doc._id); // true
}
```

See [`SchemaType.get()`](https://mongoosejs.com/docs/api.html#schematype_SchemaType-get). I'm no expert on GraphQL, but the custom getter approach would be better on the Mongoose side because it doesn't pollute `ObjectId.prototype`.

I'll look into why the tests are failing in the meantime.

vkarpov15:
I'm going to close this since I got https://github.com/alexmcmillan/graphql-mongodb-query-fail-demo/pull/1 working without polluting `ObjectId.prototype`. Making this change is pretty risky - the test failure was due to one of our internal plugins checking for `valueOf()`. I imagine other plugins could run into similar issues.

Drop in this one-liner (requires Mongoose `>= 5.4.0`) and your GraphQL issues should go away:

```javascript
mongoose.ObjectId.get(v => v == null ? v : v.toString());
```

[See this for more info](http://thecodebarbarian.com/whats-new-in-mongoose-54-global-schematype-configuration.html).

PR #7361, opened 2018-12-23
update connections.html

JIBIN-P:
Summary:
I noticed the mongoose.connect() doesn't work properly without adding {useNewUrlParser: true} as an option. So I thought of updating this since many new users are facing unecessary trouble for this. This fix was actually recommended by @nax3t in the Q&A for the web developer bootcamp course in udemy. If this kind of PR is already rejected then please point me to the reason why if not then plzz merge it and also point me in the right direction if i can do anything more to improve this PR.

Screenshots:
Error Screenshot:
errorconnectmongoose

Solved Error Screenshot:
solvedconnectmongoose

Thank you.

### DIFF
--- a/docs/connections.html
+++ b/docs/connections.html
@@ -11,7 +11,7 @@
 </div>
 
 <p>You can connect to MongoDB with the <code>mongoose.connect()</code> method.</p>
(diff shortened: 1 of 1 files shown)


JIBIN-P:
Ok I will do this asap, thank you for guiding me.

PR #7365, opened 2018-12-24
Update lambda.jade with mongoose v5 connect method

raymondsiu:

If you're making a change to documentation, do **not** modify a `.html` file directly. Instead find the corresponding `.jade` file or test case in the `test/docs` directory.

**Summary**

Trying to use this code sample never worked for me, but by updating to the `connect()` method instead solved the issue.

**Examples**

N/A


raymondsiu:
@vkarpov15 I guess it could have been a one-off in my case, but once I switched to use the connect method instead, it seemed to solve my issue. I will try testing again to verify.

PR #7438, opened 2019-01-22
Fix broken link

shihabmridha:
The [(QueryCursor)](https://mongoosejs.com/docs/queries.html#streaming) hyperlink was pointing to something else which I think is not valid.


vkarpov15:
Thanks :+1:

Issue #7421, opened 2019-01-15
feature: option for validating only changed fields or a specified subset of fields 


captaincaius:
**Do you want to request a *feature* or report a *bug*?**

*feature*

**What is the current behavior?**

when using save() on a given object, validation is black-or-white - i.e. either all fields are validated (even unchanged ones) or none.

**If the current behavior is a bug, please provide the steps to reproduce.**

The use case:
Situations when you don't want to collect ALL field values for a document all in one place (e.g. multi-step forms, etc).  Especially in the case of onboarding a user, this poses a problem, because some of a document's other required fields may not even be set while working on that particular subset of fields, but you want to persist their progress to the database along the way.

**What is the expected behavior?**

The ability to tell save() to either:
- only validate updated fields
- only validate a specified subset of fields

**Workarounds and why they're not ideal**

1. bypass validation entirely: you WANT to use the power of the validation for the fields you are saving
2. use update instead of save: you lose hooks - plus, working at this level of abstraction is part of what makes mongoose awesome.
3. don't persist to the database along the way: users :-/ plus the extra work needed on the client-side.
4. split up the schema: IMO the data model shouldn't be driven by UX decisions as it makes changes a lot harder

**Please mention your node.js, mongoose and MongoDB version.**
node 10 / mongoose 5 / mongo 3.2

vkarpov15:
Not a bad idea. We support passing a list of paths to validateSync: https://mongoosejs.com/docs/api.html#document_Document-validateSync . But if you just want to work around making certain fields required depending on the user's status, you can use conditional required:

```
new Schema({
  status: String,
  approvedAt: { type: Date, required: function() { return this.status === 'APPROVED'; } }
});
```

Is that the problem you have or is there something else?

captaincaius:
Hmm I think I get what you mean - add a field for each subset of fields that submit together and make all their validations dependent on whether that field is set.

Thanks for the suggestion!  I'll try it.  It does couple the schema to UX a bit which I was hoping I could avoid, but it's the cleanest option I've got right now it seems.

I initially suggested this because I noticed in the code that something very similar is happening with isSelected(), but you pointed out validateSync and it seems its approach would be simpler (i.e. get all paths then filter).

Would this be a welcome PR for a new contributor?  Doesn't seem hard but I wouldn't want to add review burden without asking first ;).

vkarpov15:
[...]

PR #7492, opened 2019-02-06
feat(document): validateUpdatedOnly option. only validates modified

captaincaius:
schema option that can be overridden on validate, validateSync, and save

**Summary**

#7421 - ignore validation on unmodified paths

Abandoned yucky post-filtering approach in #7483.

- [x] Still unresolved: https://github.com/Automattic/mongoose/pull/7483#discussion_r253687434 - are you sure I should add it?

**Examples**

see tests ;)

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1731,7 +1731,7 @@ Document.prototype.validate = function(options, callback) {
     options = null;
   }
 
-  return utils.promiseOrCallback(callback, cb => this.$__validate(function(error) {
+  return utils.promiseOrCallback(callback, cb => this.$__validate(options, function(error) {
     cb(error);
   }), this.constructor.events);
 };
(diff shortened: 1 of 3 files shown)


captaincaius:
FYI for some reason I didn't see the problem I had with the hooks the first time I tried it "the right way".  I probably just had a mistake, but if anything pops into your head that I should check, LMK.

captaincaius:
Cool beans. Is there anything else you'd like done differently or added? Any other tests perhaps?

exentrich:
Unfortunately this option doesn't work.

I have this schema and options:
```js
const Schema = mongoose.Schema({
  title: String,
  coverId: Number
}, {
  validateModifiedOnly: true
}
```

When I try to update existing document `title` field, it fails because `coverId` field value is NaN (invalid). My use-case to ignore such validations.
Can you please help? Maybe I do something wrong.

PR #7515, opened 2019-02-12
Doc ValidationError

JulioJu:
Actually ValidationError seems to be an hash of CastError and ValidatorError. Therefore I make the proposal to say that in the doc. 

Proof:

```sh
$ grep "new ValidationError" * -R
document.js:    this.$__.validationError = new ValidationError(this);
helpers/query/castUpdate.js:  aggregatedError = aggregatedError || new ValidationError();
helpers/updateValidators.js:        const err = new ValidationError(null);
```

Furthermore, there I've seen a small typo error under  test/document.test.js therefore I propose to push the correction in this current PR.

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -1011,13 +1011,13 @@ describe('document', function() {
   });
 
   describe('Errors', function() {
-    it('MongooseErrors should be instances of Error (gh-209)', function(done) {
+    it('MongooseError should be instances of Error (gh-209)', function(done) {
       const MongooseError = require('../lib/error');
       const err = new MongooseError('Some message');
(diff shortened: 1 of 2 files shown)


test/document.test.js
```diff
@@ -1011,13 +1011,13 @@ describe('document', function() {
   });
 
   describe('Errors', function() {
-    it('MongooseErrors should be instances of Error (gh-209)', function(done) {
+    it('MongooseError should be instances of Error (gh-209)', function(done) {
```
vkarpov15: The test name as written was grammatically correct, but now it isn't. Both 'MongooseError' and 'instance' should be both singular or both plural.

Issue #7520, opened 2019-02-14
Populate getter not called with "getters: true"


nocksapp:
**Do you want to request a *feature* or report a *bug*?**
Bug considering the "getter" option on a ".populate" call.

**What is the current behavior?**
The "getter" option isn't working on populate with an embedded schema path like "bars.bar" 

**If the current behavior is a bug, please provide the steps to reproduce.**
```javascript
const mongoose = require('mongoose');
const db = 'GITHUB_ISSUE';
mongoose.connect(`mongodb://localhost:27017/${db}`);

const { Schema, model } = mongoose;

const BarSchema = new Schema({
  _id: {
    type: String,
  },
});

const FooSchema = new Schema({
  bar: {
    type: String,
    ref: 'Bar',
    get: (v) => {
      if (!v || typeof v !== 'string') {
        return v;
      }

      // Is called when populated with .populate({ path: 'bar', options: { getters: true } })
      console.log('called on populate');

      return v.slice(0, 4);
    },
  },
  bars: [new Schema({
    bar: {
      type: String,
      ref: 'Bar',
      get: (v) => {
        if (!v || typeof v !== 'string') {
          return v;
        }

        // Is not called when populated with .populate({ path: 'bars.bar', options: { getters: true } })
        console.log('not called on populate');

        return v.slice(0, 4);
      },
    },
  })],
});

const Bar = model('Bar', BarSchema);
const Foo = model('Foo', FooSchema);

const fooId = '5c657e2beca6210c78090082';

return Promise.all([
  Bar.create({ _id: '1234', description: 'first bar' }),
  Foo.create({ _id: fooId, bar: '123456', bars: [{ bar: '123456' }] }),
])
  .then(() => Foo.findOne({ _id: fooId })
    // Here the getter options is working as expected
    .populate({ path: 'bar', options: { getters: true } })
    // But when using a path to an embedded schema the getter won't get called
    .populate({ path: 'bars.bar', options: { getters: true } })
    .exec()
    .then((doc) => {
      // doc.bar is populated as expected
      console.log(doc.bar);
      // doc.bars[0].bar is null
      console.log(doc.bars[0].bar);
    }));
```

**What is the expected behavior?**
The expected behavior is that the "getter" function in bars.bar is called when doing a ".populate({ path: 'bars.bar', options: { getters: true } })"

**Please mention your node.js, mongoose and MongoDB version.**
Node: v10.15.1
Mongoose: 5.4.12
MongoDB: 3.6.5

PR #7521, opened 2019-02-14
Fixes populate "getter" option on embedded documents

nocksapp:
Fix for #7520

```javascript
// Will call getters for "bars.bar"
Foo.findOne({ _id: '5c657e2beca6210c78090082' }).populate({ path: 'bars.bar', options: { getters: true } });
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -4190,7 +4190,7 @@ function getModelsMapForPopulate(model, docs, options) {
 
     const localFieldPathType = modelSchema._getPathType(localField);
     const localFieldPath = localFieldPathType === 'real' ? modelSchema.paths[localField] : localFieldPathType.schema;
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -4188,7 +4188,8 @@ function getModelsMapForPopulate(model, docs, options) {
       foreignField = foreignField.call(doc);
     }
 
-    const localFieldPath = modelSchema.paths[localField];
+    const localFieldPathType = modelSchema._getPathType(localField);
+    const localFieldPath = localFieldPathType === 'real' ? modelSchema.paths[localField] : localFieldPathType.schema;
```
vkarpov15: There's an awful lot of test failures here because `localFieldPathType.schema` may not contain a `getters` array.

nocksapp: Added a check for the "getters" property. This was also the cause of the test failures

PR #7528, opened 2019-02-16
Added support for Maps for translateAliases, and logical query operators support.

chrischen:
**Summary**

- Added support for Maps passed in to translateAliases. Maps can be passed to the sort() call, which would need to be translated if aliases are used.

- Added initial support for query operators. Currently only logical query operators (can be nested too) such as $and, $or, $nor, and $not are officially supported. TBH not sure if this the best way to approach this as it's not straightforward to parse the schema and make sure all query operators are interpreted correctly. I've tried to keep the implementation as general as possible, and not hardcode any specific query operators.

**Examples**

SomeModel.find().sort(SomeModel.translateAliases(new Map([['someKey', 'someValue']])))


lib/model.js
```diff
@@ -1588,35 +1588,68 @@ Model.discriminators;
...
+  if (typeof fields === 'object') {
+    // Fields is an object (query conditions or document fields)
+    if (fields.set) {
+      // Infer that a Map was supplied
+      for (const [key, value] of new Map(fields)) {
+        fields = translate(key, value);
+      }
+    } else {
+      // Infer a regular object was supplied
+      for (const key in {...fields}) {
```
vkarpov15: We still support Node 4, 5, 6, and 7, so can't use rest operator. Not sure why you need to clone here, can you just do `for (const key of Object.keys(fields))`?

chrischen: I believe I cloned the object as iterating+modifying fields during iteration was causing some issues. Object.keys should work.

chrischen:
I pushed the new changes to my master branch. Do I need to redo this pull request?

lib/model.js
```diff
@@ -1588,35 +1588,68 @@ Model.discriminators;
...
+      else
+        // Delete from object
         delete fields[key]; // We'll be using the translated key instead
+    return fields;
+  }
+
+  if (typeof fields === 'object') {
+    // Fields is an object (query conditions or document fields)
+    if (fields.set) {
+      // Infer that a Map was supplied
```
vkarpov15: Why not just use `fields instanceof Map` instead of checking for `fields.set()`? The latter check seems very brittle

lib/model.js
```diff
@@ -1588,35 +1588,68 @@ Model.discriminators;
...
+      for (const [key, value] of new Map(fields)) {
```
vkarpov15: No [array destructuring](http://thecodebarbarian.com/an-overview-of-destructuring-assignments-in-node-js) in Node.js 4 or 5, so you'll need to do `for (const key of fields) {}`

lib/model.js
chrischen: My rationale is  that we depend on the set function anyways, and checking for set is cheaper than checking instanceof. Thoughts?

chrischen: Nevermind, checking for set will be problematic if a key used is called `set`. I would have to check it's a function anyways...  so I'll change this to instanceof.

lib/model.js
chrischen: Ok changed that.

PR #7541, opened 2019-02-20
Stub model for browser

caub:
It allows to import objects from a file using mongoose models, without having it crash

### DIFF
--- a/lib/browser.js
+++ b/lib/browser.js
@@ -128,10 +128,10 @@ exports.Document = DocumentProvider();
  *
  * @method model
  * @api public
- * @return {object} an empty object
+ * @return null
  */
 exports.model = function() {
-  return {};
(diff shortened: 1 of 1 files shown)


lib/browser.js
```diff
@@ -123,6 +123,17 @@ exports.utils = require('./utils.js');
...
 
+/**
+ * function stub for model
+ *
+ * @method model
+ * @api public
+ * @return {object} an empty object
+ */
+exports.model = function() {
+  return {};
```
vkarpov15: Why not `return null` or something like that? I'm hesitant to just return an empty object because then it will have an inconsistent and incompatible return type vs `model()` in server land.

caub: Yes, good idea, done

PR #7628, opened 2019-03-21
Minor documentation-corrections + fixed function-signature "Query.comment"

720degreeLotus:
This PR fixes some minor formating-issues within the documentation and, in addition to that, corrects the function-signature of "Query.prototype.comment()" to a `String` instead `Number`.

This is my first PR here, so please tell me, if there is anything that i can do better in the future.

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -799,7 +799,7 @@ Query.prototype.mod = function() {
  * @method comment
  * @memberOf Query
  * @instance
- * @param {Number} val
+ * @param {String} val
  * @see comment http://docs.mongodb.org/manual/reference/operator/comment/
  * @api public
  */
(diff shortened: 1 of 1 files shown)


720degreeLotus:
In addition to these fixes, i am not sure about the usage of the word "atomic" within the documentation.
There are at least 3 different formats used:
```
atomic
$atomic
`atomic`
```
I would suggest to use one format an stick to it to reduce confusion and i would go for just plain
```
atomic
```
. What are your thoughts on this?

720degreeLotus:
> In addition to these fixes, i am not sure about the usage of the word "atomic" within the documentation.
> There are at least 3 different formats used:
> 
> ```
> atomic
> $atomic
> `atomic`
> ```
> I would suggest to use one format an stick to it to reduce confusion and i would go for just plain
> 
> ```
> atomic
> ```
> . What are your thoughts on this?

@vkarpov15 What are your thoughts on this?

vkarpov15:
@720degreeLotus I agree, sounds good :+1: feel free to put in a PR

Issue #7634, opened 2019-03-23
wrong partialFilterExpression index with discriminators


egorovli:
**Do you want to request a *feature* or report a *bug*?**

A bug.

**What is the current behavior?**

The docs say it is possible to use three parameters to Model.discriminator(). If the third one is specified, it is used as a value for the discriminator. However the generated index in MongoDB uses the first parameter (the model name).

Consider the following example:

```js
let BasePaymentSchema = new mongoose.Schema({
  amount: mongoose.Types.Decimal128
}, {
  discriminatorKey: 'provider'
})

let BasePayment = mongoose.model('Payment', BasePaymentSchema)

let PaypalPaymentSchema = new mongoose.Schema({
  transactionId: {
    type: String,
    unique: true
  }
})

let PaypalPayment = BasePayment.discriminator(
  'PaypalPayment',
  PaypalPaymentSchema,
  'paypal' // This value should be in discriminator field
)
```

But if I do `db.payments.getIndexes()` I get:

```json
{
  "v": 2,
  "unique": true,
  "key": {
    "transactionId": 1
  },
  "name": "transactionId_1",
  "ns": "my-database.payments",
  "background": true,
  "partialFilterExpression": {
    "provider": "PaypalPayment"
  }
}
```

So basically the model name (`PaypalPayment`) is used as a `partialFilterExpression` instead of the supplied discriminator value (`paypal`). This way the `unique` constraint doesn't work at all.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Node 11.12.0
Mongoose 5.4.19
MongoDB 4.0.6

PR #7635, opened 2019-03-24
fix modelName being used as value to partialFilterExpression index

egorovli:
**Summary**

The issue is described here: #7634. If the third parameter is used in a `Model.discriminator()` method it should be used as a value to `partialFilterExpression` index instead of the `model.modelName`.

**Examples**

Consider the following example:

```js
let BasePaymentSchema = new mongoose.Schema({
  amount: mongoose.Types.Decimal128
}, {
  discriminatorKey: 'provider'
})

let BasePayment = mongoose.model('Payment', BasePaymentSchema)

let PaypalPaymentSchema = new mongoose.Schema({
  transactionId: {
    type: String,
    unique: true
  }
})

let PaypalPayment = BasePayment.discriminator(
  'PaypalPayment',
  PaypalPaymentSchema,
  'paypal' // This value should be in discriminator field
)
```

Before the fix `db.payments.getIndexes()` returns:

```json
{
  "v": 2,
  "unique": true,
  "key": {
    "transactionId": 1
  },
  "name": "transactionId_1",
  "ns": "my-database.payments",
  "background": true,
  "partialFilterExpression": {
    "provider": "PaypalPayment"
  }
}
```

After the fix:

```json
{
  "v": 2,
  "unique": true,
  "key": {
    "transactionId": 1
  },
  "name": "transactionId_1",
  "ns": "my-database.payments",
  "background": true,
  "partialFilterExpression": {
    "provider": "paypal"
  }
}
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1494,8 +1494,14 @@ function _decorateDiscriminatorIndexOptions(model, indexOptions) {
   if (model.baseModelName != null && indexOptions.unique &&
       !('partialFilterExpression' in indexOptions) &&
       !('sparse' in indexOptions)) {
+    
+    const value = (
+      model.schema.discriminatorMapping &&
+      model.schema.discriminatorMapping.value
+    ) || model.modelName
+
     indexOptions.partialFilterExpression = {
-      [model.schema.options.discriminatorKey]: model.modelName
+      [model.schema.options.discriminatorKey]: value
     };
   }
   return indexOptions;


lib/model.js
```diff
@@ -1494,8 +1494,14 @@ function _decorateDiscriminatorIndexOptions(model, indexOptions) {
   if (model.baseModelName != null && indexOptions.unique &&
       !('partialFilterExpression' in indexOptions) &&
       !('sparse' in indexOptions)) {
+    
+    const value = (
+      model.schema.discriminatorMapping &&
```
egorovli: I'm not really familiar with `mongoose` internals, maybe this `&&` check isn't necessary.

PR #7690, opened 2019-04-05
Allow passing of options to a method similar to Model.create(...)

giovannilobitos:
Model.create currently doesn't (can't?) support passing of options. There will be times when a developer needs to pass options like validateBeforeSave when creating a new document.

This pull request introduces a new method on Model called createWithOptions. This is similar to Model.create, except this method allows and requires passing of options.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2869,6 +2869,117 @@ Model.create = function create(doc, options, callback) {
   }, this.events);
 };
 
+
+/**
+ * Another shortcut (aside from Model.create) for saving one or more documents to the database.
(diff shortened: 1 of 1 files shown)


giovannilobitos:
Oh, yes. I missed that one. But I’m still not so sure about converting your data into array (given that I only need to create one document) just to pass the options.

vkarpov15:
It is a bit annoying, but that's the only workaround now because we allow passing a spread of documents to `Model.create()`, like `Model.create(doc1, doc2, doc3)`. If the first arg isn't an array, Mongoose has no way of knowing whether the 2nd arg is options or a doc to save.

vkarpov15:
Thanks for the contribution though :+1:

Issue #6354, opened 2018-04-16
Model includes() does not works as indexOf(). 


mapo-job:
**Report *bug***

**What is the current behavior?**
Model includes() does not works as indexOf().  As define here: [Includes  ES6 ](https://nodesource.com/blog/six-of-the-most-exciting-es6-features-in-node-js-v6-lts/#checkingarrayvalueswitharrayincludes)

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Use these schema:
```
var User = new Schema({
    name:{
        type: String,
        default: ''
    }
});

var relationSchema = new Schema({
    user:  { type: mongoose.Schema.Types.ObjectId, ref: 'User' },
    friends: [ { type: mongoose.Schema.Types.ObjectId, ref: 'User' } ]
});
```

2. Insert doc in DB

User:
```
{ 
     "_id" : ObjectId("5acd4da4c271813c48bcd843"), 
    "name" : "Alice"
},
{
    "_id" : ObjectId("5ad4eb200a67063840c158d5"), 
    "name" : "Bob"
}
```
Relation:
```
{ 
    "_id" : ObjectId("5acd7fa7f7be5030fc626dd0"), 
    "friends" : [
        ObjectId("5ad4eb200a67063840c158d5")
    ], 
    "user" : ObjectId("5acd4da4c271813c48bcd843"), 
}
```

3. Test the method.
```
let userId= "5acd4da4c271813c48bcd843";
let friendId= "5ad4eb200a67063840c158d5";

//Relation contains the friendId in friends
Relations.findOne({user: userId})
    .then((rel) => {
        consloe.log(rel.friends.includes(friendId)); // false
        console.log(rel.friends.indexOf(friendId)); //0
```

**What is the expected behavior?**
indexOf and includes must return the same results

**Please mention your node.js, mongoose and MongoDB version.**
Node v8.9.1
Mongoose v5.0.14
MongoDb v3.4.4

mapo-job:
Hi @lineus,
    Thanks for the clarification, I understand that adding `MongooseArray.includes()` would be a backward compatibility issue and prefer if it's in a major release. Until then, I'll use indexOf...  

I try with ObjectID but the object is not the same so it does not work.

```
console.log(relation.friends.indexOf(user1.id)); //1
console.log(relation.friends.indexOf(new ObjectID(user1.id))); //1

console.log(relation.friends.includes(user1._id)); // true note the use of _id
[...]

PR #7732, opened 2019-04-23
feat: support MongooseArray#includes

hansemannn:
**Summary**

Expose the `MongooseArray#includes` method (fixes https://github.com/Automattic/mongoose/issues/6354).

**Examples**

See `types.array.test.js` for details :-)

### DIFF
--- a/lib/types/array.js
+++ b/lib/types/array.js
@@ -798,16 +798,8 @@ MongooseArray.mixin = {
    * @memberOf MongooseArray
    */
 
-  includes: function indexOf(obj) {
-    if (obj instanceof ObjectId) {
-      obj = obj.toString();
-    }
-    for (let i = 0, len = this.length; i < len; ++i) {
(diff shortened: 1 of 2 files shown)


lib/types/array.js
```diff
@@ -786,6 +786,28 @@ MongooseArray.mixin = {
...
+   * Return whether or not the `obj` is included in the array.
+   *
+   * @param {Object} obj the item to check
+   * @return {Boolean}
+   * @api public
+   * @method includes
+   * @memberOf MongooseArray
+   */
+
+  includes: function indexOf(obj) {
```
Fonger: You forget to rename the function and I think it's better to directly return `this.indexOf(obj) !== -1` in this function for simplicity like `addToSet` and `pull` do.

hansemannn: @Fonger updated, thanks!

PR #7736, opened 2019-04-24
fix(query): pass correct callback for _legacyFindAndModify

### LINKED ISSUE #7734: findByIdAndUpdate behavior changed from 5.5.2 to 5.5.3
There is a behavioural change between mongoose 5.5.2 and 5.5.3 - I want to know if this is intentional or a bug.

We used to rely on the following pattern:

```
let obj = myModel.findByIdAndUpdate('some_id', {foo: "bar"}, {new: true});
await doSomethingWithTheObject(obj);
return obj.set('processed', true).save();
```

Starting from 5.5.3, last row throws an error - .set is not a function. Also virtual field obj.id does not exist. It seems, that `findByIdAndUpdate` now returns lean object instead of mongoose object. At the same time, docs have not changed and it still should return Query object unless "lean" parameter is used (which is not the case currently).

Tried (and managed to reproduce) with  NodeJS version 8.15.1 and latest 10.15.3 with MongoDB version 4.0.8.

### LINKED ISSUE #7735: Bug: raw documents returned in v5.5.3
The following is a breaking change from v5.5.2

```js

const FooSchema = new mongoose.Schema({
  foo: String
});

FooSchema.methods.updateUser = async function () {
};
FooSchema.statics.create = async function (foo) {
  const Model = this;

  const o = await Model.findOneAndUpdate({ foo: 'qux' }, { foo: 'bar' }, { upsert: true, new: true });
  console.log(typeof o.updateUser);
  // outputs 'function' with v5.5.2
  // outputs 'undefined' with v5.5.3
  // here o.updateUser is undefined instead of being a function (the method defined above)
};
const Foo = mongoose.model('Foo', FooSchema);

(async () => {
  try {
    await mongoose.connect('mongodb://localhost:27017/foo-test');
    await Foo.create();
  } catch (e) {
    console.error(e)
  }
  mongoose.disconnect();
})()
```

The same problem happens without async/away (with callbacks for all functions above)

@vkarpov15 I think you might have an idea about the changes in v5.5.3

Fonger:
**Summary**

Fix breaking changes introduced in 5.5.3
The callback function was `cb`, in 0b849710508566e86194a925f1248a2c3d200f16 became `callback`

```diff
-  this._collection.findAndModify(castedQuery, castedDoc, opts, _wrapThunkCallback(_this, -function(error, res) {
-    return cb(error, res ? res.value : res, res);
-  }));
!  _legacyFindAndModify.call(_this, castedQuery, castedDoc, opts, callback);
```

This PR simply changes `callback` to `cb`
```diff
!  _legacyFindAndModify.call(_this, castedQuery, castedDoc, opts, callback);
+  _legacyFindAndModify.call(_this, castedQuery, castedDoc, opts, cb);
```

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -3419,7 +3419,7 @@ Query.prototype._findAndModify = function(type, callback) {
       if (error) {
         return callback(error);
       }
-      _legacyFindAndModify.call(_this, castedQuery, castedDoc, opts, callback);
+      _legacyFindAndModify.call(_this, castedQuery, castedDoc, opts, cb);
     };
 
     try {
@@ -3428,7 +3428,7 @@ Query.prototype._findAndModify = function(type, callback) {
       callback(error);
     }
   } else {
-    _legacyFindAndModify.call(_this, castedQuery, castedDoc, opts, callback);
+    _legacyFindAndModify.call(_this, castedQuery, castedDoc, opts, cb);
   }
 
   return this;


vkarpov15:
Thanks!

Issue #7800, opened 2019-05-14
Issue with $regex/$options when using field option "uppercase"


mathpere:
Hello,

I'm using the latest version of mongoose. I encounter an issue with $regex/$options when using field option `uppercase`.

With the following model: 

```
const schema = new mongoose.Schema({
  myField: {type: String, uppercase: true}
});

const Model = mongoose.model('Car', schema);
```

When I query like this: 
```js
const model = await Model.findOne({myField: {$regex: 'abc', $options: 'i'}});
```
I get this exception: 

```js
{message: 'MongoError: Feature not supported: regexp option: I'
ok: 0
errmsg: 'Feature not supported: regexp option: I'
code: 303
name: 'MongoError'}
```
When I remove the option `uppercase: true`, this works:

```js
const schema = new mongoose.Schema({
  myField: {type: String}
});
```
I took a look in the code, and I saw `$options` value was taken into account with `castForQuery` and I'm not sure why. Maybe is it related?

PS: I'm using AWS DocumentDB which doesn't support `{$options: 'I'}` instead of `{$options: 'i'}`

PR #7802, opened 2019-05-14
Run string cast on $options

Fonger:
**Summary**

Apply the fix in #1462 $conditionalHandlers for String schema #7800 

https://github.com/Automattic/mongoose/blob/c08002049442d0f87c99819529177f9ba148048b/lib/schema/array.js#L439-L442

Maybe we can throw an error if `$options` is not of type string.

**Examples**

See test

Close&Reopen to re-trigger travis-ci test.

### DIFF
--- a/lib/schema/string.js
+++ b/lib/schema/string.js
@@ -577,7 +577,7 @@ SchemaString.prototype.$conditionalHandlers =
       $gte: handleSingle,
       $lt: handleSingle,
       $lte: handleSingle,
-      $options: handleSingle,
+      $options: String,
       $regex: handleSingle,
       $not: handleSingle
     });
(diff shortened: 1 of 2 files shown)


lib/schema/string.js
```diff
@@ -577,7 +577,7 @@ SchemaString.prototype.$conditionalHandlers =
       $gte: handleSingle,
       $lt: handleSingle,
       $lte: handleSingle,
-      $options: handleSingle,
```
vkarpov15: Why is this change necessary? `handleSingle()` will cast it to a string, no?

Fonger: handleSingle() will cast it to a string, but also apply uppercase/lowercase setter to reproduce #7800.
Change to String here like the way `SchemaArray` handles `$options`

Issue #7812, opened 2019-05-16
Split API docs into separate pages to make docs more google-able


vkarpov15:
Re: https://github.com/Automattic/mongoose/issues/7519

Fonger:
Jade has been renamed to pug and was deprecated 4 years ago. Consider migrating to pug?

vkarpov15:
Done :+1: had a couple issues upgrading but managed to get us on the latest version of pug

PR #7864, opened 2019-06-03
Add height: 100% to #menu

robsonniemeyer:
mongoose1
mongoose2

**Summary**

This issue is causing the menu to not fill the entire page height.


robsonniemeyer:
Did you clear the cache?

vkarpov15:
This is likely because of your ad blocker. Thanks for pointing this out, didn't realize that an ad blocker would leave a bunch of dangling whitespace. We'll fix this as part of our work on #7812 since we're already doing some layout work on the docs.

PR #7885, opened 2019-06-11
[5.6] update variables and docs to use 'pug'

Fonger:
**Summary**

- This targets branch 5.6. re: #7880
- Update docs to use `pug`
- Update variables to use `pug`

### DIFF
--- a/website.js
+++ b/website.js
@@ -2,7 +2,7 @@
 
 const acquit = require('acquit');
 const fs = require('fs');
-const jade = require('pug');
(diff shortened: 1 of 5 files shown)


website.js
```diff
@@ -2,7 +2,7 @@
 
 const acquit = require('acquit');
 const fs = require('fs');
-const jade = require('pug');
```
vkarpov15: Boo, I was hoping I could get away with this one :rofl:

PR #7890, opened 2019-06-12
Fix #7889

### LINKED ISSUE #7889: Value pushed to array not persisted

Bug

If you push a value to an array, populate the said array, push nothing to the said array, and finally save the document, the originally pushed value is not persisted to the database.

```js
mongoose.connect("mongodb://localhost:27017/array-empty-push", {useNewUrlParser: true});

console.log(`Mongoose ${mongoose.version}`);

const Car = mongoose.model("Car", new mongoose.Schema({
  model: Number
}));

const Player = mongoose.model("Player", new mongoose.Schema({
  cars: [{type: mongoose.Schema.Types.ObjectId, ref: "Car"}]
}));

async function test() {
  await Player.deleteMany({});
  await Car.deleteMany({});

  let car = await Car.create({model: 0});
  await Player.create({cars: [car._id]});
  let player = await Player.findOne({});
  await player.populate("cars").execPopulate();

  console.log(player.cars.length);

  car = await Car.create({model: 1});
  player.cars.push(car);
  await player.populate("cars").execPopulate();

  car = await Car.create({model: 2});
  player.cars.push(car);

  console.log(player.cars.length);

  await player.save();
  player = await Player.findOne({});

  console.log(player.cars.length);

  mongoose.disconnect();
}

test();
```

The three numbers this script should print are 1, 3, 3. The actual values it prints are 1, 3, 2.

* Node.js 12.4.0
* Mongoose 5.5.14
* MongoDB driver 3.2.7
* MongoDB 3.4.6

jyrkive:
This pull request fixes bug #7889.

I'd like to also write a unit test for it, but I don't really know what would be the right place to put the test. Any suggestions?

### DIFF
--- a/lib/types/array.js
+++ b/lib/types/array.js
@@ -349,6 +349,10 @@ MongooseArray.mixin = {
       undefined, { skipDocumentArrayCast: true });
     const ret = [].push.apply(this, values);
 
+    if (values.length === 0) {
+      return ret;
+    }
+
     this._registerAtomic('$push', values);
     this._markModified();
     return ret;


jyrkive:
Actually, now when I have had the lunch break to think, this isn't the right fix. This solves the problem when pushing nothing, but not if I push an actual value.

I'm afraid it's necessary to preserve _atomics when populating.

PR #7905, opened 2019-06-17
feat(mongoose): throw an error if calling `mongoose.connect()` multiple times while connected

### LINKED ISSUE #7813: Saving not working in local database, works in remote database
local: mongoose@5.4.13, running on MacOs Mojave 10.14.4
remote: mongoose@5.5.8, running on Linux Debian Jesse 8

Report a bug.

Save fails in the local database but works in the remote database.

In the remote database, the record gets saved without a problem and added to the database.
In the local database, the _id for the document is not generated, and I get the following error:
```
{ MongooseError: document must have an _id before saving
  at new MongooseError (/path/to/api/node_modules/mongoose/lib/error/mongooseError.js:13:11)
 at Timeout._onTimeout (/path/to/api/node_modules/mongoose/lib/model.js:257:18)
 at listOnTimeout (timers.js:327:15)
 at processTimers (timers.js:271:5)
 message: 'document must have an _id before saving',
 name: 'MongooseError' }
```

This is a gist reproducing the error:
https://gist.github.com/DigitalLeaves/06b514f19ef43e74b5979f6e0d099440

It should save both documents in both databases.

Remote:
Node.js v11.10.0
mongoose@5.5.8, running on Linux Debian Jesse 8
MongoDB shell version v3.4.20
git version: 447847d93d6e0a21b018d5df45528e815c7c13d8
OpenSSL version: OpenSSL 1.0.1t  3 May 2016
allocator: tcmalloc
modules: none
build environment:
    distmod: debian81
    distarch: x86_64
    target_arch: x86_64

Local:
Node.js v11.10.0
mongoose@5.4.13, running on MacOs Mojave 10.14.4
MongoDB shell version v3.6.3
git version: 9586e557d54ef70f9ca4b43c26892cd55257e1a5
OpenSSL version: OpenSSL 1.0.2o  27 Mar 2018
allocator: system
modules: none
build environment:
    distarch: x86_64
    target_arch: x86_64

### LINKED ISSUE #7814: Throw an error if calling `mongoose.connect()` multiple times while connected
   Bug
    When  a document with a field `username` is updated twice(with another field `token` changed),
    I get a validation error with message  path `username` validation is invalid. Note that it happens alternately.

It shouldn't give any validation error

Node.js v8.10.0 mongoose v5.5.9 mongodb v4.0.9

More detailed issue can be found [here](https://stackoverflow.com/questions/56145471/how-to-fix-mongoose-giving-validation-error-on-alternate-doc-save-execution-w) on stackoverflow.

Thanks!

Fonger:
**Summary**

This may prevent developers from the misuse of mongoose. Resolves #7814.

EDIT: sorry for the wrong issue reference (not 7813)

### DIFF
--- a/docs/connections.pug
+++ b/docs/connections.pug
@@ -153,7 +153,7 @@ block content
 
     // Or using promises
     mongoose.connect(uri, options).then(
-      () => { /** ready to use. The `mongoose.connect()` promise resolves to undefined. */ },
+      () => { /** ready to use. The `mongoose.connect()` promise resolves to mongoose instance. */ },
       err => { /** handle initial connection error */ }
     );
     ```
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -726,11 +726,11 @@ Connection.prototype._close = function(force, callback) {
   this._closeCalled = true;
 
   switch (this.readyState) {
-    case 0: // disconnected
+    case STATES.disconnected:
       callback();
       break;
 
-    case 1: // connected
+    case STATES.connected:
       this.readyState = STATES.disconnecting;
       this.doClose(force, function(err) {
         if (err) {
@@ -741,13 +741,13 @@ Connection.prototype._close = function(force, callback) {
       });
 
       break;
-    case 2: // connecting
+    case STATES.connecting:
       this.once('open', function() {
         _this.close(callback);
       });
       break;
 
-    case 3: // disconnecting
+    case STATES.disconnecting:
       this.once('close', function() {
         callback();
       });
--- a/lib/index.js
+++ b/lib/index.js
@@ -287,6 +287,9 @@ Mongoose.prototype.createConnection = function(uri, options, callback) {
 Mongoose.prototype.connect = function() {
   const _mongoose = this instanceof Mongoose ? this : mongoose;
   const conn = _mongoose.connection;
+  if (conn.readyState !== STATES.disconnected) {
+    throw new _mongoose.Error('You can not `mongoose.connect()` multiple times while connected.');
+  }
   return conn.openUri(arguments[0], arguments[1], arguments[2]).then(() => _mongoose);
 };
 
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -679,7 +679,28 @@ describe('mongoose module:', function() {
     const Model = mongoose.model('gh6813', new Schema({ name: String }));
     assert.ok(Model);
   });
+  it('throws an error if calling `mongoose.connect()` multiple times while connected (gh-7814)', function() {
+    const mong = new Mongoose();
+    return co(function*() {
+      const errorRegex = /multiple times while connected/;
 
+      // throws while connecting
+      const connectPromise = mong.connect(uri, options);
+      assert.throws(() => mong.connect(uri, options), errorRegex);
+
+      // throws after connected
+      yield connectPromise;
+      assert.throws(() => mong.connect(uri, options), errorRegex);
+
+      // throws while disconnecting
+      const disconnectPromise = mong.disconnect();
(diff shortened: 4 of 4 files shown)


sarora2073:
This fix to prevent multiple connections ignores valid scenarios where we may want multiple separate connections to the DB e.g. multiple background processing connections that may intentionally have the own separate connections setup e.g. so they can be moved around servers. I hope this "fix" will be modified e.g. a parameter like "allowMultipleConnections", if it doesn't already exist.

dankadmin:
I think that overall, this is a very good thing, and it can help isolate some unexpected problems. However, I think I would agree that having a option to create multiple connections might be nice. For instance, while running asynchronous unit tests, where I'm now getting the error.

I currently use mocha and have one unit test file per model, and each is opening its own connection and closing it when the tests are done, causing them to potentially overlap. Is it bad form to run simultaneous tests like this, or could we have an option to allow such uses to override this check?

Fonger:
@saroza2073 @dankadmin

I think in that case, you should call mongoose.createConnection instead.

sarora2073:
@Fonger - thanks! Didn’t know that was an option.

DoubleHub:
Seriously!?!?!? You did this breaking change in a MINOR patch!?!?!? Oh my god!!!!!

DoubleHub:
Seriously, you did this in a MINOR patch!?

Fonger:
Sorry for the trouble and breaking changes.

This change has been reverted ca243778d3b261c063e975d2002714212c65d2d0 and probably will be released as 5.6.5 shortly.

However, this changes may be applied in mongoose 6 again because **calling mongoose.connect multiple times is still a misuse and it may lead to potential bugs** (even if it worked).

If your code rely on mongoose.connect multiple times, it's highly recommended to update your code asap.

sarora2073:
FWIW - i almost freaked out at first when i thought i wasn't going to be able to use mongoose for multiple connections due to this change BUT as per above discussion really this change is just enforcing the intended usage of connect vs createConnection as detailed in docs and here: https://stackoverflow.com/questions/40818016/connect-vs-createconnection

To that end, it shouldn't be consider a "breaking change" so much as an attempt to enforce expected usage, even though as a practical matter it may require you to update your code to use the correct connection method, if you have multiple connections intentionally.

dankadmin:
It is true that this is not removing any previously intended functionality, and it is only enforcing the use of the software as it has been intended, and I even found a mention of wanting to enforce this behavior from back in 2012.

On the other hand, most, in not all, of the tutorials and documentation that I can find on the web uses the single connection method. That's no surprise, though, considering that even the official documentation almost exclusively seems to use mongoose.connect. I understand that it's the most simple thing to use, but I believe that many professional projects are going to need multiple connections, if for nothing more than asynchronous unit testing.

It's not a trivial change in code to switch, considering that you have to modify the way your models work to ensure that you are using the correct connection, and there is a lot of code re-factoring.  So, I would recommend that if you are going make this change, please also update the official documentation to use mongoose.createConnection for the majority of the examples, including the way that models are used. Then, maybe future tutorials on the web will begin to update as well to make it easier on new users.

vkarpov15:
There's no problem with using `mongoose.connect()`, that's sufficient for most apps. The problem is only with calling `mongoose.connect()` multiple times, which is likely to cause nasty surprises - what happens if you call `mongoose.connect()` in the middle of a `populate()`?

I've had a couple people reach out saying they used multiple `mongoose.connect()` calls as an alternative to [`useDb()`](https://mongoosejs.com/docs/api/connection.html#connection_Connection-useDb) because they weren't aware of `useDb()`. Is there another reason to call `mongoose.connect()` multiple times?

PR #7955, opened 2019-07-03
docs: add async/await guide

Rossh87:

**Summary**

Resolves some confusion re: async/await patterns with Mongoose, and particularly the difference between Queries and native Promises.

**Examples**

Side nav now has a link under 'Guides' that points to an Async/Await guide.


docs/async-await.pug
```diff
@@ -0,0 +1,161 @@
...
+            handleError(err);
+        }
+    }
+    ```
+  :markdown
+    Note that the specific fulfillment values of different Mongoose methods vary, and may be affected by configuration. Please refer to the [API documentation](./api.html) for information about specific methods.
+
+  h3(id="async-functions") Async Functions
+  :markdown
+    All JavaScript functions earmarked with the keyword ‘async’ must return a native JavaScript promise that will fulfill with whatever value the async function returns.  It is never possible to synchronously return a value from an async function, even if that value is awaited in the function body.
```
vkarpov15: Minor point, but this could be read as saying that you as the person writing JS code are responsible for returning a promise from an async function, which is incorrect. [There's a distinction between the "resolved value" and the return value of an async function](http://thecodebarbarian.com/async-functions-in-javascript.html#an-async-function-always-returns-a-promise) that would be handy to clarify here.

docs/async-await.pug
```diff
@@ -0,0 +1,161 @@
...
+
+        console.log(isPromise(query));  //false
+
+        const awaitedValue = await query;
+
+        query.then(chainedValue => console.log(chainedValue === awaitedValue)); //false
+    }
+    ```
+  :markdown
+    You are most likely to accidentally re-execute queries in this way when mixing callbacks with async/await.  This is never necessary and should be avoided.  If for some reason you need a Query to return a genuine promise, [Query.prototype.exec](./api/query.html#query_Query-exec) can meet that need.
```
vkarpov15: We like to avoid comments like "for some reason", they tend to sound pedantic and don't convey any relevant information. I'd suggest using "If you need a fully fledged promise instead of a thenable, you should use `Query#exec()`."

docs/async-await.pug
```diff
@@ -0,0 +1,161 @@
...
+    function isPromise(thenable) {
+        return Promise.resolve(thenable) === thenable;
+    };
+
+    // The fulfillment value of the promise returned by user.save() will always be the same,
+    // regardless of how, or how often, we observe it.
+    async function observePromise() {
+        const user = await User.findOne({firstName: 'bill', lastName:'brasky'});
+
+        user.deceased = true;
```
vkarpov15: Minor nitpick, but we try to avoid examples that reference sensitive topics. You could reuse the FDR example above.

docs/async-await.pug
```diff
@@ -0,0 +1,161 @@
...
+
+    ```javascript
+    // Idiomatic way to determine if a thenable is a genuine promise
+    function isPromise(thenable) {
+        return Promise.resolve(thenable) === thenable;
```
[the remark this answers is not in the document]
Rossh87: I believed the `Promise.resolve` implementation was more reliable.  After some reading/testing, I realize I was wrong ;)  I'll change this, good catch!

docs/async-await.pug
Rossh87: Should be less ambiguous now.  See what you think.

docs/async-await.pug
Rossh87: Great suggestion, done.

Rossh87:
I've believe I addressed all the changes.  Sorry for the messy commit history, if it's a problem I can reset and clean it up in a different PR.  Thanks!

docs/async-await.pug
```diff
@@ -0,0 +1,192 @@
...
+      //Await the promise returned from calling getUser.
+      const user = await getUser();
+
+      console.log(user);  //Prints '{name: 'bill', admin: false}'
+    }
+
+  h3(id="queries") Async/Await with Mongoose Queries
+  :markdown
+    Under the hood, [*async/await* is syntactic sugar](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await)
+<<<<<<< HEAD
```
vkarpov15: Looks like there's some merge conflicts in async-await.pug? Can you please clean those up? Otherwise, should be good to merge.

Rossh87: This should be cleared up.  However, due to a botched git workflow, this PR has an issue with the Makefile and possibly other problems.  I opened a [PR for a new branch of this repo](https://github.com/Automattic/mongoose/pull/7969) that includes all suggested changes and builds correctly.  Should be a simple fast-forward onto master.  Sorry for the confusion.  Thanks!

PR #8004, opened 2019-07-22
fix(query): add missing options for deleteOne and deleteMany in Query

Fonger:
**Summary**

Apply the same fix #6810 #7860 for Query instance. They were only for Model.

This will solve @krlozadan's issue. It seems he gets the model via Query instance.
See https://github.com/Automattic/mongoose/pull/6810#issuecomment-513302970


Fonger:
Wow you merge it so fast xD :100:

PR #8037, opened 2019-07-30
perf(documentarray): refactor to use ES6 classes instead of mixins, 30% speedup

vkarpov15:
Before: 

```
$ time node gh-7895.js 

real	0m10.777s
user	0m11.311s
sys	0m0.354s
```

After:

```
$ time node gh-7895.js 

real	0m7.058s
user	0m7.194s
sys	0m0.048s
```

Should be fine to release in 5.6.8, but I want to see if we can make this a little bit faster, because this is still slightly slower than Mongoose 4.13

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -260,9 +260,7 @@ Model.prototype.$__handleSave = function(options, callback) {
     this.$__version(true, obj);
     this[modelCollectionSymbol].insertOne(obj, saveOptions, function(err, ret) {
       if (err) {
-        _this.isNew = true;
-        _this.emit('isNew', true);
-        _this.constructor.emit('isNew', true);
+        _setIsNew(_this, true);
 
         callback(err, null);
         return;
@@ -271,9 +269,7 @@ Model.prototype.$__handleSave = function(options, callback) {
       callback(null, ret);
     });
     this.$__reset();
-    this.isNew = false;
-    this.emit('isNew', false);
-    this.constructor.emit('isNew', false);
+    _setIsNew(this, false);
     // Make it possible to retry the insert
     this.$__.inserting = true;
   } else {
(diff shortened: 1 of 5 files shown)


vkarpov15:
It looks like I managed to fix the tests by removing event emitter registration in the doc array constructor - it isn't strictly necessary because we can just set `isNew` normally. However, this means I'd like to defer this change to 5.7.0. It shouldn't be backwards breaking, but it feels a little too risky for a patch release.

brettkiefer:
This looks great! I found something kind of funny running our test suite against it that I thought I'd alert you to. I'm not consider if you consider it to be a bug in `util.inspect`, in this change, or neither.

If you run this script on 336e995ab (with node 10 and 12, at least), you'll get infinite recursion in `inspect`'s `formatValue` and ultimately `RangeError: Maximum call stack size exceeded`, whereas with e.g. 5.6.5, it would inspect the subdoc array as expected.

```
'use strict';

const mongoose = require('.');
const { inspect } = require('util');

const SubdocSchema = mongoose.Schema({ a: String, b: String, c: String, d: String, e: String  });
const Schema = mongoose.Schema({ name: String, subdocs: { type: [SubdocSchema] } });
const Model = mongoose.model('Test', Schema);
const data = { _id: new mongoose.Types.ObjectId(), name: 'test', subdocs: [{a: 'a'}] };
const model = new Model();
model.init(data);
console.log(inspect(model.subdocs));
```


That may be something you'd expect consumers to work around, but I thought it was worth calling out.

vkarpov15:
Thanks @brettkiefer , I will look into this before merging

brettkiefer:
To close the loop, https://github.com/Automattic/mongoose/commit/f96a04132635b78d36897b12031c5a069271f97c fixed the issue in our test suite that initially brought the infinite recursion issue up. Thank you!

PR #8070, opened 2019-08-14
Update handleImmutable.js

rich-earth:
check schemaType.options not undefined

**Summary**
I'm hitting a problem where queries fail with a Schema that has arrays of mixed objects, for some reason throws an error here when tries to find "immutable" of undefined.

Thanks


rich-earth:
I should add this only happens after upgrading from 5.3 to latest...

PR #8089, opened 2019-08-21
Fix issue when using find in a collection with nested schema array

felipeplets:
**Summary**

When casting an Schema type array of objects the `applySetters` is called with no scope leading `immutableSetter` to have this as null and thus breaking at line 20 with the following error:

`Cannot read property '$__' of undefined`

This issue was introduced on version 5.6.0, I can't replicate in version 5.5.15

This change fix the error.

### DIFF
--- a/lib/helpers/schematype/handleImmutable.js
+++ b/lib/helpers/schematype/handleImmutable.js
@@ -17,7 +17,7 @@ module.exports = function(schematype) {
 
 function createImmutableSetter(path) {
   return function immutableSetter(v) {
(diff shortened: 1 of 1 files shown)


vkarpov15:
Thanks :+1: We introduced immutable in 5.6.0, so this bug wouldn't affect 5.5.x

felipeplets:
@vkarpov15 do you know when we can expect it to be live?
This issue prevents me to update my project to Mongoose 5.6.

vkarpov15:
Early next week.

PR #8130, opened 2019-09-04
fix(parallel): fix parallelLimit when fns is empty

### LINKED ISSUE #8128: Cannot read property 'filter' of undefined
bug

it is breaking with the following error:

```jsx
TypeError: Cannot read property 'filter' of undefined

      at node_modules/mongoose/lib/model.js:3251:32
      at parallelLimit (node_modules/mongoose/lib/helpers/parallelLimit.js:15:12)
      at Function.Object.<anonymous>.Model.$__insertMany (node_modules/mongoose/lib/model.js:3245:3)
      at process.nextTick (node_modules/kareem/index.js:369:33)
```

Everything worked well on `5.6.11` but it broke on `5.6.12`

it should not break

mongoose: 5.6.12
node: 10.15.0
mongo: v4.0.5

sibelius:
fix #8128

**Summary**

new parallelLimit has a bug in an edge case

**Examples**

check #8128

### DIFF
--- a/lib/helpers/parallelLimit.js
+++ b/lib/helpers/parallelLimit.js
@@ -12,7 +12,7 @@ function parallelLimit(fns, limit, callback) {
   let error = null;
 
   if (fns.length === 0) {
-    return callback();
+    return callback(null, []);
   }
 
   for (let i = 0; i < fns.length && i <= limit; ++i) {


sibelius:
Where can we add a regression test?

vkarpov15:
@sibelius I added one :point_up:

PR #8138, opened 2019-09-07
Fix significant performance impact of index creation link in guide

joebowbeer:
**Summary**

Update link; avoid extra-long source line.


vkarpov15:
Thanks!

PR #8228, opened 2019-10-08
feat(schema): make pojo paths optionally become subdoc instead of Mixed (closes #7494)

### LINKED ISSUE #7181: Make `foo: { type: { name: String } }` behave the same as `{ foo: new Schema({ name: String }) }`
Re: #7143. Right now the below declarations are different, and its a very subtle footgun that's hard to notice. 

```javascript
// `foo.name` is a string
new Schema({ foo: { name: String } });

// `foo` is mixed
new Schema({ foo: { type: { name: String } } });
```

Demo using Node shell:

```
$ node
> const { Schema } = require('mongoose')
undefined
> new Schema({ foo: { type: { name: String } } }).path('foo')
Mixed {
  path: 'foo',
  instance: 'Mixed',
  validators: [],
  setters: [],
  getters: [],
  options: { type: { name: [Function: String] } },
  _index: null,
  [Symbol(mongoose:schema_mixed)]: true }
> new Schema({ foo: { type: { name: String } } }).path('foo.name')
undefined
> 
> 
> new Schema({ foo: { name: String } }).path('foo')
undefined
> new Schema({ foo: { name: String } }).path('foo.name')
SchemaString {
  enumValues: [],
  regExp: null,
  path: 'foo.name',
  instance: 'String',
  validators: [],
  setters: [],
  getters: [],
  options: { type: [Function: String] },
  _index: null }
> 
```

Anybody have thoughts on this? I'm thinking it makes sense to deprecate this behavior, and make it so that `new Schema({ foo: { type: { name: String } } })` makes `foo.name` a string in Mongoose 6.x.

@lineus  @fonger ?

### LINKED ISSUE #7408: Schema.path('path to object') returns undefined

bug

`MySchema.path('path to object')` returns `undefined`.

Having:
```ts
  test1: {
    a: String
  },
  test2: {
    type: {
      a: String
    }
  },
  test3: [{
    a: String
  }]
});
```
Will make
`MySchema.path('test1')` == `undefined`.

While the other ones work:
`MySchema.path('test2')` != `undefined`.
`MySchema.path('test3')` != `undefined`.

I would expect `test1` to work the same way as `test2`.

Node v11.6.0
Mongoose: 5.4.3
MongoDB: v4.0.4

### LINKED ISSUE #7494: feature request: nested schemas - fully declarative pojo schema definitions

*feature*

This is very related to #7181 but after starting to write there I decided it's probably off-topic.

TL;DR: 
- hail mary request: nested objects in schema configuration objects make child schemas by default - just like arrays already do.
- next best thing request: an boolean OPTION to do it without needing to use { type: new Schema(...) }

- If you want validation on multiple levels (and don't want Mixed), you need a child schema.
- When a schema configuration has an array, a child schema is created automatically, and a simple pojo can fully describe your schema (including subschemas) \o/.

But when it's not an array, it's impossible to do so without:

child: { type: new Schema({...}) }

That means there are certain schemas that can't be fully described with passing a single configuration pojo to Schema() (namely when you need single subdocuments).

"What's the big deal?  Just wrap the child in 'new Schema()' and move on", you say?

I share my schema definitions between /packages/client and /packages/server, and I currently can't have just a simple pojo in /packages/shared.

Instead of this:

``` javascript
export const schema = {
  sub: {
    type: {
      from: Number,
      to: Number
    },
    validate: ...
  }
}

```

I have to turn it into a factory of sorts so I can use the right "mongoose" - something like this:

``` javascript
export const schemaFactory = (mongoose) => ({
  sub: {
    type: new mongoose.Schema({
      from: Number,
      to: Number
    }),
    validate: ...
  }
});

```

Making the two equivalent would be great, but I don't think everyone would want it.  So if instead, we can have a subSchema: true option or something like that it would be awesome possum.

captaincaius:
Note: Might also close #7181 and close #7408 too...

- default behaves as previous releases did
- embeddeddocument test copied to cover declarative syntax

- otherwise, type: somePojo turns somePojo into subdocument and Mixed paths are only created when explicitly requested via Mixed or {}

I think it's probably also a good idea to make this option a base option too, but wanted to ask for blessing first.

**Summary**

It creates an option that makes type keys valued as POJOs create a subschema instead of Mixed.  The existing behavior is unaltered by default.

This also helps full-stack monorepo use-cases that share the schema configuration object between the client-side project and the server-side project, because the schema configuration object can specify subschemas without importing Mongoose.Schema (so client-side can use mongoose/browser and server-side can use mongoose itself).

**Examples**

See test ;).

### DIFF
--- a/test/types.embeddeddocumentdeclarative.test.js
+++ b/test/types.embeddeddocumentdeclarative.test.js
@@ -53,7 +53,7 @@ describe('types.embeddeddocumentdeclarative', function() {
       const ParentSchema = new mongoose.Schema(ParentSchemaDef, {typePojoToMixed: false});
       it('interprets the POJO as a subschema (gh-7494)', function(done) {
         assert.equal(ParentSchema.paths.child.instance, 'Embedded');
-        assert.equal(ParentSchema.paths.child['$isSingleNested'], true);
+        assert.strictEqual(ParentSchema.paths.child['$isSingleNested'], true);
         done();
       });
       it('enforces provided schema on the child path, unlike Mixed (gh-7494)', function(done) {
@@ -68,7 +68,7 @@ describe('types.embeddeddocumentdeclarative', function() {
         const princessZelda = kingDaphnes.child.toObject();
 
         assert.equal(princessZelda.name, 'Princess Zelda');
-        assert.equal(princessZelda.mixedUp, undefined);
+        assert.strictEqual(princessZelda.mixedUp, undefined);
         done();
       });
     });
@@ -112,16 +112,16 @@ describe('types.embeddeddocumentdeclarative', function() {
         child: {
           name: 'Prince Komali',
           type: 'Medli',
-          confidence: 0,
+          confidence: 1,
         }
       });
 
       assert.equal(grandmother.child.name, 'Rito Chieftan');
       assert.equal(grandmother.child.type, 'Mother');
-      assert.equal(grandmother.child.confidence, undefined);
+      assert.strictEqual(grandmother.child.confidence, undefined);
       assert.equal(ritoChieftan.child.name, 'Prince Komali');
       assert.equal(ritoChieftan.child.type, 'Medli');
-      assert.equal(ritoChieftan.child.confidence, undefined);
+      assert.strictEqual(ritoChieftan.child.confidence, undefined);
       done();
     });
   });
--- a/test/types.embeddeddocumentdeclarative.test.js
+++ b/test/types.embeddeddocumentdeclarative.test.js
@@ -1,0 +1,58 @@
+'use strict';
+
+/**
+ * Module dependencies.
+ */
+
+const assert = require('assert');
+const start = require('./common');
+
+const mongoose = start.mongoose;
+const Schema = mongoose.Schema;
+
+/**
+ * Test.
+ */
+
+describe('types.embeddeddocumentdeclarative', function() {
+  let GrandChildSchema;
+  let ChildSchema;
+  let ParentSchema;
+
+  before(function() {
+    GrandChildSchema = {
+      name: String
+    };
+
+    ChildSchema = {
+      name: String,
+      children: [GrandChildSchema]
+    };
+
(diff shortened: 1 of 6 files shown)


captaincaius:
Please let me know if:

1. This is the right approach (so I can update docs too).
2. You're okay with adding a base option for this too.

captaincaius:
Is it just me or are tests flaky now? :(

test/types.embeddeddocumentdeclarative.test.js
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    ChildSchema = {
+      name: String,
+      children: [GrandChildSchema]
+    };
+
+    ParentSchema = new Schema({
+      name: String,
+      typePojoToMixed: false,
```
vkarpov15: I really don't like this syntax - mixing options with paths is very messy.

captaincaius: Good catch - yeah this was a total typo...  it was supposed to be in the options object.

the fact that this test passes tells me my lazy copy-the-other-test approach isn't sufficient for this.  I'll write tests from scratch and ensure there's no regression when the option isn't set (a Mixed path is made) and that it truly creates subdocuments when the option IS set.

captaincaius: Fixed.

The fact that this test passed scared the life out of me so I rewrote the test entirely - from a coverage POV it's much better.  But from a test independence POV there may be a little redundancy.  I think that if typePojoToMixed ever becomes true by default, a little test refactor might be in order.

captaincaius:
test fixed...  still pending:
- updated docs
- global default for the option (so it doesn't need to be set on every Schema)

captaincaius:
mongoosewide option added - only docs left...  I think I'll plan to update the following:

- subdocument guide: mention the ability to do it without new Schema() with the new option
- SchemaTypes guide: Mixed section

...and looks like that's it AFAICT

captaincaius:
all done - ready for review.

PR #8289, opened 2019-10-28
WIP: fix(document): allow validation of moddelless documents on node side

### LINKED ISSUE #8237: fature/bug: databaseless facade on server-side mongoose

*feature*

but in reality it's a little bit from column A, a little bit from column B

```
captainlocal@cappielappiebuntu:~/dev/mongooseygoose/mongoose$ node
> const goosey = require('./lib');
undefined
> const Document = goosey.Document
undefined
> const intentUserDoc = new Document({}, { name: String });
TypeError: Cannot read property 'options' of undefined
    at new Document (/home/captainlocal/dev/mongooseygoose/mongoose/lib/document.js:83:34)
> 

```

server-side behaves the same way as client-side (i.e. doesn't crash)

tl;dr from below - requresting something like this that works on both normal mongoose and mongoose/browser:

```
const User = mongoose.ModelWithoutDB(UserSchema);
const user = new User();

```

I know it sounds silly to want this on the server-side, but SSR is the motivation.  While working on the PR for #7219 I realized that when replacing references to my own 'wb-mongoose-browser' module with 'mongoose' itself, all the sudden I got that error above.

This is because on the SSR build, even though the codebase is using browser syntax, webpack grabs the normal mongoose instead of the 'browser' key in package.json (it targets 'node' after all).  As a workaround, you can use resolve.alias in webpack, but it would be much nicer if it wasn't necessary.

node 10, mongoose master

### LINKED ISSUE #8272: bug: validateSync() fails on modelless documents

*bug*

relates to #8237 - now you can create a document, but you can't validate it.

```
> const goose = require('./index.js');
undefined
> const schemo = new goose.Schema({num: {type: Number, default: 4}});
undefined
> const dokk = new goose.Document({}, schemo);
undefined
> dokk
{ num: 4 }
> dokk.validateSync();
TypeError: _this.constructor.emit is not a function
    at Document.validateSync (/home/captainlocal/dev/mongooseygoose/mongoose/lib/document.js:2347:21)
> 
```

validateSync() works as expected

node 10, mongoose master

captaincaius:
Hi - this PR is incomplete because it lacks tests, docs, etc. because I wanted to see if you're okay with this approach first.  To me it improves DX and is also a little more maintainable b/c it makes the code paths the same on the client-side and server-side with respect to non-persistent documents.

If you're okay with this direction, please let me know if you have a better idea for a name than "ephemodel" (ephemeral model) :-/. 

Fix #8237, fix #8272, and improve DX by making ephemeral models similar
to normal ones

**Summary**

This fixes modelless document validation on the server-side by sharing its exact same code path with that of browser-side documents.

Browser-side syntax is currently broken on the server-side, so making models for DTO-type sanitizers can currently only be done by making a REAL model.

**Examples**

```javascript
const Ephey = goosey.ephemodel(new goosey.Schema({hair: Number}));
const dockey = new Ephey({hair: 'red', eyes: 'green'});
console.log(dockey.validateSync());
console.log(dockey.toObject());
```
...prints...

> { ValidationError: Validation failed: hair: Cast to Number failed for value "red" at path "hair"
...and...
> {}

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -23,6 +23,7 @@ const STATES = require('./connectionstate');
 const Types = require('./types');
 const Query = require('./query');
 const Model = require('./model');
+const BrowserDocument = require('./browserDocument');
 const Document = require('./document');
 const applyPlugins = require('./helpers/schema/applyPlugins');
 const get = require('./helpers/get');
@@ -395,6 +396,25 @@ Mongoose.prototype.pluralize = function(fn) {
   }
   return _mongoose._pluralize;
 };
+
+
+Mongoose.prototype.ephemodel = function(schema, skipInit) {
+  let ephemodel = function Ephemodel(doc, fields, skipId) {
+    if (doc instanceof Schema) {
+      throw new TypeError('Argument to `Ephemodel` must be a POJO, ' +
+        '**not** a schema. Make sure you\'re calling `mongoose.ephemodel()`, not ' +
+        '`mongoose.Ephemodel()`.');
+    }
+    BrowserDocument.call(this, doc, this.schema, fields, skipId, skipInit);
+  };
+  ephemodel.__proto__ = BrowserDocument;
+
+  ephemodel.prototype = Object.create(BrowserDocument.prototype);
+  ephemodel.prototype.constructor = ephemodel;
+
+  ephemodel.prototype.schema = schema;
+  return ephemodel;
+}
 
 /**
  * Defines a model or retrieves it.


vkarpov15:
I'm not sure I can support this approach because I really don't want to import browser document into node code. If the only issue is what you saw in #8272, that's a pretty easy fix.

captaincaius:
Thanks for the quick review.

Okay...  I had a hunch you'd say that, so I'm glad I didn't get too elaborate with it.  I'll close this.

FWIW, I was going to rename browserDocument because a databaseless document is useful beyond the browser, but never mind.

Thanks again.

vkarpov15:
Yeah 'databaselessDocument' is overkill since the `Document` class is databaseless anyway :) The reason why mongoose's `Model` class inherits from the `Document` class is that `Document` is intended to hold all the db-independent stuff: validation, etc. `Model` has everything that actually requires a database connection.

PR #8307, opened 2019-11-06
feat(clone) Check for & use `cloneOf` method when cloning objects

### LINKED ISSUE #8299: Add `cloneOf` or `clone` method check in clone function to handle odd cloning situations
Feature
`clone` function falls back to `valueOf`, leading to a situation where `valueOf` may not give you the cloned value you seek.

See https://github.com/mongoosejs/mongoose-double/issues/7 for background on the issue that lead to this ticket.

`valueOf` should return a clone but this is not always possible.  Thus we need a method that we can seek and call to generate a proper clone

Suggest adding a `cloneOf` or `clone` method (similar to what is done for SchemaTypes) so we can support this would be great.

Per private email conversation with @vkarpov15 

> At first glance I think it is better to make Mongoose clone handle doubles without relying on valueOf as a clone. Please open up an issue in the mongoose GitHub repo.

stieg:
Adds a check in the `clone` function to see if the given parameter has
a `cloneOf` method.  If the method exists then the `clone` method
assumes that this method will return a clone of the called object and
thus invokes it to complete its task.

Fixes #8299

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -199,6 +199,10 @@ exports.clone = function clone(obj, options, isArrayChild) {
     return obj.toObject(options);
   }
 
+  if (typeof obj.cloneOf === 'function') {
+    return obj.cloneOf();
+  }
+
   if (obj.constructor) {
     switch (exports.getFunctionName(obj.constructor)) {
       case 'Object':
--- a/test/utils.test.js
+++ b/test/utils.test.js
@@ -254,6 +254,22 @@ describe('utils', function() {
 
       return Promise.resolve();
     });
+
+    it('invokes cloneOf method on object before valueOf if exists', function(done) {
+      const o = Object.create({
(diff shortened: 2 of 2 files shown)


stieg:
@vkarpov15 The failure of the CI appears to be unrelated to this change.  Looking for feedback on this when you have a moment.  This PR currently blocks any resolution to https://github.com/mongoosejs/mongoose-double/issues/7 so as soon as a path forward is cleared here I can move forward with PRs for that project.  Cheers.

stieg:
I agree with your assessment above and the potential proposed solution.  Question I have is if I put something like that together are you ready to review/accept it or do you simply need more time to review #8299 before even considering it.  I would like to drive this issue to a conclusion soon as is reasonably possible but first I need your blessing on the solution path before I can do so.  LMK.

PR #8326, opened 2019-11-12
fix(model): delete $versionError after saving to prevent memory leak

Fonger:
The original fix in PR #8048 has a typo. This PR really fixes the issue.


vkarpov15:
I'll merge this for now, but can you also write a test for this please? Basically make sure that `$versionError` exists before and no longer exists after.

PR #8335, opened 2019-11-13
Fixes #6899

### LINKED ISSUE #6899: Throw an error on setting an invalid global option
Using the following doesn't let the user know that they entered a wrong option, which could be a typo.

```js
mongoose.set('invalidOption', true);
```

Planning to make a PR on the weekend that throws an error to inform the user on setting an invalid option.

AbdelrahmanHafez:


AbdelrahmanHafez:
Done.

First time actually benefiting from CI build, used to wait for tests to run for 20 good minutes to see the changes before committing every time. So useful!

P.S: The Node 5 build #7684 seems to be having a problem with installing node, do you think you can trigger a restart on that build?

Also, I've been curious for a while now, are we actually supporting Node 4 on mongoose 5? Would it be too crazy if we used post node 4 features such as `Array.prototype.includes(...)`?

AbdelrahmanHafez:
I am starting to think that maybe we should merge this on the next major release, since it may be a breaking change for some people who have invalid options in their codebase. Thoughts?

vkarpov15:
We can push it back to a minor release like 5.8. You're right that this change might be a bit too heavy for a patch release.

PR #8351, opened 2019-11-17
Fixes #8317 map(...)

### LINKED ISSUE #8317: TypeError: Cannot read property '$__' of undefined at _updateParentPopulated
Hello, this seems similar to #3889.

I am using node 8.10.0, mongodb 3.2.11 and mongoose 5.7.9

I have the following code:

```
var forms = this.getFormalization();
var form = forms.find(f => f._id == id);
  if (typeof form.lastIndependentDate === 'undefined' || this.lastUpdate > 
      form.lastIndependentDate) {
    forms = forms.slice()
    forms.splice(forms.indexOf(form), 1)
```

The last line throws the error:

TypeError: Cannot read property '$__' of undefined
    at _updateParentPopulated (/home/shaolin/Documents/software/normative-gui/node_modules/mongoose/lib/types/documentarray.js:323:14)
    at CoreDocumentArray.splice (/home/shaolin/Documents/software/normative-gui/node_modules/mongoose/lib/types/documentarray.js:228:5)
    at model.theorySchema.methods.isIndependent (/home/shaolin/Documents/software/normative-gui/models/theory.js:145:11)
    at /home/shaolin/Documents/software/normative-gui/controllers/theoryController.js:202:14
    at /home/shaolin/Documents/software/normative-gui/node_modules/mongoose/lib/model.js:4604:16
    at /home/shaolin/Documents/software/normative-gui/node_modules/mongoose/lib/query.js:4351:12
    at process.nextTick (/home/shaolin/Documents/software/normative-gui/node_modules/mongoose/lib/query.js:2849:28)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)

The model is
```
var theorySchema = new Schema({
    lastUpdate        : Date,
    name 		          : String,
    description       : String,
    content           : String, // annotated XML
    comment           : String,
    vocabulary        : [{symbol: String, original: String, full: String}],
    autoVocabulary    : [{symbol: String, original: String, full: String}],
    formalization     : [{original: String, json: Object, formula: String, active: Boolean, lastIndependent: Boolean, lastIndependentDate: Date}],
    autoFormalization : [{original: String, json: Object, formula: String, active: Boolean, lastIndependent: Boolean, lastIndependentDate: Date}],
		user 							: { type: Schema.Types.ObjectId, ref: 'User' },
    clonedForm        : { type: Schema.Types.ObjectId, ref: 'Theory' },
    lastConsistency   : Boolean,
    lastConsistencyDate: Date,
    writeProtected    : Boolean
});


theorySchema.methods.getFormalization = function() {
  return this.formalization.concat(this.autoFormalization)
}
```
Downgrading to 4.4.3 also fixes it in my case.

Thank you.

AbdelrahmanHafez:
Will be testing other array methods soon, and fix any broken ones.

@vkarpov15 I am curious, what are the exact changes that introduced this bug?

### DIFF
--- a/lib/types/documentarray.js
+++ b/lib/types/documentarray.js
@@ -180,6 +180,15 @@ class CoreDocumentArray extends CoreMongooseArray {
     return arr;
   }
 
+  map() {
+    const arr = super.map.apply(this,arguments);
+
+    arr[arrayParentSymbol] = this[arrayParentSymbol];
+    arr[arrayPathSymbol] = this[arrayPathSymbol];
+
+    return arr;
+  }
+
   /**
    * Wraps [`Array#push`](https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Array/push) with proper change tracking.
    *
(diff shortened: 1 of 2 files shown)


Fonger:
I think it's #8037

AbdelrahmanHafez:
To understand correctly, what you're saying is `map()` and `push()` always had this bug? Because if that's the case, I've been using the combination and the test only started to fail lately.

You're right though, it makes more sense to return a good old JS array on mapping. Changes coming your way.

AbdelrahmanHafez:
Got some failing tests on Node 4 and 5, will check them soon.

lib/types/documentarray.js
```diff
@@ -180,6 +180,11 @@ class CoreDocumentArray extends CoreMongooseArray {
     return arr;
   }
 
+  map() {
+    const arr = [].concat(Array.prototype.map.apply(this,arguments));
```
vkarpov15: I think this change is currently unnecessary. I like the idea of `map()`, `filter()`, etc. returning vanilla JavaScript arrays rather than mongoose arrays, but I think that's a bit too heavy of a change for a patch release. I think that all we need for this release is the `if (!parent || parent.$__.populated == null) return;` change.

PR #8428, opened 2019-12-14
Fixes #8371

### LINKED ISSUE #8371: first  remove, then save, why there is no data？
please ignore the contradiction of code logic. after the following code is executed, there is no data in mongodb. who can tell me why? according to my understanding, execute  ```document.save()```can insert data, but now there is no data. it's strange.

```

let acl = await Acl.findOneAndRemove({name: "nodejs"});
console.log(acl);
console.log(acl instanceof mongoose.Document); // true
let newAcl = await acl.save();
console.log(newAcl);

```
nodejs：10.16.3    
mongoose: 5.6.9

AbdelrahmanHafez:
Add test for #8371 verifying that setting `document.isNew = true;` makes `document.save()` tries to insert a new document.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -309,7 +309,7 @@ Model.prototype.$__handleSave = function(options, callback) {
         callback(null, ret);
       });
     } else {
-      this.constructor.exists({_id:this._id})
+      this.constructor.exists(this.$__where())
         .then((documentExists)=>{
           if (!documentExists) throw new DocumentNotFoundError(this.$__where(),this.constructor.modelName);
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -309,8 +309,14 @@ Model.prototype.$__handleSave = function(options, callback) {
         callback(null, ret);
       });
     } else {
-      this.$__reset();
-      callback();
+      this.constructor.exists(this)
+        .then((documentExists)=>{
+          if (!documentExists) throw new DocumentNotFoundError(this.$__where(),this.constructor.modelName);
+
+          this.$__reset();
+          callback();
+        })
+        .catch(callback);
       return;
     }
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -309,7 +309,7 @@ Model.prototype.$__handleSave = function(options, callback) {
         callback(null, ret);
       });
     } else {
-      this.constructor.exists(this)
+      this.constructor.exists({_id:this._id})
         .then((documentExists)=>{
           if (!documentExists) throw new DocumentNotFoundError(this.$__where(),this.constructor.modelName);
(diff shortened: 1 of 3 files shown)


AbdelrahmanHafez:
I keep getting random timeouts, gonna close and re-open the PR to trigger a new build.

lib/model.js
```diff
@@ -309,8 +309,14 @@ Model.prototype.$__handleSave = function(options, callback) {
         callback(null, ret);
       });
     } else {
-      this.$__reset();
-      callback();
+      this.constructor.exists(this.$__where())
```
vkarpov15: Unfortunately this is not going to work - Mongoose shouldn't be executing additional MongoDB queries on `save()`, that can cause perf issues for some apps.

I think setting `isNew = true` on `findOneAndRemove()` would be a better approach.

AbdelrahmanHafez: @vkarpov15 
I see your point, this is an added query and that will have a negative impact on performance.

However, I don't think the overhead is significant, considering that it's an edge case where the user is trying to save a document with no changes, we'll be fetching only the `_id` (considering the current implementation of exists).

I am not entirely comfortable with the idea of setting `isNew = true` on `findOneAndRemove(...)` because it not really a _new_ document, if any other considerations were to be made to new docs, this would be mistaken for an actual new document.

Currently, making changes to a deleted doc (by any function/process), then saving, throws an error. (the third test passes without this PR)

With the `findOneAndRemove(...)` and setting `isNew = true` solution to fix the issue will still create inconsistency in the case of trying to save a document that is deleted from the collection if we don't verify its existence it could be removed by another function/process and the document will not be aware of it. Using `doc.save()` on a deleted document with no changes will silently fail, which I think is an issue that's worth the added existence verification overhead.

One could argue that the user would expect .save(...) to execute a query to the database anyway, in the past we tried to do the smart thing and not send anything when we didn't have to, but as the need surfaced (this bug) we're getting back to the initial status which is sending a read operation.

It seems to me our options are:

A) Merge this PR, and check for existence when no changes are present on the document.
B) Make this behavior optional, with it being turned on by default, and make users who will have a problem with the performance opt-out of it.
C) make findOneAndRemove(...) set isNew = true, but that will make the second test fail, which IMO shouldn't.

Please let me know how you wanna proceed.

vkarpov15: Ah ok I didn't read your PR closely enough. I didn't notice that the `exists()` check only happens when there aren't any changes to save. I'll think about this PR a little more, but right now I'm leaning towards merging this, I think you're right that we should surface a `DocumentNotFoundError` if the document isn't there but there's no changes to save.

PR #8435, opened 2019-12-17
allow browser build to be published (fix #8427)

### LINKED ISSUE #8427: Npm publish ignores browser build

*bug*

Although `npm prePublish` builds it, it doesn't end up in the published package because `dist` is in `.npmignore`.

dist gets published

Mongoose 5.8.0

captaincaius:

**Summary**

Fixes missing dist folder in published package

**Examples**

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -15,7 +15,6 @@ bin/
 tools/31*
 *.key
 data/
-dist/
 
 *.txt
 *.png


captaincaius:
I realized this is against master - should I redo (or cherrypick) starting from 5.8 instead?  The README says all PRs should be to master, but it's looking like you may have evolved your strategy since it was written.

vkarpov15:
No worries about making this PR against `master` and no need to cherrypick, all the commits that are in 5.8 are in master. The README is correct IMO, most PR's should be against master because they usually aren't significant enough to go in a minor or major release.

PR #8498, opened 2020-01-10
[WIP] Add support for default options per schema type  (Fixes #8487)

### LINKED ISSUE #8487: [Feature] default options for specific data type
There are types that I almost always use specific options with.

For example, I always use `trim: true` with fields that have `type: String`, and I require everything by default, the exception being things that are `required: false`, and numbers almost always have `min: 0`.

Is there a way we can add default options per field type, which can be overwritten by specifying some other value?

The way I imagine such an API would be as follows:
```js
// maybe per schema?
mongoose.fieldOptions({ type: String, defaultOptions: { trim: true, required: true  } });
```

Thoughts?

AbdelrahmanHafez:
@vkarpov15 
This is a solution that _works_ on `SchemaString`, but it clutters the `SchemaString` constructor, and I don't like that. I am trying to centralize it on `SchemaType` and make all the other types inherit the behavior from it.

Any pointers would be greatly appreciated.


AbdelrahmanHafez:
TODO:
1- Add support for Map, DocumentArray, EmbeddedDocument, SingleNestedPath (help needed).
2- Add tests in a separate file `setDefaultOption.js` and test all types there.
3- Add to documentation in jade files, and JSDoc.

AbdelrahmanHafez:
*TRIGGERS another build because of a timeout failing test*

AbdelrahmanHafez:
That timeout test seems to be failing randomly, I'll just assume this is not related to this PR.

test/types.string.test.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+describe('types.string', function() {
+  describe('Schema.Types.String.setDefaultOptions(...)', function() {
+    it('when given an option, sets it', () => {
+      // Arrange
+      const mongooseInstance = new mongoose.Mongoose();
+
+      // Act
+      mongooseInstance.Schema.Types.String.setDefaultOption('trim',true);
```
vkarpov15: I'd recommend just `set()` rather than `setDefaultOption()`. That's more consistent with [`mongoose.set()`](https://mongoosejs.com/docs/api/mongoose.html#mongoose_Mongoose-set)

lib/helpers/schematype/getDefaultOptionSetter.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+function getDefaultOptionSetter(type) {
```
vkarpov15: Why is this function necessary? Seems like a lot of work to avoid copying a one-liner.

Especially since you could just define `set()` as a function on SchemaType.prototype.

AbdelrahmanHafez: I tried to do that, had a hard time inheriting a static.

Apparently, I just needed to
```js
SchemaType.set = function set (option, value){
  this.defaultOptions[option] = value;
};
```

and on the SchemaTypes
```js
SchemaString.set = SchemaType.set;
```

test/types.string.test.js
AbdelrahmanHafez: I agree, `set()` sounds better, consistency-wise.

AbdelrahmanHafez:
@vkarpov15 I am not sure whether it makes sense to add `set(...)` to `DocumentArray`, `SingledNestedPath`, and `Array`. I haven't used them explicitly a lot, and not sure if/how we would use `set(...)` with them.

AbdelrahmanHafez:
Currently, all tests are passing except for the flaky timeout test.

lib/schematype.js
```diff
@@ -139,6 +150,10 @@ SchemaType.cast = function cast(caster) {
   return this._cast;
 };
 
+SchemaType.set = function set(option, value) {
+  this.defaultOptions[option] = value;
```
vkarpov15: Let's add a check here for `this.hasOwnProperty('defaultOptions')` to ensure that `set()` never modifies the base class's default options. Something like:

```
if (!this.hasOwnProperty('defaultOptions')) {
  this.defaultOptions = Object.assign({}, this.defaultOptions);
}
this.defaultOptions[option] = value;
```

I ran into this issue a lot when implementing support for `SchemaNumber.cast()`, etc. See [Static Properties in JavaScript Classes with Inheritance](http://thecodebarbarian.com/static-properties-in-javascript-with-inheritance.html) for some learnings on that. Also see [`hasOwnProperty()`](https://masteringjs.io/tutorials/fundamentals/hasownproperty).

AbdelrahmanHafez:
Sounds great, I'd prefer it if we wait till the implementation is finished before rebasing.

I'll probably be able to handle the rest of the tasks on the to-do-list after a week or so. Meanwhile, I'd like to know what you think about this matter: https://github.com/Automattic/mongoose/pull/8498#issuecomment-575472421

vkarpov15:
For https://github.com/Automattic/mongoose/pull/8498#issuecomment-575472421, I think we should add `.set()` for those. You can still set options like `required` and `validate` for those. But we can postpone that until later if it is too complicated.

AbdelrahmanHafez:
Alright, I think this is good to go. Only remaining is to add `.set(...)` to `EmbeddedDocument` and `SingleNestedPath`. Been having some difficulty with these.

Also, I hope rebasing my branch over master wouldn't make it difficult reviewing this PR.  I would compare [the changes](https://github.com/Automattic/mongoose/compare/master...AbdelrahmanHafez:feat/set-default-option-per-type) from my branch to master.

vkarpov15:
Merged into 5.9 branch :+1:

PR #8532, opened 2020-01-24
test and Fixes #8531 - only call validate once for subdocuments

### LINKED ISSUE #8531: "Can't validate() the same doc multiple times in parallel" erroneously thrown
This is a bug

I haven't tested enough to be certain what all different cases can reproduce this, but the case I tried was starting with a document  with a single nested path inside a document in an array inside a document inside an array, if you modify the contents of that nested path document and then save() the document you will get this error:

Can't validate() the same doc multiple times in parallel.

In our code we are updating the doc like this:

            const q: any = assn.answer_keys[0].questions[0];
            q.clip = {rows: 1, row_height: 3.5};

The error is occuring on the "clip" subdocument.

From digging in it seems to be calling the validator twice on that path, once as "questions.0.clip" and then again as "clip" -- but either the second time occurs as part of the validation for the first (async nesting) or else the first is never calling the callback because when it gets called a second time the validation error has not been cleared yet.

You can run the unit test from https://github.com/taxilian/mongoose/tree/feature/validate_in_parallel_issue

or the gist https://gist.github.com/taxilian/0dc71f915eded06eed74112146de7887

It should save and not throw an exception =]

node.js 10.17.0, mongoose 5.8.9

----

I will try to fix this myself, but if you have any suggestions on how to do it I'd appreciate it. I don't know if you frequent any discords or anything where I could get some suggestions. I've been hunting this through async stack traces for about 8 hours now.

taxilian:
It turns out that due to a check which I can't find a reason for the logic for removing subdocuments from the "paths" list when running validators only ran on the root document -- this is not a problem most of the time, it just results in the validators running multiple times on some embedded paths, but in some slightly weird cases it can result in validate being called twice -- and the second time is deeper in the stack from the first validate.

Combined with the check for multiple parallel calls to `validate()` this resulted in a fatal error when saving a document with changes to those documents which triggered that case.

A test has been added in this pull request which reproduced the error and which after the fix no longer reproduces the error.

This fixes #8531

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2103,7 +2103,8 @@ function _getPathsToValidate(doc) {
   const modifiedPaths = doc.modifiedPaths();
   for (i = 0; i < len; ++i) {
     subdoc = subdocs[i];
-    if (doc.isModified(subdoc.$basePath, modifiedPaths) &&
+    if (subdoc.$basePath &&
+        doc.isModified(subdoc.$basePath, modifiedPaths) &&
         !doc.isDirectModified(subdoc.$basePath) &&
         !doc.$isDefault(subdoc.$basePath)) {
       // Remove child paths for now, because we'll be validating the whole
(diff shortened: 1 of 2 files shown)


jloveridge:
Would love to see this in an official release.

PR #8536, opened 2020-01-24
Copy plugins from base schema when creating a discriminator

samgladstone:
**Summary**

While debugging a problem I was having, I noticed the following issue: the base schema's plugins are not being copied to the discriminator's plugin. I have investigated the change and have not identified anything that would be impacted. At least, as far as I can see, the plugin functions are only called when creating schema or calling `schema.plugin()`. Still, I lost some time trying to work out if this was the cause of my problem and so figured this could be helpful if it saves someone else going down this rabbit hole.

My only concern is that it may be confusing to see a plugin listed that is designed to manipulates the schema when the schema has not been manipulated. I struggled to think of an appropriate example (that wasn't redundant) and the best I could come up with was:

`js
const simplePlugin = (schema) => {
    const valuePath = schema.path('value');

    if (valuePath && !valuePath.isRequired)
        valuePath.required(true);
}
`

I figure it would be confusing to see this plugin listed for a discriminator when value was not on the base schema but is for the discrimintaing one and so is not required. So, alternatively if you prefer, we could flag that the plugin was only applied to the base by doing something along the lines of:

`js
schema.plugins = Array.prototype.slice.call(baseSchema.plugins).forEach(p => { p.baseSchemaOnly = true });
`

### DIFF
--- a/lib/helpers/model/discriminator.js
+++ b/lib/helpers/model/discriminator.js
@@ -158,7 +158,7 @@ module.exports = function discriminator(model, name, schema, tiedValue, applyPlu
     schema.options.id = id;
     schema.s.hooks = model.schema.s.hooks.merge(schema.s.hooks);
 
-    schema.plugins = Array.prototype.slice(baseSchema.plugins);
+    schema.plugins = Array.prototype.slice.call(baseSchema.plugins);
     schema.callQueue = baseSchema.callQueue.concat(schema.callQueue);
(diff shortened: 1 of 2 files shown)


vkarpov15:
Looks like this mistake was introduced in this commit: https://github.com/Automattic/mongoose/commit/014cbbd095393e7b871f276822e3e6912930469f .

PR #8542, opened 2020-01-27
Fix SchemaType.clone()

samgladstone:
I noticed that `SchemaType.clone()` was not copying values where updates had been made after instantiation. 

I figured that testing using `assert.deepStrictEqual` made the most sense here, but perhaps it's overkill and the if (x !== undefined) are unecessary? I'll happily adjust if you have a preference.

Also, I thought the rename schema.type.test.js → test/schematype.test.js made sense as it seems to be testing shared functionality, again will adjust if you prefer the old style.


samgladstone:
Currently linting/looking at the fail

samgladstone:
There we go 😄

Issue #7704, opened 2019-04-11
5.5.x causing CastError(s) in subdocs


paglias:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
We are currently using 5.4.19 because we were affected by this bug https://github.com/Automattic/mongoose/issues/7681 and https://github.com/Automattic/mongoose/issues/7660 so I'm not sure if it got introduced in 5.5.1 or in between the two versions but since upgrading to 5.5.1 we're seeing several test failures regarding casting errors inside embedded docs

```
User validation failed: purchased.plan.consecutive.offset: Cast to Number failed for value "NaN" at path "consecutive.offset", purchased.plan.consecutive.gemCapExtra: Cast to Number failed for value "NaN" at path "consecutive.gemCapExtra", ... (other fields here...)
```
In our schema `Purchased` is a subdoc of `User`

and 

```
CastError: Cast to embedded failed for value "{ completed: false,\n  text: \'Checklist Item 1\',\n  id: \'274db346-3baa-46c6-bff3-ea5cff557301\' }" at path "undefined"
```

In this case it's also happening inside a subdoc (although an array of docs not a single one like `purchased`)

**If the current behavior is a bug, please provide the steps to reproduce.**
I've not been able to create a repro script yet but will look into it

**What is the expected behavior?**
No errors

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Node 10, mongo 3.4, mongoose 5.5.1

paglias:
Confirmed in 5.5.0 as well, not a full repro script but looks like it's being caused by stuff like this

```
          task.checklist.push({
            text: 'Checklist Item 1',
            completed: false,
          });
```

where task.checklist is a doc of this schema https://github.com/HabitRPG/habitica/blob/develop/website/server/models/task.js#L266

The error thrown is

```
CastError: Cast to embedded failed for value "{ completed: false,\n  text: \'Checklist Item 1\',\n  id: \'6194d00c-0635-4ba6-86e1-49cd6a692deb\' }" at path "undefined"
```

and this is the stack trace

```
      at new CastError (node_modules/mongoose/lib/error/cast.js:29:11)
      at DocumentArray.cast (node_modules/mongoose/lib/schema/documentarray.js:409:19)
      at DocumentArray.SchemaType.applySetters (node_modules/mongoose/lib/schematype.js:864:12)
[...]

Issue #7435, opened 2019-01-20
Setting schema options via a plugin fails in 5.4.5


sebastian-nowak:
This code used to work in 5.4.0 but fails in 5.4.5:

```js
#!/usr/bin/env node
(async function () {
  const mongoose = require('mongoose');
  mongoose.set('applyPluginsToDiscriminators', true);
  mongoose.connect('mongodb://localhost:27017/test', {useNewUrlParser: true});

  mongoose.plugin((schema) => {
    schema.options.versionKey = false;
    schema.options.minimize = false;
  });
  const eventSchema = new mongoose.Schema({
    kind: {type: String}
  }, {
    discriminatorKey: 'kind'
  });

  const testEventSchema = new mongoose.Schema({
    inner: {
      type: new mongoose.Schema({
        _id: false,
        bool: {type: Boolean, required: true}
      })
    }
  });

  const schema = new mongoose.Schema({
    events: {
      type: [eventSchema]
    }
  });

  schema.path('events').discriminator('test-event', testEventSchema);
})();
```

```
Error: Can't customize discriminator option minimize (can only modify toJSON, toObject, _id, id)
```

I think it might be caused by the fix to https://github.com/Automattic/mongoose/issues/7370.

PR #8543, opened 2020-01-28
Discriminator: Remove unnecessary merge of schema with itself

samgladstone:
This is a fix to an issue I have been having. 

My understanding of `utils.merge` is that if you merge an object with itself nothing should happen. Therefore, I think the below merge is redundant.

However, more importantly to me, this merge actually has quite of a lot of (I presume) unintended side effects, which I am not clear on the best approach to fix. I will raise this as a separate issue.

### DIFF
--- a/lib/helpers/model/discriminator.js
+++ b/lib/helpers/model/discriminator.js
@@ -72,8 +72,6 @@ module.exports = function discriminator(model, name, schema, tiedValue, applyPlu
     if (baseSchema.paths._id &&
         baseSchema.paths._id.options &&
         !baseSchema.paths._id.options.auto) {
-      const originalSchema = schema;
-      utils.merge(schema, originalSchema);
       delete schema.paths._id;
       delete schema.tree._id;
     }
--- a/lib/schema/SingleNestedPath.js
+++ b/lib/schema/SingleNestedPath.js
@@ -291,7 +291,7 @@ SingleNestedPath.prototype.doValidateSync = function(value, scope, options) {
  */
 
 SingleNestedPath.prototype.discriminator = function(name, schema, value) {
-  discriminator(this.caster, name, schema, value);
+  schema = discriminator(this.caster, name, schema, value);
 
   this.caster.discriminators[name] = _createConstructor(schema, this.caster);
 
--- a/test/docs/discriminators.test.js
+++ b/test/docs/discriminators.test.js
@@ -213,8 +213,8 @@ describe('discriminator docs', function () {
     var event1 = new ClickedLinkEvent({ _id: 'custom id', time: '4pm' });
     // Woops, clickedLinkSchema overwrites the `time` path, but **not**
     // the `_id` path because that was implicitly added.
-    assert.ok(typeof event1._id === 'string');
-    assert.ok(typeof event1.time === 'string');
+    assert.strictEqual(typeof event1._id, 'string');
+    assert.strictEqual(typeof event1.time, 'string');
 
     // acquit:ignore:start
     done();
--- a/test/model.discriminator.test.js
+++ b/test/model.discriminator.test.js
@@ -312,7 +312,7 @@ describe('model', function() {
       });
 
       it('inherits statics', function(done) {
-        assert.strictEqual(Employee.findByGender, EmployeeSchema.statics.findByGender);
+        assert.strictEqual(Employee.findByGender, PersonSchema.statics.findByGender);
         assert.strictEqual(Employee.findByDepartment, EmployeeSchema.statics.findByDepartment);
         assert.equal(Person.findByDepartment, undefined);
         done();
@@ -1329,7 +1329,7 @@ describe('model', function() {
       // Delete every model
       afterEach(function() { mongoose.deleteModel(/.+/); });
 
-      it('does not modify _id path of the passed in schema the _id is not auto generated', function() {
+      it('does not modify _id path of the passed in schema the _id is not auto generated (gh-8543)', function() {
         const model = mongoose.model('Model', new mongoose.Schema({ _id: Number }));
         const passedInSchema = new mongoose.Schema({});
         model.discriminator('Discrimintaor', passedInSchema);
@@ -1338,7 +1338,7 @@ describe('model', function() {
 
       function throwErrorOnClone() { throw new Error('clone() was called on the unrelated schema'); };
 
-      it('when the base schema has an _id that is not auto generated', function() {
+      it('when the base schema has an _id that is not auto generated (gh-8543) (gh-8546)', function() {
         const unrelatedSchema = new mongoose.Schema({});
         unrelatedSchema.clone = throwErrorOnClone;
         mongoose.model('UnrelatedModel', unrelatedSchema);
--- a/test/model.discriminator.test.js
+++ b/test/model.discriminator.test.js
@@ -1329,6 +1329,13 @@ describe('model', function() {
       // Delete every model
       afterEach(function() { mongoose.deleteModel(/.+/); });
 
+      it('does not modify _id path of the passed in schema the _id is not auto generated', function() {
+        const model = mongoose.model('Model', new mongoose.Schema({ _id: Number }));
+        const passedInSchema = new mongoose.Schema({});
+        model.discriminator('Discrimintaor', passedInSchema);
+        assert.equal(passedInSchema.path('_id').instance, 'ObjectID');
+      });
+
       function throwErrorOnClone() { throw new Error('clone() was called on the unrelated schema'); };
 
       it('when the base schema has an _id that is not auto generated', function() {
--- a/test/model.discriminator.test.js
+++ b/test/model.discriminator.test.js
@@ -1324,6 +1324,22 @@ describe('model', function() {
 
       mongoose.deleteModel(/Model/);
     });
+
+    describe('does not have unintended side effects', function() {
+      // Delete every model
+      afterEach(function() { mongoose.deleteModel(/.+/); });
+
+      function throwErrorOnClone() { throw new Error('clone() was called on the unrelated schema'); };
+
+      it('when the base schema has an _id that is not auto generated', function() {
+        const unrelatedSchema = new mongoose.Schema({});
+        unrelatedSchema.clone = throwErrorOnClone;
+        mongoose.model('UnrelatedModel', unrelatedSchema);
+
+        const model = mongoose.model('Model', new mongoose.Schema({ _id: mongoose.Types.ObjectId }, { _id: false }));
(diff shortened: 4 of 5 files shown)


samgladstone:
This was introduced in 9fa31d3369a68e141f738ec73b3226ac5768dbd2

samgladstone:
I see, good spot! It does modify the original schema. I've added a test for that. 

However, using `schema = schema.clone()` is causing my code to throw the following:

`Project validation failed: fields.1._id: Cast to ObjectID failed for value "2" at path "_id", fields.0._id: Cast to ObjectID failed for value "1" `
`at path "_id"`

It's also, breaking these tests:
[document / push() onto discriminator doc array (gh-7704):](https://github.com/Automattic/mongoose/blob/master/test/document.test.js#L7624) 
[discriminator docs / Handling custom _id fields](https://github.com/Automattic/mongoose/blob/master/test/docs/discriminators.test.js#L194)

I can take a look at what's happening tomorrow.

> Also, what error do you get when calling clone()?

I'm not entirely sure what you mean with this question (unless it's what I've described above)? 

Otherwise, I've described the wider issue in a lot more detail in #8546

samgladstone:
Ok, I've had a look through whether using `schema = schema.clone()` is a feasible alternative. I don't believe it is, particularly because it breaks recursive embedded discriminators.

The tests that were failing above were because the schema is not passed out of that [merge function ](https://github.com/Automattic/mongoose/blob/master/lib/helpers/model/discriminator.js#L69) doesn't return the schema, so cloning it within this function results in the function doing nothing. I adjusted the function to account for this and that passed all the tests:

```js
function merge(schema, baseSchema) {
    // Retain original schema before merging base schema
    schema._baseSchema = baseSchema;
    if (baseSchema.paths._id &&
        baseSchema.paths._id.options &&
        !baseSchema.paths._id.options.auto) {
      schema = schema.clone() // ADDED
      delete schema.paths._id;
      delete schema.tree._id;
    }
    ...
    delete schema._requiredpaths; // reset just in case Schema#requiredPaths() was called on either schema
    return schema; // ADDED 
  }

  // merges base schema into new discriminator schema and sets new type field.
  //merge(schema, model.schema);
  schema = merge(schema, model.schema); // ADDED
```
However, I figured that if it was safe to do this as above, it should be safe to do this when creating every discriminator. This would also surface this change in a lot more tests to throw up potential issues. So, with that in mind, I moved the clone to the beginning of the discriminator function:

```js
module.exports = function discriminator(model, name, schema, tiedValue, applyPlugins) {
  if (!(schema && schema.instanceOfSchema)) {
    throw new Error('You must pass a valid discriminator Schema');
  }

  schema = schema.clone(); // ADDED

  if (model.schema.discriminatorMapping &&
      !model.schema.discriminatorMapping.isRoot) {
    throw new Error('Discriminator "' + name +
        '" can only be a discriminator of the root model');
  }
```

Doing this has exposed a further 22 test failures. Granted, they are edge cases as you have to have the auto id off on the base schema, but it still seems like they are possible. 

### Identified Fixable Issues 

A couple of test failures were inconsistent and seemed like they should be fixed. I've pushed these fixes below **TODO add commit ref**

#### SingleNestedPath Should Reassign Schema

When creating a discriminator from a [model](https://github.com/Automattic/mongoose/blob/master/lib/model.js#L1105), the returned schema from the `discriminator` function is assigned to the input schema. This does not happen for [SingleNestedPath](https://github.com/Automattic/mongoose/blob/master/lib/schema/SingleNestedPath.js#L294). These should be consistent, so I have assigned the return value in `SingleNestedPath`.

```js
SingleNestedPath.prototype.discriminator = function(name, schema, value) {
  //discriminator(this.caster, name, schema, value);
  schema = discriminator(this.caster, name, schema, value);

  this.caster.discriminators[name] = _createConstructor(schema, this.caster);

  return this.caster.discriminators[name];
};
```

This fixes the following tests:
* [document.test.js L6008](https://github.com/Automattic/mongoose/blob/master/test/document.test.js#L6008)
* [document.test.js L6221](https://github.com/Automattic/mongoose/blob/master/test/document.test.js#L6221)
* [model.discriminator.test.js L1042](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.test.js#L1042)
* [model.discriminator.test.js L1084](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.test.js#L1084)
* [model.discriminator.test.js L1252](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.test.js#L1252)
* [query.test.js L2523](https://github.com/Automattic/mongoose/blob/master/test/query.test.js#L2523)
* [query.test.js L2572](https://github.com/Automattic/mongoose/blob/master/test/query.test.js#L2572)

#### Test Against Base Statics

In the test [model /  discriminator() / root schema inheritance / inherits statics](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.test.js#L314), the first assert should be checking against the base schema (`PersonSchema`), not the modified discrimnating schema (`EmployeeSchema`) as that is where it is inheriting the method from

**TODO PUSH**

### NOT FIXABLE: Nested Discriminators Break

Nested Discriminators break as they rely on the schema being modified to create the nesting as in the docs [Discriminators: Embedded discriminators in arrays](https://mongoosejs.com/docs/discriminators.html#embedded-discriminators-in-arrays). I have added a test cases for this in the commit **ADD COMMIT**

### POSSIBLY BREAKING: Schema Modification is Used in Tests

Modifications to the schema are used in the below tests. Making the passed in schema not modified breaks mongoose's tests => it's a breaking change?

All tests are model / discriminator() /
* [find / hydrates correct models](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L111) relies on schema modification on [line 128](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L128)
* schema modiciation on [line 159](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#159) is relied upon by:
  * [find / hydrates correct models when fields selection set as string](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L173)
  * [find / hydrates correct models when fields selection set as object](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L177)
* [findOne/ hydrates correct model](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L367) relies on schema modification on [line 338](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L338)
* schema modiciation on [line 427](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#427) is relied upon by:
  * [findOne/ hydrates correct model when fields selection set as string inclusive](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L448)
  * [findOne/ hydrates correct model when fields selection set as string exclusive](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L452)
  * [findOne/ hydrates correct model when fields selection set as empty string](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L456)
  * [findOne/ hydrates correct model when fields selection set as object inclusive](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L460)
  * [findOne/ hydrates correct model when fields selection set as object exclusive](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L464)
  * [findOne/ hydrates correct model when fields selection set as empty object](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L468)


However, this can be fixed in mongoose by using the result of calling discriminator's schema. For example, in [model / discriminator() / find / hydrates correct models](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.querying.test.js#L111), the following changes resolve the test failure (Changes marked // CHANGED):

```js
it('hydrates correct models', function(done) {
    const baseEvent = new BaseEvent({name: 'Base event'});
    const impressionEvent = new ImpressionEvent({name: 'Impression event'});
    const conversionEvent = new ConversionEvent({name: 'Conversion event', revenue: 1.337});

    baseEvent.save(function(err) {
        assert.ifError(err);
        impressionEvent.save(function(err) {
            assert.ifError(err);
            conversionEvent.save(function(err) {
                assert.ifError(err);
                BaseEvent.find({}).sort('name').exec(function(err, docs) {
                assert.ifError(err);
                assert.ok(docs[0] instanceof BaseEvent);
                assert.equal(docs[0].name, 'Base event');

                assert.ok(docs[1] instanceof ConversionEvent);
                //assert.deepEqual(docs[1].schema.tree, ConversionEventSchema.tree); // CHANGED
                assert.deepEqual(docs[1].schema.tree, ConversionEvent.schema.tree);
                assert.equal(docs[1].name, 'Conversion event');
                assert.equal(docs[1].revenue, 1.337);

                assert.ok(docs[2] instanceof ImpressionEvent);
                //assert.deepEqual(docs[2].schema.tree, ImpressionEventSchema.tree); // CHANGED
                assert.deepEqual(docs[2].schema.tree, ImpressionEvent.schema.tree);
                assert.equal(docs[2].name, 'Impression event');
                done();
                });
            });
        });
    });
});
````

### POSSIBLY BREAKING: Schema is Used for Comparison in Tests

The following tests use the Schema instance that was used to create a discriminator as a way of checking something. Again, this breaks mongoose's tests => it's a breaking change?

* [mongoose module / global plugins with applyPluginsToDiscriminators (gh-7435)](https://github.com/Automattic/mongoose/blob/master/test/index.test.js#L375) uses the original schema at [line 407](https://github.com/Automattic/mongoose/blob/master/test/index.test.js#L407)
* [model / hydrate / works correctly with model discriminators]https://github.com/Automattic/mongoose/blob/master/test/model.hydrate.test.js#L70) uses the original schema at [line 88](https://github.com/Automattic/mongoose/blob/master/test/model.hydrate.test.js#L88)

### Reference change

[model / discriminator() / root schema inheritance / inherits validators](https://github.com/Automattic/mongoose/blob/master/test/model.discriminator.test.js#L288) would need to use a `deepStrictEqual()` as the reference is changed during the `schema.clone()`

vkarpov15:
Upon further review, I'm going to merge this and work through the issues you pointed out in #8546. Thanks for the detailed writeup and investigation :+1:

samgladstone:
Shall I leave that failing test to you? It's the one that is making sure the original schema's _id is not modified when creating a discriminator. I added it to test the case mentioned above.

Sorry for not being totally clear when I said the state of the branch was ok!

vkarpov15:
Yeah leave the failing test to me, I'll take care of it.

samgladstone:
Thanks 👍

PR #8548, opened 2020-01-28
Feature/single validate subpaths and instrumentation

taxilian:
After my recent pull request to fix parallel validation errors I discovered that we were still seeing the issue in production; it's less likely to happen, but can still happen. I added some instrumentation which stores stack traces to allow you to see how a given part of the tree was called which I was able to use to verify that it was still getting called twice.

It turns out that the check I had added before was still not catching everything; I believe this PR fixes that. Because it also ended up adding a lot of duplicate entries to the `paths` array and those are stepped through in a bunch of subsequent loops I did some testing and decided that using a Set for paths was the most efficient way to deal with preventing duplicates; that required some further refactors of the function in order to iterate through the Set, which can't be accessed by index (as far as I know). This might break some older versions of node; I don't know how far back you're wanting to support, so if you need me to change things to work better on older versions please let me know.

I believe all unit tests are still passing; some don't ever pass locally (I don't have sharding set up), so I'm waiting to see how the automated tests work.

### DIFF
--- a/lib/error/parallelValidate.js
+++ b/lib/error/parallelValidate.js
@@ -13,10 +13,18 @@ const MongooseError = require('./mongooseError');
  * @api private
  */
 
-function ParallelValidateError(doc) {
+function ParallelValidateError(doc, opts) {
   const msg = 'Can\'t validate() the same doc multiple times in parallel. Document: ';
   MongooseError.call(this, msg + doc._id);
   this.name = 'ParallelValidateError';
+  if (opts && opts.parentStack) {
+      // Provide a full async stack, most recent first
+      this.stack = this.stack + "\n\n" + opts.parentStack.join('\n\n');
+  }
+  // You need to know to look for this, but having it can be very helpful
+  // for tracking down issues when combined with the deepStackTrace schema
+  // option
+  this.conflictStack = opts && opts.conflictStack;
 }
 
 /*!
--- a/lib/error/parallelValidate.js
+++ b/lib/error/parallelValidate.js
@@ -18,8 +18,8 @@ function ParallelValidateError(doc, opts) {
   MongooseError.call(this, msg + doc._id);
   this.name = 'ParallelValidateError';
   if (opts && opts.parentStack) {
-      // Provide a full async stack, most recent first
-      this.stack = this.stack + "\n\n" + opts.parentStack.join('\n\n');
+    // Provide a full async stack, most recent first
+    this.stack = this.stack + '\n\n' + opts.parentStack.join('\n\n');
   }
   // You need to know to look for this, but having it can be very helpful
   // for tracking down issues when combined with the deepStackTrace schema
--- a/lib/document.js
+++ b/lib/document.js
@@ -2086,7 +2086,7 @@ function _getPathsToValidate(doc) {
   _evaluateRequiredFunctions(doc);
 
   // only validate required fields when necessary
-  let paths = Object.keys(doc.$__.activePaths.states.require).filter(function(path) {
+  let paths = new Set(Object.keys(doc.$__.activePaths.states.require).filter(function(path) {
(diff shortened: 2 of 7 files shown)


lib/error/parallelValidate.js
```diff
@@ -13,10 +13,18 @@ const MongooseError = require('./mongooseError');
  * @api private
  */
 
-function ParallelValidateError(doc) {
+function ParallelValidateError(doc, opts) {
   const msg = 'Can\'t validate() the same doc multiple times in parallel. Document: ';
   MongooseError.call(this, msg + doc._id);
   this.name = 'ParallelValidateError';
+  if (opts && opts.parentStack) {
+    // Provide a full async stack, most recent first
```
vkarpov15: Hmm I'd rather not support this option going forward - it seems like a one-off just to work around this particular issue. Would it be fine to just remove the `deepStackTrace` option? The rest of the PR looks great - I love the idea of switching to sets.

taxilian: Switching to sets was the most performant method I could find to avoid duplicates in the paths array -- I spent some time doing some benchmarks to be sure.

It's certainly up to you if you want to pull out the `deepStackTrace` option, but TBH I think it might be a better idea to expand it a little as long as we're now passing some state through all of the validators; specifically, I have been contemplating adding something that could track (only when an option is provided to avoid performance penalties otherwise) which validators have been called so that unit tests could be easily built to ensure that the validation is run for each path exactly once; based on what I've seen here it's clear that the validators are being run twice for some paths and I am less certain than I'd like to be that there isn't a way that validators could be skipped.

My personal feeling is that sometimes it makes sense to support a little bit of extra instrumentation if it enables more effective troubleshooting in the future.

That said, it's your project and I won't be offended if you disagree with me; let me know and I can strip it back out. I will say that when troubleshooting these issues having that option initially would have saved me hours or work.

vkarpov15: I'm ok with it being for instrumentation only, but right now it is a new schema level option, and I don't think it is useful enough for non-Mongoose devs to be that visible. So I'd say to remove it for now.

I'd love to write some tests verifying that validators only run once in all cases, but the tricky part is figuring out what "all cases" means. For example, none of our existing tests throw a ParallelValidateError, so clearly we're missing something. Do you know what code on your end causes that error?

taxilian: The reason the existing tests don't throw the error is that it's purely a timing issue; the only time it was happening is when the validation was on a subdocument which caused it to be async w/ nextTick and then it still had to start the second call before the first ended, which doesn't always happen. The test I added before reproduced it but there is an edge case which I haven't been able to totally narrow down where deep paths (like `keys.1.questions.0.clip.size`) are in `doc.$__.activePaths.states.modify` but the checks for `doc.isModified`, `!doc.isDirectModified`, and `!doc.$isDefault` in that block do not pass, which previously meant that it did not filter those subpaths out and so `paths` contained a lot of deep paths instead of just shallow paths; because of that the validator would run at this level but then would run at a later level as well.

The more changes there are to deep subdocuments the more likely it was to reproduce it, but since it is a timing issue it was very tricky to catch; I actually had to go back to 5.8.9 (with my earlier patch) to reproduce the issue locally because on 5.8.10 it didn't happen. I *think* that's because of the speed improvements in 5.8.10 but I'm not totally sure.

I am definitely not married to having it schema level, that was just the easiest place I could find to put it that I could get to. Is there a better place to put it? Maybe a global debug option of some sort? If so I could just change that. if not I can remove it

vkarpov15: I'll merge this and remove it for now. We can always bring it back, but if we remove ParallelValidateError for subdocuments, there won't be much need for that flag for the foreseeable future.

PR #8568, opened 2020-02-06
Fix: allow dynamic foreignField in virtual populate

BJvdA:
**Summary**

Allows the use of a function in `foreignField` when doing a virtual populate, so it can be dynamic just like `localField`.

https://mongoosejs.com/docs/populate.html#populate-virtuals

**Examples**

```
AssignmentSchema.virtual('assigned', {
  ref: doc => doc.adminId ? 'Admin' : 'User',
  localField: '_id',
  foreignField: doc => doc.adminId ? 'adminId' : 'userId',
  justOne: true
});
```

### DIFF
--- a/lib/helpers/populate/getModelsMapForPopulate.js
+++ b/lib/helpers/populate/getModelsMapForPopulate.js
@@ -92,6 +92,7 @@ module.exports = function getModelsMapForPopulate(model, docs, options) {
     const virtual = _virtualRes == null ? null : _virtualRes.virtual;
 
     let localField;
+    let foreignField;
     let count = false;
     if (virtual && virtual.options) {
       const virtualPrefix = _virtualRes.nestedSchemaPath ?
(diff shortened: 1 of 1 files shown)


lib/helpers/populate/getModelsMapForPopulate.js
```diff
@@ -101,13 +102,16 @@ module.exports = function getModelsMapForPopulate(model, docs, options) {
       } else {
         localField = virtualPrefix + virtual.options.localField;
       }
+      if (typeof virtual.options.foreignField === 'function') {
+        foreignField = virtualPrefix + virtual.options.foreignField.call(doc, doc);
```
vkarpov15: Looks like there's a bunch of test failures due to this change. I think they should be fixed if you get rid of the `virtualPrefix +` bit here. Should be just `foreignField = virtual.options.foreignField.call(doc, doc)` I think.

PR #8651, opened 2020-03-07
Correctly set message for ValidationError; npm audit fixes

dskrvk:
**Summary**

After commit d9163f561311642e36c79be4d40d396efe3f40af `ValidationError.message` is always an empty string. This reverts to the previous behavior where it was an informative value that could be printed.

Also fix some`npm audit` warnings for dev dependencies.

**Examples**

See tests.


package.json
```diff
@@ -127,6 +127,7 @@
       "no-buffer-constructor": "warn",
       "no-console": "off",
       "no-multi-spaces": "error",
+      "no-prototype-builtins": "warn",
```
vkarpov15: Why is this change here?

package.json
```diff
@@ -45,12 +45,12 @@
...
     "co": "4.6.0",
     "dox": "0.3.1",
-    "eslint": "5.16.0",
+    "eslint": "6.8.0",
     "eslint-plugin-mocha-no-only": "1.1.0",
     "highlight.js": "9.1.0",
     "lodash.isequal": "4.5.0",
     "lodash.isequalwith": "4.4.0",
-    "marked": "0.6.2",
+    "marked": "0.8.0",
```
vkarpov15: This change is a bit risky for a simple devDependency - a breaking change could break our docs build.

vkarpov15:
I undid the package.json changes, otherwise works for me :+1:

package.json
dskrvk: FWIW, I ran website.js and didn't see any errors - the docs were generated successfully.

package.json
dskrvk: See comment message - it's a check that's enabled in eslint-6.8 by default and is error-level.

dskrvk:
The motivation for upgrading the packages is to resolve npm audit warnings. I still believe they should be addressed.

dskrvk:
I can redo the change and run more verifications - is it just `make docs` in this case?

vkarpov15:
@dskrvk `make gendocs`, `make docs` will try to deploy :) If marked@0.8.0 doesn't change the output HTML at all, we can upgrade.

PR #8685, opened 2020-03-19
fix: typo in docs related to Schema.index

f3r:

**Summary**

When trying to use the [Schema.prototype.index()](https://mongoosejs.com/docs/api/schema.html#schema_Schema-index) method, to add a compound index to our schema, our team found that it didn't work as expected (it still allowed duplicated documents on those keys to be inserted).

```
schema.index({ first: 1, last: -1 }, {unique: true)
```

However, by research and chance, we found that by adding simple quotes to the keys, the problem was solved:

```
schema.index({ 'first': 1, 'last': -1 }, {unique: true)
```

Therefore, we have updated the documentation to prevent future users  from coming to this bug.

Thanks!

ps) Many thanks to @Ollebacx & @Marcosmm1 for their support!

### DIFF
--- a/docs/guide.pug
+++ b/docs/guide.pug
@@ -74,7 +74,7 @@ block content
     Notice above that if a property only requires a type, it can be specified using
     a shorthand notation (contrast the `title` property above with the `date`
     property).
-    
+
     Keys may also be assigned nested objects containing further key/type definitions
     like the `meta` property above.  This will happen whenever a key's value is a POJO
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
I created an issue with a reproduction script, trying with both single quotes and without doesn't seem to have an effect.

Could you check the out [the issue](https://github.com/Automattic/mongoose/issues/8688), and modify the script so it shows the issue you're having?

PR #8690, opened 2020-03-20
Fix broken HTML id references

AbdelrahmanHafez:
Fixes broken ids references in docs, for some reason some ids were wrapped in hyphens.


AbdelrahmanHafez:
Happy to help 😄

Issue #8331, opened 2019-11-13
[Bug] Model.bulkWrite wipes all the document data if no update provided


AbdelrahmanHafez:
If we for some reason (mostly by mistake) did not provide an `update` object in `updateOne` or `updateMany` operation in Model.bulkWrite, it wipes all the data inside the document except for the `_id` field.

Here's a test reproducing the issue.
```js
const mongoose = require('./index');
const { Schema } = mongoose;
const assert = require('assert');

const userSchema = new Schema({ name: { type: String, required: true } });

const User = mongoose.model('User', userSchema);

const user = new User({ name: 'Hafez' });

mongoose.connect('mongodb://127.0.0.1:27017/test');
async function run () {
  let threw = false;

  await User.deleteMany();

  await user.save();

  try {
    await User.bulkWrite([{ updateOne: { filter: { _id: user._id } } }]);
  }
  catch (error) {
    threw = true;
  }
  finally {
    assert.equal(threw, true, 'Not providing an update object inside bulk write should throw an error');

    const user = await User.findOne();
    assert.equal(user.name, 'Hafez', 'User data should not be wiped if did not provide an update object.');
  }

}


run().catch(console.error);
```

**What is the expected behavior?**

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

This usually happens when a developer is inserting an `update` object off by one curly brace, which is fairly easy to miss:
```js
User.bulkWrite([{ updateOne: { filter: { _id: user._id } }, update:{ name: 'Sam' } }])
```

The expected behavior: for bulkWrite to throw an error instead of just wiping all the data inside the documents .

PR #8693, opened 2020-03-20
refactor(model.test): prevent false positive in #8331 test

AbdelrahmanHafez:


test/model.test.js
```diff
@@ -5540,25 +5540,26 @@ describe('Model', function() {
 
         return co(function*() {
           const createdUser = yield User.create({ name: 'Hafez' });
-          let threw = false;
+
+          let err;
+
           try {
```
vkarpov15: Better to just do `const err = yield User.bulkWrite(ops).then(() => null, err => err);` so you don't need to write out the whole `try/catch` block.

vkarpov15: More on [golang-in-JS style error handling](http://thecodebarbarian.com/async-await-error-handling-in-javascript.html#golang-in-js)

AbdelrahmanHafez: Yes, that looks much cleaner. I just opened another PR with that modification.

PR #8732, opened 2020-03-27
Remove mongoose.set('debug', true); from test

AbdelrahmanHafez:


AbdelrahmanHafez:
Oh, didn't know that. I still find it more convenient to use `mongoose.set('debug', true);` before and after the parts I am interested in.

A bit more error-prone, but focused. Thanks for the tip!

Issue #8760, opened 2020-04-02
[Bug] populate with `clone: true` throws an error with `lean`


AbdelrahmanHafez:
**What is the current behavior?**
Can not use `clone: true` in populate along with `lean()`, it throws an error:
```
TypeError: doc.constructor.hydrate is not a function
    at assignRawDocsToIdStructure (node_modules\mongoose\lib\helpers\populate\assignRawDocsToIdStructure.js:58:29)
    at assignRawDocsToIdStructure (node_modules\mongoose\lib\helpers\populate\assignRawDocsToIdStructure.js:41:7)
    at assignVals (node_modules\mongoose\lib\helpers\populate\assignVals.js:26:3)
    at _assign (node_modules\mongoose\lib\model.js:4564:3)
    at _done (node_modules\mongoose\lib\model.js:4411:7)
    at _next (node_modules\mongoose\lib\model.js:4403:7)
    at node_modules\mongoose\lib\model.js:4837:16
    at node_modules\mongoose\lib\helpers\promiseOrCallback.js:24:16
    at node_modules\mongoose\lib\model.js:4860:21
    at node_modules\mongoose\lib\query.js:4366:11
    at node_modules\kareem\index.js:135:16
    at processTicksAndRejections (internal/process/task_queues.js:79:11)
Emitted 'error' event on Function instance at:
    at node_modules\mongoose\lib\model.js:4839:13
    at node_modules\mongoose\lib\helpers\promiseOrCallback.js:24:16
    [... lines matching original stack trace ...]
    at processTicksAndRejections (internal/process/task_queues.js:79:11)
```
**If the current behavior is a bug, please provide the steps to reproduce.**
```js
const mongoose = require('mongoose');
const { Schema } = mongoose;
const assert = require('assert');

mongoose.connect('mongodb://localhost:27017/test', { useNewUrlParser: true, useUnifiedTopology: true });

const blogPostSchema = new Schema({
  title: { type: String },
  commentsIds: [{ type: Schema.ObjectId, ref: 'Comment' }]
});

const commentSchema = new Schema({
  content: String
});

const BlogPost = mongoose.model('BlogPost', blogPostSchema);
const Comment = mongoose.model('Comment', commentSchema);

async function run () {
  await BlogPost.deleteMany();

  const blogPost = new BlogPost({ title: 'Important post' });
  const comment = new Comment({ content: 'Cool post.' });

  blogPost.commentsIds = [comment._id];

  await Promise.all([
[...]

PR #8761, opened 2020-04-02
Allow `clone: true` with `lean`

AbdelrahmanHafez:
fixes: #8760

### DIFF
--- a/lib/helpers/populate/assignRawDocsToIdStructure.js
+++ b/lib/helpers/populate/assignRawDocsToIdStructure.js
@@ -24,7 +24,7 @@ module.exports = assignRawDocsToIdStructure;
  * @api private
  */
 
-function assignRawDocsToIdStructure(rawIds, resultDocs, resultOrder, options, recursed) {
+function assignRawDocsToIdStructure(Model, rawIds, resultDocs, resultOrder, options, recursed) {
   // honor user specified sort order
   const newOrder = [];
   const sorting = options.sort && rawIds.length > 1;
@@ -38,7 +38,7 @@ function assignRawDocsToIdStructure(rawIds, resultDocs, resultOrder, options, re
 
     if (Array.isArray(id)) {
       // handle [ [id0, id2], [id3] ]
-      assignRawDocsToIdStructure(id, resultDocs, resultOrder, options, true);
+      assignRawDocsToIdStructure(Model, id, resultDocs, resultOrder, options, true);
       newOrder.push(id);
       continue;
     }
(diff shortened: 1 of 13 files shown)


lib/helpers/populate/assignRawDocsToIdStructure.js
```diff
@@ -55,7 +55,7 @@ function assignRawDocsToIdStructure(rawIds, resultDocs, resultOrder, options, re
     doc = resultDocs[sid];
     // If user wants separate copies of same doc, use this option
     if (options.clone) {
-      doc = doc.constructor.hydrate(doc._doc);
+      doc = Model.hydrate(options.lean ? doc : doc._doc);
```
vkarpov15, line 58: So this fixes the error, but this also makes it so that the child doc gets hydrated despite `lean()`. If memory serves, `lean()` propagates to `populate()` queries as well, so the expected behavior is that we don't hydrate the populated doc. We would need to clone it using `utils.clone()`.

AbdelrahmanHafez: Ah, good catch!

I changed that in the last commit, but for some reason using `doc = utils.clone(doc)` makes it so that the document is not assigned to the array. I verified that it does return a different copy from `doc`.

Can you think of a reason why it would do that?

AbdelrahmanHafez:
Rebased on master added many unnecessary commits to the PR, to review the files that actually changed: https://github.com/Automattic/mongoose/compare/master...AbdelrahmanHafez:8760

lib/helpers/populate/assignRawDocsToIdStructure.js
vkarpov15: No I'm not sure, I would need to do some debugging. I'll take a look later this week if you can't figure it out.

AbdelrahmanHafez: Yep, I spent 3 good hours trying to figure it out with no luck. Will need your help with this one.

vkarpov15:
Just FYI, I'd recommend not rebasing. There's virtually no benefit to rebase over squash+merge, it just causes headache in my experience. And we don't even do squash+merge because Mongoose wouldn't derive much benefit from it.

AbdelrahmanHafez:
Noted. I'll just pull the latest changes from `Automattic/mongoose` into my fork before creating a new branch.

vkarpov15:
@AbdelrahmanHafez I just made the changes in https://github.com/Automattic/mongoose/commit/b107d902fbb57a36855a08f4bc5dcb956efe5bf3 to avoid the extra commits in this branch. The issue is that you also need to update `leanPopulateMap` when cloning the populated doc - `populate()` filters out POJOs that aren't the result of a populate query:

https://github.com/Automattic/mongoose/blob/b5f744ed212e7db25cde944887f45b1aca33db13/lib/helpers/populate/assignVals.js#L193-L196
https://github.com/Automattic/mongoose/blob/b5f744ed212e7db25cde944887f45b1aca33db13/lib/helpers/populate/assignVals.js#L238

PR #8764, opened 2020-04-03
[WIP] pass custom options from save to Model.exists(...) when no changes are present

### LINKED ISSUE #8739: Wrong Options Used in Model.$__handleSave
Bug

When you try to save a document that has no changes, mongoose will run a `Model.exists()` query to make sure the document is in the DB (I don't know why this is done, but that's a different story). That query is done with a subset of the options that were passed to `Document.save()` call, not the original options. This breaks my middleware because `Model.exists()` triggers a `Document.findOne()` middleware, but the extra options that I have passed in are not sent with this new query.

The relevant line is 316 in lib/model.js.

```js
    } else {
      this.constructor.exists(this.$__where(), saveOptions)
        .then((documentExists)=>{
          if (!documentExists) throw new DocumentNotFoundError(this.$__where(),this.constructor.modelName);

          this.$__reset();
          callback();
        })
        .catch(callback);
      return;
    }
```

`saveOptions` there is a stripped down version of the original `options` passed in. I think `options` should be used there, or something similar to it.

```js
const doc = new MyModel({name: 'john'});
await doc.save({ customOpt: 1 }); // works
await doc.save({ customOpt: 1 }); // customOpt isn't passed to the Model.exists query
```

This seems like a far fetched case. But I ran into this by having a `Document.save()` call inside a `session.withTransaction()` call. `withTransaction` would then rerun the function after a `TransientTransactionError`, triggering this edge case where the document is saved twice.

The custom options should be passed through to the `Model.exists()` method.

Node: 12.14.0
Mongoose: 5.9.6
MongoDB: 4.2.3

AbdelrahmanHafez:
fixes #8739

This is WIP, creating a pull request to use the CI to see if I am making any breaking changes quickly.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -314,7 +314,8 @@ Model.prototype.$__handleSave = function(options, callback) {
         callback(null, ret);
       });
     } else {
-      this.constructor.exists(this.$__where(), saveOptions)
+      const optionsWithCustomValues = Object.assign({}, options, saveOptions);
+      this.constructor.exists(this.$__where(), optionsWithCustomValues)
         .then((documentExists)=>{
           if (!documentExists) throw new DocumentNotFoundError(this.$__where(), this.constructor.modelName);
 
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -2078,7 +2078,7 @@ describe('document', function() {
       });
     });
 
-    it('passes save custom options to Model.exists(...) when no changes are present', function() {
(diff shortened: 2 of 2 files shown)


AbdelrahmanHafez:
Alright, this should be good to go.

makinde:
Thanks guys

AbdelrahmanHafez:
Happy to help

PR #8795, opened 2020-04-14
Upgrade mongodb driver to v3.5.6

AbdelrahmanHafez:


AbdelrahmanHafez:
Upgrading broke a test with casting, will do some debugging soon.

vkarpov15:
:scream_cat: looks like MongoDB upgraded to bson 4.x in a patch release. Bold move. We need to upgrade our bson dep to match.

vkarpov15:
Fixed in https://github.com/Automattic/mongoose/commit/56a346d763d58c439ac391a37145b9f84c861345 . For some reason 2328c1d was still failing. Might be because of package-lock or some sort of caching on the Travis side.

PR #8799, opened 2020-04-15
Fix for updatedAt on sub-sub-documents

jskonst:
**Summary**
During the update of documents of the subdocument, we have faced the problem that field `updatedAt`, not updates on subdocument and it's a parent. 

**Examples**
Test `updates updatedAt when calling update on sub child` was provided in `timestamps.test.js` file. It fails on master branch and passes on our PR.


jskonst:
Will fix failed tests

PR #8817, opened 2020-04-18
Make Model.create() return an empty array when passed an empty array

### LINKED ISSUE #8792: Model.create(...) with an empty array should return an empty array
Currently `await Model.create([ ])` returns `undefined`, I think it would make things more consistent if we were to return an empty array instead of `undefined`, so people won't have to check if dynamically constructed arrays have items in them or not.

I found some tests asserting that create with an empty array returns undefined, is this for internal testing? Or would it count as breaking changes?

AbdelrahmanHafez:
fixes #8792


AbdelrahmanHafez:
Happy to help 👍

PR #8818, opened 2020-04-18
Make helpers.applyPaths(...) accept user provided value

### LINKED ISSUE #8806: $elemMatch (projection) is ignored due selected field in schema
$elemMatch (projection) is ignored due selected field in schema, for example schema:
```js
{name: String, attributes: {
  select: true,
  type: [{name: String, group: String}],
}}
```
Query:
```js
product.findOne({name: 'test'}).select({attributes: {$elemMatch: {group: 'beta'}}});
```
Executed query:
```js
product.findOne({name: 'test'}, {projection: {attributes: 1});
```

It seems that the issue happens in [applyPaths](https://github.com/Automattic/mongoose/blob/master/lib/queryhelpers.js#L119)

`$elemMatch` should be preserved

**What are the versions of Node.js, Mongoose and MongoDB you are using?**
Node.js@10, Mongoose@5.9.3, MongoDB: 4.x

AbdelrahmanHafez:
fixes #8806 

Also refactored analyzePaths to be defined at the end of the function instead of in the middle of it, following [the stepdown rule](https://dzone.com/articles/the-stepdown-rule).

### DIFF
--- a/lib/queryhelpers.js
+++ b/lib/queryhelpers.js
@@ -120,18 +120,17 @@ exports.applyPaths = function applyPaths(fields, schema) {
   // determine if query is selecting or excluding fields
   let exclude;
   let keys;
-  let ki;
-  let field;
+  let keyIndex;
 
   if (fields) {
     keys = Object.keys(fields);
-    ki = keys.length;
+    keyIndex = keys.length;
 
-    while (ki--) {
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
Happy to help 👍

PR #8823, opened 2020-04-20
[WIP] refactor: use (for-of) instead of for (let i = 0...) when iterating on arrays

AbdelrahmanHafez:
I've always found it shorter, and clearer to use for..of
```js
for (const item of arr) {
  // do stuff
}
```
instead of using indexes
```js
for (let i = 0; i < arr.length; i++) {
  const item = arr[i]; 
  // do stuff
}
```

Of course there are cases where indexes are necessary, those will not be changed, waiting to hear if @vkarpov15 approves of these changes, I'd like to take the time and refactor this wherever possible in the codebase.

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -370,8 +370,7 @@ Aggregate.prototype.unwind = function() {
   const args = utils.args(arguments);
 
   const res = [];
-  for (let i = 0; i < args.length; ++i) {
-    const arg = args[i];
+  for (const arg of args) {
     if (arg && typeof arg === 'object') {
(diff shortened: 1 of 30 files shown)


AbdelrahmanHafez:
Sweet, I'll change the PR from Draft to Open once I am done with the rest.

AbdelrahmanHafez:
Still working on this, re-opened it because I was trying to trigger a travis build. Apparently that's because of a conflict with master.

AbdelrahmanHafez:
Aaaand we should be good to go.

PR #8834, opened 2020-04-22
add node 13, and 14 to travis, and fix stream close event listener called multiple times on node 14

### LINKED ISSUE #8835: [Bug] closing a cursor on Node 14 calls on('close', callback); twice
On node 14 [streams close automatically](https://github.com/nodejs/node/pull/31408) which caused `cursor.on('close', callback);` to call callback twice, since we emit 'close' manually in mongoose.

To fix this we need to check whether the cursor is destroyed or not, if destroyed, do nothing.

AbdelrahmanHafez:
fixes #8835

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 sudo: false
-node_js: [12, 11, 10, 9, 8, 7, 6, 5, 4]
+node_js: [14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4]
 install:
   - travis_retry npm install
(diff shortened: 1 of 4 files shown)


AbdelrahmanHafez:
For some reason `stream.on('close', cb);` calls the callback twice on node 14, not sure if this is a bug in mongoose or node 14, will dig some more into this.

The test passes by replacing `stream.on(...)` with `stream.once(...)`, but I am sure there are users out there who are relying on `stream.on(...)` working correctly.

Issue #8821, opened 2020-04-19
Arrays don't validate empty strings


erezarnon:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Array elements which are empty strings don't run through validators

**If the current behavior is a bug, please provide the steps to reproduce.**

const TestModelSchema = new mongoose.Schema({
    array: {type: [{type: String, minlength: 1, maxlength: 128, required: true}], default: undefined},
})

const TestModel = mongoose.model('TestM', TestModelSchema);
TestModel.validate({array: ['']}).then(console.log)

**What is the expected behavior?**
Validation should give an error.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
nodejs 12.16.2
mongo 3.6.3
mongoose 5.9.9

AbdelrahmanHafez:
I can confirm that validation doesn't throw an error in this case, repro script:

```js
const mongoose = require('mongoose');
const { Schema } = mongoose;
const assert = require('assert');

run().catch(console.error);

async function run () {
  const userSchema = new Schema({
    friends: [{ type: String, required: true }]
  });

  const User = mongoose.model('User', userSchema);
  const err = await User.validate({ friends: [null] }).catch(err => err);

  assert.ok(err);

  console.log('Done.');
}
```

PR #8848, opened 2020-04-24
fix: validate nested paths on Model.validate(...)

AbdelrahmanHafez:
fixes: #8821

but I've been facing an issue with the errors being returned, I am getting {PATH} replaced with `friends.$` instead of `friends.0`, which would make all the errors in an array be omitted except for the last one.

@vkarpov15 I probably spent too much trying to figure this one out, any idea how I can make `SchemaType#doValidate(...)` change `friends.$` into `friends.0`?

**EDIT:** I figured it out, just needed to add `doValidateOptions: { path }`

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2136,10 +2136,10 @@ function _getPathsToValidate(doc) {
   }));
 
 
-  function addToPaths(p) { paths.add(p); }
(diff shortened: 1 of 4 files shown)


AbdelrahmanHafez:
Happy to help 👍

PR #8850, opened 2020-04-24
fix(schemaType): treat `select: undefined` and `select: null` as not specified

### LINKED ISSUE #8849: Schema select: undefined is treated as select: false
According to [Schema docs](https://mongoosejs.com/docs/api.html#schematype_SchemaType-select) select accepts boolean. Therefore when `select` has a value of `undefined` I expect that it will be ignored. Instead, it is treated as `false`.
Ignoring undefined would allow a more flexible definition of dynamic schemas. For example, skip fields on a particular environment without effecting other environments:
```js
new Schema({
  name: String,
  secretData: {
    select: process.env.NODE_ENV === 'production' ? false : undefined,
    type: String
  }
});
```

**Script to reproduce.**
```js
const { Schema } = mongoose;

run().catch(console.error);

async function run () {
    useNewUrlParser: true,
    useUnifiedTopology: true
  });

  const productSchema = new Schema({
    name: String,
    secretData: {
      select: undefined,
      type: String
    }
  });

  const Product = mongoose.model('Product', productSchema);

  await Product.create({ name: 'test', secretData: '42' });

  const product = await Product.findOne({ name: 'test' });

  assert.equal(product.secretData, 42);

  console.log('Done.');
}
```

**What are the versions of Node.js, Mongoose and MongoDB you are using?**

Node.js@10, Mongoose@5.9.10, MongoDB: 4.x

AbdelrahmanHafez:
fixes #8849

### DIFF
--- a/lib/schematype.js
+++ b/lib/schematype.js
@@ -55,10 +55,15 @@ function SchemaType(path, options, instance) {
   }
 
 
+  if (options.select == null) {
+    delete options.select;
+  }
+
   const Options = this.OptionsConstructor || SchemaTypeOptions;
   this.options = new Options(options);
   this._index = null;
-  this.selected;
+
+
   if (utils.hasUserDefinedProperty(this.options, 'immutable')) {
     this.$immutable = this.options.immutable;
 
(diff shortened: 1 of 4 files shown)


AbdelrahmanHafez:
I agree, removing the `done()` calls should've been in a separate PR.

What about refactors related to the issue I am working on? Refactoring is part of the process where I try to understand the code in question. Would separating the refactoring in different commits from the test/fix/feature commits be acceptable?

vkarpov15:
Yep refactoring in separate commits is fine.

Issue #8776, opened 2020-04-09
insertMany fails silently by returning empty array if all the documents are invalid


tusharf5:
First of all thanks to all the contributors & the author for such a useful and amazing MongoDB library.

**Do you want to request a *feature* or report a *bug*?**

Bug

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Create a schema with a required field.
2. Create array of 100 documents that don't have that required field.
3. Insert those docs using the below options
```js
const res = await Model.insertMany(docs, { ordered: false, rawResult: true });
// res is []
```

**What is the expected behavior?**
It should return with a object with `mongoose.validationErrors` property instead of silently failing and returning an empty array.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Node - node:12.16-stretch-slim
mongoose - ~5.9.7

AbdelrahmanHafez:
This is due to the fact that we're passing `ordered: false`. If you omit that, an error will be thrown.

I found the following in [the docs](https://mongoosejs.com/docs/api/model.html#model_Model.insertMany):

>Mongoose always validates each document before sending insertMany to MongoDB. So if one document has a validation error, no documents will be saved, unless you set the ordered option to false.

I think this behavior is by design and is inherited from the MongoDB driver since the native driver behaves [the same way](https://docs.mongodb.com/manual/reference/method/db.collection.insert/#db.collection.insert) when using `ordered: false`, it fails silently.

The script below passes, but correctly fails when we add `ordered: false`.
```js
const mongoose = require('mongoose');
const { Schema } = mongoose;
const assert = require('assert');

mongoose.connect('mongodb://localhost:27017/test', { useNewUrlParser: true, useUnifiedTopology: true });

const userSchema = new Schema({
  name: { type: String, required: true },
  age: Number
});

const User = mongoose.model('User', userSchema);

run().catch(console.error);

async function run () {
  await User.deleteMany();

  const err = await User.insertMany([{ age: 25 }]).catch(err => err);
  assert.ok(err.message.includes('Path `name` is required'));
}
```

tusharf5:
It does report a list of errors if they are coming from the database.

According to me `ordered: false` implies that if a doc fails (either **client side validation** or **database write error**) it won't stop the other docs from saving if they are valid and at the end it returns a list of all the validation or write errors of unsuccessful documents that were not saved.

[...]

Issue #7822, opened 2019-05-20
insertMany return blank if all inserted docs is error (duplicated error) but not if one insertable item to return validation error


eisenhaus335:
When i insert docs with everything is duplicated items i got blank response 
its really just give me this response [] (from console.log)

But when i give one something insertable. Its give me this response
```
{
    "result": {
        "ok": 1,
        "n": 1,
        "opTime": {
            "ts": "6693001219876585484",
            "t": 16
        }
    },
    "ops": [
        {
            "_id": "5ce2512c3dfc5d1704812804",
            "sku": "test15",
            "name": "barang",
            "qty": 20,
            "type": "pcs",
            "__v": 0
        }
    ],
    "insertedCount": 1,
    "insertedIds": {
        "0": "5ce2512c3dfc5d1704812804"
    },
    "mongoose": {
        "validationErrors": [
            {
                "errors": {
                    "sku": {
                        "message": "SKU already exists",
                        "name": "ValidatorError",
                        "properties": {
                            "message": "SKU already exists",
                            "type": "user defined",
                            "path": "sku",
                            "value": "test11"
                        },
                        "kind": "user defined",
                        "path": "sku",
                        "value": "test11"
                    }
                },
                "_message": "product validation failed",
                "message": "product validation failed: sku: SKU already exists",
                "name": "ValidationError"
            }
        ]
    }
}
```

```
const Product = require('../models/product')

async function add(ctx, next) {
    await Product.insertMany(ctx.request.body, { ordered: false, rawResult: true })
        .then(function (result) {
            console.log(result)
            ctx.body = {
                ...result
            }
        })
        .catch(function (err) {
            console.log(err)
            ctx.throw(401, 'something is wrong')
        })
}
```
PS: SKU is unique

Node.js: v12.2.0
"mongoose": "^5.5.9"
Mongodb: 4.0.9

vkarpov15:
What does your schema look like? Also, can you please post the custom validator that's reporting the 'SKU already exists' message?

eisenhaus335:
```
ProductSchema.path('sku').validate(async (value) => {
    const SKUCount = await mongoose.models.product.countDocuments({ sku: value });
    return !SKUCount;
}, 'SKU already exists');
```
const CreatedSchema = new mongoose.Schema({
    user: {
        type: mongoose.Schema.ObjectId,
        ref: 'user'
    },
    time: { type: Date, default: Date.now() }
}, { _id: false })
[...]

PR #8853, opened 2020-04-24
fix(model): return validation errors when all docs are invalid & rawR…

tusharf5:
fixes: #8776 , #7822 , partially #5337 

**Summary**

The current behaviour of `insertMany` when all the documents are invalid and `order : false` is returning a `[]`. (I think later on we would want to be able to provide those validation errors to the caller instead of just a `[]` as discussed here #5337)

When `rawResult` is also `true` and all the documents are invalid, currently it still returns a `[]`. By merging the PR, mongoose in this scenario will properly return a `{ mongoose : { validationErrors: [...errors] }` which it should as `rawResult : true`.

I've added tests which covers all the possible configurations for the above scenario using `order` & `rawResult` when the docs are all invalid.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3350,6 +3350,14 @@ Model.$__insertMany = function(arr, options, callback) {
     });
     // Quickly escape while there aren't any valid docAttributes
     if (docAttributes.length < 1) {
+      if (rawResult) {
+        const res = {
+          mongoose: {
+            validationErrors: validationErrors
+          }
+        };
+        return callback(null, res);
+      }
       callback(null, []);
       return;
     }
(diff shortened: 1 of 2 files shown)


tusharf5:
This is strange 🤔 One check failed on Node 7 with this error

```
schema select option
       "after each" hook for "all inclusive/exclusive combos work":
     Error: all indexes on collection dropped
```

I don't know if it's necessarily caused by the changes in this PR.

AbdelrahmanHafez:
Looks good to me, thank you 👍
As a side note, you should try co/yield, it's basically async/await for older environments.

Here's [an example](https://github.com/Automattic/mongoose/blob/49e78aced9812c75375a3151e9dfb691b18dbbe4/test/model.test.js#L6567).

tusharf5:
Sure, Thank you.

PR #8866, opened 2020-04-26
docs(model): add options.overwrite to findOneAndUpdate and findByIdAn…

### LINKED ISSUE #8865: [Docs] Documentation does not mention the "overwrite" option
When using functions such as `findByIdAndUpdate()`, we can pass the `overwrite: true` option to implement PUT instead of PATCH.

```
model.findByIdAndUpdate(id, modelData, { overwrite: true })
```

Unfortunately, the documentation does not mention that in any way.
There is a brief mention in the [Model.update](https://mongoosejs.com/docs/api/model.html#model_Model.update) section, but it does not suggest that this option can also be used with functions such as `findByIdAndUpdate()`.

AbdelrahmanHafez:
fixes #8865


mwanago:
Much appreciation for creating a PR in less than 20 minutes after an [issue](https://github.com/Automattic/mongoose/issues/8865) being submitted.

AbdelrahmanHafez:
Happy to help 😄

mwanago:
By the way, another alternative would be using `findOneAndReplace()`.
Unfortunately, it is not added to the TypeScript definitions in `@types/mongoose` which is a bit troublesome.
There seem to be some efforts made to change that, but not finalized. I might create a PR for that.

PR #8868, opened 2020-04-26
BREAKING CHANGE: call setters with `priorVal` as 2nd parameter, and with new subdocument as context if creating new document

### LINKED ISSUE #8629: Cleaner setup for setter context with single nested subdocs
Re: #8603, #7595, #7660, #5363.

Context is a bit messy when setting single nested subdocs because we want to show the prior subdoc's values to setters, but also we want to allow changing the document with `this.set()`.

```javascript
 const nestedSchema = Schema({
      name: String,
      status: { type: String, default: 'Pending' }
    });

    const Test = db.model('Test', {
      nested: {
        type: nestedSchema,
        set: subdoc => {
          // Subdoc should have old values... but then also `this.set()` should modify the updated document, while still handling defaults.
        }
      }
    });
```

Our current approach is:

1) Create new subdoc
2) Assign old subdoc's values to new subdoc
3) Run all setters and assignments
4) Remove all properties that weren't explicitly set, or set them to their defaults

For 6.0, I'd like to rework this to be simpler and more robust.

vkarpov15:
Fix #8629

There's some messy edge cases that come in when you have custom setters underneath nested subdocs, and you're creating a new subdoc. For example:

```javascript
      const Model = db.model('Test', Schema({
        name: String,
        profile: {
          type: Schema({ age: Number }, { _id: false }),
          set: function(v, priorVal) {
            // What should this be if `doc.profile.toObject() === { age: 29 }`
            // and I do `doc.profile = { age: 30 }`?
            this.age;
            // If `this.age === 29`, what happens if I change `this.age` in the setter?
            return v;
          }
        }
      }));
```

We've created some messy workarounds, but I think the cleanest way to do this going forward is to pass the current value as the 2nd parameter to the setter. In Mongoose 5, `this.age` will be 29, but with this PR `this.age` will be undefined but `priorVal` will be 29. I think this approach gives us the least code to maintain while also giving us a useful feature: accessing the prior value in a setter without relying on `this`.

What do you think @AbdelrahmanHafez ?

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -920,8 +920,6 @@ Document.prototype.$set = function $set(path, val, type, options) {
 
       return this;
     }
-  } else {
-    this.$__.$setCalled.add(path);
   }
 
   function _handleIndex(i) {
@@ -953,7 +951,6 @@ Document.prototype.$set = function $set(path, val, type, options) {
         !(this.schema.paths[pathName] &&
         this.schema.paths[pathName].options &&
         this.schema.paths[pathName].options.ref)) {
-      this.$__.$setCalled.add(prefix + key);
       this.$set(path[key], prefix + key, constructing, options);
     } else if (strict) {
       // Don't overwrite defaults with undefined keys (gh-3981)
--- a/lib/internal.js
+++ b/lib/internal.js
@@ -30,7 +30,6 @@ function InternalCache() {
   this.pathsToScopes = {};
   this.cachedRequired = {};
   this.session = null;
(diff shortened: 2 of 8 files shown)


AbdelrahmanHafez:
I like this is an elegant solution, from a user point of view, it's much less confusing than relying on `this`, and having access to the previous value and the current value is quite helpful, and the maintenance relief is a huge plus.

I think I would call the parameters `(newValue, previousValue)` instead of `(v, priorVal)` though.

Issue #8869, opened 2020-04-26
[Bug] reserved schema keys don't throw when type is nested object


valiant1x:
**Do you want to request a *feature* or report a *bug*?**
*bug*

**What is the current behavior?**

Passing the Schema constructor an object with the key `errors` causes a very obscure, non-obvious error to be thrown only once an associated Document is `save()`d or `create()`d, producing an error about priorDoc being undefined:

```
(node:6930) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'priorDoc' of undefined
    at /home/node/qproxy-api/node_modules/mongoose/lib/document.js:1118:27
    at model.$set (/home/node/qproxy-api/node_modules/mongoose/lib/document.js:1125:5)
    at model.set [as errors] (/home/node/qproxy-api/node_modules/mongoose/lib/helpers/document/compile.js:137:20)
    at model.Document (/home/node/qproxy-api/node_modules/mongoose/lib/document.js:85:15)
    at model.Model (/home/node/qproxy-api/node_modules/mongoose/lib/model.js:105:12)
    at new model (/home/node/qproxy-api/node_modules/mongoose/lib/model.js:4669:15)
    at /home/node/qproxy-api/node_modules/mongoose/lib/model.js:3106:22
    at /home/node/qproxy-api/node_modules/mongoose/lib/model.js:3142:7
    at Array.forEach (<anonymous>)
    at /home/node/qproxy-api/node_modules/mongoose/lib/model.js:3141:15
    at /home/node/qproxy-api/node_modules/mongoose/lib/helpers/promiseOrCallback.js:31:5
    at new Promise (<anonymous>)
    at promiseOrCallback (/home/node/qproxy-api/node_modules/mongoose/lib/helpers/promiseOrCallback.js:30:10)
    at Function.create (/home/node/qproxy-api/node_modules/mongoose/lib/model.js:3076:10)
    at Object.<anonymous> (/home/node/qproxy-api/test.js:11:7)
    at Module._compile (internal/modules/cjs/loader.js:1158:30)
(node:6930) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
[...]

Issue #7958, opened 2019-07-04
Conflicting mongoose once function (this.once is not a function)


venkateshwarlu-bavandla:
I have "once" field in my model doc. Which is conflicting while updating the existing doc, But while saving fresh doc am not getting.

Code:
```javascript
var mongoose = require('mongoose')

var options = {discriminatorKey: 'kind'};

var eventSchema = new mongoose.Schema({once: {
    booleanField: {type: String}
  }}, options)



var Event = mongoose.model('Event', eventSchema);

// ClickedLinkEvent is a special type of Event that has
// a URL.
var ClickedLinkEvent = Event.discriminator('ClickedLink',
  new mongoose.Schema({lookups: {
    type: [
      {
        hi: { type: String }
      }
    ]},url: String}, options));

var clickedEvent =
  new ClickedLinkEvent({
    once: {booleanField: true},
    lookups: [
    {
      hi: 'address1',
      name: 'address2',
    }], url: 'google.com'});

console.log(clickedEvent)
```
While updating existing doc am getting below error:
```javascript
Uncaught TypeError: this.once is not a function
      at model.Document (/node_modules/mongoose/lib/document.js:127:10)
      at model.Model (/node_modules/mongoose/lib/model.js:93:12)
      at new model (/node_modules/mongoose/lib/model.js:4542:13)
      at Object.createModel (/node_modules/mongoose/lib/queryhelpers.js:130:10)
      at completeOne (/node_modules/mongoose/lib/query.js:2663:26)
      at model.Query.Query._completeOne (/node_modules/mongoose/lib/query.js:1951:7)
      at Immediate.Query.base.findOne.call (node_modules/mongoose/lib/query.js:1993:10)
      at Immediate.<anonymous> (/node_modules/mquery/lib/utils.j
```

Can anybody suggest how can resolve this issue.

vkarpov15:
Found the issue and fixed it in master, fix will be in 5.6.4.

In the future, if you provide a code sample, can you please make sure the code sample demonstrates your issue? A code sample that works correctly and doesn't throw the error you indicated is not very helpful.

PR #8875, opened 2020-04-27
throw error on schema with reserved key with type of object

AbdelrahmanHafez:
fixes: #8869

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -572,7 +572,6 @@ reserved['prototype'] =
 // EventEmitter
 reserved.emit =
 reserved.on =
-reserved.once =
 reserved.listeners =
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
@vkarpov15 
We're having a test failing due to #7958, we probably allowed that because we didn't know that reserved keywords throw errors on simple types, and nested arrays, but not nested objects.

If my assumption about the reason we did that is correct, I suggest we allow `once` as an exception only to nested objects, and remove this exception as a breaking change on v6.0. Another option would be to move this whole PR to v6.0.
What do you think?

AbdelrahmanHafez:
Great, thanks for the tip!

PR #8885, opened 2020-04-28
inherit config for useDB from default connection

### LINKED ISSUE #8267: connection.UseDb not adopting connect options
BUG

When switching databases, the connect options are not adopted nor can we write them

1. Create a new mongoose connection via `mongoose.connect(<uri>, { <options> })`
2. Create a model that uses a different database 
3. Call `mongoose.connection.useDb('NewDB') 

Outcome: `(node:12285) DeprecationWarning: collection.ensureIndex is deprecated. Use createIndexes instead.`

That the useDb method would adopt the options from the original connect object. 

	`"mongoose": "^5.7.5",`

AbdelrahmanHafez:
fixes #8267


AbdelrahmanHafez:
Happy to help 👍

PR #8891, opened 2020-04-29
Update api.html

QuentinDanjou:
Aggregation cursor eachAsync function only get the document as parameter

> If you're making a change to documentation, do **not** modify a `.html` file directly. Instead find the corresponding `.pug` file or test case in the `test/docs` directory.

I haven't found how in the  mongoose/docs/api.pug  I could do this modification.

### DIFF
--- a/docs/api.html
+++ b/docs/api.html
@@ -4186,7 +4186,7 @@ aggregate.append(pipeline);</code></pre></div><hr class="separate-api-elements">
 <h4>Example:</h4>
 
 <pre><code><span class="hljs-keyword">var</span> cursor = Model.aggregate(..).cursor({ batchSize: <span class="hljs-number">1000</span> }).exec();
-cursor.eachAsync(<span class="hljs-function"><span class="hljs-keyword">function</span>(<span class="hljs-params">error, doc</span>) </span>{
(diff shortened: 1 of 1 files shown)


docs/api.html
```diff
@@ -4186,7 +4186,7 @@ <h4>Examples:</h4>
 <h4>Example:</h4>
 
 <pre><code><span class="hljs-keyword">var</span> cursor = Model.aggregate(..).cursor({ batchSize: <span class="hljs-number">1000</span> }).exec();
-cursor.eachAsync(<span class="hljs-function"><span class="hljs-keyword">function</span>(<span class="hljs-params">error, doc</span>) </span>{
+cursor.eachAsync(<span class="hljs-function"><span class="hljs-keyword">function</span>(<span class="hljs-params">doc</span>) </span>{
```
AbdelrahmanHafez, line 4189: Welcome @QuentinDanjou

The documentation is generated from two different places:
1- The API is generated from the JSDoc in the JS files.
2- The rest of the docs are generated from the .pug files.

So you probably wanna revert the HTML changes, and make those changes here instead:
https://github.com/Automattic/mongoose/blob/e8dc41c88949d020f4b7d08fabf696d317cdf4ce/lib/aggregate.js#L823

QuentinDanjou: Thank you! I will do it asap 😄

AbdelrahmanHafez:
fixed by #8973

PR #8896, opened 2020-04-29
allow adding descending indexes on schema

### LINKED ISSUE #8895: field level descending index?
documentation mentions;

```
  var animalSchema = new Schema({
    name: String,
    type: String,
    tags: { type: [String], index: true } // field level
  });

  animalSchema.index({ name: 1, type: -1 }); // schema level
```

is it even possible to create a descending index within field level?

tried
`    tags: { type: [String], index: -1 }`

but can't get it working?

AbdelrahmanHafez:
fixes #8895


AbdelrahmanHafez:
Should add an example in the docs to make it know that it's possible.

PR #8897, opened 2020-04-29
fix(virtualtype): order in applyGetters and applySetters

### LINKED ISSUE #8892: Schema.loadClass overwrite methods from inherited class
bug

if an class extends another class and has the same names for something like an getter / setter, then the bottom-most class's getter / setter is used

(Typescript)
```ts
// NodeJS: 12.16.2
// MongoDB: 4.2-bionic (Docker)
import * as mongoose from "mongoose"; // mongoose@5.9.10

class C1 {
  public get hello() {
    return 1;
  }
}

class C2 extends C1 {
  public get hello() {
    return 2;
  }
}

(async () => {
  await mongoose.connect(`mongodb://localhost:27017/`, { useNewUrlParser: true, dbName: "mongooseLoadClassInheritance", useCreateIndex: true, useUnifiedTopology: true });

  const C1Schema = new mongoose.Schema({});
  C1Schema.loadClass(C1);
  const C1Model = mongoose.model("C1", C1Schema);
  console.log("C1Model", ((new C1Model()) as any).hello); // expected: "1", result: "1"

  const C2Schema = new mongoose.Schema({});
  C2Schema.loadClass(C2);
  const C2Model = mongoose.model("C2", C2Schema);
  console.log("C2Model", ((new C2Model()) as any).hello); // expected: "2", result: "1"

  await mongoose.disconnect();
})();
```

to use the top-most class's getter/setter-

mongoose: 5.9.10
typescript: 3.8.3
nodejs: 12.16.2

ggurkal:
**Summary**

Virtual's getter and setter functions are called in reverse order. As for the example below, second function is called first, then the first function called next. However, the opposite would be the logical order.

```
SomeSchema.virtual('something').get(function() {
  return 'first!';
}).get(function() {
  return 'second';
})
```

Fixes #8892


hasezoey:
@AbdelrahmanHafez from what i see this pr is just extending the test for getters & setters (to also test class inheritance)

PR #8929, opened 2020-05-04
fix(query): Query#select({ field: false }) should not overwrite schema selection options

### LINKED ISSUE #8923: Model doesn't respect its subschema {select: false|true} properties

Model A has a subschema which it has a field marked as `select: false`. Model A projection doesn't respect its subschema and returns the whole subschema.

```node
// SubSchema
{
  firstName: {type: String, select: false},
  lastName: {type: String}
}

// Model A
{
  username: {type: String},
  subSchema: {type: SubSchema}
}

// Find one along with a projection
ModelA.findById(id, {username: false}).exec();

// Expected result > { subSchema: { lastName: 'Lenon' } }
// Actual result > { subSchema: { firstName: 'John', lastName: 'Lenon' } }
```

From the sample above, returns `{ subSchema: { lastName: 'Lenon' } }`.

`MongoDB: 4.2.5`
`Mongoose: 5.9.11`
`Node.js: 12.10.0`

AbdelrahmanHafez:
fixes #8923

### DIFF
--- a/lib/queryhelpers.js
+++ b/lib/queryhelpers.js
@@ -135,7 +135,7 @@ exports.applyPaths = function applyPaths(fields, schema) {
       if (!isDefiningProjection(field)) {
         continue;
       }
-      exclude = field === 0;
+      exclude = field === 0 || field === false;
       break;
     }
   }
--- a/lib/queryhelpers.js
+++ b/lib/queryhelpers.js
@@ -135,7 +135,7 @@ exports.applyPaths = function applyPaths(fields, schema) {
       if (!isDefiningProjection(field)) {
         continue;
       }
-      exclude = field === 0 || field === false;
+      exclude = !field;
       break;
     }
   }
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
I usually try to start a PR with a failing test, then wait till CI finishes up, then push the fix.

Apparently doing both commits too quickly causes the CI to use the commit with (test+fix) and falsely pass. Therefore the 3rd, and 4th commit.

lib/queryhelpers.js
```diff
@@ -135,7 +135,7 @@ exports.applyPaths = function applyPaths(fields, schema) {
       if (!isDefiningProjection(field)) {
         continue;
       }
-      exclude = field === 0;
+      exclude = field === 0 || field === false;
```
vkarpov15, line 138: I think it would be better for this to be `exclude = !field`. Better to use [falsy](https://masteringjs.io/tutorials/fundamentals/falsy) rather than hard coding `0` and `false` as the allowed values while leaving `0n` as equivalent to `true`.

vkarpov15:
I'd err on the side of not putting in an extra commit to trigger the test failure. Separate test and fix commits are fine though.

lib/queryhelpers.js
AbdelrahmanHafez: Sounds reasonable.

PR #8949, opened 2020-05-07
Add useMongoTimestamp schema option to use $currentDate for updatedAt

neuro-sys:
**Summary**

When using `timestamps` option on schema, Mongoose creates the updatedAt field in a pre save hook, which means the timestamp generation is not atomic with the update operation, and this results in out of order timestamps to be written to the database. Although this may be okay for many use cases, this renders `updatedAt` as not reliable for doing time based concurrency checks for consumers of this document.

This PR introduces a `useMongoTimestamp` option which, when set as true, will skip the pre-save generated updatedAt and instead use MongoDB `$currentDate` operator to generate the timestamp.

[$currentDate](https://docs.mongodb.com/v2.6/reference/operator/update/currentDate/) is available in MongoDB since version 2.6

**Examples**

Here's a session output from the test case provided with console.logs at important events:

```
2020-05-07T18:27:48.886Z: saved document with id: 5eb45323fa71b410d6d6b8ae updatedAt value: 2020-05-07T18:27:48.878Z
2020-05-07T18:27:49.384Z: saved document with id: 5eb45323fa71b410d6d6b8ae updatedAt value: 2020-05-07T18:27:47.879Z
2020-05-07T18:27:49.389Z: read  document with id: 5eb45323fa71b410d6d6b8ae updatedAt value: 2020-05-07T18:27:47.879Z
```

You will notice that the `updatedAt` field was inserted at `18:27:48.878Z`, and then the next save operation overwrote that with `18:27:47.879Z`.


lib/model.js
```diff
@@ -292,6 +292,11 @@ Model.prototype.$__handleSave = function(options, callback) {
 
     const delta = this.$__delta();
 
+    const useMongoTimestamp = this.schema.options.useMongoTimestamp;
+    if (useMongoTimestamp) {
+      delta[1]['$currentDate'] = { updatedAt: true };
```
neuro-sys, line 297: This is probably not correct since the user defined key may be different than `updatedAt` I suppose.

neuro-sys:
> Also, I'd rather then option be nested under timestamps, like new Schema({ name: String }, { timestamps: { createdAt: true, updatedAt: { useMongoTimestamp: true } } }), than make it a separate top-level schema option that doesn't do anything if timestamps isn't set.

I just noticed that this format conflicts with specifying custom key for `updatedAt`. How would you recommend to have it so that user can enable `useMongoTimestamp` and also specify a custom `updatedAt` key?

vkarpov15:
@neuro-sys that's a good point re: conflicting with custom path. I think `timestamps: { useMongoTimestamp: true } }` is a better alternative. The only downside is that we can't use `$currentDate` with `insertOne()` as you mentioned, but I think that's fine.

neuro-sys:
@vkarpov15 What are your thoughts, do you want to proceed or close the PR?

Issue #7671, opened 2019-04-01
Immutable properties


hugree:
**Do you want to request a *feature* or report a *bug*?**
Feature 

**What is the current behavior?**
There's no ability to tell what fields have changed if the upsert is called as the findOneAndUpdate executes. 

**What is the expected behavior?**
Added ways to check what fields have changed within a schema. Some immutability options would be nice too.

I know people have asked for this a lot. Some answers point to some pre/post hook solutions that work only for the standard save method and not the findOneAndUpdate. Others who want immutability use some 3rd party plugins, but from my investigation having those plugins and doing toObject transformations do not get along.

What's the best way of doing "let's make sure nobody changes this or that field" ever with mongoose?

vkarpov15:
There's no built-in way, but we can add support for this in the future. Can you point me to some plugins you've looked at for this case?

vkarpov15:
There's no good way that I know of to tell what changed in a `findOneAndUpdate()` if you use `new: true`.

PR #8952, opened 2020-05-08
fix(update): immutable properties are ignored in bulk upserts

### LINKED ISSUE #8951: Immutable properties are ignored in bulk upserts
This is almost the same issue as #8467 but the fix introduced in 8.9.5 (https://github.com/Automattic/mongoose/commit/a7c72e6fc8d4a190a63dd87ca760b6b710922015) does not seem to work for bulk upserts.

Here is the adapted repro from #8467:

```javascript
const Model = db.model('Test', Schema({
  name: String,
  age: { type: Number, default: 25, immutable: true }
}));

return co(function*() {
  yield Model.bulkWrite([
    {
      updateOne: {
        filter: { name: 'John' },
        update: { name: 'John', age: 20 },
        upsert: true,
        setDefaultsOnInsert: true
      }
    }
  ]);

  const doc = yield Model.findOne().lean();
  assert.equal(doc.age, 20);
});
```

I've spent some time debugging this and I think I have a fix, I'll open a PR in a minute.

philippejer:
Fixes #8951 

The cause of this issue, as I understand it:

- In `castUpdate()`, the keys of the update are taken _before_ the call to `moveImmutableProperties()` but this call can modify the keys (i.e. it adds a `$setOnInsert` key if it does not exist already and moves immutable properties under it).

- In the case of `bulkWrite()`, contrary to e.g. `updateOne()`, `_decorateUpdateWithVersionKey()` is not called, so the update object does not already have a `$setOnInsert` key. It is then ignored by `castUpdate()`  because `moveImmutableProperties()` is called too late.

The proposed solution:

- Moved `moveImmutableProperties()` a bit earlier, before the keys of the update are taken, so that `castUpdate()` works on the correct set of keys.

- I thought that would be enough, but actually it breaks a test for the `strict: 'throw'` behavior (see test for #7671), because if `moveImmutableProperties()` is called for a regular update (not an upsert), it will move the immutable properties under `$setOnInsert` and bypass the strictness check.

- The solution is to call `moveImmutableProperties()` only for upserts (so I had to propagate the `upsert` option to `castUpdate()` for this).

Edit: rebased to remove trailing blank lines.

### DIFF
--- a/lib/helpers/model/castBulkWrite.js
+++ b/lib/helpers/model/castBulkWrite.js
@@ -46,7 +46,8 @@ module.exports = function castBulkWrite(model, op, options) {
         });
         op['updateOne']['update'] = castUpdate(model.schema, op['updateOne']['update'], {
           strict: strict,
-          overwrite: false
+          overwrite: false,
+          upsert: op['updateOne'].upsert
         });
 
         if (op['updateOne'].setDefaultsOnInsert) {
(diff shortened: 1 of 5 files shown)


philippejer:
Happy to contribute, thanks for all your work 👍

PR #8970, opened 2020-05-11
test(model): fix flaky text for model.populate

AbdelrahmanHafez:
This test sometimes falsely fails, because documents aren't always fetched in the order they are `create(...)`ed in.

I'll look more into this, and see if create inserts documents in the order they're passed in, meanwhile, this should fix this test.


AbdelrahmanHafez:
I was under the impression that if no sort is specified, mongo will sort ascendingly by `_id`. Thanks for sharing the knowledge. 👍

PR #8973, opened 2020-05-11
fix(cursor): add index as second parameter to eachAsync callback

### LINKED ISSUE #8972: pass index as a second argument for eachAsync
re: #8962

Similar to how Bluebird's [Promise.map(...)](http://bluebirdjs.com/docs/api/promise.map.html) passes an index to parallel operations because maintaining your own counter in parallel can easily create bugs.
```js
let counter = 0;

await User.find().sort('order').cursor().eachAsync((doc, i) => {
  doStuffWithCounter(counter);

  // the right hand-side can be evaluated two times in parallel, where it will cause
  // incorrect end result for counter
  counter = counter + 1;
});
```

instead, we should:
```js
doStuffWithCounter(i);
```

Admittedly, I haven't been able to reproduce the parallel reassignment issue with the script above, but I've been bitten by that once before, and I am sure the people will find different ways to utilize it.

AbdelrahmanHafez:
fixes #8972 #8891

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -820,7 +820,7 @@ Aggregate.prototype.option = function(value) {
  * ####Example:
  *
  *     var cursor = Model.aggregate(..).cursor({ batchSize: 1000 }).exec();
- *     cursor.eachAsync(function(error, doc) {
+ *     cursor.eachAsync(function(doc, i) {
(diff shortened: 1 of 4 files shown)


AbdelrahmanHafez:
Happy to help 👍

Issue #8691, opened 2020-03-20
Improve stack trace of "Cast to ObjectId failed for value..." error


adamreisnz:
**Do you want to request a *feature* or report a *bug*?**
Improvement

**What is the current behavior?**
Current stack trace for this error is not very helpful, as it only logs Mongoose internals:

![image](https://user-images.githubusercontent.com/490562/77140080-4075cc80-6add-11ea-900f-b02827309539.png)

It is not possible to deduce where this error was thrown in the app.

Is there any way this can be improved so that the stack trace contains actual app lines to help us better figure out where this error is coming from?

For context, the error does specify the path and model, but we use this model in so many places that it's hard to find out what exactly is going wrong.

vkarpov15:
New stack trace looks like this:

```
$ node gh-8691.js 
{ CastError: Cast to ObjectId failed for value "bad" at path "_id" for model "Test"
    at new CastError (/home/val/Workspace/MongoDB/mongoose/lib/error/cast.js:29:11)
    at model.Query.exec (/home/val/Workspace/MongoDB/mongoose/lib/query.js:4331:21)
    at run (/home/val/Workspace/MongoDB/troubleshoot-mongoose/gh-8691.js:20:39)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

Just make sure you use `exec()` if you're using async/await. `await Model.findOne()` without `exec()` will unfortunately remove the calling file from the stack trace because of how V8's `await` handles custom thenables.

adamreisnz:
Hmm, we do use async/await extensively everywhere, and only use `exec()` when populating. Are you recommending usage of `exec()` everywhere for every query when using async/await?

vkarpov15:
@adamreisnz I've recently come to realize that using `exec()` gives you better stack traces. If you just use `await Model.findOne()`, the `exec()` call happens within `then()` in V8, and your stack trace won't include the `await Model.findOne()` line. I opened up an issue to investigate whether it is possible to get a better stack trace without explicitly calling `exec()`, but for now I'd recommend adding `exec()` if you're trying to track down where a cast error is coming from.

adamreisnz:
Ok thanks for the insight. It's not practical to replace it _everywhere_ in our code, but I'll try to add it in future code and slowly sprinkle it around our existing codebase.

I hope this can be fixed and the stack trace improved without calling `.exec()` in the future.

vkarpov15:
[...]

PR #8976, opened 2020-05-11
fix(model): add `kind` to cast errors thrown by query execution

### LINKED ISSUE #8953: Emitted error for Model.findById() with malformatted ID no longer contains field error.kind

Report a potential bug.

In version `^5.8.9` of mongoose error emitted after attempting to find a document using a malformatted (too short) ID contains field `error.kind` with value `"ObjectId"`. In version `^5.9.12` that I'm using the field is `undefined` in the same exact scenario.

`<someModel>.findById(<someTooShortID>)` throws an error not containing a field that was contained in an earlier version of mongoose, i. e. `error.kind`

`error.kind === "ObjectId"` should evaluate to `true` for the emitted error.

`"mongoose": "^5.9.12"`

AbdelrahmanHafez:
fixes #8953

This is caused by what I believe is a typo `type` => `kind` in https://github.com/Automattic/mongoose/commit/ee135ef90ec7afd48e12d83fbce9a137112fdac8 for #8691


AbdelrahmanHafez:
Happy to help 👍

Issue #8924, opened 2020-05-03
Virtual population with discriminator: TypeError: Cannot read property 'options' of null 


andreialecu:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**What is the current behavior?**

See the repro below, here's the error:

```
/home/aandrei/mongoosebug/node_modules/mongoose/node_modules/mongodb/lib/utils.js:132
      throw err;
      ^

TypeError: Cannot read property 'options' of null
    at getModelsMapForPopulate (/home/aandrei/mongoosebug/node_modules/mongoose/lib/helpers/populate/getModelsMapForPopulate.js:280:34)
    at populate (/home/aandrei/mongoosebug/node_modules/mongoose/lib/model.js:4313:21)
    at _populate (/home/aandrei/mongoosebug/node_modules/mongoose/lib/model.js:4283:5)
    at /home/aandrei/mongoosebug/node_modules/mongoose/lib/model.js:4258:5
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```js
"use strict";
const mongoose = require("mongoose");
const { Schema } = mongoose;
const assert = require("assert");

mongoose.set("debug", true);

run()
  .then(process.exit)
  .catch(console.error);

async function run() {
  await mongoose.connect(
    "mongodb://localhost:27017,localhost:27018,localhost:27019/test?replicaSet=rs",
    {
      useNewUrlParser: true,
      useUnifiedTopology: true,
    },
  );

  await mongoose.connection.dropDatabase();

  const userSchema = new Schema({ name: { type: String } });

  const User = mongoose.model("User", userSchema);

  const postSchema = new Schema({
    title: String,
    userId: { type: Schema.ObjectId },
  });
  postSchema.virtual("user", {
    ref: "User",
    localField: "userId",
    foreignField: "_id",
    justOne: true,
  });

  const Post = mongoose.model("Post", postSchema);

  const otherUserSchema = new mongoose.Schema({
    postId: { type: Schema.ObjectId },
  });
  otherUserSchema.virtual("post", {
    ref: "Post",
    localField: "postId",
    foreignField: "_id",
    justOne: true,
  });

  const OtherUser = User.discriminator("OtherUser", otherUserSchema);

  await User.init();

  const user = await User.create({ name: "User" });
  const post = await Post.create({ title: "JS 2020", userId: user._id });

  const user2 = await OtherUser.create({
    name: "Other User",
    postId: post._id,
  });

[...]

PR #8978, opened 2020-05-11
fix(populate): populate virtuals defined on child discriminators

AbdelrahmanHafez:
fixes: #8924

This attempts to find virtuals on the current schema, if nothing is found, it traverses depth-first in discriminators for that field.


AbdelrahmanHafez:
Yes, that will be an issue that people will need to know of, if the same virtual is present on two child discriminators, it'll just return the first matched virtual. In that case, they'll need to specifically use that discriminator.

One other issue is that with the current implementation we're searching depth-first, I think it'd make more sense to search breadth-first. I already created an issue #8980 for that.

PR #8991, opened 2020-05-13
fix(model): allow POJOs as schemas for model.discriminator(...)

### LINKED ISSUE #8984: make model.discriminator accept a POJO as a schema
Currently mongoose supports passing a POJO to `mongoose.model` and creates a schema out of it, but not `model.discriminator`. Would be nice if we support POJOs as a discriminator schema as well.

```js
// Works
const User = mongoose.model('User', {});

// Throws
const SuperUser = User.discriminator('SuperUser', { ability: String });
```

AbdelrahmanHafez:
fixes #8984

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1112,6 +1112,10 @@ Model.discriminator = function(name, schema, value) {
 
   _checkContext(this, 'discriminator');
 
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
Happy to help 👍

PR #9000, opened 2020-05-14
refactor: assert.doesNotThrow on tests instead of comments, remove done on synchronous tests

### LINKED ISSUE #8999: Remove unnecessary done calls on synchronous tests, explicit assert.doesNotThrow
Having `done()` calls on synchronous tests has no effect, since assertions throw an error on failure, which will cause the test to fail anyway.

I'll keep this issue open to remove those unnecessary `done()` calls to make it clear that when someone sees `done()` this test is asynchronous and requires waiting.

AbdelrahmanHafez:
fixes #8999


AbdelrahmanHafez:
>Plus, with assert.doesNotThrow(), you get an assertion error and lose the original type of the error.

I try to avoid writing comments that can be replaced with code, but the error type being lost is a fair reason to prefer comments over assert.doesNotThrow().

>@AbdelrahmanHafez any objection to merging this and continuing this work in a future PR?

Not at all, it's clear by now that I abandoned the project of refactoring more tests. Feel free to merge this PR, and #9024, I'll open new PRs if I come back to this again.

Issue #6566, opened 2018-06-07
10x slowdown on Windows when upgrading from Mongoose 4 to 5.


deklanw:
My tests are taking 10x as long on Windows when I upgrade from Mongoose 4 to 5. The slowdown doesn't appear to happen on my Ubuntu laptop.

I created a repo with two branches to illustrate the problem,
https://github.com/deklanw/pensieve-mongoose-4-to-5-speed-testing

I included the .env file for testing. If need be just change this line
`MONGODB_URI=mongodb://localhost:27017/pensieve-test`
to reflect your testing environment.

Just,
`yarn install`
`yarn test`

I tested on my Windows desktop and laptop to make sure. They're both Windows 10 x64. The problem occurred on both. The problem doesn't occur on my Ubuntu laptop.

Here are the versions:
`Windows 10 x64 laptop:
mongodb v3.6.5
node v10.1.0
yarn v1.6.0`
`Windows 10 x64 desktop:
mongodb v3.6.3
node v10.3.0
yarn v1.7.0`
The Ubuntu laptop has the same versions of everything as the Windows 10 desktop.

@lineus Tested this on OSX, I believe, and didn't find the slowdown. So it appears to be Windows related.

deklanw:
Here is the test output for the Mongoose 4 and 5 branches, respectively, on my Windows 10 desktop. Running the tests repeatedly gave similar results.

```$ node scripts/test.js
✨  Test database loaded  Cards controller
    GET /api/cards
GET /api/cards 200 21.914 ms - 345
      √ should return array of cards for user (44ms)
    POST /api/cards
POST /api/cards 200 24.228 ms - 278
      √ should create single card for user
    DELETE /api/cards/:id
DELETE /api/cards/5b195db0dd72452680504f5c 200 2.678 ms - 14
      √ should delete single card for user

  Decks controller
    GET /api/decks
GET /api/decks 400 0.135 ms - 64
      √ should return 400 if no authentication provided
GET /api/decks 200 7.843 ms - 335
      √ should return decks if token is provided
    POST /api/decks
POST /api/decks 200 2.426 ms - 275
      √ should create single deck for user
    DELETE /api/decks/:id
DELETE /api/decks/5b195db0dd72452680504f6d 200 2.033 ms - 14
      √ should delete single deck for user

  Sessions controller
    GET /api/sessions/:id
GET /api/sessions/5b195db0dd72452680504f67 200 4.788 ms - 368
      √ should return a single session for user

[...]

PR #9004, opened 2020-05-15
docs(faq): remove faq ipv6

AbdelrahmanHafez:
We added this question due to #6566

This is fixed in the driver somewhere along the way, and we don't need to manually use `family: 4` anymore.

```js

run().catch(console.error);

async function run() {
  const before = Date.now();
  console.log(`Connected in: ${Date.now() - before}ms.`);
}
```

Output:
```
Connected in: 22ms.
```


AbdelrahmanHafez:
Happy to help 👍

PR #9007, opened 2020-05-15
docs: add note about throwing error only after validation and fix broken reference to api/CastError

### LINKED ISSUE #8993: NaN casting doesnt behave as expected
Report a bug.

Entering non-number value for a path with Number as its SchemaType just casts to undefined.

```javascript
// Setup, taken from the docs
const mongoose = require("mongoose")
const schema = new Schema({ age: Number });
const Car = mongoose.model('Car', schema);

// All as expected:
new Car({ age: '15' }).age; // 15 as a Number
new Car({ age: true }).age; // 1 as a Number
new Car({ age: false }).age; // 0 as a Number
new Car({ age: { valueOf: () => 83 } }).age; // 83 as a Number

// These should result in a CastError, but instead return undefined
new Car({ age: NaN }).age;
new Car({ age: "test"}).age;
new Car({ age: [1, 2, 3, 4]}).age;
```

The [current docs](https://mongoosejs.com/docs/schematypes.html#numbers) say 

> NaN, strings that cast to NaN, arrays, and objects that don't have a valueOf() function will all result in a CastError.

So those last three statements should all result in a CastError. Instead, they return `undefined`.

Node.js: 12.16.2
Mongoose: 5.9.13 (same behavior on 5.7.7)
MongoDB: 4.2.6

AbdelrahmanHafez:
fixes #8993

### DIFF
--- a/docs/schematypes.pug
+++ b/docs/schematypes.pug
@@ -348,7 +348,7 @@ block content
     The values `null` and `undefined` are not cast.
 
     NaN, strings that cast to NaN, arrays, and objects that don't have a `valueOf()` function
-    will all result in a [CastError](/docs/api.html#mongooseerror_MongooseError.CastError) once validated, meaning that it will not throw on initialization, only when validated.
+    will all result in a [CastError](/docs/validation.html#cast-errors) once validated, meaning that it will not throw on initialization, only when validated.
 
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez:
Happy to help 👍

Issue #8259, opened 2019-10-18
Improve stack traces for MongoErrors


rvillane:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

 "mongoose": "5.7.1",

**What is the current behavior?**

I randomly get this error below

the connection code uses these settings:

```
mongoose.connect(dbURI, {
    reconnectTries: Number.MAX_VALUE,
    useNewUrlParser: true,
    useUnifiedTopology: true,
    useFindAndModify: false,
});
```
any recommendations ?

MongoError: server instance pool was destroyed
    at basicWriteValidations (/dist/node_modules/mongodb/lib/core/sdam/server.js:388:12)
    at executeWriteOperation (/dist/node_modules/mongodb/lib/core/sdam/server.js:415:17)
    at Server.insert (/dist/node_modules/mongodb/lib/core/sdam/server.js:337:5)
    at topology.selectServer (/dist/node_modules/mongodb/lib/core/sdam/topology.js:998:15)
    at selectServers (/dist/node_modules/mongodb/lib/core/sdam/topology.js:394:9)
    at selectServers (/dist/node_modules/mongodb/lib/core/sdam/topology.js:816:12)
    at NativeTopology.selectServer (/dist/node_modules/mongodb/lib/core/sdam/topology.js:381:5)
    at executeWriteOperation (/dist/node_modules/mongodb/lib/core/sdam/topology.js:966:12)
    at NativeTopology.insert (/dist/node_modules/mongodb/lib/core/sdam/topology.js:535:5)
    at NativeTopology.insert (/dist/node_modules/mongodb/lib/topologies/native_topology.js:58:11)
    at insertDocuments (/dist/node_modules/mongodb/lib/operations/common_functions.js:260:19)
    at InsertOneOperation.execute (/dist/node_modules/mongodb/lib/operations/insert_one.js:26:5)
    at executeOperation (/dist/node_modules/mongodb/lib/operations/execute_operation.js:83:26)
    at Collection.insertOne (/dist/node_modules/mongodb/lib/collection.js:476:10)
    at NativeCollection.(anonymous function) [as insertOne] (/dist/node_modules/mongoose/lib/drivers/node-mongodb-native/collection.js:166:28)
    at model.Model.$__handleSave (/dist/node_modules/mongoose/lib/model.js:262:33)
[...]

PR #9009, opened 2020-05-15
perf: delay constructing ServerSelectionError

zbjornson:
**Summary**

`MongooseServerSelectionError` is showing up as a dominator in our CPU profiles. As far as I can tell, this change has no behavior impact. The stack should be the same since `err`'s stack is assigned to `serverSelectionError` via `assimilateError`.

**Examples**

Run `node --inspect-brk ./benchmarks/benchjs/read.js` with everything but `Read - Mongoose - With lean` commented out, open the profiler (chrome devtools), start profiling, ...

Before:
image

After:
image

The benchmark values seem to be very noisy, but after a bunch of runs it comes out ~12% faster.

(The other error classes that show up in the above profile haven't shown up in our application profiles as dominators.)

---

Changing the Mongoose errors to use real ECMAScript classes that extend `Error` nearly doubles their performance (although avoiding them in the first place is better). I thought Mongoose needs to support back to Node.js v4 (classes without flags didn't come until v6), but I see plenty of usages of `class` in the code base. Are they okay to use?

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -4903,11 +4903,11 @@ Model.$handleCallbackError = function(callback) {
  */
 
 Model.$wrapCallback = function(callback) {
-  const serverSelectionError = new ServerSelectionError();
   const _this = this;
 
   return function(err) {
     if (err != null && err.name === 'MongoServerSelectionError') {
+      const serverSelectionError = new ServerSelectionError();
       arguments[0] = serverSelectionError.assimilateError(err);
     }
     if (err != null && err.name === 'MongoNetworkError' && err.message.endsWith('timed out')) {


AbdelrahmanHafez:
Thanks 👍

As for `Error` extension, we do need to support node 4, but that shouldn't stop us since we already use `'use strict';` on all files, and already use ES6 classes in some places of our code.

### 9009.js
```js
'use strict'; // would fail if we omit this

class ValidationError extends Error {
  constructor () {
    super();
    this.customField = 'custom value';
  }
}

const validationError = new ValidationError();
assert.equal(validationError.customField, 'custom value');

console.log('Node version: ', process.version);
```

### Output
```
Node version: v4.9.1
All assertions passed.
```

vkarpov15:
There is a behavior change with this change: different stack trace for server selection errors. I think you're assuming that `assimilateError()` takes the stack trace of the thrown error because of [this `Object.assign()`](https://github.com/Automattic/mongoose/blob/14bba6f80beea436f9aa003cf586c7327c96ffae/lib/error/serverSelection.js#L52-L54), but that doesn't happen because the `stack` property isn't enumerable on Mongoose errors.

For example, given the below script:

```javascript
'use strict';
  
const { Schema } = mongoose;

async function run() {
    useNewUrlParser: true,
    useUnifiedTopology: true,
    serverSelectionTimeoutMS: 1000,
    bufferCommands: false
  });

  const schema = Schema({ num: Number });

  console.log('Waiting...');
  await new Promise(resolve => setTimeout(resolve, 5000));

  await Model.create({ num: 42 });
  console.log('done');
}
```

I get the below stack without your change:

```
$ node gh-9009.js 
Waiting...
MongooseError [MongooseServerSelectionError]: connect ECONNREFUSED 127.0.0.1:27017
    at new MongooseServerSelectionError (/mongoose/lib/error/serverSelection.js:24:11)
    at Function.Model.$wrapCallback (/mongoose/lib/model.js:4928:32)
    at /mongoose/lib/model.js:475:27
    at promiseOrCallback (/mongoose/lib/helpers/promiseOrCallback.js:9:12)
    at model.Model.save (/mongoose/lib/model.js:474:10)
    at /mongoose/lib/model.js:3119:16
    at /mongoose/lib/model.js:3149:7
    at Array.forEach (<anonymous>)
    at /mongoose/lib/model.js:3148:15
    at /mongoose/lib/helpers/promiseOrCallback.js:31:5
    at new Promise (<anonymous>)
    at promiseOrCallback (/mongoose/lib/helpers/promiseOrCallback.js:30:10)
    at Function.create (/mongoose/lib/model.js:3083:10)
    at run (/troubleshoot-mongoose/gh-9009.js:27:15) {
  message: 'connect ECONNREFUSED 127.0.0.1:27017',
  name: 'MongooseServerSelectionError',
  [Symbol(mongoErrorContextSymbol)]: {}
}
```

Note that the `gh-9009.js` script is actually in the stack trace. With your change the stack trace won't show the calling script. This change was for #8259 because server selection error stack traces are not very helpful.

Any other way you can think of to speed up these errors? I would like to keep the stack traces if possible, although this may be a reason for us to start using async functions for [async stack traces](https://thecodebarbarian.com/async-stack-traces-in-node-js-12).

zbjornson:
Drat. Thanks for the repro script, but how do you get `ECONNREFUSED`? I tried using a port that I don't have MongoDB running on, but (a) I get identical stacks between master and this branch and (b) it's not at all the same stack as yours.

```
MongooseError [MongooseServerSelectionError]: Server selection timed out after 1000 ms
    at new MongooseServerSelectionError (C:\Users\zachb\git\mongoose\lib\error\serverSelection.js:24:11)
    at NativeConnection.Connection.openUri (C:\Users\zachb\git\mongoose\lib\connection.js:823:32)
    at Mongoose.connect (C:\Users\zachb\git\mongoose\lib\index.js:333:15)
    at run (C:\Users\zachb\git\mongoose\9009.js:12:18)
    at Object.<anonymous> (C:\Users\zachb\git\mongoose\9009.js:9:1)
    at Module._compile (internal/modules/cjs/loader.js:936:30)
    ⋮ (node.js stuff)
```

If this issue is simply that `stack` isn't enumerable, then I think changing `assimilateError` to explicitly copy that property would work? Or `Object.getOwnPropertyDescriptors/Names/Symbols` if we need to handle more properties than that.

(That script also made me realize that `reason` is not an own-property of the assimilated error in my other PR #9021, fixed.)

vkarpov15:
@zbjornson the reason why you're getting the same stack trace is because `mongoose.connect()` creates a server selection error here: https://github.com/Automattic/mongoose/blob/7aa470ea9c68272243e5b4e064a729fb44fd48fa/lib/connection.js#L823 . In my example, you need to shut down mongodb between the initial `mongoose.connect()` and the `create()` call to get this error.

The purpose of creating a `ServerSelectionError` in `$wrapCallback()` is to make it so that queries have readable stack traces in the event of a server selection error. If we wait to create the `ServerSelectionError` until we get a response from the MongoDB server, the initial stack trace gets lost.

Either way, I'm going to close this PR in favor of #9021. Although we should do the same thing for `CastError`, `VersionError`, `ParallelValidateError`, and `ParallelSaveError`. Because we use the same pattern of eagerly creating errors to keep a useful stack trace for all those errors, which is why they show up in your profiler.

PR #9013, opened 2020-05-16
fix(documentarray): make sure you can call `unshift()` after `map()`

### LINKED ISSUE #9012: Error when calling unshift() after map()
This is almost the same issue as #8399 (root cause seems to be #8356).

Here is a repro:

```javascript
const MyModel = db.model('Test', Schema({
  myArray: [{ name: String }]
}));

const doc = new MyModel({
  myArray: [{ name: 'b' }, { name: 'c' }]
});
let myArray = doc.myArray;

myArray = myArray.map(val => ({ name: `${val.name} mapped` }));

myArray.unshift({ name: 'a inserted' });

assert.deepEqual(myArray.map(v => v.name), [
  'a inserted',
  'b mapped',
  'c mapped'
]);
```

I'll create a PR in a minute for this issue.

philippejer:
Fix #9012

### DIFF
--- a/lib/types/core_array.js
+++ b/lib/types/core_array.js
@@ -879,7 +879,7 @@ class CoreMongooseArray extends Array {
    *
    * ####Note:
    *
-   * _marks the entire array as modified, which if saved, will store it as a `$set` operation, potentially overwritting any changes that happen between when you retrieved the object and when you save it._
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
LGTM.
Thanks 👍

Issue #6578, opened 2018-06-11
Configuration to make runValidators default to true?


joeytwiddle:
Until I discovered [runValidators](http://mongoosejs.com/docs/validation.html#update-validators) recently, I had always assumed mongoose did validation on updates by default. But it doesn't!

Would it be possible to add a global config or a schema config that would make `runValidators` default to `true` for all updates? (We could then override with `false` on individual queries, if we need to.)

----

The documentation says:

> Be careful: update validators are off by default because they have several caveats.

And the caveats are well documented below that.

In our apps, we tend to use only simple validators, so most of the caveats don't apply. If we do expand our validators, we could either turn off the option for that schema, or adapt our validators and update queries as per the docs.

vkarpov15:
That's a good idea. Right now the way to do it would be with a global plugin:

```javascript
mongoose.plugin(schema => {
  schema.pre('findOneAndUpdate', setRunValidators);
  schema.pre('updateMany', setRunValidators);
  schema.pre('updateOne', setRunValidators);
  schema.pre('update', setRunValidators);
});

function setRunValidators() {
  this.setOptions({ runValidators: true });
}
```

Would be neat to have this as a one-liner though.

joeytwiddle:
Some other nice global one-liners might be:

- Make all fields `required: true` by default.
- Make all schemas `strict: 'throw'` by default.

I have been using those policies in our projects (implemented in our common model builder function) because it helps to detect typos and other developer mistakes sooner. We override the defaults when we need to.

So if you are thinking of introducing some global settings in future, let's try to find something that works neatly for all of those!

Anyway the example above is good enough for me, for now, thanks.

vkarpov15:
The `required: true` one I'm not sure is a good idea. I like the `strict: 'throw'` idea though, opened up #6658 to track.

AbdelrahmanHafez:
The code below works, and throws an error as expected.

```js
const mongoose = require('mongoose');
const { Schema } = mongoose;

mongoose.plugin(schema => {
  schema.pre('findOneAndUpdate', setRunValidators);
  schema.pre('updateMany', setRunValidators);
  schema.pre('updateOne', setRunValidators);
  schema.pre('update', setRunValidators);
});

function setRunValidators () {
  this.setOptions({ runValidators: true });
}


mongoose.connect('mongodb://localhost:27017/test', { family: 4, useNewUrlParser: true });


const userSchema = new Schema({
  name: { type: String, required: true },
[...]

PR #9016, opened 2020-05-16
fix(base): allow global option mongoose.set('strictQuery', true);

### LINKED ISSUE #6658: Global strict mode configuration
https://github.com/Automattic/mongoose/issues/6578#issuecomment-402029769

`mongoose.set('strict', 'throw')` to make all schemas use `strict: 'throw'` by default.
`mongoose.set('strictQuery', 'throw')` to do the same for [`strictQuery`](http://mongoosejs.com/docs/guide.html#strictQuery)

AbdelrahmanHafez:
fixes #6658 for strictQuery

### DIFF
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -2431,33 +2431,32 @@ describe('schema', function() {
     });
   });
 
-  it('setting `strictQuery` on base sets strictQuery to schema (gh-6658)', function() {
-    // Arrange
-    const originalValue = mongoose.get('strictQuery');
-    mongoose.set('strictQuery', 'some value');
+  describe('mongoose.set(`strictQuery`, value); (gh-6658)', function() {
+    let strictQueryOriginalValue;
 
-    // Act
-    const schema = new Schema();
+    this.beforeEach(() => strictQueryOriginalValue = mongoose.get('strictQuery'));
+    this.afterEach(() => mongoose.set('strictQuery', strictQueryOriginalValue));
 
-    // Assert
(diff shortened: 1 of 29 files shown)


test/schema.test.js
```diff
@@ -2430,4 +2430,33 @@ describe('schema', function() {
       assert.throws(buildInvalidSchema, /`db` may not be used as a schema pathname/);
     });
   });
+
+  describe('mongoose.set(`strictQuery`, value); (gh-6658)', function() {
+    let strictQueryOriginalValue;
+
+    this.beforeEach(() => strictQueryOriginalValue = mongoose.get('strictQuery'));
+    this.afterEach(() => mongoose.set('strictQuery', strictQueryOriginalValue));
```
AbdelrahmanHafez, line 2438: @vkarpov15 I'd like to know what you think of this approach.

I initially made the cleanup in each test, but if one of those tests had failed, the cleanup wouldn't have happened, causing more tests to fail, making it difficult to find the root cause of all the failing tests.

I think this mitigates this concern, but I still don't like it. How would you do that?

vkarpov15: This approach is ok, but the better approach would be to create a new Mongoose instance using `const _mongoose = new mongoose.Mongoose();` and use `_mongoose.set()`. Check out the test for #6578 for an example.

AbdelrahmanHafez: Wouldn't this attempt to create a new connection with the same URI for the database? Making tests run significantly slower?

PR #9017, opened 2020-05-16
refactor: remove CoreMongooseArray#map, defined twice

AbdelrahmanHafez:
No need to add it, already defined on https://github.com/Automattic/mongoose/blob/bfccbe53d75bd71eda2a45361bf7c1196672e273/lib/types/core_array.js#L961


AbdelrahmanHafez:
Happy to help 👍

PR #9021, opened 2020-05-16
refactor(error): convert errors to classes extending Error

zbjornson:
**Summary**

Converts all errors to real ECMAScript classes that extend Error. This gives a significant performance improvement.

CPU profiles for one of the read benchmarks:

Before:
image

After:
image

I also corrected a few of the JSDoc type annotations that looked like copy/paste errors along the way.

### DIFF
--- a/lib/error/objectExpected.js
+++ b/lib/error/objectExpected.js
@@ -6,33 +6,25 @@
 
 const MongooseError = require('./');
 
-/**
- * Strict mode error constructor
- *
- * @param {String} type
- * @param {String} value
- * @inherits MongooseError
- * @api private
- */
 
-function ObjectExpectedError(path, val) {
-  const typeDescription = Array.isArray(val) ? 'array' : 'primitive value';
-  MongooseError.call(this, 'Tried to set nested object field `' + path +
-    `\` to ${typeDescription} \`` + val + '` and strict mode is set to throw.');
-  this.name = 'ObjectExpectedError';
-  if (Error.captureStackTrace) {
-    Error.captureStackTrace(this);
-  } else {
-    this.stack = new Error().stack;
+class ObjectExpectedError extends MongooseError {
+  /**
+   * Strict mode error constructor
+   *
+   * @param {string} type
+   * @param {string} value
+   * @api private
+   */
+  constructor(path, val) {
+    const typeDescription = Array.isArray(val) ? 'array' : 'primitive value';
+    super('Tried to set nested object field `' + path +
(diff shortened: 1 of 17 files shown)


lib/error/objectExpected.js
```diff
@@ -6,33 +6,25 @@
...
- * @inherits MongooseError
- * @api private
- */
 
-function ObjectExpectedError(path, val) {
-  const typeDescription = Array.isArray(val) ? 'array' : 'primitive value';
-  MongooseError.call(this, 'Tried to set nested object field `' + path +
-    `\` to ${typeDescription} \`` + val + '` and strict mode is set to throw.');
-  this.name = 'ObjectExpectedError';
-  if (Error.captureStackTrace) {
```
AbdelrahmanHafez, line 23: I noticed we're omitting all `Error.captureStackTrace` here, I am not sure how it works. Is it not required when using the ES6 class syntax somehow? If so, could you explain why that is?

zbjornson: Right, it's unnecessary. Here's a short example of class-based, function-based and function-based with `captureStackTrace`: https://repl.it/repls/RipeOrdinaryBucket

Output:

```
Class-based
MyErrorSub
    at fn3 (/home/runner/RipeOrdinaryBucket/class.js:6:24)
    at fn2 (/home/runner/RipeOrdinaryBucket/class.js:5:18)
    at fn1 (/home/runner/RipeOrdinaryBucket/class.js:4:46)
    at /home/runner/RipeOrdinaryBucket/index.js:9:5
    ...

Function-based
[MyErrorSub]

Function with captureStackTrace
MyErrorSub
    at new MyErrorSub (/home/runner/RipeOrdinaryBucket/fnWithCapture.js:12:8)
    at fn3 (/home/runner/RipeOrdinaryBucket/fnWithCapture.js:19:24)
    at fn2 (/home/runner/RipeOrdinaryBucket/fnWithCapture.js:18:18)
    at fn1 (/home/runner/RipeOrdinaryBucket/fnWithCapture.js:17:66)
    at /home/runner/RipeOrdinaryBucket/index.js:9:5
    ...
````

Note that the stack frame for `MyErrorSub` itself is omitted in the class-based version compared to `captureStackTrace` (good).

Also, I used `Object.defineProperty(proto, 'name')` so that `name` is not enumerable. This matches native JS Errors.

AbdelrahmanHafez: Ah great, thanks 👍

PR #9023, opened 2020-05-17
update returns updateWriteOpResult

dandv:
I think `update` returns [updateWriteOpResult](https://mongodb.github.io/node-mongodb-native/3.6/api/Collection.html#~updateWriteOpResult) rather than writeOpResultObject, but please check.

Also made a few absolute URLs, relative.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3636,7 +3636,7 @@ Model.hydrate = function(obj) {
  * @see response http://docs.mongodb.org/v2.6/reference/command/update/#output
  * @param {Object} filter
  * @param {Object} doc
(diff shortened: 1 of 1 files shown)


lib/model.js
```diff
@@ -3636,8 +3636,8 @@ Model.hydrate = function(obj) {
  * @see response http://docs.mongodb.org/v2.6/reference/command/update/#output
  * @param {Object} filter
  * @param {Object} doc
- * @param {Object} [options] optional see [`Query.prototype.setOptions()`](http://mongoosejs.com/docs/api.html#query_Query-setOptions)
```
AbdelrahmanHafez, line 3639: You mistakenly removed `/docs` here, please add it

PR #9031, opened 2020-05-19
fix(bulkwrite): cast filter & update to schema after applying timestamps

### LINKED ISSUE #9030: BulkWrite won't honor data type for timestamps when using updateOne / updateMany
Report a bug

- When using `bulkWrite` for `updateOne` / `updateMany`, it does not honors data type for timestamps.
- According to docs, a timestamp can be stored as an `ISODate` (Default) or an epoch timestamp (via setting the timestamp field as `Number`).
- Mongoose does honors this requirement when using standard `updateOne` / `updateMany` operations, but fails and reverts back to `ISODate` when using same operation via `bulkWrite`

```js
// mongoose v5.9.15
// mongodb v4.0.4

// connection
mongoose.set('debug', true);
mongoose.connect('mongodb://localhost:27017/sandbox', {
  useNewUrlParser: true,
  useUnifiedTopology: true
});

// model
  name: String,
  created_at: Number,
  updated_at: Number,
}, {
  timestamps: {
    createdAt: 'created_at',
    updatedAt: 'updated_at',
  },
});
const User = mongoose.model('User', UserSchema);

// operation
async function run() {
  await User.deleteMany();
  const userInitialTimestamp = Date.now();

  const user1 = await User.create({name: 'user1'});

  assert(user1 instanceof User);
  assert(typeof user1.updated_at === 'number');
  assert(user1.updated_at > userInitialTimestamp);

  await User.updateOne({
    _id: user1._id,
  }, {
    $set: {
      name: 'userAnother1',
    },
  });

  // works!
  // the actual POJO object still has epoch timestamp for updated_at field
  const userAnother1 = await User.findById(user1._id).lean();
  assert(!(userAnother1 instanceof User));
  assert(userAnother1.name === 'userAnother1');
  assert(typeof userAnother1.updated_at === 'number');
  assert(userAnother1.updated_at > user1.updated_at);

  await User.bulkWrite([{
    updateOne: {
      filter: {
        _id: user1._id,
      },
      update: {
        $set: {
          name: 'userAnother2',
        },
      },
    },
  }]);
  // bulkwrite won't honor the data type for update_at field
  // sends ISODate over the wire instead of epoch timestamp (check debug)

  // this will still work!
  // mongoose will convert the ISODate to epoch timestamp
  const userAnother2 = await User.findById(user1._id);
  assert(userAnother2 instanceof User);
  assert(userAnother2.name === 'userAnother2');
  assert(typeof userAnother2.updated_at === 'number');
  assert(userAnother2.updated_at > userAnother1.updated_at);

  // but this wont!
  // because actual value stored is ISODate, not epoch timestamp anymore
  const userAnother3 = await User.findById(user1._id).lean();
  assert(!(userAnother3 instanceof User));
  assert(userAnother3.name === 'userAnother2');
  assert(typeof userAnother3.updated_at === 'number');
  assert(userAnother3.updated_at > userAnother1.updated_at);
}

run()
  .catch(error => console.error(error.stack))
  .finally(async () => {
    await mongoose.connection.close();
  });
```

Mongoose should have stored epoch timestamp instead of ISODate when `bulkWirte` was run.

- Mongoose - 5.9.15
- MongoDb - 4.0.4
- Node - 10.14.1

AbdelrahmanHafez:
fixes #9030

### DIFF
--- a/lib/helpers/model/castBulkWrite.js
+++ b/lib/helpers/model/castBulkWrite.js
@@ -45,15 +45,15 @@ module.exports = function castBulkWrite(originalModel, op, options) {
         const strict = options.strict != null ? options.strict : model.schema.options.strict;
 
         _addDiscriminatorToObject(schema, op['updateOne']['filter']);
-        op['updateOne']['filter'] = cast(model.schema, op['updateOne']['filter'], {
-          strict: strict,
-          upsert: op['updateOne'].upsert
-        });
-        op['updateOne']['update'] = castUpdate(model.schema, op['updateOne']['update'], {
-          strict: strict,
-          overwrite: false,
-          upsert: op['updateOne'].upsert
-        });
+
+        if (model.schema.$timestamps != null && op['updateOne'].timestamps !== false) {
+          const createdAt = model.schema.$timestamps.createdAt;
+          const updatedAt = model.schema.$timestamps.updatedAt;
+          applyTimestampsToUpdate(now, createdAt, updatedAt, op['updateOne']['update'], {});
+        }
+
+        applyTimestampsToChildren(now, op['updateOne']['update'], model.schema);
+
 
         if (op['updateOne'].setDefaultsOnInsert) {
           setDefaultsOnInsert(op['updateOne']['filter'], model.schema, op['updateOne']['update'], {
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
Happy to help 👍

Issue #9039, opened 2020-05-20
Default fields in embedded document are overwritten with undefined


vitorhnn:
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
Passing undefined as a value to a key in a embedded document causes Mongoose to first set the key to its default value and then to overwrite it with undefined
**If the current behavior is a bug, please provide the steps to reproduce.**
```javascript
it('Sets default when passing undefined as value for a key in a nested subdoc (gh-9039)', function () {
  const Test = db.model('Test', {
    nested: {
      prop: {
        type: String,
        default: 'some default value'
      }
    }
  });

  return co(function* () {
    const doc = yield Test.create({ nested: { prop: undefined } });
    assert.equal(doc.nested.prop, 'some default value');
  });
});
```

**What is the expected behavior?**
See the provided test

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Node: 14.2.0
Mongoose: 5.9.15
MongoDB: 3.6.17

AbdelrahmanHafez:
I think this is by design, many people would count on mongoose to set a field to the default if the value provided to it is `undefined`. If you'd like to bypass the default by explicitly setting the value to unknown, `null` is the correct option. 

`null` is known (explicit) unknown, `undefined` is unknown unknown.

PR #9040, opened 2020-05-20
fix(document): use the complete path name of a property when avoiding overwriting default values with undefined

vitorhnn:
**Summary**
This prevents Mongoose from overwriting default values with undefined for nested documents

**Examples**
See #9039

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -957,9 +957,9 @@ Document.prototype.$set = function $set(path, val, type, options) {
       this.$__.$setCalled.add(prefix + key);
       this.$set(path[key], prefix + key, constructing, options);
     } else if (strict) {
-      // Don't overwrite defaults with undefined keys (gh-3981)
+      // Don't overwrite defaults with undefined keys (gh-3981) (gh-9039)
       if (constructing && path[key] === void 0 &&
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
I think this is by design, many people would count on mongoose to set a field to the default if the value provided to it is `undefined`. If you'd like to bypass the default by explicitly setting the value to unknown, `null` is the correct option.

`null` is known (explicit) unknown, `undefined` is unknown unknown.

vitorhnn:
My apologies if I wasn't clear. I'm aware that Mongoose treats `undefined` as unknown unknown, and this PR respects that (there are tests checking that, see `test/docs/defaults.js:29`).

The issue this PR fixes is that Mongoose does _not_ set `undefined` to the default value if it has been passed in a nested document, i.e:
```javascript
const Test = model('Test', { nested: { prop: { type: String, default: 'some default' } } })

await Test.create({ nested: { prop: undefined } })
// Should generate { nested: { prop: 'some default' } }
// but currently generates { nested: { prop: undefined } }
// which is persisted to mongodb as {}
```

AbdelrahmanHafez:
Oh, okay.
Thank you 👍

PR #9053, opened 2020-05-25
enhancement(connection): make server selection error less deterministic

### LINKED ISSUE #9006: Misleading error message when connecting to MongoDB Atlas with wrong password
When you connect to MongoDB Atlas from your application with wrong password using Mongoose, you get error message:

> Could not connect to any servers in your MongoDB Atlas cluster. Make sure your current IP address is on your Atlas cluster's IP whitelist: https://docs.atlas.mongodb.com/security-whitelist

Even though you have whitelisted your ip.

This is not a MongoDB Atlas error it is mongoose that assumes connection error to be due to ip address not being whitelisted.

You can reproduce the error if you put wrong password in dburl:

```
const dburl = 'mongodb-url-with-username-and-password';

mongoose.connect(dburl, { useNewUrlParser: true, useUnifiedTopology: true }).catch(err => {
	console.log(err);
});
```

This is the line of code that throws the error as mentioned by Jesse from MongoDB team:

[https://github.com/Automattic/mongoose/blob/54dd5cbf9a69a6079f1d88afceb480cf6fa87476/lib/error/serverSelection.js#L41](https://github.com/Automattic/mongoose/blob/54dd5cbf9a69a6079f1d88afceb480cf6fa87476/lib/error/serverSelection.js#L41)

Node js version: v13.6.0
Mongoose version: 5.9.14
MongoDB version: v4.2.6

*this issue is a bug*

AbdelrahmanHafez:
fixes #9006

The current error message could give the impression that this **is** the reason the connection failed, not just a common reason. According to https://github.com/Automattic/mongoose/issues/9006#issue-618882548 there are other possible reasons (such as incorrect credentials), this PR makes it so that the message is less deterministic.

### DIFF
--- a/lib/error/serverSelection.js
+++ b/lib/error/serverSelection.js
@@ -38,9 +38,10 @@ MongooseServerSelectionError.prototype.constructor = MongooseError;
  * ignore
  */
 
-const atlasMessage = 'Could not connect to any servers in your MongoDB Atlas ' +
-  'cluster. Make sure your current IP address is on your Atlas cluster\'s IP ' +
-  'whitelist: https://docs.atlas.mongodb.com/security-whitelist/.';
+const atlasMessage = 'Could not connect to any servers in your MongoDB Atlas cluster. ' +
+  'One common reason is that you\'re trying to access the database from ' +
+  'an IP that isn\'t whitelisted. Make sure your current IP address is on your Atlas ' +
+  'cluster\'s IP whitelist: https://docs.atlas.mongodb.com/security-whitelist/';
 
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez:
Happy to help 👍

osher:
I got same problem, and spent a lot of time on it.

The root cause was a misconfigured password, so I'm afraid is an as prominent reason as the IP whitelisting.

IMHO - what can be done:
 -   use an even  more generic language, such as `connection could not be established with the provided mongoUrl`
    OR
 -  be more specific and include a failed authentication in that description.
  OR
 - Ideally - distinct between server selection and authentication, if that's whatsoever possible

PR #9059, opened 2020-05-27
Clarify parameters passed to .debug. Fix #9029.

### LINKED ISSUE #9029: Please document all parameters to .debug
Maybe I haven't found the right documentation page, but the best information I could get about `.debug` was [here](https://mongoosejs.com/docs/api.html#mongoose_Mongoose-set):

> `mongoose.set('debug', function(collectionName, methodName, arg1, arg2...) {}); // use custom function to log collection methods + arguments`

What are the other args?

It's odd that [this SO answer](https://stackoverflow.com/questions/18762264/log-all-queries-that-mongoose-fire-in-the-application/18763020#18763020) has more information than the official docs :) 

What I'm particularly interested in is a way to describe the operation that Mongoose sent to MongoDB at a higher level, e.g. see `populate` in the debugging info.

dandv:


dandv:
I've also noticed that in the debug callback, `this` is set to the NativeCollection. That was useful to me in order to log Mongoose debugging output to both a file and the console, using $print and $printoToStream.

Should `this` be documented?

vkarpov15:
@dandv yes please, if you can add a PR that would be much appreciated.

PR #9075, opened 2020-05-30
Update faq.pug

tigransimonyan:
before
```
Model.init().then(function() {
  assert.ifError(err);
```
after
```
Model.init().then(function(err) {
  assert.ifError(err);
```

### DIFF
--- a/docs/faq.pug
+++ b/docs/faq.pug
@@ -112,8 +112,7 @@ block content
     // Promise based alternative. `init()` returns a promise that resolves
     // when the indexes have finished building successfully. The `init()`
     // function is idempotent, so don't worry about triggering an index rebuild.
-    Model.init().then(function(err) {
-      assert.ifError(err);
+    Model.init().then(function() {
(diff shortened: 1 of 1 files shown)


docs/faq.pug
```diff
@@ -112,7 +112,7 @@ block content
     // Promise based alternative. `init()` returns a promise that resolves
     // when the indexes have finished building successfully. The `init()`
     // function is idempotent, so don't worry about triggering an index rebuild.
-    Model.init().then(function() {
+    Model.init().then(function(err) {
       assert.ifError(err);
```
AbdelrahmanHafez, line 116: Welcome @tigransimonyan 

The function passed to `.then(...)` is only invoked in case of success, so in this case, we should remove `assert.ifError(err);` and the parameter you just added.

tigransimonyan: Done! :D

AbdelrahmanHafez: Thanks 👍

PR #9099, opened 2020-06-05
fix(document): allow using $locals for virtual setters

### LINKED ISSUE #9098: Virtual setters can't utilize `doc.$locals` on document initialization
Bug.

Virtual setters cannot use `doc.$locals` on `new MyModel({ myVirtual: 'someValue' })` since `doc.$locals` doesn't get initialized until after the setters have been called.

https://gist.github.com/ryankeener/4949e0c90eb73d56f6cad9a822d635e4

Virtual setters should be able to utilize `$locals` on `new`.

mongoose: `5.9.10`
node: `v12.16.2`

I tried moving https://github.com/Automattic/mongoose/blob/master/lib/document.js#L161 to the top of the constructor and it resolved my issue but I'm unsure of whether that's the best solution.

The example gist is simplified to show the effect but the 'why' of this is that I have a document on which I'd like to set fields `key` & `version` indirectly via a virtual, i.e. `doc.myVirtual = 'someKey@1'`. Later (validation/pre-save) I do an async call against another collection to find the latest `@1` version (e.g. `1.1.0`) and then set `doc.version= '1.1.0'; doc.key='someKey'`.

AbdelrahmanHafez:
fixes #9098

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -85,6 +85,7 @@ function Document(obj, fields, skipId, options) {
   this.isNew = 'isNew' in options ? options.isNew : true;
   this.errors = undefined;
   this.$__.$options = options || {};
+  this.$locals = {};
 
   if (obj != null && typeof obj !== 'object') {
     throw new ObjectParameterError(obj, 'obj', 'Document');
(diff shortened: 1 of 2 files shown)


vkarpov15:
Woops sorry, I missed this PR and fixed this in f9ddbc02f163b0d6b9d85b8f149ccd97e81ac222 . I'll close this PR but I'll give it a shout out in the changelog.

AbdelrahmanHafez:
It's alright, the important thing is that it got fixed.
Thanks for the shoutout. 👍

Issue #9055, opened 2020-05-25
Investigate async stack trace support


vkarpov15:
Re: #9009. In Mongoose 5, we eagerly create `Error` instances for certain errors, like server selection errors, in order to ensure the user gets a stack trace that points to a line in their code rather than node internals. However, this pattern has significant performance implications because creating error instances takes a disproportionate amount of CPU time.

If we use `async function`s all the way through, we should be able to ensure readable stack traces with [async stack traces](https://thecodebarbarian.com/async-stack-traces-in-node-js-12).

daukadolt:
Hi! Is this something community a newbie can contribute to?

If there is a way to contribute, I'd be thrilled to!

vkarpov15:
@daukadolt you're most welcome to contribute. If you can get even one Mongoose function to work with async stack traces all the way through, that would be very helpful. Just make sure you base any pull requests for this issue against the `6.0` branch.

daukadolt:
@vkarpov15 in the PR above you've mentioned bluebird.

Is supporting bluebird required for compatibility with pre-es2015 javascript?

vkarpov15:
@daukadolt for better or for worse, a lot of people still use bluebird. We don't support pre-ES6 JavaScript (Mongoose 5 requires node 4 or higher), it's more about not wanting to break the guarantee that Mongoose always returns a Bluebird promise if you say `mongoose.Promise = require('bluebird')` given that a fair number of people use that.

vkarpov15:
Going to close this re: https://github.com/Automattic/mongoose/pull/9100#issuecomment-819608011 . Something to consider in the future when there's less friction on async stack traces.

PR #9100, opened 2020-06-05
[WIP] proof of concept for async stack traces with `Query#exec()`

vkarpov15:
Proof of concept for async stack trace integration. The key insight is that you don't need async functions all the way through the stack, you just need to intercept the error in the top-level async function and create the stack trace.

For example, the below script:

```javascript
'use strict';
  
const { Schema } = mongoose;

async function run() {
    useNewUrlParser: true,
    useUnifiedTopology: true
  });

  const schema = new Schema({ name: { type: String, unique: true } });

  //const q = Model.findOne({ name: 'foo' }).orFail();
  //await q.exec();

  await Model.init();
  await Model.create({ name: 'foo' });
  await Model.findOneAndUpdate({ _id: new mongoose.Types.ObjectId() }, { name: 'foo' }, { upsert: true }).exec();
}
```

Before this change, the stack trace doesn't include the actual calling code:

```
MongoError: E11000 duplicate key error collection: test.tests index: name_1 dup key: { name: "foo" }
    at MessageStream.messageHandler (/mongoose/node_modules/mongodb/lib/cmap/connection.js:261:20)
    at MessageStream.emit (events.js:315:20)
    at processIncomingData (/mongoose/node_modules/mongodb/lib/cmap/message_stream.js:144:12)
    at MessageStream._write (/mongoose/node_modules/mongodb/lib/cmap/message_stream.js:42:5)
    at writeOrBuffer (_stream_writable.js:352:12)
    at MessageStream.Writable.write (_stream_writable.js:303:10)
    at Socket.ondata (_stream_readable.js:712:22)
    at Socket.emit (events.js:315:20)
    at addChunk (_stream_readable.js:302:12)
    at readableAddChunk (_stream_readable.js:278:9)
    at Socket.Readable.push (_stream_readable.js:217:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:186:23) {
  ok: 0,
  errmsg: 'E11000 duplicate key error collection: test.tests index: name_1 dup key: { name: "foo" }',
  code: 11000,
  codeName: 'DuplicateKey',
  keyPattern: { name: 1 },
  keyValue: { name: 'foo' },
  [Symbol(mongoErrorContextSymbol)]: {}
}
```

After with Node v12 and v14:

```
MongoError: E11000 duplicate key error collection: test.tests index: name_1 dup key: { name: "foo" }
    at model.Query.exec (/mongoose/lib/query.js:4446:7)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async run (/gh-9055.js:25:3) {
  name: 'MongoError',
  ok: 0,
  errmsg: 'E11000 duplicate key error collection: test.tests index: name_1 dup key: { name: "foo" }',
  code: 11000,
  codeName: 'DuplicateKey',
  keyPattern: { name: 1 },
  keyValue: { name: 'foo' },
  [Symbol(mongoErrorContextSymbol)]: {}
}
```

Note that the original point where `await Model.findOneAndUpdate(...).exec()` was called is in the stack trace :+1:

This change has several nuances though:

1) We need a way to reliably create a new error with a different stack trace for Mongoose errors and mongo errors. That's a bit tricky since Mongoose error constructors all have slightly different syntaxes. This PR hacks around it with an `_args` property on errors, but I don't want to release this if that is the solution.
2) `mongoose.Promise` breaks. Async functions always return native promises, so there's no way to make an async `exec()` return a bluebird promise.


vkarpov15:
Re: #9055

vkarpov15:
That's fair. I'm still thinking we might want to offer some option for backwards compatibility in case people are tightly tied to bluebird, but I'm comfortable making an async function that returns a native promise the default.

AbdelrahmanHafez:
Perhaps something like this would work.
```js
Query.prototype.exec = function exec () {
  if (this.options.useMongoosePromise) {
    // do things the old way
    return;
  }

  return (async function exec () {

  });
};
```

vkarpov15:
@AbdelrahmanHafez unfortunately that wouldn't help, because the instant you leave an async function you lose async stack traces. So if `exec()` isn't async, you lose async stack traces.

vkarpov15:
Upon further review, I'm going to close this branch. Async stack traces aren't standardized, so this behavior is subject to change and it's hard to justify such a major change for non-standard behavior. We may consider making a plugin to do this in the future.

PR #9127, opened 2020-06-13
Performance Bottleneck with creation of ServerSelectionError and CastError

Uzlopak:
**Summary**

While running performance tests, I realized, that mongoose creates alot of unnecessary ServerSelectionErrors and CastErrors. My Flamegraphs get hot.

**Examples**

before:
Bildschirmfoto von 2020-06-13 03-40-38

after fixing ServerSelectionError

Bildschirmfoto von 2020-06-13 03-40-55

after fixing CastError
Bildschirmfoto von 2020-06-13 03-41-03

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -4892,11 +4892,11 @@ Model.$handleCallbackError = function(callback) {
  */
 
 Model.$wrapCallback = function(callback) {
-  const serverSelectionError = new ServerSelectionError();
   const _this = this;
 
   return function(err) {
     if (err != null && err.name === 'MongoServerSelectionError') {
+      const serverSelectionError = new ServerSelectionError();
       arguments[0] = serverSelectionError.assimilateError(err);
     }
     if (err != null && err.name === 'MongoNetworkError' && err.message.endsWith('timed out')) {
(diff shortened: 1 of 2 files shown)


Uzlopak:
Can somebody explain to me, why the tests are failing?

Uzlopak:
@vkarpov15 How can these changes break the unit tests?

Uzlopak:
I reverted it to only cover the ServerSelectionError, for now. The unit test seems to be randomly failing?

vkarpov15:
@Uzlopak unfortunately we create `CastError` and `ServerSelectionError` eagerly in order to provide better stack traces, see #9100 and #9009. If you create the error in a callback, you lose the original stack trace of where the user called Mongoose.

We're working on making Mongoose integrate with [async stack traces](https://thecodebarbarian.com/async-stack-traces-in-node-js-12) for Mongoose 6. If this is causing performance issues for you, perhaps we can add an option to opt out of eager error instantiation?

Sorry about the test failures, we've been experiencing significantly elevated test flakiness over the last couple weeks, haven't fixed it yet.

Uzlopak:
@vkarpov15 
I made more tests and realized, that in node10 the flamegraph gets for these errors hot, but in node12 we have not this issue. So for me this is not an issue anymore so I close this PR as it is a desired behaviour by design

:)

PR #9132, opened 2020-06-14
fix(model): allow empty arrays for bulkWrite

### LINKED ISSUE #9131: Model.bulkWrite(...) throws an error with an empty arrays
Currently `Model.bulkWrite([ ]);` throws an error, this can happen due to a dynamically created array of operations that can have length of 0.

I'll put in a PR that returns a default bulkwrite result in this case instead of going all the way to Mongo and getting an error.

AbdelrahmanHafez:
fixes #9131


AbdelrahmanHafez:
Happy to help 👍

PR #9137, opened 2020-06-15
docs: specify the array field syntax for invalidate

### LINKED ISSUE #9133: Document the notation for invalidating nested array fields
At https://mongoosejs.com/docs/api/document.html#document_Document-invalidate, the `path` parameter is documented as,

> path «String» the field to invalidate

What is the format for nested array fields? `array.2.field`, `array[2].field`?

dandv:
This PR fixes #9133.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2561,7 +2561,7 @@ Document.prototype.validateSync = function(pathsToValidate, options) {
  *               value: 14 } } }
  *     })
  *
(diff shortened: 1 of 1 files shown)


dandv:
> A minor improvement would be to use `i` instead of `X`, since it's the standards for indexes.

Good point, fixed!

AbdelrahmanHafez:
Thanks 👍

PR #9144, opened 2020-06-18
docs(validation): add validateBeforeSave and invalidate

dandv:
Add [validateBeforeSave](https://mongoosejs.com/docs/guide.html#validateBeforeSave) and `invalidate()` to the [Validation guide](https://mongoosejs.com/docs/validation.html).

Is there a way to mark a field as valid, so it will be skipped by the pre-save validation?


vkarpov15:
Re: "Is there a way to mark a field as valid, so it will be skipped by the pre-save validation?", [`Document#$markValid()`](https://mongoosejs.com/docs/api/document.html#document_Document-$markValid) is the inverse of `invalidate()`.

PR #9176, opened 2020-06-25
fix(document): disallow `transform` functions that return promises

### LINKED ISSUE #9163: Throw error when `transform` returns a promise
Am I doing something wrong defining a transformer on a schema? Here's my schema code:

```js
const bookSchema = new Schema({
  title: {
    type: String,
    transform: async function(v) {
      return await Promise.resolve(v)
    }
  },
});
```

When I create a new `Book`, I get `title: {}`. Do transformers have to be synchronous?

AbdelrahmanHafez:
fixes #9163


AbdelrahmanHafez:
That is clearer, thank you.

PR #9185, opened 2020-06-29
Update schema.js

joaoritter:
**Summary**

I've been trying to debug the following problem: when a custom `typeKey` and `typePojoToMixed: false` are passed as options to a schema, schema properties under the new typeKey that are a POJO get set as Mixed in the schema. This is incorrect, the type should be dynamically converted to a subschema, as per the line I have changed in this PR.
My change preserves the `typeKey` during the Object.assign, instead of using the default key of `type`.

**Examples**

```
const schema = mongoose.Schema({ a: { $type: { b: String, c: String } }}, { typeKey: '$type', typePojoToMixed: false });

// current buggy behavior: 
console.log({ schema }) // Schema { ..., paths: {  'a': [Mixed] } }   

// expected behavior: 
console.log({ schema }) // Schema { ..., paths: {  'a.b': [SchemaString], 'a.c': [SchemaString] } }   
```

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -509,7 +509,7 @@ Schema.prototype.add = function add(obj, prefix) {
         // Propage `typePojoToMixed` to implicitly created schemas
         const opts = { typePojoToMixed: false };
         const _schema = new Schema(obj[key][this.options.typeKey], opts);
-        const schemaWrappedPath = Object.assign({}, obj[key], { type: _schema });
+        const schemaWrappedPath = Object.assign({}, obj[key], { [this.options.typeKey]: _schema });
         this.path(prefix + key, schemaWrappedPath);
       } else {
         // Either the type is non-POJO or we interpret it as Mixed anyway


joaoritter:
I also noticed there are no test cases for setting a typeKey to a POJO. It's probably worth adding one if you intend to support it, or add to documentation if not.

joaoritter:
This is blocking me right now, I'd appreciate a quick acknowledgement of this PR. I understand if it takes a bit to merge, but I just want to make sure im not missing something simple. I could attempt other work arounds if this is in fact a bug and it'll take a while to merge a fix.

joaoritter:
I dug up this commit where this line was last changed. It would be useful to add a test with a typeKey here. 
https://github.com/Automattic/mongoose/commit/b4b61d312b57af710df33853311d12fb00075f08

Thanks!

joaoritter:
100%, thank YOU for all your work on this project.

PR #9189, opened 2020-07-02
add mongoose default option to set `returnOriginal`

### LINKED ISSUE #9183: global option to set `new` on all findAndUpdate operations
`new` option on findOneAndUpdate is a common gotcha to people using mongoose, I personally have been confused by it before.

I believe most people would prefer to have `findOneAndUpdate` return the new document by default, we should provide a global option that allows people to control the default behavior.

AbdelrahmanHafez:
fixes #9183

Also made a typo fix in the docs for `findOneAndReplace`


lib/index.js
```diff
@@ -147,6 +147,7 @@ Mongoose.prototype.driver = require('./driver');
  *
  * Currently supported options are:
  * - 'debug': If `true`, prints the operations mongoose sends to MongoDB to the console. If a writable stream is passed, it will log to that stream, without colorization. If a callback function is passed, it will receive the collection name, the method name, then all arugments passed to the method. For example, if you wanted to replicate the default logging, you could output from the callback `Mongoose: ${collectionName}.${methodName}(${methodArgs.join(', ')})`.
+ * - 'defaultNewOnFindAndUpdateOrReplace': If `true`, changes the default `new` option to `findOneAndUpdate()`, `findByIdAndUpdate` and `findOneAndReplace()` to true.
```
vkarpov15, line 150: I like this change, but I think the name is a bit too verbose and isn't consistent with other naming conventions: we typically give global, schema-level, and connection-level options the same name as their child options.

Admittedly, `mongoose.set('new', true)` suffers from the opposite problem: `new` is too generic to be a global option name.

Maybe we instead name the global option `returnOriginal`? We already support a `returnOriginal` option, so perhaps to opt in to always getting the equivalent of `new: true` we ask users to do `mongoose.set('returnOriginal', false)`?

AbdelrahmanHafez: I agree, I started with `defaultNewOnFindAndUpdate` but I was concerned that users would expect it to not work on `findOneAndReplace`, so I added the last bit, which became too verbose. I justified that by the fact that this is written only once in a project, so I leaned toward the more verbose name.

However, `returnOriginal` has an advantage, it is clear enough to deliver the meaning and generic enough that we can apply the option to more operations in the future, add the changes in the documentation, and users can then look at it later. We're not limiting ourselves by the option name.

lib/query.js
```diff
@@ -3004,20 +3004,25 @@ Query.prototype.findOneAndUpdate = function(criteria, doc, options, callback) {
...
+  }
+  if (options.fields) {
+    this.select(options.fields);
+    delete options.fields;
+  }
+
+
+  const returnOriginal = get(this, 'model.base.options.returnOriginal');
+  if (options.new == null && returnOriginal != null) {
+    options.new = !returnOriginal;
```
vkarpov15, line 3021: I'd recommend setting `options.returnOriginal = returnOriginal` for consistency's sake. [We convert `new` to `returnOriginal` under the hood anyway](https://github.com/Automattic/mongoose/blob/f168f17526fc2e468313ca2d857e5452f6c05031/lib/query.js#L3404-L3407).

test/model.test.js
```diff
@@ -6868,4 +6868,57 @@ describe('Model', function() {
...
+
+        const createdUser = yield User.create({ name: 'Hafez' });
+
+        const user1 = yield User.findOneAndUpdate({ _id: createdUser._id }, { name: 'Hafez1' }, { new: false });
+        assert.equal(user1.name, 'Hafez');
+
+        const user2 = yield User.findByIdAndUpdate(createdUser._id, { name: 'Hafez2' }, { new: false });
+        assert.equal(user2.name, 'Hafez1');
+
+        const user3 = yield User.findOneAndReplace({ _id: createdUser._id }, { name: 'Hafez3' }, { new: false });
```
vkarpov15, line 6919: I'd also add tests to make sure you can pass in `returnOriginal`, like `yield User.findOneAndUpdate({ _id: createdUser._id }, { name: 'Hafez1' }, { returnOriginal: true })`

lib/query.js
AbdelrahmanHafez: Sounds reasonable, I'll take a look at those changes in the next weekend, we can move this to 5.10.1 if it will delay things.

PR #9204, opened 2020-07-05
docs(guide): refactor transactions examples to async/await

AbdelrahmanHafez:


docs/transactions.pug
```diff
@@ -46,7 +46,36 @@ block content
...
+
+    await Customer.createCollection();
+    const session = await mongoose.startSession();
+
+    // Start a transaction
+    session.startTransaction();
+
+    // This `create()` is part of the transaction because of the `session`
+    // option.
+    await Customer.create([{ name: 'Test' }], { session: session });
```
js-kyle, line 59: I wonder if async await examples should use `exec()` for better stack traces?

docs/transactions.pug
```diff
@@ -46,7 +46,36 @@ block content
     operation in a transaction, you need to pass the `session` as an option.
 
     ```javascript
-    [require:transactions.*basic example]
```
vkarpov15, line 49: I like that these are now using async/await, but I don't like that this gets rid of our automated tests. I think it would be better to just convert `transactions.test.js` to use async/await, and then move it behind `test/es-next` like what we did for the `findOneAndUpdate()` tutorial:

* https://github.com/Automattic/mongoose/blob/0c7eab161c2746df64786238194f32ccd0aede25/test/es-next/findoneandupdate.test.es6.js
* https://github.com/Automattic/mongoose/blob/0c7eab161c2746df64786238194f32ccd0aede25/test/es-next.test.js

docs/transactions.pug
vkarpov15: I agree that it's better to use [`Query#exec()`](https://mongoosejs.com/docs/api/query.html#query_Query-exec) in examples. Although for this it isn't necessary: `create()` returns a promise, and doesn't have an `exec()` function.

vkarpov15:
I made this change in the `transactions.test.js` file so I'm closing this one :+1:

Issue #9146, opened 2020-06-18
null should not be an allowed value on Primitive types.


sgpinkus:
**Do you want to request a *feature* or report a *bug*?**
It's either a bug or a request for a option to allow us to not allow null values for primitives or at least cast them since `null` is not a legal Boolean, Number, or String. null's type is actually object in Node:

```
> typeof null === 'object'
true
```

[ECMAScript spec](https://www.ecma-international.org/ecma-262/6.0/index.html#sec-null-value) actually says its type is `Null` but states `null` is a:

> primitive value that represents the intentional absence of any *object* value

not  the intentional absence of a primitive or any other type. It's like a null pointer in other languages. Allowing null in a property explicitly declared to to have a type (other than Object or Mixed) sort of degrades the meaningfulness of that declaration. Means more work for client to account for edge cases.

Secondly `null` actually has reasonable casts for primitives, so it confusing to not cast null while other mismatched types that have reasonable casts will be cast by Mongoose:

```
> Boolean(null)
false
> Number(null)
0
> String(null)
'null'
```
---

I realize you can use `required: true` to force non null, but 1. You have to make sure you do this for every primitive type which is a lot of work. 2. It's not exactly equivalent to not allowing null because you can not allow null while still allowing the value to be undefined.

**What is the current behavior?**
You can set property declared to be of type Boolean, Number String to null:

```
var mongoose = require('mongoose');

async function main() {
  await mongoose.connect('mongodb://localhost:27019/test', {});
  await mongoose.connection.dropDatabase();
  var TestSchema = new mongoose.Schema({
    a: { type: Boolean, default: false },
    b: { type: Number, default: 0 },
    c: { type: String, default: '' },
  });
  var Test = mongoose.model('Test', TestSchema);
  var myTest = await new Test({ a: null, b: null, c: null });
  await myTest.save();
  console.log(myTest)
}
main()
  .then(() => process.exit())
  .catch((e) => { console.error(e); process.exit(); })
```
Gives:
```
{ a: null, b: null, c: null, _id: 5eeaef8d8885f90cf6088e5f, __v: 0 }
```

**What is the expected behavior?**
See top.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

```
> process.version
'v12.18.0'
> mongoose.version
'5.9.17'
```

vkarpov15:
[...]

PR #9223, opened 2020-07-09
Allow null in Boolean.convertTo(True|False) functions (#9146).

sgpinkus:

**Summary**
Allow Boolean.convertToFalse Set to take null. See [comment](https://github.com/Automattic/mongoose/issues/9146#issuecomment-656084916) on #9146 for motivation.

**Examples**

```
const mongoose = require('/home/sam/src/mongoose/index.js');
const { Schema } = mongoose;
mongoose.Schema.Types.Boolean.convertToFalse.add(null);
console.log(mongoose.Schema.Types.Boolean.convertToFalse);
const M = mongoose.model('Test', new Schema({ b: Boolean }));
console.log(new M({ b: null })); // false
```

### DIFF
--- a/lib/schema/boolean.js
+++ b/lib/schema/boolean.js
@@ -224,6 +224,8 @@ SchemaBoolean.prototype.castForQuery = function($conditional, val) {
   return this._castForQuery($conditional);
 };
 
+SchemaBoolean.prototype.castNull = true;
+
 /*!
  * Module exports.
  */
--- a/lib/cast/boolean.js
+++ b/lib/cast/boolean.js
@@ -14,15 +14,14 @@ const CastError = require('../error/cast');
  */
 
 module.exports = function castBoolean(value, path) {
-  if (value == null) {
-    return value;
-  }
-
   if (module.exports.convertToTrue.has(value)) {
     return true;
   }
   if (module.exports.convertToFalse.has(value)) {
     return false;
+  }
+  if (value == null) {
+    return value;
   }
   throw new CastError('boolean', value, path);
 };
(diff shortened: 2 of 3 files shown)


lib/schema/boolean.js
```diff
@@ -224,6 +224,8 @@ SchemaBoolean.prototype.castForQuery = function($conditional, val) {
   return this._castForQuery($conditional);
 };
 
+SchemaBoolean.prototype.castNull = true;
```
vkarpov15, line 227: Is this change necessary given your other changes? Your code example doesn't make any reference to this new `castNull` option.

AbdelrahmanHafez: It's being used in [this line](https://github.com/Automattic/mongoose/blob/18296bb5d6fdc9975e2825dd4fee995fe340c516/lib/schematype.js#L1058).

sgpinkus: I just used it because it's more extensible than say type checking for Boolean in the line @AbdelrahmanHafez referred to. If in future, other types want to cast null, they'd set `castNull = true` to get passed that conditional check, although currently other type cast implementations (like ex. [Number](https://github.com/Automattic/mongoose/blob/43288f869366ade4c287b4dd06e9c7016c6104c1/lib/cast/number.js#L17)) don't do anything with null even if castNull is true. Boolean is a bit of a special case because there is an existing documented mechanism (convertToTrue|False) exposed to easily customize null casting.

AbdelrahmanHafez: If this is approved, it would be great if you add this option to the documentation using JSDoc.

sgpinkus:
> I made an alternative fix in [415ba3d](https://github.com/Automattic/mongoose/commit/415ba3de0486727209f6c2ffc9d6edd360235fc2) . Very similar, just no need for the extra `castNull` option. If you add `null` to `convertToFalse`, that should be all you have to do.

Awesome thanks! Yeah, castNull wasn't ideal.

PR #9249, opened 2020-07-16
fix call save and then modify, save failed again

wlx199x:
document call save after and markModified , markModified possible by previous save call reset, so again save failed

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -315,6 +315,7 @@ Model.prototype.$__handleSave = function(options, callback) {
         ret.$where = where;
         callback(null, ret);
       });
+      this.$__reset();
     } else {
       const optionsWithCustomValues = Object.assign({}, options, saveOptions);
       this.constructor.exists(this.$__where(), optionsWithCustomValues)
@@ -344,12 +345,7 @@ Model.prototype.$__save = function(options, callback) {
         callback(error, this);
       });
     }
-
-    // store the modified paths before the document is reset
-    const modifiedPaths = this.modifiedPaths();
-
-    this.$__reset();
-
+    
     let numAffected = 0;
     if (get(options, 'safe.w') !== 0 && get(options, 'w') !== 0) {
       // Skip checking if write succeeded if writeConcern is set to
@@ -377,7 +373,7 @@ Model.prototype.$__save = function(options, callback) {
         if (numAffected <= 0) {
           // the update failed. pass an error back
           const err = this.$__.$versionError ||
-            new VersionError(this, version, modifiedPaths);
+            new VersionError(this, version, this.$__.$modifiedPaths);
           return callback(err);
         }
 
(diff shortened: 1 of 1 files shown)


wlx199x:
> Can you add test cases demonstrating the issue?

it('reset ""modified', function(done){
      const UniqueSchema = new Schema({
        changer: String,
        unique: {
          type: Number,
          unique: true
        }
      });

      const Unique = db.model('Test', UniqueSchema);
      const u1 = new Unique({
        changer: 'a',
        unique: 5
      });

      Unique.on('index', function(){
        u1.save(function(err){
          assert.ifError(err);
          assert.ok(!u1.isModified('changer'));
          u1.changer = 'b';
          u1.save();
          u1.changer = 'c';
          setTimeout(() => {
            assert.ok(!u1.isModified('changer'));
            Unique.collection.drop(done);
          }, 10);
        })
      })
    })

wlx199x:
> Can you add test cases demonstrating the issue?

it('reset ""modified', function(done){
  const UniqueSchema = new Schema({changer: String});
  const Unique = db.model('Test', UniqueSchema);
  const u1 = new Unique({changer: 'a'});

  u1.save(function(err){
    u1.changer = 'b';
    u1.save(); // this save reset next line modified
    u1.changer = 'c'; // this again modified
    setTimeout(() => {
      assert.ok(u1.isModified('changer')); //but this failing
      Unique.collection.drop(done);
    }, 10);
  })
})

vkarpov15:
The below script works fine, and I can't merge this PR because it breaks tests. Please modify the below script to demonstrate your issue:

```javascript
'use strict';
  
const { Schema } = mongoose;

async function run() {
    useNewUrlParser: true,
    useUnifiedTopology: true
  });

  const UniqueSchema = new Schema({changer: String});
  const Unique = mongoose.model('Test', UniqueSchema);
  const u1 = new Unique({changer: 'a'});

  await u1.save();

  u1.changer = 'b';
  assert.ok(u1.isModified('changer'));

  await u1.save();

  u1.changer = 'c';
  assert.ok(u1.isModified);

  console.log('Done');
  process.exit(0);
}

```

AbdelrahmanHafez:
A minor typo fix to the test case @vkarpov15 posted, which still passes the test:
```
  u1.changer = 'c';
  assert.ok(u1.isModified()); // function call was missing
```

wlx199x:
You are right.
My usage scenario is that I sometimes lack await to find the problem，however, I think it makes more sense to call this.$__reset before the save returns，should not be waiting for the updateOne callback. @vkarpov15

vkarpov15:
@wlx199x not necessarily. We wait until after `save()` is done to call `$__reset()` because the `save()` may fail (for example, because of lost connectivity to the database).

wlx199x:
Ok, I see. Thank you very much. @vkarpov15

PR #9299, opened 2020-07-31
Upgrade to mongodb driver 3.6

vkarpov15:


package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "bson": "^1.1.4",
```
billouboq, line 22: I think they now use bson 4.0.4, maybe this needs to be updated too

vkarpov15: Nope, they still use 1.1.x for 3.6. It would be a breaking change if they used 4.x: https://github.com/mongodb/node-mongodb-native/blob/d84d14b2c61e639b4e773c6b32ac509e9da53420/package.json#L28

Kamikadze4GAME:
Good morning, would this upgrade cover new features of MongoDB 4.4 (new aggregate stages and others)?

package.json
billouboq: Oh my bad 🤗 
Thanks for the explanation

Issue #8810, opened 2020-04-17
Make mongoose connections not thenable, instead add `Connection#asPromise()` function


vkarpov15:
#8706 and other issues have expressed some concern that `createConnection()` returns a connection with `then()` and `catch()` methods, but then connections delete `then()` and `catch()` once connection is established.

This behavior is an ugly hack to make it so that `createConnection()` returns a connection and also a thenable that resolves to a connection - it is impossible to make a promise resolve to itself. Basically we want to support these 2 cases:

1. Synchronously create a connection and use it immediately: `const conn = mongoose.createConnection(uri); conn.model('Test', schema);`
2. Asynchronously create a connection and use it once connection has been established: `const conn = await mongoose.createConnection(uri); conn.model('Test', schema);`

Supporting both of these cases is surprisingly messy. I'd rather err on the side of supporting (1), because the synchronous case is more limiting, whereas when you have `await` refactoring async logic is easier.

With the proposed change, (1) would stay the same, (2) would become `await mongoose.createConnection(uri).asPromise();`.

@AbdelrahmanHafez @lineus comments most welcome :+1:

vkarpov15:
Related: #5404, https://github.com/Automattic/mongoose/commit/343b7a79651b4fe75e5e9b03d92659a76de323b3

PR #9358, opened 2020-08-25
fix(connection): openUri return promise-like on existing connection

### LINKED ISSUE #9335: TypeError: conn.openUri(...).then is not a function
bug

I'm using the following to connect to MongoDB via Mongoose:
```
 conn = Mongoose.connect(conn_str, {
      useNewUrlParser: true,
      useUnifiedTopology: true,
      family: 4,
    }).catch((error) => {
      throw error
    })
```
Once I run the script with around 20-100 connections, I start getting the following error:
```
TypeError: conn.openUri(...).then is not a function
at Mongoose.connect (/src/node_modules/mongoose/lib/index.js:339:47)
```
I can't figure out what is causing this.

Look at the above. I can't replicate this reliably because it occurs only when there is a lot of connections and data flying.

There wouldn't be an error.

Node 12.16.3, Mongoose 5.10, Mongo 3.6.9.

DiegoRBaquero:
**Summary**
Fixes #9335

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -669,11 +669,11 @@ Connection.prototype.openUri = function(uri, options, callback) {
     if (typeof callback === 'function') {
       callback(null, this);
     }
-    
+
     this.then = (resolve) => {
-      resolve(this)
+      resolve(this);
     };
-    this.catch = () => {}
+    this.catch = () => {};
     return this;
   }
 
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -671,6 +671,8 @@ Connection.prototype.openUri = function(uri, options, callback) {
     }
 
     this.then = (resolve) => {
+      delete this.then;
+      delete this.catch;
       resolve(this);
       return this;
     };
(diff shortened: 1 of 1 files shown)


lib/connection.js
```diff
@@ -669,6 +669,12 @@ Connection.prototype.openUri = function(uri, options, callback) {
     if (typeof callback === 'function') {
       callback(null, this);
     }
+
+    this.then = (resolve) => {
```
clhuang, line 673:
```suggestion
    this.then = (resolve) => {
      delete this.then;
      delete this.catch;
      resolve(this);
      return this
    }
```
if you don't do this, if you try to call `Promise.resolve(conn)` or `await conn` your interpreter will go in an infinite loop of calling `resolve` recursively as it tries to flatten the promise, see https://github.com/Automattic/mongoose/issues/8810

DiegoRBaquero: makes, sense, fixed!

clhuang: I guess if you really wanted to you could do
```js
this.then = (resolve) => {
  process.nextTick(() => {
    delete this.then;
    delete this.catch;
    resolve(this);
  });
  return this;
}
```

PR #9414, opened 2020-09-12
Replace var with const in docs and test files

jmadankumar:
**Summary**

Documentation and test cases are using the old **var** keyword for variable declaration. This makes the document look outdated. This PR fixes the variable declaration using const and let in the documentation and test cases.


vkarpov15 (approved):
Thanks!

PR #9425, opened 2020-09-17
`mongoose.model() --> mongoose.model()

tphobe9312:
MD link format fixed

**Summary**

MD url format is fixed in this PR

**Examples**

The problem : MD code was being rendered as is : [`mongoose.model()](/docs/api/mongoose.html#mongoose_Mongoose-model).


AbdelrahmanHafez (approved):

vkarpov15 (approved):
Thanks :+1:

PR #9457, opened 2020-10-01
fix(transaction): fix saving new documents w/ arrays in transactions

PenguinToast:
**Summary**

Fix a bug with transaction tracking when saving a new document that has array fields:
```
     TypeError: Cannot convert undefined or null to object
      at Function.assign (<anonymous>)
      at mergeAtomics (lib/plugins/trackTransaction.js:87:31)
      at _getAtomics (lib/plugins/trackTransaction.js:66:31)
      at model.<anonymous> (lib/plugins/trackTransaction.js:26:30)
```

**Examples**

See added test for minimal reproduction.


vkarpov15 (approved):
Thanks!

PR #9488, opened 2020-10-18
Update connections.pug

tareqdayya:

**Summary**

Make it clear that mongoose.connection.on('error') does not fire once connection is lost any time after the initial connection. The text seems to suggest, to me at least, that it's the code to use to listen to drops in connection after the initial connection.
Only 3 Lines were added: 148-150.

**Examples**

### DIFF
--- a/docs/connections.pug
+++ b/docs/connections.pug
@@ -145,6 +145,10 @@ block content
     });
     ```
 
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #9497, opened 2020-10-23
fix(connection): make connection.then(...) resolve to a connection instance

### LINKED ISSUE #9496: 5.10.0 -> 5.10.1 Introduces a breaking change (still in 5.10.9)

**What is the current behavior?** in 5.10.0 mongoose.connection is resolved to instance (?) but in 5.10.1 it resolves to undefined. (using .then())

I was using [node rate limiter flexible](https://github.com/animir/node-rate-limiter-flexible/wiki/Mongo) which then after running a yarn upgrade and reruninng project I saw errors. after some time of debugging I pinpointed the issue with downgrading the mongoose version from latest to where rate limiter was still working.

**What is the expected behavior?** connection.then should resolve to the instance like it was doing so in it's previous versions ?

It's still happening in 5.10.9

AbdelrahmanHafez:
fixes #9496

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -811,7 +811,7 @@ Connection.prototype.openUri = function(uri, options, callback) {
       throw err;
     });
   this.then = function(resolve, reject) {
-    return this.$initialConnection.then(resolve, reject);
+    return this.$initialConnection.then(() => resolve(_this), reject);
   };
   this.catch = function(reject) {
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

PR #9499, opened 2020-10-23
Update Options for Model.findOneAndUpdate

radamson:
**Summary**

The current documentation for [`Model.findOneAndUpdate`](https://mongoosejs.com/docs/api/model.html#model_Model.findOneAndUpdate) inconsistently document the `overwrite` and `upsert` options between the narrative and tagged parts of the source documentation.  The [`Query.prototype.setOptions`](https://mongoosejs.com/docs/api.html#query_Query-setOptions) does not list the `overwrite`.  From what I could tell it looks like `overwrite` is valid for all write operations.

I also found I had to run `npm install` before running the the documentation make commands, so I through that in the instructions to make it obvious.

_Sidenote:_  I noticed some of the other options are inconsistently documented between the tagged and narrative sections of the documentation. Is there a good place to look in the source code to identify if/where they are applicable so the documentation for those can be updated accordingly?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,6 +51,7 @@ To contribute to the [guide](http://mongoosejs.com/docs/guide.html) or [quick st
 
 If you'd like to preview your documentation changes, first commit your changes to your local master branch, then execute:
 
+* `npm install`
 * `make docclean`
 * `make gendocs`
 * `node static.js`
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks! We don't have a good centralized list of all options for `update()`, because some of those options go to the mongodb server, so the available options depend on mongodb server version. But if you find options that aren't documented, please feel free to add a PR.

PR #9507, opened 2020-10-28
 fix(connection): avoid executing promise handler unless it's a function

### LINKED ISSUE #9505: TypeError: resolve is not a function
In v5.10.10 I get the following error:
_TypeError: resolve is not a function at
node_modules/mongoose/lib/connection.js:814:47
    at processTicksAndRejections (internal/process/task_queues.js:97:5)_

Works fine with 5.10.9 and lower

My code is:
````js

const DB = mongoose.createConnection(process.env.DB, {
  useNewUrlParser: true,
  useCreateIndex: true,
  useFindAndModify: false,
  useUnifiedTopology: true,
});
````

### PR DESCRIPTION
**Pull request by AbdelrahmanHafez**
fixes #9505

This PR makes `connection.then` consistent with native promises in the sense that they don't throw execute `.then(...)` arguments unless they were functions.
```js
const p = new Promise((resolve, reject) => {
  setTimeout(() => {
    // only executes `resolve` if it's a function
    resolve('Hello');
  }, 10);
});

// does not throw
p.then(null);
```

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -814,7 +814,11 @@ Connection.prototype.openUri = function(uri, options, callback) {
       throw err;
     });
   this.then = function(resolve, reject) {
-    return this.$initialConnection.then(() => resolve(_this), reject);
+    return this.$initialConnection.then(() => {
+      if (typeof resolve === 'function') {
+        resolve(_this);
+      }
+    }, reject);
   };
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

PR #9527, opened 2020-11-07
Remove redundant code.

ksullivan:
**Summary**

Query.prototype.mongooseOptions is set to a function and then immediately overwritten with another function. This change removes the first redundant assignment.


vkarpov15 (approved):
Thanks :+1:

PR #9530, opened 2020-11-08
Update middlewar.pug file

mustafaKamal-fe:
**Summary**

The documentation does not specificly addresses that both properties 'query' and 'document' must be assigned in order for the 'remove' middleware to execute for the right context (that is on the query or the document).

**Examples**
`   	
			
	  const schema = new mongoose.Schema({...});
		
	//middleware on document remove only
	//like in doc.remove()	
      schema.pre("remove", {query: false, document: true});
    
    //middleware on query remove only 
	//like in Model.remove()
      schema.post("remove", {query: true, document: false});
`

### DIFF
--- a/docs/middleware.pug
+++ b/docs/middleware.pug
@@ -333,7 +333,7 @@ block content
 
     ```javascript
     // Only document middleware
-    schema.pre('remove', { document: true, query: false}, function() {
+    schema.pre('remove', { document: true, query: false }, function() {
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

mustafaKamal-fe:
> Thanks 👍

i have just realized that you are the mind behind mongoose. thank you so much for it 🤘🚀

Issue #9535, opened 2020-11-10
Question: Mongoose enums in schema, using Typescript enums?


ajmas:
I originally a[sked this question ](https://stackoverflow.com/questions/64759521/typescript-enums-assigned-to-a-mongoose-enum)on Stack Overflow, but hoping this will be a better place for answers:

We have a Typescript based NodeJs project that makes use of Mongoose. We are trying to find an appropriate way to define an enum field on a Mongoose schema, based on a Typescript enum. I did look through the docs, but I could not find anything of note.

Taking an example enum:

```ts
enum ETransactionType {
    Buy = 'buy',
    Sell = 'sell',
}
```

and a Mongoose schema:

```ts
const promotionSchema = new Schema({
   user: { type: Schema.Types.ObjectId, ref: 'User' },
   product: { type: Schema.Types.ObjectId, ref: 'Product' },
   transactionType: {
      type: ETransactionType,
      default: ETransactionType.Buy
   }
});
```

Is this sufficient or should we also be extracting the values from the TS enum, such that:

```ts
transactionType: {
   type: ETransactionType,
   default: ETransactionType.Buy,
   enum: Object.values(ETransactionType)
}
```

AbdelrahmanHafez:
The first option will not have an `enum` validator run in runtime, it's useful only for development-time.

Also, it's worth noting that mongoose currently does not support TypeScript, [we're planning to have official TypeScript definitions in version 5.11.](https://github.com/Automattic/mongoose/issues/8108)

vkarpov15:
Here's how you should handle enums:

```typescript
const promotionSchema = new Schema({
   user: { type: Schema.Types.ObjectId, ref: 'User' },
   product: { type: Schema.Types.ObjectId, ref: 'Product' },
   transactionType: {
      type: String,
      default: ETransactionType.Buy,
      enum: Object.values(ETransactionType)
   }
});
```

Mongoose doesn't have built-in support for TypeScript enums, and it really can't because TypeScript enums are just [POJOs](https://masteringjs.io/tutorials/fundamentals/pojo) at runtime. But one thing we should do is eliminate the need for the `Object.values()` part.

DubonYaar:
That should do the trick

`Object.keys(authTypes).splice(Object.keys(authTypes).length / 2)`

AbdelrahmanHafez:
@DubonYaar 
[...]

Issue #3044, opened 2015-06-02
null in enum


windkomo:
Is there a reason why null is not treated the same as undefined in enums ?
Shouldn't setting required to false (or omitting the attribute) allow null value ?

I've seen in tests that null is added to enums.

vkarpov15:
Because null is not the same thing as undefined :) The high level semantics are that undefined means a value hasn't been set, whereas null means the value has been explicitly set to a value that ["represents the intentional absence of any object value"](http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.11).

tusbar:
@vkarpov15 though the required validator checks for both `undefined` and `null`.

I find it very counterintuitive to have to specify `null` in an enum’s values to allow `null`s while you could specify `required: true` to forbid them.

vkarpov15:
I disagree. Absence of a value and value never defined makes sense for required IMO, whereas enum generally means either the value was never defined or the value is one of the allowed values. Arguing about what's more intuitive is typically not very constructive though - I'm down to change if its a real pain point for a lot of users, but the difference is a one-liner.

tusbar:
Fair enough. :)

qwemaze:
@vkarpov15 

> [undefined value](http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.9) -- primitive value used when a variable has not been assigned a value
> [null value](http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.11) -- primitive value that represents the intentional absence of any object value

They both are *primitive values* and they both tell that there's no actual value, the only difference is that *null* is intentional. That's why *required* does not check against them.

 
> whereas enum generally means either the value was never defined or the value is one of the allowed values

So basically enum verifies the value against a set of values, right?

And why should it try to verify *null* if *null* is actually a no-value?

Enum should not try to verify either null or undefined because they are both, in the first place, are no-values.

varunjayaraman:
i think the point is that `null` is a value while `undefined` isn't. Specifically, the value that `null` holds is the absence of any value. You're intentionally setting something to what you refer to as a `no-value`

qwemaze:
They both are primitive values.

adamreisnz:
I find that I am also running into this.
[...]

PR #9547, opened 2020-11-14
feat(schema): allow using a POJO for `enum` to support TypeScript enums

### LINKED ISSUE #9546: Support passing a TypeScript enum to `enum` validator in schema
Right now you need to use `Object.values()` to get TypeScript enums to work with the `enum` validator:

```typescript
const promotionSchema = new Schema({
   user: { type: Schema.Types.ObjectId, ref: 'User' },
   product: { type: Schema.Types.ObjectId, ref: 'Product' },
   transactionType: {
      type: String,
      default: ETransactionType.Buy,
      enum: Object.values(ETransactionType)
   }
});
```

Would be nice if Mongoose handled the `Object.values()` part automatically and treated `{ foo: 'FOO', bar: 'BAR' }` as equivalent to `['FOO', 'BAR']` for enums.

AbdelrahmanHafez:
fixes #9546 #9535

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -245,7 +245,7 @@ SchemaArray.prototype.enum = function() {
 
   let enumArray = arguments;
   if (!Array.isArray(arguments) && utils.isObject(arguments)) {
-    enumArray = Object.values(enumArray);
+    enumArray = utils.object.vals(enumArray);
   }
 
   arr.caster.enum.apply(arr.caster, enumArray);
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -242,7 +242,13 @@ SchemaArray.prototype.enum = function() {
     }
     break;
   }
-  arr.caster.enum.apply(arr.caster, arguments);
+
+  let enumArray = arguments;
+  if (!Array.isArray(arguments) && utils.isObject(arguments)) {
+    enumArray = Object.values(enumArray);
+  }
+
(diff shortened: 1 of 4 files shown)


vkarpov15 (approved):
Great work, we'll release this with v5.11 :+1:

AbdelrahmanHafez:
Thank you, that works. 👍

joergbaier:
@AbdelrahmanHafez @vkarpov15 My team began updating our `enum` values to typescript enums. However there is not a way to allow `null` in this case. This is somewhat related to the discussion here https://github.com/Automattic/mongoose/issues/3044

```typescript
enum Color {
  Red = "red",
  Blue = "blue",
  ...
}

const SomeSchema = new Schema({
  color: { type: String, enum: Color },
  ...
});
```

As far as I can tell, there is no way to allow the color field to be nullable with this syntax. Typescript doesn't allow null in the enum and there is not an option to allow it in mongoose. We would need to use the previous syntax in this case (converting the enum to an array and concatenating null).

It looks like typegoose added a `addNullToEnum` option to handle this.
https://github.com/typegoose/typegoose/issues/235

Any thoughts on exposing a nullable option? I imagine that allowing null with `required: false` would not make sense.

vkarpov15:
@joergbaier the syntax you put would allow `color` to be `null`. The issue you posted says the following:

"Mongoose requires to add null to the enum array to allow an empty value to be used, even when required is false."

That statement is incorrect. You don't need to add `null` to the enum.

Issue #9571, opened 2020-11-20
Make `deepEqual()` treat objects as different if key order changed


vkarpov15:
https://github.com/Automattic/mongoose/blob/3a867422ac365c149f346024ff9103335eb96d7f/lib/utils.js#L133-L134 means that two objects are considered equal if key order is different. This has some implications because, if key order changes in the document, that's a change we need to persist.

PR #9572, opened 2020-11-21
fix(utils.deepEqual): make `deepEqual` treat objects with different order of keys as different

AbdelrahmanHafez:
re #9571


vkarpov15 (approved):
Nice, thanks :+1: My benchmarks show that this doesn't have a significant impact on performance, but it does make `deepEqual()` more correct as far as MongoDB is concerned.

AbdelrahmanHafez:
Happy to help 👍

PR #9587, opened 2020-11-26
fix(document): allow setting fields after an undefined field

### LINKED ISSUE #9585: 5.10.16 is badly broken!!!! It should be rescinded/removed!
We create discriminator-based documents that are highly nested, and we often create plain js objects to use in the model constructors to populate these structures in a new document when the document is first instantiated.

Something changed in 5.10.16 so that many of the fields and structures in the plain JS object we pass to the document constructor no longer end up in the document itself!

5.10.15 works fine.....so something new in .16 is the culprit.

Our code is too complex to create a test scenario easily. Sorry about that.

AbdelrahmanHafez:
fixes #9585

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -973,24 +973,26 @@ Document.prototype.$set = function $set(path, val, type, options) {
         options = Object.assign({}, options, { _skipMinimizeTopLevel: true });
       }
 
-      if (typeof path[key] === 'object' &&
-          !utils.isNativeObject(path[key]) &&
-          !utils.isMongooseType(path[key]) &&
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks and sorry for this bug!

AbdelrahmanHafez:
Happy to help!

maximilianschmid:
Problems of 5.10.16 resolved in 5.10.17. Our test suite now passes again. Great work, thank you.

AbdelrahmanHafez:
Glad to know, @maximilianschmid.
Happy to help.

PR #9611, opened 2020-12-01
Fix TypeScript intellisense

alecgibson:
This change aims to fix intellisense for developing type definitions in
mongoose.

In `.ts` files, we currently get a compilation warning from TypeScript
when trying to `import mongoose from 'mongoose'`:

```
Cannot find module 'mongoose' or its corresponding type declarations.ts(2307)
```

As such, none of the intellisense in VS Code works correctly.

This change adds `mongoose` as its own dependency so we can more easily
develop TypeScript definitions.

**Summary**

**Examples**


alecgibson:
The alternative to this approach is to replace all `.ts` instances of `import {} from 'mongoose'` with a relative dependency: `import {} from '../'` or similar

AbdelrahmanHafez:
Welcome @alecgibson 

I'd advise against using the `"dependency": "file:./"` fix, as we're suffering from its ramifications in our project as we used it as an alternative to [link-module-alias](https://www.npmjs.com/package/link-module-alias) to replace `require('../../../../helpers/numbers.js');` with `require('dependency/helpers/numbers.js');`.

The reason I would advise against it is the issues that arise when we try to upgrade any dependency, running `npm install` would result in an error with failing installation.
```
npm WARN tar ENOENT: no such file or directory, lstat 'E:\Info\speero\node_modules\.staging\mongoose-cd43e65c\lib\helpers\topology'
```

We've tried to resolve the issue and came to the conclusion that the only way around it is to get rid of the `file:./` and use the native node `exports` option. The only issue is that TypeScript is still yet to support it which means we wouldn't have Intellisense support.
Follow the following issue for progress https://github.com/microsoft/TypeScript/issues/33079.

That being said, I am not familiar enough with TypeScript, and that issue may be caused due to something I am not aware of in our projects. Are there any alternatives that do not include the `file:./` fix?

vkarpov15:
Thanks @AbdelrahmanHafez for your insights. I had similar concerns about this change - too risky because it affects non-typescript installs, although I’ve never tried this pattern before.

We should look into alternatives. I recall looking into this but I don’t think @types/mongoose supports this import syntax.

As a workaround, you can just do import {}, mongoose from ‘mongoose’

alecgibson:
Okay, to be fair it did feel a bit hacky when I did it. Just pushed a more semantic fix, which is to include a `tsconfig.json`, which IDEs should pick up (although I've only tested in VS Code).

vkarpov15 (commented):
Would putting a tsconfig.json in a test directory help? I don't think this will affect end users...

alecgibson:
Sorry maybe I was unclear; this change is to fix developing type definitions in `mongoose` itself. I quite pointedly _don't_ want clients to pick up `tsconfig.json`. Does that make sense?

vkarpov15:
That works 👍

vkarpov15 (approved):

Issue #9610, opened 2020-12-01
TypeScript: `createConnection` missing non-promise overload


alecgibson:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Can't return a non-Promise `Connection` from `mongoose.createConnection(url)`;

**If the current behavior is a bug, please provide the steps to reproduce.**

```ts
const connection: Connection = mongoose.createConnection(url);
```

This fails compilation:

```
TS2740: Type 'Promise<Connection>' is missing the following properties from type 'Connection': close, collection, collections, config, and 40 more.

22 const connection: Connection = mongoose.createConnection('');
         ~~~~~~~~~~
```

**What is the expected behavior?**

Being able to return a non-Promise `Connection` object.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

 - Node.js v10.23.0
 - Mongoose v5.11.0
 - MongoDB v3.6.3
 - TypeScript v4.1.2

alecgibson:
We're just missing overloads in the type definition. For some reason you can get a non-`Promise` if you don't provide any config, but if you provide config, you _must_ have a `Promise`: https://github.com/Automattic/mongoose/blob/07b141234827df4ca46085b047126ba2465ac6c4/index.d.ts#L58-L61

PR #9612, opened 2020-12-01
TS: Return `Connection` from `createConnection`

alecgibson:
Fixes https://github.com/Automattic/mongoose/issues/9610

This change fixes the TypeScript type definitions so that we return a
full `Connection` object from `createConnection(url)`, not just a
`Promise<Connection>`.

**Summary**

**Examples**


vkarpov15 (approved):
Thanks :+1:

PR #9621, opened 2020-12-02
fix(index.d.ts): allow `useCreateIndex` in connection options

alecgibson:
Updated the TypeScript example to mirror the example given in the
[deprecation warning docs][1].

[1]: https://mongoosejs.com/docs/deprecations.html


vkarpov15 (approved):
Thanks, sorry for the inconvenience. We will ship a release today with these changes :+1:

PR #9636, opened 2020-12-03
enhancement(schemaType): pass document to default functions

### LINKED ISSUE #9633: enhancement: pass document to default functions
Currently when a path has `default` set to a function, we can access the document via `this`, would be nice if the document gets passed as an argument to the function, similar to what we've done to `required` validator in https://github.com/Automattic/mongoose/issues/9433
```js
let documentIsThis;
let documentFromDefault;
  name: { type: String },
  age: {
    type: Number,
    default (doc) {
      // should evaluate to true so people can use arrow functions
      documentIsThis = doc === this;
     documentFromDefault = doc;
    }
  }

});

new User({ name: 'Hafez' });

assert.equal(documentIsThis, true);
assert.equal(documentFromDefault.name, 'Hafez');
```

AbdelrahmanHafez:
fixes #9633

### DIFF
--- a/lib/schematype.js
+++ b/lib/schematype.js
@@ -1028,7 +1028,7 @@ SchemaType.prototype.ref = function(ref) {
 
 SchemaType.prototype.getDefault = function(scope, init) {
   let ret = typeof this.defaultValue === 'function'
-    ? this.defaultValue.call(scope)
+    ? this.defaultValue.call(scope, scope)
     : this.defaultValue;
 
   if (ret !== null && ret !== undefined) {
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -9767,4 +9767,25 @@ describe('document', function() {
     const person = new Person({ items: undefined, email: 'test@gmail.com' });
     assert.equal(person.email, 'test@gmail.com');
   });
+
+  it('passes document to `default` functions (gh-9633)', function() {
+    let documentFromDefault;
+    const userSchema = new Schema({
+      name: { type: String },
+      age: {
+        type: Number,
+        default: function(doc) {
+          documentFromDefault = doc;
+        }
+      }
+
+    });
+
+    const User = db.model('User', userSchema);
+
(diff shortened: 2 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

vkarpov15:
@AbdelrahmanHafez I'm going to have to revert this. There are a couple cases in our tests where we use functions that rely on arguments as defaults. For example:

```javascript
const schema = new Schema({
  myId: {
    type: 'ObjectId',
    default: mongoose.Types.ObjectId
  }
}); 
```

This change will mean Mongoose calls `mongoose.Types.ObjectId` with `doc` as the first argument. So this change is unfortunately a little too risky for a patch release. We can make this change for v6.0.

AbdelrahmanHafez:
Ah yes, I didn't think of that.

I actually think now we're better off leaving things as is to avoid incorrectly passing the document as an argument to functions that do not necessarily expect the document.

For example, in our orders models we save `VATPercentage` which expects some parameters.

```js
VATPercentage: {
  type: Number,
  default: VATHelper.getVatPercentage
}
```

This would break if we pass any arguments to it since it expects the first argument to be a date.

```js
// vat.helper.js

const firstOfJuly2020 = new Date('2020-07-01');

module.exports = {
  // passing doc here would likely break the function
  getVatPercentage(date = new Date()) {
    if (date >= firstOfJuly2020) {
       return 0.15;
    }

    return 0.1;
 }
}
```

AbdelrahmanHafez:
@vkarpov15 Do you think we should add this as a breaking change in v6.0 or leave things as is? I can see arguments on both sides, what do you think?

vkarpov15:
@AbdelrahmanHafez I'm leaning towards putting this in 6.0. Not 100% convinced, but I think consistency with other cases where we pass `doc` as the first parameter and allowing arrow functions for defaults is more important here. We can add workarounds for some common cases, like `Date.now()` or `ObjectId()` to minimize breakage.

PR #9638, opened 2020-12-03
added missing custom error messages

SaifAlsabe:
Example

```

    name: {
        type: String,
        required: [true, "please enter a name"],
        minlength: [3, "Minimum name length is 3"],
    }
});
```

now we can add custom error messages


vkarpov15 (approved):
Thanks :+1:

Issue #9646, opened 2020-12-04
BUG: Ambiguous argument matching for bulkWrite's overload functions in TypeScript


davecroman:
Bug Report

Performing a bulkWrite causes an overload match to fail when expecting a Promise to be returned.

**Steps to reproduce**

1. Call `bulkWrite` with a single argument (an array of write operations)
2. Handle the resolved Promise by chaining with a `then`

```
MyModel.bulkWrite(writes)
      .then((results) => {
          // handle results here
      })
```

**What is the expected behavior?**
Compilation should succeed and the Promise should be resolved if successful

**Actual behavior**
An error is thrown during compilation
```
TSError: ⨯ Unable to compile TypeScript:
src/route/api/**.ts:120:8 - error TS2339: Property 'then' does not exist on type 'void'.

120       .then((results) => {
           ~~~~
```

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Node.js: 14.15.1
Mongoose: 5.11.3
MondoDB: (irrelevant)

**Perhaps noteworthy**
I've had a look at the TypeScript file and my guess is that the overload functions aren't matching properly:

```
bulkWrite(writes: Array<any>, options?: mongodb.CollectionBulkWriteOptions, cb?: (err: any, res: mongodb.BulkWriteOpResultObject) => void): void;
bulkWrite(writes: Array<any>, options?: mongodb.CollectionBulkWriteOptions): Promise<mongodb.BulkWriteOpResultObject>;
```

In the above example, the code matches both functions, but because the compiler uses the first match, it assumes that the function returns a void. Potential solutions:

1. Swap the order of the overloaded functions so that the first match is the function that returns a Promise, given no callback is provided.
2. Remove the argument match ambiguity by declaring `cb` as a required parameter for the first overload function and set `options` as the third argument but still optional.

I also see this issue in the following overloaded functions (this may not be an all-encompassing list):

1. `eachAsync`
2. `delete`
3. `deleteOne`
4. `remove`

vkarpov15:
Fixed by #9647

PR #9647, opened 2020-12-04
fix(index.d.ts): order when cb is optional in method

CatsMiaow:
As-Is
image

Even when there is no `cb`, the `cb` type is designated according to the type order.

To-be
image


vkarpov15 (approved):
Thanks :+1: This also fixes #9646

PR #9652, opened 2020-12-04
fix(index.d.ts): add `Document#__v` so documents have a Version by default

sahasayan:
fix(index.d.ts): add `Document#__v` so documents have a Version by default

**Summary**

__v  was missing in Document.

**Examples**

See more https://mongoosejs.com/docs/guide.html#versionKey


vkarpov15 (approved):
Thanks :+1:

PR #9655, opened 2020-12-04
fix(index.d.ts): add the missing generic declaration for Schema

sahasayan:
**Summary**

add the missing generic declaration for Schema

**Examples**

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mongoose/index.d.ts#L886


vkarpov15:
Ugh I hate doing this because there’s no reason that I’m aware of for Schema to be generic, but we can do this for backwards compat

PR #9665, opened 2020-12-06
fix(index.ts): allow the next() argument to be optional

### LINKED ISSUE #9664: Typings: optional argument for next()
bug

`next()` call requires an err argument.

``` javascript
UserSchema.pre<IUser>('save', function (next) {
    if (this.isModified('password')) {
        this.password = bcrypt.hashSync(this.password, 10);
    }

    const isSomethingWrong = false;

    if (isSomethingWrong) {
        next(new Error('something is wrong'));
    }

    next(); // An argument for 'err' was not provided.
});
```

`next()` call must be allowed without an err param.

Node.js 14.15.0
Mongoose 5.11.4
MongoDB 4.2.11

sahasayan:
**Summary**
allow the next() argument to be optional

fixes #9664


vkarpov15 (approved):
Thanks :+1:

PR #9680, opened 2020-12-08
fix(index.d.ts): Fix return type of Model#aggregate()

orgads:
`Type 'Promise<unknown>' is not assignable to type 'Promise<Document>'.`


vkarpov15 (approved):
Thanks :+1:

PR #9682, opened 2020-12-08
fix(index.d.ts): Refine default return type of aggregate

orgads:
`any` was copied from @types/mongoose, but it looks like `T` is the right choice here.


vkarpov15 (commented):
I disagree with this change, `aggregate()` can change the shape of result documents, so assuming that an the result of an aggregation pipeline is an array of documents by default seems a bit off. Why do you think this is the right choice?

orgads:
You're right. Thanks for the review.

PR #9685, opened 2020-12-08
fix(index.d.ts): allow 2 generic types in mongoose.model function

### LINKED ISSUE #9678: How to define types of statics with mongoose 5.11 ?

bug

I cannot get type definitions for statics on Models in.

```
error TS2339: Property 'findOrCreateProject' does not exist on type 'Model<ProjectDoc>'.

93     const project = await Project.findOrCreateProject(projectName);

error TS2558: Expected 1 type arguments, but got 2.

42 export const Project = mongoose.model<ProjectDoc, ProjectModelDoc>("Project", ProjectSchema);
                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is what I did previously (with 5.10 and @types/mongoose), but it does not work anymore. Model does not have two type arguments anymore:

```js
import mongoose, { Schema, Document, Model } from "mongoose";

export interface ProjectDoc extends Document {
  name: string;
}

interface ProjectModelDoc extends Model<ProjectDoc> {
  findOrCreateProject(name: string): Promise<ProjectDoc>;
}

const ProjectSchema = new Schema(
  {
    n: { type: String, required: true, unique: true, alias: "name" },
  },
  { _id: true, timestamps: { createdAt: "ca", updatedAt: "ua" } }
);

ProjectSchema.statics.findOrCreateProject = async function (
  name: string
): Promise<ProjectDoc> {
  ...
};

export const Project = mongoose.model<ProjectDoc, ProjectModelDoc>("Project", ProjectSchema);
```

No error

mongoose: 5.11.5

sahasayan:
fixes #9678

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -99,6 +99,12 @@ declare module "mongoose" {
   export function isValidObjectId(v: any): boolean;
 
   export function model<T extends Document>(name: string, schema?: Schema, collection?: string, skipInit?: boolean): Model<T>;
+  export function model<T extends Document, U extends Model<T>>(
+    name: string,
+    schema?: Schema,
+    collection?: string,
+    skipInit?: boolean
+  ): U;
 
   /** Returns an array of model names created on this instance of Mongoose. */
   export function modelNames(): Array<string>;


icopp:
From my testing with something similar on an existing codebase, it looks like there may still be some issues here with methods like `create`, which will expect that the methods from the statics are included on the supplied object. Fixing it may take changing the `Model` definition to directly accommodate statics as a generic type param which `create` and the like can then exclude.

vkarpov15 (approved):
Thanks :+1:

PR #9686, opened 2020-12-08
Allow array of validators in SchemaTypeOptions

cjroebuck:
**Summary**

Allows schemas with an array of validators (`ValidateOpts`) to be passed, currently the typescript type only allows to pass a single validator.

**Examples**

https://github.com/Automattic/mongoose/blob/master/lib/schematype.js#L713-L717


vkarpov15 (approved):
Thanks :+1:

PR #9692, opened 2020-12-10
fix(index.d.ts): add missing Aggregate#skip() & Aggregate#limit()

### LINKED ISSUE #9691: Typescript: Properties 'limit' and 'skip' do not exist on type Aggregate
I'd like to report a bug

I updated from Mongoose 5.10.19 with @types/mongoose to 5.11.6 and got a new error from typescript: 
`error TS2339: Property 'limit' does not exist on type 'Aggregate<any[]>'`
on a line like this:
`const data = await Model.aggregate(aggregationPipeline).skip(skip).limit(limit);`

Compiling the following with typescript should reproduce the issue:
```

mongoose.connect('mongodb://localhost:27017/test', {
  useNewUrlParser: true,
  useCreateIndex: true,
  useUnifiedTopology: true,
});

const personSchema = new mongoose.Schema({name: String})
const personModel = mongoose.model('Person', personSchema);

const doSomething = async () => {
  const data = await personModel.aggregate().skip(10).limit(10);
  console.log(data);
}

doSomething();
```

The program should compile as before

NodeJs: 14.11.0
Mongoose: 5.11.6
MongoDB: 4.2.9

sahasayan:
fixes #9691

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2236,6 +2236,12 @@ declare module "mongoose" {
     /** Sets the hint option for the aggregation query (ignored for < 3.6.0) */
     hint(value: object | string): this;
 
+    /**
+     * Appends a new $limit operator to this aggregate pipeline.
+     * @param num maximum number of records to pass to the next stage
+     */
+    limit(num: number): this;
+                                          
     /** Appends new custom $lookup operator to this aggregate pipeline. */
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks!

PR #9698, opened 2020-12-12
fix(index.d.ts): use document _id type for findById methods

orblazer:
**Summary**

This PR apply strict type for `findById` methods for prevent mistakes and for only accept the right type of `_id`.

**Examples**

When we make an Document with custom id like :
```ts
interface Account extends Document {
  _id: number
}
```

The `findById` methods accept all entry like :
```ts
accountModel.finById('test')
accountModel.finById(ObjectId())
accountModel.finById(1)
```

And now only the last `findById` is valid (`accountModel.finById(1)`).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -680,7 +680,7 @@ declare module "mongoose" {
      * equivalent to `findOne({ _id: id })`. If you want to query by a document's
      * `_id`, use `findById()` instead of `findOne()`.
      */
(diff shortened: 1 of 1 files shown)


vkarpov15:
I don’t think this change is the right choice. Mongoose casting does let you pass any type, like strings when ObjectId is allowed. Plus this may be backwards breaking in this case and others. any is the right choice here. Thanks for the PR though.

PR #9699, opened 2020-12-12
fix(index.d.ts): allow specify type of _id

orblazer:
**Summary**

When we want make an complex type and change the type of `_id` the typescript say its is `any` if we don't make something like that :
```ts
type FixedDocument<T> = (T extends {_id?: any } ? Omit<Document, '_id'> : Document) & T
// OR
interface Account extends Omit<Document, '_id'> {
  _id: number
  username: string
}
```

**Examples**

Now this is more easier and we no longer need `Omit<Document, '_id'>`.
```ts
interface Account extends Document<number> {
  username: string
}

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -363,11 +363,11 @@ declare module "mongoose" {
     getIndexes(): any;
   }
 
-  class Document {
+  class Document<T = any> {
     constructor(doc?: any);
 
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks

PR #9700, opened 2020-12-12
fix(index.d.ts): make options optional for `toObject`

orblazer:
**Summary**

When we want use `toObject` on `Map` or `Array` types we need pass unnecessary options.

**Examples**

Before this we need to pass an empty object for call `toObject` like :
```ts
found.array.toObject({})
found.map.toObject({})
```

After this PR we can omit the options like :
```ts
found.array.toObject()
found.map.toObject()
```


vkarpov15 (approved):
Thanks

PR #9718, opened 2020-12-15
Update populate.pug with enable virtuals

getzelus:
I tried the example of the band with members. 
It worked only by adding the line :  
mongoose.set('toJSON', { virtuals: true });
on the configuration of mongoose (at the start of the server). 

I found that example thanks to that website :
https://masteringjs.io/tutorials/mongoose/virtuals

### DIFF
--- a/docs/populate.pug
+++ b/docs/populate.pug
@@ -601,6 +601,9 @@ block content
 
     <h3 id="populate-virtuals"><a href="#populate-virtuals">Populate Virtuals</a></h3>
 
+    You need to enable virtuals on mongoose first :
+    mongoose.set('toJSON', { virtuals: true });
+
     So far you've only populated based on the `_id` field. However, that's
     sometimes not the right choice.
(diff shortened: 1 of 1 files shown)


vkarpov15 (commented):
I'd disagree with this, it isn't strictly necessary to set `toJSON.virtuals` to use populate virtuals. I'd put this at the end of the section and link to this section in the virtuals docs: https://mongoosejs.com/docs/tutorials/virtuals.html#virtuals-in-json

vkarpov15:
I took a closer look and the `virtuals` option is already in this section of the docs:

```
Keep in mind that virtuals are _not_ included in `toJSON()` output by default. If you want populate virtuals to show up when using functions
    that rely on `JSON.stringify()`, like Express'
    [`res.json()` function](http://expressjs.com/en/4x/api.html#res.json),
    set the `virtuals: true` option on your schema's `toJSON` options.
```

PR #9727, opened 2020-12-20
feat(model): Model.bulkSave()

### LINKED ISSUE #9673: [Feature] Model.bulkSave(...)
One common use case in our applications is finding documents, doing some changes on them, then saving them.
```js
const users = await User.find({ cityId: 1 });
await Promise.all(users.map(async function (user){
  let cart = await  Cart.findOne({ userId: user._id });
  if (!cart) {
    cart = new Cart({ userId:user._id });
  }

  cart.itemsIds.addToSet(item._id);
  await cart.save();
}));
```

Looking at the code snippet above, we have around 600k users, running the code above could have the following for carts:
- 300K carts are unchanged, no sent operations to MongoDB, thanks to mongoose checking if the documents are unchanged.
- 150K carts are inserted (isNew === true).
- 150K carts are updated.

This means we send 300K separate operations, which can be very slow.

Would be great if we have a function that can take an array of documents, and `bulkSave` them in a single operation using `Model.bulkWrite(...)`.

It would work like this:
```js
const users = await User.find({ cityId: 1 });
const carts = await Promise.all(users.map(async function (user){
  let cart = await  Cart.findOne({ userId: user._id });
  if (!cart) {
    cart = new Cart({ userId:user._id });
  }

  cart.itemsIds.addToSet(item._id);
  return cart;
}));

// writeResult is a bulkWrite result
// all carts must be instanceof Document
const writeResult = await Cart.bulkSave(carts); // calls Cart.bulkWrite under the hood
```

What do you think? @vkarpov15 
I'm willing to add this feature if accepted.

AbdelrahmanHafez:
fixes #9673


lib/model.js
```diff
@@ -3499,6 +3499,46 @@ Model.bulkWrite = function(ops, options, callback) {
...
+
+    const changes = document.getChanges();
+    if (document.isNew) {
+      accumulator.push({
+        insertOne: { document }
+      });
+    } else if (!utils.isEmptyObject(changes)) {
+      accumulator.push({
+        updateOne: {
+          filter: { _id: document._id },
```
vkarpov15, line 3530: It would be better if this applied versioning and custom `where`. To do that, you'd need to do something like `document.$__version(_applyCustomWhere(document, document.$__where()), changes)`. Take a look at how `save()` does it.

AbdelrahmanHafez: Great, will do.

I've been thinking about how this is going to play with middlewares, and how it's going to reflect on documents in case of failure. I am starting to think that we should change that into a helper:
```js
const writeOperations = Model.buildBulkWriteFromDocuments(documents);
await Model.bulkWrite(writeOperations);
```

I still haven't looked in the error object that's returned from bulkWrite failures, but unless it has a reference for the `_id` of the document that failed, we won't be able to know which documents to avoid marking as not-modified.

vkarpov15: @AbdelrahmanHafez is this ready to merge on your end? LGTM to me and I'd like to ship 5.12.0 in the next couple of weeks.

AbdelrahmanHafez: @vkarpov15 
I've been giving this some thought, I'd like to expose two different functions:
A) `Model.buildWriteOperations(...)`: which takes an array of documents and returns an array of write operations, the implementation of this is quite simple.
B) `Model.bulkSave(...)`: This would utilize the first function and do a `bulkSave`, error-handling this one can have its nuances, we could fail mongoose-side validation errors early without sending any updates to the database, but for mongo-side errors, handling them can be a bit confusing to users.

How are we going to inform users as to which documents saves have succeeded and which ones failed?
How are we going to inform users of whether or not this is a mongo error or mongoose error?

I am starting to think that this will create more confusion than its worth, and for these reasons would rather just expose the first function as a nice utility that helps users bulkWrite and let them be responsible for how they use those operations.

I don't know if that's a thing in open-source, but I'm open to a pair-programming session where we brainstorm this. Let me know if that's something you'd be interested in.

I pushed the changes I have on mind.

As to the first utility, I've been running an experiment to compare the performance difference of using `updateMany` with the same `update` vs multiple `updateOne`s, and the results show around 2.5x performance improvement. The common update is a common use-case in our company (e.g. updating order `status` or some special identifier.). This may be a room for improvement in the first functionality. If you're interested in the details of the experiment, please check it out [here](https://stackoverflow.com/questions/65831219/mongodb-bulkwrite-multiple-updateone-vs-updatemany/65835186#65835186).

vkarpov15: I'd prefer to keep (A) internal, and expose (B). I think it's fine to just return a bulkWriteResult and let the client app figure out the bulkWriteResult as necessary. Like how `save()` does it: most of the time, users just look at the document to tell whether it was successfully saved or not.

The one big concern I have with this as implemented is `save()` hooks: right now `bulkSave()` doesn't trigger `save()` hooks, but perhaps it should. What do you think?

AbdelrahmanHafez:
@vkarpov15 This is ready from my end, let me know what you think.

lib/model.js
```diff
@@ -3522,6 +3522,115 @@ Model.bulkWrite = function(ops, options, callback) {
...
+    }
+
+    const changes = document.getChanges();
+    if (document.isNew) {
+      accumulator.push({
+        insertOne: { document }
+      });
+    } else if (!utils.isEmptyObject(changes)) {
+      accumulator.push({
+        updateOne: {
```
vkarpov15, line 3621: I'd still like for :point_down: to apply versioning and custom `$__where()` as suggested below.

lib/model.js
```diff
@@ -3522,6 +3522,115 @@ Model.bulkWrite = function(ops, options, callback) {
...
+  if (!Array.isArray(documents)) {
+    throw new Error(`bulkSave expects an array of documents to be passed, received \`${documents}\` instead`);
+  }
+
+
+  const writeOperations = documents.reduce((accumulator, document, i) => {
+    if (!(document instanceof Document)) {
+      throw new Error(`documents.${i} was not a mongoose document, documents must be an array of mongoose documents (instanceof mongoose.Document).`);
+    }
+    const validationError = document.validateSync();
```
vkarpov15, line 3609: This shouldn't be necessary because we run `validate()` as a pre save hook.

vkarpov15 (changes requested):
Couple of suggestions, but looks good overall :+1:

vkarpov15 (approved):
Nice, thanks :+1: I'll merge this into the 5.13 branch

Issue #8798, opened 2020-04-15
Allow `schema` as a schema path name


songguangyu:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When i set schema name `schema`, 
![image](https://user-images.githubusercontent.com/6869168/79321677-4a54e900-7f3e-11ea-8e59-85a2de3fca93.png)

mongoose throw err 
![image](https://user-images.githubusercontent.com/6869168/79321430-f0ecba00-7f3d-11ea-8206-f40725ef9f10.png)

`mongoose/lib/schema.js` line 528

![image](https://user-images.githubusercontent.com/6869168/79321525-17aaf080-7f3e-11ea-9e6c-a2d2c9d20961.png)

**If the current behavior is a bug, please provide the steps to reproduce.**

`
const schema = new Schema({
  'schema': { type: String, default: '{}'},
});
`
**What is the expected behavior?**
I can set name `schema` in Schema
**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Mongoose version 5.4.7
5.9.9 alse have this problem 
https://github.com/Automattic/mongoose/blob/master/lib/schema.js#L645
Nodejs version 12

AbdelrahmanHafez:
https://mongoosejs.com/docs/api.html#schema_Schema.reserved

songguangyu:
@AbdelrahmanHafez Thanks for you answer, I am working on service migration. If the schema field already exists in the original database, how should I deal with this situation?

AbdelrahmanHafez:
In that case I would rename the schema field into a different field name.
I've always found [thesaurus](https://www.thesaurus.com/browse/schema#) quite helpful in finding synonyms.

songguangyu:
@AbdelrahmanHafez  Maybe another mongo client is a better solution.

AbdelrahmanHafez:
I can't make that choice for you, depending on the effort of renaming the field, and the features mongoose provides. See what outweighs what, and make your decision.

Perhaps @vkarpov15 has a better idea.

songguangyu:
@AbdelrahmanHafez thanks!

vkarpov15:
Some path names unfortunately do conflict with existing methods and properties. However, there are ways for us to work around this. I'll keep this open for a future release.

JatinRanka:
@vkarpov15 Can I take up this issue to work on?

vkarpov15:
[...]

PR #9734, opened 2020-12-22
fix : Allow  schema as a schema path name #8798

Edu93Jer:

**Summary**

I am new at the open-source and I wanted to contribute to one of the libraries I use the most, look for a new good issue to solve and here I am. Mostly what it was about, was to allow a "schema" as a schema path name.

**Examples**

It fixs the issue #8798 actually I modify the test, at schema.test.js line 1441, so that way I make sure the solution was correct.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -101,17 +101,17 @@ function Document(obj, fields, skipId, options) {
     throw new ObjectParameterError(obj, 'obj', 'Document');
   }
 
-  const schema = this.schema;
+  const reserved_schema = this.schema;
 
   if (typeof fields === 'boolean' || fields === 'throw') {
     this.$__.strictMode = fields;
     fields = undefined;
   } else {
-    this.$__.strictMode = schema.options.strict;
+    this.$__.strictMode = reserved_schema.options.strict;
     this.$__.selected = fields;
   }
 
-  const requiredPaths = schema.requiredPaths(true);
+  const requiredPaths = reserved_schema.requiredPaths(true);
   for (const path of requiredPaths) {
     this.$__.activePaths.require(path);
   }
@@ -177,7 +177,7 @@ function Document(obj, fields, skipId, options) {
     const keys = Object.keys(this._doc);
 
     keys.forEach(function(key) {
-      if (!(key in schema.tree)) {
+      if (!(key in reserved_schema.tree)) {
         defineKey(key, null, _this);
       }
     });
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -589,7 +589,7 @@ reserved.isNew =
 reserved.populated =
 reserved.remove =
 reserved.save =
-reserved.schema =
+reserved.reserved_schema =
 reserved.toObject =
 reserved.validate = 1;
 
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -1438,9 +1438,9 @@ describe('schema', function() {
 
       assert.throws(function() {
         new Schema({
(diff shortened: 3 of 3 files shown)


Edu93Jer:
@vkarpov15 I made a PR for the issue #8798 , but I don´t understand why did the CI build failed, I looked at the details but I don't understand why it failed, looks like a connection, but I can´t understand at all, could you please help me understand better.

lib/document.js
```diff
@@ -101,17 +101,17 @@ function Document(obj, fields, skipId, options) {
     throw new ObjectParameterError(obj, 'obj', 'Document');
   }
 
-  const schema = this.schema;
+  const reserved_schema = this.schema;
```
vkarpov15, line 104: This unfortunately won't work in the case that `schema` is a path name. In that case, `this.schema` may be a string or something else, so this assignment won't work.

vkarpov15 (changes requested):
I may be missing something but I don't think this fixes the issue. Try adding a test case, I may be wrong.

Also, don't mind the CI failure, we have a couple of connection tests that are unreliable.

lib/document.js
Edu93Jer: Well I tried what you told me to change all the places in `lib/document.js` where we use `this.schema` there was actually 45 references but when you run the test it gives you 3 more errors.

mongoose_test

And actually, the first error that makes reference to the path name stays the same, is there a way to see what does the first test is expecting? `AssertionError [ERR_ASSERTION]: Missing expected exception.` 

Because I guess I renamed the schema property where it wasn't needed, and that brings me the 3 more errors, so when I try to compile the app, it says this...

mongoose_test1
mongoose_test2

I don't know if I'm explaining myself enough so you can understand me, sorry if I am making this more complicated. I have been trying many ways but can not find the correct one. Hope you can explain me a little more.

Cheers!

Edu93Jer: Do you recommend me to push the changes so you can review them? I am kind of lost now, I didn't push them because well they didn't pass the test.

vkarpov15: Given that there's 45 references, this change may be a little more complex than I'd imagined. It likely isn't a good first issue, there's a lot of small refactoring to be done. You'd need to change each of those 45 references. Is that still something you're interested in working on?

Edu93Jer: Yes, of course, I don't wanna give up, just if you could help me a little more understanding the code, is being kind of difficult for me to understand it. 
I already change the name on the 45 references, but it crashes with other documents as I can understand because it gives 3 more errors, which are the 2), 3), and 4) of the image above. It can not read the property emit and options from files `model.js` and `query.js`
And when I tried to test it with another project it gives the error that can not find the property [scopeSymbol] from undefined, so I am definitely changing the variable, where maybe it is not supposed to be changed.

vkarpov15: The error message in (2) is indicative of the fact that you changed `this.schema` in a static function on the `Model` class. Remember that `Model.init()` is a static function, not a method, so `this.schema` doesn't need to change here: https://github.com/Automattic/mongoose/blob/21f1f180e3ae23df28275e30e2522db913d553f2/lib/model.js#L1228 .

Remember that `Document` is a class, and `Model` inherits from `Document`. So `Model` has both a static property `schema` and an instance property `schema` from `Document`. The instance property needs to change, the static property doesn't.

vkarpov15:
I'm going to close this PR, it is long out of date and unlikely to end up in a mergeable state.

PR #9744, opened 2020-12-28
fix(index.d.ts): add missing overloaded function for Document#populate()

sahasayan:
**Summary**

add missing overloaded function for Document#populate()

**Examples**

```
await post.populate('author', 'id username').execPopulate();
```


vkarpov15 (approved):
Thanks :+1:

PR #9750, opened 2020-12-30
fix(index.d.ts): add missing function for Aggregate#group()

coro101:
**Summary**

It adds a missing group() function to the Aggregate class(index.d.ts).

**Examples**

It will make the following code works in Typescript:
```Typescript
const doc = await Document.aggregate().group({
  _id: '$groupId',
  total: { $sum: 1 }
});
```


vkarpov15 (approved):
Thanks :+1:

PR #9761, opened 2021-01-04
Make SchemaDefinition accept a model

mroohian:
Make SchemaDefinition accept a model for more stricter type checking and making the schema consistent with typescript models.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1161,10 +1161,11 @@ declare module 'mongoose' {
     virtualpath(name: string): VirtualType | null;
   }
 
-  interface SchemaDefinition {
-    // eslint-disable-next-line @typescript-eslint/ban-types
-    [path: string]: SchemaTypeOptions<any> | Function | string | Schema | Schema[] | Array<SchemaTypeOptions<any>> | Function[] | SchemaDefinition | SchemaDefinition[];
-  }
+  type SchemaDefinitionProperty<T = undefined> = SchemaTypeOptions<any> | Function | string | Schema | Schema[] | Array<SchemaTypeOptions<any>> | Function[] | SchemaDefinition<T> | SchemaDefinition<T>[];
+
+  type SchemaDefinition<T = undefined> = T extends undefined
+    ? { [path: string]: SchemaDefinitionProperty; }
+    : { [path in keyof T]-?: SchemaDefinitionProperty<T[path]>; };
 
   interface SchemaOptions {
     /**


vkarpov15 (commented):
Thanks for the PR, but I unfortunately don't understand what this change does. Can you please elaborate on what issues this is meant to catch with some code samples?

mroohian:
First of all, this change is backward compatible with how SchemaDefintion works right now if T is undefined, However it allows you to define a SchemaDefinition that complies with a typescript interface or rather a Model. If a field is not present in the model the schema would cause compile error. This is specially important for big project with huge models which the model properties can change over time causing regression errors. Specially if the fields are optional this would cause silent bugs. As an example:

```ts
import { Model, SchemaDefinition, Schema } from 'mongoose';

interface User {
  readonly email: string;
  readonly password: string;
  readonly someOptionalField?: string;
  readonly anotherProperty?: string;
}

const userSchemaDefinition: SchemaDefinition<User> = {
  password: { type: String, required: true },
  someOptionalField: String,

  nonExistingField: String, // 1. This causes an error

  // 2. anotherProperty is defined in the interface but not defined in the schema and triggers an error.
}

/*
 3. `SchemaDefinition` without any params results in `SchemaDefinition<undefined>` which disables these checks
  and is backward compatible
*/

const UserSchema = new mongoose.Schema(userSchemaDefinition);

interface UserDocument extends Document, User ...

```

vkarpov15:
This is a good idea. But how does it handle the case where `interface User extends Document`? That's a common pattern and this would still allow you to add properties that are in `Document` but not in `interface User`. Maybe you should do something like `[path in keyof LeanDocument<T>]-?`?

Also, what does the `-?` syntax do? I haven't seen that before.

mroohian:
The syntax `-?` makes all the optional fields mandatory and `+?` makes all the fields optional.
I am not quite sure about `LeanDocument` mostly because I haven't used it often. In any case, I would not use the same interface for documents and models. As an example:

```ts
interface User {
  // ...
}

interface UserDocument extends User, Document {
  // ...
}

const UserSchemaDefinition: SchemaDefinition<User> = {
  // ...
}

const UserSchema = new mongoose.Schema(UserSchemaDefinition);

interface UserModel extends mongoose.Model<UserDocument> {
  // ...
}

const userModel = mongoose.model<UserDocument, UserModel>('User', UserSchema);

```

BTW, the constructor for `Schema` could also accept `SchemaDefinition<T>` instead of `SchemaDefinition`.

vkarpov15 (approved):
Thanks :+1:

davidkhierl:
How to work with boolean type
`Type 'boolean' is not assignable to type 'SchemaDefinitionProperty<undefined>`

mroohian:
> SchemaDefinitionProperty

why would you pass a boolean as the schema value? do you have a sample code?

davidkhierl:
> > SchemaDefinitionProperty
> 
> why would you pass a boolean as the schema value? do you have a sample code?

I posted a question at SO thought I may doing it wrong, https://stackoverflow.com/questions/65635637/mongoose-5-11-11-schemadefinition-typing

Was it normal to define the type as Boolean in the SchemaDefinition  because most of my schema have a Boolean property

mroohian:
@vkarpov15 Hi, as mentioned here, the Schema should also be updated too which was forgotten in this PR.

> 
> BTW, the constructor for `Schema` could also accept `SchemaDefinition<T>` instead of `SchemaDefinition`.

@davidkhierl Thank you for reporting! I can confirm the bug and already created a PR for it: https://github.com/Automattic/mongoose/pull/9789

PR #9763, opened 2021-01-04
Add missing projection typing

vorticalbox:
**Summary**

Adds missing project function to Aggregate to type definitions 

**Examples**

```
Users.aggregate().project({username: 1})
```


vorticalbox:
Technically the type should be

```
project(arg: { [k:string]: 1 | 0 }): this
```

as projection takes `key: 1` for example

```

Users.aggregate().project({username: 1, password: 0});
```

but didn't to keep consistency with the rest of the class.

vkarpov15 (approved):
Thanks :+1:

PR #9766, opened 2021-01-04
feat: implemented getPopulatedDocs()

IslandRhythms:

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -3911,7 +3911,7 @@ Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
     }
     else if (typeof value === 'object' && value !== null) {
       result.push(value);
-    } 
+    }
   }
   return result;
 };
--- a/lib/document.js
+++ b/lib/document.js
@@ -3128,8 +3128,6 @@ Document.prototype.$__getAllSubdocs = function() {
   return subDocs;
 };
 
-
-
 /*!
  * Runs queued functions
  */
@@ -3907,8 +3905,7 @@ Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
     const value = this.get(key);
     if (Array.isArray(value)) {
       result = result.concat(value);
-    }
-    else if (typeof value === 'object' && value !== null) {
+    } else if (value instanceof Document) {
       result.push(value);
     }
   }
--- a/lib/document.js
+++ b/lib/document.js
@@ -3901,10 +3901,9 @@ Document.prototype.populate = function populate() {
 
 /* Returns an array of all populated documents associated with the query. */
 Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
-  const pathway = (Object.keys(this.$__.populated));
+  const keys = (Object.keys(this.$__.populated));
   let result = [];
-
-  for (const key of pathway) {
+  for (const key of keys) {
     const value = this.get(key);
     if (Array.isArray(value)) {
       result = result.concat(value);
--- a/lib/document.js
+++ b/lib/document.js
@@ -3906,10 +3906,10 @@ Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
 
   for (const key of pathway) {
     const value = this.get(key);
-    if (Array.isArray(value)){
+    if (Array.isArray(value)) {
       result = result.concat(value);
     }
-    else if (typeof value === 'object' && value !== null){
+    else if (typeof value === 'object' && value !== null) {
       result.push(value);
     } 
   }
--- a/lib/document.js
+++ b/lib/document.js
@@ -3902,7 +3902,7 @@ Document.prototype.populate = function populate() {
 /* Returns an array of all populated documents associated with the query. */
 Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
   const pathway = (Object.keys(this.$__.populated));
-  const result = [];
+  let result = [];
 
   for (const key of pathway) {
     const value = this.get(key);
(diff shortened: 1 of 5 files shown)


lib/document.js
```diff
@@ -3897,6 +3899,24 @@ Document.prototype.populate = function populate() {
...
 };
 
+/* Returns an array of all populated documents associated with the query. */
+Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
+  const index = (Object.keys(this.$__.populated));
+  // console.log(index);
+  let result = [];
+
+  for (const key of index) {
+    let value = this.get(key);
```
vkarpov15, line 3909: Make this `const` instead of `let`

lib/document.js
```diff
@@ -3897,6 +3899,24 @@ Document.prototype.populate = function populate() {
   return this;
 };
 
+/* Returns an array of all populated documents associated with the query. */
+Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
+  const index = (Object.keys(this.$__.populated));
```
vkarpov15, line 3904: Let's rename this variable to `keys` or something that's more indicative of how this is used.

lib/document.js
```diff
@@ -3897,6 +3899,24 @@ Document.prototype.populate = function populate() {
...
+  let result = [];
+
+  for (const key of index) {
+    let value = this.get(key);
+    if (Array.isArray(value)) result = result.concat(value);
+    else if (typeof value === 'object' && value !== null)
+      result.push(value);
+
+  }
+  console.log(result);
```
vkarpov15, line 3915: Remove this `console.log()`

lib/document.js
```diff
@@ -3897,6 +3899,24 @@ Document.prototype.populate = function populate() {
...
 
+/* Returns an array of all populated documents associated with the query. */
+Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
+  const index = (Object.keys(this.$__.populated));
+  // console.log(index);
+  let result = [];
+
+  for (const key of index) {
+    let value = this.get(key);
+    if (Array.isArray(value)) result = result.concat(value);
```
vkarpov15, line 3910: Mongoose style guidelines generally say no one-line ifs. Curly braces `{}` around `if` always, so this should be:

```javascript
if (Array.isArray(value)) {
  result = result.concat(value);
} else if (typeof value === 'object' && value !== null) {
  result.push(value);
}
```

test/document.test.js
```diff
@@ -9834,4 +9834,35 @@ describe('document', function() {
...
+
+  it('supports getting a list of populated docs (gh-9702)', function() {
+    const Child = db.model('Child', Schema({ name: String }));
+    const Parent = db.model('Parent', {
+    children: [{ type: ObjectId, ref: 'Child' }],
+    child: {type: ObjectId, ref: 'Child'}
+    });
+    
+    return co(function*() {
+    const c = yield Child.create({ name: 'test' });
```
vkarpov15, line 9846: There's some issues with indentation here. Run `npm run lint -- --fix` and that should help fix these errors.

test/document.test.js
```diff
@@ -9834,4 +9834,35 @@ describe('document', function() {
...
+    yield Parent.create({
+    children: [c._id],
+    child: c._id
+    });
+    
+    const p = yield Parent.findOne().populate('children child');
+    
+    p.children; // [{ _id: '...', name: 'test' }]
+    
+    assert.equal(p.$getPopulatedDocs().length, 1);
```
vkarpov15, line 9856: Should be `assert.equal(p.$getPopulatedDocs().length, 2);` because both 'children' and 'child' are populated

vkarpov15 (changes requested):

test/document.test.js
```diff
@@ -9834,4 +9834,35 @@ describe('document', function() {
...
+      p.children; // [{ _id: '...', name: 'test' }]
+
+      assert.equal(p.$getPopulatedDocs().length, 2);
+      assert.equal(p.$getPopulatedDocs()[0], p.children[0]);
+      assert.equal(p.$getPopulatedDocs()[0].name, 'test');
+    });
+  });
+
 });
+
```
vkarpov15, line 9863: Please remove the trailing whitespace

test/document.test.js
```diff
@@ -9834,4 +9834,35 @@ describe('document', function() {
...
+        children: [c._id],
+        child: c._id
+      });
+
+      const p = yield Parent.findOne().populate('children child');
+
+      p.children; // [{ _id: '...', name: 'test' }]
+
+      assert.equal(p.$getPopulatedDocs().length, 2);
+      assert.equal(p.$getPopulatedDocs()[0], p.children[0]);
```
vkarpov15, line 9857: Let's also add a test that `populatedDocs[1]` is strictly equal to `p.child`

lib/document.js
```diff
@@ -3897,6 +3899,23 @@ Document.prototype.populate = function populate() {
   return this;
 };
 
+/* Returns an array of all populated documents associated with the query. */
+Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
+  const pathway = (Object.keys(this.$__.populated));
```
vkarpov15, line 3904: Also not a fan of this name. Let's just use `keys`

vkarpov15 (changes requested):
Can you also add this function to

test/document.test.js
```diff
@@ -9834,4 +9834,30 @@ describe('document', function() {
...
+      assert.equal(p.$getPopulatedDocs()[0].name, 'test');
+      assert.equal(p.$getPopulatedDocs()[1],p.child);
```
vkarpov15, line 9859: ESLint fails here because you need a space after `,`. It's a bit pedantic, I know, but it helps with readability since code is typically read more times than it is written.

vkarpov15 (changes requested):

vkarpov15 (approved):

PR #9768, opened 2021-01-05
Added recommended connection option.

Fernando-Lozano:
"You should set this option to true, except for the unlikely case that it prevents you from maintaining a stable connection."

**Summary**

**Examples**


AbdelrahmanHafez (approved):

vkarpov15 (approved):
Thanks :+1:

PR #9769, opened 2021-01-05
feat: Implemented setting explain flag

IslandRhythms:

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1079,7 +1079,7 @@ Model.exists = function exists(filter, options, callback) {
     return;
   }
   console.log(options);
-  if (typeof options != undefined) {
+  if (typeof options !== 'undefined') {
     if (!('explain' in options)) {
       return query.then(doc => !!doc);
     }
--- a/lib/model.js
+++ b/lib/model.js
@@ -1078,8 +1078,12 @@ Model.exists = function exists(filter, options, callback) {
     });
     return;
   }
-
-  return query.then(doc => !!doc);
+  if (!('explain' in options)) {
+    return query.then(doc => !!doc);
+  }
+  else {
+    return query.then(doc => doc);
+  }
 };
 
 /**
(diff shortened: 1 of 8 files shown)


lib/model.js
```diff
@@ -1079,7 +1078,15 @@ Model.exists = function exists(filter, options, callback) {
     });
     return;
   }
-
+  console.log(options);
```
vkarpov15, line 1081: Get rid of this `console.log()` please

lib/model.js
```diff
@@ -1079,7 +1078,15 @@ Model.exists = function exists(filter, options, callback) {
...
+  if (typeof options !== 'undefined') {
```
vkarpov15, line 1082: This should be `if (options != null)`, because otherwise if `options` is `null` this will error out.

A better alternative would be to do `options = options || {};` so you don't even need this `if` statement. That ensures that `options` is an object.

lib/model.js
```diff
@@ -1079,7 +1078,15 @@ Model.exists = function exists(filter, options, callback) {
...
+    if (!('explain' in options)) {
```
vkarpov15, line 1083: I think this should be `if (!options.explain)`, because `explain` can be `false` as well.

lib/model.js
```diff
@@ -1079,7 +1078,15 @@ Model.exists = function exists(filter, options, callback) {
...
+      return query.then(doc => !!doc);
+    }
+    else {
```
vkarpov15, line 1086: For consistent style, let's do `} else {` instead of `}\nelse {`

vkarpov15 (changes requested):

lib/model.js
```diff
@@ -1079,8 +1078,12 @@ Model.exists = function exists(filter, options, callback) {
     });
     return;
   }
-
-  return query.then(doc => !!doc);
+  options = options || {};
+  if (!options.explain) {
+    return query.then(doc => !!doc);
+  } else {
+    return query.then(doc => doc);
```
vkarpov15, line 1085: No need to do this `then()`, can just do `query.exec()` to return a promise.

Issue #9723, opened 2020-12-18
Unable to set a base typescript interface


judemanutd:
**What is the current behavior?**
I had this particular setup working on mongoose version 5.10.15 but cannot get it working with 5.11.8. Here is the basic gist

I have the initial interface that extends Document from mongoose

```typescript
export interface IOrder extends IOrderBase, Document {
  _id: ObjectID;
  id: ObjectID;
}
```
And then the lean interface which contains the properties for the model

```typescript
export interface IOrderBase {
  _id: ObjectID;
  id: ObjectID;
  publicId: string;
  id_customer: ObjectID;

 .... other columns

  createdAt: Date;
  updatedAt: Date;
}
```

This is what my schema looks like 

```typescript
const OrdersSchema = new Schema(
  {
    publicId: { type: String, default: null },
    id: {
      type: SchemaTypes.ObjectId,
      default: null,
    },
    id_customer: {
      type: SchemaTypes.ObjectId,
      required: true,
    },
    ...... other columns
  },
  {
    timestamps: true,
    validateBeforeSave: true,
    minimize: false,
  },
);

const Orders = model<IOrder>("Orders", OrdersSchema);
```

**What is the expected behavior?**

The above config allowed for type casting lean queries in the following manner
1. If I wanted a mongoose document I could simply run the following `Orders.findOne({})` and it would return a document of type `IOrder`
2. If I wanted to run a lean query I would use the following `Orders.findOne({}).lean<IOrderBase>(true)` and it would return a document of type `IOrderBase`

This allowed me to differentiate between a mongoose document and a lean document but is failing with the latest version.

________________________________

For the line `export interface IOrder extends IOrderBase, Document {` I get the following error 
```
Interface IOrder incorrectly extends interface 'Document<any>'
```
For the line `const Orders = model<IOrder>("Orders", OrdersSchema);` I get the following error
```
Type IOrder does not satisfy the constraint 'Document<any>'
```

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
node version : 14.15-alpine3.10
mongoose : 5.11.8
@types/mongoose - removed and deleted node_modules

What is the updated config to reflect the output type of a query when its lean?

I have tried the following as the test indicates 
```typescript
const result: LeanDocument<IOrderBase> = await Orders.findOne({}).lean(fetchLean);
```
but the compiler throws the following error

```typescript
[...]

PR #9773, opened 2021-01-06
fix(index.d.ts): id's are more type strict

BorntraegerMarc:
**Summary**

With this change the document ID's will reflect better what is actually going on.

As far as I know usually the `id` is a string. But sometimes it can be a `ObjectId`. For example when a sub-sub-document was not populated.

These changes make the types better. Because I don't know that `id` is ever anything else than `string` or `ObjectId`

These changes do not contradict with https://github.com/Automattic/mongoose/issues/9723 as the use case is still valid that ID's can be anything (valid).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -477,8 +477,8 @@ declare module 'mongoose' {
      */
     getChanges(): UpdateQuery<this>;
 
-    /** The string version of this documents _id. */
-    id?: any;
+    /** The string version of this documents _id. Sometimes it's ObjectId when sub-documents were not populated */
+    id?: string | mongodb.ObjectId;
 
     /** Signal that we desire an increment of this documents version. */
     increment(): this;
--- a/index.d.ts
+++ b/index.d.ts
@@ -478,7 +478,7 @@ declare module 'mongoose' {
     getChanges(): UpdateQuery<this>;
 
     /** The string version of this documents _id. Sometimes it's ObjectId when sub-documents were not populated */
-    id?: string | mongodb.ObjectId;
+    id?: string | mongodb.ObjectId;
 
     /** Signal that we desire an increment of this documents version. */
     increment(): this;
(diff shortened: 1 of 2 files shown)


vkarpov15:
Thanks for the PR, but we unfortunately can't merge this. The problem is that `id` actually can be any type depending on your schema.

```javascript
// Perfectly valid
const schema = new Schema({ id: Number });
```

If you define an `id` property in your schema, you overwrite Mongoose's default `id` getter. So this change would be massively backwards breaking and prevent TypeScript users from defining their own `id` property.

Samuel-Beslogic:
Can't it be infered from the Document using the DocType generic parameter?
Something like
```js
class Document<T = any, TQueryHelpers = any, DocType = { id: string }> {
  ...
  id: DocType['id']
  ...
}
```
image
image

aphex:
Can we not just set `id` to `T` also, just like we do for `_id`. Isn't id just a virtual to _id anyway? This way someone can set `T` on the document to make it whatever type they like?

https://github.com/Automattic/mongoose/blob/936f6f0b0fc32ecd37279539c4ba23ffefbde1c0/index.d.ts#L373

vkarpov15:
`id` is a virtual that gets `_id` as a string

aphex:
@vkarpov15 oh great, just confirmed this is the case even if you choose to make it something else.

```
const Cat = mongoose.model('Cat', { _id: Number, name: String });
```

Typeof `id` is number but `_id` is string

Can we just type id as string then instead of any? https://github.com/Automattic/mongoose/blob/master/index.d.ts#L587

Then if we can get something like suggested in this PR (https://github.com/Automattic/mongoose/pull/10699) users can set the type for the ID as well as returning documents with full typing and keeping doctype separate from document.

aphex:
@vkarpov15 well that sucks. What if we provided a way for a user to pass in a new type for it. A generic just for the `id` virtual that defaulted to `any`. This should keep all those extenders happy but allow the rest of us good citizens to strongly type things?

Samuel-Beslogic:
If extends Document is so problematic. Maybe a long-term solution with migration steps for a major release might be preferable.

PR #9788, opened 2021-01-09
Update index.d.ts

amirhh:
fix? `type` is not assignable to type 'never' in update operators
image

using mongoose "^5.11.11"
i removed @types/mongoose
i tried chanding @types/mongodb .d.ts file, no luck

this change is working and is good enough for me

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2168,7 +2168,7 @@ declare module 'mongoose' {
   type FunctionPropertyNames<T> = {
     // The 1 & T[K] check comes from: https://stackoverflow.com/questions/55541275/typescript-check-for-the-any-type
     // eslint-disable-next-line @typescript-eslint/ban-types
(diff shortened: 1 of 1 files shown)


vkarpov15:
Unfortunately this PR breaks a few TypeScript tests. We'll investigate this further for v5.11.12 :+1:

vkarpov15:
I actually managed to repro this issue, but it turns out this issue was due to my interface being defined as:

```javascript
interface ITest extends Document {
  tags?: string;
}
```

Rather than:

```javascript
interface ITest extends Document {
  tags?: string[];
}
```

I get this error when `tags` is a string, but it works fine when `tags` is an array as shown here: https://travis-ci.org/github/Automattic/mongoose/builds/753706893

I'm going to close this PR. Please open a new issue and follow the issue template if this is still an issue.

PR #9795, opened 2021-01-11
Gh9398

IslandRhythms:

### DIFF
--- a/test/types.documentarray.test.js
+++ b/test/types.documentarray.test.js
@@ -154,13 +154,6 @@ describe('types.documentarray', function() {
     }
     assert.equal(threw, false);
 
-    // test the _id option, noId is deprecated
-    NoId = new Schema({
-      title: { type: String }
-    }, { _id: false });
-
-    Subdocument = TestDoc(NoId);
-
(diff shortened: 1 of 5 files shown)


test/types.documentarray.test.js
```diff
@@ -142,13 +142,6 @@ describe('types.documentarray', function() {
     assert.notEqual(a.id({ one: 'rolling', two: 'rock' }).title, 'rock-n-roll');
     assert.equal(a.id({ one: 'rock', two: 'roll' }).title, 'rock-n-roll');
 
-    // test with no _id
-    let NoId = new Schema({
```
vkarpov15, line 146: Need to remove `Noid` below too

vkarpov15 (changes requested):

test/types.documentarray.test.js
IslandRhythms: I had thought that was more so a test to see if it would work with the absence of an id since it was a variable assignment with _id = false and the comment says test with no _id. I'll remove it now.

vkarpov15 (approved):
Thanks :+1:

PR #9805, opened 2021-01-13
Gh discriminator key

IslandRhythms:


lib/helpers/query/castUpdate.js
```diff
@@ -42,6 +41,11 @@ module.exports = function castUpdate(schema, obj, options, context, filter) {
     return obj;
   }
 
+  if (schema.options.strict === 'throw') {
+    throw new StrictModeError();
```
vkarpov15, line 45: This should have an error message. You also need to check that `obj` actually contains the discriminator key

lib/helpers/query/castUpdate.js
```diff
@@ -42,6 +41,11 @@ module.exports = function castUpdate(schema, obj, options, context, filter) {
...
+  } else if (!('overwriteDiscriminatorKey' in context._mongooseOptions)) {
```
vkarpov15, line 46: Should use [`hasOwnProperty()`](https://masteringjs.io/tutorials/fundamentals/hasownproperty) instead.

lib/helpers/query/castUpdate.js
```diff
@@ -42,6 +41,11 @@ module.exports = function castUpdate(schema, obj, options, context, filter) {
...
+    delete obj.type;
```
vkarpov15, line 47: This needs to be `delete obj[schema.options.discriminatorKey]` because the user can set the `discriminatorKey` option to whatever they want, it won't always be `type`

vkarpov15 (changes requested):

lib/helpers/query/castUpdate.js
vkarpov15: Also need to check if `context._mongooseOptions != null`

PR #9828, opened 2021-01-18
Gh 7653

IslandRhythms:


lib/cast.js
```diff
@@ -47,6 +49,10 @@ module.exports = function cast(schema, obj, options, context) {
     schema = getSchemaDiscriminatorByValue(schema, obj[schema.options.discriminatorKey]) || schema;
   }
 
+  if (options.upsert && schema.tree._id.required) {
```
vkarpov15, line 52: Also need to check if `obj._id === null`, it's fine to do an upsert on a doc with a required `_id` if `_id` is set or not defined.

vkarpov15 (changes requested):

vkarpov15:
Since this PR didn't make it into 6.0, I'm going to have to close it unfortunately. Sorry!

PR #9830, opened 2021-01-18
Gh 7951

IslandRhythms:
Read the tests and it said that the only thing failing is the map reduce so I think we should be good. Sorry its not the right commit format as I thought I had more steps to take.


vkarpov15 (approved):
Thanks :+1:

PR #9834, opened 2021-01-19
TransactionOptions support within transaction wrapper

pnutmath:
**Summary**

Currently I am using `withTransaction` to handle the transaction in my code. As mongoose introduced transaction wrapper to handle Mongoose document state https://mongoosejs.com/docs/api/connection.html#connection_Connection-transaction

There is no option to provide the `mongodb.TransactionOption` basically supported by `withTransaction` method.

Added support to pass the `mongodb.TransactionOptions`

**Examples**

      const doc = new Person({ name: 'Will Riker' });
      await db.transaction(async function setRank(session) {
        doc.rank = 'Captain';
        await doc.save({ session });
        doc.isNew; // false
 
        // Throw an error to abort the transaction
        throw new Error('Oops!');
      },{ readPreference: 'primary' }).catch(() => {});
 
      // true, `transaction()` reset the document's state because the
      // transaction was aborted.
      doc.isNew;

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -448,7 +448,7 @@ Connection.prototype.startSession = _wrapConnHelper(function startSession(option
  *
  *       // Throw an error to abort the transaction
  *       throw new Error('Oops!');
- *     }).catch(() => {});
+ *     },{ readPreference: 'primary' }).catch(() => {});
  *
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #9845, opened 2021-01-21
fix: replaced this.schema with this.$__schema where appropriate

IslandRhythms:


lib/query.js
```diff
@@ -81,7 +81,7 @@ function Query(conditions, options, model, collection) {
 
   if (model) {
     this.model = model;
-    this.schema = model.schema;
```
vkarpov15, line 84: No need to make this change in `query.js`. Queries don't have user-defined properties. We just need to make this change for documents.

lib/schema/SingleNestedPath.js
```diff
@@ -38,7 +38,7 @@ function SingleNestedPath(schema, path, options) {
   this.caster = _createConstructor(schema);
   this.caster.path = path;
   this.caster.prototype.$basePath = path;
-  this.schema = schema;
```
vkarpov15, line 41: This change is also unnecessary. SchemaTypes can keep `this.schema`, so stuff under `lib/schema` doesn't need to change. Files under `lib/types` are the ones that need to change.

lib/document.js
```diff
@@ -213,7 +213,7 @@ for (const i in EventEmitter.prototype) {
  * @instance
  */
 
-Document.prototype.schema;
+Document.prototype.$__schema;
```
vkarpov15, line 216: We should keep `Document.prototype.schema` for backwards compatibility. `$__schema` is what Mongoose will use internally, but there will still be a `schema` property that developers can use if they need to. Also, `$__schema` should be marked with `@api private` so it isn't part of the public API.

vkarpov15 (changes requested):

vkarpov15:
Also, tests all fail, so please make sure to run some tests, say `npm test -- -g "ocument"`, to make sure that this change doesn't break things too badly.

test/schema.test.js
```diff
@@ -1696,57 +1696,57 @@ describe('schema', function() {
...
+      const ret = this.$__schema.clone().remove('g');
       assert.ok(ret instanceof Schema);
     });
 
     it('removes a single path', function(done) {
-      assert.ok(this.schema.paths.a);
-      this.schema.remove('a');
-      assert.strictEqual(this.schema.path('a'), undefined);
-      assert.strictEqual(this.schema.paths.a, void 0);
+      assert.ok(this.$__schema.paths.a);
```
vkarpov15, line 1704: Tests should still use `this.schema`, not `this.$__schema`. `this.$__schema` is for internal use in case where the user decides to add a `schema` path to their schema, but we should still make sure `schema` works.

IslandRhythms:
So when I run "npm test -- -g ocument" a lot of tests fail.

IslandRhythms:
No test failures on my local machine

vkarpov15 (approved):

vkarpov15 (approved):

PR #9855, opened 2021-01-23
Upgrade to MongoDB Driver 4.x

vkarpov15:


vkarpov15:
Notes on some breaking changes:

1) The `overwrite` option to `update()`, `updateOne()`, etc. is no longer supported. You need to use `replaceOne()` instead
2) `Connection#host`, `Connection#port`, `Connection#name` will not be set until after connection is established.
3) MongoDB driver 4.x requires Node.js 10 or higher.

vkarpov15:
4. MongoDB driver no longer supports `bufferMaxEntries`
5. The result of `updateOne()` and `updateMany()` operations no longer has `ok`, `n`, or `nModified` properties. These properties are replaced with `acknowledged`, `matchedCount`, and `modifiedCount`
6. The result of `deleteOne()` and `deleteMany()` operations no longer has an `n` property, it has been renamed `deletedCount`

vkarpov15:
7. No more `geoSearch()` or `geoHaystackSearch()` re: https://docs.mongodb.com/manual/reference/command/geoSearch/
8. `poolSize` option is gone, replaced by `minPoolSize` and `maxPoolSize`

PR #9878, opened 2021-01-29
Add missing overload to Model.validate (#9877)

### LINKED ISSUE #9877: Missing overloads of Model.validate
Bug

Only 2 overloads of Model.validate() are defined. The docs for [Model.validate](https://mongoosejs.com/docs/api/model.html#model_Model.validate) differs from the [type declarations](https://github.com/Automattic/mongoose/blob/5a999819cc60811996b0bcf4e84046249a8ba8c5/index.d.ts#L753)

Type Model.validate({key: value}, ['key']) in a ts project

**If you are using TypeScript, please include your `tsconfig.json`**
```
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "target": "es2017",
    "module": "commonjs",
    "lib": ["es2017"],
    "rootDir": "./../",
    "noEmit": true,
    "strict": true,
    "noImplicitAny": false,
    "moduleResolution": "node",
    "baseUrl": "./",
    "paths": {
      "@App/*": ["../src/*"],
      "@Mocks/*": ["./__mocks__/*"]
    },
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true
  },
  "include": ["../src/**/*", "../environment.d.ts", "./"],
  "exclude": ["node_modules"]
}
```

Typescript does not complain about missing overloads

node: 15.3.0
mongoose: 5.11.11
mongo: 4.4.3

jonamat:

**Summary**

This PR fixes #9877

**Examples**

Now if you type Model.validate({key: "value"}, ["key"]) on a ts project, ts compiler doesn't warn about invalid arguments 
`Argument of type 'string[]' is not assignable to parameter of type '(err: any) => void'. Type 'string[]' provides no match for the signature '(err: any): void'.`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -753,6 +753,7 @@ declare module 'mongoose' {
     /** Casts and validates the given object against this model's schema, passing the given `context` to custom validators. */
     validate(callback?: (err: any) => void): Promise<void>;
     validate(optional: any, callback?: (err: any) => void): Promise<void>;
+    validate(optional: any, pathsToValidate: string[], callback?: (err: any) => void): Promise<void>;
 
     /** Watches the underlying collection for changes using [MongoDB change streams](https://docs.mongodb.com/manual/changeStreams/). */
     watch(pipeline?: Array<Record<string, unknown>>, options?: mongodb.ChangeStreamOptions): mongodb.ChangeStream;


vkarpov15 (approved):
Thanks :+1:

PR #9886, opened 2021-02-01
actually updated all the isSelected to $__isSelected

IslandRhythms:
sorry again

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -9878,4 +9878,61 @@ describe('document', function() {
       assert.equal(ss.nested[0].toHexString(), updatedElID);
     });
   });
+  it('gh9884', function() {
+    // mongoose.set('useFindAndModify', false);
+    // run().catch(err => console.log(err));
+
+    return co(function*() {
+
+      const obi = new Schema({
+        eType: {
+          type: String,
+          required: true,
+          uppercase: true
+        },
+        eOrigin: {
+          type: String,
+          required: true
+        },
+        eIds: [
+          {
+            type: String
+          }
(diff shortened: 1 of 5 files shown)


test/document.test.js
```diff
@@ -9878,4 +9878,61 @@ describe('document', function() {
       assert.equal(ss.nested[0].toHexString(), updatedElID);
     });
   });
+  it('gh9884', function() {
+    // mongoose.set('useFindAndModify', false);
```
vkarpov15, line 9882: Get rid of these comments please

test/document.test.js
```diff
@@ -9878,4 +9878,61 @@ describe('document', function() {
...
+          {
+            eType: 'X',
+            eOrigin: 'Y',
+            eIds: ['Y', 'Z']
+          }
+        ]
+      };
+
+      const Model = db.model('Test', schema);
+      console.log('Here');
```
vkarpov15, line 9932: Get rid of the `console.log()` calls here

test/document.test.js
```diff
@@ -9878,4 +9878,61 @@ describe('document', function() {
...
+      yield Model.create(newDoc);
+
+      console.log('Done', yield Model.findOne(), new Date());
```
vkarpov15, line 9935: and this one too. Just do `const doc = yield Model.findOne(); assert.ok(doc);` and that should be enough

vkarpov15 (changes requested):

vkarpov15 (approved):
Thanks :+1:

PR #9888, opened 2021-02-02
fix(index.d.ts): allow required function in array definition

Ugzuzg:
**Summary**

The `required` field supports defining an array with a function and a message. Modifying the typescript definitions to reflect that.


vkarpov15 (approved):
Thanks :+1:

PR #9890, opened 2021-02-02
fix(index.d.ts): reorder create typings to allow array desctructuring

Ugzuzg:
**Summary**

Current typings do not allow array destructuring, as TypeScript chooses `create<DocContents = T | DocumentDefinition<T>>(doc: DocContents): Promise<T>;` definition even while using `create` with arrays.


vkarpov15 (approved):
Thanks :+1:

PR #9891, opened 2021-02-02
changes not being reflected

IslandRhythms:
These changes worked in my local environment but for some reason when I run it, the code ignores everything

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -9933,7 +9933,7 @@ describe('document', function() {
   });
   it('gh9880', function(done) {
     const testSchema = new Schema({
-      prop:  String,
+      prop: String,
       nestedProp: {
         prop: String
       }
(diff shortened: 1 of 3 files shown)


test/document.test.js
```diff
@@ -9931,4 +9931,63 @@ describe('document', function() {
       assert.ok(doc);
     });
   });
+  it('gh9880', function() {
+    const mongoose = require('mongoose');
+    const { Schema } = mongoose;
+    mongoose.connect('mongodb://localhost/test', {useNewUrlParser: true, useUnifiedTopology: true});
```
vkarpov15, line 9937: I'd just get rid of lines 9935-9937, not necessary

IslandRhythms:
The failure's are not caused by my test. They are caused by gh9884

IslandRhythms:
I checked gh9884 on my machine and it passed no problem

IslandRhythms:
I believe the two failures are unrelated to my changes so should be good to go.

test/document.test.js
```diff
@@ -9931,4 +9931,58 @@ describe('document', function() {
...
+          updatedDoc.nestedProp; // 'MongooseDocument { null }'
+
+          // now this will throw a TypeError:
+          const failing = new Test({
+            prop: 'Test 3',
+            nestedProp: updatedDoc.nestedProp
+          });
+          failing;
+        });
+        done();
```
vkarpov15, line 9984: `done()` should be on line 9982 instead of `failing;`, the test isn't done until the `new Test({ prop: 'Test 3' })` code is done running

vkarpov15 (changes requested):

vkarpov15 (approved):
Nice :+1:

PR #9896, opened 2021-02-03
Refactor: improve types

### LINKED ISSUE #9850: [types] Support custom Query Method types
types

not being able to specify custom query types

being able to specify custom query types

nodejs: 14.15.4
mongoose: 5.11.11

---

in the old types, Query Methods could be defined with the `QueryHelpers` generic in [`Model`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c68565effdc8cef247f8ed1029f78ade9515264c/types/mongoose/index.d.ts#L3193)

ggurkal:
Implement query method types and improve `create` method's types by excluding internals and infering children types.

Fixes
#9850

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -590,15 +590,26 @@ declare module 'mongoose' {
     validateSync(pathsToValidate?: Array<string>, options?: any): NativeError | null;
   }
 
-  type CreateDoc<T> = {
-    [P in keyof T as
-      T[P] extends Function
-      ? never
-      : P
-    ]: T[P] extends infer R
-      ? CreateDoc<R>
-      : T[P];
-  };
+  type CreateDoc<T> = Omit<
+    {
+      [P in keyof T as
+        T[P] extends Function
+        ? never
+        : P
+      ]: T[P] extends infer R
+        ? CreateDoc<R>
+        : T[P];
+    },
+    | '$locals'
+    | '$op'
+    | '$where'
+    | 'db'
+    | 'isNew'
+    | 'modelName'
+    | 'schema'
+    | 'baseModelName'
+    | 'errors'
+  >;
 
   export const Model: Model<any>;
   // eslint-disable-next-line no-undef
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1046,7 +1067,7 @@ declare module 'mongoose' {
   type SchemaPreOptions = { document?: boolean, query?: boolean };
   type SchemaPostOptions = { document?: boolean, query?: boolean };
 
-  class Schema<DocType extends Document = Document, M extends Model<DocType> = Model<DocType>, SchemaDefinitionType = undefined> extends events.EventEmitter {
+  class Schema<DocType extends Document = Document, M extends Model<DocType> = Model<DocType>, TQueryHelpers = {}, SchemaDefinitionType = undefined> extends events.EventEmitter {
```
vkarpov15, line 1070: This is unfortunately a breaking change, you'd have to re-order these

index.d.ts
```diff
@@ -622,19 +643,19 @@ declare module 'mongoose' {
...
+    countDocuments(callback?: (err: any, count: number) => void): Query<number, T, TQueryHelpers>;
+    countDocuments(filter: FilterQuery<T>, callback?: (err: any, count: number) => void): Query<number, T, TQueryHelpers>;
 
     /** Creates a new document or documents */
-    create<DocContents = T | DocumentDefinition<T>>(docs: DocContents[], options?: SaveOptions): Promise<T[]>;
-    create<DocContents = T | DocumentDefinition<T>>(doc: DocContents): Promise<T>;
-    create<DocContents = T | DocumentDefinition<T>>(...docs: DocContents[]): Promise<T[]>;
-    create<DocContents = T | DocumentDefinition<T>>(docs: DocContents[], callback: (err: CallbackError, docs: T[]) => void): void;
-    create<DocContents = T | DocumentDefinition<T>>(doc: DocContents, callback: (err: CallbackError, doc: T) => void): void;
+    create<DocContents = T | DocumentDefinition<T>>(docs: CreateDoc<DocContents>[], options?: SaveOptions): Promise<T[]>;
```
vkarpov15, line 654: Why is this change necessary? Seems unrelated.

index.d.ts
```diff
@@ -1149,7 +1170,7 @@ declare module 'mongoose' {
     pre<T extends Model<DocType> = M>(method: 'insertMany' | RegExp, options: SchemaPreOptions, fn: (this: T, next: (err: CallbackError) => void) => void): this;
 
     /** Object of currently defined query helpers on this schema. */
-    query: any;
+    query: TQueryHelpers;
```
vkarpov15, line 1173: Ideally this should fall back to `any` if `TQueryHelpers` isn't set

vkarpov15 (changes requested):
Added a couple of comments. Also, can you please fix the test failures? Looks like there's some TS-related test failures that you should fix. Run `env D=1 npm test -- -g "typescript.*queries"` to get the errors.

index.d.ts
ggurkal: it's there to remove internals, such as `$locals`, `$where`, etc., from the `.create` function, so you only get the interface properties and `_id`, `id` and `__v`.

index.d.ts
ggurkal: i made it fall back to `Record<string, any>` would be better, if you agree.

ggurkal:
@vkarpov15 the TS errors are there for the `master` branch as well, so i don't think they're related to my changes. anyway, i did make two changes (`(res: any)`) in `queries.ts` to at least prevent two of them, but i'm left with this:
```Cannot find module 'mongoose' or its corresponding type declarations.```

vkarpov15:
Going to merge this into a test branch so I can more easily tinker with it :+1:

PR #9914, opened 2021-02-10
📝 fixed confusing sentence in Schema docs

namenyi:
**Summary**

I've found a sentence in [the Schema Aliases section](https://mongoosejs.com/docs/guide.html#aliases) that seems to be mistaken and confusing.

---

```
Now accessing `name` will get you the value of `n`, and setting `n` will set the value of `name`
```

I think it should say ```setting `name` will set the value of `n` ```.

The example a few lines below also supports my assumption:

```js
person.name = 'Not Val';
console.log(person); // { n: 'Not Val' }
```

### DIFF
--- a/docs/guide.pug
+++ b/docs/guide.pug
@@ -378,7 +378,7 @@ block content
     const personSchema = new Schema({
       n: {
         type: String,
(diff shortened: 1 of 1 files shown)


marsonya (approved):

vkarpov15 (approved):
Thanks :+1:

Issue #9857, opened 2021-01-24
Issue with typings on SchemaDefinition


amorimfilipe:
**Do you want to request a *feature* or report a *bug*?** 

Not sure if it is a bug or a question.

**What is the current behavior?**

```ts
interface User {
    name: string
    active: boolean
    points: number
}

type UserDocument = mongoose.Document<User>
type UserSchemaDefinition = mongoose.SchemaDefinition<User>
type UserModel = mongoose.Model<UserDocument>
```

_Behavior A_

```ts
const schemaDefinition: UserSchemaDefinition = {
    name: String, // if add other type instead of String as interface requires string, it don't shows an error.
    active: Boolean,
    points: Number
}

const schema = new mongoose.Schema<UserDocument, UserModel>(schemaDefinition) // error on arg "schemaDefinition" below

// Argument of type '{ name: SchemaDefinitionProperty<string>; active: SchemaDefinitionProperty<boolean>; points: SchemaDefinitionProperty<number>; }' is not assignable to parameter of type '{ [path: string]: SchemaDefinitionProperty<undefined>; }'.
//   Property 'active' is incompatible with index signature.
//     Type 'SchemaDefinitionProperty<boolean>' is not assignable to type 'SchemaDefinitionProperty<undefined>'.
//       Type 'boolean' is not assignable to type 'SchemaDefinitionProperty<undefined>'.ts(2345)
```

if add `boolean | number | date ` to ` type SchemaDefinitionProperty` it don't gives an error.

--- 

_Behavior B_

```ts
const schemaDefinition: UserSchemaDefinition = {
    name: {
        type: String
    },
    active: {
        type: Boolean
    },
    points: {
        type: Number
    }
}

// Type '{ type: StringConstructor; }' is not assignable to type 'SchemaDefinitionProperty<string>'.
//   Types of property 'type' are incompatible.
//     Type 'StringConstructor' is not assignable to type 'string'.ts(2322)

// Type '{ type: BooleanConstructor; }' is not assignable to type 'SchemaDefinitionProperty<boolean>'.
//   Types of property 'type' are incompatible.
//     Type 'BooleanConstructor' is not assignable to type 'boolean | undefined'.
//       Type 'BooleanConstructor' is not assignable to type 'true'.ts(2322)

// Type '{ type: NumberConstructor; }' is not assignable to type 'SchemaDefinitionProperty<number>'.
//   Types of property 'type' are incompatible.
[...]

PR #9927, opened 2021-02-13
SchemaTypeOptions.type is optional

moander:
type must be optional to support schemas using the [typeKey option](https://mongoosejs.com/docs/guide.html#typeKey) and [nested paths](https://mongoosejs.com/docs/subdocs.html#subdocuments-versus-nested-paths).

This PR reverts a change that was merged a few days ago:
https://github.com/Automattic/mongoose/commit/968bd8a57d735f97a2cf1686ba1efc8bf23593d4#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R1367

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1365,7 +1365,7 @@ declare module 'mongoose' {
   }
 
   interface SchemaTypeOptions<T> {
-    type: T;
+    type?: T;
 
     /** Defines a virtual with the given name that gets/sets this path. */
     alias?: string;


vkarpov15:
Unfortunately this change prevents us from throwing an error if the `type` is incorrect. We'll investigate this further, I'd like to avoid merging this PR as is because of #9857, but we'll see if we can get the best of both.

vkarpov15:
I took a closer look and the issue comes down to the fact that the value `String` is a valid `SchemaTypeOptions` because `SchemaTypeOptions` has no required keys. That leaves us in a tough spot with supporting checking incorrect schema definitions where the interface has a path set to `number` but the `Schema` has that path set to `String` while still supporting `typeKey`, because `typeKey` isn't necessarily known at compile time. The approach on a485c40 will handle the case where the user uses `{ type: String }` rather than just `String`, that's the best we can do for now :+1:

Issue #9603, opened 2020-11-30
Cannot find module './drivers/node-mongodb-native/connection' using Parcel


julesmons:
**Do you want to request a *feature* or report a *bug*?**
*bug*

**What is the current behavior?**
I'm using parcel v2 to bundle an entire node api into one JS file. 
It's a relatively small api, and i need to bundle all of it because of the hosting env it will ultimately be placed on.

The API uses mongoose and when it's bundled & 'compiled' by parcel it doesn't raise any errors. However, when running the application it doesn't work because certain parts of mongoose are missing, that's where the error "Cannot find module './drivers/node-mongodb-native/connection'" comes from. _The runtime environment can't find this module because it's never bundled._

The cause of this problem has been found in other issues from other libraries. See:
https://github.com/parcel-bundler/parcel/issues/4031
https://github.com/evanw/esbuild/issues/480

The esbuild issue is from a different library, but it boils down to the same problem. 
Some parts of mongoose are required dynamically using a variable. See: https://github.com/Automattic/mongoose/blob/1c2978e31f4c46be135e6ee69e668027ceb4bf5e/lib/index.js#L729-L735

The bundlers don't pick this up and this results in those parts not being bundled, and therefore not being available at runtime.

**If the current behavior is a bug, please provide the steps to reproduce.**

To reproduce this issue, i've included a sample project below. Just use "npm run start" to see the error.

<details>
    <summary>The source files</summary>

    #### package.json
    ```json
    {
        "name": "test",
        "version": "1.0.0",
        "description": "",
        "author": "Jules Mons",
        "license": "ISC",
        "main": "./dist/js/main.js",
        "source": "./src/js/main.js",
        "targets": {
            "main": {
                "distDir": "./dist",
                "context": "node",
                "engines": {
                    "node": "12"
                },
                "includeNodeModules": true
            }
        },
        "scripts": {
            "start": "npm-run-all build serve",
            "serve": "node .",
            "build": "cross-env NODE_ENV=production parcel build ./src/js/main.js --target main"
        },
        "dependencies": {
            "mongoose": "5.10.15"
        },
        "devDependencies": {
            "npm-run-all": "4.1.5",
            "parcel": "^2.0.0-nightly.460"
        }
    }
    ```

    #### src/js/main.js
    ```javascript
    import mongoose from "mongoose";

    mongoose.connect(

        "mongodb://localhost:27017", 
        { useNewUrlParser: true }, 
        (err) => {

            if (!err) {

[...]

Issue #9604, opened 2020-11-30
Remove MONGOOSE_DRIVER_PATH, use a `setDriver()` function on Mongoose instance instead


vkarpov15:
See https://github.com/Automattic/mongoose/issues/9603

PR #9937, opened 2021-02-15
remove dynamic import statements

raphaelschwinger:
**Summary**

Importing from paths generated at runtime breaks 
parcel, see (https://github.com/parcel-bundler/parcel/issues/4031),
or esbuild, see (https://github.com/evanw/esbuild/issues/480).
By removing the calls to the deprecated global MONGOOSE_DRIVER_PATH
this issue can be resolved.

**Examples**
Lines linke this cause problems because the path will not be resolved by eslint etc.
```
const Connection = require(driver + '/connection');
```
Fixes: https://github.com/Automattic/mongoose/issues/9603

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -4,16 +4,7 @@
  * Module dependencies.
  */
 
-if (global.MONGOOSE_DRIVER_PATH) {
-  const deprecationWarning = 'The `MONGOOSE_DRIVER_PATH` global property is ' +
-    'deprecated. Use `mongoose.driver.set()` instead.';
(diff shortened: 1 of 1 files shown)


raphaelschwinger:
Thank you @hugdru for your code suggestions!

vkarpov15 (commented):
Unfortunately we can't merge this into master because it would be a backwards breaking change. I can rebase this change to be against the 6.0 branch, we're planning on removing this flag with #9604 anyway.

vkarpov15 (approved):

Issue #9374, opened 2020-08-30
Make `autoIndex` and `autoCreate` default to `false` if connection's read preference is 'secondary' or 'secondaryPreferred'


chumager:
**Do you want to request a *feature* or report a *bug*?**
Help
**What is the current behavior?**
when I set the readPreference to secondayPreferred in my app I get a mongo error:
```
You have triggered an unhandledRejection, you may have forgotten to catch a Promise rejection:                                                                
MongoError: not master                                                                                                                                        
    at MessageStream.messageHandler (/home/jcmunoz/.yarn/berry/cache/mongodb-npm-3.6.0-1deea15e1a-3.zip/node_modules/mongodb/lib/cmap/connection.js:266:20)   
    at MessageStream.emit (events.js:314:20)                                                                                                                  
    at processIncomingData (/home/jcmunoz/.yarn/berry/cache/mongodb-npm-3.6.0-1deea15e1a-3.zip/node_modules/mongodb/lib/cmap/message_stream.js:144:12)        
    at MessageStream._write (/home/jcmunoz/.yarn/berry/cache/mongodb-npm-3.6.0-1deea15e1a-3.zip/node_modules/mongodb/lib/cmap/message_stream.js:42:5)         
    at writeOrBuffer (_stream_writable.js:352:12)                                                                                                             
    at MessageStream.Writable.write (_stream_writable.js:303:10)                                                                                              
    at TLSSocket.ondata (_stream_readable.js:713:22)                                                                                                          
    at TLSSocket.emit (events.js:314:20)                                                                                                                      
    at addChunk (_stream_readable.js:303:12)                                                                                                                  
    at readableAddChunk (_stream_readable.js:279:9)                                                                                                           
    at TLSSocket.Readable.push (_stream_readable.js:218:10)                                                                                                   
    at TLSWrap.onStreamRead (internal/stream_base_commons.js:188:23)                                                                                          
```
I tested it like in #9309 and it works fine, but as the stack doesn't provide any useful info I can't trace the error.
is there a way to increase the stack trace so I can figure out where is my problem? 

If I avoid the readPreference config I works ok, and I made a little plugin in the **pre /find/ hook** to set the readPreference in the query and it also works fine, so obviously it's not a mongoose problem, but I can't trace it.

It should come from another module or an error from myself, but it's weird because the driver should take care of doing insert or updates in the primary and querying in secondary. 
**If the current behavior is a bug, please provide the steps to reproduce.**

 too much code involve...
[...]

PR #9941, opened 2021-02-16
Gh 9374 from 6.0

Rossh87:
**Summary**
Implements sensible defaults discussed in #9374 if connection readPreference is 'secondary'  or 'preferSecondary'.  If readPreference will be set to 'nearest' or 'primaryPreferred', it warns but otherwise behaves as before.
**Examples**
Several tests added for Mongoose.prototype.connect and Mongoose.prototype.createConnection.

### DIFF
--- a/lib/helpers/processConnectionOptions.js
+++ b/lib/helpers/processConnectionOptions.js
@@ -1,0 +1,65 @@
+'use strict';
+
+const clone = require('./clone');
+const MongooseError = require('../error/index');
+
+function processConnectionOptions(uri, options) {
+  const opts = options ? options : {};
+  const readPreference = opts.readPreference
+    ? opts.readPreference
+    : getUriReadPreference(uri);
+
+  const resolvedOpts = readPreference
+    ? resolveOptsConflicts(readPreference, opts)
+    : opts;
+
+  return clone(resolvedOpts);
+}
+
+function resolveOptsConflicts(pref, opts) {
+  // don't silently override user-provided indexing options
+  if (setsIndexOptions(opts) && setsSecondaryRead(pref)) {
+    throwReadPreferenceError();
+  }
+
+  // if user has not explicitly set any auto-indexing options,
+  // we can silently default them all to false
+  else {
+    return defaultIndexOptsToFalse(opts);
+  }
+}
+
+function setsIndexOptions(opts) {
+  const configIdx = opts.config && opts.config.autoIndex;
+  const { autoCreate, autoIndex, useCreateIndex } = opts;
+  return !!(configIdx || autoCreate || autoIndex || useCreateIndex);
+}
+
+function setsSecondaryRead(prefString) {
+  return !!(prefString === 'secondary' || prefString === 'secondaryPreferred');
+}
+
+function getUriReadPreference(connectionString) {
+  const exp = /(?:&|\?)readPreference=(\w+)(?:&|$)/;
+  const match = exp.exec(connectionString);
+  return match ? match[1] : null;
+}
+
+function defaultIndexOptsToFalse(opts) {
+  opts.config = { autoIndex: false };
+  opts.autoCreate = false;
+  opts.autoIndex = false;
+  opts.useCreateIndex = false;
+  return opts;
+}
+
+function throwReadPreferenceError() {
+  throw new MongooseError(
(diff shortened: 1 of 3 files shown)


lib/helpers/processConnectionOptions.js
```diff
@@ -0,0 +1,93 @@
...
+    const configIdx = opts.config && opts.config.autoIndex;
+    const { autoCreate, autoIndex, useCreateIndex } = opts;
+
+    // don't silently override user-provided indexing options
+    if (configIdx || autoCreate || autoIndex || useCreateIndex) {
+        const conflictLevel = getWarningLevel(pref);
+
+        if (conflictLevel === warnLevels.FAIL) {
+            throwReadPreferenceError();
+        } else if (conflictLevel === warnLevels.WARN) {
```
vkarpov15, line 27: I'd recommend getting rid of the `WARN` logic entirely, unless we add an extra option to explicitly suppress the warning.

test/collection.test.js
```diff
@@ -7,140 +7,140 @@ const assert = require('assert');
...
-        db.close();
-        done();
-      });
-    }
-
-    collection.insertOne({ foo: 'bar' }, {}, function(err, result) {
-      assert.ok(connected);
-      insertedId = result.insertedId;
-      finish();
+describe('collections:', function () {
```
vkarpov15, line 10: It looks like some linter or formatter made a lot of unnecessary changes in this. Can you please undo these unnecessary changes and just commit the tests you're adding without the extra formatting diff?

lib/helpers/processConnectionOptions.js
```diff
@@ -0,0 +1,93 @@
...
+        opts.config = { autoIndex: false };
+        opts.autoCreate = false;
+        opts.autoIndex = false;
+        opts.useCreateIndex = false;
+        return opts;
+    }
+}
+
+function getUriReadPreference(connectionString) {
+    const exp = /(?:&|\?)readPreference=(\w+)(?:&|$)/;
```
vkarpov15, line 47: Nice :+1:

vkarpov15 (changes requested):

lib/helpers/processConnectionOptions.js
Rossh87: No problem.  It seems like we could eliminate the warning behavior in 1 of 4 ways:
1.  Permit the user to set index-creation options for connections that *could* connect to a non-primary and fail somewhat cryptically.
2. Throw for `nearest` and `primaryPreferred` in the same way as `secondary` and `secondaryPreferred`.
3. Override the user's preference and default to false anyway.

I *think* I prefer 2, but after some consideration I'm not sure that silently overriding is as bad as I first thought.  Maybe a user who's savvy enough to notice the override is also savvy enough not to set indexing options on a secondary connection, so we'd end up only throwing for less-familiar users who might prefer a connection that Just Works(TM).  In that case we could cut out the error path entirely, which would simplify the code a fair bit.  Do you have a preference?

test/collection.test.js
Rossh87: Yep, sorry, should have run the linter before commit.

lib/helpers/processConnectionOptions.js
vkarpov15: I prefer (1) actually. Index builds only happen once when the app starts up, so it is reasonable to connect using `primaryPreferred` for durability with the assumption that, when the app starts, the primary will be fine, and you're ok with the index build failing if there isn't a primary.

Rossh87: Cool, sounds good.  I've got all these changes ready to go.  It looks like 6.0 has moved on since I forked it--would you like me to merge/rebase those changes into this PR, or just fix the commented issues?  Thanks!

Rossh87: I went ahead and implemented all the changes from this PR/comments against the latest commit of 6.0, and rolled the changes up into a single commit.  The warn logic and all the unnecessary formatting diff should be gone.  The tests that are failing on this commit are failing on the tip commit of 6.0, so I don't think I introduced any new issues.  Please let me know if there's anything else you'd like changed.  Thanks!

vkarpov15 (approved):
Nice, thanks :+1:

PR #9959, opened 2021-02-21
Improve types

btd:
**Summary**

After 5.11 mongoose started to provide its own types. Some parts of types still use `any (compare the same places with @types/mongoose e.g)

**Examples**

```ts
  const { deletedCount } = await Model.deleteMany({ fieldFromModel });
```

while with some tsconfig settings tsc will not complain. But you will need to do the same for typescript-eslint.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -659,14 +659,14 @@ declare module 'mongoose' {
      * Behaves like `remove()`, but deletes all documents that match `conditions`
      * regardless of the `single` option.
      */
-    deleteMany(filter?: any, options?: QueryOptions, callback?: (err: CallbackError) => void): Query<any, T>;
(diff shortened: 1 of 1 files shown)


btd:
Also question why Document defined type of key as any and not ObjectId?

vkarpov15 (approved):
LGTM :+1:

vkarpov15:
Can you please clarify what you mean by "Also question why Document defined type of key as any and not ObjectId?"

btd:
@vkarpov15 on this line https://github.com/Automattic/mongoose/blob/master/index.d.ts#L369 Document defined as `Document<T = any>`. Why it is not ObjectId? I do not have problem with this now, just interested.

vkarpov15:
Because `_id` can be any type. Nothing wrong with saying `new Schema({ _id: Number })`

PR #9968, opened 2021-02-25
fix: Add generic to plugin schema definition

emiljanitzek:
**Summary**

Update the typescript definition for `plugin()` on Schema to add generic support. Currently, the plugin() complains when trying to add a plugin with custom Document/Model definitions.

**Examples**
```
const myDocumentSchema = new Schema<MyDocument, MyModel>({ name: String });

export default function myPlugin(mySchema: Schema<MyDocument, MyModel>): void {
    mySchema.add({
        plugin: Boolean
    })
}

myDocumentSchema.plugin(myPlugin); // <-- this does not work since Schema in plugin does not support generics
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1123,7 +1123,7 @@ declare module 'mongoose' {
     pathType(path: string): string;
 
     /** Registers a plugin for this schema. */
-    plugin(fn: (schema: Schema, opts?: any) => void, opts?: any): this;
+    plugin(fn: (schema: Schema<DocType, Model<DocType>, SchemaDefinitionType>, opts?: any) => void, opts?: any): this;
 
     /** Defines a post hook for the model. */
     post<T extends Document = DocType>(method: MongooseDocumentMiddleware | MongooseDocumentMiddleware[] | RegExp, fn: (this: T, res: any, next: (err?: CallbackError) => void) => void): this;


index.d.ts
```diff
@@ -1123,7 +1123,7 @@ declare module 'mongoose' {
     pathType(path: string): string;
 
     /** Registers a plugin for this schema. */
-    plugin(fn: (schema: Schema, opts?: any) => void, opts?: any): this;
+    plugin<T extends Document = DocType, L extends Model<T> = M>(fn: (schema: Schema<T, L, SchemaDefinitionType>, opts?: any) => void, opts?: any): this;
```
vkarpov15, line 1126: Why are the generics necessary? Adding `schema: Schema<DocType, Model<DocType>, SchemaDefinitionType>` should be sufficient for this case, shouldn't it?

emiljanitzek: That's correct, it would be sufficient for the case above. I was following the same pattern as found with post/pre methods just below. I can update according to your suggestion if you would prefer that?

vkarpov15: Yes please, I'd rather avoid adding unnecessary generic params if possible.

vkarpov15 (approved):
Thanks :+1:

Issue #9930, opened 2021-02-14
DeprecationWarning: Listening to events on the Db class has been deprecated and will be removed in the next major version


Systerr:
**Do you want to request a *feature* or report a *bug*?**
bug (deprecation warning)

**What is the current behavior?**
On start we have deprecation warning into console 

**If the current behavior is a bug, please provide the steps to reproduce.**
```
(node:9455) DeprecationWarning: Listening to events on the Db class has been deprecated and will be removed in the next major version.
    at _setClient (node_modules/mongoose/lib/connection.js:943:6)
    at node_modules/mongoose/lib/connection.js:839:7
    at node_modules/mongodb/lib/utils.js:693:5
    at node_modules/mongodb/lib/mongo_client.js:227:7
    at connectCallback (node_modules/mongodb/lib/operations/connect.js:366:5)
    at node_modules/mongodb/lib/operations/connect.js:602:5
    at Object.connectHandler [as callback] (/node_modules/mongodb/lib/core/sdam/topology.js:297:43)
    at processWaitQueue (node_modules/mongodb/lib/core/sdam/topology.js:1048:21)
    at NativeTopology.serverUpdateHandler (node_modules/mongodb/lib/core/sdam/topology.js:567:7)
    at Server.emit (events.js:315:20)
```
**What is the expected behavior?**
No deprecation warning
**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Latest 5.11.16

I digged in into warning and found why this happens:
1) This introduced into mongo driver on version v3.6.4 https://github.com/mongodb/node-mongodb-native/pull/2610 as part on preparation of driver v4 release
2) Latest mongoose  5.11.16 introduced mongo drive update https://github.com/Automattic/mongoose/pull/9893
3) Original change happens because of https://github.com/mongodb/node-mongodb-native/pull/2251

Systerr:
https://github.com/Automattic/mongoose/blob/5.11.16/lib/connection.js#L943

pnutmath:
https://jira.mongodb.org/browse/NODE-1709?focusedCommentId=2962047&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2962047

Only MongoClient will receive events not DB class in future

PurpShell:
how does this get fixed?

lantw44:
The warning still exists in mongoose 5.11.8:
```
[...]

PR #10004, opened 2021-03-08
Fix Event Listeners to comply with (new v3 & v4) MongoDB Driver API changes

hillct:
Listening to events on the DB object is deprecated in v3.x and will be unsupported in v4 of the mongodb drivers. Adjusted remaining listeneers to use the DB client object, as per the MongoDB Driver API docs

resolving https://github.com/Automattic/mongoose/issues/9930

**Summary**

**Examples**

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -929,7 +929,7 @@ function _setClient(conn, client, options, dbName) {
         }
       });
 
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #10028, opened 2021-03-14
fix: query object types on schema

francescov1:
**Summary**

The `query` field on the Schema class is typed incorrectly. It types `this` as the Model type, whereas it should actually be the Query type.

Only thing I'm not sure about is if query functions always return another query, or if they can return other values. Let me know if I need to change anything!

**Examples**


vkarpov15 (approved):
You're right that `this` should be a Query instance, but query helpers can technically return anything. We also should default to `Query<any, any, any>`

Issue #9980, opened 2021-03-01
[Typescript] SchemaType.cast (non static) probably has wrong type


hasezoey:
**Do you want to request a *feature* or report a *bug*?**
types

**What is the current behavior?**
it seems like `SchemaType.cast` has value "Function" and wants an "Function" to be returned (plainly speaking)
https://github.com/Automattic/mongoose/blob/abf7b025c3dd17195d3f9df456e68a6a0ce956b0/index.d.ts#L2472-L2489

**If the current behavior is a bug, please provide the steps to reproduce.**

so things like the following dont work:
```ts
class CustomInt extends mongoose.SchemaType {
  constructor(key: string, options: any) {
    super(key, options, 'CustomInt');
  }

  public cast(val) { // Error: Type 'number' is not assignable to type '(v: any) => any'
    return Number(val);
  }
}
```

and as an example, `SchemaNumber` has 3 parameters listed, which are not typed:

https://github.com/Automattic/mongoose/blob/abf7b025c3dd17195d3f9df456e68a6a0ce956b0/lib/schema/number.js#L342-L351

**What is the expected behavior?**

the provided example to not error and have correct types

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
5.11.18

---

am i just maybe using the wrong type to extend from?

PR #10039, opened 2021-03-18
invalid  cast method for instance method of SchemaType

shlomiassaf:
The current implementation is using the same signature for the static level `cast` and instance level `cast`.

However, the instance level `cast` is different.

The use of `any` is crucial here, to allow sub types to specify their narrowed types.
This is true as long as generics is not introduced to `SchemaType`


vkarpov15 (approved):
Unfortunate timing, we just fixed this 30 minutes ago with #9980. We are missing the `prev` and `options` parameters, so we'll add those

Issue #10048, opened 2021-03-20
bulkWrite throws an error when set `timestamps: false` in schema


SoftwareSing:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**

When set `timestamps: false` in schema, `bulkWrite` with `insertOne` or `replaceOne` will throw an error.
But it works fine when set `timestamps: null` or just don't set timestamps.

**If the current behavior is a bug, please provide the steps to reproduce.**

```javascript
const mongoose = require('mongoose')

module.exports = async function bulkInsertOne () {
  const schema = new mongoose.Schema({ name: String }, { timestamps: false })
  const Model = mongoose.model('MyModel', schema)
  await Model.bulkWrite([{
    insertOne: {
      document: { name: 'test' }
    }
  }])
}
```

error:

```
(node:24085) UnhandledPromiseRejectionWarning: TypeError: doc.initializeTimestamps is not a function
    at /home/miku/_____/code/test-for-mongoose-timestamps/node_modules/mongoose/lib/helpers/model/castBulkWrite.js:24:13
    at /home/miku/_____/code/test-for-mongoose-timestamps/node_modules/mongoose/lib/model.js:3500:35
    at each (/home/miku/_____/code/test-for-mongoose-timestamps/node_modules/mongoose/lib/helpers/each.js:11:5)
    at /home/miku/_____/code/test-for-mongoose-timestamps/node_modules/mongoose/lib/model.js:3500:5
    at /home/miku/_____/code/test-for-mongoose-timestamps/node_modules/mongoose/lib/helpers/promiseOrCallback.js:31:5
    at new Promise (<anonymous>)
    at promiseOrCallback (/home/miku/_____/code/test-for-mongoose-timestamps/node_modules/mongoose/lib/helpers/promiseOrCallback.js:30:10)
    at Mongoose._promiseOrCallback (/home/miku/_____/code/test-for-mongoose-timestamps/node_modules/mongoose/lib/index.js:1149:10)
    at Function.Model.bulkWrite (/home/miku/_____/code/test-for-mongoose-timestamps/node_modules/mongoose/lib/model.js:3498:23)
    at bulkInsertOne (/home/miku/_____/code/test-for-mongoose-timestamps/src/bulkInsertOne.js:6:15)
    at main (/home/miku/_____/code/test-for-mongoose-timestamps/src/main.js:7:9)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

[...]

PR #10050, opened 2021-03-20
fix: make bulkWrite can work with `timestamps: false`

SoftwareSing:
**Summary**

bulkWrite use `if (model.schema.options.timestamps != null)` to check that it should run `initializeTimestamps()` or not, so it will got error when set `timestamps: false` in schema.
I saw `Model.$__insertMany` use `if (doc.initializeTimestamps)` to check, so I make bulkWrite use the same condition to fix issue.

**Examples**

#10048

### DIFF
--- a/lib/helpers/model/castBulkWrite.js
+++ b/lib/helpers/model/castBulkWrite.js
@@ -20,7 +20,7 @@ module.exports = function castBulkWrite(originalModel, op, options) {
       const model = decideModelByObject(originalModel, op['insertOne']['document']);
 
       const doc = new model(op['insertOne']['document']);
-      if (doc.initializeTimestamps) {
+      if (model.schema.options.timestamps) {
         doc.initializeTimestamps();
       }
       if (options.session != null) {
@@ -147,7 +147,7 @@ module.exports = function castBulkWrite(originalModel, op, options) {
 
       // set `skipId`, otherwise we get "_id field cannot be changed"
       const doc = new model(op['replaceOne']['replacement'], strict, true);
-      if (doc.initializeTimestamps) {
+      if (model.schema.options.timestamps) {
         doc.initializeTimestamps();
       }
       if (options.session != null) {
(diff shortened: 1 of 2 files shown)


lib/helpers/model/castBulkWrite.js
```diff
@@ -20,7 +20,7 @@ module.exports = function castBulkWrite(originalModel, op, options) {
       const model = decideModelByObject(originalModel, op['insertOne']['document']);
 
       const doc = new model(op['insertOne']['document']);
-      if (model.schema.options.timestamps != null) {
+      if (doc.initializeTimestamps) {
```
vkarpov15, line 23: Shouldn't this be `if (model.schema.options.timestamps)` ?

lib/helpers/model/castBulkWrite.js
```diff
@@ -147,7 +147,7 @@ module.exports = function castBulkWrite(originalModel, op, options) {
 
       // set `skipId`, otherwise we get "_id field cannot be changed"
       const doc = new model(op['replaceOne']['replacement'], strict, true);
-      if (model.schema.options.timestamps != null) {
+      if (doc.initializeTimestamps) {
```
vkarpov15, line 150: Similarly, shouldn't this be `if (model.schema.options.timestamps)`?

vkarpov15 (changes requested):

lib/helpers/model/castBulkWrite.js
SoftwareSing: I considered using it at first, but I tried to find where is also using `doc.initializeTimestamps()`, and I found this:

https://github.com/Automattic/mongoose/blob/9cf48f50d36791f43b517030c5c87a89964bcf75/lib/model.js#L3324-L3326

So I think it's better to use the same condition, if you need to change it in the future, it will be easier to find where you need to edit.
But it just my opinion, if you think it's better to use `if (model.schema.options.timestamps)`, I can change it.

vkarpov15: I think that is meant to handle a separate case. I'll make the changes on my own and merge this :+1:

PR #10054, opened 2021-03-23
fix(index.d.ts): add a missing type for discriminator

coro101:
**Summary**

It adds a missing type for discriminator.

**Examples**

This commit will make the following code works in TS:
```Typescript
Image.discriminator<ImageChildDoc, ImageChildModel>('ImageChild', imageChildSchema);
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -782,6 +782,7 @@ declare module 'mongoose' {
 
     /** Adds a discriminator type. */
     discriminator<D extends Document>(name: string, schema: Schema, value?: string): Model<D>;
+    discriminator<T extends Document, U extends Model<T, TQueryHelpers>, TQueryHelpers = {}>(name: string, schema: Schema<T, U>, value?: string): U;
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -782,6 +782,7 @@ declare module 'mongoose' {
 
     /** Adds a discriminator type. */
     discriminator<D extends Document>(name: string, schema: Schema, value?: string): Model<D>;
+    discriminator<T extends Document, U extends Model<T, TQueryHelpers>, TQueryHelpers = {}>(name: string, schema: Schema<T, U>, value?: string): U;
```
vkarpov15, line 785: Is there any need for `TQueryHelpers` here? It doesn't seem necessary since it's just in the `Model` generic.

coro101: That's true! We don't need `TQueryHelpers` here.

Thanks for careful reading. 😃

vkarpov15 (approved):
Thanks :+1:

PR #10055, opened 2021-03-23
Fix for 'ee.listeners' error

emrebass:
Fix for 'ee.listeners is not a function' error.


vkarpov15 (approved):
I took a quick look and wasn't able to repro this issue. It would be nice if you had a script that demonstrated this issue, because then we could figure out why `ee` isn't an EventEmitter here. But I don't see any harm in merging this as is :+1:

PR #10062, opened 2021-03-24
Gh 10025

IslandRhythms:
Some tests are failing but they were failing before I made these changes.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -70,7 +70,12 @@ function Connection(base) {
   this._closeCalled = false;
   this._hasOpened = false;
   this.plugins = [];
-  this.id = id++;
+  // this.id = id++;
+  if (typeof base === 'undefined' || !base.connections.length) {
+    this.id = 0;
+  } else {
+    this.id = base.connections.length;
+  }
 }
 
 /*!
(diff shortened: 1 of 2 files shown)


test/connection.test.js
```diff
@@ -1299,4 +1299,21 @@ describe('connections:', function() {
       assert.ifError(errorOnDisconnect);
     });
   });
+  it('Connection id should be scoped per Mongoose Instance (gh-10025)', function() {
+    const m = new mongoose.Mongoose;
+    // console.log('First Instance', m.connection.id); // should be x
```
vkarpov15, line 1304: Get rid of these log statements please

lib/connection.js
```diff
@@ -70,7 +70,12 @@ function Connection(base) {
   this._closeCalled = false;
   this._hasOpened = false;
   this.plugins = [];
-  this.id = id++;
+  // this.id = id++;
+  if (typeof base === 'undefined' || !base.connections.length) {
```
vkarpov15, line 74: I'd just get rid of this `if` statement, shouldn't be necessary because `Connection` constructor is private so `base` should always be a Mongoose instance.

lib/connection.js
```diff
@@ -25,7 +25,7 @@ const parseConnectionString = require('mongodb/lib/core').parseConnectionString;
 const arrayAtomicsSymbol = require('./helpers/symbols').arrayAtomicsSymbol;
 const sessionNewDocuments = require('./helpers/symbols').sessionNewDocuments;
 
-let id = 0;
+// let id = 0;
```
vkarpov15, line 28: Can you get rid of this line too please?

vkarpov15 (changes requested):

lib/connection.js
IslandRhythms: capture
without this if statement these errors pop up

vkarpov15 (approved):
Nice :+1:

PR #10063, opened 2021-03-24
collation not added to text indexes

IslandRhythms:

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1633,6 +1633,9 @@ function _ensureIndexes(model, options, callback) {
 
     const indexFields = utils.clone(index[0]);
     const indexOptions = utils.clone(index[1]);
+    const textBlock = Object.keys(indexFields).map(function(key) {
+      return indexFields[key];
+    });
     delete indexOptions._autoIndex;
     _decorateDiscriminatorIndexOptions(model, indexOptions);
     if ('safe' in options) {
@@ -1652,7 +1655,7 @@ function _ensureIndexes(model, options, callback) {
       indexOptions.background = options.background;
     }
     if (model.schema.options.hasOwnProperty('collation') &&
-        !indexOptions.hasOwnProperty('collation') && Object.values(indexFields) != 'text') {
+        !indexOptions.hasOwnProperty('collation') && textBlock != 'text') {
       indexOptions.collation = model.schema.options.collation;
     }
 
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -1634,8 +1633,16 @@ function _ensureIndexes(model, options, callback) {
 
     const indexFields = utils.clone(index[0]);
     const indexOptions = utils.clone(index[1]);
+    const textBlock = () => {
+      const results = Object.keys(indexFields).map(function(key) {
+        return indexFields[key];
+      });
+      if (results.indexOf('text') !== -1) {
```
vkarpov15, line 1640: Why not just check `indexFields.hasOwnProperty('text')`? That way you don't need this extra function, and it seems silly to convert an object to an array of keys just to use [`indexOf()`](https://masteringjs.io/tutorials/fundamentals/includes#indexof) to search through that array because `indexOf()` is much slower than accessing a document's keys.

More on [`hasOwnProperty()`](https://masteringjs.io/tutorials/fundamentals/hasownproperty).

lib/model.js
```diff
@@ -1654,7 +1661,7 @@ function _ensureIndexes(model, options, callback) {
       indexOptions.background = options.background;
     }
     if (model.schema.options.hasOwnProperty('collation') &&
-        !indexOptions.hasOwnProperty('collation')) {
+        !indexOptions.hasOwnProperty('collation') && textBlock()) {
```
vkarpov15, line 1664: I'd just pull in a check for `indexFields.hasOwnProperty(...)` here

test/model.indexes.test.js
```diff
@@ -666,5 +666,24 @@ describe('model', function() {
...
+            strength: 2
+          }
+        });
+        userSchema.index({ username: 'text' }, { unique: true });
+        const User = db.model('User', userSchema, 'User');
+
+        yield User.init();
+        const indexes = yield User.listIndexes();
+        assert.ok(!indexes[1].collation);
+        yield User.collection.drop();
```
vkarpov15, line 685: Doesn't look like this `drop()` call is necessary. Does it cause tests to fail if we remove it?

vkarpov15 (changes requested):

test/model.indexes.test.js
IslandRhythms: does not, will remove

lib/model.js
IslandRhythms: but doesn't `hasOwnProperty()` check the keys, not the value? When I run `indexFields.hasOwnProperty('text')` it returns false even though `'text'` does exist in `indexFields`.

vkarpov15 (approved):

PR #10094, opened 2021-04-02
fix: add missing asyncInterator to Query type def

### LINKED ISSUE #10093: Query's Typescript definition misses an asyncIterator
Report a bug

In Typescript, it is not possible to use a Query instance in a `for await` loop

According to [documentation](https://mongoosejs.com/docs/queries.html#streaming), I should be able to do the following:
```
for await (const doc of Person.find()) {
  console.log(doc); // Prints documents one at a time
}
```

My `tsconfig.json`
```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2019",
    "lib": [],
    "esModuleInterop": true,
    "strict": true,
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "sourceMap": true,
    "outDir": "dist",
    "baseUrl": ".",
  },
  "include": [
    "src/**/*.ts*",
  ]
}
```

The code above should compile

Node: v12.21.0
Typescript: v4.2.3
Mongoose: v5.12.3
MongoDB: v3.6.5 (DT: 3.6.11)

**Workaround**
Explicitly create a QueryCursor:
```
for await (const doc of Person.find().cursor()) {
  console.log(doc); // Prints documents one at a time
}
```

borfig:

**Summary**

Fixes #10093. According to [documentation](https://mongoosejs.com/docs/queries.html#streaming), a `Query` instance has an async iterator (which is in fact the `.cursor()`'s async iterator). This pull request makes it possible in Typescript.

**Examples**
Now it is possible to do the following in Typescript:
```
for await (const doc of Person.find()) {
  console.log(doc); // Prints documents one at a time
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1850,6 +1850,13 @@ declare module 'mongoose' {
   class Query<ResultType, DocType extends Document, THelpers = {}> {
     _mongooseOptions: MongooseQueryOptions;
 
+    /**
+     * Returns a wrapper around a [mongodb driver cursor](http://mongodb.github.io/node-mongodb-native/2.1/api/Cursor.html).
+     * A QueryCursor exposes a Streams3 interface, as well as a `.next()` function.
+     * This is equivalent to calling `.cursor()` with no arguments.
+     */
+    [Symbol.asyncIterator]: QueryCursor<DocType>;
+
     /** Executes the query */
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Nice, thanks :+1:

PR #10098, opened 2021-04-02
fix(model): `insertedDocs` may contain docs that weren't inserted

olnazx:
Fixes the bug when `insertedDocs` property may contain docs that were not actually inserted. It only happens when `Model.insertMany()` called with `ordered: true` option. 

Example:

```js
const Test = mongoose.model('Test', new mongoose.Schema({ _id: { type: String } }), 'tests');

Test.insertMany(
  [
    { _id: 'test_id' },
    { _id: 'test_id_2' },
    { _id: 'test_id' },
    { _id: 'test_id_3' }
  ],

  { ordered: true }
).catch(error => {
  console.log(error.insertedDocs);

  // =>
  //
  // [
  //   { _id: 'test_id', __v: 0 },
  //   { _id: 'test_id_2', __v: 0 },
  //   { _id: 'test_id_3', __v: 0 }
  // ]

  // mongo$ > db.tests.find()
  //
  // { "_id" : "test_id", "__v" : 0 }
  // { "_id" : "test_id_2", "__v" : 0 }
});
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3343,8 +3343,24 @@ Model.$__insertMany = function(arr, options, callback) {
 
         // `insertedDocs` is a Mongoose-specific property
         const erroredIndexes = new Set(get(error, 'writeErrors', []).map(err => err.index));
+
+        let firstErroredIndex = -1;
         error.insertedDocs = docAttributes.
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Nice, thanks :+1:

Issue #10100, opened 2021-04-02
Model.save does not store fields that set to be exclusive at schema-level


ptantiku:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Cannot create a new document with correct fields, if some fields are set to be exclusive at schema-level by using `select: false` and schema option to `toObject.useProjection: true`.

**If the current behavior is a bug, please provide the steps to reproduce.**

```js
const mongoose = require('mongoose')

const { Schema } = mongoose;
mongoose.set('debug', true);

async function main() {
  await mongoose.connect('mongodb://localhost:27017/test', {
    useNewUrlParser: true,
    useUnifiedTopology: true,
    useFindAndModify: false,
  });

  const User = mongoose.model('users', new Schema({
    username: { type: String, required: true },
    password: { type: String, required: true, select: false },  // hidden by default
  }, {
    toObject: {
      useProjection: true,  // to enable projection in schema-level
    },
  }));

  const user = new User({ username: 'testuser', password: 'password' });
  await user.save();
  /* Mongoose's debug message shows:
  Mongoose: users.insertOne({ _id: ObjectId("60676b786d511d503cc8ee34"), username: 'testuser', __v: 0}, { session: null })
  */
  console.log(user);  // no "password" field here
  console.log(user.toObject());  // no "password" field here
  console.log(user.toObject({ useProjection: false }));  // "password" field is shown here

  await mongoose.connection.dropDatabase()
}

main().catch(err => console.log(err));
```

**What is the expected behavior?**
It should do the insert document containing both `username` and `password` field.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Node.js   v14.15.4
Mongoose   v5.11.16
MongoDB   v4.4.2

---

I guess the code at [this line](https://github.com/Automattic/mongoose/blob/master/lib/model.js#L259) is the cause of the problem. This function uses `toObject()` (with [default arguments](https://github.com/Automattic/mongoose/blob/master/lib/options.js#L7)) to create a plain object of data before saving it. But with default arguments, the hidden field does not come out from `toObject()` so it does not get saved into DB. And I don't know if this can be overridden from the outside of the library.

Perhaps, adding `useProjection: false` to [/lib/options](https://github.com/Automattic/mongoose/blob/master/lib/options.js#L7) would solve this?

PR #10101, opened 2021-04-02
fix: Model.save does not store fields set with select:false in schema-level

ptantiku:
**Summary**

This fixes issue #10100 by adding `useProjection: false` into `internalToObjectOptions` variable in "/lib/option.js" file so that when calling `Model.save()`, it would save all the fields, including the fields set `select: false` in schema-level.

**Examples**

Model.save() should store all the fields.
```js
const User = mongoose.model('users', new Schema({
  username: { type: String, required: true },
  password: { type: String, required: true, select: false },  // hidden by default
}, {
  toObject: {
    useProjection: true,  // to enable projection in schema-level
  },
}));

const user = new User({ username: 'testuser', password: 'password' });
await user.save();    // this should do users.insertOne({_id: ObjectId("xxx"), username: 'testuser', password: 'password', __v: 0});
```

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -11,5 +11,5 @@ exports.internalToObjectOptions = {
   _skipDepopulateTopLevel: true,
   depopulate: true,
   flattenDecimals: false,
-  useProjection: false,
+  useProjection: false
 };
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
I confirmed and the test failures are due to issues in master. Thanks :+1:

PR #10109, opened 2021-04-06
refactor(index.d.ts): move `discriminator()` to common interface

LoneRifle:
**Summary**

To ensure consistency in `discriminator()` declarations across
Model, Array and DocumentArray, create a common interface that
has the declarations found in Model, and have the other two classes
implement the interface

- Move `Model.discriminator()` declarations to AcceptsDiscriminator
- Adjust the declarations so that both accept a Schema with at least
  the type argument for Document, to line up with expected type argument
  of the returned Model
- Implement AcceptsDiscriminator in Array and DocumentArray by copying
  the method declarations into the respective class declarations

**Example**

```ts
interface SubDocument extends Document {
  // ...
}

const subSchema = new Schema<SubDocument>({
  // ...
})

const baseSchema = new Schema<Document>({
  // ...
})

const Base = mongoose.model('base', baseSchema)
const Sub = Base.discriminator('Sub', subSchema)
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -598,9 +598,15 @@ declare module 'mongoose' {
     validateSync(pathsToValidate?: Array<string>, options?: any): NativeError | null;
   }
 
+  interface AcceptsDiscriminator {
+    /** Adds a discriminator type. */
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Nice, thanks :+1:

PR #10119, opened 2021-04-08
Gh 10068

IslandRhythms:

**Summary**

**Examples**

### DIFF
--- a/lib/helpers/populate/assignVals.js
+++ b/lib/helpers/populate/assignVals.js
@@ -233,7 +233,7 @@ function valueFilter(val, assignmentOpts, populateOptions, allIds) {
     }
     return val;
   }
-
+  // console.log(populateOptions);
   // findOne
   if (isPopulatedObject(val)) {
     maybeRemoveId(val, assignmentOpts);
@@ -243,6 +243,7 @@ function valueFilter(val, assignmentOpts, populateOptions, allIds) {
   if (val instanceof Map) {
     return val;
   }
+  // if(typeof populateOptions._queryProjection === 'undefined') return transform(val, allIds)
   if (populateOptions.justOne === true) {
     return val == null ? transform(val, allIds) : transform(null, allIds);
   }
--- a/lib/helpers/populate/assignVals.js
+++ b/lib/helpers/populate/assignVals.js
@@ -242,11 +242,11 @@ function valueFilter(val, assignmentOpts, populateOptions, allIds) {
   if (val instanceof Map) {
     return val;
   }
-
+  /*
   if (typeof populateOptions._queryProjection === 'undefined') {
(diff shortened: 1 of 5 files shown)


lib/helpers/populate/assignVals.js
```diff
@@ -166,7 +165,7 @@ module.exports = function assignVals(o) {
     // If lean, need to check that each individual virtual respects
     // `justOne`, because you may have a populated virtual with `justOne`
     // underneath an array. See gh-6867
-    mpath.set(o.path, valueToSet, docs[i], lookupLocalFields, setValue, false);
+    return mpath.set(o.path, valueToSet, docs[i], lookupLocalFields, setValue, false);
```
vkarpov15, line 168: This causes every test that populates multiple docs to fail, because this is an early `return` that prevents every doc after the first from being populated. This change may accidentally fix this one use case, but it breaks a lot of others.

lib/helpers/populate/assignVals.js
```diff
@@ -244,9 +243,14 @@ function valueFilter(val, assignmentOpts, populateOptions, allIds) {
     return val;
   }
 
+  if (typeof populateOptions._queryProjection === 'undefined') {
```
vkarpov15, line 246: This change looks unnecessary as well

vkarpov15 (changes requested):

lib/model.js
```diff
@@ -4554,8 +4548,7 @@ function _assign(model, vals, mod, assignmentOpts) {
   let val;
 
   // Clone because `assignRawDocsToIdStructure` will mutate the array
-  const allIds = utils.clone(mod.allIds);
-
+  let allIds = utils.clone(mod.allIds);
```
vkarpov15, line 4551: Please undo this change, it is unnecessary and causes a linter warning

test/model.populate.test.js
```diff
@@ -10232,4 +10232,80 @@ describe('model: populate:', function() {
...
+      // Get the books and make book2 author in contributors just an id instead of an object
+      const populatedBooks = (yield Books.find().populate(populateOptions)).map(aBook => {
+        aBook = aBook.toObject();
+        if (!aBook._id.equals(aBook1.id)) {
+          aBook.contributors[0].author = aBook.contributors[0].author._id;
+        }
+        return aBook;
+      });
+
+      // console.log('populatedBooks = ' + util.inspect(populatedBooks, false, null, true))
```
vkarpov15, line 10263: Please remove the commented out code, it isn't necessary

vkarpov15 (changes requested):
Couple minor comments but looks good overall

vkarpov15 (approved):

PR #10131, opened 2021-04-14
fix(schemaType): make type Mixed cast error objects to pojos

### LINKED ISSUE #10126: [Possible Bug] support error objects as a value for paths with type `Mixed`
When creating a document, and passing an error object to path that has type `Object`, it should enumerate over its properties and save the values in that object.

It does work in memory when creating a document, but when finding that document from the database, the message and stack properties are lost, only manually added data to the error object persist.
### 10126.js
```js
'use strict';
const { Schema } = mongoose;

run().catch(console.error);

async function run () {
    useNewUrlParser: true,
    useUnifiedTopology: true
  });

  const userSchema = new Schema({ err: Object });

  const err = new Error('I am a bad error');
  err.moreInformation = { reasons: ['Cloudflare is down', 'DNS'] };

  const user = await User.create({ err });
  assertErrorProperties(user); // passes

  const userFromDB = await User.findOne();
  assertErrorProperties(userFromDB); // fails

}

function assertErrorProperties (user) {
  assert.equal(user.err.message, 'I am a bad error');
  assert.ok(user.err.stack);
  assert.equal(user.err.moreInformation.reasons[0], 'Cloudflare is down');
}
```
### Output
```
$ node 10126.js
AssertionError [ERR_ASSERTION]: undefined == 'I am a bad error'
    at assertErrorProperties (E:\Info\localhost\Web Development\NodeJS\speero-backend\test.js:33:10)
    at run (E:\Info\localhost\Web Development\NodeJS\speero-backend\test.js:27:3)
    at processTicksAndRejections (internal/process/task_queues.js:93:5) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: undefined,
  expected: 'I am a bad error',
  operator: '=='
}
```

AbdelrahmanHafez:
fixes #10126

### DIFF
--- a/lib/schema/mixed.js
+++ b/lib/schema/mixed.js
@@ -7,6 +7,7 @@
 const SchemaType = require('../schematype');
 const symbols = require('./symbols');
 const isObject = require('../helpers/isObject');
+const utils = require('../utils');
 
 /**
  * Mixed SchemaType constructor.
@@ -103,6 +104,9 @@ Mixed.set = SchemaType.set;
  */
 
 Mixed.prototype.cast = function(val) {
+  if (val instanceof Error) {
+    return utils.errorToPOJO(val);
+  }
   return val;
 };
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -920,3 +920,16 @@ exports.getOption = function(name) {
  */
 
 exports.noop = function() {};
+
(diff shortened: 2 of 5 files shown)


vkarpov15 (approved):
Works for me :+1: The fact that JS makes `Error#message` and `Error#stack` not enumerable causes a lot of headache.

PR #10165, opened 2021-04-21
Gh 10074

IslandRhythms:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -368,7 +368,7 @@ declare module 'mongoose' {
   }
 
   class Document<T = any, TQueryHelpers = {}> {
-    constructor(doc?: any);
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

serhalp:
@vkarpov15 Am I missing something or does this change not make any sense?

https://github.com/Automattic/mongoose/pull/9699 added a generic type `T` to `Document` which is is the type of the `_id` field, but this PR started using that same type `T` to mean the type of the constructor arg. These two uses of the same generic type aren't compatible.

vkarpov15:
@serhalp thanks for pointing this out, opened a new issue :point_up: to investigate

PR #10167, opened 2021-04-21
Gh 10083

IslandRhythms:
failure is from a test that needs to be adjusted

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -243,10 +243,7 @@ Model.prototype.$__handleSave = function(options, callback) {
     if ('wtimeout' in options.writeConcern) {
       saveOptions.wtimeout = options.writeConcern.wtimeout;
     }
-    if ('checkKeys' in options.writeConcern) {
-      saveOptions.checkKeys = options.writeConcern.checkKeys;
-    }
-    saveOptions = { writeConcern: saveOptions };
+    saveOptions.writeConcern = options.writeConcern;
   } else {
     if ('w' in options) {
       saveOptions.w = options.w;
@@ -257,9 +254,9 @@ Model.prototype.$__handleSave = function(options, callback) {
     if ('wtimeout' in options) {
       saveOptions.wtimeout = options.wtimeout;
     }
-    if ('checkKeys' in options) {
-      saveOptions.checkKeys = options.checkKeys;
-    }
+  }
+  if ('checkKeys' in options) {
+    saveOptions.checkKeys = options.checkKeys;
   }
   const session = this.$session();
   if (!saveOptions.hasOwnProperty('session')) {
--- a/lib/model.js
+++ b/lib/model.js
@@ -233,6 +233,7 @@ Model.prototype.$__handleSave = function(options, callback) {
   }
 
   applyWriteConcern(this.$__schema, options);
+
   if (typeof options.writeConcern != 'undefined') {
     if ('w' in options.writeConcern) {
       saveOptions.w = options.writeConcern.w;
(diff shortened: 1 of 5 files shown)


lib/model.js
```diff
@@ -231,20 +231,36 @@ Model.prototype.$__handleSave = function(options, callback) {
...
+    if ('j' in options.writeConcern) {
+      saveOptions.j = options.writeConcern.j;
+    }
+    if ('wtimeout' in options.writeConcern) {
+      saveOptions.wtimeout = options.writeConcern.wtimeout;
+    }
+    if ('checkKeys' in options.writeConcern) {
+      saveOptions.checkKeys = options.writeConcern.checkKeys;
+    }
+    saveOptions = { writeConcern: saveOptions };
```
vkarpov15, line 249: This really should be `saveOptions.writeConcern = options.writeConcern` or something like that. Why overwrite all `saveOptions`?

lib/model.js
```diff
@@ -231,20 +231,36 @@ Model.prototype.$__handleSave = function(options, callback) {
...
+      saveOptions.w = options.writeConcern.w;
+    }
+    if ('j' in options.writeConcern) {
+      saveOptions.j = options.writeConcern.j;
+    }
+    if ('wtimeout' in options.writeConcern) {
+      saveOptions.wtimeout = options.writeConcern.wtimeout;
+    }
+    if ('checkKeys' in options.writeConcern) {
+      saveOptions.checkKeys = options.writeConcern.checkKeys;
```
vkarpov15, line 247: `checkKeys` isn't a write concern option. I think this is only an issue because you're mistakenly overwriting all of `saveOptions` below.

vkarpov15 (changes requested):

vkarpov15 (changes requested):
I took a look and the query test still fails. We're going to have to make some changes to how `Query.prototype.j`, `Query.prototype.w`, and `Query.prototype.wtimeout` functions work, see https://github.com/Automattic/mongoose/blob/9f9173c81b1daa6a6cb70421575404fb6cd6fae8/lib/query.js#L1209 . That will need an `if` statement like:

```javascript
if (this.options.writeConcern != null) {
  this.options.writeConcern.j = val;
} else {
  this.options.j = val;
}
```

lib/model.js
vkarpov15: You really shouldn't be overwriting `saveOptions` here. You'd be removing all other top-level `saveOptions`, like `session`. Instead, do something like:

```javascript
saveOptions.writeConcern = Object.assign({}, options.writeConcern);
```

lib/query.js
```diff
@@ -1168,7 +1169,12 @@ Query.prototype.w = function w(val) {
   if (val == null) {
     delete this.options.w;
   }
-  this.options.w = val;
+  if (this.options.writeConcern != null) {
+    this.options.writeConcern.w = val;
+  }
+  else {
```
vkarpov15, line 1175: Formatting: `else` on the same line as `{`

lib/query.js
```diff
@@ -1684,7 +1699,20 @@ Query.prototype._optionsForExec = function(model) {
   if (options.upsert !== void 0) {
     options.upsert = !!options.upsert;
   }
-
+  if (options.writeConcern) {
+    if (options.j) {
+      options.writeConcern.j = options.j;
```
vkarpov15, line 1704: Nice :+1: Good job finding the right place to fix this.

vkarpov15 (changes requested):

IslandRhythms:
So something I noticed after the commit is that the first operation ran from 10083 looks like this with `Object.assign` 
```
Mongoose: Test.insertOne({ _id: 12345678, op: 'help', size: 54, totalSize: 104 }, { w: 'majority', j: true, wtimeout: 15000, writeConcern: { w:
 'majority', j: true, wtimeout: 15000 }, session: null})

```
its fine after that

IslandRhythms:
I got it

vkarpov15 (approved):
Thanks :+1:

PR #10168, opened 2021-04-22
fix(index.d.ts): add parameter type in merge

yoonhoGo:
**Summary**

[In Mongoose 5.12.x official document](https://mongoosejs.com/docs/api/query.html#query_Query-merge), type of `source` parameter in `Query.prototype.merge()` is `Query | Object`.
So, I add `FilterQuery<DocType>` to `source` parameter type.

Please check it :)


vkarpov15 (approved):
Thanks :+1:

PR #10169, opened 2021-04-22
style: update carbon format

sayzlim:
**Summary**

Update the Carbon format on the homepage and documentation page so it's still visible in smaller viewport without using display: none. The CTR will affect the earnings, so this update will increase the ad revenue. I've also made some changes to how the sponsors are displayed:

- Remove border surrounding the image.
- Change to flex layout with margin between each image.
- Restrict the height to 50px to ensure they look sharper and aligned.

You can check out the screenshot for desktop, docs, and mobile here: https://dropover.cloud/fcb1e7

### DIFF
--- a/docs/css/mongoose5.css
+++ b/docs/css/mongoose5.css
@@ -76,6 +76,10 @@ pre code {
   text-transform: none;
 }
 
+.pure-menu {
+  padding-bottom: 180px;
+}
+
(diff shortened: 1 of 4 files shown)


vkarpov15 (approved):
Thanks @sayzlim  :+1:

PR #10170, opened 2021-04-23
chore: update carbon + native ads integration

sayzlim:
**Summary**

The legacy native.js is no longer supported. I've replaced it with an official script that is approved on most ad blockers. Few changes I've made in this PR:

- Create `native` as a separate component.
- Replaced `native.js` with `//m.servedby-buysellads.com/monetization.custom.js`
- Replaced carbon ad code with the latest version from sell.buysellads.com

Ideally, we don't have to include native in each `pug` documents and load it inside the layout, but I believe there is a reason why you want to have more control which pages the ad should appear.


vkarpov15 (approved):
Thanks :+1:

sayzlim:
@vkarpov15 do we need to rebuild the docs? Looks like the gh-pages doesn't load the changes made here: https://mongoosejs.com/docs/middleware.html

PR #10171, opened 2021-04-23
Added a navbar for responsive screen

ad99526:
Added a navbar so that doesn't overlap in devices with screen width less than 1100px

**Summary**

**Examples**


vkarpov15 (approved):
Normally we wouldn't accept this PR since it modifies an HTML file directly, but in this case the docs are so out of date that there's no other option.

Do you know how you ended up on this page? Google search, SO link, etc.?

ad99526:
Thanks for approving the pull request. I came upon this page from Google searching for a solution to a problem I was facing.

PR #10172, opened 2021-04-23
chore: rebuild docs

sayzlim:
**Summary**

With the changes from each pug files. It seems that we need to rebuild the docs for the new update to show up on the documentation site.


vkarpov15 (approved):
Correct, you do.

Issue #10147, opened 2021-04-16
Incorrect typescript typings (Document.validateSync return type)


michaln-q:
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
Current typings provided here: https://github.com/Automattic/mongoose/blob/master/index.d.ts#L599 don't match return type described here: https://mongoosejs.com/docs/api.html#document_Document-validateSync (ValidationError !== NativeError)

**What is the expected behavior?**
That function returns ValidationError instead of NativeError
**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Mongoose: 5.12.3
Node: 12.15.1
Typescript: 3.8.3
MongoDB: 3.3

I'm happy to provide a PR if you agree this is a bug.

PR #10175, opened 2021-04-26
fix(index.d.ts): change document.validateSync return type to Validati…

michaln-q:
**Summary**

Should fix https://github.com/Automattic/mongoose/issues/10147


vkarpov15 (approved):
Thanks :+1:

Issue #10146, opened 2021-04-16
Missing typescript typings (CastError constructor)


michaln-q:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

There is no option to create CastError (https://mongoosejs.com/docs/api.html#mongoose_Mongoose-CastError)

**If the current behavior is a bug, please provide the steps to reproduce.**

```
class CustomId extends mongoose.SchemaType {
    public constructor(path: string, options?: mongoose.SchemaTypeOpts<unknown>) {
        super(path, options, 'CustomId');
    }

    public cast(value: unknown): mongoose.Types.ObjectId | string {
        if (typeof value === 'string' and value.startsWith('PROJ')) {
           return value as mongoose.Types.ObjectId;
        }
        throw new mongoose.Error.CastError('CustomId', value, this.path);
    }
}
```

```
{
  "compilerOptions": {
    "esModuleInterop": true,
    "noImplicitAny": true,
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "target": "es2019",
    "baseUrl": "./src",
    "outDir": "./dist",
    "strict": true,
    "paths": {
      "~/*": [
        "./*"
      ]
    },
    "typeRoots": [
      "./node_modules/@types",
      "./src/@types"
    ],
    "noEmit": true
  },
  "include": [
    "src/**/*",
    "tests/**/*",
    "dev/**/*",
    "*"
  ]
}
```

**What is the expected behavior?**
https://mongoosejs.com/docs/api.html#mongoose_Mongoose-CastError
Add a constructor to CastError class, like:

```
 export class CastError extends Error {
      name: 'CastError';
      stringValue: string;
      kind: string;
      value: any;
      path: string;
      reason?: NativeError | null;
      model?: any;
      
      constructor(type: string, value: any, path: string, reason?: NativeError);
    }
```

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Node: 12.15.1
Mongoose: 5.12.3
MongoDB: 3.3.2

michaln-q:
(I'm happy to provide PR with that change)

vkarpov15:
@michaln-q sure, PRs are welcome :+1:

PR #10176, opened 2021-04-26
fix(index.d.ts): add CastError constructor

michaln-q:
**Summary**

Should fix https://github.com/Automattic/mongoose/issues/10146

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2665,6 +2665,8 @@ declare module 'mongoose' {
       path: string;
       reason?: NativeError | null;
       model?: any;
+
+      constructor(type: string, value: any, path: string, reason?: NativeError, schemaType?: SchemaType);
     }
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -2665,6 +2665,8 @@ declare module 'mongoose' {
       path: string;
       reason?: NativeError | null;
       model?: any;
+
+      constructor(type: string, value: any, path: string, reason?: NativeError);
```
vkarpov15, line 2669: There should be 1 more parameter: `schemaType?: SchemaType` . Can you add that please?

vkarpov15 (changes requested):

index.d.ts
michaln-q: Sure, fixed.

vkarpov15 (approved):
Thanks :+1:

PR #10190, opened 2021-04-29
Fix: broken link in document.js

joostdecock:
This links to an anchor, but the target is on a different page, so this link is broken.

I checked the other links to see how to link, but there are three different approaches used throughout the docs:

 - Relative links: `/docs/api/model.html#model_Model.replaceOne`
 - http links: `http://mongoosejs.com/docs/api/model.html#model_Model.replaceOne`
 - https links: `https://mongoosejs.com/docs/api/model.html#model_Model.replaceOne`

I opted for the https link (`https://mongoosejs.com/docs/api/model.html#model_Model.replaceOne`) since that seemed like the best choice to me.

**Summary**

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -805,7 +805,7 @@ Document.prototype.updateOne = function updateOne(doc, options, callback) {
  *
  * ####Valid options:
  *
- *  - same as in [Model.replaceOne](#model_Model.replaceOne)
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #10198, opened 2021-05-03
Added Document.prototype.parent to type declaration

HunterKohler:
**Summary**

The type declaration file was missing `Document.prototype.parent()` this is a method as can be seen here 

https://github.com/Automattic/mongoose/blob/80245edf5aeab304411681035f7f4153dbd2b692/lib/document.js#L3725

It is reflected in the docs [here](https://mongoosejs.com/docs/api/document.html#document_Document-parent).

I just added: the definition, and the description from the documentation. Made sure to put it in alphabetical order with the rest.

**Note**

You still need to keep the type declarations for `EmbeddedDocument.prototype.parent()` and `Subdocument.prototype.parent()` because they narrow the return type from `Document | undefined` to `Document`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -543,6 +543,12 @@ declare module 'mongoose' {
     overwrite(obj: DocumentDefinition<this>): this;
 
     /**
+     * If this document is a subdocument or populated document, returns the 
+     * document's parent. Returns undefined otherwise.
+     */
+    parent(): Document | undefined;
+
+    /**
      * Populates document references, executing the `callback` when complete.
      * If you want to use promises instead, use this function with
      * [`execPopulate()`](#document_Document-execPopulate).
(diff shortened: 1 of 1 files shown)


vkarpov15 (changes requested):
Unfortunately we can't merge this as written because it would break any code that uses `parent` as a document property along with the `extends Document` pattern:

```javascript
interface ITest extends Document {
  parent?: Types.ObjectId;
}
```

This is a TypeScript limitation: no support for child classes overwriting base class properties.

Mongoose documents also have a [`$parent()` function](https://mongoosejs.com/docs/api/document.html#document_Document-$parent) that's an alias for `parent()` in cases where you can't use `parent()`. Can we add that to the TS definitions instead?

vkarpov15 (approved):

PR #10213, opened 2021-05-07
fix: (index.d.ts) don't require error value in pre/post hooks

michaln-q:
**Summary**

Basing on https://mongoosejs.com/docs/middleware.html

```
  // If you call `next()` with an argument, that argument is assumed to be
  // an error.
```  

In the scenario posted below executing just `return next()` is not possible due to TS errors;

```
const RandomSchema = new mongoose.Schema({
    id: {
        type: String,
        required: true,
        maxlength: 128,
        trim: true
    }
}, { _id: false });

RandomSchema.post('findOneAndUpdate', (doc, next) => {
  if (doc === null) {
    return next();
  }
  // do sth
} 
```

The only way is to pass `null` as `next` param, but it isn't intuitive.


vkarpov15 (approved):
Nice, thanks :+1:

PR #10238, opened 2021-05-12
added onlyVirtuals to loadClass

IslandRhythms:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1131,7 +1131,7 @@ declare module 'mongoose' {
      * [methods](http://mongoosejs.com/docs/guide.html#methods).
      */
     // eslint-disable-next-line @typescript-eslint/ban-types
-    loadClass(model: Function, onlyVirtuals: boolean): this;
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

index.d.ts
```diff
@@ -1131,7 +1131,7 @@ declare module 'mongoose' {
      * [methods](http://mongoosejs.com/docs/guide.html#methods).
      */
     // eslint-disable-next-line @typescript-eslint/ban-types
-    loadClass(model: Function): this;
+    loadClass(model: Function, onlyVirtuals: boolean): this;
```
vkarpov15, line 1134: Actually, on second thought, `onlyVirtuals` should be optional, so this should be `onlyVirtuals?: boolean`.

`onlyVirtuals?: boolean` means you can call `loadClass()` without specifying the `onlyVirtuals` parameter.

vkarpov15 (changes requested):

vkarpov15 (approved):

PR #10245, opened 2021-05-14
Feat 7287

IslandRhythms:
I feel like this could be better but I'm not sure how.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -70,6 +70,7 @@ const specialProperties = utils.specialProperties;
  */
 
 function Document(obj, fields, skipId, options) {
+  console.log('doc', new Error().stack)
   if (typeof skipId === 'object' && skipId != null) {
     options = skipId;
     skipId = options.skipId;
@@ -96,7 +97,7 @@ function Document(obj, fields, skipId, options) {
   this.$__.$options = options || {};
   this.$locals = {};
   this.$op = null;
-
+  console.log(this);
   if (obj != null && typeof obj !== 'object') {
     throw new ObjectParameterError(obj, 'obj', 'Document');
   }
@@ -136,6 +137,7 @@ function Document(obj, fields, skipId, options) {
     // By default, defaults get applied **before** setting initial values
     // Re: gh-6155
     if (defaults) {
+      console.log('z')
       $__applyDefaults(this, fields, skipId, exclude, hasIncludedChildren, true, {
         isNew: this.isNew
       });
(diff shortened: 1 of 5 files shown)


lib/document.js
```diff
@@ -167,6 +165,10 @@ function Document(obj, fields, skipId, options) {
     $__applyDefaults(this, fields, skipId, exclude, hasIncludedChildren, false, options.skipDefaults, {
       isNew: this.isNew
     });
+  } else if (!defaults) {
+    EventEmitter.prototype.once.call(this, 'init', () => {
```
vkarpov15, line 169: I'm not sure why you need this - if `defaults` correctly propagates down from the query's `_mongooseOptions`, you shouldn't need this. Why are defaults still getting set if `defaults` is false?

lib/document.js
```diff
@@ -454,6 +455,22 @@ function $__applyDefaults(doc, fields, skipId, exclude, hasIncludedChildren, isB
...
+/**
+ * Checks which paths have defaults
+ * to then delete them from the doc
+ *
+ * @param {Object} doc
+ */
+function $__unapplyDefaults(doc) {
+  let values = Object.values(doc.$__schema.paths);
+  for (let i = 0; i < values.length; i++) {
+    if (values[i].defaultValue && values[i].instance != 'ObjectID') {
```
vkarpov15, line 467: You should use`$isDefault()` to check if a value is set to its default value.

vkarpov15: https://mongoosejs.com/docs/api/document.html#document_Document-$isDefault

lib/query.js
```diff
@@ -2950,7 +2956,7 @@ function completeOne(model, doc, res, options, fields, userProvidedFields, pop,
     return null;
   }
 
-  const casted = helpers.createModel(model, doc, fields, userProvidedFields);
+  const casted = helpers.createModel(model, doc, fields, userProvidedFields, options);
```
vkarpov15, line 2959: This generally looks right, but maybe should be `this._mongooseOptions` instead of `options`?

vkarpov15 (changes requested):

lib/document.js
IslandRhythms: I tried negating the defaults if statement to see if that would solve it but to no avail. The defaults are already set because when mongoose creates the document, if there are values with defaults it inserts them with the default values. So the only way to get rid of them it seems is to remove them after the fact.

lib/document.js
IslandRhythms: Object.values is unfortunately incompatible with older versions of node so I have updated the current method.

lib/query.js
IslandRhythms: for some reason it throws an error stating cannot read property mongooseOptions of undefined

lib/document.js
IslandRhythms: will do

IslandRhythms: I'm not exactly sure how to use it. I do `doc.$isDefault('name')` which returns false even though it should return true.

vkarpov15 (approved):
Nice :+1:

PR #10248, opened 2021-05-15
fix(typescript): id to be string

hannojg:
**Summary**

When documentation says that its the string version then its type should be string. Otherwise we would need to case the id to a string again, which feels unneccessary.

**Examples**

```typescript
const model = new Model({});
const id = model.id;

if (id === someOtherString) {
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -487,7 +487,7 @@ declare module 'mongoose' {
     getChanges(): UpdateQuery<this>;
 
     /** The string version of this documents _id. */
-    id?: any;
+    id?: string;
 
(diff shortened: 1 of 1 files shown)


vkarpov15 (commented):
Unfortunately we can't merge this, because this would break any code that overwrites Mongoose's default `id` getter. For example, it is perfectly ok to define `id` as a number:

```javascript
const schema = new Schema({ id: Number });
```

But then anyone that uses `extends Document` would get a compiler error.

See further discussion on #9773, which this PR is essentially a dupe of.

PR #10251, opened 2021-05-15
ts: fix query async iterator definition

### LINKED ISSUE #10250: It is not possible to use for-await with queries in typescript
a *bug*

It is not possible to use `for await (const doc in query)` syntax with `Query` instances.

The following does not compile:
```
import { Schema, model } from "mongoose";
interface Test {
  test: string;
}
const schema = new Schema<Test>({
  test: { type: String, required: true },
});
const TestModel = model<Test>("Test", schema);
async function main() {
  const query = TestModel.find({});
  for await (const test of query) {
    console.log(test);
  }
}
```

My `tsconfig.json`:
```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2019",
    "lib": [
      "dom",
      "es2019",
      "es2020.bigint",
      "es2020.string",
      "es2020.symbol.wellknown",
    ],
    "esModuleInterop": true,
    "strict": true,
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "sourceMap": true,
    "outDir": "dist",
    "baseUrl": ".",
    "jsx": "react"
  },
  "include": [
    "src/**/*.ts*",
  ]
}
```

The sample code should compile

- Mongoose v5.12.9
- Node.js v12.22.1
- MongoDB v4.4.6
- TypeScript v4.2.4

borfig:
Fixes #10250

1. The async iterator is supposed to be a method and not a property;
2. According to underlying implementation, it returns an asyncIterator
   and not a query cursor.

**Summary**

This PR addresses #10250

**Examples**

The following code now compiles:
```
import { Schema, model } from "mongoose";
interface Test {
  test: string;
}
const schema = new Schema<Test>({
  test: { type: String, required: true },
});
const TestModel = model<Test>("Test", schema);
async function main() {
  const query = TestModel.find({});
  for await (const test of query) {
    console.log(test);
  }
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1876,7 +1876,7 @@ declare module 'mongoose' {
      * A QueryCursor exposes a Streams3 interface, as well as a `.next()` function.
      * This is equivalent to calling `.cursor()` with no arguments.
      */
-    [Symbol.asyncIterator]: QueryCursor<DocType>;
+    [Symbol.asyncIterator](): AsyncIterableIterator<DocType>;
 
     /** Executes the query */
     exec(): Promise<ResultType>;


vkarpov15 (approved):
Great, thanks! :+1:

PR #10253, opened 2021-05-15
ts: strictly typed query cursor async iterator

### LINKED ISSUE #10252: Strictly typed Query Cursor async iterator
IMHO, it is a *feature*.

When using `for await` with query cursors, the returned objects are of type `any`, which can hide possible bugs.

The following *buggy* TypeScript code compiles, and the bug will be discovered only in runtime:
```
import { Schema, model } from "mongoose";
interface Test {
  test: string;
}
const schema = new Schema<Test>({
  test: { type: String, required: true },
});
const TestModel = model<Test>("Test", schema);
async function main() {
  const cursor = TestModel.find({}).cursor();
  for await (const test of cursor) {
    console.log(test.tedt); // intentional bug
  }
}
main()
```

My `tsconfig.json`:
```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2019",
    "lib": [
      "dom",
      "es2019",
      "es2020.bigint",
      "es2020.string",
      "es2020.symbol.wellknown",
    ],
    "esModuleInterop": true,
    "strict": true,
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "sourceMap": true,
    "outDir": "dist",
    "baseUrl": ".",
    "jsx": "react"
  },
  "include": [
    "src/**/*.ts*",
  ]
}
```

I am proposing to change the `QueryCursor`'s `[Symbol.asyncIterator]` to return a strictly-typed `AsyncIterableIterator<DocType>`, instead of the inherited `AsyncIterableIterator<any>`

- Mongoose v5.12.9
- Node.js v12.22.1
- MongoDB v4.4.6
- TypeScript v4.2.4

borfig:
Closes #10252

**Summary**

This PR implements #10252 

**Examples**

Currently, the following TypeScript code compiles, with the bug in the code:
```
import { Schema, model } from "mongoose";
interface Test {
  test: string;
}
const schema = new Schema<Test>({
  test: { type: String, required: true },
});
const TestModel = model<Test>("Test", schema);
async function main() {
  const cursor = TestModel.find({}).cursor();
  for await (const test of cursor) {
    console.log(test.tedt); // intentional bug
  }
}
main()
```
The proposed change makes the above not compile

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2368,6 +2368,8 @@ declare module 'mongoose' {
     T;
 
   class QueryCursor<DocType extends Document> extends stream.Readable {
+    [Symbol.asyncIterator](): AsyncIterableIterator<DocType>;
+
     /**
      * Adds a [cursor flag](http://mongodb.github.io/node-mongodb-native/2.2/api/Cursor.html#addCursorFlag).
      * Useful for setting the `noCursorTimeout` and `tailable` flags.


vkarpov15 (approved):
Thanks :+1:

PR #10265, opened 2021-05-18
upgrade mongodb to v3.6.7

AbdelrahmanHafez:


lib/connection.js
```diff
@@ -910,7 +910,7 @@ function _setClient(conn, client, options, dbName) {
       });
       client.on('serverDescriptionChanged', ev => {
         const newDescription = ev.newDescription;
-        if (newDescription.type === 'Standalone') {
+        if (newDescription.type === 'Standalone' || newDescription.type === 'Unknown') {
```
vkarpov15, line 913: This is not likely to be the correct fix here, because if the topology description is "Unknown" we shouldn't consider that as "reconnected."  Is this to try to work around the 'timeout' issue in the first commit's test failures?

vkarpov15 (changes requested):

lib/connection.js
vkarpov15: @AbdelrahmanHafez I think I identified the issue and opened up a PR in the MongoDB driver. We'll see what they say, I'm not sure whether this was an intentional change or not.

vkarpov15 (commented):
I upgraded to mongodb@3.6.8 in master, so this PR is no longer necessary. Thanks :+1:

Issue #10200, opened 2021-05-04
[Bug] casting deeply nested paths does not work with `updateOne` but works with `create`


AbdelrahmanHafez:
Passing deeply nested paths on `updateOne` should convert the path to a deeply structured object, which works with new Model(...) and Model.create(...), but does not work with `Model.updateOne(...)` and `Model.bulkWrite(...)`.

Repro script:
### 10200.js
```js
describe('converts dot separated paths to nested structure', () => {
  it('works with new Model(...)', () => { // passes in 5.12.7
    const Payment = getPaymentModel();
    const paymentPOJO = getPaymentWithDotSeparatedPaths();
    const payment = new Payment(paymentPOJO);
    assertDocumentStructure(payment.toObject());
  });
  it('works with Model.create(...)', () => { // passes in 5.12.7
    return co(function*() {
      const Payment = getPaymentModel();
      const paymentPOJO = getPaymentWithDotSeparatedPaths();
      const payment = yield Payment.create(paymentPOJO);
      assertDocumentStructure(payment);
    });
  });
  it('works with Model.updateOne(...)', () => { // fails in 5.12.7
    return co(function*() {
      const User = getPaymentModel();
      const userPOJO = getPaymentWithDotSeparatedPaths();

      const emptyUser = yield User.create({});
      yield User.updateOne({ _id: emptyUser._id }, userPOJO);
      const user = yield User.findOne({ _id: emptyUser._id }).lean();

      assertDocumentStructure(user);
    });
  });
  it('works with Model.bulkWrite(...)', () => { // fails in 5.12.7
    return co(function*() {
      const Payment = getPaymentModel();
      const paymentPOJO = getPaymentWithDotSeparatedPaths();

      const emptyPayment = yield Payment.create({});

      yield Payment.bulkWrite([
        { updateOne: { filter: { _id: emptyPayment._id }, update: paymentPOJO } }
      ]);
      const payment = yield Payment.findOne({ _id: emptyPayment._id }).lean();

      assertDocumentStructure(payment);
    });
  });


  function getPaymentModel() {
    const paymentSchema = new Schema({
      paymentFor: String,
      externalServiceResponse: {
        id: String,
        resultDetails: {
          clearingInstituteName: String,
          transaction: {
            receipt: String,
            authorizationCode: String,
            acquirer: { settlementDate: String }
          },
          response: { acquirerCode: String, acquirerMessage: String },
          authorizationResponse: { stan: String },
[...]

PR #10266, opened 2021-05-18
test(model): add more test cases re #10200

AbdelrahmanHafez:
This PR adds more test cases to verify the behavior in #10200 works with `new Model(...)` as well as `Model.create(...)`.
It also verifies that nested fields that do not exist in the schema are removed by default.


vkarpov15 (approved):
Thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

Issue #9105, opened 2020-06-07
Updating nested object with timestamps causes ConflictingUpdateOperators error


tHBp:
**Do you want to request a *feature* or report a *bug*?** Bug

**What is the current behavior?** If the `timestamps` property is targeting a nested object in the schema, any updates to the aforementioned object fail with `ConflictingUpdateOperators` error.

**If the current behavior is a bug, please provide the steps to reproduce.** Code attached below.

**What is the expected behavior?** It should be able to update the object

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

| Software | Version |
|-|-|
| NodeJS | 12.18.0 |
| mongoose | 5.9.18 |
| mongodb | 4.2.6 |

<details>
<summary>Minimal code to reproduce bug</summary>

```js
const mongoose = require("mongoose");

const SampleSchema = new mongoose.Schema(
  {
    someComplexObject: {
      someProperty: String,
    },
  },
  {
    timestamps: {
      updatedAt: "someComplexObject.timestamp",
    },
  }
);

const model = mongoose.model("SampleModel", SampleSchema);

(async () => {
  // Connect to database
  await mongoose.connect("mongodb://localhost:27017/dummy", {
    useNewUrlParser: true,
    useUnifiedTopology: true,
    useFindAndModify: false,
  });

  // Insert a dummy value to the db
  const createResult = await model.create({
    someComplexObject: {
      someProperty: "someValue",
    },
  });

  try {
   // Try to update the same nested object
    const updateResult = await model.findByIdAndUpdate(createResult.id, {
      someComplexObject: {
        someProperty: "updatedString",
      },
    });
  // Should log the result here
    console.log("Result", updateResult);
  } catch (error) {
  // Instead throws the error here
    console.error(error);
  } finally {
    await mongoose.disconnect();
  }
})();
```

</details>

<details>
<summary>Error</summary>

```
MongoError: Updating the path 'someComplexObject' would create a conflict at 'someComplexObject'
    at MessageStream.messageHandler (/mongoose-issue/node_modules/mongodb/lib/cmap/connection.js:261:20)
    at MessageStream.emit (events.js:315:20)
    at processIncomingData (/mongoose-issue/node_modules/mongodb/lib/cmap/message_stream.js:144:12)
    at MessageStream._write (/mongoose-issue/node_modules/mongodb/lib/cmap/message_stream.js:42:5)
    at doWrite (_stream_writable.js:403:12)
[...]

PR #10267, opened 2021-05-18
enhancement: timestamps createdAt property immutable by default

IslandRhythms:
So there is a failing test, however I'm not sure If the test needs to be fixed or my solution needs to be changed. The test is gh-9105

### DIFF
--- a/lib/helpers/timestamps/setupTimestamps.js
+++ b/lib/helpers/timestamps/setupTimestamps.js
@@ -32,7 +32,7 @@ module.exports = function setupTimestamps(schema, timestamps) {
   }
 
   if (createdAt && !schema.paths[createdAt]) {
-    schemaAdditions[createdAt] = Date;
+    schemaAdditions[createdAt] = {type:Date, immutable:true};
   }
 
   schema.add(schemaAdditions);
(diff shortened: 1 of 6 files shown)


AbdelrahmanHafez (changes requested):
I'd say that such change is a breaking change and should not be included in a minor/patch release.
I think it's fine to set `createdAt` to be `immutable` by default, but if we choose to go that way, we need to move it to v6.0.

Also, would be neat to offer a global option to set `createdAt` to `false` for those who choose to.

vkarpov15:
@AbdelrahmanHafez you're right about this being potentially backwards breaking, I'll merge this into 6.0 just to be safe.

Any ideas for syntax re: setting `createdAt` to `false` globally? Maybe `mongoose.set('timestamps.createdAt', false)`?

AbdelrahmanHafez:
@vkarpov15 
I'd go for `mongoose.set('timestamps.createdAt.immutable', false)`, otherwise it can mislead people to think that this option sets timestamps to *not* add a `createdAt` field.

vkarpov15:
@AbdelrahmanHafez would having an immutable `createdAt` break your code?

AbdelrahmanHafez:
@vkarpov15 
No, but I am anticipating that someone out there is updating the value of createdAt, and it would really make their lives difficult.

We can ship this as an option to set immutable globally on createdAt, to be false by default in v5, then change it to true by default in v6.

miladezzat (approved):

PR #10269, opened 2021-05-19
fix typo: preferences

sanjib:
Fixes a small documentation typo for the function Query.prototype.read

before: preferrences 

after: preferences


vkarpov15 (approved):
Thanks :+1:

PR #10276, opened 2021-05-20
`pathsToSkip` option for `validate()` and `validateSync()`

IslandRhythms:

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2460,7 +2460,10 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
   if (typeof pathsToValidate === 'string') {
     pathsToValidate = pathsToValidate.split(' ');
   }
-  if (pathsToSkip) {
+  if (Array.isArray(pathsToValidate)) {
+    paths = _handlePathsToValidate(paths, pathsToValidate);
+  }
+  else if (pathsToSkip) {
     // make a new variable that is paths intersection !options.pathsToSkip
     const valPaths = new Set(paths);
     for (const path of pathsToSkip) {
@@ -2469,10 +2472,7 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
       }
     }
     // then send that to be validated
-    paths = _handlePathsToValidate(paths, valPaths);
-  }
-  if (Array.isArray(pathsToValidate)) {
-    paths = _handlePathsToValidate(paths, pathsToValidate);
+    paths = _handlePathsToSkip(paths, valPaths);
   }
   if (paths.length === 0) {
     return process.nextTick(function() {
@@ -2591,6 +2591,21 @@ function _handlePathsToValidate(paths, pathsToValidate) {
       ret.push(path);
     } else if (parentPaths.has(path)) {
       ret.push(parentPaths.get(path));
+    }
+  }
+  return ret;
+}
+
+/*!
+ * ignore
+ */
+
+function _handlePathsToSkip(paths, valPaths) {
+  const _valPaths = new Set(valPaths);
+  const ret = [];
+  for (const path of paths) {
+    if (_valPaths.has(path)) {
+      ret.push(path);
     }
   }
   return ret;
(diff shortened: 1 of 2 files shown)


lib/document.js
```diff
@@ -2455,10 +2459,20 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
     pathDetails[0].filter((path) => this.isModified(path)) :
     pathDetails[0];
   const skipSchemaValidators = pathDetails[1];
-
   if (typeof pathsToValidate === 'string') {
     pathsToValidate = pathsToValidate.split(' ');
   }
+  if (pathsToSkip) {
+    // make a new variable that is paths intersection !options.pathsToSkip
+    let valPaths = new Set(paths);
```
vkarpov15, line 2467: I'd recommend making `pathsToSkip` a set and [iterating over `paths` with a `for/of`](https://masteringjs.io/tutorials/fundamentals/array-iterate) instead of making `paths` a set and iterating over `pathsToSkip`. That's likely cleaner

lib/document.js
```diff
@@ -2406,6 +2406,10 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
       (typeof options === 'object') &&
       ('validateModifiedOnly' in options);
 
+  const pathsToSkip = options &&
+      (typeof options === 'object') &&
```
vkarpov15, line 2410: Careful, remember you wrote about [typeof object](https://masteringjs.io/tutorials/fundamentals/typeof-object) and why using this to check if a value is an object is dangerous.

I'd instead do `const pathsToSkip = get(options, 'pathsToSkip', null)`

lib/document.js
vkarpov15: `paths = paths.filter(p => !pathsToSkip.has(p))`, see [filter arrays in JavaScript](https://masteringjs.io/tutorials/fundamentals/filter).

IslandRhythms: Ok I see now but `_handlePathsToValidate` takes all the paths `paths` and then the paths to validate `valPaths` as the arguments. if I do that re-assignment if will mess up the functino execution.

lib/document.js
```diff
@@ -2584,6 +2596,21 @@ function _handlePathsToValidate(paths, pathsToValidate) {
   return ret;
 }
 
+/*!
+ * ignore
+ */
+
+function _handlePathsToSkip(paths, valPaths) {
+  const _valPaths = new Set(valPaths);
```
vkarpov15, line 2604: Why are you creating a set when `valPaths` is already a set?

lib/document.js
```diff
@@ -2455,13 +2457,23 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
...
+  else if (pathsToSkip) {
+    // make a new variable that is paths intersection !options.pathsToSkip
+    const valPaths = new Set(paths);
+    for (const path of pathsToSkip) {
+      if (valPaths.has(path)) {
+        valPaths.delete(path);
+      }
+    }
+    // then send that to be validated
+    paths = _handlePathsToSkip(paths, valPaths);
```
vkarpov15, line 2475: Why is the separate `valPaths` set here? Why not just pass `pathsToSkip` to `_handlePathsToSkip()`?

lib/document.js
```diff
@@ -2455,13 +2457,23 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
...
   }
   if (Array.isArray(pathsToValidate)) {
     paths = _handlePathsToValidate(paths, pathsToValidate);
   }
+  else if (pathsToSkip) {
+    // make a new variable that is paths intersection !options.pathsToSkip
+    const valPaths = new Set(paths);
+    for (const path of pathsToSkip) {
+      if (valPaths.has(path)) {
+        valPaths.delete(path);
```
vkarpov15, line 2471: I think you missed the point here. The point is to move this logic into `_handlePathsToSkip()`, convert `pathsToSkip` into a set, and use `paths = paths.filter(p => !pathsToSkip.has(p))`

vkarpov15 (changes requested):

vkarpov15 (approved):
Thanks :+1:

PR #10288, opened 2021-05-25
`.map`'s `ResultType` must be an array if used over an array

quantumsheep:
**Summary**
When using `.map()`, I noticed that the return type wasn't right.

```js
users.find().map((user) => user); // Query<UserDocument, UserDocument>
```

or it should be `Query<UserDocument[], UserDocument>`.

This PR fixes that issue while keeping non-array queries capable of using `.map()`.


vkarpov15 (approved):
Thanks :+1:

PR #10326, opened 2021-06-04
fix

IslandRhythms:


vkarpov15 (changes requested):
Can you add a function to test this in `test/typescript/document.ts`?

test/typescript/document.ts
```diff
@@ -15,4 +15,10 @@ void async function main() {
 
   const p: Promise<ITest> = doc.remove();
   await p;
+
+}();
+
+void async function run() {
+  const user = new Test({ name: {}, address: {} });
+  user.validateSync();
```
vkarpov15, line 23: This doesn't test anything. You'll want to do `const error = user.validateSync()`, and then try accessing `error.errors.address.errors`, which should fail without the fix in this PR.

vkarpov15 (changes requested):
Test doesn't do anything. In general, if you write a test for a change, the test should fail without your change.

test/typescript/document.ts
IslandRhythms: Doesn't main.test.js check for errors though? if something in document.ts throws an error, doens't main.test.js store it and do an assertion if there are no errors?

vkarpov15 (approved):

PR #10343, opened 2021-06-11
Models should follow the schema (Please god I want this for my sanity)

ShadiestGoat:
**Summary**

See, when I create a model, there is never a time where I ever have to create a model for a schema that isn't defined. This is basically to make it so that you NEED a proper document for the proper schema.
Thats it

**Examples**

```ts
new MyModel({must be a schema})
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -619,7 +619,7 @@ declare module 'mongoose' {
   export const Model: Model<any>;
   // eslint-disable-next-line no-undef
   interface Model<T, TQueryHelpers = {}, TMethods = {}> extends NodeJS.EventEmitter, AcceptsDiscriminator {
-    new(doc?: T | any): EnforceDocument<T, TMethods>;
+    new(doc?: T): EnforceDocument<T, TMethods>;
 
     aggregate<R = any>(pipeline?: any[]): Aggregate<Array<R>>;
     // eslint-disable-next-line @typescript-eslint/ban-types


ShadiestGoat:
Any checks that weren't passed are irrelevant

vkarpov15:
There's no way we're going to merge this unfortunately, it is perfectly valid to pass any object to a model constructor. Changing that would be a massive backwards breaking change and be completely opposed to how Mongoose is meant to be used.

Why is this change good for your sanity?

ShadiestGoat:
The reason for this is for autocomplete, mostly. I mean what'd be the point of typescript here if you its just any?

Wouldn't it be better then to just set the default type to be any, rather than having it or any? Cause putting the type to be `X | any` means it'd just be `any`, meaning typing is useless there. 
This basically sets it be any by default, but if a schema is present, only the schema will be passed through. So, if you really wanted to pass any object through, then you could still do that, just don't define the type to be that schema...

ShadiestGoat:
Imo this should be reconsidered

vkarpov15:
How about this instead?

```typescript
new(doc?: T): EnforceDocument<T, TMethods>;
new(doc?: any): EnforceDocument<T, TMethods>;
```

That might help.

ShadiestGoat:
That the exact same thing? 
T is any by default, so if you want to pass anything through there, just don't define T to be anything:
```ts
//explicitly state it's of type any
const mmodel:Model<any> = model('Foo', Bar)
//any by default (assuming SchemaDefinition is empty)
const mmodel2 = model('Foo2', Bar2)

//or, you could just ask ts to ignore
const mmodel:Model<FooBar> = model('Foo', Bar)
//@ts-ignore
new mmodel({wrongType: "Barrr"})
```

vkarpov15:
I think it might be different for VS code autocomplete, because it would show autocomplete for `T` first. But if that doesn't help, there's nothing we can do here because Mongoose is meant to validate potentially untrusted data coming in to the `Model()` constructor, so it needs to accept `any` one way or another.

PR #10360, opened 2021-06-15
fix(model): make Model.validate use object under validation as context by default

### LINKED ISSUE #10346: Validate doesn't work with conditional required fields
Bug.

Conditional `required` fields don't throw when using `Model.validate()`.

```js

run().catch(console.error);

async function run () {
    age: { type: Number, required () {return this.name === 'John';} }
  });

  const err = await User.validate({ name: 'John' }).then(() => null, err => err);
  assert.ok(err);
}
```

Expects method to throw a validaiton error path `age` is requried.

Node.js: 14.15.5
Mongoose: 5.12.13
MongoDB:  4.4.3

AbdelrahmanHafez:
fixes #10346


lib/model.js
```diff
@@ -4062,9 +4062,9 @@ Model.aggregate = function aggregate(pipeline, callback) {
  */
 
 Model.validate = function validate(obj, pathsToValidate, context, callback) {
-  if (context === void 0 && obj instanceof Model) {
-    // For convenience, if we're validating a document, make `context` default to
-    // the model so users don't have to always pass `context`, re: gh-10132.
+  if (context === void 0) {
```
vkarpov15, line 4065: I think we should make this only happen if the user doesn't specify `context`. So either `arguments.length < 3`, or `arguments.length === 3 && typeof arguments[2] === 'function'`.

vkarpov15 (commented):
I agree that this is the right way to go. Using `undefined` by default is a bit messy. Maybe we should release this in 5.13 though, might be a bit heavy for a patch release.

AbdelrahmanHafez:
Sounds good to me.

AbdelrahmanHafez:
@vkarpov15 
Regarding the `context` change, and how we're trying to guess what parameter is what. There's this pattern [RORO](https://www.freecodecamp.org/news/elegant-patterns-in-modern-javascript-roro-be01e7669cbd/), receive an object, return an object.
It makes function parameters similar to python's named parameters, so that you won't have to guess anymore. We've been using those everywhere around our codebases and never went back.

We've also implemented a layer over mongoose to make mongoose's API comply with that pattern.
That pattern has been specifically helpful in testing, because stubbing chainable APIs can be really difficult.
```js
    Comment.dal.find({
      filter: { userId: user._id },
      skip,
      limit,
      sort: sortBy,
      populate: { path: 'author', select: 'name username email' },
      lean: true
    }),
```

Does converting mongoose APIs to take objects with named parameters like that seem like an idea you'd pursue someday? Should I open an issue so we can discuss it further?

vkarpov15:
Yeah using objects to simulate [named parameters in JavaScript](https://masteringjs.io/tutorials/fundamentals/parameters) is a neat pattern, and I'm not opposed to potentially supporting it in the future.

miladezzat (approved):

vkarpov15 (approved):
Thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

PR #10362, opened 2021-06-15
Feat 10316: Can now do a dryRun of syncIndexes using Model.diffIndexes()

IslandRhythms:

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1419,7 +1419,8 @@ Model.syncIndexes = function syncIndexes(options, callback) {
  * 
  * @param {Object} options not used at all.
  * @param {Function} callback optional callback
- * @returns {Object} {toDrop, toCreate} Indexes that would be dropped in mongodb and indexes that would be created in mongodb.
+ * @returns {Promise} which containts an object, {toDrop, toCreate}, which
+ * are indexes that would be dropped in mongodb and indexes that would be created in mongodb.
  */
 
 Model.diffIndexes = function diffIndexes(options, callback) {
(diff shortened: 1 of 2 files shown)


test/model.indexes.test.js
```diff
@@ -683,5 +683,17 @@ describe('model', function() {
         yield User.collection.drop();
       });
     });
+    it('should do a dryRun feat-10316', function() {
+      return co(function*() {
+        const userSchema = new mongoose.Schema({ username: String }, { password: String }, { email: String });
+        userSchema.index({ password: 1 });
+        userSchema.index({ email: 1 });
+        const User = db.model('Upson', userSchema);
```
vkarpov15, line 691: Add a `yield User.init()` here to wait for all indexes to finish building

lib/model.js
```diff
@@ -1411,6 +1411,42 @@ Model.syncIndexes = function syncIndexes(options, callback) {
...
+          continue;
+        }
+
+        for (const schemaIndex of schemaIndexes) {
+          const key = schemaIndex[0];
+          const options = _decorateDiscriminatorIndexOptions(this,
+            utils.clone(schemaIndex[1]));
+          if (isIndexEqual(key, options, index)) {
+            found = true;
+          } else {
```
vkarpov15, line 1436: You may end up adding the same index to `toCreate` multiple times. If you have 3 indexes in MongoDB, you'll execute the `schemaIndexes()` loop 3 times and end up adding the same schema index twice each time.

You would likely need to write a separate `for` loop to correctly compute `toCreate`

vkarpov15 (changes requested):

IslandRhythms:
Do you want me to go back in and add the comments to describe what diffIndexes() does or do you got it?

vkarpov15:
@IslandRhythms please add comments, that would be very helpful :+1:

vkarpov15 (approved):
Good work, this looks great :+1:

PR #10365, opened 2021-06-16
Update transactions.md

JimLynchCodes:
Maybe this is addition I made is a very obvious thing to many people... but as someone who landed on this page from google I can say having this line right there in the docs page would have saved me some time... 😉


vkarpov15 (approved):
Thanks :+1:

Issue #10345, opened 2021-06-12
LeanDocument contain id field


tbhaxor:
**Do you want to request a *feature* or report a *bug*?**

Issue #10282 continued

**bug**

**What is the current behavior?**

LeanDocument contain `.id` field

**If the current behavior is a bug, please provide the steps to reproduce.**

1. npm init -y
2. npm i mongoose @types/node
3. Create a user model (https://gist.github.com/tbhaxor/1deb83bb4d8a962f7959a3ac41487cd9)
4. Import typings (https://gist.github.com/tbhaxor/367e33692470c26cab5fec5cef749d91)
5. Import user model in another file and lean document `User.findOne().lean()`

![Screenshot_20210612_163817](https://user-images.githubusercontent.com/28386721/121773890-99c39300-cb9c-11eb-93c4-667b6b754e91.png)

**What is the expected behavior?**

Id should not exits

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Mongoose: 5.12.13

_This problem is with mongoose typescript, mongodb and node version are not required_

IslandRhythms:
From the closed issue:
> 
> 
> I looked into this a bit more and there isn't much we can do about this. The reason is that your document can have a user-defined `id` type, and there's no way for `LeanDocument` to conditionally strip out `id` only if `id` is inherited from `Document` as opposed to user-defined.
> 
> This is one of the reasons why our [TypeScript docs](https://mongoosejs.com/docs/typescript.html) recommend not using `extends Document`:
> 
> ```ts
> // We recommend replacing this:
> interface Store extends Document { ... }
> 
> // With this:
> interface Store { ... }
> ```
> 
> Makes it much easier for Mongoose to infer the correct type when using `lean()`.

tbhaxor:
@IslandRhythms I know typescript for a long time. I am not using a document that is why I reopened this

Please give me the mongoose typescript repo details. I will open a pull request for the fix, in case your team didn't manage to fix this issue earlier

tbhaxor:
@IslandRhythms I am not using Document

tbhaxor:
@IslandRhythms @vkarpov15 Yours typings are not working. Please check this ~3 mins video

https://www.loom.com/share/9b90007366fb4fc49c8f9843e1db4b02

vkarpov15:
01c25ee improves this behavior for `toObject()` and `toJSON()`, we'll see what we can do for `lean()`

Issue #10282, opened 2021-05-24
LeanDocument typings contain undefined property id in intellisense


tbhaxor:
**Do you want to request a *feature* or report a *bug*?**

**bug**

**What is the current behaviour?**

LeanedDocument exposes `id` property which indeed is not the case when you `.lean()` the document

![image](https://user-images.githubusercontent.com/28386721/119340535-443c4a80-bcb0-11eb-9b6e-4c5094550f89.png)

**If the current behaviour is a bug, please provide the steps to reproduce.**

1. Create a mongoose typescript project
2. lean any document
3. Try to access the `.id` property. You will see in the IntelliSense but when you console.log it, this would be undefined

```JSON
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "outDir": "./dist",
    "strict": true,
    "noImplicitAny": true,
    "typeRoots": ["./ctypings", "./node_modules/@types"],
    "types": ["express", "express-session", "process", "express-request-id"],
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true
  },
  "exclude": ["node_modules", "./node_modules", "./node_modules/*", "./node_modules/@types/node/index.d.ts"],
  "include": ["src/**/*.ts", "ctypings/**/*.ts"]
}
```

**What is the expected behaviour?**

Omit the `id` from LeanDocument

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

The problem is with `@types/mongoose`, version of that is 5.10.3

tbhaxor:
@IslandRhythms If i can know the repo, maybe I can help in this

tbhaxor:
The possible solution could be

```ts
Omit<T, keyof Document> & { _id?: Types.ObjectId, __v?: number }
```

vkarpov15:
I looked into this a bit more and there isn't much we can do about this. The reason is that your document can have a user-defined `id` type, and there's no way for `LeanDocument` to conditionally strip out `id` only if `id` is inherited from `Document` as opposed to user-defined.

This is one of the reasons why our [TypeScript docs](https://mongoosejs.com/docs/typescript.html) recommend not using `extends Document`:

```typescript
// We recommend replacing this:
interface Store extends Document { ... }

// With this:
interface Store { ... }
```

Makes it much easier for Mongoose to infer the correct type when using `lean()`.

PR #10371, opened 2021-06-18
Omit id field in lean document typings

tbhaxor:

**Summary**

Closes https://github.com/Automattic/mongoose/issues/10345

**Examples**

Screenshot_20210618_193841

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2437,7 +2437,7 @@ declare module 'mongoose' {
     T[K];
   };
 
-  export type LeanDocument<T> = Omit<_LeanDocument<T>, Exclude<keyof Document, '_id' | 'id' | '__v'> | '$isSingleNested'>;
+  export type LeanDocument<T> = Omit<_LeanDocument<T>, Exclude<keyof Document, '_id' | '__v'> | '$isSingleNested'>;
 
   export type LeanDocumentOrArray<T> = 0 extends (1 & T) ? T :
     T extends unknown[] ? LeanDocument<T[number]>[] :


AbdelrahmanHafez (approved):

vkarpov15:
Realistically I don't think we can merge this for the same reasons I described here: https://github.com/Automattic/mongoose/issues/10282#issuecomment-853103354 . Any document with a user-defined `id` property will lose the `id` property in the return value from `toObject()`

```typescript
import { model, Schema } from 'mongoose';

interface User {
  name: string;
  id: number;
}

const UserModel = model<User>('User', new Schema({ name: String, id: Number }));

const doc = new UserModel({ name: 'test', id: 42 });

const leanDoc = doc.toObject();
leanDoc.id = 43; // Property 'id' does not exist on type 'LeanDocument<User & Document<any, any>>'.
```

I think the better solution would be to pass `User` rather than `User & Document` internally to `LeanDocument`, and have `LeanDocument` avoid stripping document properties from types that don't `extends Document`. What do you think @tbhaxor ?

tbhaxor:
> pass `User` rather than `User & Document`

@vkarpov15  It seems reasonable to me

vkarpov15:
I think that should be up to the end user to add `_id` and `__v` to their document interface, if they choose to use them.

```typescript
interface User {
  _id: mongoose.Types.ObjectId;
  name: string;
}
```

You can disable `_id` and `versionKey` from the schema constructor options, so technically Mongoose can't know whether the document has `_id` and `__v` at compile time without the end user telling us so.

tbhaxor:
@vkarpov15 Again, your point made sense 😄

tbhaxor:
So, a Leaned Document **MUST** return the generic type passed

PR #10373, opened 2021-06-19
Added typings for near() in model aggregation

### LINKED ISSUE #10366: Property 'near' does not exist on type 'Aggregate<any[]>

Typescript Bug

There's a type error in v5.12.14 that says `Property 'near' does not exist on type 'Aggregate<any[]>'`

```js
ModelName.aggregate().near()
```

No Typescript error. Based on the [docs](https://mongoosejs.com/docs/api/aggregate.html#aggregate_Aggregate-near), the `.near` function is a part of Aggregate class

```
NodeJS: v12.13.0
mongoose: v5.12.14
```

tbhaxor:

**Summary**

FIxes #10366

**Examples**

Here is POC &dArr;
Screenshot_20210619_170401

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2562,6 +2562,12 @@ declare module 'mongoose' {
      */
     model(model: any): this;
 
+    /**
+     * Append a new $near operator to this aggregation pipeline
+     * @param arg $near operator contents
+     */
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

Issue #10367, opened 2021-06-17
[Feature] support named parameters


AbdelrahmanHafez:
re https://github.com/Automattic/mongoose/pull/10360#issuecomment-862245274
It would be nice if we support named parameters for mongoose methods.
It should be straightforward to implement, I'll start with the following:
* Model.find
* Model.findOne
* Model.updateOne
* Model.updateMany
* Model.countDocuments
* Model.validate
* Model.aggregate
* Model.exists

Then we can decide what other functions we can apply the pattern to.

swayamsingh2004:
I'd like to work on this. Since this covers multiple methods, I'll start by implementing named parameter support for Model.findOne as a first step, and we can decide on the rest after. Let me know if that scope works.

AbdelrahmanHafez:
Hi @swayamsingh2004, I'll probably be slow to review, but will get to it eventually. If that works, your plan sounds great.

swayamsingh2004:
While working on `Model.findOne`, I ran into a design question I'd like input on before writing code.

The challenge is detecting whether a call is using the old positional style or the new named-parameters style, since both pass a single object as the first argument in some cases:

```js
User.findOne({ email: "x" })                          // old style, plain filter
User.findOne({ filter: { email: "x" }, projection: "name" })  // new style, named params
```

A simple "does the object have a `filter` key" check breaks if someone's actual schema has a real field literally named `filter`:

```js
User.findOne({ filter: "some-value" })  // ambiguous: named-params wrapper, or a real filter on a `filter` field?
```

A couple of options I see:
1. Accept this as a known, documented limitation (collisions with a real `filter`/`projection`/`options` field are rare in practice).
2. Use a clearly distinct signature for the new style so it can never collide with a normal filter object (e.g. requiring a special marker, or a separate method name).

Wanted to check which direction you'd prefer before I commit to an implementation. If there's no strong preference, I'll go with option 1 as the default and move forward, just wanted to flag the tradeoff first.

AbdelrahmanHafez:
We need to go with option 2, we generally support arbitrary schema paths unless they conflict with mongo server, should continue to support here for consistency and better support.

swayamsingh2004:
Thanks for the clarification — that makes sense given the general philosophy of supporting arbitrary schema paths.

[...]

PR #10375, opened 2021-06-19
feat(document): add support for pathsToSkip in validate and validateSync

### LINKED ISSUE #10353: [Feature] Document#validate(...)  `pathsToSkip` option to exclude fields
Currently, we can make an inclusive list of fields that we want to validate using `pathsToValidate`.

It would be nice if we could have an option to skip specific fields `pathsToSkip`, and a user can either select fields to skip, or fields to validate, never both.

@vkarpov15 What do you think? I'd happily work on that if accepted.

AbdelrahmanHafez:
fixes #10353

This still doesn't add support for `pathsToSkip` for Model.validate(...)` because the current parameters are:
```js
Model.validate = function validate(obj, pathsToValidate, context, callback) { ... }
```
We can not add it as a property to any of those parameters, so I'll defer implementing it until #10367 is implemented, then it'll make our lives much easier to extend our API.


lib/document.js
```diff
@@ -2250,6 +2251,9 @@ Document.prototype.validate = function(pathsToValidate, options, callback) {
     options = pathsToValidate;
     pathsToValidate = null;
   }
+  if (options && typeof options.pathsToSkip === 'string') {
+    options.pathsToSkip = options.pathsToSkip.split(' ');
```
vkarpov15, line 2255: For perf reasons, let's make this `options.pathsToSkip = options.pathsToSkip.indexOf(' ') === -1 ? [options.pathsToSkip] : options.pathsToSkip.split(' ');` . `split()` is very slow unfortunately, even if there's nothing to split.

lib/document.js
```diff
@@ -2646,8 +2649,12 @@ Document.prototype.validateSync = function(pathsToValidate, options) {
     shouldValidateModifiedOnly = this.$__schema.options.validateModifiedOnly;
   }
 
+  let pathsToSkip = get(options, 'pathsToSkip');
+
   if (typeof pathsToValidate === 'string') {
     pathsToValidate = pathsToValidate.split(' ');
+  } else if (typeof pathsToSkip === 'string') {
+    pathsToSkip = pathsToSkip.split(' ');
```
vkarpov15, line 2657: Same for this one.

lib/document.js
```diff
@@ -2646,8 +2649,12 @@ Document.prototype.validateSync = function(pathsToValidate, options) {
     shouldValidateModifiedOnly = this.$__schema.options.validateModifiedOnly;
   }
 
+  let pathsToSkip = get(options, 'pathsToSkip');
```
vkarpov15, line 2652: Let's just do `pathsToSkip = options.pathsToSkip`, `get()` is handy for dotted paths and cases where we don't know whether `options` exists, but it also has some non-trivial performance overhead.

IslandRhythms (commented):
ditto everything Val said

vkarpov15 (approved):
Thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

PR #10385, opened 2021-06-23
Gh 10364: fixed typos in the docs

IslandRhythms:

### DIFF
--- a/docs/discriminators.md
+++ b/docs/discriminators.md
@@ -15,6 +15,16 @@ key (defaults to the model name). It returns a model whose schema
 is the union of the base schema and the discriminator schema.
 
 ```javascript
+const db = mongoose.createConnection('mongodb://localhost:27017/mongoose_test');
+const eventSchema = new mongoose.Schema({ time: Date });
+Event = db.model('_event', eventSchema);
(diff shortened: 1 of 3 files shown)


docs/discriminators.md
```diff
@@ -15,7 +15,17 @@ key (defaults to the model name). It returns a model whose schema
 is the union of the base schema and the discriminator schema.
 
 ```javascript
-[require:The `model.discriminator()` function]
+const db = mongoose.createConnection('mongodb://localhost:27017/mongoose_test');
```
vkarpov15, line 18: Why is this copy/pasted in here? Acquit should be pulling this code from the discriminators.test.js file.

vkarpov15 (changes requested):

docs/discriminators.md
IslandRhythms: I tried to get that section in using acquit but it wouldn't do it.

```
beforeEach('Test', function(done) {
...
});
```
in discriminators.md

```javascript
[require:Test]
```

vkarpov15: Then you need to figure out why the example is not getting pulled in :)

PR #10388, opened 2021-06-24
fix(TypeScript): Query.cast()

lkho:
**Summary**
Query.cast() args should be optional

reference source:
https://github.com/Automattic/mongoose/blob/6904cdeb83e936ea4e8d370496d1667d0b747594/lib/query.js#L4906

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1975,7 +1975,14 @@ declare module 'mongoose' {
     box(val: any): this;
     box(lower: number[], upper: number[]): this;
 
-    cast(model: Model<any, THelpers> | null, obj: any): UpdateQuery<DocType>;
+    /**
+     * Casts this query to the schema of `model`.
+     *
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1975,7 +1975,14 @@ declare module 'mongoose' {
...
 
-    cast(model: Model<any, THelpers> | null, obj: any): UpdateQuery<DocType>;
+    /**
+     * Casts this query to the schema of `model`.
+     *
+     * @param {Model} [model] the model to cast to. If not set, defaults to `this.model`
+     * @param {Object} [obj] If not set, defaults to this query's conditions
+     * @return {Object} the casted `obj`
+     */
+    cast(model?: Model<any, THelpers> | null, obj?: any): any;
```
vkarpov15, line 1985: Thanks :+1:

vkarpov15 (approved):

PR #10393, opened 2021-06-26
Replace mongodb-topology-manager

YC:
**Summary**
~~Upgrades CI environment to latest LTS version.~~
Refactored connection tests to remove use of `mongodb-topology-manager`.
The latest version on npm doesn't support the newer version of mongod (likely due to change https://github.com/mongodb/mongo/commit/e628734fdce256588793b8eda0998f04f5b696e6, which capitalised 'waiting' in the stdout logging). 
The latest version on Github doesn't `npm install` correctly for Node 15/16. 

This commit introduces `mkdirp` and `rimraf` to `devDependencies`, but they seem to be imported by other dependencies anyway.


test/connection_server.js
```diff
@@ -0,0 +1,79 @@
...
+    return new Promise(function(resolve, reject) {
+      self.purge().then(function() {
+        // Spawn server process
+        self.server = child_process.spawn(self.binary,
+          ['--port', self.port, '--bind_ip',
+            self.bind_ip, '--dbpath', self.dbpath]);
+
+        // Wait for ready
+        self.server.stdout.on('data', function(data) {
+          if (data.toString().includes('Waiting for connections')) {
```
vkarpov15, line 29: This change means that this logic won't work with MongoDB < 4.4, right?

.github/workflows/test.yml
```diff
@@ -26,16 +26,16 @@ jobs:
 
       - name: Setup
         run: |
-          wget -q http://downloads.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1804-4.0.2.tgz
-          tar xf mongodb-linux-x86_64-ubuntu1804-4.0.2.tgz
+          wget -q https://downloads.mongodb.org/linux/mongodb-linux-x86_64-ubuntu2004-4.4.6.tgz
```
vkarpov15, line 29: Is it necessary to upgrade to 4.4 in this change if mongodb-topology-manager doesn't support it? I'd rather remove mongodb-topology-manager in a separate release.

test/connection_server.js
YC: I'll look into it, and try to improve compatibility with older versions.

.github/workflows/test.yml
YC: According to https://www.mongodb.com/download-center/community/releases/archive, 4.4.0 is the first version with a Ubuntu 20.04 archive/package. 
I'll revert `test.yml` for now.

test/connection_server.js
YC: I've reverted `test.yml`, so it should be back to testing 4.0.2.

vkarpov15 (approved):
Works for me. We should move off of mongodb-topology-manager anyway, it doesn't look like it'll be maintained going forward.

PR #10406, opened 2021-06-30
[doc] Fix outdated URL about unindexed upsert

grimmer0125:
**Summary**

The original URL is outdated and it needs to use the new URL on the MongoDB site.


vkarpov15 (approved):
Thanks :+1:

PR #10414, opened 2021-07-02
Allow usage of schema reserved keywords

### LINKED ISSUE #9010: Remove all reserved keywords from schema
re https://github.com/Automattic/mongoose/issues/8940#issuecomment-628279424 https://github.com/Automattic/mongoose/issues/8940#issuecomment-629479042

It would be great if we could get rid of entirely of schema reserved keywords.

It's a bit tricky, but I believe it's doable. AFAIK the reason we're having reserved keywords is that we don't want to overwrite document methods.
Even for those, we could allow them and offer alternatives to users, and use that alternative internally. Symbols seem like a good option for internal implementation.

This could be quite useful for people migrating from other types of DBs to MongoDB without having to change their fields' names and don't want their apps to break other apps that depend on their data being the same names.
```js
// user wants to use `validate` as a field name
// it's already present in their SQL DB which they are migrating
const userSchema = new Schema(
  { validate: { type: Boolean } },
  { suppressReservedWarning: true }
);

// if suppressReservedWarning is `false` (the default) compiling this schema
// should log a warning that `validate` field will overwrite a method for documents
// you'll need to use an alternative for those methods
// for more info: https://link.to.docs/alternatives
// this should include a list of alternatives that would still work

const user = new User({ validate: false });


// one alternative would be exporting an immutable object on mongoose
// that has a list of symbols, which we'll use internally
// and let the user use for themselves in this case.

const { validateSymbol } = mongoose.symbols;
await user[validateSymbol]();

// another option, that I don't really like
// is defining methods as statics on model that behave differently
// depending on the arguments.

await User.validate(user);
// at Model.validate, the first line should be
// if (obj instanceof Document) {
//   return obj[validateSymbol]();
// }
```

Thoughts? @vkarpov15 @lineus

AbdelrahmanHafez:
fixes #9010


AbdelrahmanHafez:
I think we should replace the error throwing with a warning when someone uses a key that is previously reserved, the warning should include a link to an FAQ letting people know the history of reserved, why it's there, and advise them to prefer `$method` over `method`, then we'll need to offer an option to shut that warning off on each schema and globally.

lib/types/subdocument.js
```diff
@@ -161,7 +161,7 @@ Subdocument.prototype.isModified = function(paths, modifiedPaths) {
       paths = this.$__pathRelativeToParent();
     }
 
-    return parent.isModified(paths, modifiedPaths);
+    return parent.$isModified(paths, modifiedPaths);
```
AbdelrahmanHafez, line 164: @vkarpov15 
Is it possible that `parent` here can be a subdocument? If that's the case it's possible that `$isModified` won't exist on it, we'd need to handle that.

vkarpov15: It is possible, but subdocuments inherit from documents so they should both have `$isModified()`

vkarpov15:
I agree with this comment: https://github.com/Automattic/mongoose/pull/10414#issuecomment-876863778

My only concern thus far is changing the docs to use `$save()` instead of `save()`. Does that change feel too heavy to you? We will always support `.save()` _unless_ the user explicitly creates a `save` property on the schema, so I'm a little wary of changing it in the docs because it implies that `.save()` will no longer work.

lib/types/subdocument.js
AbdelrahmanHafez: Then we're safe, thank you.

AbdelrahmanHafez:
Actually, I agree. I don't think we should change it in the docs now that we'll warn users when they're using reserved keys, my initial thought was that we'll prefer `$save(...)` as a default over `save(...)`, but that's not true for users, it should be the default for plugin authors and mongoose internal usage.

I'll revert the changes in the docs.

AbdelrahmanHafez:
Assumption to be tested:
any object that `isMongooseObject(obj) === true` inherits from `Document`, therefore it has access to `$toObject(...)`

AbdelrahmanHafez:
@vkarpov15 
All ready except for the following:
* preferring `$toObject` over `toObject` internally, the name `toObject` is used all over the place and I am not sure what inherits from what
allowing `schema.reserved.collection` and using `collection` in schema results in an error that I yet haven't been able to resolve:
```
TypeError: Cannot read property 'Symbol(mongoose#Document#scope)' of undefined
    at Model.set [as collection] (lib\helpers\document\compile.js:176:32)
    at Function.compile (lib\model.js:4804:30)
    at Mongoose._model (lib\index.js:554:27)
    at NativeConnection.Connection.model (lib\connection.js:1139:23)
    at NativeConnection.conn.model (test\common.js:106:18)
    at Context.<anonymous> (test\document.test.js:10536:25)
    at processImmediate (internal/timers.js:464:21)
```

vkarpov15 (approved):
Great work, thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

steven266:
@AbdelrahmanHafez Thank you for your great work!

https://github.com/Automattic/mongoose/pull/10414#issuecomment-893060615

> Allowing schema.reserved.collection and using collection in schema results in an error that I yet haven't been able to resolve:

It seems like `collection` is still not usable, right? Is there any way I can help to make `collection` usable as a schema key?

AbdelrahmanHafez:
Yeah, that's still one of the words that are unusable yet.
The way I would do it is pull the repository locally, start a new branch starting from master, go to the following line:
https://github.com/Automattic/mongoose/blob/ebab6646a8f33805b7a2af0cea5dff8f82164bee/test/schema.test.js#L1441


Uncomment `collection`, and to run the tests:
```
$ npm i
$ npm run tdd
```

This test would fail, I'd see where it's failing and try to make it pass, we then need to make an alias for `collection` called `$collection`, and if anything internally in mongoose is using `collection`, we need to make it using `$collection` instead.

PR #10417, opened 2021-07-03
fix: issue 10415

SFVII:
Fix issue  10415, just adding "@types/node": "^14.14.20" to override an older version installed by a submodule


vkarpov15 (commented):
Sorry we missed this, we fixed this in https://github.com/Automattic/mongoose/commit/11596317d4ba2b860aa9b5994290b8cacb600782.

PR #10439, opened 2021-07-08
always pass unpopulated path to validator closes #8042 closes #8018

### LINKED ISSUE #8018: Populating an id key that doesn't exist anymore in the reference collection (single, not an array) and then saving gives a validation error if the id key is a required value
Hi, I wanted to point out this error I am getting that I think is sort of a bug, since technically we are just executing READ commands on the document(s), and the content changes so that the validation fails.

When trying to save a document, after it has been populated on a required key, and the populated key is null because the reference object was not found anymore, trying to save it fails due to a validation error in the populated key being null after the populate.

```javascript
mongoose.set('debug', true);

const GITHUB_ISSUE = `gh8018`;
const connectionString = `mongodb://localhost:27017/${GITHUB_ISSUE}`;

run().then(() => console.log('done')).catch(error => console.error(error.stack));

async function run() {
    const db = await mongoose.connect(connectionString);

    const schema = mongoose.Schema({
        placeHolderKey: {
            type: String,
            default: "TEST"
        }
    });

    const schema2 = mongoose.Schema({
        keyToPopulate: {
            ref: "SecondaryModel",
            required: true
        }
    });


    const secondaryModelThatIsPopulatedOnTheMainModel = db.model('SecondaryModel', schema);
    const mainModelThatFailsValidation = db.model('MainModelThatFailsRequiredValidation', schema2);

    let secondaryModelObject = await secondaryModelThatIsPopulatedOnTheMainModel.create({placeHolderKey: "TEST"});
    let mainModelObject = await mainModelThatFailsValidation.create({keyToPopulate: secondaryModelObject._id});
    await secondaryModelObject.remove();

    const doc = await mainModelThatFailsValidation.find({_id: mainModelObject._id})
        .populate("keyToPopulate")
        .exec();

    doc[0].save(); //This one normally throws the error

    console.log(doc[0].toObject()); //Removing the `await secondaryModelObject.remove();` will throw no errors
}
```

I think that `populate` shouldn't affect the actual content to the point of making it different from the previous one, since we didn't change anything at all.

Mongoose: 5.5.2
Node: 8.12.0

### LINKED ISSUE #8042: Always pass unpopulated path to validators
Issues like #8018 make it very confusing to write validation for populated paths.

IslandRhythms:

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2525,7 +2525,7 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
       // shouldn't fail (gh-8018). We should always fall back to the populated
       // value.
       let pop;
-      if (val == null && (pop = _this.populated(path)) || !(pop = _this.populated(path))) {
+      if (val == null) {
         val = pop;
       }
       const scope = path in _this.$__.pathsToScopes ?
--- a/lib/document.js
+++ b/lib/document.js
@@ -2524,9 +2524,8 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
       // If you `populate()` and get back a null value, required validators
       // shouldn't fail (gh-8018). We should always fall back to the populated
       // value.
-      let pop;
-      if (val == null) {
-        val = pop;
+      if (val == null && _this.populated(path)) {
+        val = _this.depopulate(path);
       }
       const scope = path in _this.$__.pathsToScopes ?
         _this.$__.pathsToScopes[path] :
--- a/lib/document.js
+++ b/lib/document.js
@@ -2524,8 +2524,9 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
       // If you `populate()` and get back a null value, required validators
       // shouldn't fail (gh-8018). We should always fall back to the populated
       // value.
-      if (val == null && _this.populated(path)) {
-        val = _this.depopulate(path);
+      let pop;
+      if ((pop = _this.populated(path))) {
+        val = pop;
       }
       const scope = path in _this.$__.pathsToScopes ?
         _this.$__.pathsToScopes[path] :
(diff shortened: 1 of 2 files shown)


lib/document.js
```diff
@@ -2524,9 +2524,8 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
       // If you `populate()` and get back a null value, required validators
       // shouldn't fail (gh-8018). We should always fall back to the populated
       // value.
-      let pop;
-      if (val == null && (pop = _this.populated(path))) {
-        val = pop;
+      if (val == null && _this.populated(path)) {
+        val = _this.depopulate(path);
```
vkarpov15, line 2528: It looks like this would depopulate the path on the document, rather than depopulating just for the validator. We don't want the depopulate to persist.

lib/document.js
```diff
@@ -2524,9 +2524,8 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
       // If you `populate()` and get back a null value, required validators
       // shouldn't fail (gh-8018). We should always fall back to the populated
       // value.
-      let pop;
-      if (val == null && (pop = _this.populated(path))) {
-        val = pop;
+      if (val == null && _this.populated(path)) {
```
vkarpov15, line 2527: We should remove this `val == null` check. This `null` check is the crux of this issue: we should always use `val = pop`

vkarpov15 (changes requested):

vkarpov15 (changes requested):
Nice, but can you add a test for this please?

IslandRhythms:
I copied your test from 8018 and just didn't populate it. Not sure if that's how to properly test this fix. I am a bit confused on how to test for passing an unpopulated path to a validator, as I thought that was already default behavior on model creation when you have a field in the document set to `required: true`

vkarpov15 (approved):
Thanks :+1:

Issue #10441, opened 2021-07-09
Inconsistency when setting the defaults on a model


semirturgay:
**What is the current behavior?**

**Defaults not set for models contains multiple objects some with defaults**

Given the following model: 
```
const DefaultTestObject = db.model('Test',
        new Schema({
          foo: {
            x: { type: String },
            y: { type: String }
          },
          boo: {
            x: { type: Boolean, default: false },
            y: { type: String }
          },
          bee: {
            x: { type: Boolean, default: false },
            y: { type: Boolean, default: false }
          }
        })
      );
```
---
When initializing the following model: 

```
 const myTest = new DefaultTestObject({
        foo: { x: 'a', y: 'b' },
        bee: {},
        boo: {}
  });
```

then running the following assertion fails: 

```
myTest.save(function(err, doc) {
        assert.ifError(err);
        assert.equal(doc.boo.x, false); // we have the defaults set for boo {x: false}
        assert.equal(doc.bee.x, false); // ☠️  but defaults are not set for *bee* it is {}
        assert.equal(doc.bee.y, false);
});       
``` 
---
The behavior gets more interesting when I try: 

```
const myTest = new DefaultTestObject({
        foo: { x: 'a', y: 'b' },
        bee: {x: true},
        boo: {}
  });
``` 

now: 

```
myTest.save(function(err, doc) {
        assert.ifError(err);
        assert.equal(doc.bee.x, true); // the  defaults are set for *bee* it is {x: true, y: false}
        assert.equal(doc.bee.y, false); 
        assert.equal(doc.boo.x, false); // ☠️  but defaults are not set for *boo* it is {}
});       
``` 
---
Now making the things even more interesting:

```
// I will not set a value for *foo*
const myTest = new DefaultTestObject({
        bee: {x: true},
        boo: {}
  });
``` 

now all the assertions are good (the defaults are set properly):  

```
myTest.save(function(err, doc) {
        assert.ifError(err);
        assert.equal(doc.bee.x, true); 
        assert.equal(doc.bee.y, false); 
        assert.equal(doc.boo.x, false); 
});       
``` 

**If the current behavior is a bug, please provide the steps to reproduce.**

- Have a model with a schema that has 3 objects 
- Set defaults at the schema definition for some keys of at least two objects
- Try to initialize the model with the object having no default and pass an empty object for the objects having defaults
- Save the model, and check the DB, you will see the defaults are not populated correctly 
[...]

PR #10442, opened 2021-07-09
fix(model): applying object defaults

semirturgay:
**Summary**
Proof of concept and fix for #10441 

**Examples**
check #10441

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -994,6 +994,9 @@ Document.prototype.$set = function $set(path, val, type, options) {
         delete this._doc[key];
         // Make sure we set `{}` back even if we minimize re: gh-8565
         options = Object.assign({}, options, { _skipMinimizeTopLevel: true });
+      } else {
+        // Make sure we set `{_skipMinimizeTopLevel: false}` if don't have overwrite: gh-10441
+        options = Object.assign({}, options, { _skipMinimizeTopLevel: false });
       }
 
       const someCondition = typeof path[key] === 'object' &&
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -3826,6 +3826,57 @@ describe('Model', function() {
         });
       });
     });
+    it('is saved object with proper defaults', function(done) {
+      const DefaultTestObject = db.model('Test',
+        new Schema({
+          foo: {
+            x: { type: String },
+            y: { type: String }
+          },
+          boo: {
+            x: { type: Boolean, default: false }
+          },
+          bee: {
+            x: { type: Boolean, default: false },
+            y: { type: Boolean, default: false }
+          }
+        })
+      );
+      const myTest = new DefaultTestObject({
+        foo: { x: 'a', y: 'b' },
+        bee: {},
+        boo: {}
+      });
+
+      myTest.save(function(err, doc) {
+        assert.ifError(err);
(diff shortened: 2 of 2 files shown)


test/model.test.js
```diff
@@ -3826,6 +3826,57 @@ describe('Model', function() {
...
+          }
+        })
+      );
+      const myTest = new DefaultTestObject({
+        foo: { x: 'a', y: 'b' },
+        bee: {},
+        boo: {}
+      });
+
+      myTest.save(function(err, doc) {
```
AbdelrahmanHafez, line 3851: Using callback is considered old-fashioned, a more maintainable option would be to use `co/yield`, [here](https://github.com/Automattic/mongoose/blob/c01685a56eccca995c6880dcc6ea51c71d4dd43f/test/document.test.js#L9076)'s an example of how we use it in tests.

AbdelrahmanHafez (changes requested):

test/model.test.js
semirturgay: Hello thanks for the comment, this pr created to demonstrate the issue https://github.com/Automattic/mongoose/issues/10441,

lib/document.js
```diff
@@ -1072,18 +1072,19 @@ Document.prototype.$set = function $set(path, val, type, options) {
           this.$__.savedState[path + '.' + key] = priorVal[key];
         }
       }
+      if (!utils.isEmptyObject(val)) {
```
vkarpov15, line 1075: This unfortunately breaks setting a nested field to an empty object (see test failures), so this isn't a correct solution. I haven't been able to look closely enough to have any suggestions about how to fix this issue though.

vkarpov15 (changes requested):

lib/document.js
semirturgay: @vkarpov15 thanks for the comment, 

I see the problem this fix creates.  
It is a blocker for us unfortunately without this fixed We can't update mongoose,  
it will be great if you can have a look and provide some help to fix this.

semirturgay: @vkarpov15 

I had a deeper look at the repo, and I believe this time I find the correct solution for this.  

It was basically happening because of not restoring the _skipMinimizeTopLevel to false back after we set it to overwrite the nested object value if it is set by the user. 

```
 if (type === true &&
          !prefix &&
          path[key] != null &&
          pathtype === 'nested' &&
          this._doc[key] != null &&
          Object.keys(this._doc[key]).length === 0) {
        delete this._doc[key];
        // Make sure we set `{}` back even if we minimize re: gh-8565
        options = Object.assign({}, options, { _skipMinimizeTopLevel: true });
      } else {
        // Make sure we set `{_skipMinimizeTopLevel: false}` if don't have overwrite: gh-10441
        options = Object.assign({}, options, { _skipMinimizeTopLevel: false });
      }
```

vkarpov15 (approved):
Thanks. Fix will be in v5.13.3 :+1:

nikolasleblanc:
Hey @vkarpov15, wondering if you have a timeline on when this release might go out? Appreciate your work on this.

AbdelrahmanHafez:
@nikolasleblanc FYI, this is released now.

Issue #10437, opened 2021-07-08
bulkSave() throws an error when an unchanged document is passed


lucashpmelo:
**Do you want to request a *feature* or report a *bug*?**

A bug

**What is the current behavior?**

When an unmodified object is passed to bulkSave() an error is raised:
`TypeError: Cannot read property '0' of undefined`

**If the current behavior is a bug, please provide the steps to reproduce.**

```javascript
const mongoose = require('mongoose')
const { Schema } = mongoose

const testSchema = new Schema({
  name: {
    type: String,
    required: true,
  },
  date: {
    type: Date,
    required: true,
  },
  value: {
    type: Number,
    required: true,
    default: 0,
  },
})

new mongoose.model('Test', testSchema)
const Test = mongoose.model('Test')

async function bulkSave(data) {
  const bulk = []

  for (const i in data) {
    const { name, date, value } = data[i]

    let test = await Test.findOne({ name: name, date: date })

    if (!test) {
      test = new Test(data[i])
    }

    test.value = value

    bulk.push(test)
  }

  return await Test.bulkSave(bulk)
}

async function run() {
  await mongoose.connect('mongodb://localhost:27017/test', {
    useNewUrlParser: true,
    useUnifiedTopology: true,
  })

  await mongoose.connection.dropDatabase()

  const data1 = [
    { name: 'Test 1', date: '2021-08-01T00:00:00.000Z', value: 10 },
    { name: 'Test 1', date: '2021-08-02T00:00:00.000Z', value: 20 },
    { name: 'Test 2', date: '2021-08-01T00:00:00.000Z', value: 15 },
  ]

  await bulkSave(data1)

  const data2 = [
    { name: 'Test 1', date: '2021-08-01T00:00:00.000Z', value: 15 },
    { name: 'Test 1', date: '2021-08-02T00:00:00.000Z', value: 20 }, // <-- ERROR
    { name: 'Test 2', date: '2021-08-02T00:00:00.000Z', value: 25 },
  ]

  await bulkSave(data2)

  console.log('Done')
}

run().catch((err) => console.log(err))
```

**What is the expected behavior?**

bulkSave() handle whether the document has changes or not.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Node v13.11.0
Mongoose v5.13.2
MongoDB 4.2.5

PR #10464, opened 2021-07-17
test(model): assert Model.bulkSave(...) works if some document is not modified

AbdelrahmanHafez:
re #10437

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -7567,8 +7567,8 @@ describe('Model', function() {
       return co(function*() {
         const user = yield User.create({ name: 'Hafez' });
 
-        const res = yield User.bulkSave([user]);
-        assert.ok(res.result.ok);
+        const err = yield User.bulkSave([user]).then(() => null, err => err);
+        assert.ok(err == null);
       });
     });
   });
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -7557,6 +7557,20 @@ describe('Model', function() {
         assert.equal(usersFromDatabase[1].name, 'name from pre-save');
       });
     });
+    it('works if some document is not modified (gh-10437)', () => {
+      const userSchema = new Schema({
+        name: String
+      });
+
+      const User = db.model('User', userSchema);
+
+      return co(function*() {
+        const user = yield User.create({ name: 'Hafez' });
+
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

test/model.test.js
```diff
@@ -7557,6 +7557,20 @@ describe('Model', function() {
...
+        name: String
+      });
+
+      const User = db.model('User', userSchema);
+
+      return co(function*() {
+        const user = yield User.create({ name: 'Hafez' });
+
+        const res = yield User.bulkSave([user]);
+        assert.ok(res.result.ok);
```
vkarpov15, line 7571: I'd prefer if we didn't assert on particular properties in `res.result`. I'd just make this check that no error was thrown using `const err = yield User.bulkSave([user]).then(() => null, err => err); assert.ifError(err);`

AbdelrahmanHafez: I agree, that's the way to go.

AbdelrahmanHafez: I agree, that's the way to go.

AbdelrahmanHafez:
Happy to help 👍

Kamikadze4GAME:
Maybe to change this lines (`L3731-L3756`)

https://github.com/AbdelrahmanHafez/mongoose/blob/gh-10437/lib/model.js#L3731-L3756

into 

```js
    const isANewDocument = document.isNew;

    if (isANewDocument) {
      accumulator.push({
        insertOne: { document }
      });

      return accumulator;
    }

    const delta = document.$__delta();
    const isDocumentWithChanges = delta != null && !utils.isEmptyObject(delta[0]);

    if (isDocumentWithChanges) {
      const where = document.$__where(delta[0]);
      const changes = delta[1];

      _applyCustomWhere(document, where);

      document.$__version(where, delta);

      accumulator.push({
        updateOne: {
          filter: where,
          update: changes
        }
      });
      
      return accumulator;
    }

    return accumulator;
```

AbdelrahmanHafez:
@Kamikadze4GAME Done 👍

Kamikadze4GAME (approved):

vkarpov15 (approved):

PR #10465, opened 2021-07-17
refactor(test): prefer co/yield over callbacks re #10441

AbdelrahmanHafez:
re #10441 #10442

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -3826,57 +3826,47 @@ describe('Model', function() {
         });
       });
     });
-    it('is saved object with proper defaults', function(done) {
-      const DefaultTestObject = db.model('Test',
-        new Schema({
-          foo: {
-            x: { type: String },
-            y: { type: String }
-          },
-          boo: {
-            x: { type: Boolean, default: false }
-          },
-          bee: {
-            x: { type: Boolean, default: false },
-            y: { type: Boolean, default: false }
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1: feel free to put in PRs for changing from callbacks to co whenever you feel like it

AbdelrahmanHafez:
Happy to help.
I'm planning to refactor a lot of the code after we release v6.0 to use all the new features we couldn't use in order to support older versions of node.

Kamikadze4GAME:
Hello @AbdelrahmanHafez.
Do you mean to change all `co/yield` into `async/await`, right?

AbdelrahmanHafez:
Hello @Kamikadze4GAME 

I think we should refactor both code that is `co/yield` and callbacks to use async/await, but we can't do that before v6.0 because we're dropping support for node <= 12 in v6.0, and refactoring now on v6.0 branch will cause a lot of conflicts, so it's just better we leave it until we merge 6.0 into master, then take it from there.

vkarpov15:
I agree with @AbdelrahmanHafez . We want to switch to using async/await for tests, but not until after 6.0 is released. We can easily make those changes after 6.0, but in the meantime we need to focus on user-facing changes for 6.0 and minimize unnecessary merge conflicts.

PR #10481, opened 2021-07-22
refactor(defineKey): delete unused parameter, and prefer named parameters over unnamed params

### LINKED ISSUE #10473: [refactor] get rid of defineKey `keys` parameter
The function `defineKey` takes a parameters `keys` that is unused, we should get rid of it, and verify that in the function usage we're passing the parameters correctly.

We could even make function use [named parameters](https://www.freecodecamp.org/news/elegant-patterns-in-modern-javascript-roro-be01e7669cbd/) instead of passing parameters directly, that way we can add parameters or remove them without having to worry about current usage.

AbdelrahmanHafez:
fixes #10473

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -176,7 +176,7 @@ function Document(obj, fields, skipId, options) {
 
     keys.forEach(function(key) {
       if (!(key in schema.tree)) {
-        defineKey(key, null, _this);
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Works for me, thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

PR #10494, opened 2021-07-28
Make LeanDocument work with PopulatedDoc

juhdanad:
**Summary**

When we call `toObject()` on a populated document, it also modifies the subdocuments. The type signature of `LeanDocument` does not reflect this change. The reason is that `(ObjectId | Document)` does not extend `Document`, so `LeanDocument` ignores it. This PR modifies the definition of LeanDocument, so it becomes distributive over union types.

**Examples**

```typescript
import { PopulatedDoc, Document } from "mongoose";
import { ObjectId } from "mongodb";
interface Child {
  name: string;
}
interface Parent {
  child?: PopulatedDoc<Child & Document<ObjectId>>,
}
const parent : Parent & Document = {} as Document; // just for type check
const leanChild = parent.toObject().child;
if (leanChild != null && !(leanChild instanceof ObjectId)) {
  leanChild.save(); // works, but should not work
}
```

**Test case**
The attached test case imports everything from the "mongoose" package (not the local development files), so I think it will not see the changes and shold fail. I did not modify the import as other tests use the same pattern. Also, I used the `ObjectId` from the "mongodb" package, to make `instanceof` type inference possible.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2438,12 +2438,17 @@ declare module 'mongoose' {
   type LeanArray<T extends unknown[]> = T extends unknown[][] ? LeanArray<T[number]>[] : LeanType<T[number]>[];
 
   export type _LeanDocument<T> = {
-    [K in keyof T]:
-    0 extends (1 & T[K]) ? T[K] : // any
-    T[K] extends unknown[] ? LeanArray<T[K]> : // Array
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Nice, thanks :+1:

PR #10498, opened 2021-07-28
proper aggregation cursor

IslandRhythms:


lib/aggregate.js
```diff
@@ -839,7 +839,7 @@ Aggregate.prototype.cursor = function(options) {
     this.options = {};
   }
   this.options.cursor = options || {};
-  return this;
+  return new AggregationCursor(this); // return this;
```
vkarpov15, line 842: Fix this unnecessary comment

vkarpov15 (changes requested):
One quick fix, otherwise LGTM

PR #10501, opened 2021-07-28
Update depopulate documentation for document.js

gfrancz:

**Summary**

Currently the documentation indicates that if a path for depopulate(path) is not provided, then it's a no-op. This is absolutely not true -- in fact, all fields are depopulated. This change fixes the documentation issue.

**Examples**

E.g., depopulate() is not a no-op, it depopulates all populated virtuals.


vkarpov15 (approved):
Thanks :+1:

PR #10515, opened 2021-07-30
perf: improve typescript type checking performance

andreialecu:
**Summary**

Related #10349 

You can use the repro at https://github.com/andreialecu/repro-mongoose-slow-ts

Output of `yarn tsc --extendedDiagnostics`.

Before:
```
Identifiers:                 82983
Symbols:                    128042
Types:                       42997
Instantiations:             291774
Memory used:               148795K
Assignability cache size:    13123
Identity cache size:             3
Subtype cache size:              0
Strict subtype cache size:       0
I/O Read time:               0.02s
Parse time:                  0.39s
ResolveModule time:          0.03s
ResolveTypeReference time:   0.00s
Program time:                0.45s
Bind time:                   0.19s
Check time:                  1.21s
transformTime time:          0.00s
commentTime time:            0.00s
I/O Write time:              0.00s
printTime time:              0.01s
Emit time:                   0.01s
Total time:                  1.87s
✨  Done in 2.37s.
```

After:
```
Identifiers:                 82983
Symbols:                     98359
Types:                       23540
Instantiations:             153096
Memory used:               133310K
Assignability cache size:     7426
Identity cache size:             1
Subtype cache size:              0
Strict subtype cache size:       0
I/O Read time:               0.02s
Parse time:                  0.38s
ResolveModule time:          0.02s
ResolveTypeReference time:   0.00s
Program time:                0.44s
Bind time:                   0.19s
Check time:                  0.76s
transformTime time:          0.00s
commentTime time:            0.00s
I/O Write time:              0.00s
printTime time:              0.01s
Emit time:                   0.01s
Total time:                  1.40s
✨  Done in 1.62s.
```

Notice a huge decrease in `types` and `instantiations`, and also in `Check time`.

I'm not sure if this is the right change, but there was some overlap with `_AllowStringsForIds` being used by `DocumentDefinition` doing most of the same work as `_FilterQuery` is doing. That seems unnecessary and the `DocumentDefinition<T>` seems like could be applied to just a specific part of the type.

This doesn't seem to break anything in my code base, and I got most of the speed back.

PR #10536, opened 2021-08-03
perf: Improve type checking performance

perenSHERE:
### Extends #10515
**Summary**
**Visual Studio code intellicode lags due to the declared Schema Type**
 for testing the speed of the intellicode i used a simple typescript command as `npm run tsc -- --extendedDiagnostics`
**Steps to reproduce**
- clone the repository at : [repro script](https://github.com/perenSHERE/mongoose-repro-intellicode-javascript)
- run  `npm i`
- then run `npm run tsc -- --extendedDiagnostics`
**Results (Before):**
image
- apply the changes in the **Pull Request**
**Results (After):**
image

> This gives me back the snappy intellicode and also doesn't effect the types that much 

**Examples**
Before applying the changes in the Pull Request.

https://user-images.githubusercontent.com/73697546/128023769-6fbb0137-c845-48fc-9372-b9a207fbbcbe.mp4

After applying the changes in the Pull Request.

https://user-images.githubusercontent.com/73697546/128024398-76fbeca4-2971-4b62-bff9-6f545c84a71e.mp4

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1232,7 +1232,7 @@ declare module 'mongoose' {
   type PostMiddlewareFunction<ThisType, ResType = any> = (this: ThisType, res: ResType, next: (err?: CallbackError) => void) => void | Promise<void>;
   type ErrorHandlingMiddlewareFunction<ThisType, ResType = any> = (this: ThisType, err: NativeError, res: ResType, next: (err?: CallbackError) => void) => void;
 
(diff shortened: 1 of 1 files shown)


andreialecu:
Note that this change breaks the types, some lines below become invalid/error out.

vkarpov15:
We made these improvements along with the necessary patches to make our tests pass in fefebb3. Unfortunately this makes our type checking less strict, but the performance impact seems too good to pass up right now. Since our tests all pass, we don't think this change _should_ be backwards breaking, but we'll keep an eye out for any bug reports.

PR #10538, opened 2021-08-03
update type declarations of Schema.index method

Raader:
added proper types to arguments of index method on Schema class.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1257,7 +1257,6 @@ declare module 'mongoose' {
     /** Iterates the schemas paths similar to Array#forEach. */
     eachPath(fn: (path: string, type: SchemaType) => void): this;
 
-    
     /** Defines an index (most likely compound) for this schema. */
     index(fields: mongodb.IndexSpecification, options?: IndexOptions): this;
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1231,6 +1231,8 @@ declare module 'mongoose' {
   type PreSaveMiddlewareFunction<T> = (this: T, next: (err?: CallbackError) => void, opts: SaveOptions) => void | Promise<void>;
   type PostMiddlewareFunction<ThisType, ResType = any> = (this: ThisType, res: ResType, next: (err?: CallbackError) => void) => void | Promise<void>;
   type ErrorHandlingMiddlewareFunction<ThisType, ResType = any> = (this: ThisType, err: NativeError, res: ResType, next: (err?: CallbackError) => void) => void;
+  
+  type Index = { [key: string]: boolean | number | string | Index }
```
vkarpov15, line 1235: Why is this `| Index` here? I don't know of any recursive index definitions.

Raader: My bad, changed it to mongodb's index type, It's my first ever pr on a project like this. it is not much but it should help with intellisense.

vkarpov15 (approved):
Thanks :+1:

Gotos:
This seems to break index creation. `mongodb.IndexSpecification` doesn't seem to be the right type for the fields parameter. According to the docu and how we used mongoose so far, the fields are "just" an object with the fields that you want to use for the index as keys and values for those set to 1. IndexSpecification is vastly different from that. I think this is likely an error? Or am I missing something?

Gotos:
Nevermind, I just realized this was fixed in 5.13.7. Sorry!

PR #10546, opened 2021-08-06
updated array push type definition

shahriar-shojib:
**Summary**

Array Push Type accepts the T type parameter, it was any

**Examples**

```ts
//typescript
const x = [1,2,3];
x.push('x') // error

//mongoose array
const x = [1,2,3];
x.push('x') // ok
```
```ts
//mongoose array
const x = [ { a: 1 }, { a: 2 } ];
x.push('x') // ok
x.push({ a: 3 }) // no intellisense

```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1970,10 +1970,10 @@ declare module 'mongoose' {
       isMongooseArray: true;
 
       /** Pushes items to the array non-atomically. */
-      nonAtomicPush(...args: any[]): number;
+      nonAtomicPush(...args: T[]): number;
 
       /** Wraps [`Array#push`](https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Array/push) with proper change tracking. */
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1970,10 +1971,10 @@ declare module 'mongoose' {
       isMongooseArray: true;
 
       /** Pushes items to the array non-atomically. */
-      nonAtomicPush(...args: any[]): number;
+      nonAtomicPush(...args: T[]): number;
```
vkarpov15, line 1974: I'd make this `AnyKeys<T> & AnyObject` instead of `T[]`. Mongoose handles casting, so you don't necessarily have to push an instance of `T`.

vkarpov15 (changes requested):

index.d.ts
shahriar-shojib: Hello,
I have made the changes you requested

PR #10547, opened 2021-08-06
fix(index.d.ts): add missing type for diffIndexes

bvgusak:
**Summary**

Add missing types for Model's `diffIndexes()` function introduced in 5.13.0.

**Examples**

Will allow using `diffIndexes()` in TypeScript:
```
const MyModel = mongoose.model("MyModel", mySchema);
const indexesDiff = await MyModel.diffIndexes()
```


vkarpov15 (approved):
Thanks :+1:

PR #10550, opened 2021-08-07
refactor(index.d.ts): add `Callback` type to avoid code repetition

thiagokisaki:


index.d.ts
```diff
@@ -2616,8 +2616,8 @@ declare module 'mongoose' {
      */
     eachAsync(fn: (doc: DocType) => any, options?: { parallel?: number }): Promise<void>;
     eachAsync(fn: (doc: DocType[]) => any, options: { parallel?: number, batchSize: number }): Promise<void>;
-    eachAsync(fn: (doc: DocType) => any, options?: { parallel?: number, batchSize?: number }, cb?: (err: CallbackError) => void): void;
-    eachAsync(fn: (doc: DocType[]) => any, options: { parallel?: number, batchSize: number }, cb?: (err: CallbackError) => void): void;
+    eachAsync(fn: (doc: DocType) => any, options?: { parallel?: number, batchSize?: number }, cb?: Callback): void;
```
vkarpov15, line 2619: This looks like it might be a problem: this change makes the `eachAsync()` callback take a param. Is this intentional or maybe we should have a `CallbackNoResult` type?

thiagokisaki: You're right. It might be a problem. Although `Callback` still allows you to pass a callback receiving only the `error`, it can mislead users that callback is called with two arguments. It's better to create a `CallbackNoResult` as you mentioned.

vkarpov15 (approved):
Great, thanks :+1:

PR #10562, opened 2021-08-09
Fix type for schema.index() to reflect actual mongoose API

### LINKED ISSUE #10561: Bug in Typescript definitions for Schema.index()
bug (in Typescript types)

The definition of Schema.index() (at https://github.com/Automattic/mongoose/blob/master/index.d.ts#L1279-L1280) says that the `fields` argument is of type `mongodb.IndexSpecification`, but mongoose actually takes its own shorthand syntax for the indexed fields, namely an object containing the indexed fields and the sort order (see https://mongoosejs.com/docs/api/schema.html#schema_Schema-index and https://github.com/Automattic/mongoose/blob/master/lib/schema.js#L1549-L1560).

This is as easy as the following schema definition (as a `.ts` file):

```ts

export const schema = new mongoose.Schema({ whatever: String });

schema.index({ whatever: 1 });
```
... you'll get an error on the `whatever: 1` in the `schema.index()` call, stating:
```text
error TS2345: Argument of type '{ whatever: number; }' is not assignable to parameter of type 'IndexSpecification'.
```

Here's the `tsconfig.json` file, though it shouldn't matter much:
```json
{
  "extends": "@tsconfig/node14/tsconfig.json",

  "compilerOptions": {
    "rootDir": "./",
    "outDir": "./dist",

    "allowJs": true,
    "checkJs": false,

    "typeRoots": [
      "@types",
      "node_modules/@types"
    ]
  },

  "include": ["**/*"]
}
```

No error, the `fields` argument to `schema.index()` should probably have a type definition similar to `SchemaDefinition<>`/`DocumentDefinition<>` that leverages the `SchemaDefinitionType` to define the allowable keys.  Maybe something like
```ts
export type IndexDefinition<T> = { [K in keyof T]: 1 | -1 };
// ...

/** Defines an index (most likely compound) for this schema. */
index(fields: IndexDefinition<DocumentDefinition<SchemaDefinitionType>>>, options?: IndexOptions): this;
```

(The `SchemaDefinition<>` type might need to be in that chain as well... it's not clear to me what specific value that layer adds.)

Node.js: 14.16.0
mongoose: 5.13.6
mongodb: irrelevant

JaredReisinger:
**Summary**

See issue #10561: the Typescript typings for `Schema.index()` were recently broken.  Commit 1c66cc972c90529bd0ffe8496e9de866b92a97fa broke the typing for `Schema.index()` by making `fields` look like mongo's `IndexSpecification`, when it's not: mongoose has its own idea about what the index specification looks like.  This not only undoes that breaking change, but adds additional type-safety so that only actual schema fields can be used in the index definition.

(I'm not 100% sure about the changed return value on `Schema.indexes()`... there may be prefixing and a `background` field added, as per https://github.com/Automattic/mongoose/blob/master/lib/helpers/schema/getIndexes.js.)

Fixes #10561.

**Examples**

With this change, the following code (as Typescript): 
```ts

export const schema = new mongoose.Schema({ whatever: String });

schema.index({ whatever: 1 });
```
will not cause an error.  Further, specifying an index of `{ whatever: 2 }` _**will**_ be an error, as only `1` and `-1` are now valid values for the index keys.  Also, `{ somethingElse: 1 }` will be an error as `somethingElse` is not defined as a schema field.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1246,7 +1246,8 @@ declare module 'mongoose' {
   type ExtractQueryHelpers<M> = M extends Model<any, infer TQueryHelpers> ? TQueryHelpers : {};
   type ExtractMethods<M> = M extends Model<any, any, infer TMethods> ? TMethods : {};
 
-  type IndexDefinition<T> = { [K in keyof T]: 1 | -1; };
+  type IndexDirection = 1 | -1;
+  type IndexDefinition<T> = { [K in keyof T]: IndexDirection; };
 
   type PreMiddlewareFunction<T> = (this: T, next: (err?: CallbackError) => void) => void | Promise<void>;
   type PreSaveMiddlewareFunction<T> = (this: T, next: (err?: CallbackError) => void, opts: SaveOptions) => void | Promise<void>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1246,6 +1246,8 @@ declare module 'mongoose' {
   type ExtractQueryHelpers<M> = M extends Model<any, infer TQueryHelpers> ? TQueryHelpers : {};
   type ExtractMethods<M> = M extends Model<any, any, infer TMethods> ? TMethods : {};
 
+  type IndexDefinition<T> = { [K in keyof T]: 1 | -1; };
+
   type PreMiddlewareFunction<T> = (this: T, next: (err?: CallbackError) => void) => void | Promise<void>;
   type PreSaveMiddlewareFunction<T> = (this: T, next: (err?: CallbackError) => void, opts: SaveOptions) => void | Promise<void>;
   type PostMiddlewareFunction<ThisType, ResType = any> = (this: ThisType, res: ResType, next: (err?: CallbackError) => void) => void | Promise<void>;
@@ -1277,13 +1279,13 @@ declare module 'mongoose' {
     eachPath(fn: (path: string, type: SchemaType) => void): this;
 
     /** Defines an index (most likely compound) for this schema. */
-    index(fields: mongodb.IndexSpecification, options?: IndexOptions): this;
+    index(fields: IndexDefinition<SchemaDefinition<DocumentDefinition<SchemaDefinitionType>>>, options?: IndexOptions): this;
 
     /**
      * Returns a list of indexes that this schema declares, via `schema.index()`
      * or by `index: true` in a path's options.
      */
-    indexes(): Array<any>;
+    indexes(): Array<IndexDefinition<SchemaDefinition<DocumentDefinition<SchemaDefinitionType>>>>;
 
     /** Gets a schema option. */
     get<K extends keyof SchemaOptions>(key: K): SchemaOptions[K];
--- a/index.d.ts
+++ b/index.d.ts
@@ -1246,9 +1246,9 @@ declare module 'mongoose' {
   type ExtractQueryHelpers<M> = M extends Model<any, infer TQueryHelpers> ? TQueryHelpers : {};
   type ExtractMethods<M> = M extends Model<any, any, infer TMethods> ? TMethods : {};
 
-  type IndexDirection = 1 | -1;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1246,6 +1246,8 @@ declare module 'mongoose' {
   type ExtractQueryHelpers<M> = M extends Model<any, infer TQueryHelpers> ? TQueryHelpers : {};
   type ExtractMethods<M> = M extends Model<any, any, infer TMethods> ? TMethods : {};
 
+  type IndexDefinition<T> = { [K in keyof T]: 1 | -1; };
```
thiagokisaki, line 1249: Could you include a [IndexDirection](https://github.com/mongodb/node-mongodb-native/blob/4.1/src/operations/indexes.ts#L62) type and change this line to use it?

JaredReisinger: Done!

JaredReisinger: ... Let me know if you want the additional sorting directions that the newer mongo types allow.  I've only ever used 1/-1, and I'm not sure what the built-in mongo version expectations are that mongoose has.

thiagokisaki: Considering that mongoose v5.x depends on mongodb v3.6 and that mongodb driver supports MongoDB v4.4, I think we should add the other sorting directions. In [MongoDB v4.4 index docs](https://docs.mongodb.com/v4.4/indexes/) they mention 'text', '2d', '2dsphere' and 'geoHaystack' indexes.

thiagokisaki: I just noticed a problem in allowing the index creation for only the properties declared in schema definition. That's because it's possible to create indexes for fields in embedded documents using "dot notation". So an index for `'location.state'`, for example, is also valid. See https://docs.mongodb.com/manual/core/index-single/#create-an-index-on-an-embedded-field

JaredReisinger: Hmm.... tricky.  I don't think Typescript can climb cross-schema references to get an exhaustive list.  Next-best would probably be "any prop, or a string that starts with a prop name followed by a period"... but I don't think that's really representable in Typescript.  Template literal types (https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) would allow for expanding/augmenting the property names, but only to a discrete set of concrete values.  There's no "startsWith" equivalent in Typescript.  All of which leaves us back at "any property name is okay"; not so great for adding type-safety, but at least it shouldn't cause errors.

I'll add the other sorting directions as well.

JaredReisinger: I spent a little bit of time seeing if I could create a recursive type that actually _would_ exhaustively generate the allowable index keys... but the generic type parameters on Schema are opaque enough to me that I couldn't tell whether what I was writing would work or not.  (In my particular case, we're using a tool that generates the doc/model/schema types from the schema definition code, and I suspect it uses an incomplete definition for the schema... the `SchemaDefinitionType` generic type parameter is never provided, which means I have no idea what this value is *supposed* to be/look-like.)  I do believe it's possible to do, it's just more intricate than I can tackle at the moment.

In any case, `IndexDefinition` now allows any string property, and `IndexDirection` includes:

* `1`, `-1`: mentioned at https://docs.mongodb.com/v4.4/indexes/#index-names
* `2d`: mentioned at https://docs.mongodb.com/v4.4/tutorial/build-a-2d-index/#create-a-2d-index
* `2dsphere`: mentioned at https://docs.mongodb.com/v4.4/core/2dsphere/#create-a-2dsphere-index
* `geoHaystack`: mentioned at https://docs.mongodb.com/v4.4/tutorial/build-a-geohaystack-index/#create-a-haystack-index
* `hashed`: mentioned at https://docs.mongodb.com/v4.4/core/index-hashed/#create-a-hashed-index
* `text`: mentioned at https://docs.mongodb.com/v4.4/core/index-text/#create-text-index

vkarpov15 (approved):
Thanks for finding this issue and sorry for the inconvenience. It looks like we're missing tests on index specifications, we'll add that :+1:

PR #10563, opened 2021-08-10
Define `IndexOptions` in terms of `mongodb.IndexOptions`

JaredReisinger:
**Summary**

The current `IndexOptions` definition is missing some mongo-allowed properties, like `partialFilterExpression`.  Since mongoose simply passes the value along (after handling the added `expires` property), the type should be defined in the same way: as a simple extension to the mongo `IndexOptions` type.

**Examples**

This allows the following to work properly (in Typescript)

```ts

export const schema = new mongoose.Schema({ whatever: String, useWhatever: Boolean });

schema.index({ whatever: 1 }, { partialFilterExpression: { useWhatever: true } });
```


vkarpov15 (approved):
Great, thanks :+1:

PR #10604, opened 2021-08-25
fix: allow calling `Model.aggregate()` with options

amitbeck:
**Summary**

I want Mongoose to support passing options to `Model.aggregate()` and to not have to call `Aggregate.option()` manually in a separate operation.

**Examples**

Until now:
```js
// error
Model.aggregate([ /* pipeline */], { allowDiskUse: true }).exec() // Callback must be a function, got [object Object]

// workaround, requires 2 separate operations including storing the `Aggregate` instance as variable
const aggregate = Model.aggregate([ /* pipeline */]);
await aggregate.option({ allowDiskUse: true }).exec();
```

After this PR:
```js
Model.aggregate([ /* pipeline */], { allowDiskUse: true }).exec() // works
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -732,7 +732,8 @@ declare module 'mongoose' {
     new(doc?: AnyKeys<T> & AnyObject): EnforceDocument<T, TMethods>;
 
     aggregate<R = any>(pipeline?: any[], options?: Record<string, unknown>): Aggregate<Array<R>>;
-    aggregate<R = any>(pipeline: any[], options?: Record<string, unknown>, cb: Function): Promise<Array<R>>;
+    aggregate<R = any>(pipeline: any[], cb: Function): Promise<Array<R>>;
+    aggregate<R = any>(pipeline: any[], options: Record<string, unknown>, cb: Function): Promise<Array<R>>;
 
     /** Base Mongoose instance the model uses. */
     base: typeof mongoose;


index.d.ts
```diff
@@ -731,8 +731,8 @@ declare module 'mongoose' {
   interface Model<T, TQueryHelpers = {}, TMethods = {}> extends NodeJS.EventEmitter, AcceptsDiscriminator {
     new(doc?: AnyKeys<T> & AnyObject): EnforceDocument<T, TMethods>;
 
-    aggregate<R = any>(pipeline?: any[]): Aggregate<Array<R>>;
-    aggregate<R = any>(pipeline: any[], cb: Function): Promise<Array<R>>;
+    aggregate<R = any>(pipeline?: any[], options?: Record<string, unknown>): Aggregate<Array<R>>;
+    aggregate<R = any>(pipeline: any[], options?: Record<string, unknown>, cb: Function): Promise<Array<R>>;
```
vkarpov15, line 735: I'm pretty sure this will cause a TS error because a mandatory param can't follow an optional param. We'd need to make `options` required here, and add a separate overload for `aggregate(pipeline, cb)`

vkarpov15 (approved):
Quick comment re: TS, but otherwise LGTM

index.d.ts
amitbeck: Wow I can't believe I actually didn't notice this lol

amitbeck:
@vkarpov15 thanks for the fix, funny that I missed this.

PR #10630, opened 2021-08-27
 Support removing fields from projections

### LINKED ISSUE #10628: Allow removing fields from inclusion projection

**Feature**

Currently, you can't use `.select()` to remove fields on an inclusive query projection. Running `User.find({}).select({ name: 1, age: 1 }).select({ age: 0 })` yields a projection of `{ name: 1, age: 0 }` and the error: `MongoError: Cannot do exclusion on field age in inclusion projection`.

```js
const User = mongoose.model('User', new mongoose.Schema({
  name: String,
  age: Number,
}));

User.find({}).select({ name: 1, age: 1 }).select({ age: 0 }).exec();
```

I propose that the second `select()` call would remove the `age` field from the projection, producing a combined projection of `{ name: 1 }`. Conversely, `User.find({}).select({ name: 0, age: 0 }).select({ age: 1 }).exec();` would also remove the `age` field and give a projection of `{ name: 0 }`. I'd be willing to try to put a PR together if this change in functionality is accepted.

Node v16.8.0
Mongoose v5.13.8
MongoDB v5.0.2

canac:
Closes #10628, including new documentation.

**Summary**

This PR allows you to remove fields from a projection that have previously been added. `query.select({ a: 1, b: 1 }).select({ b: 0 })` now produces the valid projection `{ a: 1 }` instead of the invalid projection `{ a: 1, b: 0 }`.

**Examples**

Check out the tests.

### DIFF
--- a/lib/helpers/projection/isSubpath.js
+++ b/lib/helpers/projection/isSubpath.js
@@ -9,15 +9,5 @@
  */
 
 module.exports = function isSubpath(path1, path2) {
-  const path1Parts = path1.split('.');
-  const path2Parts = path2.split('.');
-  for (let i = 0; i < Math.min(path1Parts.length, path2Parts.length); ++i) {
-    // If any path parts don't match, then path1 is not a subpath of path2
-    if (path1Parts[i] !== path2Parts[i]) {
-      return false;
-    }
-  }
-  // If all corresponding parts match, but path1 has more parts than path2, then path1 is not a subpath of path2
-  // e.g. "a.b" is not a subpath of "a.b.c"
-  return path1Parts.length <= path2Parts.length;
+  return path1 === path2 || path2.startsWith(path1 + '.');
 };
--- a/lib/helpers/projection/isSubpath.js
+++ b/lib/helpers/projection/isSubpath.js
@@ -1,0 +1,23 @@
+'use strict';
+
+/*!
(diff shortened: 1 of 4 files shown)


lib/helpers/projection/isSubpath.js
```diff
@@ -0,0 +1,23 @@
...
+/*!
+ * Determines if `path2` is a subpath of or equal to `path1`
+ *
+ * @param {string} path1
+ * @param {string} path2
+ * @return {Boolean}
+ */
+
+module.exports = function isSubpath(path1, path2) {
+  const path1Parts = path1.split('.');
```
vkarpov15, line 12: I'd much rather this function be `path1 === path2 || path1.startsWith(path2 + '.') || path2.startsWith(path1 + '.')`. The `split()` function is relatively expensive performance-wise, and we try to avoid it where possible.

vkarpov15 (changes requested):
One small suggestion, looks good otherwise

lib/helpers/projection/isSubpath.js
canac: Good idea! Fixed a minor bug in your logic and incorporated that suggestion.

vkarpov15 (approved):
Going to merge this into 6.1 branch for release with 6.1.0 :+1:

PR #10633, opened 2021-08-28
refactor tests/examples to use async/await

AbdelrahmanHafez:


vkarpov15 (approved):
Great work! Any objection to me merging this? Would rather any additional work be in a sperate PR, because this PR is already making GitHub lag :)

AbdelrahmanHafez:
@vkarpov15 
Not at all, I usually only ask to review once I think it's ready to merge.

I'm planning to do more refactoring in a separate PR once this is merged.

PR #10649, opened 2021-08-31
add omitUndefined to types

harisvsulaiman:
**Summary**

This will add the missing types for `omitUndefined`.

**Examples**
```js
Model.findOneAndUpdate({}, {}, {omitUndefined: true}) // this will throw a type error 

```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1066,6 +1066,11 @@ declare module 'mongoose' {
     timestamps?: boolean;
     upsert?: boolean;
     writeConcern?: any;
+    /**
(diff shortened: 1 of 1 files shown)


vkarpov15 (commented):
This option was removed by design in v6.0.0, see changelog: https://github.com/Automattic/mongoose/blob/master/CHANGELOG.md#600-rc0--2021-08-03 .

You no longer need to set `omitUndefined: true` because that's now Mongoose's default behavior and there's no way to opt out of `omitUndefined` :+1:

PR #10670, opened 2021-09-03
Refactor tests from callbacks to async await

AbdelrahmanHafez:


test/aggregate.test.js
```diff
@@ -1277,25 +1264,23 @@ describe('aggregate: ', function() {
       const mySchema = new Schema({ name: String, qty: Number });
       mySchema.index({ qty: -1, name: -1 });
       const M = db.model('Test', mySchema);
-      M.on('index', function(error) {
+      M.on('index', async function(error) {
```
vkarpov15, line 1267: We should replace `M.on('index', async function(error) { ... })` with `await M.init(); ...`

vkarpov15 (approved):
1 quick comment, otherwise LGTM :+1:

test/aggregate.test.js
AbdelrahmanHafez: Ah yes, I've avoided modifying those tests.
`await M.init();` seems like the way to go. 👍

test/model.indexes.test.js
```diff
@@ -47,7 +47,7 @@ describe('model', function() {
       const IndexedModel = db.model('Test', Indexed);
       let assertions = 0;
 
-      await new Promise((resolve) => IndexedModel.on('index', () => resolve()));
+      await Indexed.init();
```
vkarpov15, line 50: Should be `IndexedModel.init()`

test/model.indexes.test.js
```diff
@@ -83,33 +83,28 @@ describe('model', function() {
...
-              assertions++;
-            }
-            if (index[0] === 'blogposts._id') {
-              assertions++;
-            }
-            if (index[0] === 'blogposts.title') {
-              assertions++;
-            }
-          }
+      const mongoIndexes = await UserModel.collection.getIndexes();
```
vkarpov15, line 88: Looks like `getIndexes()` resolves to an object now, `Object.values(await UserModel.collection.getIndexes())` should be what you're looking for.

vkarpov15 (changes requested):

test/model.indexes.test.js
AbdelrahmanHafez: Thanks, I'll take a look at those failing tests sometime during the weekend, then we can merge and I can work on another batch on another PR.

vkarpov15 (approved):
Nice :+1:

PR #10679, opened 2021-09-05
ci: add test for ubuntu-20.04

YC:
**Summary**
Adds test for latest Ubuntu LTS, while preserving 4.0 (which has built binaries up to 18.04). 
Can add other versions if/when necessary.


vkarpov15 (approved):
Thanks :+1:

PR #10699, opened 2021-09-09
fix(types): Model / Document definitions

aphex:
This PR allow a model to understand document vs a doc type

Given the following setup

```
interface Event {
  name: string
  code?: string
}

const EventSchema = new Schema<Event>(
  {
    code: { type: String },
  }
)

export const EventModel = model<Event>('Event', EventSchema)
```

Attempting to create the model with the following
```
  const event = new EventModel({code: '1234'})
```

Will not give us a type error, and it should as we are missing `name` from the event model. The fix in this PR to `AnyKeys` will cause this to report properly and inform the user they are missing `name`. I am not sure why `AnyKeys` was turning everything optional prior to this.

**However to continue;** 
The resulting `event` is properly an `Event` but this is incorrect as it is missing the `id`, `_id`  and `__v` fields that are automatically applied to every document. So my suggestion, for those that want full type completion with the actual ID's of their models is to adjust `index.d.ts` as I have here and then setup the following.

```
interface Event {
  name: string
  code?: string
}

// first `any` here in Document would allow for `id` typing as well!
export type EventDocument = Document<any, any, Event> & Event

const EventSchema = new Schema<Event>(
  {
    name: { type: String, required: true},
    code: { type: String },
  }
)

export const EventModel = model<EventDocument, {}, {}, Event>(
  'Event',
  EventSchema,
)```

This will allow `new EventModel(def)` to confirm `def` adheres to the `Event` interface as well as returning an `EventDocument` instead of an `Event` when doing various the various query operations (`find`, `findOne`, `findById`, etc) giving the user full document typing with `_id`, `id` and `__v` as they are actually there to work with. 

This is essential for projects that are just letting Mongo deal with ObjectIDs for unique identifiers. This also is only an enhancement as "normally" users will pass the `Event` interface in like the following

```
export const EventModel = model<Event>('Event', EventSchema)
```

Which then everything will work as normal including type checking for `Event` when creating a new `EventModel`. They simply will not get `id`. `_id` and `__v` in there code completion for `event`. Which maybe they are cool with 🤷‍♂️ .

The only issue this presents is for users that extend Document with their interfaces. like this

```
interface Event extends Document {}
```

This seems like a poor practice in general as now we have no way to know what are actual Schema properties and what are just document properties. Document also adds a ton of methods and crap we don't want. So when you go to create a new `EventModel` you could actually pass in document properties not knowing which belong to which. 

Going the route of something like PR suggests will allow us to keep separation between `DocType` and a `Document` and get real code completion and error checking all the way through creation and querying.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -107,6 +107,7 @@ declare module 'mongoose' {
     collection?: string,
     skipInit?: boolean
   ): U;
+  export function model<T, TQueryHelpers = {}, TMethods = {}, TSchemaDefinitionType = T>(name: string, schema?: Schema<any>, collection?: string, skipInit?: boolean): Model<T, TQueryHelpers, TMethods, TSchemaDefinitionType>;
 
   /** Returns an array of model names created on this instance of Mongoose. */
   export function modelNames(): Array<string>;
@@ -712,7 +713,7 @@ declare module 'mongoose' {
     discriminator<T, U>(name: string | number, schema: Schema<T, U>, value?: string | number | ObjectId): U;
   }
 
-  type AnyKeys<T> = { [P in keyof T]?: T[P] | any };
+  type AnyKeys<T> = { [P in keyof T]: T[P] | any };
   interface AnyObject { [k: string]: any }
 
   type Require_id<T> = T extends { _id?: any } ? (T & { _id: T['_id'] }) : (T & { _id: Types.ObjectId });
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by aphex** [comment]
Added new check to `Model` to ignore `readonly` properties from new model creation checking. Given the following example

```
interface Thing {
  text: string
  readonly startTime: Date
}

const ThingSchema = new Schema<Thing>(
  {
    text: { type: String },
    startTime: { type: Date, default: () => new Date() },
  }
)
```

This will allow models to be created like this
```
const thing = new ThingModel({ text: 'foo' })
```

This will not throw an error saying `startTime` needs to be defined. This works well for default fields you **never** want to set. 

Sadly I cannot come up with a great indicator of an optional field to **set**. For example assume we wanted `startTime` to have a default value, but also allow it to be provided at creation as an override. We can no longer use `readonly` here as it is writable. We could use the optional `?` indicator, which is probably the bets bet, but then anytime you query that model the document will report back an optional `startTime` which is technically false as it will always have a value... 

Guess you cant win them all 🤷‍♂️ but this helps at least the `readonly` situation where you want to only use a default value.

vkarpov15 (changes requested):
`const event = new EventModel({code: '1234'})` not giving a type error is by design. That's because `event` can get updated later before `save()`, via plugins or subsequent sets. If you want stricter typing, that's currently left up to the developer.

Either way, we can't merge this because it breaks our existing tests.

aphex:
@vkarpov15 ah totally makes sense on adding properties later, of course. So the error would come during `save` and this would be a runtime error, no way to catch this at build time I assume. I can pull out the additions that throw errors for `new Event({code: '1234'}` and just focus on the `_id`, `id`, `_v` properties of a instance of an Event. Right now its a bit of a mess to get code completion for these in a Document and LeanDocument. I will also review the tests and see what I can do. 

What are the policies on tests? Can tests not change until a major release?

vkarpov15:
@aphex re: tests, our policy is we can add to tests, but not remove or change tests in order to support new functionality without a major release.

One option to get better autocomplete is to add `_id` and `__v` to your `Event` interface.

PR #10727, opened 2021-09-15
Fix username parameter for MongoDB driver

### LINKED ISSUE #10726: MongoParseError: credentials must be an object with 'username' and 'password' properties

Bug

As described in https://mongoosejs.com/docs/connections.html#options:

```js
let connect = mongoose.connect('mongodb://localhost:27018/test', {
	user: 'admin',
    pass: 'password',
});
```

Should forward the auth to the nodejs driver. Looking at the object sent to the driver:

```js
{
  authSource: 'admin',
  auth: { user: 'admin', password: 'password' },
  driverInfo: { name: 'Mongoose', version: '6.0.5' }
}
```

Looking at https://github.com/mongodb/node-mongodb-native/blob/4.0/docs/CHANGES_4.0.0.md#authentication, it says the expected format is `username` and not `user`, as the error suggests:

`MongoParseError: credentials must be an object with 'username' and 'password' properties`.

Noting that the following works:

```js
let connect = mongoose.connect('mongodb://localhost:27018/test', {
	auth: {
		username: 'admin',
		password: 'password',
	}
});
```

Either the doc need to be updated to remove the user/pass backwards compat (which appears broken) or it needs to set `username` instead of `user` in the `auth` object.

MongoDB 3.4
Mongoose 6.0.5
NodeJS v14.17.6

Here is the code responsible for the bug:

https://github.com/Automattic/mongoose/blob/e6820eecbccb66e35d316f84287d3d1484f789dc/lib/connection.js#L742-L750

saveman71:
Fix #10726 

Didn't see any test to update

### DIFF
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -140,7 +140,7 @@ Mongoose passes options to the driver without modification, modulo a few
 exceptions that are explained below.
 
 * `bufferCommands`    - This is a mongoose-specific option (not passed to the MongoDB driver) that disables [Mongoose's buffering mechanism](http://mongoosejs.com/docs/faq.html#callback_never_executes)
-* `user`/`pass`       - The username and password for authentication. These options are Mongoose-specific, they are equivalent to the MongoDB driver's `auth.user` and `auth.password` options.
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #10728, opened 2021-09-15
Improve Document#populate documentation, tests

saveman71:
When migrating to mongoose 6 I found that the new populate syntax for `Document#populate` was missing some documentation and it was hard to migrate without having a look at the code and guessing that the syntax was the same as `Query#populate`.

I'm not so familiar with documentation, but I copied the one from `Query#populate` which seems like an improvement.

Maybe a "sub-page" could describe what arguments `Query#populate, `Model#populate` and `Document#populate` accept in a single place instead of maintaining that in 3 different places, but I'll leave that up to you.

Also added a few tests to demonstrate the arguments I was adding do work.

### DIFF
--- a/docs/migrating_to_6.md
+++ b/docs/migrating_to_6.md
@@ -282,7 +282,13 @@ However, Mongoose 6 does **not** buffer commands while a connection is disconnec
 
 <h3 id="removed-execpopulate"><a href="#removed-execpopulate">Removed `execPopulate()`</a></h3>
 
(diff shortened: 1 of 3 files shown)


saveman71:
Also, for these 3 documentation, I guess it's missing `options.populate` to do nested population?

vkarpov15:
@saveman71 that's correct, `options.populate` is the option for nested populate.

PR #10742, opened 2021-09-17
fix createdAt set condition

kaishu16:
**Summary**

If I set schema like the code below, then it won't pass the condition of 
L56 at `mongoose/lib/helpers/timestamps/setupTimestamps.js`

so I changed `!this.get(createdAt)` into `!this.$__getValue(createdAt)`to set proper timestamps.

```
const schema = Schema({
      createdAt: {
        type: Schema.Types.Number,
        get: (createdAt) => new Date(createdAt * 1000),
      },
      updatedAt: {
        type: Schema.Types.Number,
        get: (updatedAt) => new Date(updatedAt * 1000),
      },
      name: String
    }, {
      timestamps: { currentTime: () => Math.floor(now / 1000) }
    });
```


vkarpov15 (approved):
Thanks :+1:

Issue #10761, opened 2021-09-22
Invalid schema configuration: `True` is not a valid type at path `createdAt.immutable`


nicecatch:
I updated to Mongoose 6.0.7 and all of my models use `timestamps: true`

Here an example:

```
export interface IDocumentProjectRequest extends IProjectRequest, Document {}

const projectRequestSchema = new Schema(
  {
    user: { type: Schema.Types.ObjectId, ref: "users", required: true },
    diagnostic: {
      type: Schema.Types.ObjectId,
      ref: "diagnostic",
      required: true,
    },
    project: { type: Schema.Types.ObjectId, ref: "project" },
  },
  { collection: "projectRequests", timestamps: true }
);

export const ProjectRequest = mongoose.model<IDocumentProjectRequest>(
  "projectRequest",
  projectRequestSchema
);
````

but when I try to run the project I get this error

```
TypeError: Invalid schema configuration: `True` is not a valid type at path `createdAt.immutable`. See http://bit.ly/mongoose-schematypes for a list of valid schema types.
    at Schema.interpretAsType (C:\src\server\node_modules\mongoose\lib\schema.js:1019:11)
    at Schema.path (C:\src\server\node_modules\mongoose\lib\schema.js:677:27)
    at Schema.add (C:\src\server\node_modules\mongoose\lib\schema.js:495:12)
    at Schema.add (C:\src\server\node_modules\mongoose\lib\schema.js:506:12)
    at setupTimestamps (C:\src\server\node_modules\mongoose\lib\helpers\timestamps\setupTimestamps.js:36:10)
    at Schema.setupTimestamp (C:\src\server\node_modules\mongoose\lib\schema.js:1221:10)
    at new Schema (C:\src\server\node_modules\mongoose\lib\schema.js:141:8)
    at Object.<anonymous> (C:\src\serverr\src\db\models\Account.ts:7:23)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Module._compile (C:\src\server\node_modules\source-map-support\source-map-support.js:568:25)
[ERROR] 09:09:08 TypeError: Invalid schema configuration: `True` is not a valid type at path `createdAt.immutable`. See http://bit.ly/mongoose-schematypes for a list of valid schema types.
```

this seems to be linked to this pull request https://github.com/Automattic/mongoose/pull/10267
because when I manually revert
`schemaAdditions[createdAt] = { type: Date, immutable: true };`
to 
[...]

PR #10768, opened 2021-09-23
:bug: Fix error on timesStamps when typeKey is set

jclaudan:
**Summary**
Fix issue #10761


vkarpov15 (approved):
Good find, thanks :+1:

Issue #10770, opened 2021-09-23
Model.translateAlias doesn't iterate through sub-Schemas if the key wasn't an alias


frisbee09:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When using `Model.translateAlias`, if you have a __not aliased key__ that refers to a sub-schema, it will not be picked up to translate keys further along in the key

**If the current behavior is a bug, please provide the steps to reproduce.**
* Create a `FirstSchema`, where the top level field does not have an alias and refers to a `SubSchema`
* Inside the `SubSchema`, use aliases
* Perform a `model(FirstSchema).translateAlias()` and see that the `SubSchema` keys are not translated

The current `translate` function uses this iterative loop
```javascript
// in model.js, Model.translateAlias()
for (const i in fieldKeys) {
      const name = fieldKeys[i];
      if (currentSchema && currentSchema.aliases[name]) {
        alias = currentSchema.aliases[name];
        // Alias found,
        translated.push(alias);
      } else {
        // Alias not found, so treat as un-aliased key
        // This is where this problem is, since we do not push name to alias
        translated.push(name);
      }

      // Check if aliased path is a schema
      // If the key wasn't aliased, we have no way to find the subschema
      if (currentSchema && currentSchema.paths[alias]) {
        currentSchema = currentSchema.paths[alias].schema;
      }
      else
        currentSchema = null;
    }
```

**What is the expected behavior?**
I propose a one line fix
```javascript
for (const i in fieldKeys) {
      const name = fieldKeys[i];
      if (currentSchema && currentSchema.aliases[name]) {
        alias = currentSchema.aliases[name];
        // Alias found,
        translated.push(alias);
      } else {
        // Alias not found, so treat as un-aliased key
        // Here
        alias = name;
        translated.push(name);
      }

      // Check if aliased path is a schema
      if (currentSchema && currentSchema.paths[alias]) {
        currentSchema = currentSchema.paths[alias].schema;
      }
      else
        currentSchema = null;
    }
```

frisbee09:
Here is a test case:
```
it('should translate nested schema keys when the top level is not an alias', function() {
    const Syntax = new mongoose.Schema({
      s: { type: String, alias: 'syntax' }
    });

    // Note that d does not have an alias
    const Character = mongoose.model('Character', new mongoose.Schema({
      dot: { type: Syntax },
      name: { type: String, alias: '名' },
      bio: {
[...]

PR #10772, opened 2021-09-23
Amending Model.translateAliases to observe non-aliased sub schemas

frisbee09:
**Summary**

A fix for issue #10770. `translateAliases` wouldn't pick up nested sub-schemas if the key it just observed wasn't an alias itself.

**Examples**

I amended the translateAliases test case to include an example of the behaviour and made sure it failed before and passed afterwards.


vkarpov15 (approved):
Thanks :+1:

PR #10784, opened 2021-09-25
docs: remove array indexes section from FAQ

Duchynko:
**Summary**

Starting in mongoose 6x, you don't need to use markModified anymore (reference: https://mongoosejs.com/docs/migrating_to_6.html#arrays-are-proxies)

Would it make sense to remove it from FAQ? Is this going to change for the latest (6x) docs only?

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -12,48 +12,6 @@ hr {
 }
 </style>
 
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks!

Issue #10790, opened 2021-09-27
Typescript definition missing for `Query.allowDiskUse`


coyotte508:
This is a **bug**.

`allowDiskUse` was introduced for queries in https://github.com/Automattic/mongoose/commit/d65070ea95ff4956df027aec9d38d6effda51763 but the typescript definition is missing:

```
const results = SomeModel.find({}).sort({a: 1}).allowDiskUse(true); // TS error
```

PR #10791, opened 2021-09-27
fix(query): Add type definitions for allowDiskUse

coyotte508:
Fixes https://github.com/Automattic/mongoose/issues/10790


vkarpov15 (approved):
Thanks :+1:

PR #10810, opened 2021-09-29
chore: upgrade to mongodb driver 4.1.2

orgads:
And support working with other minor versions.

**Summary**

If the including package has both mongoose and mongodb installed with different versions, it leads to errors on build.

**Examples**

```
myfile.ts(40,31): error TS2345: Argument of type 'import("node_modules/mongoose/node_modules/mongodb/mongodb").ChangeStream<any>' is not assignable to parameter of type 'import("node_modules/mongodb/mongodb").ChangeStream<any>'.
  The types of 'options.session.clientOptions.autoEncrypter' are incompatible between these types.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "bson": "^4.2.2",
     "kareem": "2.3.2",
-    "mongodb": "4.1.1",
+    "mongodb": "^4.1.2",
     "mpath": "0.8.4",
     "mquery": "4.0.0",
     "ms": "2.1.2",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "bson": "^4.2.2",
     "kareem": "2.3.2",
-    "mongodb": "^4.1.2",
+    "mongodb": "4.1.2",
     "mpath": "0.8.4",
     "mquery": "4.0.0",
     "ms": "2.1.2",


orgads:
ping

vkarpov15 (approved):
Changed it to use exactly `4.1.2`, LGTM otherwise :+1:

orgads:
Why use an exact version? It will break on next mongodb upgrade. See my first comment.

vkarpov15:
Because we've had quite a few cases where minor and/or patch releases of the MongoDB node driver break something in Mongoose, and we'd like to avoid the combinatorial explosion of having to figure out whether every 6.0.9+ version of Mongoose supports MongoDB 4.2.

"If the including package has both mongoose and mongodb installed with different versions, it leads to errors on build." <-- that's a feature, not a bug. Mixing and matching ObjectId instances, Decimal128 instances, etc. between incompatible versions of MongoDB and Mongoose is unsupported and likely to cause bugs.

orgads:
> "If the including package has both mongoose and mongodb installed with different versions, it leads to errors on build." <-- that's a feature, not a bug. Mixing and matching ObjectId instances, Decimal128 instances, etc. between incompatible versions of MongoDB and Mongoose is unsupported and likely to cause bugs.

I was referring to minor/patch releases, which are not supposed to break anything. But if you say from your experience that it does break, I'll take it.

vkarpov15:
In theory minor/patch releases don't break anything, but Mongoose uses a very wide surface area of the MongoDB Node driver, so compatibility issues are not uncommon.

If you need a workaround, you can always just put `"mongodb": "4.x"` as a top-level dependency in your `package.json`, and npm should dedupe against what version of the MongoDB driver Mongoose uses.

PR #10812, opened 2021-09-29
made requested refactor

IslandRhythms:


vkarpov15:
Change it to `require('../common')`, and if that doesn't work you can just get rid of `debug.test.js`.

PR #10846, opened 2021-10-05
fix: accepts uppercase values in mongoose.isValidObjectId

foxadb:
### Summary

ObjectId are case-insensitives in MongoDB as long as the value is hexadecimal.

`mongoose.isValidObjectId` function should consider uppercase ObjectId as valid.

### Example

Adding the two following assertions to the test suite:
```js
assert.ok(mongoose.isValidObjectId('5f5c2d56f6e911019ec2acdc')); // passing before and after change
assert.ok(mongoose.isValidObjectId('608DE01F32B6A93BBA314159')); // passing only after change
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -960,7 +960,7 @@ Mongoose.prototype.isValidObjectId = function(v) {
       if (typeof v === 'string' && v.length === 12) {
         return true;
       }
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #10877, opened 2021-10-13
Update index.d.ts -> Add missing constructor typing for VersionError

StefanoA1:
Adds missing constructor typing for VersionError as it extends Error it wrongfully expects to receive a string as a parameter (not 3 parameters)

**Summary**

Adds missing constructor typing for VersionError as it extends Error it wrongfully expects to receive a string as a parameter (not 3 parameters)

**Examples**

```
// this gives the following error: Expected 1 arguments, but got 3. ts(2554)
const error = new VersionError({}, 2, ['a', 'b']);
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3088,6 +3088,8 @@ declare module 'mongoose' {
       name: 'VersionError';
       version: number;
       modifiedPaths: Array<string>;
+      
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

Issue #10876, opened 2021-10-13
Cursor 'close' event emitted twice 


iovanom:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When the cursor is used with pause/resume the event 'close' is emitted twice.

**If the current behavior is a bug, please provide the steps to reproduce.**

```javascript
cursor = TestModel.find({}).cursor();
cursor.on('data', (obj) => {
  cursor.pause();
  // simulate the async actions using obj
  setTimeout(() => cursor.resume(), 50);
});
cursor.on('close', () => console.log('close'));
```

The code below will get two 'close' in the console;

**What is the expected behavior?**

The expected behavior is to receive only one 'close' event.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
node - 14.18.0
mongoose - 5.13.11
MongoDB - 4.4.9

IslandRhythms:
close onnly fires once for me

```javascript

const mongoose = require('mongoose');

const testSchema = new mongoose.Schema({
    name: String
});

const Test = mongoose.model('Test', testSchema);



async function run() {
    await mongoose.connect('mongodb://localhost:27017/', {useNewUrlParser: true,
    useUnifiedTopology: true,});
    await mongoose.connection.dropDatabase();
    const cursor =  await Test.find({}).cursor();

    cursor.on('data', (obj) => {
      cursor.pause();
      // simulate the async actions using obj
      setTimeout(() => cursor.resume(), 50);
    });
    cursor.on('close', () => console.log('close'));
  
}


run();




```

iovanom:
> close only fires once for me
> 
> ```js
> const mongoose = require('mongoose');
> 
> const testSchema = new mongoose.Schema({
>     name: String
> });
> 
> const Test = mongoose.model('Test', testSchema);
> 
> 
> 
> async function run() {
>     await mongoose.connect('mongodb://localhost:27017/', {useNewUrlParser: true,
>     useUnifiedTopology: true,});
>     await mongoose.connection.dropDatabase();
>     const cursor =  await Test.find({}).cursor();
> 
>     cursor.on('data', (obj) => {
>       cursor.pause();
>       // simulate the async actions using obj
>       setTimeout(() => cursor.resume(), 50);
>     });
>     cursor.on('close', () => console.log('close'));
>   
> }
> 
> 
> run();
> ```

you don't have any document in collection. Have a look on my tests that I have added.

[...]

Issue #10875, opened 2021-10-12
Use of awaitdata vs. awaitData in test suite


bazineta:
The query test suite contains a test for use of `awaitdata` on a tailable cursor:

https://github.com/Automattic/mongoose/blob/b7992659d6e55d2ed9d30b71a735e2b3b535f3cd/test/query.test.js#L1318

It appears that the while the native driver in the past accepted the all-lowercase syntax as synonymous with the camel-case `awaitData`, it at present only regards the camel-case as being correct. At least, that certainly seems to be the case to me, with support for the deprecated old option removed in this commit of October 15, 2020:

https://github.com/mongodb/node-mongodb-native/commit/0e6375a22fc3d1f4ce10846b98dafaa65044ef6d

bazineta:
After some more investigation through the native driver change log, it's actually this line that documents the `awaitData` change:
```
NODE-2816: remove deprecated find options (#2571)
```
Which, chasing things down, eventually leads to this Jira item:

https://jira.mongodb.org/browse/NODE-2816

On the `tailableRetryInterval` option, there's this:

https://jira.mongodb.org/browse/NODE-2358

I haven't been able to find anything documented with respect to `numberOfRetries'.

PR #10878, opened 2021-10-13
[gh-10875] Use stream destroy method on close to prevent emit 'close'…

iovanom:
… event twice

**Summary**

These changes fix the issue with twice emitted 'close' events. 
https://github.com/Automattic/mongoose/issues/10876

**Examples**


vkarpov15 (approved):
Thanks! :+1:

iovanom:
@vkarpov15 Could you add this fix also in version 5? I'm using version 5 in prod for my project and I need this fix.

iovanom:
> This PR breaks my setup, some of my streams now never ends.. some of them do though.
> 
> my simple setup is like this: image
> 
> the finish event is in 80% not fired.

what node version do you use?

PR #10897, opened 2021-10-18
[gh-10875] Use stream destroy method on close to prevent emit 'close'…

iovanom:
… event twice

**Summary**

These changes fix the issue with twice emitted 'close' events for version 5. 
https://github.com/Automattic/mongoose/issues/10876

The PR https://github.com/Automattic/mongoose/pull/10878 with fix for lastest version is already merged.  This one is for version 5.

**Examples**


vkarpov15 (approved):
Thanks :+1:

PR #10899, opened 2021-10-18
fix(index.d.ts): More precise type for Schema.clone()

coyotte508:
**Summary**

Currently, `schema.clone()` returns a generic `Schema`, without using typing info already available on `schema`.

**Examples**


vkarpov15 (approved):
Thanks :+1:

Issue #10902, opened 2021-10-19
cursor never finishes


simllll:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
this PR https://github.com/Automattic/mongoose/pull/10878 breaks our setup, some streams now never finish.

**If the current behavior is a bug, please provide the steps to reproduce.**
It#s hard to tell what's casuing this, because some stream do finish, but some do not. It seems around 60-70% never finish.
![image](https://user-images.githubusercontent.com/5757263/137886016-1afa1c81-10d1-46da-bbd3-934920262d9f.png)

**What is the expected behavior?**
stream should fire finish event like it was before.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
mongoose 6.0.11: not working
mongoose 6.0.10: works

simllll:
It seems the only event that is fired is a "drain" event, after that..nothing happens anymore.

iovanom:
PR: https://github.com/Automattic/mongoose/pull/10906

PR #10906, opened 2021-10-20
[gh-10902] Remove the logic for emit close, autodestroy will emit clo…

iovanom:
…se. Works only for node>=12.

**Summary**
The query/aggregate cursor should emit 'end' event and the 'end' event should be emitted before 'close' event.
It can be fixed using the auto destroy feature that exists in node >= 12.
I've added the autoDestroy=true for stream options because for node 12 this option is false by default.
This fix will work for node >= 12.

**Examples**


simllll:
LGTM :-)

PR #10911, opened 2021-10-21
chore: upgrade to mongodb driver 4.1.3

orgads:


vkarpov15 (approved):
Thanks :+1: We don't do semver ranges for our deps, except for bson where we do it to line up with mongodb's bson dependency.

PR #10916, opened 2021-10-22
[gh-10902 v5] Emit end event in before close

iovanom:

**Summary**
The 'end' event should be emit before 'close' event.
See [PR](https://github.com/Automattic/mongoose/pull/10906) from master.
Issue: https://github.com/Automattic/mongoose/issues/10902#event-5500362317

**Examples**


vkarpov15 (approved):
Thanks!

PR #10942, opened 2021-10-30
Fix ts types for query set

jneal-afs:
**Summary**

The Query `set` method does not function properly in TypeScript projects. It currently only supports, the `string, value` implementation, but not the `object` only implementation. See https://github.com/Automattic/mongoose/blob/5.x/lib/query.js#L1835-L1848 for reference.

**Examples**

```ts
// Does not work
Model.updateOne({ id }).set({ name: 'new' });

// Works
Model.updateOne({ id }).set('name', 'new');
```


jneal-afs:
I should note, my project uses 5.x and that is why I opened this PR on the 5.x branch. However, this issue is also present in the master branch, so it should be cherry-picked there as well.

Issue #10349, opened 2021-06-12
Mongoose slows down performance of VSCode (TSServer) by a lot


ShadiestGoat:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

It slows down vscode by a lot, even with small projects. I have a very small project, with like 2 small schemas, and the 2 other files, containing on average 160 lines of code, and vscode slows down a lot as soon as I install mongoose. This does not happen when I install most other libraries (haven't tested them all lmao). If I uninstall mongoose, it comes back and works perfectly normal. 
To clarify "slowing down" means that typescript language server needs 3-5 seconds to realise that I'm importing somethign invalid, and just vscode in general to realise I'm importing something that's never used takes around 2-4 seconds. Usually it happens in under a second so like.. you know...
And it's not like vscode is taking a lot of ram/cpu either, it just grows slow. 
Btw I think this is a mongoose issue, since it happens with no other library, its a special case here
Reloading typescript server or project doesn't help

**If the current behavior is a bug, please provide the steps to reproduce.**

Have vscode to edit your code
Install mongoose on your current project (`npm i mongoose` 

My tsconfig.json:
```json
{
    "compilerOptions": {
        "allowJs": true,
        "moduleResolution": "Node",
        "esModuleInterop": true,
        "target": "ES6",
        "resolveJsonModule": true,
        "outDir": "./dist",
        "checkJs": true,
        "module": "CommonJS",
        "lib": ["ESNext"]
    },
    "include": ["src"]
}
```
**What is the expected behavior?**

It to not be like this, lmao.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Node: v16.2.0
Mongoose: 5.12.13
Linux: 5.12.7-arch1-1

vkarpov15:
Interesting, can you please provide an example project that demonstrates this? I haven't noticed this but I might be missing something.

ShadiestGoat:
Mm well I can replicate it with very little:
```ts
import { model, Model, Schema } from "mongoose";

export interface UserSH {
    username: string,
    password: string,
    maxMb: number,
    submitted: string[],
    id: string
}

export const UserSchema = new Schema<UserSH, Model<UserSH>, UserSH>({
    id: "string",
    maxMb: "number",
    username: "string",
    submitted: ["string"],
    password: "string"
})


uiyasdjklsadias //error this baby is undefined
```
[...]

PR #10951, opened 2021-11-03
Fix Schema's clone function typing

StefanoA1:
**Summary**
The Schema's clone function does not preserve the typing (and on its best case, it loses the name of the type and just types the cloned schema as a nameless object)

**Examples**
Before:
clone returns an untyped schema (or at best it returns an object but the type name/interface/class is lost)

After:
clone returns a typed schema which can be the original type or a type that extends the base type (would work for type extending it with partial params)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1258,7 +1258,7 @@ declare module 'mongoose' {
     childSchemas: { schema: Schema, model: any }[];
 
     /** Returns a copy of this schema */
-    clone<T extends DocType = DocType>(): Schema<T>;
+    clone<T extends Schema<DocType> = this>(): T;
 
     /** Object containing discriminators defined on this schema */
     discriminators?: { [name: string]: Schema };
--- a/index.d.ts
+++ b/index.d.ts
@@ -1258,7 +1258,7 @@ declare module 'mongoose' {
     childSchemas: { schema: Schema, model: any }[];
 
     /** Returns a copy of this schema */
-    clone<T extends Schema<DocType> = this>(): T;
+    clone<T = this>(): T;
 
     /** Object containing discriminators defined on this schema */
     discriminators?: { [name: string]: Schema };
--- a/index.d.ts
+++ b/index.d.ts
@@ -1258,7 +1258,7 @@ declare module 'mongoose' {
     childSchemas: { schema: Schema, model: any }[];
 
     /** Returns a copy of this schema */
-    clone(): this;
+    clone<T extends DocType = DocType>(): Schema<T>;
 
     /** Object containing discriminators defined on this schema */
     discriminators?: { [name: string]: Schema };


vkarpov15 (changes requested):
I'm supportive of adding a generic param to `clone()`, but I'm not sure making `clone()` return something other than `this` is the right call. For one thing, you're missing a bunch of generic parameters to `Schema`. And I can confirm that [`this` respects generics](https://www.typescriptlang.org/play?#code/MYGwhgzhAECyCeBhcUA8AVAfNA3gWAChpoA3MEALmnQG5DDjgB7AOwgBcAnAV2Hac4AKMpWoBKXA2LR2ACwCWEAHQjoAXlLk6RaAF96O0KwCmgsVTmLJO4p2PtunFjIURtxXdqkBze+jNU6NbS0HYOTi6KKlpS+gRxhMxs7NBMAEYAVurQLMYA7nBIKBCoHJzyLN6YggDk7MYcNWJeBEkcqZkATNnpGUpGuWYtbSllPV1Kvuz+zYRlSvwAqgAOy8aciJCmzUA)

Can't we do `clone<T = this>(): T`?

StefanoA1:
> I'm supportive of adding a generic param to `clone()`, but I'm not sure making `clone()` return something other than `this` is the right call. For one thing, you're missing a bunch of generic parameters to `Schema`. And I can confirm that [`this` respects generics](https://www.typescriptlang.org/play?#code/MYGwhgzhAECyCeBhcUA8AVAfNA3gWAChpoA3MEALmnQG5DDjgB7AOwgBcAnAV2Hac4AKMpWoBKXA2LR2ACwCWEAHQjoAXlLk6RaAF96O0KwCmgsVTmLJO4p2PtunFjIURtxXdqkBze+jNU6NbS0HYOTi6KKlpS+gRxhMxs7NBMAEYAVurQLMYA7nBIKBCoHJzyLN6YggDk7MYcNWJeBEkcqZkATNnpGUpGuWYtbSllPV1Kvuz+zYRlSvwAqgAOy8aciJCmzUA)
> 
> Can't we do `clone<T = this>(): T`?

I agree, but then T should ensure to extend Schema<DocType> at least (to ensure that not any type is taken as a generic here), what are your thoughts on:
`clone<T extends Schema<DocType> = this>(): T;`
then ?

vkarpov15:
Unfortunately not. 2 reasons: 1) `T extends Schema<DocType>` is missing a bunch of generic parameters, and 2) `extends` sometimes has some serious performance issues, see #10349 .

vkarpov15 (approved):

StefanoA1:
@vkarpov15 ah! interesting, do you happen to have any docs on why extends causes such perf issues? thanks for the link

vkarpov15:
Wish I knew, but I'm not experienced enough with TS internals to know why sometimes `extends` in generics is such a perf issue.

Issue #9405, opened 2020-09-09
Bug: Schema minimize option are overriden for references


maloguertin:
**bug**

**What is the current behavior?**

The minimize option of a referenced schema is overriden by it's parent.

**If the current behavior is a bug, please provide the steps to reproduce.**

Take this schema:

```js
const userSchema = mongoose.Schema({
  adress: { type: mongoose.Schema.ObjectId, ref: 'adresses' }
})

const adressSchema = mongoose.Schema({
  zip: { type: String },
  country: { type: Object, default: {}}
}, {minimize: false})
```

if I query userSchema and populate the adress field

`console.log(user)` outputs `{ adress: {zip: "1234"} }`

`console.log(user.adress)` outputs `{zip: "1234", country: {} }`

**What is the expected behavior?**

referenced schemas options should be taken into account instead of overriden.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

`mongoose`: `5.9.4`
`mongodb`: `4.2.8`
`node`: `10.20.0`

PR #10952, opened 2021-11-03
Fix minimize

IslandRhythms:


IslandRhythms:
The failing linter test is weird because that line that it is referencing has no code on it.

lib/document.js
```diff
@@ -3406,20 +3396,22 @@ Document.prototype.$toObject = function(options, json) {
...
-
   let _minimize;
+
+  // console.log('Schema Options', schemaOptions.minimize)
+  // console.log('options', options)
   if (options._calledWithOptions.minimize != null) {
     _minimize = options.minimize;
   } else if (defaultOptions.minimize != null) {
     _minimize = defaultOptions.minimize;
+  } else if (options._parentOptions != null && options.getters != null) {
```
vkarpov15, line 3410: Why `options.getters != null` here?

lib/document.js
```diff
@@ -3406,20 +3396,22 @@ Document.prototype.$toObject = function(options, json) {
...
   defaultOptions = utils.options(defaultOptions, clone(baseOptions));
   defaultOptions = utils.options(defaultOptions, clone(schemaOptions[path] || {}));
-
   // If options do not exist or is not an object, set it to empty object
   options = utils.isPOJO(options) ? clone(options) : {};
   options._calledWithOptions = options._calledWithOptions || clone(options);
-
   let _minimize;
+
+  // console.log('Schema Options', schemaOptions.minimize)
```
vkarpov15, line 3404: Get rid of the unnecessary commented out code please

vkarpov15 (commented):
Don't worry about the lint failure, I'll fix that in master

lib/document.js
IslandRhythms: it won't work without it. gh-9405 has tests already and this was the only combination that could get them all to pass.

vkarpov15:
I took a closer look and this is expected behavior. We made the change in #9405 that subdocs inherit parents' minimize and added tests that enforce that behavior, so no way to make this work. The `getters` trick works by accident rather than by design.

If you find yourself struggling to reconcile behavior that's asserted on in the tests with new behavior, feel free to ask me to clarify in Slack.

PR #10958, opened 2021-11-05
docs(document.js): fix link leading to old github pages site

PuneetGopinath:

**Summary**

The link on the site for Mixed schema type, was linking me to the old github pages page, you might have another type of fix, but this is the fix I found

**Examples**


vkarpov15 (approved):
Thanks :+1:

PR #10989, opened 2021-11-18
fix(index.d.ts): make populate type param optional

mohd-akram:
**Summary**

When mongoose is used from JS, there's no way to pass a type param; this was causing return types to always be `any`. This change makes the type parameter optional, as it was before.


vkarpov15 (approved):
Thanks :+1:

PR #10996, opened 2021-11-19
docs: fix default value of strictQuery

jonchurch:
I think this is the right place to update the docs seen on https://mongoosejs.com/docs/api.html#mongoose_Mongoose-set
**Summary**

This small part of the docs is inaccurate, I believe. 

By default, strict is set to true, and strictQuery is set to the value of strict, so strictQuery is true by default.


vkarpov15 (approved):
Thanks :+1:

PR #10997, opened 2021-11-19
Static sync indexes

IslandRhythms:


IslandRhythms:
Could probably use a few more tests but just want to make sure the gist of what I've written is correct. Also, the tests don't fail on my machine so I don't know why they are 1/2 failing here.

lib/connection.js
```diff
@@ -1372,6 +1372,14 @@ Connection.prototype.setClient = function setClient(client) {
   return this;
 };
 
+Connection.prototype.syncIndexes = async function syncIndexes() {
+  const result = [];
+  for (const model in this.models) {
+    result.push(await this.model(model).syncIndexes());
```
AbdelrahmanHafez, line 1378: Love how we can finally use async/await in the codebase.

test/connection.test.js
```diff
@@ -1078,4 +1078,28 @@ describe('connections:', function() {
     assert.deepStrictEqual(conn2.id, m1.connection.id + 1);
     assert.deepStrictEqual(conn3.id, m.connection.id + 2);
   });
+
+  it('Allows a syncIndexes option with connection mongoose.connection.syncIndexes (gh-10893)', async function() {
```
AbdelrahmanHafez, line 1082: I find it difficult to understand what the test is doing, it would probably be easier to read if we use more descriptive variable names and add some spaces to make it visually easier to read.

Following are a few suggestions:
`m` => `mongooseInstace`
`M` => `Model` or even better we can give it a name that's more likely to be used in a real-world scenario `User`.
`sync` => `indexesAfterFirstSync`
`indexes` => `indexesAfterSecondSync`


It would be easier to understand if we split each case into its own test case.

Otherwise, looks great to me, and the tests that are failing are probably flaky ones that will pass if we restart the testing stage.

AbdelrahmanHafez (changes requested):

lib/index.js
```diff
@@ -984,6 +984,11 @@ Mongoose.prototype.isValidObjectId = function(v) {
   return false;
 };
 
+Mongoose.prototype.syncIndexes = async function() {
```
vkarpov15, line 987: Let's not make this async. Just replace this with `= function syncIndexes() {`

lib/index.js
```diff
@@ -984,6 +984,11 @@ Mongoose.prototype.isValidObjectId = function(v) {
...
+  const result = await this.connection.syncIndexes();
```
vkarpov15, line 988: This won't quite work right in cases where people import mongoose like `const { syncIndexes } = require('mongoose')`

Use the same pattern as in `startSession()`:

```
```

test/index.test.js
```diff
@@ -794,5 +794,9 @@ describe('mongoose module:', function() {
       const goodIdString2 = '1'.repeat(12);
       assert.deepStrictEqual(mongoose.isValidObjectId(goodIdString2), true);
     });
+    it('Allows a syncIndexes shorthand mongoose.syncIndexes (gh-10893)', async function() {
+      const m = new mongoose.Mongoose();
+      assert([], await m.syncIndexes());
```
vkarpov15, line 799: `assert.deepEqual(await m.syncIndexes(), [])`

lib/connection.js
```diff
@@ -1372,6 +1372,14 @@ Connection.prototype.setClient = function setClient(client) {
   return this;
 };
 
+Connection.prototype.syncIndexes = async function syncIndexes() {
+  const result = [];
```
vkarpov15, line 1376: Instead of returning an array, let's make this return an object where the keys are model names and the value is a result. So `result[model.modelName] = await this.model(model).syncIndexes()`

vkarpov15 (changes requested):

lib/connection.js
Kamikadze4GAME: @vkarpov15

Maybe is it better to return object, where key is `modelName` and value is result (successful) from `this.model(model).syncIndexes()` or if something went wrong instance of the Error (or MongooseError)?

So the result object will look like:

```js
const syncIndexesResult = await connection.syncIndexes();
// `syncIndexesResult`
{
  model1: ['index-1-1', 'index-1-2'], // this is example of the successful result from `Model.syncIndexes`,
  model2: MongooseError, 
  model3: [...],
}
```

As a result, you could control future actions (ex: show in console list of failed models or manually rerun syncIndexes on failed models).

What do you think?

vkarpov15 (approved):
Nice :+1:

lib/connection.js
Kamikadze4GAME: @vkarpov15 @AbdelrahmanHafez ping

PR #11005, opened 2021-11-23
Owner doc

IslandRhythms:
comment out the owner doc function and everything will work fine


lib/document.js
```diff
@@ -3939,9 +3940,13 @@ Document.prototype.toJSON = function(options) {
   return this.$toObject(options, true);
 };
 
+
+Document.prototype.ownerDocument = function() {};
```
vkarpov15, line 3944: Make this `return this`

lib/types/subdocument.js
```diff
@@ -229,12 +229,12 @@ Subdocument.prototype.$ignore = function(path) {
  */
 
 Subdocument.prototype.ownerDocument = function() {
+  this.isSubDocument = true;
```
vkarpov15, line 232: `this.$__.isSubDocument = true`

Also, put this line in the constructor, not in `ownerDocument()`

lib/document.js
```diff
@@ -987,7 +987,8 @@ Document.prototype.$session = function $session(session) {
 
   this.$__.session = session;
 
-  if (!this.ownerDocument) {
+
+  if (!this.isSubDocument) {
```
vkarpov15, line 991: Use `this.$__.isSubDocument` instead

vkarpov15 (changes requested):
I think the reason why this PR is failing is because you're setting `isSubDocument` in `ownerDocument()` as opposed to in the subdocument constructor.

lib/document.js
IslandRhythms: Just there or all the other instances?

IslandRhythms: I ran it with only replacing that and replacing the other instances and still get a bunch of test failures

lib/document.js
```diff
@@ -3950,7 +3957,7 @@ Document.prototype.toJSON = function(options) {
  */
 
 Document.prototype.parent = function() {
-  return this.$__.parent;
+  return this.$__.parent || this;
```
vkarpov15, line 3960: This change is incorrect: a subdocument can have a null parent. At the very least, you can do `if (this.$__.isSubdocument) { return this.$__.parent; } return this;`

lib/types/subdocument.js
```diff
@@ -17,6 +17,7 @@ module.exports = Subdocument;
  */
 
 function Subdocument(value, fields, parent, skipId, options) {
+  this.$__.isSubDocument = true;
```
vkarpov15, line 20: The tests are all blowing up because of this line. You would need to put `this.$__.isSubDocument = true` **after** `Document.call(this, value, fields, skipId, options);`, because that `Document.call()` is what creates the `.$__` property.

vkarpov15 (changes requested):

IslandRhythms:
The only thing I've seen work is changing the function `ownerDoc` to a different name

vkarpov15 (changes requested):
@IslandRhythms one of the issues was that you didn't update a few cases in `lib/plugins` where we were still using `if (this.ownerDocument)` to check for subdocuments. With the most recent fixes, there's only a couple of leftover test failures. Please take a look and see if you can find any other places where we're still using `this.ownerDocument`

vkarpov15 (approved):
Thanks :+1:

PR #11006, opened 2021-11-23
Strict populate global

IslandRhythms:


lib/helpers/populate/getModelsMapForPopulate.js
```diff
@@ -17,6 +17,9 @@ const populateModelSymbol = require('../symbols').populateModelSymbol;
 const schemaMixedSymbol = require('../../schema/symbols').schemaMixedSymbol;
 
 module.exports = function getModelsMapForPopulate(model, docs, options) {
+  if (model.schema.base.options.strictPopulate == false && options.strictPopulate == null) {
```
vkarpov15, line 20: `model.schema.base.options.strictPopulate != null`. Why hardcoding false?

Also, `model.base.options` should work and be shorter.

Finally, any way to move this up to the `populate()` function in `model.js` that calls `getModelsMapForPopulate()` and shallow clone the object before modifying? A bit random to modify this option here.

vkarpov15 (changes requested):

test/index.test.js
```diff
@@ -693,6 +693,26 @@ describe('mongoose module:', function() {
     assert.ok(mongoose.isValidObjectId(new mongoose.Types.ObjectId()));
     assert.ok(!mongoose.isValidObjectId(6));
   });
+  it('allows global `strictPopulate` option (gh-10694)', async function() {
```
AbdelrahmanHafez, line 696: Splitting each test case into a different `it(...)` invocation makes it easier to understand the intention.

I'd split the test here into the following different test cases:
```js
it('global `strictPopulate` works when false')
it('`global `strictPopulate` works when true');
it('global `strictPopulate` can be overriden on a specific query set to false')
it('global `strictPopulate` can be overriden on a specific query set to true')
```

And I'd do my best to make the test case look like a real-world scenario by avoiding variable names such as `test` on line 708.

AbdelrahmanHafez:
Nice!
Other than this minor test modification, looks great to me. 👍

AbdelrahmanHafez (changes requested):
Nice!
Other than [the minor test modifications](https://github.com/Automattic/mongoose/pull/11006/files#r756568918), looks great to me. 👍

vkarpov15 (commented):
I agree with @AbdelrahmanHafez 's comments, but otherwise LGTM :+1:

AbdelrahmanHafez (approved):
Looks good to me. Thanks!

vkarpov15 (approved):
Nice, good job :+1:

Issue #10856, opened 2021-10-08
Improve support for partially denormalized subdocs, including populating denormalized subdocs


adamreisnz:
**Do you want to request a *feature* or report a *bug*?**
Request a feature

**What is the current behavior?**
When a schema is set as `strict`, it is not possible to append different fields to the model OR save those fields in the database.
When a schema is set as not `strict`, it is possible to append different fields AND save those fields in the database

**What is the proposed behavior?**
It would be nice if these two features could be split into two distinct features/flags. I have a use case where I would like to append/set additional properties to a model, and have this returned to the client when using `toJSON`, but I **don't** want these fields to be saved to the database.

Currently, this doesn't seem possible unless I disable strict for the schema altogether, and create a manual pre-save hook to clean up data that I don't want when saving. Needless to say this adds overhead and is prone to bugs if new fields are added.

**Use case**
The main use case is being able to populate denormalised data easily, which currently is not possible if using `strict` schemas.

For example:

```js
//Bar model, has a name property and some other properties that we are interested in
const BarSchema = new Schema({
  name: String,
  more: String,
  another: Number,
})
mongoose.model('Bar', BarSchema)

//Denormalised Bar schema with just the name, for use on the Foo model
const BarNameSchema = new Schema({
  _id: {
    type: Schema.Types.ObjectId,
    ref: 'Bar',
  },
  name: String,
})

//Foo model, which contains denormalized bar data (just the name
const FooSchema = new Schema({
  something: String,
  other: Number,
  bar: BarNameSchema,
})
mongoose.model('Foo', FooSchema)

//Now for some routes I want to be able to load the full `Bar` model, and set it on the `bar` property of a loaded `foo` document
const _id = ...
const foo = mongoose.model('Foo').findOne({_id})
const bar = mongoose.model('Bar').findOne({foo.bar._id})

//This won't work, mongoose will drop the additional properties because the schema is set to `strict`
foo.bar = bar 

//I could make the schema not-strict, but then saving foo will save redundant properties in the database
await foo.save() //don't want the full bar data to be saved on the model
```

So ideally, these two behaviours of `strict` could be separated somehow into a separate flag/setting. However, I am also open to other suggestions that would solve our problem.

To be honest this has been a pet peeve of mine for a few years of working with Mongoose now. It's so easy to populate data, and it's so easy to make and manage denormalized data in your database, but it seems the combination of the two still causes unnecessary friction.

[...]

PR #11008, opened 2021-11-23
feat(populate): support `ref` on subdocuments

vkarpov15:
Re: #10856

**Summary**

From the discussion of #10856, we want to support populating partially denormalized subdocuments. 

```javascript
    const blogPostSchema = Schema({
      title: String,
      author: {
        // Denormalize `_id` and `name` from User, but if you `populate('author')`, you should get the other properties
        type: new Schema({ _id: Number, name: String }),
        ref: 'User'
      }
    });
```

PR is a little rougher than I'd like, I would still like to improve a couple of things: a `cloneArrays()` function for the `if (Array.isArray(unpopulatedValue)) { ... }` case to handle deep arrays, make this less of an edge case in `getModelsMapForPopulate()`, and make sure this still works with arrays that contain subdocuments. But enough to review the general idea.

One key issue/design decision: if you `populate('author')` on a blog post, `author` will be a fully fledged populated document, **not** a subdocument. This means that updating `doc.author.name = 'foo'; await doc.save();` will not update `author.name` unless you depopulate. We can make this work in the future, but might be a bit too tricky for this PR.

@adamreisnz we'd like to hear your thoughts as well

**Examples**

### DIFF
--- a/lib/helpers/populate/assignVals.js
+++ b/lib/helpers/populate/assignVals.js
@@ -170,7 +170,7 @@ module.exports = function assignVals(o) {
     }
     if (docs[i].$__) {
       o.allOptions.options[populateModelSymbol] = o.allOptions.model;
-      docs[i].$populated(_path, o.allIds[i], o.allOptions.options);
+      docs[i].$populated(_path, o.unpopulatedValues[i], o.allOptions.options);
 
       if (valueToSet instanceof Map && !valueToSet.$isMongooseMap) {
         valueToSet = new MongooseMap(valueToSet, _path, docs[i], docs[i].schema.path(_path).$__schemaType);
--- a/lib/helpers/populate/getModelsMapForPopulate.js
+++ b/lib/helpers/populate/getModelsMapForPopulate.js
@@ -75,6 +75,23 @@ module.exports = function getModelsMapForPopulate(model, docs, options) {
     let normalizedRefPath = null;
     let schemaOptions = null;
 
+    if (schema != null && schema.instance === 'Embedded' && schema.options.ref) {
+      const data = {
+        localField: options.path + '._id',
+        foreignField: '_id'
+      };
+      let normalizedRef;
+      if (typeof schema.options.ref === 'function' && !schema.options.ref[modelSymbol]) {
+        normalizedRef = schema.options.ref.call(doc, doc);
+      } else {
+        normalizedRef = schema.options.ref;
+      }
+      const unpopulatedValue = get(doc, options.path);
+      const id = get(unpopulatedValue, ['_id']);
+      addModelNamesToMap(model, map, available, [normalizedRef], options, data, id, doc, schemaOptions, unpopulatedValue);
+      continue;
+    }
+
     if (Array.isArray(schema)) {
       const schemasArray = schema;
       for (const _schema of schemasArray) {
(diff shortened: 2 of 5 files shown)


adamreisnz:
Sweet, I will try to look into this today

adamreisnz:
I like the general idea. I think the fact that if you want to make changes you have to depopulate the document first is fine. We only tend to populate for reading data purposes anyway, so saving should not cause any issues, and with the solution of depopulating first present it seems like there's a decent workaround.

This will be very helpful for us in many different use cases!

adamreisnz:
Awesome! Will take it for a spin soon

adamreisnz:
@vkarpov15 looks like this is still not fully working as expected in 6.2.2.

The following test script is now failing to populate the `bar` subdocument when a `foo` is loaded from the database:

```js
const mongoose = require('mongoose')
const {Schema} = mongoose
const uri = `mongodb://localhost:27017/test`
console.log(`Mongoose version: ${mongoose.version}`)

//Bar model, has a name property and some other properties that we are interested in
  name: String,
  more: String,
  another: Number,
})
const Bar = mongoose.model('Bar', BarSchema)

//Denormalised Bar schema with just the name, for use on the Foo model
const BarNameSchema = new Schema({
  _id: {
    type: Schema.Types.ObjectId,
    ref: 'Bar',
  },
  name: String,
})

//Foo model, which contains denormalized bar data (just the name)
const FooSchema = new Schema({
  something: String,
  other: Number,
  bar: {
    type: BarNameSchema,
    ref: 'Bar',
  },
})
const Foo = mongoose.model('Foo', FooSchema)

//Now for some routes I want to be able to load the full `Bar` model, and set it on the `bar` property of a loaded `foo` document
async function test() {

  //Connect to database
  await mongoose.connect(uri)

  const bar = await Bar.create({
    name: 'I am Bar',
    more: 'With more data',
    another: 2,
  })
  const foo = await Foo.create({
    something: 'I am Foo',
    other: 1,
    bar,
  })

  console.log('Foo after creation:', foo)

  const fooLookup = await Foo
    .findOne({_id: foo._id})
    .populate({
      path: 'bar',
      select: 'name more another',
    })

  console.log('Populated foo looked up from db:', fooLookup)

  const bar2 = await Bar.create({
    name: 'I am another Bar',
    more: 'With even more data',
    another: 3,
  })
  const foo2 = await Foo.create({
    something: 'I am another Foo',
    other: 4,
  })

  foo2.bar = bar2
  console.log('Foo 2 with bar attached:', foo2)
}

test()
```

The output is:

```
Foo after creation: {
  something: 'I am Foo',
  other: 1,
  bar: {
    name: 'I am Bar',
    more: 'With more data',
    another: 2,
    _id: new ObjectId("621020ceea279ee27fd51e98"),
    __v: 0
  },
  _id: new ObjectId("621020ceea279ee27fd51e9a"),
  __v: 0
}
Populated foo looked up from db: {
  _id: new ObjectId("621020ceea279ee27fd51e9a"),
  something: 'I am Foo',
  other: 1,
  bar: null,
  __v: 0
}
Foo 2 with bar attached: {
  something: 'I am another Foo',
  other: 4,
  _id: new ObjectId("621020ceea279ee27fd51e9f"),
  __v: 0,
  bar: {
    name: 'I am another Bar',
    more: 'With even more data',
    another: 3,
    _id: new ObjectId("621020ceea279ee27fd51e9d"),
    __v: 0
  }
}
```

As you can see in the middle output, the foo that is looked up from the database contains `null` as value for `bar`.

It looks like it's not being saved correctly

PR #11015, opened 2021-11-24
Fix typo in populate docs

gavi-shandler:
Docs mistakenly reference foreign field "band" but the foreign field in the example is "author"


vkarpov15 (approved):
Thanks :+1:

PR #11016, opened 2021-11-26
Added note about `_id` propery by MongoDB

khairnarsaurabh23:
In Docs explain that  adding  `_id` property to document is default feature of Mongoose.
This might confuse the beginners about whether MongoDB adds its to the document or Mongoose does?
Hence added a note so the beginners will not get confused.


vkarpov15:
Mongoose adds `_id`. MongoDB would add `_id` if Mongoose didn't set it by default, so I don't think the comments in this PR are quite correct. Am I misunderstanding something?

khairnarsaurabh23:
Apologies, I misunderstood the mongoose docs.

Issue #10838, opened 2021-10-03
Unable to set a map type when a schema uses generics (conflict with lodash global map type)


RachelleJanssen:
This issues occures in version 6.0.7 after migrating from 5.9.15 using typescript 4.1.3

I'm migrating our stack and I ran into an issue with an EntityRoleSchema that has a instance method called "hasScopes". I figured out I have to make a separate interface for it that extends a Model<IEntityRole>. However, when I do this I'm getting two conflicting types on a map type. It's a field where I store translations, which can only fall within the supported Language enum (for example 'nl-NL' or 'en-US'). The following code worked fine in 5.9.15, but I'm guessing a lot more happens under the hood now that prevents this from compiling

The error that I'm getting is:

`Type '{ type: MapConstructor; of: StringConstructor; }' is not assignable to type 'SchemaDefinitionProperty<Map<Language, string>> | undefined'.
  Types of property 'type' are incompatible.
    Type 'MapConstructor' is not assignable to type 'Map<Language, string> | typeof SchemaType | Schema<any, any, any> | { clear?: SchemaDefinitionProperty<() => void> | undefined; ... 10 more ...; readonly [Symbol.toStringTag]?: SchemaDefinitionProperty<...> | undefined; } | undefined'.
      Type 'MapConstructor' is missing the following properties from type 'typeof SchemaType': cast, checkRequired, set, get`

while typing this I figured out that the MapConstructor is actually a global type coming from lodash. I guess it's being overwritten somehow but I can't figure out what to do about it.

Here is my code:

```
export interface IEntityRole {
  name: Map<Language, string>;
  protected: boolean;
  scopes?: AuthorizationScopes[];
  applicant: boolean;
  allowedProjectStatus: ProjectStatus[];
  allowedProjectPhase: Types.ObjectId[];
  filterable: boolean;
  display: EntityRoleDisplay;
}

export interface IEntityRoleModel extends Model<IEntityRole> {
  hasScope: (scopes: string[]) => boolean;
}

export const entityRoleSchema = new Schema<IEntityRole, IEntityRoleModel>(
  {
    name: { type: Map, of: String },
    protected: { type: Boolean, required: true, default: false },
    scopes: { type: [String] },
    applicant: { type: Boolean, required: true, default: false },
    allowedProjectStatus: [{ type: String, enum: getEnumValues(ProjectStatus) }],
    allowedProjectPhase: [{ type: Schema.Types.ObjectId, ref: 'EntityRole' }],
[...]

PR #11026, opened 2021-11-29
types: handle buffer type

francescov1:

**Summary**

This fixes the Buffer type issue mentioned [here](https://github.com/Automattic/mongoose/issues/10838#issuecomment-976006026). It essentially just copies the same fix that was used for Map types from the previously linked thread (this [PR](https://github.com/Automattic/mongoose/commit/317b35b7c2467e1b3d0db5e7d14110582548d600))

**Examples**


vkarpov15 (approved):
Thanks :+1:

PR #11044, opened 2021-12-03
docs(model.estimatedDocumentCount): add `await` into example to get value from the Query

olecom:
**Summary**

When doing copy/paste of the example
```
const numAdventures = Adventure.estimatedDocumentCount();
```

there is no expected numeric value assigned to the `numAdventures`. By adding `await` there will be one.

https://mongoosejs.com/docs/api.html#model_Model.estimatedDocumentCount


vkarpov15 (approved):
Thanks :+1:

PR #11055, opened 2021-12-07
Incorrect return value for UpdateOne

bbradshaw:
The docs claim that the results from updateOne() contain an "modifiedCount" field identical to Mongo, but testing with Mongoose 5.12 shows that the actual field is named "nModified" instead.

**Summary**

**Examples**

### DIFF
--- a/docs/models.md
+++ b/docs/models.md
@@ -103,7 +103,7 @@ database without returning them to your application. See the
 
 ```javascript
 Tank.updateOne({ size: 'large' }, { name: 'T-90' }, function(err, res) {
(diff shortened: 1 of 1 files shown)


vkarpov15:
That's because this behavior changed in Mongoose 6. In Mongoose 6, the field is `modifiedCount`, not `nModified`.

bbradshaw:
@vkarpov15 Perhaps this is the wrong document to edit, but this is what shows up when you are looking at the Mongoose 5 docs:

[https://mongoosejs.com/docs/5.x/docs/models.html](https://mongoosejs.com/docs/5.x/docs/models.html)

PR #11057, opened 2021-12-07
fix(index.d.ts): add boolean type for transform option

AliYusuf95:
In docs is clearly state that the default value for `toJSON()` and `toObject()` options is `{ transform: true, flattenDecimals: true }`. Also, it's possible to use `{ transform: false }` to stop document transformation.


vkarpov15 (approved):
Thanks :+1:

PR #11077, opened 2021-12-11
fix(index.d.ts): separating _id property type from other $group properties

AliYusuf95:
As stated in TypeScript docs that string [index signatures enforce that all properties match their return type](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures). It's better to separate the `_id` to achieve [any valid expression type](https://docs.mongodb.com/manual/reference/operator/aggregation/group/#mongodb-pipeline-pipe.-group), instead of having all properties have the same type `{ [op in AccumulatorOperator]?: any }`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3042,8 +3042,7 @@ declare module 'mongoose' {
 
     export interface Group {
       /** [`$group` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/group) */
-      $group: {
-        _id: any
+      $group: { _id: any } & {
         [key: string]: { [op in AccumulatorOperator]?: any }
       }
     }


vkarpov15:
I took a look at the docs you sent, but I don't understand what benefit this change achieves. Can you please provide a code sample that shows what "achieve any valid expression type" means?

AliYusuf95:
@vkarpov15 By "achieve any valid expression type" I mean that `_id` type should be an [expression](https://docs.mongodb.com/manual/meta/aggregation-quick-reference/#expressions) e.g. string. But with the current type, it's not possible to use any expression because index signatures enforce its type to `_id`.
```ts
// valid group object
const group: Group = { $group: { _id: '$someField', otherKey: { $push: '$ROOT'} } };
```
Error:
```
Type '{ _id: string; otherKey: { $push: string; }; }' is not assignable to type '{ [key: string]: { $accumulator?: any; $addToSet?: any; $avg?: any; $count?: any; $first?: any; $last?: any; $max?: any; $mergeObjects?: any; $min?: any; $push?: any; $stdDevPop?: any; $stdDevSamp?: any; $sum?: any; }; _id: any; }'.
  Property '_id' is incompatible with index signature.
    Type 'string' has no properties in common with type '{ $accumulator?: any; $addToSet?: any; $avg?: any; $count?: any; $first?: any; $last?: any; $max?: any; $mergeObjects?: any; $min?: any; $push?: any; $stdDevPop?: any; $stdDevSamp?: any; $sum?: any; }'.
```

Anyway, it seems my implementation was wrong 😅, I couldn't find a way to use index signatures without affecting `_id` type.

PR #11079, opened 2021-12-11
Fix findByIdAndUpdate behavior with undefined id

### LINKED ISSUE #10964: findByIdAndUpdate should not work with undefined id
Neither a feature or a bug, but an improvement for the API

Actually, findByIdAndUpdate(id, ...) is equivalent to findOneAndUpdate({ _id: id }, ...), as stated by the documentation.
That's working accordingly, however this is not the logically expected behavior when the id is undefined:
`findOneAndUpdate({ _id: undefined }, ...)` => this form, understandably, should update the first entry.
`findByIdAndUpdate(undefined, ...)` => this form however should fail.

I expect an argument error when using `findByIdAndUpdate` with an undefined id.

Mongoose v5.13.8

gramliu:
**Summary**

findByIdAndUpdate should not work with an undefined id, if an update is specified. Resolves #10964

**Examples**

I added a test to demonstrate this change.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2621,6 +2621,10 @@ function _decorateUpdateWithVersionKey(update, options, versionKey) {
  *     A.findByIdAndUpdate()                     // returns Query
  *
  * ####Note:
+ * 
+ * If id is undefined and update is specified, an error will be thrown.
+ * 
+ * ####Note:
  *
  * All top level update keys which are not `atomic` operation names are treated as set operations:
  *
@@ -2676,6 +2680,9 @@ Model.findByIdAndUpdate = function(id, update, options, callback) {
       throw new TypeError(msg);
     }
     return this.findOneAndUpdate({ _id: id }, undefined);
+  } else if (id === undefined) {
+    const msg = 'Model.findByIdAndUpdate(): id cannot be undefined if update is specified.\n';
+    throw new TypeError(msg);
   }
 
   // if a model is passed in instead of an id
--- a/lib/model.js
+++ b/lib/model.js
@@ -2680,8 +2680,8 @@ Model.findByIdAndUpdate = function(id, update, options, callback) {
       throw new TypeError(msg);
     }
     return this.findOneAndUpdate({ _id: id }, undefined);
-  } else if (id === undefined) {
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -2676,6 +2680,9 @@ Model.findByIdAndUpdate = function(id, update, options, callback) {
       throw new TypeError(msg);
     }
     return this.findOneAndUpdate({ _id: id }, undefined);
+  } else if (id === undefined) {
+    const msg = 'Model.findByIdAndUpdate(): id cannot be undefined if update is specified.\n';
```
vkarpov15, line 2684: Why `\n`?

lib/model.js
```diff
@@ -2676,6 +2680,9 @@ Model.findByIdAndUpdate = function(id, update, options, callback) {
       throw new TypeError(msg);
     }
     return this.findOneAndUpdate({ _id: id }, undefined);
+  } else if (id === undefined) {
```
vkarpov15, line 2683: Our tests rely on `Model.findByIdAndUpdate()` with no arguments working, so can we make this `else if (arguments.length > 0 && id === undefined)`?

vkarpov15 (changes requested):
I think this PR is helpful, but causes some test failures. We can merge this when those get sorted out

lib/model.js
gramliu: Oh, I was just following the format of the preceding error messages. I went ahead and removed it

lib/model.js
gramliu: Thank you for bringing this up. I forgot to consider that. I went ahead and added this to the condition

gramliu:
@vkarpov15 I pushed some changes that might resolve the failing tests. For the function itself, I added the check for undefined id only if the length of arguments was greater than 1. Moreover, I updated the test I wrote to properly check that the error is thrown. It turns out that my original test was wrong

vkarpov15 (approved):
Needed to patch up the tests but LGTM overall :+1:

vkarpov15:
Unfortunately I have to revert this change - it is backwards breaking as indicated in #11149. We'll consider this for our next major release.

PR #11091, opened 2021-12-14
Update subdocs.md

warlock1996:
add schema to model function


vkarpov15 (approved):
Thanks :+1:

vkarpov15:
I took a closer look and this change is incorrect. The code as written is intended to access the `Parent` model, `mongoose.model('Parent')` gets the model named 'Parent'.

warlock1996:
maybe but it throws an error with one argument

PR #11098, opened 2021-12-14
fix replaceRoot type

alibehroozi:
**Summary**

The replaceRoot type was incorrect. The current type requires `newRoot` to be passed to `replaceRoot` which results in a wrong output in an aggregation.


vkarpov15 (approved):
Thanks :+1:

PR #11102, opened 2021-12-15
fix(index.d.ts) treat _id like other fields in $project

### LINKED ISSUE #11088: [TypeScript] PipelineStage.Project interface definition is wrong
Introduced in Mongoose **6.1.0** PipelineStage.Project has incomplete definition:

```
export interface Project {
  /** [`$project` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/project/) */
  $project: { _id?: 0 | false; [field: string]: any }
}
```
And must be updated with: 
```
_id?: 0 | false | 1 | true
```

Because in some cases I need only ```_id``` in projection and nothing else. MongoDB allows to do that:
```
{
  $project: {
    _id: 1,
  }
}
```

### LINKED ISSUE #11099: Getting typings error in $group and $project stages of aggregation after upgrading to 6.1.1

**bug**

I am getting the following error in $group
```
Type '{ _id: string; root: { $first: string; }; amount: { $sum: string; }; }' is not assignable to type '{ [key: string]: { $min?: any; $max?: any; $addToSet?: any; $push?: any; $count?: any; $accumulator?: any; $avg?: any; $first?: any; $last?: any; $mergeObjects?: any; $stdDevPop?: any; $stdDevSamp?: any; $sum?: any; }; _id: any; }'.
```

and getting following error in $project

```
Type 'string' is not assignable to type 'false | 0'.
```

Please note there could be some cases where devs can use the first object's ID 

You can try running the script with TS node. It will give you error

```typescript
Model.aggregate([
        { $match: { createdAt: { $gte: startDate, $lte: endDate } } },
        { $addFields: { date: { $dateToString: { date: '$createdAt', format: format.mongoose } } } },
        { $group: { _id: '$date', root: { $first: '$$ROOT' }, amount: { $sum: '$amount' } } },
        { $sort: { _id: 1 } },
        { $project: { _id: '$root._id', date: '$_id', close: '$amount' } },
])
```

Well it should not throw any error

Nodejs - v14.18.2

Mongoose: 6.1.1

Since this is a mongoose typings bug, mongodb version is not required

rpenido:
Fix #11088 
Fix #11099

Changed type to
```
$project: { _id?: 0 | false; [field: string]: any } | {_id: 1 | true}
```

:heavy_check_mark:  OK
```
{
  $project: {
    _id: 1,
  }
}
```

:heavy_check_mark:  OK
```
{
  $project: {
    _id: 0,
    someField: 1
  }
}
```

:x: Error
```
{
  $project: {
    _id: 1,
    someField: 1
  }
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3115,7 +3115,7 @@ declare module 'mongoose' {
 
     export interface Project {
       /** [`$project` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/project/) */
-      $project: { _id?: 0 | false; [field: string]: any } | {_id: 1 | true}
+      $project: { [field: string]: any }
     }
 
     export interface Redact {
--- a/index.d.ts
+++ b/index.d.ts
@@ -3115,7 +3115,7 @@ declare module 'mongoose' {
 
     export interface Project {
       /** [`$project` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/project/) */
-      $project: { _id?: 0 | false; [field: string]: any }
+      $project: { _id?: 0 | false; [field: string]: any } | {_id: 1 | true}
     }
 
     export interface Redact {


Uzlopak:
_id can also be null and could be an issue if strictNullChecks is true.

rpenido:
> _id can also be null and could be an issue if strictNullChecks is true.

Both cases below also work because of the `?` in `_id?` type declaration


```
{
  someField: 1
}
```

```
{
  _id: undefined,
  someField: 1
}
```

Was that your concern? Could you please clarify?

Thank you!

Uzlopak:
You are  right. This is not an issue. Good Work.

intenser:
> $project: { _id?: 0 | false; [field: string]: any } | {_id: 1 | true}

I'm not sure that is a correct definition.  MongoDB allows a lot of variations for _id:

```{$project:{_id: "some text"}}```

```{$project:{_id: "$referOtherField"}}```

```{$project:{_id: {$concat: ["$firstName", "$lastName"]}}}``` // expressions

Or even:
```
{ $project: {
    _id: {
      $cond: {
        if: { $eq: ['server', '$deviceType'] },
        then: '$$REMOVE',
        else: '$_id'
      }
    },
    deviceType: 1
}}
```

rpenido:
I agree with you.

So I don't think that we should treat `_id` differently.

rpenido:
fix #11099 (the $project error)

tbhaxor (approved):

tbhaxor:
This would work. Can we merge this and release it in next version

vkarpov15 (approved):
Thanks :+1:

PR #11104, opened 2021-12-15
fix _id field of PipelineStage.Group

Uzlopak:
This PR fixes the PipelineStage.Group

_id can be set to null, which is not covered by the type definition as it wants to enforce AccumulatorOperators.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3054,8 +3054,8 @@ declare module 'mongoose' {
 
     export interface Group {
       /** [`$group` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/group) */
-      $group: { _id: null | any } | {
-        _id: null | any
+      $group: { _id: any } | {
+        _id: any
         [key: string]: { [op in AccumulatorOperator]?: any }
       }
     }
--- a/index.d.ts
+++ b/index.d.ts
@@ -3055,6 +3055,7 @@ declare module 'mongoose' {
     export interface Group {
       /** [`$group` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/group) */
       $group: { _id: null | any } | {
+        _id: null | any
         [key: string]: { [op in AccumulatorOperator]?: any }
       }
     }
--- a/index.d.ts
+++ b/index.d.ts
@@ -3054,9 +3054,9 @@ declare module 'mongoose' {
 
     export interface Group {
       /** [`$group` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/group) */
-      $group: {
+      $group: { _id: null | any } | {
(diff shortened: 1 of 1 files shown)


alibehroozi:
Still doesn't work for me.

Uzlopak:
What is your use case?

I currently have to do this to have no typing error

 ```typescript
   const group: PipelineStage.Group = {
      $group: {
        _id: null as any,
        credit: {
          $sum: "$credit",
        },
        debit: {
          $sum: "$debit",
        },
        count: {
          $sum: 1,
        },
      },
    };
```

**Comment by alibehroozi** [comment]
The major problem with this Group type is that it doesn't work if _id is of type string.
We need to somehow exclude _id from `[key: string]`

**Comment by Uzlopak** [comment]
@alibehroozi 
Try again plz

**Comment by Uzlopak** [comment]
should fix #11099

**Comment by Uzlopak** [comment]
should also fix #11059

**Comment by alibehroozi** [comment]
Works for me.
Thanks.

**Comment by rpenido** [comment]
This solves the typescript error, but allows this pattern:
```
{
  $group: {
    someField: {$sum: 1}
  }
}
```

The code above will raise the following error:
```
Error scripts: MongoServerError: a group specification must include an _id
```

**Comment by Uzlopak** [comment]
and now?

isnt there any unit tests for typings like tsd?

File: index.d.ts [hunk 1]
```diff
@@ -3054,8 +3054,8 @@ declare module 'mongoose' {
 
     export interface Group {
       /** [`$group` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/group) */
-      $group: {
-        _id: any
+      $group: { _id: null | any } | {
+        _id: null | any
```
rpenido, line 3058:
```suggestion
      $group: { _id: any } | {
        _id: any
```
```suggestion
      $group: { _id: null | any } | {
        _id: null | any
```

rpenido (commented):
@Uzlopak Check if it makes sense to you!

index.d.ts
rpenido, line 3058:
```suggestion
      $group: { _id: any } | {
        _id: any
```

rpenido (commented):
@Uzlopak Check if it makes sense to you!

vkarpov15:
Fixed in #11102 , sorry for the inconvenience!

Uzlopak:
@vkarpov15 

This is regarding group stage. The other PR is regarding project stage. Is it really fixed with the other PR?

flodaniel:
I would love to see this added to Milestone 6.1.3
For us this is the blocker from upgrading v5 to v6.

vkarpov15:
@flodaniel @Uzlopak  can you please open a new issue and follow the issue template if you're still having issues? These comments aren't very helpful.

PR #11133, opened 2021-12-23
fix(index): export ConnectionStates

orgads:
**Summary**
It is declared as an enum in ts definitions, but did not really work.

**Examples**
This code passes ts compilation, but crashes on runtime:

```typescript
import { Connection, ConnectionStates } from 'mongoose';
// ...
return (this.connection.readyState !== ConnectionStates.connected);
```


vkarpov15 (approved):
Thanks :+1:

PR #11124, opened 2021-12-20
fix(index.d.ts): fix typings in Group stage

DavideViolante:
As discussed in #11059. Thanks to @rpenido @jeremyben.

Fixes #11059 and #11067.

TypeScripts experts reviews are appreciated.

Issue #11059, opened 2021-12-08
Typescript: Group interface for type PipelineStage has conflicting index signatures


meteohr:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

If you provide a "group" stage with the mandatory `_id` property in your aggregration pipeline, typescript will throw an error:
```
TS2322: Type '{ _id: string; imageResource: { $first: string; }; }'
is not assignable to type
'{ [key: string]: { $count?: any; $accumulator?: any; $addToSet?: any; $avg?: any; $first?: any; $last?: any; $max?: any; $mergeObjects?: any; $min?: any; $push?: any; $stdDevPop?: any; $stdDevSamp?: any; $sum?: any; }; _id: any; }'.
Property '_id' is incompatible with index signature.
Type 'string' has no properties in common with type '{ $count?: any; $accumulator?: any; $addToSet?: any; $avg?: any; $first?: any; $last?: any; $max?: any; $mergeObjects?: any; $min?: any; $push?: any; $stdDevPop?: any; $stdDevSamp?: any; $sum?: any; }'.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```
 const aggregateSomething: PipelineStage = {
        $group: {
            _id: 'name',
            imageResource: { $first: 'something' },
        }
    }
```

tsconfig:
```
{
    "compilerOptions": {
        "target": "es2017",
        "module": "commonjs",
        "moduleResolution": "node",
        "outDir": "./dist",
        "experimentalDecorators": true,
        "emitDecoratorMetadata": true,
        "sourceMap": true,
        "pretty": true,
        "allowJs": false,
        "strict": true,
        "esModuleInterop": true,
        "isolatedModules": true,
        "skipLibCheck": true,
        "forceConsistentCasingInFileNames": true,
        "noUnusedLocals": true,
        "noUnusedParameters": true,
        "noUncheckedIndexedAccess": true,
        "noFallthroughCasesInSwitch": true,
        "noImplicitReturns": true,
        "types": ["jest"],
        "lib": ["esnext"],
        "importHelpers": true,
        "resolveJsonModule": true
    },
    "include": ["src/**/*", "tests/**/*"],
    "exclude": ["node_modules"]
}
```
**What is the expected behavior?**
Property `_id` accepts `any`, instead of only accepting the object with `AccumulatorOperator` as type. 

[...]

PR #11140, opened 2021-12-24
Follow-up fix of TS index signature definition for Group type.

EugeneKorshenko:
**Summary**
This PR is the follow-up PR for: #11059 and for #11124 
All explicitly defined keys should always follow the type defined at the index signature. So having `any` as a type for `_id` was just a hidden error.
Even more, `_id` as a key in the member of union type was redundant.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3058,7 +3058,7 @@ declare module 'mongoose' {
 
     export interface Group {
       /** [`$group` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/group) */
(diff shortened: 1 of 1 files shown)


EugeneKorshenko:
@vkarpov15 please, take a quick look once you have a chance🙂

DavideViolante:
It seems correct, my snippets from [#11059](https://github.com/Automattic/mongoose/issues/11059#issuecomment-992334441) works with this changes too.

vkarpov15 (approved):
Thanks :+1:

Issue #11138, opened 2021-12-23
Model.exists() returns a promise and resolves to a boolean


ahmedelshenawy25:
**Do you want to request a *feature* or report a *bug*?**
Bug confirmed by @AbdelrahmanHafez 

**What is the current behavior?**
`Model.exists()` returns a promise and resolves to a boolean

**If the current behavior is a bug, please provide the steps to reproduce.**

```js
import mongoose from 'mongoose';
const { Schema } = mongoose;
import assert from 'assert';


run().catch(console.error);

async function run () {
  await mongoose.connect('mongodb://localhost:27017/test');

  await mongoose.connection.dropDatabase();

  const userSchema = new Schema({
    name: { type: String }
  });

  const User = mongoose.model('User', userSchema);
  const user = await User.create({});

  const existingUser = await User.exists({ _id: user._id });
  assert.ok(existingUser._id);
  assert.ok((typeof existingUser) !== 'boolean');

  console.log(mongoose.version);
  console.log('All assertions passed.');
}
```

**What is the expected behavior?**
`Model.exists()` should return a query and resolve to a doc or null #8097 

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Node.js: 16.13.1
Mongoose: 6.1.3
MongoDB: 5.0.5

PR #11142, opened 2021-12-26
fix(model): make exists(...) return lean document with `_id` or `null` instead of boolean

AbdelrahmanHafez:
re #11138


Uzlopak:
This will definitely be a breaking change. I know a codebase where exists is checked to be a resolved boolean.

AbdelrahmanHafez:
@Uzlopak 
It is a breaking change that was supposed to land on mongoose v6.0, it didn't land due to a bug, but it was mentioned in the changelog, and people would be expected to change their code accordingly.

@vkarpov15 What do you think is the best action here? I think we can ship it in a minor release and make a note about it.

vkarpov15:
I'll merge this into the 6.2 branch in the interest of caution. This is a messy case where the docs say one thing, the TypeScript bindings say something else, and the actual behavior is inconsistent between the two. So let's put this in a minor release to minimize risk, but also make it so the actual behavior and TS bindings match what is in our docs.

Maximusya:
https://mongoosejs.com/docs/migrating_to_6.html is still missing this breaking change.
Though it hit us not when migrating from 5 to 6(.1), but when migrating from 6.1 to 6.2

PR #11168, opened 2022-01-03
fix(document): allow populating deeply nested models as strings

### LINKED ISSUE #11160: Mongoose 6: Deep-populate with string causing errors (undocumented breaking change?)
I'm not sure if this may be documented as a breaking change somewhere, but I couldn't find it: When using a string-based populate, it fails with Mongoose 6 while it works with 5.x:
```js
await MyModel.findById(myId).populate({ path: 'subModel', populate: 'subField' });
```
The same works with brackets:
```js
await MyModel.findById(myId).populate({ path: 'subModel', populate: ['subField'] });
```
It's fine to migrate, but devs may not know it's an issue until they run a related query.

AbdelrahmanHafez:
fixes #11160

@vkarpov15 
This change means that we won't report the full path if a user uses `populate: 'websiteId'` because we can't assign a property to a string.

Can you think of a way to overcome this?
Even with this caveat, I think it's best we have populate in the first place, then we can work on improving the error message being reported.


vkarpov15 (changes requested):
Do `if (subPopulate != null && typeof subPopulate === 'object')`

Remember that `typeof null === 'object'`

lib/model.js
```diff
@@ -4627,9 +4627,10 @@ function _execPopulateQuery(mod, match, select, assignmentOpts, callback) {
       for (const pop of subPopulate) {
         pop._fullPath = basePath + '.' + pop.path;
       }
-    } else if (subPopulate != null) {
+    } else if (typeof subPopulate === 'object') {
```
AbdelrahmanHafez, line 4630:
```suggestion
    } else if (subPopulate != null && typeof subPopulate === 'object') {
```

AbdelrahmanHafez:
@vkarpov15 
Good catch, thanks.

Kamikadze4GAME:
@vkarpov15, at first I also thought that this check is needed.  But then I looked around the code and realized that this check was unnecessary. [Line 4614](https://github.com/AbdelrahmanHafez/mongoose/blob/c8fd2cdd692745903fc5b9ad4e5da50136795fd4/lib/model.js#L4614)  already excludes falsy values. Is not it?

vkarpov15:
@Kamikadze4GAME is right, thanks for pointing that out. I'll merge this and fix in master so we can get this out.

vkarpov15 (approved):

PR #11177, opened 2022-01-04
Update docs to include the section about populating multiple paths in middleware

aleksadjukic:
Added a section that explains how to populate multiple paths in middleware.

**Summary**

This is a pretty common use case, yet it is not really straight forward to achieve, so I believe that the solution to the problem should definitely be in the documentation.

### DIFF
--- a/docs/populate.md
+++ b/docs/populate.md
@@ -807,13 +807,22 @@ MySchema.post('save', function(doc, next) {
 
 <h3 id="populating-multiple-paths-middleware"><a href="#populating-multiple-paths-middleware">Populating Multiple Paths in Middleware</a></h3>
 
-Populating multiple paths in middleware is just a little bit trickier than what you may think. Here's how you may expect it to work:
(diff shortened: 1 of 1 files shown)


docs/populate.md
```diff
@@ -804,6 +805,31 @@ MySchema.post('save', function(doc, next) {
 });
 ```
 
+<h3 id="populating-multiple-paths-middleware"><a href="#populating-multiple-paths-middleware">Populating Multiple Paths in Middleware</a></h3>
+
+Populating multiple paths in middleware is just a little bit trickier than what you may think. Here's how you may expect it to work:
+
+```javascript
+userSchema.pre('find', function (next) {
```
vkarpov15, line 813: Would be great to add the schema definition to this example: it isn't clear that `followers` and `following` reference `User`

vkarpov15 (changes requested):

docs/populate.md
aleksadjukic: Agreed, will do :)

vkarpov15 (approved):
Thanks :+1:

PR #11186, opened 2022-01-06
Gh 11071

IslandRhythms:
Needed to handle booleans

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2218,7 +2218,7 @@ Document.prototype.isSelected = function isSelected(path) {
   if (this.$__.selected == null) {
     return true;
   }
-  if(!path) {
+  if (!path) {
     return false;
   }
   if (path === '_id') {
--- a/lib/document.js
+++ b/lib/document.js
@@ -2215,7 +2215,6 @@ Document.prototype.isInit = function(path) {
  */
 
 Document.prototype.isSelected = function isSelected(path) {
-  console.log(path)
   if (this.$__.selected == null) {
     return true;
   }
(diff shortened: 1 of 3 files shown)


IslandRhythms:
I'm not sure how to fix this linter issue because it says that `_model` is an unexpected token but it is used in the test. ¯\_(ツ)_/¯

lib/document.js
```diff
@@ -2218,7 +2218,9 @@ Document.prototype.isSelected = function isSelected(path) {
   if (this.$__.selected == null) {
     return true;
   }
-
+  if (!path) {
```
vkarpov15, line 2221: This is a coarse solution. This check is helpful, but you should also add a check in `$__version()` to avoid calling `isSelected()` if `key === false`

test/model.test.js
```diff
@@ -7562,6 +7562,70 @@ describe('Model', function() {
...
+        static _instance
+        _model;
+    
+        /**
+         * @private
+         */
+        constructor() {}
+    
+        static create() {
+            if (this._instance) {return this._instance;}
```
vkarpov15, line 7577: Please simplify this test case:

1) Remove the class and methods. Just inline all the method calls.
2) Remove the `try/catch` around `insertMany()`

vkarpov15 (changes requested):

IslandRhythms:
Something I just noticed, should I add a check in document.js for `isSelected` that returns false if `path` is a boolean? Otherwise it will throw the typerror because the subsequent methods cannot be called on a boolean.

vkarpov15 (approved):
Good job :+1:

Issue #11178, opened 2022-01-05
Bug: Typescript compilation broken for mongoose version >= 6.0.0


Francesco-Lanciana:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When running `tsc` on a file that contains a mongoose import I get the following error:

```
node_modules/mongodb/mongodb.d.ts:3296:5 - error TS2416: Property 'end' in type 'GridFSBucketWriteStream' is not assignable to the same property in base type 'Writable'.
  Type '{ (): void; (chunk: Buffer): void; (callback: Callback<void | GridFSFile>): void; (chunk: Buffer, callback: Callback<void | GridFSFile>): void; (chunk: Buffer, encoding: BufferEncoding): void; (chunk: Buffer, encoding: BufferEncoding, callback: Callback<...>): void; }' is not assignable to type '{ (cb?: () => void): this; (chunk: any, cb?: () => void): this; (chunk: any, encoding: BufferEncoding, cb?: () => void): this; }'.
```

**If the current behavior is a bug, please provide the steps to reproduce.**
Pull the following [repository](https://github.com/Francesco-Lanciana/mongoose-bug) and run `npm install` and `npm run start`

**What is the expected behavior?**
No errors should appear

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Node.js -> 14.15.2
Mongoose -> 6.1.5

danchistyakov:
I also face this problem. Any ideas?

wejd:
@danchistyakov  @Francesco-Lanciana  if you are using the latest version of @types/node try to downgrade to the v17.0.5, i had the same issue and this solution solves my problem

moritz157:
@wejd Yes, it's a problem in the mongodb node package caused by the latest version of @types/node as explained here: mongodb/node-mongodb-native#3088.

It has been fixed in the latest release, therefore I suggest mongoose should update it's mongodb dependency to `4.3.0`

vkarpov15:
Peg `@types/node` to < 17.0.5. The issue with [this repo](https://github.com/Francesco-Lanciana/mongoose-bug/blob/main/package-lock.json) is that, for some reason, your lockfile pulls in `@types/node@17.0.7` even though Mongoose 6.1.5 enforces < 17.0.5.

PR #11188, opened 2022-01-07
Update mongodb dependency to 4.3.0

moritz157:

**Summary**

The latest version of @types/node introduced a typescript error into node-mongodb-native as described here: mongodb/node-mongodb-native#3088. This also affects mongoose as reported in #11178.

The bug has been solved in the latest release of the `mongodb` package. Therefore mongoose should update it's mongodb dependency to `4.3.0` as done in this PR.

**Examples**


vkarpov15 (approved):
LGTM, but we're going to merge this into the 6.2 branch because we always make minor releases when we bump mongodb minor versions.

Issue #11172, opened 2022-01-03
doc.isModified() wrong when _id: false and updated this way


johnpeb:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**

`doc.isModified()` is incorrectly returning false when a subdocument is modified under an array with `_id` disabled and when the array is completely assigned to a value that is identical, but then later an array member is changed.

See REQUIREMENT 1 and REQUIREMENT 2 below. 

**If the current behavior is a bug, please provide the steps to reproduce.**

Run the script below: `npx ts-node scripts/mongoose.ts`

Script.ts:
```
import { Schema } from 'mongoose';
import * as mongoose from 'mongoose'

run().catch(err => console.log(err));

async function run(): Promise<void> {
  await mongoose.connect('mongodb://localhost:27017/test', {
  });

  await mongoose.connection.dropDatabase();

  const Order = mongoose.model('Order', new Schema({
    cumulativeConsumption: [{
      _id: false, // REQUIREMENT 1 <<<<<
      unit: String,
      value: Number
    }]
  }));

  await Order.create({
    cumulativeConsumption: [{ unit: 'foo', value: 123 }, { unit: 'bar', value: 42 }]
  });

  const doc = await Order.findOne();
  doc.cumulativeConsumption = doc.toObject().cumulativeConsumption // REQUIREMENT 2 <<<<<
  // console.log(doc.modifiedPaths())
  const match = doc.cumulativeConsumption.find(o => o.unit === 'bar');
  match.value = 43;
  match.unit = 'baz'
  console.log(`Modified: ${doc.isModified()}, expecting true`)

  process.exit()
}
```

**What is the expected behavior?**

1. `modifiedPaths()` should reflect the change after the `  match.value = 43;` line
2. `isModified()` should be true after the  `match.value = 43;` line

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Mongoose 6.1.4, node v12.22.7, mongo v4.2.8

AbdelrahmanHafez:
Nice, thanks! 👍

PR #11196, opened 2022-01-08
fix(document): make array assignment trigger isModified: true

AbdelrahmanHafez:
re #11172

Marking a path as `isModified` when its state is `init` breaks other test cases,

The issue here is as follows:
```js
const productSchema = new Schema({ _id: false, price: Number });
const orderSchema = new Schema({ products: [productsSchema] });

const order = new Order({ products: [{ price: 42 }] });
// mongoose can't detect this change, assuming that's because it compares with `assert.equals(...)`
// since there is no `_id` in both cases , an array with an object { price: 42 } is the same.
order.products = [{ price: 42 }];
order.products[0].price = 10; // mongoose can't detect this change either
```

@vkarpov15 I'm assuming here that the issue would be solved if we can make the line `order.products = [{ price: 42 }]` change the state of `products` to `modify`.

Can you point me in the right direction on where that change might be?

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2057,21 +2057,34 @@ Document.prototype[documentModifiedPaths] = Document.prototype.modifiedPaths;
 
 Document.prototype.isModified = function(paths, modifiedPaths) {
   if (!paths) {
-    return this.$__.activePaths.some('modify');
+    return this.$__.activePaths.some('modify') || this.$__.activePaths.some('init');
   }
   if (!Array.isArray(paths)) {
     paths = paths.split(' ');
   }
   const modified = modifiedPaths || this[documentModifiedPaths]();
   const directModifiedPaths = Object.keys(this.$__.activePaths.states.modify);
+  const directInitializedPaths = Object.keys(this.$__.activePaths.states.init);
   const isModifiedChild = paths.some((path) => !!~modified.indexOf(path));
+  if (isModifiedChild) {
+    return isModifiedChild;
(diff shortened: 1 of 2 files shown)


vkarpov15:
@AbdelrahmanHafez I'd say that `order.products = [{ price: 42 }]` shouldn't mark `products` as modified, because the array hasn't changed. General rule of thumb is that, if setting a property to a value that is deep equal to the existing value, Mongoose shouldn't mark as modified to avoid sending unnecessary updates in `save()`.

`order.products[0].price = 10` _should_ mark `products.0.price` as modified, but I'll need to investigate more to know why that's happening.

AbdelrahmanHafez:
@vkarpov15 
I agree.
Following are a couple of more pieces of information on the current behavior that might help:
```js
// scenario 1
order.products = [{ price: 42 }]; // doesn't mark the path as modified
order.products[0].price = 43 // still doesn't work

// scenario 2
order.products = [{ price: 43 }]; // marks it as modified
order.products[0].price = 100 // now works

// scenario 3, product has `price` and `name`
// db has `products: [{ name: 'computer', price: 42 }]
// changing the properties order doesn't mark it as modified
order.products = [{ price: 42, name: 'computer' }]  // doesn't work
```

vkarpov15:
Closed because #11172 is now fixed

AbdelrahmanHafez:
Nice, thanks! 👍

PR #11176, opened 2022-01-04
types: Fix return type of `projection` method

EugeneKorshenko:
**Summary**

`Query.projection` should return `this` for correct completion while method chaining.

Introduced here: https://github.com/Automattic/mongoose/commit/d7aaa5568330e2dcadde7ae6847bff1522f3ae4e

PR #11210, opened 2022-01-10
types: Fixes and improves `Query.projection` method signature

EugeneKorshenko:
**Summary**

Fixes type mismatch introduced here #11176
Auto completion for `Query.projection` method should be more efficient and also allows dot notation for nested fields, etc.


EugeneKorshenko:
@vkarpov15 please, take a quick look once you have a chance :slightly_smiling_face:

index.d.ts
```diff
@@ -2117,6 +2117,8 @@ declare module 'mongoose' {
 
   type UnpackedIntersection<T, U> = T extends (infer V)[] ? (V & U)[] : T & U;
 
+  type ProjectionFields<DocType> = {[Key in keyof Omit<LeanDocument<DocType>, '__v'>]?: number} & Record<string, number>;
```
vkarpov15, line 2120: I don't like restricting projections to numbers. There are plenty of codebases out there that use true/false instead of 0/1. And MongoDB does also allow using strings in `select()`, although that has [different behavior than using true/false or 0/1](https://thecodebarbarian.com/whats-new-in-mongoose-5-13-sanitizeprojection.html).

Ideally I'd say this should be `[Key in keyof Omit<LeanDocument<DocType>, '__v'>]?: any`, but if you want to be more strict than that, we need to support:

1. Boolean
2. String
3. [Projection operators](https://docs.mongodb.com/manual/reference/operator/projection/)

vkarpov15 (changes requested):
The general idea looks good. I double checked the performance against #10349 and it doesn't look like this causes any major performance issues. Only thing to fix up is removing the restriction that RHS of projection properties should be numbers, that isn't entirely correct.

index.d.ts
EugeneKorshenko: Thank you for your review. I'll follow your version with `any`.

vkarpov15 (approved):
Merging this into 6.2 branch, thanks :+1:

PR #11222, opened 2022-01-13
Sync indexes global option

IslandRhythms:
Why can't I do `await Model.listIndexes().length` off the rip but instead have to save it to a variable? Just curious.

### DIFF
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -879,7 +879,8 @@ describe('mongoose module:', function() {
       const Indexes = await syncFalse.listIndexes();
       assert.equal(Indexes.length, 2);
       await syncFalse.collection.createIndex({ name: 1 });
-      await m.deleteModel('Sync');
+      // await m.deleteModel('Sync');
+      delete m.connection.models['Sync'];
       m.set('syncIndexes', true);
       const syncSchema = new m.Schema({ nickname: { type: String, index: true } });
       const syncTrue = db.model('Sync', syncSchema, 'Sync');
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -866,5 +866,18 @@ describe('mongoose module:', function() {
       const m = new mongoose.Mongoose();
       assert.deepEqual(await m.syncIndexes(), {});
     });
+    it('Allows a syncIndexes global option (gh-11030)', async function() {
+      const m = new mongoose.Mongoose();
+      m.set('syncIndexes', true);
+
+      const db = await m.connect('mongodb://localhost:27017/mongoose_test_11030');
+
+      const schema = new m.Schema({
+        title: String,
+      });
+
+      const Movie = db.model('Movie', schema);
+
+    });
   });
 });
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -872,23 +872,22 @@ describe('mongoose module:', function() {
       m.set('debug', true);
       const db = await m.connect('mongodb://localhost:27017/mongoose_test_11030');
       const schema = new m.Schema({
-        title: String
+        title: {type: String, index: true}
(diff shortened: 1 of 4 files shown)


vkarpov15:
"Why can't I do await Model.listIndexes().length off the rip but instead have to save it to a variable? Just curious."

Order of operations. `Model.listIndexes()` returns a promise, not an array. So `await Model.listIndexes().length` means "get the `.length` property of `Model.listIndexes()`, and `await` on the result", which isn't what you want.

You can do `(await Model.listIndexes()).length`, or `await Model.listIndexes().then(res => res.length)`

test/index.test.js
```diff
@@ -866,5 +866,28 @@ describe('mongoose module:', function() {
...
+      });
+      const syncFalse = db.model('Sync', schema, 'Sync');
+      await syncFalse.init();
+      const Indexes = await syncFalse.listIndexes();
+      console.log(Indexes);
+      assert.equal(Indexes.length, 2);
+      await syncFalse.collection.createIndex({ name: 1 });
+      delete m.connection.models['Sync'];
+      m.set('syncIndexes', true);
+      const syncSchema = new m.Schema({ title: String });
```
vkarpov15, line 885: In order to test that `syncIndexes()` correctly adds new indexes as well as drops old indexes, let's add a property that isn't `name` or `title`, and add an index to that new property in the schema.

test/index.test.js
```diff
@@ -866,5 +866,28 @@ describe('mongoose module:', function() {
...
+      const syncTrue = db.model('Sync', syncSchema, 'Sync');
+      await syncTrue.init();
+      const Index = await syncTrue.listIndexes();
+      console.log(Index);
+      assert.equal(Index.length, 1);
```
vkarpov15, line 890: Assert that this index is an index on `_id`.

vkarpov15 (changes requested):

IslandRhythms:
Mongoerror, ns not found. The test passes on my machine so I don't know why it's failing here.

test/index.test.js
```diff
@@ -866,5 +866,27 @@ describe('mongoose module:', function() {
...
+      const schema = new m.Schema({
+        title: { type: String, index: true }
+      });
+      const syncFalse = db.model('Sync', schema, 'Sync');
+      await syncFalse.collection.dropIndexes();
+      await syncFalse.init();
+      const Indexes = await syncFalse.listIndexes();
+      assert.equal(Indexes.length, 2);
+      await syncFalse.collection.createIndex({ name: 1 });
+      await m.deleteModel('Sync');
```
Uzlopak, line 882: I think this line seems to also force mongoose to drop the collection, which probably does not exist. thats why you get the ns not found error

test/index.test.js
```diff
@@ -866,5 +866,27 @@ describe('mongoose module:', function() {
...
     });
+    it('Allows a syncIndexes global option (gh-11030)', async function() {
+      const m = new mongoose.Mongoose();
+
+      const db = await m.connect('mongodb://localhost:27017/mongoose_test_11030');
+      const schema = new m.Schema({
+        title: { type: String, index: true }
+      });
+      const syncFalse = db.model('Sync', schema, 'Sync');
+      await syncFalse.collection.dropIndexes();
```
vkarpov15, line 877: I think you need to get rid of this line to fix the tests. The problem is that the 'Sync' collection doesn't exist yet when you're calling `dropIndexes()` here.

vkarpov15 (changes requested):

vkarpov15 (approved):
Thanks :+1: will merge into 6.2

PR #11224, opened 2022-01-13
 Redundant null check on object #11223

### LINKED ISSUE #11223: Redundant null check on object
BUG

There's a redundant null check on variable at [this line](https://github.com/Automattic/mongoose/blob/master/lib/document.js#L76), after checking type of variable as 'object'

It should perform the null check first, and if it passes null check then perform type check.

Nodejs: 14.18.2
Mongoose: 6.1.6

dadwalakshay:
**Summary**
Fixed redundant null check on `skipId` variable at [this line](https://github.com/Automattic/mongoose/blob/master/lib/document.js#L76), by performing null check on mentioned variable before performing type check.

Resolves #11223

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -73,7 +73,7 @@ const specialProperties = utils.specialProperties;
  */
 
 function Document(obj, fields, skipId, options) {
-  if (typeof skipId === 'object' && skipId != null) {
+  if (skipId != null && typeof skipId === 'object') {
     options = skipId;
     skipId = options.skipId;
   }


Uzlopak:
First of all, this is not a redundant null check. So the title is already confusing. But this was already said in the linked issue.

And second of all, if this is for some "performance" improvement, than it makes no sense anyway. In most cases, skipId will not be null. The only reason the non null check is there is because we have to check if skipId is an Options-Object and if so switch the paramers for backwards compatibility reasons(?). But because null is also of type Object, we have to filter out false-positives like the null-Object. Your change assumes, that null is the most provided value for skipId, which is just an unproven assumption. It can not even increase the performance, even if this should be the "bottleneck" as we will probably fail more often in the typeof check than in the null-check. So your PR is making the situation worse, as it would decrease the performance.  

So in conclusion, this PR should be rejected.

vkarpov15:
I'll close this PR. The null check isn't redundant, and the performance impact either way is negligible, so not worth worrying about any further.

PR #11234, opened 2022-01-17
Replace util is array

Uzlopak:
Replace the deprecated "util.isArray" with Array.isArray.


Uzlopak:
I guess it needs an merge of master into 6.2 to remove the additional commits

vkarpov15 (approved):
Thanks :+1:

PR #11236, opened 2022-01-17
Replace ms with faster @lukeed/ms

Uzlopak:
Replace ms with faster @lukeed/ms

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -4,7 +4,7 @@
  * Module dependencies.
  */
 
-const ms = require('ms');
+const ms = require('@lukeed/ms');
 const mpath = require('mpath');
 const sliced = require('sliced');
 const Decimal = require('./types/decimal128');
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
I took a closer look at the @lukeed/ms repo and we'll merge this into 6.2.

My only concern here is that this change will have negligible impact: we only use `ms()` when creating a new schema, which is typically only done when the app is starting up.

sean-daley:
Was there a particular motivation for switching from a library with 100 million weekly downloads to a fork of that library with 5000 or so? Unless you really need the performance improvements provided by this (which it doesn’t seem like you use it very much?) that seems like a curious decision to make.

vkarpov15:
@sean-daley you make a fair point. We merged because, although we don't use `ms()` much, more perf is nice, plus lukeed looks like a reputable dev and we wanted to support this work. Does using this module cause some issues for you?

Uzlopak:
@lukeed

Fyi

sean-daley:
> @sean-daley you make a fair point. We merged because, although we don't use `ms()` much, more perf is nice, plus lukeed looks like a reputable dev and we wanted to support this work. Does using this module cause some issues for you?

No, it doesn’t really cause issues for us I was just more curious at the reasoning for inclusion. (ms) seems to be a pretty popular library and if it’s not really solving a performance issue you’re having, you’re now imposing another dependency on someone who uses both ms and mongoose. Given the similarities between the two it won’t be a huge addition but it’s still “one more library” from an auditing and/or vetting perspective.

Regardless it’s not a major issue for us, I was just more curious than anything else.

vkarpov15:
@sean-daley you're right about imposing another dependency being a bit of a pain. Nobody wants more dependabot alerts. We'll revert this change for now since we don't need the extra perf.

PR #11239, opened 2022-01-18
move utils.random to test folder

Uzlopak:
The "private" function is only used for tests, therefore it should be not in the productive code.


AbdelrahmanHafez (approved):
Thanks 👍

vkarpov15 (approved):
Thanks :+1:

PR #11242, opened 2022-01-18
check for all flags in a regex deepequal

Uzlopak:
This PR modifies the deepEqual check in mongoose. It checks a regex for all 6 flags of a Regex.

As it is more strict, it could be breaking(?)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -84,7 +84,8 @@ exports.deepEqual = function deepEqual(a, b) {
         a.global === b.global &&
         a.dotAll === b.dotAll &&
         a.unicode === b.unicode &&
-        a.sticky === b.sticky;
+        a.sticky === b.sticky &&
+        a.hasIndices === b.hasIndices;
   }
 
   if (a == null || b == null) {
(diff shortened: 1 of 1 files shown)


Uzlopak:
The failing of the test seems to be not related to the change.

lib/utils.js
```diff
@@ -81,7 +81,10 @@ exports.deepEqual = function deepEqual(a, b) {
     return a.source === b.source &&
         a.ignoreCase === b.ignoreCase &&
         a.multiline === b.multiline &&
-        a.global === b.global;
+        a.global === b.global &&
+        a.dotAll === b.dotAll &&
+        a.unicode === b.unicode &&
+        a.sticky === b.sticky;
```
vkarpov15, line 87: It looks like we also need to check the `hasIndices` flag: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#advanced_searching_with_flags

vkarpov15 (changes requested):
Overall I'm ok to merge this, just looks like there might still be one flag missing. I think this is a fairly clear bug.

lib/utils.js
Uzlopak: hasIndices is new in node16, but you are right.

Uzlopak: Done :)

AbdelrahmanHafez (approved):
Good catch 👏
Appreciate all the good work 👍

vkarpov15 (approved):
Thanks :+1:

PR #11243, opened 2022-01-18
move "allowSyntheticDefaultImports" into "compilerOptions" in tsconfi…

Uzlopak:
…g.json

allowSyntheticDefaultImports is an option of compilerOptions and not an attribute of the root.


Uzlopak:
@vkarpov15 

Could you please merge this? Then I could kill my forked repo and refork it to have all the branches from upstream

vkarpov15 (approved):
Thanks :+1:

PR #11246, opened 2022-01-19
Update mongodb driver to 4.3.1

lorand-horvath:


vkarpov15:
Looks like it'll be a bit harder for us to upgrade to 4.3.1: it looks like `findOne()` now adds `& { _id: any }`, so we'll have to figure out a workaround.

lorand-horvath:
> Looks like it'll be a bit harder for us to upgrade to 4.3.1: it looks like `findOne()` now adds `& { _id: any }`, so we'll have to figure out a workaround.

Is there any documentation on this?

mark-langer:
It would be nice to have the driver updated because MongoDB's Node driver typing currently produces the error `TS2322: Type 'string' is not assignable to type 'ObjectId'.` when using 
```
connection.collection('myCollection').find<MyRecord>({
    _id: {
        $in: ['id1', 'id2'] // <-- error here
    }
});
```
if `MyRecord` is typed with `_id: string`.

image

I just upgraded to mongodb 4.3.1 manually; that removes the error. ✅

vkarpov15:
@lorand-horvath not to my knowledge.

We did some digging and the issue was introduced in https://github.com/mongodb/node-mongodb-native/pull/3102 . Long story short, the MongoDB Node driver's `WithId<>` type is _slightly_ incompatible with Mongoose's `Require_id<>` type. In theory, these two types should be exactly equivalent: return the generic type param, but with a required `_id` property. In practice, TypeScript thinks they're different.

We solved this issue, for now, by pulling in the `ModifyResult` type in e0ed2e8 . Trying to replace Mongoose's `Require_id<>` type caused some issues with our TypeScript maps tests, among others, so this seems like the simplest and cleanest way to work around this issue.

Issue #10662, opened 2021-09-02
$expr is getting ignored after upgrade to v6


pnutmath:
**Do you want to request a *feature* or report a *bug*?**
This seems a bug

**What is the current behavior?**
$expr is getting ignored after upgrade to v6

**If the current behavior is a bug, please provide the steps to reproduce.**

**Schema**
```javascript

const SampleSchema = new mongoose.Schema(
  {
    date: { type: Date, required: true },
  },
  {
    timestamps: { createdAt: 'created_at', updatedAt: 'updated_at' },
  }
);

SampleSchema.index({ date: -1 });
module.exports = moonracer.model('Sample', SampleSchema);
```

While executing this in code
```javascript
 const SampleModel =  require('./sample)

 SampleModel.find( { $expr: { $eq: [{ $year: '$date' }, 2021] } }, {date: 1}).sort({ date: -1 });
```
**mongoose generating query,log is**

`Mongoose: samples.find({  }, { sort: { date: -1 },  projection: {  date: 1,  _id: -1 }})
`

**What is the expected behavior?**

**mongoose should call following query**

`Mongoose: samples.find({ $expr: { $eq: [{ $year: '$date' }, 2021] } }, { sort: { date: -1 },  projection: {  date: 1,  _id: -1 }})
`

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Node: v12.22.1
MongoDB: 4.2.8
Mongoose: 6.0.4

Issue is starting from v6.x.x

IslandRhythms:
> module.exports = moonrace.model('Sample', SampleSchema);

That's a typo right?

IslandRhythms:
The query does go through in older versions, its in 6.x that this issue pops up

```javascript
const mongoose = require('mongoose');
mongoose.set('debug', true)
const testSchema = new mongoose.Schema({ date: {type: Date, required: true }}, {timestamps: {createdAt: 'created_at', updatedAt: 'updated_at'}})

testSchema.index({ date: -1});

const Test = mongoose.model('Test', testSchema);

async function test() {
    await mongoose.connect('mongodb://localhost:27017/test', {
        useNewUrlParser: true,
        useUnifiedTopology: true,
      });
      await mongoose.connection.dropDatabase();

      await Test.find( { $expr: { $eq: [{ $year: '$date' }, 2021] } }, {date: 1}).sort({ date: -1 });
}

test();
```

PR #11254, opened 2022-01-22
Casting for `$expr` in queries

vkarpov15:
Mostly for comparisons, like `$eq`, etc. like in #10662 .

Missing support for a few things, like `$dateFromString` and similar operators, which we can add. But primarily wanted to support casting comparisons of paths and literals, which is what this PR does.

### DIFF
--- a/lib/helpers/query/cast$expr.js
+++ b/lib/helpers/query/cast$expr.js
@@ -44,6 +44,16 @@ const arithmeticOperatorNumber = new Set([
   '$degreesToRadians',
   '$radiansToDegrees'
 ]);
+const dateOperators = new Set([
+  '$year',
+  '$month',
+  '$week',
+  '$dayOfMonth',
+  '$dayOfYear',
+  '$hour',
+  '$minute',
+  '$second'
+]);
 
 module.exports = function cast$expr(val, schema, strictQuery) {
   if (typeof val !== 'object' || val == null) {
@@ -60,9 +70,13 @@ function _castExpression(val, schema, strictQuery) {
   }
 
   if (val.$cond != null) {
-    val.$cond.if = _castExpression(val.$cond.if, schema, strictQuery);
-    val.$cond.then = _castExpression(val.$cond.then, schema, strictQuery);
-    val.$cond.else = _castExpression(val.$cond.else, schema, strictQuery);
+    if (Array.isArray(val.$cond)) {
+      val.$cond = val.$cond.map(expr => _castExpression(expr, schema, strictQuery));
+    } else {
+      val.$cond.if = _castExpression(val.$cond.if, schema, strictQuery);
+      val.$cond.then = _castExpression(val.$cond.then, schema, strictQuery);
+      val.$cond.else = _castExpression(val.$cond.else, schema, strictQuery);
+    }
   } else if (val.$ifNull != null) {
     val.$ifNull.map(v => _castExpression(v, schema, strictQuery));
   } else if (val.$switch != null) {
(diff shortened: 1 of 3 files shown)


test/helpers/query.cast$expr.test.js
```diff
@@ -0,0 +1,90 @@
...
+const cast$expr = require('../../lib/helpers/query/cast$expr');
+
+describe('castexpr', function() {
+  it('casts comparisons', function() {
+    const testSchema = new Schema({ date: Date, spent: Number, budget: Number });
+
+    let res = cast$expr({ $eq: ['$date', '2021-06-01'] }, testSchema);
+    assert.deepEqual(res, { $eq: ['$date', new Date('2021-06-01')] });
+
+    res = cast$expr({ $eq: [{ $year: '$date' }, 2021] }, testSchema);
```
AbdelrahmanHafez, line 14:
```suggestion
    res = cast$expr({ $eq: [{ $year: '$date' }, '2021'] }, testSchema);
```

test/helpers/query.cast$expr.test.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
+
+const { Schema } = require('../common').mongoose;
+const assert = require('assert');
+const cast$expr = require('../../lib/helpers/query/cast$expr');
+
+describe('castexpr', function() {
+  it('casts comparisons', function() {
```
AbdelrahmanHafez, line 8: Would be nice if we could support `$cond: [myCondition, trueCase, falseCase]` as well.

Would happily work on that in a next PR or on this one. @vkarpov15

test/helpers/query.cast$expr.test.js
```diff
@@ -0,0 +1,90 @@
...
+    assert.deepStrictEqual(res, { $eq: [{ $year: '$date' }, { $literal: 2021 }] });
+
+    res = cast$expr({ $eq: [{ $year: '$date' }, { $literal: '2021' }] }, testSchema);
+    assert.deepStrictEqual(res, { $eq: [{ $year: '$date' }, { $literal: 2021 }] });
+
+    res = cast$expr({ $gt: ['$spent', '$budget'] }, testSchema);
+    assert.deepStrictEqual(res, { $gt: ['$spent', '$budget'] });
+  });
+
+  it('casts conditions', function() {
```
AbdelrahmanHafez, line 30: Would be nice if we could support `$cond: [myCondition, trueCase, falseCase]` as well.

Would happily work on that in a next PR or on this one. @vkarpov15

AbdelrahmanHafez:
Great work, I think there is still more space for testing some of the features in the PR, but we can add that in a future PR along with any possible refactorings.

Thanks 👍

lib/helpers/query/cast$expr.js
```diff
@@ -0,0 +1,229 @@
...
+        val[1] = { $literal: schematype.cast(val[1].$literal) };
+      } else {
+        val[1] = schematype.cast(val[1]);
+      }
+    } else if (caster != null) {
+      if (is$literal) {
+        try {
+          val[1] = { $literal: caster(val[1].$literal) };
+        } catch (err) {
+          throw new CastError(caster.name.slice(4 /* 'cast'.length */), val[1], path + '.$literal');
```
AbdelrahmanHafez, line 194: Perhaps using String#replace would show intention more clearly.
```suggestion
          throw new CastError(caster.name.replace('cast', ''), val[1], path + '.$literal');
```

lib/helpers/query/cast$expr.js
```diff
@@ -0,0 +1,229 @@
...
+        }
+      } else {
+        try {
+          val[1] = caster(val[1]);
+        } catch (err) {
+          throw new CastError(caster.name.slice(4 /* 'cast'.length */), val[1], path);
```
AbdelrahmanHafez, line 200:
```suggestion
          throw new CastError(caster.name.replace('cast', ''), val[1], path);
```

lib/helpers/query/cast$expr.js
```diff
@@ -0,0 +1,229 @@
...
+        path = lhs.$month.slice(1) + '.$month';
+        caster = castNumber;
+      } else if (isPath(lhs.$week)) {
+        path = lhs.$week.slice(1) + '.$week';
+        caster = castNumber;
+      } else if (isPath(lhs.$dayOfMonth)) {
+        path = lhs.$dayOfMonth.slice(1) + '.$dayOfMonth';
+        caster = castNumber;
+      } else if (isPath(lhs.$dayOfYear)) {
+        path = lhs.$dayOfMonth.slice(1) + '.$dayOfYear';
```
AbdelrahmanHafez, line 168:
```suggestion
        path = lhs.$dayOfYear.slice(1) + '.$dayOfYear';
```

lib/helpers/query/cast$expr.js
```diff
@@ -0,0 +1,229 @@
...
+  const lhs = val[0];
+
+  if (isLiteral(val[1])) {
+    let path = null;
+    let schematype = null;
+    let caster = null;
+    if (isPath(lhs)) {
+      path = lhs.slice(1);
+      schematype = schema.path(path);
+    } else if (typeof lhs === 'object' && lhs != null) {
```
AbdelrahmanHafez, line 154: There is [room for refactoring](https://github.com/Automattic/mongoose/compare/gh-10663...AbdelrahmanHafez:gh-10663) here that reduces the need for duplication and makes the code less error-prone.

vkarpov15: You're right, this code needed to be DRY-ed up. Fixed in 29b323c

lib/helpers/query/cast$expr.js
vkarpov15: Fixed in 29b323c by removing this repetitive `if`

lib/helpers/query/cast$expr.js
vkarpov15: Fixed in 29b323c

test/helpers/query.cast$expr.test.js
vkarpov15: Fixed in 29b323c

PR #11255, opened 2022-01-23
add mongodb-memory-server to test easier locally. 

Uzlopak:
To run unit tests locally i added the option to start a mongodb-memory-server.

`npm run mongo` and voilá

if you want to run e.g. mongo 4.4.12 you would do 

`npm run mongo -- 4.4.12`

### DIFF
--- a/tools/repl.js
+++ b/tools/repl.js
@@ -10,6 +10,9 @@ async function run () {
 
   // Create new instance
   const replSet = new ReplSet({
+    binary: {
+      version: process.argv[3]
+    },
     instanceOpts: [
       // Set the expiry job in MongoDB to run every second
       { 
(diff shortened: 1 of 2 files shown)


tools/repl.js
```diff
@@ -6,31 +6,29 @@ run().catch(error => {
...
+      // Set the expiry job in MongoDB to run every second
+      { 
+        port: 27017,
+        args: ["--setParameter", "ttlMonitorSleepSecs=1"] },
+    ],
+    dbName: 'mongoose_test',
+    replSet: {
+      name: "rs0",
+      count: 2,
+      storageEngine: "wiredTiger",
```
vkarpov15, line 26: Minor thing to note for the future: I typically use `inMemory` storage engine for tests locally, because it is much faster. I only use WT for testing transactions, etc.

But it would be great to use this script to set up MongoDB for CI so we can run a replica set and test transactions in CI.

vkarpov15 (approved):
Nice work :+1:

PR #11262, opened 2022-01-24
fix casting benchmark

Uzlopak:
fix the casting benchmark


AbdelrahmanHafez (approved):
Good catch, thanks 👍

vkarpov15 (approved):
Thanks :+1: we haven't used these benchmarks in a while though.

Uzlopak:
I use the benchmarks to find the potential hot paths and performance bottlenecks.

PR #11263, opened 2022-01-24
perf: improve performance of parentPaths

Uzlopak:
My Benchmark for optimizing parentPaths.

var0 = old
var1 = new

```
first
var0 x 6,011,762 ops/sec ±0.76% (90 runs sampled)
var1 x 122,339,355 ops/sec ±0.96% (95 runs sampled)
first.second
var0 x 5,210,949 ops/sec ±0.42% (91 runs sampled)
var1 x 7,159,684 ops/sec ±0.96% (90 runs sampled)
first.second.third
var0 x 4,773,409 ops/sec ±0.29% (93 runs sampled)
var1 x 5,326,110 ops/sec ±0.78% (92 runs sampled)
first.second.third.fourth.fifth
var0 x 4,134,150 ops/sec ±0.60% (91 runs sampled)
var1 x 3,836,214 ops/sec ±0.78% (92 runs sampled)
a.b.c.d.e.f
var0 x 3,665,917 ops/sec ±0.35% (94 runs sampled)
var1 x 3,452,551 ops/sec ±1.22% (90 runs sampled)
```

```
'use strict';

const Bench = require('benchmark');
const { expect } = require('chai');

function parentPathsVar0(path) {
  const pieces = path.split('.');
  let cur = '';
  const ret = [];
  for (let i = 0; i < pieces.length; ++i) {
    cur += (cur.length > 0 ? '.' : '') + pieces[i];
    ret.push(cur);
  }

  return ret;
};
const dotRE = /\./g
function parentPathsVar1(path) {
  if (path.indexOf('.') === -1) {
    return [path];
  }
  const pieces = path.split(dotRE);
  const len = pieces.length;
  const ret = new Array(len);
  let cur = '';
  for (let i = 0; i < len; ++i) {
    cur += (cur.length !== 0) ?  '.' + pieces[i] : pieces[i];
    ret[i] = cur;
  }

  return ret;
};

const validate = (value, expected) => {
  expect(parentPathsVar0(value), "var0").to.be.eql(expected);
  expect(parentPathsVar1(value), "var1").to.be.eql(expected);
}

validate('first', ['first']);
validate('first.second', ['first', 'first.second']);
validate('', ['']);

const bench = (value) => {
  console.log(value)
  new Bench.Suite()
    .add('var0', function () {
      parentPathsVar0(value);
    })
    .add('var1', function () {
      parentPathsVar1(value);
    })
    .on('cycle', function (e) {
      const s = String(e.target);
      console.log(s);
    })
    .run();
}


bench("first");
bench("first.second");
bench("first.second.third");
bench("first.second.third.fourth.fifth");
bench("a.b.c.d.e.f");
```

### DIFF
--- a/lib/helpers/path/parentPaths.js
+++ b/lib/helpers/path/parentPaths.js
@@ -1,16 +1,18 @@
 'use strict';
 
+const dotRE = /\./g;
 module.exports = function parentPaths(path) {
   if (path.indexOf('.') === -1) {
     return [path];
   }
-  let pos = path.indexOf('.');
-  const ret = [];
-  let i = 0;
-  while (pos !== -1) {
-    ret[i++] = (path.substring(0, pos));
-    pos = path.indexOf('.', pos + 1);
+  const pieces = path.split(dotRE);
+  const len = pieces.length;
+  const ret = new Array(len);
+  let cur = '';
+  for (let i = 0; i < len; ++i) {
+    cur += (cur.length !== 0) ? '.' + pieces[i] : pieces[i];
+    ret[i] = cur;
   }
-  ret[i] = path.substring(0);
+
   return ret;
 };
--- a/lib/helpers/path/parentPaths.js
+++ b/lib/helpers/path/parentPaths.js
@@ -1,13 +1,16 @@
 'use strict';
 
 module.exports = function parentPaths(path) {
-  const pieces = path.split('.');
-  let cur = '';
-  const ret = [];
-  for (let i = 0; i < pieces.length; ++i) {
-    cur += (cur.length > 0 ? '.' : '') + pieces[i];
-    ret.push(cur);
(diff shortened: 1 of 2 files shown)


Uzlopak:
I have to reevaluate this.

Uzlopak:
I think now it is more stable in performance. You could only improve perf by caching...

vkarpov15:
Based on your original comment, it looks like with these changes the perf degrades relative to the current code if you increase the number of `.` in the string. Is that still the case after your more recent changes?

Uzlopak:
The original Code was giving false Benchmarks. More optimistic than it was actually. So I decided to make it more conservative code change and don't try to outsmart V8.

Uzlopak:
Will be part of another PR

PR #11265, opened 2022-01-24
remove obsolete code after upgrading to bson4

Uzlopak:
There was a remark to remove a code block if we upgraded to bson4....

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -36,14 +36,6 @@ module.exports = function cast(schema, obj, options, context) {
     return obj;
   }
 
-  // bson 1.x has the unfortunate tendency to remove filters that have a top-level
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez (approved):
Good catch, thanks 👍

vkarpov15 (approved):
I took a closer look and confirmed the original issue no longer happens in Mongoose 6. Below script errors out with `BSONTypeError: Unrecognized or invalid _bsontype: ObjectID22`

```javascript
'use strict';
  
mongoose.set('debug', true);

const { Schema } = mongoose;

async function run() {

  const Test = mongoose.model('Test', Schema({ name: String }));

  await Test.create({ name: 'foo' });
  const res = await Test.findOne({ _bsontype: 'ObjectID22' });
  console.log(res);
}
```

PR #11266, opened 2022-01-24
enhancement(connection&base): add support for `continueOnError`

### LINKED ISSUE #11039: [SUGGESTION] Make `mongoose.syncIndexes()` and `Connection#syncIndexes()` always successfully fulfilled (error independent)
_Originally posted by @Kamikadze4GAME in https://github.com/Automattic/mongoose/pull/10997#discussion_r755016633_

Maybe is it better to return object, where key is `modelName` and value is result (successful) from `this.model(model).syncIndexes()` or if something went wrong instance of the Error (or MongooseError)?

So the result object will look like (it's just an example):

```js
const syncIndexesResult = await connection.syncIndexes();
// `syncIndexesResult`
{
  model1: ['index-1-1', 'index-1-2'], // this is example of the successful result from `Model.syncIndexes`,
  model2: MongooseError, 
  model3: [...],
}
```

As a result, you could control future actions (ex: show in console list of failed models or manually rerun syncIndexes on failed models).

What do you think?

@vkarpov15 @AbdelrahmanHafez @IslandRhythms 

Ref: #10893, #10997

AbdelrahmanHafez:
fixes #11039


index.d.ts
```diff
@@ -3303,6 +3304,11 @@ declare module 'mongoose' {
       type?: string): this;
   }
 
+  export interface SyncIndexesOptions {
```
vkarpov15, line 3307: `SyncIndexesOptions` needs to at minimum support the same options as `createIndexes()`. Maybe try `type SyncIndexesOptions = mongodb.CreateIndexesOptions & { continueOnError?: boolean }`?

lib/connection.js
```diff
@@ -1380,11 +1381,37 @@ Connection.prototype.setClient = function setClient(client) {
...
+
+  for (const model of Object.values(this.models)) {
+    try {
+      result[model.modelName] = await model.syncIndexes();
+    } catch (err) {
+      if (!continueOnError) {
+        errorsMap[model.modelName] = err;
+        break;
+      } else {
+        result[model.modelName] = err;
```
vkarpov15, line 1404: Nice :+1:

lib/connection.js
```diff
@@ -1380,11 +1381,37 @@ Connection.prototype.setClient = function setClient(client) {
...
+      }
+    }
   }
+
+  if (!continueOnError && Object.keys(errorsMap).length) {
+    const message = Object.values(errorsMap).map(err => err.message).join(' ');
```
vkarpov15, line 1410: What do you think about making this formatting consistent with CastErrors, so we do `${key}: ${err.message}` and join with `,`?

vkarpov15 (changes requested):
Couple of minor comments, but :+1: overall

vkarpov15 (approved):
Nice :+1: merging this into 6.2

PR #11267, opened 2022-01-24
chore: remove chalk as devDependency

Uzlopak:
We use chalk only in one place: To write in red, that we dont run the sharding tests. We just need to use the correct escape character, which can be easily googled, to write in red color. 

Reduces the potential risk of a supply chain attack.

### DIFF
--- a/test/shard.test.js
+++ b/test/shard.test.js
@@ -9,10 +9,11 @@ const mongoose = start.mongoose;
 const Schema = mongoose.Schema;
 
 const uri = process.env.MONGOOSE_SHARD_TEST_URI;
+const redColorEscapeCharacter = '\x1b[31m';
 
 if (!uri) {
   console.log([
-    '\x1b[31m',
+    redColorEscapeCharacter,
     'You\'re not testing shards!',
     'Please set the MONGOOSE_SHARD_TEST_URI env variable.',
     'e.g: `mongodb://localhost:27017/database',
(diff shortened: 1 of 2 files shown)


test/shard.test.js
```diff
@@ -12,14 +11,13 @@ const Schema = mongoose.Schema;
...
-  console.log(
-    chalk.red(
-      '\n', 'You\'re not testing shards!',
-      '\n', 'Please set the MONGOOSE_SHARD_TEST_URI env variable.', '\n',
-      'e.g: `mongodb://localhost:27017/database', '\n',
-      'Sharding must already be enabled on your database'
-    )
-  );
+  console.log([
+    '\x1b[31m',
```
AbdelrahmanHafez, line 15: Nitpick: I'd add `const redColorEscapeCharacter = '\x1b[31m';` and use it in the console.log to reveal intent.

AbdelrahmanHafez (changes requested):

AbdelrahmanHafez:
Other than the minor comment, LGTM. Thanks 👍

test/shard.test.js
Uzlopak: done

AbdelrahmanHafez (approved):

vkarpov15 (approved):
Good work on this. In general, if we can replace a dependency with a one liner, we should.

Issue #2070, opened 2014-05-10
ObjectId comparison broken since 3.8.9


lcalvy:
Hi,

Since version 3.8.9, operator equality comparison on ObjectId field doesn't match. 

Using lodash to find _id in array of document also failed.

exports.findAchievementModuleByChapterId = function(achievement,chapterId){
        return _.find(achievement.modules, { 'currentChapter_id' : chapterId });
    };

ERR1399756867174 Error: ERR0001 : tried to find module for user 52fceec51b01cb0200aeba38 on slide 532c6a3336fe9e9abbdf41db but module can't be found
    at Promise.<anonymous> (/Users/lcalvy/Documents/Boulot/Sources/coorpacademy/core/lib/achievementHelper.js:39:37)
    at Promise.<anonymous> (/Users/lcalvy/Documents/Boulot/Sources/coorpacademy/node_modules/mongoose/node_modules/mpromise/lib/promise.js:177:8)
    at Promise.EventEmitter.emit (events.js:95:17)
    at Promise.emit (/Users/lcalvy/Documents/Boulot/Sources/coorpacademy/node_modules/mongoose/node_modules/mpromise/lib/promise.js:84:38)
    at Promise.fulfill (/Users/lcalvy/Documents/Boulot/Sources/coorpacademy/node_modules/mongoose/node_modules/mpromise/lib/promise.js:97:20)
    at /Users/lcalvy/Documents/Boulot/Sources/coorpacademy/node_modules/mongoose/lib/query.js:1394:13
    at model.Document.init (/Users/lcalvy/Documents/Boulot/Sources/coorpacademy/node_modules/mongoose/lib/document.js:250:11)
    at completeOne (/Users/lcalvy/Documents/Boulot/Sources/coorpacademy/node_modules/mongoose/lib/query.js:1392:10)
    at Object.cb (/Users/lcalvy/Documents/Boulot/Sources/coorpacademy/node_modules/mongoose/lib/query.js:1151:11)
    at Object.<anonymous> (/Users/lcalvy/Documents/Boulot/Sources/coorpacademy/node_modules/mongoose/node_modules/mquery/lib/utils.js:137:16)

PR #11268, opened 2022-01-24
chore: remove lodash.isEqual and lodash.isEqualWith

Uzlopak:
Well I hope I replaced everything with the correct equivalents...

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,8 +46,6 @@
     "eslint": "8.5.0",
     "eslint-plugin-mocha-no-only": "1.1.0",
     "highlight.js": "9.18.3",
-    "lodash.isequal": "4.5.0",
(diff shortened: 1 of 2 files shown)


vkarpov15 (commented):
I have some concerns here. I think the intent of this test is specifically to make sure Lodash's `isEqual()` works with ObjectIds, take a look at #2070. And Lodash's `isEqual()` isn't a trivial function.

Admittedly Mongoose isn't necessarily responsible for making sure ObjectIds behave with Lodash, but we like to watch out for these sort of regressions.

Uzlopak:
Makes sense. Maybe modifying the unit test name would be appropriate?

vkarpov15:
Sure, let's go with just modifying the test name.

PR #11269, opened 2022-01-24
chore: remove uuid-parse, upgrade uuid

Uzlopak:
uuid reports some Math.random sec-vuln. So upgraded it and used the builtin parse from 8.3 uuid to replace the not maintained uuid-parse package. 

Reduce Supply Chain Attack Risks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,8 +60,7 @@
     "semver": "5.5.0",
     "serve-handler": "6.1.3",
     "typescript": "4.5.3",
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

vkarpov15 (approved):
I pinned to an exact version of uuid, but otherwise LGTM

PR #11271, opened 2022-01-25
chore: remove semver from dev dependencies

Uzlopak:
We did not use semver. I took the chance to make the onlyTestAtOrAbove-method more readable (well in my opinion atleast ;).

### DIFF
--- a/test/aggregate.test.js
+++ b/test/aggregate.test.js
@@ -53,14 +53,21 @@ function setupData(db, callback) {
  *
  * @param {String} semver, `3.4`, specify minimum compatible mongod version
  * @param {Object} ctx, `this`, so that mocha tests can be skipped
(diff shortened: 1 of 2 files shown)


test/aggregate.test.js
```diff
@@ -53,14 +53,21 @@ function setupData(db, callback) {
  *
  * @param {String} semver, `3.4`, specify minimum compatible mongod version
  * @param {Object} ctx, `this`, so that mocha tests can be skipped
- * @return {Void}
+ * @return {Promise.<void>}
```
AbdelrahmanHafez, line 56:
```suggestion
 * @return {Promise<void>}
```

vkarpov15 (approved):
Thanks :+1:

PR #11272, opened 2022-01-25
chore: replace object-sizeof with native v8.serialize(...).length

Uzlopak:
object-sizeof is an unnecessary dependency, which does not even return the correct memory footprint of a variable in v8/node. To determine the correct size of a variable, we can use the serialize method of v8 and use the length property of the Buffer to determine the actual size of the variable, when stored in v8.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
     "mocha": "9.1.3",
     "moment": "2.x",
     "mongodb-memory-server": "^8.2.0",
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
Nice, thanks 👍

vkarpov15 (approved):
Thanks :+1:

PR #11273, opened 2022-01-25
fix: port for documentation testing in CONTRIBUTING.md

Uzlopak:
n.T.

### DIFF
--- a/static.js
+++ b/static.js
@@ -4,6 +4,6 @@ const handler = require('serve-handler');
 
 require('http').createServer(function(req, res) {
   handler(req, res, { public: '.' }).catch(err => res.statusCode(500).send(err.message));
-}).listen(8089);
+}).listen(8088);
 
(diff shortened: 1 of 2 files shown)


static.js
```diff
@@ -4,6 +4,6 @@ const handler = require('serve-handler');
 
 require('http').createServer(function(req, res) {
   handler(req, res, { public: '.' }).catch(err => res.statusCode(500).send(err.message));
-}).listen(8089);
+}).listen(8088);
```
AbdelrahmanHafez, line 7: @vkarpov15 Changed this 7 months ago to 8089, this probably affects how https://mongoosejs.com is served, potentially a breaking change.

Uzlopak: reverted it, also made it possible to set the port like

`$ PORT=8080 node static`

vkarpov15 (approved):
Thanks :+1:

PR #11278, opened 2022-01-26
chore: replace substr with substring

Uzlopak:
Not a big change, but substr is annoying as it always shows me in vscode, that .substr is "deprecated".


AbdelrahmanHafez (approved):
LGTM, thanks 👍

vkarpov15 (approved):
In general we prefer [`slice()` over `substring()` and `substr()`](https://masteringjs.io/tutorials/fundamentals/substring), but not a big deal either way.

PR #11285, opened 2022-01-27
add docs regarding addToSet adding $each to query

dor-benatia:
Signed-off-by: Dor Ben-Atia <dorb@activefence.com>

Added a caveat about using `$addToSet` - this will add `$each` operator on the values passed.

### DIFF
--- a/docs/schematypes.md
+++ b/docs/schematypes.md
@@ -510,6 +510,8 @@ const Empty3 = new Schema({ any: [Schema.Types.Mixed] });
 const Empty4 = new Schema({ any: [{}] });
 ```
 
+Note: when updating arrays with $addToSet, an $each will be applied on the arguments passed to addToSet.
+
 <h4 id="maps">Maps</h4>
 
 _New in Mongoose 5.1.0_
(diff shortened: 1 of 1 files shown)


dor-benatia:
exposing `$addToSet` caveat in underlying adding `$each` as an update operator.

AbdelrahmanHafez:
Welcome @dor-benatia 👋

I'm not sure I understand the scenario you're adding a note of, can you provide a code example?

vkarpov15 (commented):
I agree with @AbdelrahmanHafez 's comments, would be great to have a short 1-2 line code example demonstrating what this comment is saying.

AbdelrahmanHafez:
I think that's a bug, mongoose assumes that a path with the type `[Schema.Types.Mixed]` has an array of flat items, while it is a valid use-case that a user wants to store an array of arrays.

Mongoose should not have any assumptions for `Schema.Types.Mixed`.

AbdelrahmanHafez:
This should be fixed by https://github.com/Automattic/mongoose/commit/875ada1529081d67098f24ccc117bebf971a83ea

vkarpov15:
I agree with Hafez, should be fixed by https://github.com/Automattic/mongoose/commit/875ada1529081d67098f24ccc117bebf971a83ea

PR #11292, opened 2022-01-29
chore: rename unit test to clarify its purpose

Uzlopak:
follow up to #11268 

It was easier to rebranch then to revert the changes :P


vkarpov15 (approved):
Thanks :+1:

PR #11293, opened 2022-01-29
fix: set default for dotted path projection

noseworthy:
**Summary**

When a schema included a subdocument that had a default value of empty
object and that subdocument itself included some defaults, the
subdocument was always null if a query was run with a dotted path
projection to the subdocument's default values.

Check not only that the curent path being evaluated to have defaults
applied, but also check if the current path is part of the included
children in the projection.

**Examples**

A test has been provided that fails before the change, but assume the following Schemas:

```javascript

const childSchema = new mongoose.Schema({
  name: {
    default: () => 'child',
  },
});

  name: mongoose.Schema.Types.String,
  child: {
    type: childSchema,
    default: () => ({}),
  },
});

```

If there was a document in the database that didn't include the `child` field and a query was made that included a projection that had a dotted path to the child's default field, the returned document would have a value of `null` for the `child` field.:

```javascript
/* Assume the database contains:
{
  _id: ObjectId(...),
  name: 'Parent'
}
*/

const  parentDoc = await ParentModel.findOne({name: 'Parent'}, {'child.name': 1});

parentDoc.child === null;
```

This doesn't happen when using nested paths.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -479,7 +479,7 @@ function $__applyDefaults(doc, fields, exclude, hasIncludedChildren, isBeforeSet
           break;
         }
       } else if (exclude === false && fields && !included) {
-        if (curPath in fields) {
+        if (curPath in fields || type.$isSingleNested && hasIncludedChildren[curPath]) {
           included = true;
         } else if (!hasIncludedChildren[curPath]) {
           break;
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -10998,4 +10998,66 @@ describe('document', function() {
     assert.equal(typeof doc.somethingElse, 'string');
     delete String.type;
   });
+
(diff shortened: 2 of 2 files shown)


noseworthy:
Hey @vkarpov15, thanks for kicking off the actions workflow on a previous version of this PR. Sorry that I missed a couple of test cases when I was verifying this locally and the tests failed last time. I've since fixed them and re-run all the tests locally again. I think this version of the PR should pass all the tests. Would it be possible to get another workflow approval? This bug is preventing us from using `Typegoose` and `mongoose-graphql-compose` effectively together. Thanks for taking the time to look at this, and thanks for your work on `mongoose`!

vkarpov15 (approved):
Thanks :+1:

vkarpov15 (approved):
Just need to make a quick fix for the tests in prod, need to skip these tests on MongoDB server 4.0

PR #11296, opened 2022-01-29
improve-dx: make it easier to start development for the project

Uzlopak:
This PR improves the Developer Experience:
 - Added a npm script to automatically symlink mongoose so that typescript tests can run without a hassle. As it uses npm link, it should work reliably across all platforms.
- Added information regarding npm run sym-link-mongoose and npm run mongo.
- modified the github actions script
- fixed repl.js to do actually work with different versions.


vkarpov15 (approved):
Thanks :+1:

PR #11298, opened 2022-01-30
perf: Improve performance especially of validate and clone

Uzlopak:
This PR improves the performance of especially validate and clone. 

before: 

```
MONGOOSE_DEV=1 node validate
invalid x 18,325 ops/sec ±1.40% (89 runs sampled)
valid x 46,240 ops/sec ±1.30% (91 runs sampled)

MONGOOSE_DEV=1 node clone
Simple x 124,034 ops/sec ±1.41% (91 runs sampled)
AllSchema x 17,409 ops/sec ±0.52% (94 runs sampled)

MONGOOSE_DEV=1 node create
CheckItem x 113,922 ops/sec ±0.34% (94 runs sampled)
Board x 112 ops/sec ±1.26% (74 runs sampled)

node populate
1436 completed queries on mongoose version 6.1.8
```

after:
```
MONGOOSE_DEV=1 node validate
invalid x 24,576 ops/sec ±0.89% (88 runs sampled)
valid x 119,331 ops/sec ±0.12% (95 runs sampled)

MONGOOSE_DEV=1 node clone
Simple x 148,691 ops/sec ±1.36% (87 runs sampled)
AllSchema x 22,479 ops/sec ±0.91% (95 runs sampled)

MONGOOSE_DEV=1 node create
CheckItem x 112,564 ops/sec ±0.26% (93 runs sampled)
Board x 122 ops/sec ±0.45% (77 runs sampled)

node populate
1528 completed queries on mongoose version 6.1.8
1528 completed queries on mongoose version 6.1.8

```

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1333,7 +1333,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
   // When using the $set operator the path to the field must already exist.
   // Else mongodb throws: "LEFT_SUBFIELD only supports Object"
 
-  if (parts.length < 2) {
+  if (parts.length <= 1) {
     pathToMark = path;
   } else {
     const len = parts.length;
--- a/lib/document.js
+++ b/lib/document.js
@@ -1135,7 +1135,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
       // them to ensure we keep the user's key order.
       if (type === true &&
           !prefix &&
-          path[key] != null &&
+          valForKey != null &&
           pathtype === 'nested' &&
           this._doc[key] != null) {
         delete this._doc[key];
@@ -1313,7 +1313,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
   let curPath = '';
   for (i = 0; i < parts.length - 1; ++i) {
     cur = cur[parts[i]];
-    curPath += (curPath.length > 0 ? '.' : '') + parts[i];
+    curPath += (curPath.length !== 0 ? '.' : '') + parts[i];
     if (!cur) {
       this.$set(curPath, {});
       // Hack re: gh-5800. If nested field is not selected, it probably exists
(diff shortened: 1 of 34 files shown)


Uzlopak:
I think the biggest performance bottleneck is the ES6 Proxy in DocumentArray and MongooseArray. Playing currently aroung with using WeakSets to see if the isMongooseArray and isMongooseDocumentArray can be boosted or not. 

Also trying to figure out if we can reduce the roundtrips through the proxy layer by using forEach etc. when iterating through the elements of those special arrays.

Uzlopak:
@vkarpov15 
Should i split it into more granular PRs?

lib/document.js
```diff
@@ -1333,10 +1333,11 @@ Document.prototype.$set = function $set(path, val, type, options) {
   // When using the $set operator the path to the field must already exist.
   // Else mongodb throws: "LEFT_SUBFIELD only supports Object"
 
-  if (parts.length <= 1) {
+  if (parts.length < 2) {
```
vkarpov15, line 1336: Is < faster than <=? If so, I'll make sure to use that going forward

lib/schematype.js
```diff
@@ -1306,42 +1316,43 @@ SchemaType.prototype.doValidateSync = function(value, scope, options) {
...
+    const v = validators[i];
 
     if (v == null || typeof v !== 'object') {
-      return;
+      continue;
     }
 
     const validator = v.validator;
-    const validatorProperties = utils.clone(v);
+    const validatorProperties = Object.assign({}, v);
```
vkarpov15, line 1338: This is a little risky because Mongoose clone() does a deep clone, so this may cause issues if there's a validator with nested objects

vkarpov15 (commented):
Couple of comments, but general idea looks good

lib/document.js
Uzlopak: This is an artifact. 
```suggestion
  if (parts.length <= 1) {
```

vkarpov15 (approved):
LGTM, I'll just make the `clone()` change in master.

PR #11313, opened 2022-02-02
Docs: Remove documentation for `useNestedStrict`

mark-langer:
**Summary**
This option has been deprecated in 6.0. It doesn't show up in code files any more when searching the repo, so I guess removing the docs is a good thing to do.

> If you're making a change to documentation, do **not** modify a `.html` file directly. Instead find the corresponding `.pug` file or test case in the `test/docs` directory.

The modified file is the only place the Github search finds the option `useNestedStrict`, therefore I edited that directly. Maybe it can be updated by recompiling the docs as well?

**Examples**
None


AbdelrahmanHafez (approved):
LGTM, thanks 👍

vkarpov15 (approved):
Thanks :+1:

PR #11317, opened 2022-02-02
convert Collection into a class

jimmywarting:
**Summary**

Wanted to modernize some things to classes, if you are happy with this kind of change, then there is more of this kind of PR ahead if you want

### DIFF
--- a/lib/collection.js
+++ b/lib/collection.js
@@ -4,43 +4,261 @@
  * Module dependencies.
  */
 
-const EventEmitter = require('events').EventEmitter;
+const EventEmitter = require('events');
 const STATES = require('./connectionstate');
 const immediate = require('./helpers/immediate');
 
-/**
- * Abstract Collection constructor
- *
- * This is the base class that drivers inherit from and implement.
- *
- * @param {String} name name of the collection
- * @param {Connection} conn A MongooseConnection instance
- * @param {Object} opts optional collection options
- * @api public
- */
+class Collection {
+  /**
+   * Abstract Collection constructor
+   *
+   * This is the base class that drivers inherit from and implement.
+   *
+   * @param {String} name name of the collection
+   * @param {Connection} conn A MongooseConnection instance
+   * @param {Object} opts optional collection options
+   * @api public
+   */
+  constructor(name, conn, opts) {
+    if (opts === void 0) {
+      opts = {};
+    }
+    if (opts.capped === void 0) {
+      opts.capped = {};
+    }
 
-function Collection(name, conn, opts) {
-  if (opts === void 0) {
-    opts = {};
+    if (typeof opts.capped === 'number') {
+      opts.capped = { size: opts.capped };
+    }
+
+    this.opts = opts;
+    this.name = name;
+    this.collectionName = name;
+    this.conn = conn;
+    this.queue = [];
+    this.buffer = true;
+    this.emitter = new EventEmitter();
+
+    if (STATES.connected === this.conn.readyState) {
+      this.onOpen();
+    }
   }
-  if (opts.capped === void 0) {
-    opts.capped = {};
+
+  /**
+   * Called when the database connects
+   *
+   * @api private
+   */
+  onOpen() {
+    this.buffer = false;
+    immediate(() => this.doQueue());
   }
 
-  if (typeof opts.capped === 'number') {
-    opts.capped = { size: opts.capped };
+  /**
+   * Called when the database disconnects
+   *
+   * @api private
+   */
+  onClose() { }
+
+  /**
+   * Queues a method for later execution when its
+   * database connection opens.
+   *
+   * @param {String} name name of the method to queue
+   * @param {Array} args arguments to pass to the method when executed
+   * @api private
+   */
+  addQueue(name, args) {
+    this.queue.push([name, args]);
+    return this;
   }
 
-  this.opts = opts;
-  this.name = name;
-  this.collectionName = name;
-  this.conn = conn;
-  this.queue = [];
-  this.buffer = true;
-  this.emitter = new EventEmitter();
+  /**
+   * Removes a queued method
+   *
+   * @param {String} name name of the method to queue
+   * @param {Array} args arguments to pass to the method when executed
+   * @api private
+   */
+  removeQueue(name, args) {
+    const index = this.queue.findIndex(v => v[0] === name && v[1] === args);
+    if (index === -1) {
+      return false;
+    }
+    this.queue.splice(index, 1);
+    return true;
+  }
 
-  if (STATES.connected === this.conn.readyState) {
-    this.onOpen();
+  /**
+   * Executes all queued methods and clears the queue.
+   *
+   * @api private
+   */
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez:
Welcome @jimmywarting 👋

I personally am not a fan of ES6 classes, and would use regular functions/assignments over the class syntax any day.

One of the reasons is that with functions assignments you can split your code across multiple files making for smaller files and you can re-use code whenever you want, which is something you can't do with ES6 classes.
```js
'use strict';

class Calculator {
  constructor() {}
  add(x, y) {
    return x + y;
  }
  subtract(x, y) {
    return x - y;
  }
  multiply(x, y) {
    return x * y;
  }
  divide(x, y) {
    return x / y;
  }
}
```

This is fine for files that are 50 or even 500 lines of code. However, when files grow to be over ~1000 lines of code, things can get difficult to work with.

With regular functions, you can split each method in a different file, which makes testing easier, smaller chunks of code to look at, 

```js
// calculator.js
const add = require('./add');
const subtract = require('./subtract');
const multiply = require('./multiply');
const divide = require('./divide');

function Calculator(){
  this.add = add;
  this.subtract = subtract;
  this.multiply = multiply;
  this.divide = divide;
}

// add.js
export default function add(x, y) {
  return x + y;
}
```

Not that the performance differences are great, but according to this [post](https://gregsolo.medium.com/es6-classes-vs-prototypes-performance-overview-dcab1e2fca9b), ES6 classes are slower. Admittedly, I would still pick regular functions assignments even if classes were slightly faster. Making the code easier to work with enables us to enhance performance than making the code difficult to work with and squeezing microseconds of performance.

One more reason which is a bit personal, I always find it a bit difficult to read code with `this` in it, so I try to avoid using things that depend on context whenever I can.


All that being said, many people would disagree with me. @ahmedelshenawy25 to name one.


Would happily know what you think @vkarpov15 @Uzlopak @IslandRhythms

ahmedelshenawy25:
@AbdelrahmanHafez personally I find following prototype every where in code is harder and I'd rather have everything encapsulated in its class, I'd feel less intimated to dig into the codebase that way.

Uzlopak:
Well to be honest I personally write more functional code. So I have not a strong opinion on this, if it is regarding classes or prototypical Object construction. For me are only three questions relevant:
1. Is the code in the same style like the rest of the codebase? Or is it planned in the near future to apply that new introduced code style.
2. Does it maybe help to reduce mistakes?
3. Does it have a significant performance drawback? If yes, is it the performance bottleneck in the hot path?

Based on this, I can determine that we already use ES6 Classes in the Codebase. So it would be actually indicating that an ES6 transformation is desired.

2. Not applicable

3. In my opinion the performance bottlenecks are not in this class/function. The hotpath is definitely in the bson deserialize and serialize and in various data mutations and geration of Objects like in clone. 

So based on this principles I think this PR fits to mongoose.

jimmywarting:
something that you can't do with prototype based is private class fields/methods. but that is easily worked around by using weakmaps...

Uzlopak:
I benchmarked and I can conclude that extending classes from builtins result in piss poor performance. With this in mind, I strongly recommend to avoid any classes

jimmywarting:
could you share the benchmark?

Uzlopak:
@jimmywarting 
https://github.com/Uzlopak/es6-classes-vs-prototypes

I modified the original benchmark to not create the class in a function. 

But I dont trust it very much, so I added my bench.js

```
node index.js 
Classes vs Protos:
Classes: Min:  0  Max:  5 Med:  1 Mean:  0.5814185814185814
Protos: Min:  0  Max:  3 Med:  1 Mean:  0.5314685314685315
Classes vs Protos (inheritance):
Classes: Min:  1  Max:  6 Med:  1 Mean:  1.3656343656343657
Protos: Min:  0  Max:  3 Med:  1 Mean:  0.5114885114885115
Classes vs Protos (inheritance from build-in object):
Classes: Min:  320  Max:  406 Med:  334 Mean:  337.13386613386615
Protos: Min:  0  Max:  2 Med:  0 Mean:  0.053946053946053944
```
```
node bench.js
Error x 351,821 ops/sec ±0.32% (95 runs sampled)
ExtendedError x 307,811 ops/sec ±0.39% (91 runs sampled)
Date x 782,318 ops/sec ±0.52% (94 runs sampled)
ExtendedDate x 781,884 ops/sec ±0.64% (93 runs sampled)
Number x 59,980,457 ops/sec ±0.24% (95 runs sampled)
ExtendedNumber x 54,283,834 ops/sec ±0.32% (94 runs sampled)
```

So in benchmark.js it has a perf hit of about 10 %. But in the original bench it has it has piss poor performance.

vkarpov15 (commented):
As written, this PR breaks our tests, because `lib/drivers/node-mongodb-native/collection` needs to change the way it inherits from `Collection`.

In general, I'm not convinced there's any benefit to us converting to classes. We have used classes in some files, but realistically all they add is an extra indent. Not sure it is worth the risk, especially since any place where we call `Collection()` without `new` will break. What do you think?

Uzlopak:
My thoughts on this:
If you are in the javascript world you see various names pop up very often. I know jimmywartings name for about 5 years now and he is a good programmer. I  think I saw him regularly that he proposes in various projects to convert code into ES6 Classes. This is per se nothing bad, on the contrary, it is good. 

Also in my personal experience, when people use Typescript and in their typescript code they use ES6 classes. But in transpilation they still target ES5 but use ES6 and newer as library to avoid actual ES6 classes in the transpiled code. You can see this in js-bson, but I think this is mere coincidence then an active decision :D. 

In my opinion performance should be the goal of every heavy used framework. if you have to leverage some so called bad practices, to achieve performance, then why not? What is the gain to have readable code but performance is not the best? E.g. a classical for loop should use a variable instantiated with var and not with let. 

f 
```javascript
for (var i = 0, il = arr.length; i < il; ++i) {
}
```

Now some js-devs will say "Hey eslint recommends to use let and not the old var, lets use everywhere let instead of var" - unknowingly reducing the performance of the for-loop. 

People complain about bad performance of mongoose. ES6 Classes are slower. So from my point of view, it should not be adapted. When the Errors were converted to ES6 Classes, the reason behind this was because the rewrite had supposedly performance gains and there was a benchmark on it (which I actually dont think was a reliable benchmark because it was just a profiling) but it was an active decision for performance. 

Maybe node 18 or 20 will have the same performance between classes and prototypical inheritance. Maybe node 22... or never. who knows. 

Lastly, we will have breaking compatibility. We would need to release a new major.

In my opinion a return of investment is not given and the PR should be rejected.

jimmywarting:
just did a own perf test on a rectangle class that extends a shape class in both prototype and es classes and calling a `move()` method... the prototype was faster in node, chrome, safari
but in Deno, it was equally fast

vkarpov15:
I'm just going to close this PR. While I appreciate the PR, based on discussion here I don't think merging this PR provides enough benefit to justify the effort to change our code to use Collection as a class and the potential risk that comes with that change.

Issue #11302, opened 2022-01-31
Mongoose can't straddle MongoDB 4.4 and 5?


JavaScriptBach:
I am using Mongoose 5.13.10 and would like to upgrade MongoDB from 4.4 to 5. However, the [docs](https://mongoosejs.com/docs/compatibility.html) imply that no version of Mongoose is compatible with both versions of MongoDB:

> Below are the semver ranges representing which versions of mongoose are compatible with the listed versions of MongoDB server.
> * MongoDB Server 4.4.x: mongoose ^5.10.0
> * MongoDB Server 5.x: mongoose ^6.0.0

Is this a documentation mistake? If not, how can I safely upgrade MongoDB?

IslandRhythms:
The docs are stating that MongoDB Server 4.4.x is compatible with all versions of mongoose equal to and higher than 5.10.0
and MongoDB Server 5.x is compatible with versions of mongoose equal to or higher than 6.

JavaScriptBach:
> The docs are stating that MongoDB Server 4.4.x is compatible with all versions of mongoose equal to and higher than 5.10.0 and MongoDB Server 5.x is compatible with versions of mongoose equal to or higher than 6. Meaning You can stay on 4.4 and use mongoose v6

In that case I think the docs need to be updated then, because ^ includes minor releases only (https://docs.npmjs.com/about-semantic-versioning), e.g. ^5.10.0 does not include 6 and higher. You can see that major releases are also explicitly called out separately in the other bullet points like this:

> * MongoDB Server 3.0.x: mongoose ^3.8.22, 4.x, or 5.x

vkarpov15:
In order to upgrade to MongoDB server 5.x, you would need to upgrade to Mongoose 6 first.

PR #11320, opened 2022-02-02
Mark Mongoose 6 as compatible with MongoDB 4

JavaScriptBach:
Based on the conversation in #11302, I assume Mongoose 6 is compatible with MongoDB 4. Update docs so people know that a safe upgrade is possible.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -11,9 +11,9 @@ Below are the [semver](http://semver.org/) ranges representing which versions of
 * MongoDB Server 3.2.x: mongoose `^4.3.0` or `5.x`
 * MongoDB Server 3.4.x: mongoose `^4.7.3` or `5.x`
 * MongoDB Server 3.6.x: mongoose `5.x`
-* MongoDB Server 4.0.x: mongoose `^5.2.0`
-* MongoDB Server 4.2.x: mongoose `^5.7.0`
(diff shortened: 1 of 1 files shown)


docs/compatibility.md
```diff
@@ -11,9 +11,9 @@ Below are the [semver](http://semver.org/) ranges representing which versions of
 * MongoDB Server 3.2.x: mongoose `^4.3.0` or `5.x`
 * MongoDB Server 3.4.x: mongoose `^4.7.3` or `5.x`
 * MongoDB Server 3.6.x: mongoose `5.x`
-* MongoDB Server 4.0.x: mongoose `^5.2.0`
-* MongoDB Server 4.2.x: mongoose `^5.7.0`
-* MongoDB Server 4.4.x: mongoose `^5.10.0`
+* MongoDB Server 4.0.x: mongoose `^5.2.0` or `6.x`
```
AbdelrahmanHafez, line 14: AFAIK, we don't drop support for older MongoDB versions as we upgrade versions, so I assume it's safe to change this with:
```suggestion
* MongoDB Server 4.0.x: mongoose >= `^5.2.0`
```

vkarpov15: I think that `^5.2.0` or `6.x` is less brittle, because otherwise what happens with Mongoose 7?

vkarpov15 (approved):
Thanks :+1:

Issue #11318, opened 2022-02-02
https://github.com/mongoosejs/mongoose-lean-getters/issues/11


vkarpov15:
https://github.com/mongoosejs/mongoose-lean-getters/issues/11

PR #11325, opened 2022-02-03
test: use tsd for tests re: #11311

vkarpov15:
@Uzlopak


Uzlopak:
Nice, so you want to completely replace the typescript tests with tsd?

vkarpov15:
@Uzlopak yes. I would prefer to get rid of the current iteration of TypeScript tests because they're very slow and brittle. tsd seems like it does exactly what we need: assert that certain lines error out.

The only reason I can think of to keep the current iteration of TypeScript tests is that the current architecture makes it easy to run tests with different TypeScript configs - some tests use strict mode, some don't, and we can configure other options as well.

Uzlopak:
The only reason why i would think that the Typescript Integration Test makes sense is to determine if we are able to import also as esm module like this and if it still works. 

```typescript
import * as mongoose from "mongoose"; // don't know how this is called
import mongoose from "mongoose"; // esm
import { connect } from "mongoose"; // named import
```

Like the issue was reported in #11318 could have been avoided, because we would have seen that the definitions not match on an code base level.

But on the other hand you just need to do the do once the holy Trinity of the exports like this, to ensure that esm import is also possible. 

```javascript
module.exports = mongoose
module.exports.default = mongoose
module.exports.mongoose = mongoose
```

So removing typescript integration makes fully sense, as the only breaking point as far as I know is the import, and that should be not a biggy if we just make the exports right...

Uzlopak:
Well actually for testing if it is possible to esm import it we would need just to test if default export and mongoose export is mongoose object. So no need for TS integration test at all.

Uzlopak:
How can we coordinate the tsd implementation? :)

PR #11327, opened 2022-02-04
perf: remove regexp-clone

Uzlopak:
Cloning an regexp does not need an external library.

### DIFF
--- a/lib/helpers/clone.js
+++ b/lib/helpers/clone.js
@@ -1,7 +1,5 @@
 'use strict';
 
-
-const cloneRegExp = require('regexp-clone');
 const Decimal = require('../types/decimal128');
 const ObjectId = require('../types/objectid');
 const specialProperties = require('./specialProperties');
@@ -144,5 +142,13 @@ function cloneArray(arr, options) {
     ret.push(clone(item, options, true));
   }
 
+  return ret;
+}
+
+function cloneRegExp(regexp) {
+  const ret = new RegExp(regexp.source, regexp.flags);
+  if (typeof regexp.lastIndex === 'number') {
+    ret.lastIndex = regexp.lastIndex;
+  }
   return ret;
 }
--- a/lib/helpers/clone.js
+++ b/lib/helpers/clone.js
@@ -146,7 +146,7 @@ function cloneArray(arr, options) {
 }
 
 function cloneRegExp(regexp) {
-  const ret = new RegExp(regexp);
+  const ret = new RegExp(regexp.source, regexp.flags);
 
   (typeof regexp.lastIndex === 'number') && (ret.lastIndex = regexp.lastIndex);
   return ret;
--- a/lib/helpers/clone.js
+++ b/lib/helpers/clone.js
@@ -148,6 +148,8 @@ function cloneArray(arr, options) {
 function cloneRegExp(regexp) {
   const ret = new RegExp(regexp.source, regexp.flags);
 
-  (typeof regexp.lastIndex === 'number') && (ret.lastIndex = regexp.lastIndex);
+  if (ret.lastIndex !== regexp.lastIndex) {
+    ret.lastIndex = regexp.lastIndex;
+  }
   return ret;
 }
--- a/lib/helpers/clone.js
+++ b/lib/helpers/clone.js
@@ -146,9 +146,8 @@ function cloneArray(arr, options) {
 }
 
 function cloneRegExp(regexp) {
-  const ret = new RegExp(regexp.source, regexp.flags);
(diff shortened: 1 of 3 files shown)


lib/helpers/clone.js
```diff
@@ -145,4 +143,11 @@ function cloneArray(arr, options) {
...
 
   return ret;
-}
\ No newline at end of file
+}
+
+function cloneRegExp(regexp) {
+  const ret = new RegExp(regexp);
+
+  (typeof regexp.lastIndex === 'number') && (ret.lastIndex = regexp.lastIndex);
```
AbdelrahmanHafez, line 151: I think it'd be easier to read if we convert this to a regular condition.

lib/helpers/clone.js
```diff
@@ -145,4 +143,11 @@ function cloneArray(arr, options) {
   }
 
   return ret;
-}
\ No newline at end of file
+}
+
+function cloneRegExp(regexp) {
+  const ret = new RegExp(regexp);
```
AbdelrahmanHafez, line 149: Looking at the original [regexp-clone](https://github.com/aheckmann/regexp-clone/blob/c44446929998bef515e15e4de4e73204c1a6b1f4/index.js#L14) code, we're copying the flags and adding them as a second argument for:
`new RegExp(originalRegexp, flags);`.

Do we not need to do that?

test/helpers/clone.test.js
```diff
@@ -148,9 +148,11 @@ describe('clone', () => {
 
     describe('constructor is RegExp', () => {
       it('return new equal date ', () => {
-        const base = new RegExp(/A-Z.*/);
+        const base = new RegExp(/A-Z.*/g);
+        base.lastIndex = 2;
         const cloned = clone(base);
         assert.deepStrictEqual(base, cloned);
```
AbdelrahmanHafez, line 154: Would be great if we could assert that the flags get copied over.
Not sure if `deepStrictEqual` checks for that.

AbdelrahmanHafez (changes requested):
Glad that we're getting rid of more and more dependencies.
Have a couple of questions and suggestions regarding the introduced changes, but I'm happy with that change.

test/helpers/clone.test.js
Uzlopak: I tested it. It checks for the flags. But it does not check for lastIndex. Thats why I added the separate lastIndex assertion.

lib/helpers/clone.js
Uzlopak: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/RegExp#parameters

Do we need to support less than ES2015? But I guess we can do it on the safe side.

Uzlopak:
```suggestion
  const ret = new RegExp(regexp.source, regexp.flags);
```

lib/helpers/clone.js
Uzlopak:
```suggestion
  If (typeof regexp.lastIndex === 'number') {
    ret.lastIndex = regexp.lastIndex;
  }
```

Uzlopak:
@AbdelrahmanHafez 
Fixed the comments.

AbdelrahmanHafez (approved):

Uzlopak:
@AbdelrahmanHafez 
@vkarpov15  
Also here some benchmarks:

regexp-clone x 2,177,401 ops/sec ±1.10% (87 runs sampled)
new RegExp(regex) x 4,872,620 ops/sec ±0.74% (91 runs sampled)
new RegExp(regex.source, regex.flags) x 5,274,696 ops/sec ±0.73% (89 runs sampled)

So intuitivly, we have implemented the fastest cloning

```javascript
const Benchmark = require('benchmark');
const cloneRegExp = require('regexp-clone');

const regex = /^a$/g;

new Benchmark.Suite()
  .add('regexp-clone', function () {
    const newRegex = cloneRegExp(regex)
  })
  .add('new RegExp(regex)', function () {
    const newRegex = new RegExp(regex);
    if (typeof regex.lastIndex === "number") {
        newRegex.lastIndex = regex.lastIndex;
    }
  })
  .add('new RegExp(regex.source, regex.flags)', function () {
    const newRegex = new RegExp(regex.source, regex.flags);
    if (typeof regex.lastIndex === "number") {
        newRegex.lastIndex = regex.lastIndex;
    }
  })
  .on('cycle', function (evt) {
    console.log(String(evt.target));
  })
  .run();
```

vkarpov15 (approved):
Nice, thanks :+1: I also agree about removing deps.

PR #11329, opened 2022-02-04
Fixed missing semicolon and use const instead of let

apeltop:
Fixed missing semicolon and use const instead of let. that can't be changed through reassignment


AbdelrahmanHafez (approved):
Thanks 👍

vkarpov15 (approved):
Thanks :+1:

PR #11330, opened 2022-02-04
Add parenthesis

apeltop:
There's no problem with existing code because there are no arguments in the function call.
But most other places use parentheses by no arguments in the function call. so I change it . 
I think there's a reason why you didn't change it.


AbdelrahmanHafez (approved):
LGTM, thanks 👍

vkarpov15 (approved):
Nice work, thanks :+1: no particular reason for us to prefer `new Date` over `new Date()`, the latter is more readable.

PR #11334, opened 2022-02-04
impr: better esm support, no necessity for setting allowSyntheticDefaultImports and esModuleInterop

Uzlopak:
This PR improves the esm support. 

you can now import mongoose in typescript with
```typescript
```
without being forced to set esModuleInterop or allowSyntheticDefaultImports to true.

I tested it with those settings set to true, and it doesnt break. So backwards compatibility, that means somebody set those settings to true in his existing project, is given.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,9 @@
+import events = require('events');
+import mongodb = require('mongodb');
+import mongoose = require('mongoose');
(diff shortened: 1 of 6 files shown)


Uzlopak:
created new PR to avoid annoying merge conflict solving.

PR #11337, opened 2022-02-05
Fix: Broken Link

saibbyweb:
Hyperlink for `Document.equals()` on this page: https://mongoosejs.com/docs/api/array.html#mongoosearray_MongooseArray-pull leads to (https://mongoosejs.com/docs/api/api.html#document_Document-equals) which is a 404 not found page.

The pull request intends to fix the broken link.

PR #11338, opened 2022-02-05
Fix: Broken Link (#11337) +1

saibbyweb:
The edit on Github button on this page: https://mongoosejs.com/docs/api/array.html#mongoosearray_MongooseArray-pull where I actually found the broken link #11337 is itself a broken link :|

I found another file with similar contents, not exactly sure which one corresponds to the actual website docs, so decided to modify this as well.


vkarpov15 (approved):
Thanks :+1:

PR #11344, opened 2022-02-07
remove npm script for symlinking mongoose

Uzlopak:
This is a PR following the tsd PR. We don't need to symlink any more, yeah!


AbdelrahmanHafez (approved):
This is great, thanks. 👍
The symlinks thing was a bit intimidating. I couldn't have it work on my Windows machine.

PR #11352, opened 2022-02-08
simplify a little bit in save handler

Uzlopak:
This is just a little bit golfing. Not a biggy.


twhy:
It's good to simplify the code, and the line below
```javascript
const matchedCount = !documentExists ? 0 : 1;
```
can just be
```javascript
const matchedCount = documentExists ? 1 : 0;
```

Uzlopak:
I guess i could even make it branchless by just doing

documentExists * 1

vkarpov15 (approved):

PR #11353, opened 2022-02-08
eachAsync: use nextTick instead of setTimeout and other small improvements

Uzlopak:


lib/helpers/cursor/eachAsync.js
```diff
@@ -23,22 +23,20 @@ const promiseOrCallback = require('../promiseOrCallback');
 
 module.exports = function eachAsync(next, fn, options, callback) {
   const parallel = options.parallel || 1;
-  const batchSize = options.batchSize;
+  const batchSize = options.batchSize || undefined;
```
vkarpov15, line 26: Why is this necessary? For one thing, this will cause `batchSize: 0` to throw an unreadable error.

lib/helpers/cursor/eachAsync.js
```diff
@@ -23,22 +23,20 @@ const promiseOrCallback = require('../promiseOrCallback');
...
-      if (typeof batchSize !== 'number') {
-        throw new TypeError('batchSize must be a number');
-      }
-      if (batchSize < 1) {
-        throw new TypeError('batchSize must be at least 1');
-      }
-      if (batchSize !== Math.floor(batchSize)) {
-        throw new TypeError('batchSize must be a positive integer');
-      }
+  if (typeof batchSize !== 'undefined') {
```
vkarpov15, line 29: Why move this out of `promiseOrCallback()`? We prefer to consistently return a promise, rather than sometimes return promise vs sometimes throwing a sync error - more like `async function()` semantics

vkarpov15 (changes requested):

lib/helpers/cursor/eachAsync.js
Uzlopak: Oh nice. Didn't thought about this.

Uzlopak: fixed

lib/helpers/cursor/eachAsync.js
Uzlopak: Correct.

Uzlopak:
@vkarpov15 
It should now be according to your liking :)

vkarpov15 (approved):
Thanks :+1:

PR #11356, opened 2022-02-08
simplify castPipelineOperator

Uzlopak:
Small refactor


Uzlopak:
Well, now I replaced all the lookup Objects with Sets and replaced a Regex with an indexOf, which is provable faster than the regex.

PR #11364, opened 2022-02-08
refactor fix of github issue 11295

twhy:
The original fix just fix the issue, but not telling people why.
`const versionBump = this.$__.version || this.$__schema.options.optimisticConcurrency;`

```javascript
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -336,9 +336,10 @@ Model.prototype.$__handleSave = function(options, callback) {
         callback(null, ret);
       });
     } else {
+      const optimisticConcurrency = this.$__schema.options.optimisticConcurrency;
       const optionsWithCustomValues = Object.assign({}, options, saveOptions);
       const where = this.$__where();
-      if (this.$__schema.options.optimisticConcurrency) {
+      if (optimisticConcurrency) {
         const key = this.$__schema.options.versionKey;
         const val = this.$__getValue(key);
         if (val != null) {
@@ -347,13 +348,14 @@ Model.prototype.$__handleSave = function(options, callback) {
       }
       this.constructor.exists(where, optionsWithCustomValues).
         then((documentExists) => {
-          if (!documentExists) {
+          if (documentExists) {
+            const matchedCount = 1;
+            callback(null, { $where: where, matchedCount });
+          } else {
+            if (optimisticConcurrency) { this.$__.version = VERSION_INC; }
             const matchedCount = 0;
             return callback(null, { $where: where, matchedCount });
           }
-
-          const matchedCount = 1;
-          callback(null, { $where: where, matchedCount });
         }).
         catch(callback);
       return;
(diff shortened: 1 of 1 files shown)


Uzlopak:
I modified the same place #11352 
Which one is better?

twhy:
I just comment on #11352 for a little improvement of your pull request.
For this pull, I think 
either
you merge your pull first and then I send a new pull based on latest code,
or
you merge my pull and then you send a new pull based on latest code.

I suggest the latter one because you can include the little improvement into one commit.

Uzlopak:
Tbh you made more thoughts into you PR. I am ok with retracting my PR.

vkarpov15:
I agree that this is an alternative fix, but I don't necessarily think it is an improvement. I appreciate the effort you put in to understanding the root cause of this issue, but I think the original solution in https://github.com/Automattic/mongoose/commit/d0c995203e01ec6d001c46ccbc3ac4288b0772ee is better because it lines up more closely with the semantics we want for optimistic concurrency: optimistic concurrency means we always bump the version. What do you think?

Uzlopak:
tbh, I am not quite following the discussion, as it is not really interesting for me (no offense - I try to improve the dx and performance). I just know that if it is making the code to be understood better in the future, than I would prefer a little bit more elaborate code than maybe the one liner. But I think we should first resolve the merge conflict and then we could condense it down to less code, anyway.

lib/model.js
```diff
@@ -347,13 +348,14 @@ Model.prototype.$__handleSave = function(options, callback) {
       }
       this.constructor.exists(where, optionsWithCustomValues).
         then((documentExists) => {
-          if (!documentExists) {
+          if (documentExists) {
+            const matchedCount = 1;
+            callback(null, { $where: where, matchedCount });
+          } else {
+            if (optimisticConcurrency) { this.$__.version = VERSION_INC; }
```
vkarpov15, line 355: This line is really what concerns me. I don't like the idea of bumping the version _after_ the operation is already done.

vkarpov15 (changes requested):
I took a closer look at this and you made a very good point with "shows doIncrement depends only on VERSION_INC and this.$__.version." `doIncrement` is what actually decides whether to increment the version, so the `if (optimisticConcurrency)` check I added works only by accident.

I think the better solution would be to just add `if (optimisticConcurrency) { this.$__.version = VERSION_ALL; }` on L340 rather than changing the version after `exists()`.

lib/model.js
twhy: I agree. 

The reason why I didn't set `this.$__.version = VERSION_ALL;`  is that the meaning of `VERSION_ALL` is not very clear to me as a newcomer(still not very clear), so I thought `VERSION_INC` is more appropriate. 

Now I know `VERSION_INC` means "to increase the version" instead of "version has been increased".

Let me close this PR and resend a new one based on the latest code since there's new code has been merged.

PR #11368, opened 2022-02-09
Extract PipelineStage typings from index.d.ts

Uzlopak:
Extracting typings from PipelineStage from index.d.ts
Started to use some of the examples from mongodb documentation for typings tests

Can be merged, we can add later more typing tests


.eslintrc.json
```diff
@@ -22,6 +23,8 @@
                 "@typescript-eslint"
             ],
             "rules": {
+                "@typescript-eslint/triple-slash-reference": "off",
+                "spaced-comment": ["error", "always", { "markers": ["/"] }],
```
vkarpov15, line 27: Why are these changes necessary?

Uzlopak: Without the /// Reference will be marked as linting error and will be auto fixed to // / and then the PipelineStages will not be imported and then the definitions will be broken.  Triple Slash is a typescript specific operation to import the definitions from another file.
With this change the Definitions will work.

types/index.d.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference path="./PipelineStage.d.ts" />
```
vkarpov15, line 1: While I do like breaking up `index.d.ts` into separate files, is it safe to rely on `reference path`? This means that Mongoose wouldn't work in libs that use [`noResolve`](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#using---noresolve)

Uzlopak: Interesting, never used noResolve before in my TS projects.

As I read noResolve means that typings like node typings with triple Slash reference won't throw an error if we don't have the node typings installed in our project.

I am currently just on my phone. But we could check this by setting noResolve in tsd configuration. If it throws it is an issue.

Uzlopak: I checked it. It will break. But it would also break in mongoose 6.2.0 before we merged some of my improvements. So we can safely assume that this is not an actual issue. Also I think before somebody would use noResolve he would use skipLibCheck, which should do what you actually would like to do with noResolve.

If this satisfying your question?

vkarpov15: @Uzlopak that makes sense, thanks for clarifying. I don't know enough about TypeScript to know whether anyone would actually use `--noResolve`, but it looks like that enough typings in DefinitelyTyped use `reference path` that anyone using noResolve will have a workaround. So I'll merge this :+1:

vkarpov15 (approved):

PR #11369, opened 2022-02-09
Typings: ValidationError should only contain ValidatorErrors or CastErrors

### LINKED ISSUE #11367: ValidationError.errors is wrongly typed
Report a (typescript) bug

The [documentation](https://mongoosejs.com/docs/validation.html) says that the errors property of a ValidationError is an object whose values are ValidatorErrors or CastErrors.
However in practice it can also be another ValidationError object (according to the typing). Because of this I am having issues accessing the kind property on those objects, because the kind property does not exist on a ValidationError object.
```
function handleValidationError(err: mongoose.Error.ValidationError): Array<string> {
  const errorTypes = Object.keys(err.errors).map((field: string) => err.errors[field].kind)
  return errorTypes
}
```

To reproduce, create a ValidationError instance and try to access the kind property on one of the items in the errors property like in the example above. Typescript will not compile because kind does not exist on type ValidationProperty.

tsconfig.json:
```
{
  "compilerOptions": {
    "target": "es2017",
    "module": "commonjs",
    "declaration": true,
    "outDir": "./lib",
    "strict": true
  },
  "include": ["src"],
  "exclude": ["node_modules", "**/__tests__/*"]
}
```

errors on the ValidationError class should only be of type ValidatorError or CastError and not of type ValidationError

node: 17.4.0
mongoose: 6.2.0
MongoDB: 5.0

Uzlopak:
Resolves #11367

Extracted Error-typings from index.d.ts 
renamed class Error to class MongooseError to ensure we extend from MongooseError and not from global.Error class
added typings test for #11367 
fixed the issue #11367


AbdelrahmanHafez (approved):
Nice, splitting types into multiple files will make things much easier to work with.
Thanks 👍

Uzlopak:
@AbdelrahmanHafez 

Thank you Hafez. Yes splitting the typings should make it easier to handle them better. I also extracted the Connection Typings in #11418 . Please have a look at it. Here i didnt write any additional tests, because it would be kind of waste of time. But in the Connection Types PR I wrote alot of typings tests.

vkarpov15 (approved):

Issue #11372, opened 2022-02-09
[TypeScript]  $search operator not correctly typed (highlight path option)


lmX2015:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

$search operator (aggregation) is not correctly typed (for the highlight option). For example

```typescript
import {Model, model, Schema} from "mongoose";

interface IUser {
    name: string;
    firstName: string;
    search(query: string): any;
}

// 2. Create a Schema corresponding to the document interface.
const schema = new Schema<IUser>({
    firstName: { type: String, required: true },
    name: { type: String, required: true },

});

schema.static("search", function(this: Model<IUser>, query: string) {
    return this.aggregate([
        {
            $search: {
                highlight: {
                       path : ["name", "firstname"], // "name" works fine
                },
                index: "searchLucene",
                text: {
                    path: ["name", "firstname"],
                    query,
                },
            },
        },
        {
            $project: {
                firstName: 1,
                highlights: { $meta: "searchHighlights" },
                name: 1,
                score: { $meta: "searchScore" },
        }},

    ]);
});
// 3. Create a Model.
const UserModel = model<IUser>("User", schema);


```

throws a compilation error (path should be a string not string[]).

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

From [MongoDB documentation](https://docs.atlas.mongodb.com/atlas-search/highlighting/#std-label-highlight-ref),  "highlight.path" can be a string array.
 So `highlight.path` should be at least `string | string []`
 
**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

-   mongoose: 6.2.1
-   typescript: 3.3.3
-   node : 16.3.0

vkarpov15:
Fixed in v6.2.2 with #11373

PR #11373, opened 2022-02-09
fix $search highlight path option type

lmX2015:
**Summary**

Fix this issue #11372 by adding all the possible types to highlight path ($search operator) following [MongoDB specs](https://docs.atlas.mongodb.com/atlas-search/highlighting/#std-label-highlight-ref)

**Examples**

Code provided in the issue is now working fine

```typescript
import {Model, model, Schema} from "mongoose";

interface IUser {
    name: string;
    firstName: string;
    search(query: string): any;
}

// 2. Create a Schema corresponding to the document interface.
const schema = new Schema<IUser>({
    firstName: { type: String, required: true },

});

schema.static("search", function(this: Model<IUser>, query: string) {
    return this.aggregate([
        {
            $search: {
                highlight: {
                       path : ["name", "firstname"], // "name" works fine
                },
                index: "searchLucene",
                text: {
                    path: ["name", "firstname"],
                    query,
                },
            },
        },
        {
            $project: {
                firstName: 1,
                highlights: { $meta: "searchHighlights" },
                name: 1,
                score: { $meta: "searchScore" },
        }},

    ]);
});
// 3. Create a Model.
const UserModel = model<IUser>("User", schema);
```

Note that this code requires an search index to be created on the database to be run.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -3199,7 +3199,12 @@ declare module 'mongoose' {
       $search: {
         [key: string]: any
         index?: string
-        highlight?: { path: string; maxCharsToExamine?: number; maxNumPassages?: number }
+        highlight?: {
+        /** [`highlightPath` reference](https://docs.atlas.mongodb.com/atlas-search/path-construction/#multiple-field-search) */
+          path: string | string[] | { value: string, multi: string};
+          maxCharsToExamine?: number;
+          maxNumPassages?: number;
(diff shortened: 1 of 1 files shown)


Uzlopak:
Please also add typings test for this modification

vkarpov15 (approved):
Thanks :+1:

Issue #11377, opened 2022-02-10
QueryCursor.js is throwing an error on line 144: "TypeError: Cannot assign to read only property 'map' of object '#<QueryCursor>'"


balonsom:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

QueryCursor.js is throwing an error on line 144:

`QueryCursor.prototype.map = function(fn) {`

Trying to assign to read only property 'map' of object

```
TypeError: Cannot assign to read only property 'map' of object '#<QueryCursor>'
    at Object.<anonymous> (/Users/balonsom/Documents/repos/dungeon-master/node_modules/mongoose/lib/cursor/QueryCursor.js:144:27)
    at Module._compile (node:internal/modules/cjs/loader:1097:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1151:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/Users/balonsom/Documents/repos/dungeon-master/node_modules/mongoose/lib/query.js:12:21)
    at Module._compile (node:internal/modules/cjs/loader:1097:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1151:10)
```

**If the current behavior is a bug, please provide the steps to reproduce.**

This bug only arose when I upgrade Node.js to v17.5.0. I don't think I can easily provide a script to replicate.

**What is the expected behavior?**

QueryCursor.js should not throw an error.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Mongoose: 6.2.1
Node.js: 17.5.0
MongoDB: 4.4

abeledovictor:
Same happening to
mongoose 5.7.7
MongoDB 4.4.12

Uzlopak:
I can reproduce it on node 17.5. 

@IslandRhythms  You dont need a repoductions script.

Just 
nvm install 17.5 && nvm use 17.5 && npm t

And voila you get that issue.

druv5319:
Having the same issues as well. Any fix?

druv5319:
Thanks it worked. My application automatically updated the Node engine to 17.5 and didn't realize what was causing the issue.

[...]

PR #11381, opened 2022-02-11
fix Node.js 17.5 compatibility

benjamingr:
The tc39 async iterator helpers spec adds `.map` to async iterators and Node added support for that.

As part of spec compliance Node uses defineProperty to add these properties.

This PR fixes setting `.map` in Mongoose over the property.

As a side note: it might make sense for Mongoose to deprecate `.map` since it's mutative which makes it named the same but with a different behavior from the tc39 iterator helpers proposal `.map`

Fixes: https://github.com/Automattic/mongoose/issues/11377

### DIFF
--- a/lib/cursor/AggregationCursor.js
+++ b/lib/cursor/AggregationCursor.js
@@ -137,11 +137,15 @@ if (Symbol.asyncIterator != null) {
  * @method map
  */
 
-AggregationCursor.prototype.map = function(fn) {
-  this._transforms.push(fn);
-  return this;
-};
-
+Object.defineProperty(AggregationCursor.prototype, 'map', {
+  value: function(fn) {
+    this._transforms.push(fn);
+    return this;
+  },
+  enumerable: false,
+  configurable: true,
+  writable: false
+});
 /*!
  * Marks this cursor as errored
  */
--- a/lib/cursor/QueryCursor.js
+++ b/lib/cursor/QueryCursor.js
@@ -141,10 +141,15 @@ QueryCursor.prototype._read = function() {
  * @method map
  */
 
-QueryCursor.prototype.map = function(fn) {
-  this._transforms.push(fn);
-  return this;
-};
+Object.defineProperty(QueryCursor.prototype, 'map', {
+    value: function(fn) {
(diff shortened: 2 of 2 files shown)


nateevans:
FYI - this change works, I used it as a patch, but the same issue exists here 
https://github.com/Automattic/mongoose/blob/master/lib/cursor/AggregationCursor.js#L140

benjamingr:
Thanks @nateevans anywhere else?

benjamingr:
Feel free to push to this branch or PR it if pushing directly doesn't work

nateevans:
@benjamingr that seems to be the 2 files with the issue

Grayson-code (approved):

Uzlopak:
@benjamingr 

Can you please run the lint fixer over your code?

npm run lint -- --fix

benjamingr:
@Uzlopak sure done

Slayer95:
Here it should be enumerable and configurable, shouldn't it?

Slayer95:
I mean, the current implementation does that, so the fix should keep it. Otherwise, it might be the better plan to update all methods to your approach.

Just pointing that out. Maintainer shall know better.

vkarpov15 (approved):
Thanks :+1: I'll make these properties enumerable and writable in master, if possible.

Eunit99:
This fix does not resolve #11377 as claimed. I had to downgrade my NodeJS to version 12.x. to resolve #11377. More information on how I did that is found in this [tutorial](https://dev.to/eunit/how-to-solve-querycursorjs-mongoose-error-with-nvm-2j4j).

Uzlopak:
Your tutorial is missing the point. Downgrading to node 12 is like saying: Hey your Tesla needs Electricity, but  to still use gasoline use a Ford Model-T. Despite the fact you could use a modern car, which uses gasoline.

Also our CI/CD Pipelines is testing for node18, which actually would fail if that issue was still existing. But the patch by @benjamingr  is solving that issue also for node18.

So to solve that issue, you would need to upgrade mongoose to latest version OR use any version of node 17 except 17.5.0 and node 18 and above. A downgrade to node 12 is not necessary.

PR #11387, opened 2022-02-13
add unit tests for arrayDepth

Uzlopak:
I wanted to see if there is any opportunity to improve the perf of arrayDepth. Could not improve the perf, but because I wanted to refactor it, I wrote unit tests for it.
So this is just covering all branches of this function.


vkarpov15 (approved):
Thanks :+1:

PR #11389, opened 2022-02-13
perf: use fast property access instead of get were possible

Uzlopak:
With this PR I replaced the get-method with fast property access were it is possible. I know that it looks not so sexy like the get call, but it has performance improvements as we dont need to split the path, traverse an array and stuff.

before:
$ MONGOOSE_DEV=1 node benchmarks/clone.js 
Simple x 140,319 ops/sec ±1.35% (92 runs sampled)
AllSchema x 23,986 ops/sec ±1.26% (90 runs sampled)

MONGOOSE_DEV=1 node benchmarks/validate.js 
invalid x 22,726 ops/sec ±0.62% (95 runs sampled)
valid x 61,786 ops/sec ±0.71% (93 runs sampled)

after
$ MONGOOSE_DEV=1 node benchmarks/clone.js 
Simple x 196,277 ops/sec ±0.72% (92 runs sampled)
AllSchema x 28,769 ops/sec ±0.41% (94 runs sampled)

$ MONGOOSE_DEV=1 node benchmarks/validate.js 
invalid x 23,488 ops/sec ±2.98% (95 runs sampled)
valid x 69,983 ops/sec ±1.05% (92 runs sampled

So atleast cloning is about 25 % faster.

We need more benchmarks imho... but hey. its something

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -102,15 +102,18 @@ module.exports = function cast(schema, obj, options, context) {
           const pathFirstHalf = split.slice(0, j).join('.');
           const pathLastHalf = split.slice(j).join('.');
           const _schematype = schema.path(pathFirstHalf);
-          const discriminatorKey = get(_schematype, 'schema.options.discriminatorKey');
+          const discriminatorKey = _schematype &&
+            _schematype.schema &&
+            _schematype.schema.options &&
+            _schematype.schema.options.discriminatorKey;
 
           // gh-6027: if we haven't found the schematype but this path is
           // underneath an embedded discriminator and the embedded discriminator
           // key is in the query, use the embedded discriminator schema
           if (_schematype != null &&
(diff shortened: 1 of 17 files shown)


AbdelrahmanHafez (approved):
Looks great, thanks 👍
Can't wait till we deprecate node >14 and use optional chaining inside mongoose.

vkarpov15 (approved):
Thanks, I'll merge this since this improves perf on code paths that are commonly used.

However, while we can't use optional chaining yet, I'd love to see if there's a way we could improve `get()`'s performance so we don't need as many `&&` checks. One thing we did to improve performance was accept array syntax, so you could do `get(_schematype, ['schema', 'options', 'discriminatorKey'])` and skip the perf overhead of `split()`. Does that help perf or is using `&&` still meaningfully faster?

PR #11399, opened 2022-02-15
PoC for Object assign

Uzlopak:
This PR is more a question. Why do we manually create an object and then assign each field from the source object to the target object, if we  can do the same with Object.assign?


vkarpov15 (approved):
Because Mongoose pre-dates `Object.assign()` by a few years, so there are places where we should use `Object.assign()`, but don't yet. Feel free to change to using `Object.assign()` instead of this sort of one-off manual replacement wherever you find it.

PR #11400, opened 2022-02-15
impr: isOperator

Uzlopak:
isOperator in aggregate used a .some() method to iterate over the only key in the array, so now it access it directly. Also added a null check for better safety...

the isOperator helper method used .startsWith method, which is slower than just checking the first character of the string. So I replaced it. 

It appears, the isOperator in aggregate and isOperator for queries have the same behaviour somehow, but the one for queries checks for the "special keys". Are those keys also special for the aggregation stage?

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -1105,13 +1105,13 @@ if (Symbol.asyncIterator != null) {
  */
 
 function isOperator(obj) {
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #11401, opened 2022-02-15
impr: hasDollarKeys

Uzlopak:
Small refactoring of hasDollarKeys
- added a name to the function
- the typecheck now uses !== instead of !=, as we just switched the conditions in the if-clause. So no implicit conversion of nullish values.
- uses direct character comparison instead of startsWith.


AbdelrahmanHafez (approved):
LGTM, thanks 👍

vkarpov15 (approved):
Thanks 👍

PR #11403, opened 2022-02-15
impr: aggregate unwind

Uzlopak:
- Array.from is faster for Iterators, like arguments
- added check if provided argument is not null
- replace startswith with faster single character check


Uzlopak:
I have to check something, closed for now

Issue #11397, opened 2022-02-14
Primitive array field fails when containing 10000 items or more


whatwhywhenandwho:
If a primitive array field has 10000 items or more, it fails on line `54` in `types/array/methods/index.js`:

    _basePush.apply(__array, values[i]);

With:

    Uncaught TypeError: CreateListFromArrayLike called on non-object.

The entire array logic looks like this:

    // Perf optimizations for small arrays: much faster to use `...` than `for` + `push`,
    // but large arrays may cause stack overflows. And for arrays of length 0/1, just
    // modifying the array is faster. Seems small, but adds up when you have a document
    // with thousands of nested arrays.
    if (len === 0) {
      __array = new Array();
    } else if (len === 1) {
      __array = new Array(1);
      __array[0] = values[0];
    } else if (len < 10000) {
      __array = new Array();
      _basePush.apply(__array, values);
    } else {
      __array = new Array();
      for (let i = 0; i < len; ++i) {
        _basePush.apply(__array, values[i]);
      }
    }

Why the `len < 10000` condition?

Can I create a PR and rewrite it like:

    if (len === 0) {
      __array = new Array();
    } else if (len === 1) {
      __array = new Array(1);
      __array[0] = values[0];
    } else {
      __array = [
        ...values
      ];
    }

?

Uzlopak:
Please not.

Hmm, did I fuck it up?

Has to be called with call and not apply

Uzlopak:
Yes I messed it up. Will provide a fix today.

whatwhywhenandwho:
Thank you @Uzlopak for such a swift response 🙏🏼

whatwhywhenandwho:
But sorry, just a question. Why is spreading not an option?

Uzlopak:
Imho spreading is slower than Array.from
Also I had some benchmark for this and it really seemed that this approach is a little bit faster than just using Array.from.
But I will bench again.

vkarpov15:
Yep it should be `.call()`, thanks @Uzlopak .

The 10k is an arbitrary limit to prevent stack overflows, see the comment. Basically, if the array is huge, using `...` or `.apply()` can cause a hard-to-diagnose error because `.apply()` puts every array value on the stack.

Uzlopak:
It is resolved with version 6.2.2
Can be closed

PR #11405, opened 2022-02-15
fix broken MongooseArray

Uzlopak:
@vkarpov15 
I fixed mongooseArray error #11397 but this PR is just about to fix the issue for now. I plan to benchmark further so that I know if there is a better solution (performance-wise).


whatwhywhenandwho:
❤️

AbdelrahmanHafez (approved):
LGTM, thanks 👍
Would be great if we could add a test case to prevent that from happening once again.

vkarpov15 (approved):
Nice, thanks :+1:

PR #11409, opened 2022-02-16
initial version of msgFunction

Uzlopak:
@AbdelrahmanHafez 
@vkarpov15 

I wonder if you can figure out what I did here ;).

### DIFF
--- a/lib/error/msgFunction.js
+++ b/lib/error/msgFunction.js
@@ -1,16 +1,16 @@
 'use strict';
 
 module.exports = function msgFunction(msg) {
-    const tokens = msg.split(/(\{.+?\})/g);
-    let fnBody = 'return \'';
-    for (let i = 0, len = tokens.length; i < len; ++i) {
-        const token = tokens[i];
-        if (token[0] === '{' && token[token.length - 1] === '}') {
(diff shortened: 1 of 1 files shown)


Uzlopak:
Dont merge this yet ;)

vkarpov15:
Faster handling of `{VALUE}` and other properties in error messages I presume?

Uzlopak:
Exactly :). But using this everywhere relevant seems to be kind of hard...

AbdelrahmanHafez:
Thanks @Uzlopak 

I can see that this is related to parsing validation error messages, but I'm finding it a bit difficult to understand how this is meant to be used.

Other than verifying that the code works, unit tests also show examples of how the author meant for the code to be used.
Would happily help replace the code all over the place to use this once I understand how to use it.

lib/error/msgFunction.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+module.exports = function msgFunction(msg) {
```
AbdelrahmanHafez, line 3: I've always found that it helps in code readability to name functions as verbs:
```js
const replaceMessage = getMessageReplacer(msg); // vs msgFunction(msg)
const parsedMesage = replaceMessage(someVariable);
```

AbdelrahmanHafez (changes requested):

PR #11411, opened 2022-02-16
docs(populate): add virtual populate `match` option documentation

remirobichet:
**Summary**

I was initialy looking for some documentation for a custom "filter" on populate virtuals. I couldn't find one anywhere.

My use case is the following : I need a virtual field for a filtered part of a collection (e.g. author's `posts` with `archived: false` property).


vkarpov15 (approved):
Thanks :+1:

PR #11412, opened 2022-02-17
perf: add isSimpleValidator

Uzlopak:
This PR adds a helper method called isSimpleValidator. Basically it is a function to determine if a validator is just a simple object or if it is a more complex one. To avoid the usage in other places, it is not named isFlatObject. It uses a WeakMap to cache the result and gain as such speed.

With isSimpleValidator we determine if we should make a deep clone of the validator or a shallow copy of it with Object.assign.

Before:

```
MONGOOSE_DEV=1 node validate
invalid x 23,461 ops/sec ±0.78% (89 runs sampled)
valid x 75,657 ops/sec ±0.67% (90 runs sampled)
```

after
```
MONGOOSE_DEV=1 node validate
invalid x 29,047 ops/sec ±0.37% (95 runs sampled)
valid x 148,950 ops/sec ±0.29% (93 runs sampled)
```

So we basically double the performance on the valid path.


Uzlopak:
If we combine this with #11408 we get super performance speed for validation :D

Uzlopak:
With the improved benchmark from #11415 
before: 

before:
```
MONGOOSE_DEV=1 node validate
invalid async x 21,179 ops/sec ±0.71% (84 runs sampled)
valid async x 41,408 ops/sec ±1.34% (83 runs sampled)
invalid sync x 24,553 ops/sec ±0.76% (93 runs sampled)
valid sync x 76,212 ops/sec ±0.67% (92 runs sampled)
```
after:
```
 MONGOOSE_DEV=1 node validate
invalid async x 23,496 ops/sec ±1.08% (82 runs sampled)
valid async x 60,767 ops/sec ±1.08% (80 runs sampled)
invalid sync x 27,526 ops/sec ±1.27% (92 runs sampled)
valid sync x 142,613 ops/sec ±0.36% (96 runs sampled)
```

AbdelrahmanHafez (approved):
LGTM 👍

lib/helpers/isSimpleValidator.js
```diff
@@ -0,0 +1,30 @@
...
+/*!
+ * Determines if `arg` is a flat object.
+ *
+ * @param {Object|Array|String|Function|RegExp|any} arg
+ * @api private
+ * @return {Boolean}
+ */
+const isSimpleValidatorCache = new WeakMap();
+
+module.exports = function isSimpleValidator(obj) {
```
vkarpov15, line 12: My only request would be to get rid of the caching. I know that it should be faster and the risk of `WeakMap()` causing any sort of memory issue is extremely low, but I'd rather write off that risk altogether and just keep this as a simple loop.

vkarpov15:
I'll merge this and remove the WeakMap caching locally

Uzlopak:
Seems like the performance is still very good. Strange. Expected a higher performance penalty. But hey, i dont complain.

PR #11413, opened 2022-02-17
add .lgtm.yml for lgtm.com

Uzlopak:
I use lgtm.com to also find some code smells. With this PR we only check the lib folder and avoiding noise in lgtm.com

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,0 +1,3 @@
+path_classifiers:
+  test:
+    - lib


vkarpov15:
I'd prefer to not go the route of adding additional code checks. I've worked on codebases that use similar tools - DeepScan, SonarCloud, etc. In my experience, they're noisy and overly persnickety.

Uzlopak:
We actually dont add another statistical code analysis engine. It is just a yml file for the lgtm webpage. If somebody wants to check on lgtm the code quality, it will show less potential code smells because examples folder and so will not be considered. 

Dont worry: I use these tools also selective. Sometimes they have some useful remark, and then I adapt it. 

Would be happy if you add this file.

vkarpov15:
In all honesty I'd prefer to not add this file, because it's an extra config file to maintain. But, given all the great contributions you two have made, I'll merge this one if it makes your lives easier. I just don't want to set up LGTM because I'm not sure my inbox can handle it :rofl:

AbdelrahmanHafez:
I understand your concerns, we're not trying to get all AI on code 😄

Honestly, I'm just excited to cheaply test a tool like that on a codebase of the scale of mongoose.
We can always get rid of it if it becomes an issue.

PR #11414, opened 2022-02-17
perf: improve validation sync and async by replacing forEach with classic for loops

Uzlopak:
It is basically replacing the forEach with classic for loop

before
```
MONGOOSE_DEV=1 node validate
invalid x 23,482 ops/sec ±3.63% (89 runs sampled)
valid x 75,238 ops/sec ±0.58% (91 runs sampled)
```

after
```
MONGOOSE_DEV=1 node validate
invalid x 27,460 ops/sec ±0.91% (89 runs sampled)
valid x 76,128 ops/sec ±0.69% (93 runs sampled)
```

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2853,19 +2853,21 @@ Document.prototype.validateSync = function(pathsToValidate, options) {
   }
   const validating = {};
 
-  paths.forEach(function(path) {
+  for (let i = 0, len = paths.length; i < len; ++i) {
+    const path = paths[i];
+
     if (validating[path]) {
-      return;
+      continue;
     }
 
     validating[path] = true;
 
     const p = _this.$__schema.path(path);
     if (!p) {
-      return;
+      continue;
     }
     if (!_this.$isValid(path)) {
-      return;
+      continue;
     }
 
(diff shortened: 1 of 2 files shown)


lib/document.js
```diff
@@ -2853,19 +2853,21 @@ Document.prototype.validateSync = function(pathsToValidate, options) {
...
-  paths.forEach(function(path) {
+  for (let i = 0, len = paths.length; i < len; ++i) {
+    const path = paths[i];
+
     if (validating[path]) {
-      return;
+      continue;
     }
 
     validating[path] = true;
```
Uzlopak, line 2863: why do we need to keep track if we are validated a path already?

vkarpov15: I don't recall exactly, but it has something to do with subdocuments or nested paths. Having both the parent path and child path modified at the same time can cause double validation.

vkarpov15 (approved):

lib/document.js
AbdelrahmanHafez: God bless unit tests.
In an ideal world, we can _not_ do that and see what test fails.

PR #11418, opened 2022-02-17
Improve Connection Typings

Uzlopak:
This PR extracts the Connection Typings

With the first commit I extract the typings from Connection from index.d.ts into a new Connection.d.ts. 
With the second commit I do the actual improvements.

The order of the function overloads are according to Typescript Do's and Don'ts, so basically by most arity to least arity.
Attributes like pass, user, models etc. were set to readonly. nobody should use them directly for manipulating the internals of mongoose. If there is some reason to modify them, a developer should use the corresponding methods.

Just check the second commit to see relevant changes.

Also added typings tests.

### DIFF
--- a/types/Connection.d.ts
+++ b/types/Connection.d.ts
@@ -60,32 +60,32 @@ declare module 'mongoose' {
       readonly db: mongodb.Db;
 
       /**
-         * Helper for `createCollection()`. Will explicitly create the given collection
-         * with specified options. Used to create [capped collections](https://docs.mongodb.com/manual/core/capped-collections/)
-         * and [views](https://docs.mongodb.com/manual/core/views/) from mongoose.
-         */
+       * Helper for `createCollection()`. Will explicitly create the given collection
+       * with specified options. Used to create [capped collections](https://docs.mongodb.com/manual/core/capped-collections/)
+       * and [views](https://docs.mongodb.com/manual/core/views/) from mongoose.
+       */
       createCollection<T extends AnyObject = AnyObject>(name: string, options: mongodb.CreateCollectionOptions, callback: Callback<mongodb.Collection<T>>): void;
       createCollection<T extends AnyObject = AnyObject>(name: string, callback: Callback<mongodb.Collection<T>>): void;
       createCollection<T extends AnyObject = AnyObject>(name: string, options?: mongodb.CreateCollectionOptions): Promise<mongodb.Collection<T>>;
 
       /**
-         * Removes the model named `name` from this connection, if it exists. You can
-         * use this function to clean up any models you created in your tests to
-         * prevent OverwriteModelErrors.
-         */
+       * Removes the model named `name` from this connection, if it exists. You can
+       * use this function to clean up any models you created in your tests to
+       * prevent OverwriteModelErrors.
+       */
       deleteModel(name: string): this;
 
       /**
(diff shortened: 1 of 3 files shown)


types/Connection.d.ts
```diff
@@ -0,0 +1,212 @@
...
+      readonly models: Readonly<{ [index: string]: Model<any> }>;
+
+      /** Defines or retrieves a model. */
+      model<T, U, TQueryHelpers = {}>(
+            name: string,
+            schema?: Schema<T, U, TQueryHelpers>,
+            collection?: string,
+            options?: CompileModelOptions
+        ): U;
+      model<T>(name: string, schema?: Schema<T>, collection?: string, options?: CompileModelOptions): Model<T>;
```
Uzlopak, line 127: Question to all:
shouldnt we do here:
model<T extends AnyObject = AnyObject>

With the current typings we are forcing a dev to type .model(). I think better DX is to make it optional for the dev.

types/Connection.d.ts
```diff
@@ -0,0 +1,212 @@
...
+         *
+         * - 0 = disconnected
+         * - 1 = connected
+         * - 2 = connecting
+         * - 3 = disconnecting
+         * - 99 = uninitialized
+         */
+      readonly readyState: ConnectionStates;
+
+      /** Sets the value of the option `key`. */
```
Uzlopak, line 166: here we had the information that set is the same as connection.options[key] = value.

The typings do not expose the options attribute, so I removed this remark here and the same for get

types/Connection.d.ts
```diff
@@ -0,0 +1,212 @@
...
+         * Returns the [MongoDB driver `MongoClient`](http://mongodb.github.io/node-mongodb-native/3.5/api/MongoClient.html) instance
+         * that this connection uses to talk to MongoDB.
+         */
+      getClient(): mongodb.MongoClient;
+
+      /**
+         * The host name portion of the URI. If multiple hosts, such as a replica set,
+         * this will contain the first host name in the URI
+         */
+      readonly host: string;
```
Uzlopak, line 105: please give me feedback regarding the readonly on host, user, pass etc. 
Is it wrong? Is there actually a use case for having them writeable?

If I am wrong, then I would like to fix it before we have regression reports

types/Connection.d.ts
```diff
@@ -0,0 +1,212 @@
...
+       */
+      readonly models: Readonly<{ [index: string]: Model<any> }>;
+
+      /** Defines or retrieves a model. */
+      model<T, U, TQueryHelpers = {}>(
+            name: string,
+            schema?: Schema<T, U, TQueryHelpers>,
+            collection?: string,
+            options?: CompileModelOptions
+        ): U;
```
Uzlopak, line 126: ReturnType U seems odd. Should it be Model<U> or does U should extend from Model?

vkarpov15: `U` should extend from Model, but I believe we removed that for perf reasons in #10349. `U extends Model` causes perf issues for the TS compiler

types/Connection.d.ts
vkarpov15: I think `readonly` should be fine, modifying `host` is a noop.

vkarpov15 (approved):

PR #11423, opened 2022-02-18
refactor model sync indexes tests

AbdelrahmanHafez:
This PR splits the large `Model.syncIndexes(...)` test case into smaller, more specific test cases.

PR #11424, opened 2022-02-18
fix(model): handle syncIndexes on discriminators

### LINKED ISSUE #11421: syncIndexes with discriminators will drop indexes from the base schema
node v16.13.2
mongoose v6.2.0

https://mongoosejs.com/docs/discriminators.html#discriminators

If you use a discriminator schema like this
```typescript
const Event = mongoose.model(
  "Event",
  new mongoose.Schema({
    actorId: { type: mongoose.Schema.Types.ObjectId, index: true },
  })
);

const AEvent = Event.discriminator(
  "AEvent",
  new mongoose.Schema({ aField: String })
);

const BEvent = Event.discriminator(
  "BEvent",
  new mongoose.Schema({ bField: String })
);
```

And your connection like this
```typescript
mongoose.connect( ... ).then((mongooseInstance) => {
  mongooseInstance.set("debug", true);
  mongooseInstance.syncIndexes();
});
```

You will see that the syncIndexes will cause the discriminators to drop the actorId index, but then it will be created again by the base schema. It should not be dropping the actorId index.

```
...
Mongoose: events.listIndexes()
Mongoose: events.dropIndex('actorId_1')
Mongoose: events.listIndexes()
...
Mongoose: events.listIndexes()
Mongoose: events.createIndex({ 'actorId': 1 }, { background: true })
...
```

### LINKED ISSUE #11422: indexes on discriminators with `partialFilterExpression` should merge discriminator key
According to line 1804, if one index on a discriminator has `partialFilterExpression`, mongoose will not add the `discriminatorKey` to the `partialFilterExpression` object.
https://github.com/Automattic/mongoose/blob/a499c91db09e930b071b75824c1d2af5119d86aa/lib/model.js#L1799-L1815

We should assign the value of `discriminatorKey` to the `partialFilterExpression` object, and disallow users from adding a different `discriminatorKey` to the `partialFilterExpression`, since it'll probably be a mistake.

AbdelrahmanHafez:
fixes #11421
fixes #11422
I had to do some refactoring before being able to work on the issue, I built on the refactored version, so merging #11423 would make the diff easier to work with.

This turned out to be a bit more than I anticipated, I think it might be a good idea to ship it in a minor version.
This PR has the following changes:
1- `Schema#indexes` now adds `partialFilterExpression` to any index defined on a discriminator schema, not only unique ones.
2- If a `partialFilterExpression` is already present as an option on a discriminator's index, we'll inject the discriminator key/value inside that filter.
3- Model.syncIndexes behavior for models with discriminators: 
* A base model with no discriminators will sync all the indexes.
* A base model with discriminators will only sync its indexes, not its children's indexes.
* A discriminator model will only sync the indexes defined on the discriminator schema, will not attempt to sync the base models' indexes.

4- `decorateDiscriminatorIndexOptions` now relies on the `schema` object, not the `model`, which allows us to use it inside `getIndexes(...)`.

I would recommend shipping #11430 in the same release to prevent any potential issues.


lib/helpers/indexes/filterIndexes.js
```diff
@@ -0,0 +1,38 @@
...
+    const partialFilterExpression = getPartialFilterExpression(index, { isSchemaIndexes });
+    if (!partialFilterExpression) {
+      return false;
+    }
+    return partialFilterExpression[discriminatorNameKey] === discriminatorNameValue;
+  });
+}
+
+function getPartialFilterExpression(indexObject, { isSchemaIndexes }) {
+  if (isSchemaIndexes) {
```
Uzlopak, line 34: can be smaller and clearer :)

```javascript
  return isSchemaIndexes
    ? indexObject.options && indexObject.options.partialFilterExpression
    : indexObject.partialFilterExpression;
```

lib/helpers/indexes/filterIndexes.js
```diff
@@ -0,0 +1,38 @@
...
+  return indexes.filter(index => {
+    const partialFilterExpression = getPartialFilterExpression(index, { isSchemaIndexes });
+    if (!partialFilterExpression) {
+      return false;
+    }
+    return partialFilterExpression[discriminatorNameKey] === discriminatorNameValue;
+  });
+}
+
+function getPartialFilterExpression(indexObject, { isSchemaIndexes }) {
```
Uzlopak, line 33: why do you wrap it into an object in line 25 and then to destructure it on the fly in line 33?

lib/helpers/indexes/filterIndexes.js
```diff
@@ -0,0 +1,38 @@
...
+
+function filterOutNonDiscriminatorIndexes(model, indexes, { isSchemaIndexes }) {
+  if (!model.baseModelName) {
+    return indexes;
+  }
+  const discriminatorNameKey = model.schema.discriminatorMapping.key;
+  const discriminatorNameValue = model.schema.discriminatorMapping.value;
+
+
+  return indexes.filter(index => {
```
Uzlopak, line 24: nit:  Isnt it possible to integrate the functionality of getPartialFilterExpress into the filter/ or put the whole logic of that fn into the getPartialFilterExpress? Seems to me that the separation of the functionality is actually artificial.

Uzlopak:
I just read the active code, not the unit tests nor do I actually (try to) understand what the purpose of this PR is LOL.

lib/helpers/schema/getIndexes.js
```diff
@@ -108,9 +110,11 @@ module.exports = function getIndexes(schema) {
       fixSubIndexPaths(schema, prefix);
     } else {
       schema._indexes.forEach(function(index) {
-        if (!('background' in index[1])) {
-          index[1].background = true;
+        const options = index[1];
+        if (!('background' in options)) {
+          options.background = true;
         }
+        decorateDiscriminatorIndexOptions(schema, options);
```
vkarpov15, line 117: Is it necessary to add the discriminator index options here as opposed to in `syncIndexes()`? As per our discussion in Slack, I'd like to avoid tying discriminators to schemas any more than we already do.

lib/helpers/indexes/decorateDiscriminatorIndexOptions.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = function decorateDiscriminatorIndexOptions(schema, indexOptions) {
+  // If the model is a discriminator and has an index, add a
+  // partialFilterExpression by default so the index will only apply
+  // to that discriminator.
+  const discriminatorName = schema.discriminatorMapping && schema.discriminatorMapping.value;
```
vkarpov15, line 7: Opportunity for refactoring here: pass in `discriminatorKey` and `discriminatorValue` instead of `schema`

vkarpov15 (approved):
I agree that the correct behavior here is that `syncIndexes()` on a discriminator should only sync the discriminator model's indexes, at least during `Connection.prototype.syncIndexes()`. I could see potentially adding an option to make `syncIndexes()` on a discriminator also syncIndexes on the parent model, but that isn't necessary because it is a one-liner to do `mongoose.model(model.baseModelName).syncIndexes()`.

Had a couple of minor comments, but overall LGTM :+1:

PR #11432, opened 2022-02-19
upgrade dependencies

AbdelrahmanHafez:


package.json
```diff
@@ -24,39 +24,37 @@
...
-    "babel-loader": "8.1.0",
+    "babel-loader": "8.2.3",
     "benchmark": "2.1.4",
     "bluebird": "3.7.2",
-    "cheerio": "1.0.0-rc.5",
-    "dox": "0.3.1",
-    "eslint": "8.5.0",
-    "eslint-plugin-mocha-no-only": "1.1.0",
+    "cheerio": "1.0.0-rc.10",
+    "dox": "0.9.0",
```
vkarpov15, line 42: I'd rather upgrade this dep in a separate PR because how we use dox is pretty brittle and we don't have a good way of testing that the dox upgrade didn't break anything other than reviewing all the API docs manually

package.json
```diff
@@ -24,39 +24,37 @@
...
-    "mkdirp": "0.5.5",
-    "mocha": "9.2.0",
+    "mocha": "9.2.1",
     "moment": "2.x",
-    "mongodb-memory-server": "^8.2.0",
+    "mongodb-memory-server": "^8.3.0",
     "nyc": "^15.1.0",
     "pug": "3.0.2",
     "q": "1.5.1",
-    "rimraf": "2.6.3",
```
vkarpov15, line 56: Thanks

vkarpov15 (changes requested):
One issue, otherwise LGTM

package.json
AbdelrahmanHafez: Happy to help

package.json
AbdelrahmanHafez: Done 👍

PR #11439, opened 2022-02-21
chore: update declarations for schema options

hiukky:
Added declaration for the `supressReservedKeysWarning`  schema property.

**Summary**

This PR adds a declaration for the `suppressReservedKeysWarning` schema option added in version 6.

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -264,3 +264,15 @@ function gh10789() {
     }
   });
 }
+
+function gh11439() {
+  type Book = {
+    collection: string
+  }
+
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (changes requested):
Thanks @hiukky 👍
Can you add a test case using that option inside `test\types\schema.test.ts` to make sure that this will never be accidentally removed in the future?

hiukky:
@AbdelrahmanHafez I added a simple test looking at some previous examples, can you check if it's up to standard!?

AbdelrahmanHafez (approved):
LGTM, thanks 👍

Issue #11295, opened 2022-01-29
DocumentNotFoundError instead of VersionError during save


mjfwebb:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
This is the same issue as described in #10974

**If the current behavior is a bug, please provide the steps to reproduce.**

Using the same script created by @IslandRhythms in https://github.com/Automattic/mongoose/issues/10974#issuecomment-970712009
```js
const mongoose = require('mongoose');
mongoose.set('debug', true)

const robotSchema = new mongoose.Schema({
  name: String
}, { optimisticConcurrency: true });

const Robot = new mongoose.model('Robot', robotSchema);

async function run() {
  mongoose.connect('mongodb://localhost:27017/', {
    useNewUrlParser: true,
    useUnifiedTopology: true,
  });
  await mongoose.connection.dropDatabase();
  const entry = await Robot.create({ name: 'WallE' });
  const changes = await Robot.findOne({ _id: entry._id }).orFail().exec();
  const other = await Robot.findOne({ _id: entry._id }).orFail().exec();
  changes.name = 'John';
  await changes.save();
  other.name = 'WallE';
  await other.save();
}

run();
```

I am getting a DocumentNotFound:

```
DocumentNotFoundError: No document found for query "{ _id: new ObjectId("61f578fc226164d0c2fb411e"), __v: 0 }" on model "Robot"
```

**What is the expected behavior?**

A VersionError should be thrown, not a DocumentNotFound error.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Mongoose: 6.1.8
Node: v16.6.1

twhy:
Although this has been fixed, I want to explain the cause of this bug and another possible fix, in case someone else is interested at this.

The cause of this bug is the line of code below. It assigns the same `name` value as created to `other.name`.
Remove this line and you still get the same error(`DocumentNotFoundError `).
```javascript
other.name = 'WallE'; 
```
If you assign a different value to `other.name`, you'll get `VersionError` as expected, because the original code covered this case.
```javascript
other.name = 'Steve';   // get VersionError as expected
```
But why?
Because `other.name = 'WallE'; ` didn't change `other`, so mongoose doesn't think `other` is 'dirty'.
But `other.name = 'Steve'` does change `other`, so Mongoose will know `other` is dirty and handle this differently. 

[...]

PR #11447, opened 2022-02-24
refactor fix of github issue 11295

twhy:
re #11295

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -355,7 +355,7 @@ Model.prototype.$__handleSave = function(options, callback) {
       this.constructor.exists(where, optionsWithCustomValues)
         .then(documentExists => {
           const matchedCount = documentExists ? 1 : 0;
-          if (matchedCount == 0 && optimisticConcurrency) {
+          if (matchedCount === 0 && optimisticConcurrency) {
             this.$__.version = VERSION_ALL;
           }
           callback(null, { $where: where, matchedCount });
--- a/lib/model.js
+++ b/lib/model.js
@@ -342,9 +342,10 @@ Model.prototype.$__handleSave = function(options, callback) {
         callback(null, ret);
       });
     } else {
+      const optimisticConcurrency = this.$__schema.options.optimisticConcurrency;
       const optionsWithCustomValues = Object.assign({}, options, saveOptions);
       const where = this.$__where();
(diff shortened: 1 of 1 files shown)


lib/model.js
```diff
@@ -353,7 +354,10 @@ Model.prototype.$__handleSave = function(options, callback) {
       }
       this.constructor.exists(where, optionsWithCustomValues)
         .then(documentExists => {
-          const matchedCount = !documentExists ? 0 : 1;
+          const matchedCount = documentExists ? 1 : 0;
```
Uzlopak, line 357: I think I could have coded it even smaller, when I originally wrote that line. 

```suggestion
          const matchedCount = !!documentExists;
```

lib/model.js
```diff
@@ -353,7 +354,10 @@ Model.prototype.$__handleSave = function(options, callback) {
...
+          if (matchedCount == 0 && optimisticConcurrency) {
```
Uzlopak, line 358: matchedCount of 0 is equivalent to 

```suggestion
          if (!matchedCount && optimisticConcurrency) {
```

Uzlopak: We could actually rename it from matchedCount to something like "hasMatchedDocument" or so.

Uzlopak:
Just added some nitpicks, which is more my fault not having them applied in the first PR.

On the other hand: I like that you were not demotivated after the initial discussion :)

twhy:
Sorry, I can't agree with your changes, and I have written down the reason. Please review.

The discussion is great, and it helps everyone learns more about the codebase, much better than just fix the issue.

Uzlopak:
Then atleast use triple equals.

twhy:
@Uzlopak Agree.

vkarpov15:
I put together an alternative fix for this in https://github.com/Automattic/mongoose/commit/fdc88a3bf3f622a26fb45800995e7e87c9ae98eb . You're right that the original fix for #11295 worked mostly by accident rather than by design. But this fix also isn't as correct as I'd like. The fundamental problem came down to that `this.$__.version` wasn't getting set when there were no modified paths, which was throwing off optimistic concurrency. I'll close this one, but thank you for digging into this in such detail :+1:

twhy:
No problem. It's great to see a better fix. Also, thank you for maintaining this library.

PR #11450, opened 2022-02-24
docs(timestamps): explain how `timestamps` works under the hood

### LINKED ISSUE #11336: [Docs] explain how timestamps work under the hood
I've come across a case where I needed to use an update with `upsert: true`, but I didn't want to modify the `updatedAt` if the document already exists in the collection, only wanted it to be inserted if I had an insert.

I solved that with:
```js
{
      filter: { email: adminToUpsert.email },
      update: {
        $addToSet: { roles: UserRoles.ADMIN },
        $setOnInsert: {
          password: hashedPassword,
          createdAt: new Date(),
          updatedAt: new Date()
        }
      },
      options: { upsert: true, timestamps: false }
    }
```
That way if we're updating, `updatedAt` should not change, but if we're inserting a new document, we'll have an `updatedAt` field.

It would be nicer if we could provide fine control:
```js
timestamps: { onUpdate: false, onInsert: true }
```

We could change the internal implementation for 
* `timestamps: true` to be the equivalent of `timestamps: { onUpdate: true, onInsert: true }`.
* `timestamps: false` to be the equivalent of `timestamps: { onUpdate: false, onInsert: false }`.

Would need to do that for:
* Model.updateOne
* Model.updateMany
* Model.bulkWrite

Would happily work on that if you think we should include it in mongoose.

What do you think?
@vkarpov15 @Uzlopak @IslandRhythms @ahmedelshenawy25

AbdelrahmanHafez:
fixes #11336

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -1105,6 +1105,11 @@ to your schema. The type assigned is [Date](./api.html#schema-date-js).
 By default, the names of the fields are `createdAt` and `updatedAt`. Customize
 the field names by setting `timestamps.createdAt` and `timestamps.updatedAt`.
 
+The way `timestamps` works under the hood is:
+* If you create a new document, mongoose simply sets `createdAt`, and `updatedAt` to the time of creation.
+* If you update a document, mongoose will add `updatedAt` to the `$set` object.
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

PR #11465, opened 2022-03-02
feat: add update-notifier to mongoose

AbdelrahmanHafez:
This PR makes it so that when people run their apps, they'll get a prompt informing them of an update when one is present:
Screenshot_458

### DIFF
--- a/lib/helpers/updateNotifier.js
+++ b/lib/helpers/updateNotifier.js
@@ -1,0 +1,14 @@
+'use strict';
+const updateNotifier = require('update-notifier');
+const pkg = require('../../package.json');
+
+run().catch(console.error);
+
+async function run() {
+  const notifier = await updateNotifier({
+    pkg,
+    updateCheckInterval: 0
+  });
+
+  notifier.notify();
+}
--- a/lib/index.js
+++ b/lib/index.js
@@ -39,6 +39,7 @@ const sanitizeFilter = require('./helpers/query/sanitizeFilter');
 const defaultMongooseSymbol = Symbol.for('mongoose:default');
 
 require('./helpers/printJestWarning');
+require('./helpers/updateNotifier.js');
 
 /**
  * Mongoose constructor.
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "mpath": "0.8.4",
     "mquery": "4.0.2",
     "ms": "2.1.3",
-    "sift": "16.0.0"
+    "sift": "16.0.0",
+    "update-notifier": "^5.1.0"
   },
   "devDependencies": {
     "@babel/core": "7.17.5",


sean-daley:
Is this really needed? Hopefully most people are using things like dependabot or renovate nowadays and already getting notified of things like this,

AbdelrahmanHafez:
It couldn't hurt. All the projects I worked on didn't have dependabot, I added it to them.
Developers relying on GitLab can't have dependabot installed easily.

Also, people can always [opt-out](https://github.com/yeoman/update-notifier#user-settings) if they want to.

sean-daley:
We have a couple of tools outside the repo which use this so we don’t want to globally disable it altogether. Does it have a mechanism where I can check some type of config setting into the repo so it can turn off this nagware? I checked but I don’t see anything. It looks like I’d have to get all of my devs as well as all of my systems to either set an environment variable (which from what I can tell globally disables which we don’t want to do) or check in a local config on the system to disable this (which it looks like we can disable on a module by module basis which is at least better but still annoying).

I guess I just don’t really see how it helps. There are so many ways with both npm and yarn to get notified of outdated dependencies (outside of dependabot and renovate) that forcibly nagging people just seems wrong to me. Hopefully not all of my 30+ dependencies decide to do this though.

But it’s your project so I can only voice my opinion.

AbdelrahmanHafez:
Your opinion is most welcome, and I understand your concerns.
I personally have always appreciated update notifications for dependencies, my policy is to always keep dependencies up-to-date.

Regarding your concern of globally disabling the update notifications, I can not understand why someone would prefer being notified for a subset of their dependencies, but I can see that it's a use-case.

If needed, I think we can add a config in package.json to disable notifications per dependency.
Would that work for you??

```json
{
  "updateNotifierConfig":{
    "optOutFromDependencies": ["mongoose"]
  }
}
```

sean-daley:
> Regarding your concern of globally disabling the update notifications, I can not understand why someone would prefer being notified for a subset of their dependencies, but I can see that it's a use-case.

I don't want to be notified for ANY of my dependencies. What I do want to be notified for is the tools that I use which are always installed globally and are not directly tied to a repo. In my particular case, npm and ncu both use update-notifier and I don't want to disable it for them because it's actually useful there as things like renovate and dependabot and the like aren't going to tell us anything about them.

simonbrunel:
I'm not familiar with `update-notifier` (so I may be wrong on some points) but I agree with @sean-daley. While it may be useful for tools such as npm or CLI-based dependencies (usually used as dev-dependencies), I wouldn't want to see these update messages when running *my* app (e.g. an API server). 

Also, I'm thinking of a few downsides: 

- it's not always possible to update the mongoose lib, so users will be forced to explicitly opt-out this feature
- these extra calls to the npm servers *may* cause security concerns for some users
- it injects [quite a few extra dependencies](https://github.com/yeoman/update-notifier/blob/main/package.json#L36) into the production code (even if opted-out)

As mentioned by @sean-daley, there is much better ways to monitor dependencies. Just running `npm outdated` on CI seems more efficient and useful *in this case* than having to go through the server logs.

vkarpov15:
I also disagree with this PR. @simonbrunel and @sean-daley make good points. In addition, I don't necessarily think we should push everyone to always upgrade to the latest version of Mongoose. In many cases, that simply isn't practical, and that's ok. We can always mark versions with known major issues using `npm deprecate`. But notifying every time there's a new update seems a bit heavy handed, especially since there's plenty of tools that can do that for you if you prefer that.

AbdelrahmanHafez:
Fair enough.
I personally prefer always keeping my dependencies up-to-date, and do appreciate notifications for new versions wherever possible, but I understand the opposing opinion. 👍

PR #11466, opened 2022-03-02
chore: add dependabot config file

AbdelrahmanHafez:
This will make dependabot check daily for new versions of packages that we use and create PRs for each dependency.

### DIFF
--- a/.github/.dependabot.yml
+++ b/.github/.dependabot.yml
@@ -6,3 +6,5 @@ updates:
     interval: "daily"
   open-pull-requests-limit: 100
   target-branch: master
+  ignore:
+    - dependency-name: "dox"
--- a/.github/.dependabot.yml
+++ b/.github/.dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: "daily"
+  open-pull-requests-limit: 100
+  target-branch: master
--- a/.github/.dependabot.yml
+++ b/.github/.dependabot.yml
@@ -3,7 +3,7 @@ updates:
 - package-ecosystem: npm
   directory: "/"
   schedule:
-    interval: "daily"
+    interval: "monthly"
   open-pull-requests-limit: 100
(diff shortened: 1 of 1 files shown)


vkarpov15:
I'd strongly prefer not to do this. My inbox just can't handle that much noise. And most of the time it isn't worth it for us to upgrade packages.

Uzlopak:
Maybe a compromise:

Fastify Framework changed recently all their dependabot configurations to report only monthly about new packages.
So how about also set it to monthly. Should be a good balance between having an eye on new packages and not detonating @vkarpov15 Inbox :D.

AbdelrahmanHafez:
@Uzlopak 's suggestion works well for me.
What do you think? @vkarpov15

vkarpov15:
Is there a way to mute updates for certain packages? For example, I don't want to go through the effort of updating `dox` yet. There's just too much risk for too much of the docs, without any way of testing the upgrade automatically. I also don't want a bot sending me reminders every month to upgrade `dox`.

Uzlopak:
Should be possible

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#ignore

Uzlopak:
But maybe we should first update non critical packages before we activate dependabot and get annoyed from the beginning.

vkarpov15:
Works for me. Can we also change it to only run quarterly, or at least monthly? Every day is a bit aggressive.

.github/.dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: "daily"
```
Uzlopak, line 6:
```suggestion
    interval: "monthly"
```

Uzlopak:
According to the dependabot docs only monthy is possible.

vkarpov15 (approved):
Thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

PR #11469, opened 2022-03-02
fix: rename definition files to lowercase to avoid typescript bug

### LINKED ISSUE #11468: TypeScript Build error: Cannot find type definition file for 'mongoose/types/pipelinestage'
Bug

In a TypeScript project that uses a package that imports Mongoose, these errors will occur on build in a case-sensitive OS (e.g. Linux):
```
TS2688: Cannot find type definition file for 'mongoose/types/pipelinestage'.
TS2688: Cannot find type definition file for 'mongoose/types/error'.
TS2688: Cannot find type definition file for 'mongoose/types/connection'.
```

I'm working on a minimal repro, I will post it here when I'm done.

Here is what I think is going on:
https://github.com/Automattic/mongoose/pull/11368, https://github.com/Automattic/mongoose/pull/11418, https://github.com/Automattic/mongoose/pull/11369 and split `index.d.ts` into separate PipelineStage, Connection, and Error type declaration files. These are referenced in `index.d.ts` with triple-slash directives:
```
/// <reference path="./Error.d.ts" />
/// <reference path="./PipelineStage.d.ts" />
/// <reference path="./Connection.d.ts" />
```
When Typescript generates a type declaration file `example-package.d.ts` for my package that imports Mongoose, it adds this to the top of the file:
```
/// <reference types="mongoose/types/PipelineStage" />
/// <reference types="mongoose/types/Error" />
/// <reference types="mongoose/types/Connection" />
/// <reference types="mongoose" />
```
Now, when I attempt to use `example-package` in another project, a bug In TypeScript (https://github.com/microsoft/TypeScript/issues/45096) causes paths in triple-slash directives to be lowercased. This mean  that the compiler sees 
```
/// <reference types="mongoose/types/PipelineStage" />
```
and looks for a declaration file at `node_modules/mongoose/types/pipelinestage.d.ts`. Since the real declaration file is `PipelineStage.d.ts`, on case-sensitive systems the compiler does not find any declaration file and throws an error. On case-insensitive systems the compiler finds the declaration file despite searching with the wrong case. 
To work around the Typescript bug, I recommend using lower-case filenames for type declaration files: `error.d.ts`, `pipeline-stage.d.ts`, and `connection.d.ts`. I have tries this locally and confirmed that it fixes my project. 

Here is a similar issue that was traced to the above TypeScript bug: https://github.com/microsoft/TypeScript/issues/45509

tsconfig.json:
```
{
  "compilerOptions": {
    "baseUrl": ".",
    "outDir": "dist",
    "rootDir": "."
    "allowJs": true,
    "declaration": true,
    "incremental": true,
    "jsx": "react",
    "lib": ["ES2020"],
    "module": "commonjs",
    "moduleResolution": "node",
    "pretty": true,
    "removeComments": true,
    "resolveJsonModule": true,
    "sourceMap": true,
    "target": "ES2020",
    "traceResolution": false
  }
  "include": ["src"]
}
```
Note: I have tried using the `forceConsistentCasingInFileNames` option. It has no effect on this issue.

Projects that depend on packages that depend on Mongoose should compile without errors on a case-sensitive OS.

Node: `v14.18.2`
Mongoose: `v6.2.4` (issue is present in versions `^6.2.2`)
MongoDB: `4.3.1`
TypeScript: `4.4.3`

Uzlopak:
Resolves #11468 

Renaming the filenames of the extracted definitions to lowercase should avoid the issue reported in the issue #11468 .


AbdelrahmanHafez (approved):
LGTM, thanks 👍

Issue #11473, opened 2022-03-03
Typescript no Type Hint for overriding Query.


minhthinhls:
@see {@link https://github.com/Automattic/mongoose/blob/master/types/index.d.ts#L2277}
On line 2277. The path type `path: string | any` should not contains any here !!!

Uzlopak:
Well the type expanding is an issue but I guess you want also have the ability to use query operators. I guess this is the reason for the any.

vkarpov15:
Fixed in #11475

Issue #11437, opened 2022-02-21
[typescript] define or derive typings for Query.select() and Query.findX(filter, projection) args relative to the given schema


the-vampiire:
- [x] using latest version of mongoose

**Do you want to request a *feature* or report a *bug*?**

- request a feature
- copied from my [SO question](https://stackoverflow.com/questions/71199832/how-to-define-the-the-arg-type-of-query-select-or-projection-string-of-query-f)

**What is the current behavior?**

> context

when using the projection argument of a `Query.findX(filter, projection)` or chaining the `Query.select(arg)` method i would like to have typing support.

> what is happening

the current typing for both `projection` and `select` args is `any`

**What is the expected behavior?**

> what i would like to happen

define the arg Type as the fields in my schema

> example model

```js
interface IUser {
  username: string;
  password: string;
}

interface IUserDocument extends IUser, Document {}

const UserSchema = new Schema<IUserDocument>({
  username: { type: String, required: true },
  password: { type: String, required: true },
});

const User = mongoose.model("User", UserSchema);

export default User;
```

> example querying

```js
User.findOne(
  // this is correctly picking up the fields / Types (although annoyingly mixed with Document properties...)
  { username: "some name" },
  // using projection string
  "/* intellisense for IUser fields as enum */",
)
// using select Query method
.select({ /* intellisense for IUser fields as { [field: string]: 0 | 1 */ });
```

> example implementation (not the best at TS, probably a better way)

```js
// derived from IUser, used for projection string
type UserFieldProjection = "username" | "password";

type QuerySelectOption = 0 | 1;

type SelectUserField {
  [field in UserFieldProjection]: QuerySelectOption;
}
```

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

- node: `v17.3.1`
- mongoose: `^6.2.0`
- mongodb: (derived from mongoose)

the-vampiire:
made a bit of progress on trying to make this generic:



```js
export type QuerySelectOption = 0 | 1;
export type PopulateOption<T> = string & keyof T;
export type SelectFields<T> = {
  [field in PopulateOption<T>]?: QuerySelectOption;
}

export type SelectUserFields = SelectFields<IUser>;

User
  .findOne({ username: "some name" })
  // manually define "as" but at least some typing is here now!
  .select({} as SelectUserFields);
```

unfortunately not able to get it working for the `populate` arg because it needs to repeat the options with a space between. 

[...]

PR #11475, opened 2022-03-04
Typescript no Type Hint for overriding Query.

minhthinhls:
https://github.com/Automattic/mongoose/issues/11473

**Summary**

**Examples**

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2274,7 +2274,7 @@ declare module 'mongoose' {
     polygon(path: string, ...coordinatePairs: number[][]): this;
 
     /** Specifies paths which should be populated with other documents. */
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez:
Hello @minhthinhls
Please refer to [this discussion](https://github.com/Automattic/mongoose/issues/11437#issuecomment-1047644655)  where we discuss why we need `any`, this change still wouldn't give hints to what fields it can add, it only tells us that the value is a string. 

What do you think?

vkarpov15 (approved):
Based on my reading of the code, removing `any` here should be fine. First arg to populate should either be a string, or `PopulateOptions`.

PR #11479, opened 2022-03-04
fix(typescript): `aggregate.sort()` accepts a string but also `{ field: 'asc'|'ascending'|'desc'|'descending' }`

simonbrunel:
When using TypeScript, calling `model.aggregate().sort("-name")` generates the following error:

```
Argument of type 'string' is not assignable to parameter of type 'Record<string, 1 | -1 | { $meta: "textScore"; }>'
```

This PR also adds support for mongoose sort keywords: `asc`, `ascending`, `desc` and `descending` (according to the [current implementation](https://github.com/Automattic/mongoose/blob/18cb0bed7256784701d7b75e160707895a0e50db/lib/aggregate.js#L573) and  the mongoose docs).

### DIFF
--- a/test/types/aggregate.test.ts
+++ b/test/types/aggregate.test.ts
@@ -1,4 +1,5 @@
 import { Schema, model, Document, Types } from 'mongoose';
+import { expectType } from 'tsd';
 
 const schema: Schema = new Schema({ name: { type: 'String' } });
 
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (changes requested):
Please add a test case verifying the type under test/typescript to avoid this being changed some time in the future.

AbdelrahmanHafez:
Other than the test case, LGTM.
Thanks 👍

simonbrunel:
> Please add a test case verifying the type under test/typescript to avoid this being changed some time in the future.

**+1000**, I though there was no test for types (I expected to find them under `types/test` and missed the ones in `test/types`).

Since there was no test for `aggregate.sort()`, I added them for all args supported by the current implementation. Note that `asc`, `ascending`, `desc`, `descending` was also not supported by the current type definition.

Uzlopak:
Yeah, it is nice to see that the typings tests

AbdelrahmanHafez (approved):
LGTM, Thanks 👍

PR #11481, opened 2022-03-05
$set is used for updatedAt [Timestamps]

saibbyweb:
The doc mentions that both `$setOnInsert` and `$set` are used for `createdAt`. I believe it's a mistake. It should be `$set` for `updatedAt`


AbdelrahmanHafez (approved):
Good catch, thanks 👍

Issue #11482, opened 2022-03-05
Better `typeKey` support in TypeScript, or drop support for it in TypeScript altogether


vkarpov15:
Re: #9927, #11241

The fact that `type` is optional in `SchemaTypeOptions` causes unpredictable behavior, because incorrect type definitions often end up as valid `SchemaTypeOptions` because of https://github.com/Automattic/mongoose/blob/3dd0461577523f70c149cc662c5dc2f18f05b485/types/index.d.ts#L1582. Something to think about.

Uzlopak:
Yeah, that should be removed imho, as it indicates that the schema type can handle more values. 

Was this added, because than people can add more attributes for their custom types?

Probably instead of having that big T extends bla stuff, we should use type as discriminator.

vkarpov15:
We added it because Mongoose supports schema definitions like this:

```javascript
new Schema({
  name: {
    $type: String // Notice '$type` here instead of `type`... because of `typeKey`
  }
}, { typeKey: '$type' });
```

The original rationale for this was to make it easier to define schemas that have properties named 'type'. Unfortunately, this sort of pattern doesn't work well with TypeScript, and we made `type` optional in SchemaTypeOptions to avoid completely blocking using `typeKey` in TypeScript.

Also, I linked to the wrong line. `[other: string]: any` is correct, because plugins often use additional schema type options. Like the `autopopulate` option from mongoose-autopopulate.

vkarpov15:
We have this in #11487 now, just need some docs.

agus-condro-speedoc:
> We added it because Mongoose supports schema definitions like this:
> 
> ```js
> new Schema({
>   name: {
>     $type: String // Notice '$type` here instead of `type`... because of `typeKey`
>   }
> }, { typeKey: '$type' });
> ```
> 
> The original rationale for this was to make it easier to define schemas that have properties named 'type'. Unfortunately, this sort of pattern doesn't work well with TypeScript, and we made `type` optional in SchemaTypeOptions to avoid completely blocking using `typeKey` in TypeScript.
> 
> Also, I linked to the wrong line. `[other: string]: any` is correct, because plugins often use additional schema type options. Like the `autopopulate` option from mongoose-autopopulate.

Has this been resolved?

In NestJs Typescript declaring the schema's field like that doesn't work, and the recommended way [here](https://mongoosejs.com/docs/guide.html#typeKey) also doesn't work in my VSCode auto suggestions

[...]

PR #11483, opened 2022-03-05
chore: extract and refactor typings schematypeoptions

Uzlopak:
Regarding #11482

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -2,9 +2,9 @@ import { Schema, Document, SchemaDefinition, Model } from 'mongoose';
 import { expectType, expectNotType, expectError } from 'tsd';
 
 enum Genre {
-  Action,
-  Adventure,
-  Comedy
+  Action = 'Action',
+  Adventure = 'Adventure',
+  Comedy = 'Comedy'
 }
 
 interface Actor {
@@ -27,7 +27,8 @@ interface Movie {
 const movieSchema = new Schema<Document & Movie, Model<Document & Movie>>({
   title: {
     type: String,
-    index: 'text'
+    index: 'text',
+    text: true
   },
   featuredIn: {
     type: String,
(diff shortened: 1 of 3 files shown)


Uzlopak:
Dont review/merge yet. I am not yet very satisfied with the changes.

test/types/schema.test.ts
```diff
@@ -1,10 +1,10 @@
-import { Schema, Document, SchemaDefinition, Model } from 'mongoose';
+import { Schema, Document, SchemaDefinition, Model, SchemaTypes } from 'mongoose';
 import { expectType, expectNotType, expectError } from 'tsd';
 
 enum Genre {
-  Action,
-  Adventure,
-  Comedy
+  Action = 'Action',
```
vkarpov15, line 5: Is this change necessary? I kinda like the default `enum` definition. Not a big deal either way since this is test code.

types/schematypeoptions.d.ts
```diff
@@ -0,0 +1,219 @@
...
+  interface ValidateOpts<T> {
+    msg?: string;
+    message?: string | ValidatorMessageFn<T>;
+    type?: string;
+    validator: ValidateFn<T> | LegacyAsyncValidateFn<T> | AsyncValidateFn<T>;
+  }
+
+  type SchemaValidator<T> = RegExp | [RegExp, string] | Function | [Function, string] | ValidateOpts<T> | ValidateOpts<T>[];
+
+  export class SchemaTypeOptionsClass implements
```
vkarpov15, line 33: This seems a little strange. Why not just use `typeof SchemaTypeOptions` like we currently do?

types/schematypeoptions.d.ts
```diff
@@ -0,0 +1,219 @@
...
+    StringSchemaTypeOptions,
+    NumberSchemaTypeOptions,
+    ObjectIdSchemaTypeOptions,
+    DateSchemaTypeOptions,
+    MapSchemaTypeOptions,
+    BufferSchemaTypeOptions,
+    BasicSchemaTypeOptions<any> {
+  }
+
+  export type SchemaTypeOptions<T> =
```
vkarpov15, line 43: Why the separate type? It seems like `BasicSchemaTypeOptions` does similar `T extends string` logic.

PR #11484, opened 2022-03-05
chore: extract and refactor schemaoptions

Uzlopak:
This is reviewable

Extracted SchemaOptions

removed emitIndexErrors
change typings for excludeIndexes and skipVersioning from any to the types how they are documented and how they are expected by the code.


types/schemaoptions.d.ts
```diff
@@ -0,0 +1,183 @@
...
+     * field names by setting timestamps.createdAt and timestamps.updatedAt.
+     */
+    timestamps?: boolean | SchemaTimestampsConfig;
+
+    /**
+     * Using `save`, `isNew`, and other Mongoose reserved names as schema path names now triggers a warning, not an error.
+     * You can suppress the warning by setting { supressReservedKeysWarning: true } schema options. Keep in mind that this
+     * can break plugins that rely on these reserved names.
+     */
+     supressReservedKeysWarning?: boolean
```
Uzlopak, line 181: Actually this is a typo. Should have been "suppress[...]" and not "supress[...]".

types/schemaoptions.d.ts
```diff
@@ -0,0 +1,183 @@
...
+     *
+     * @default true
+     */
+    selectPopulatedPaths?: boolean;
+    /**
+     * skipVersioning allows excluding paths from versioning (i.e., the internal revision will not be
+     * incremented even if these paths are updated). DO NOT do this unless you know what you're doing.
+     * For subdocuments, include this on the parent document using the fully qualified path.
+     */
+    skipVersioning?: {[key: string]: boolean; };
```
Uzlopak, line 161: I could not find skipVersioning in the lib folder. Is this obsolete?

types/schemaoptions.d.ts
vkarpov15: You're right, that's something we should change in Mongoose 7. I opened up #11495 to track that

vkarpov15 (approved):
Thanks :+1:

Issue #11449, opened 2022-02-24
Typescript typings missing for next() function


rkreienbuehl:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
As described in the docs, in pre-hooks the possibility to use `next()` is still there. I am about to update our code to mongoose `6.2.3` and we have a lot of mongoose middleware that uses `next()`.
Now all the imports for `HookNextFunction` don't work anymore because this type definition was removed with mongoose 6.

**If the current behavior is a bug, please provide the steps to reproduce.**

```json
{
  "compileOnSave": false,
  "compilerOptions": {
    "rootDir": ".",
    "sourceMap": true,
    "declaration": false,
    "moduleResolution": "node",
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "esModuleInterop": true,
    "importHelpers": true,
    "target": "es2015",
    "module": "esnext",
    "lib": ["es2017", "dom"],
    "skipLibCheck": true,
    "skipDefaultLibCheck": true,
    "baseUrl": ".",
    "paths": {}
  },
  "exclude": ["node_modules", "tmp"]
}
```

**What is the expected behavior?**
To have a type definition for the next function.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
mongoose: `6.2.3`
node.js: `v16.13.1`

Uzlopak:
If I understood it correctly, you want to type your hook functions but you are lacking the typings of the hook functions. PR #11485 exposes the Middleware Function Types. Would this solve your issue?

vkarpov15:
I think the question is what to do now that `HookNextFunction` is no longer exported. You can still use `next()` callbacks with TypeScript in Mongoose 6, we just removed `HookNextFunction` because `HookNextFunction` is just `(err?: Error) => void`. So if you did this in Mongoose 5:

```ts
schema.pre('save', function(next: HookNextFunction) {
 // ...
});
```

You can just do this in Mongoose 6:

```ts
schema.pre('save', function(next: (err?: Error) => void) {
  // ...
});
```

wasit-shafi:
Just for Additional Info: For those interested(including me) in looking at details of when the support for type **HookNextFunction** was removed refer to official docs: https://mongoosejs.com/docs/migrating_to_6.html#typescript-changes

PR #11485, opened 2022-03-06
Impr: export and refactor types of PreMiddlewareFunction, PreSaveMiddlewareFunction, PostMiddlewareFunction, ErrorHandlingMiddlewareFunction

Uzlopak:
If I understood #11449 correctly, the ability to type the hook function is not given, because the types are not exposed.


vkarpov15 (approved):
Thanks :+1:

PR #11486, opened 2022-03-06
chore: extract and refactor document type

Uzlopak:


types/document.d.ts
```diff
@@ -0,0 +1,244 @@
+import mongodb = require('mongodb');
+
+declare module 'mongoose' {
+
+  /** A list of paths to skip. If set, Mongoose will validate every modified path that is not in this list. */
+  type pathsToSkip = string[] | string;
+
+  class Document<T = any, TQueryHelpers = any, DocType = any> {
+    constructor(doc?: any);
```
Uzlopak, line 9: Can this be

```suggestion
    constructor(doc?: Partial<T>);
```

?

vkarpov15: Unfortunately not. Probably more correct to be `doc ?: AnyObject`

vkarpov15 (approved):
Thanks :+1:

PR #11488, opened 2022-03-06
chore: extract and refactor aggregationcursor and querycursor

Uzlopak:
I unified the typings of aggregationcursor and querycursor to Cursor-type.

Now eachAsync should be more precise, as the DocType is an Array if batchSize is defined. This was not the case previously in AggregationCursor.


vkarpov15 (approved):
Thanks :+1:

PR #11489, opened 2022-03-06
fix: add missing timeseries expiration handling

### LINKED ISSUE #11229: expireAfterSeconds schema option not valid and not effective on timeseries collection
bug

I use moongose for create timeseries collection
my schema is like this : 
```
const schema = new mongoose.Schema({ timestamp: Date, metadata: metricMetadataSchema },
  {
    _id: false,
    strict: false,
    autoIndex: false,
    autoCreate: false,
    timeseries: {timeField: "timestamp", metaField: "metadata", granularity: 'minutes'},
    // @ts-ignore
    expireAfterSeconds: 60
  }
)
```
At this point the property expireAfterSeconds is not valid for typescript because expireAfterSeconds not include in type SchemaOptions
The collection is well  done created, but no info about expireAfterSeconds anywhere  and after 60 seconds data not removed.

```
{
  "compilerOptions": {
    "module": "commonjs",
    "declaration": true,
    "removeComments": true,
    "target": "es2017",
    "sourceMap": true,
    "outDir": "./dist",
    "baseUrl": "./",
    "incremental": true,
    "skipLibCheck": true
  }
}
```

node: v14.17.0
mongodb: 4.3.0
mongoose: 6.1.6

Uzlopak:
Resolves #11229 

Didnt know if I should just add expiresAfterSeconds which does only pass through the value to createCollection or like we do it in indexes with expires where we also handle date strings. So I implemented both. Removing is easier than adding.

Added unit test

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1375,6 +1375,12 @@ Model.createCollection = function createCollection(options, callback) {
   this.schema.options.timeseries;
   if (timeseries != null) {
     options = Object.assign({ timeseries }, options);
+    if (this.schema.options.expireAfterSeconds) {
+      options.expireAfterSeconds = this.schema.options.expireAfterSeconds;
+    } else if (this.schema.options.expires) {
+      options.expires = this.schema.options.expires;
+      utils.expires(options);
+    }
   }
 
   callback = this.$handleCallbackError(callback);
--- a/lib/model.js
+++ b/lib/model.js
@@ -1375,9 +1375,13 @@ Model.createCollection = function createCollection(options, callback) {
   this.schema.options.timeseries;
   if (timeseries != null) {
     options = Object.assign({ timeseries }, options);
-    if (this.schema.options.expireAfterSeconds) {
+    if (options.expireAfterSeconds != null) {
+      // do nothing
+    } else if (options.expires != null) {
+      utils.expires(options);
+    } else if (this.schema.options.expireAfterSeconds != null) {
       options.expireAfterSeconds = this.schema.options.expireAfterSeconds;
-    } else if (this.schema.options.expires) {
+    } else if (this.schema.options.expires != null) {
       options.expires = this.schema.options.expires;
       utils.expires(options);
     }
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -659,8 +659,8 @@ declare module 'mongoose' {
      * mongoose will not create the collection for the model until any documents are
      * created. Use this method to create the collection explicitly.
      */
-    createCollection<T>(options?: mongodb.CreateCollectionOptions): Promise<mongodb.Collection<T>>;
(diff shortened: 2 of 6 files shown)


types/index.d.ts
```diff
@@ -1301,6 +1301,12 @@ declare module 'mongoose' {
     /** The timeseries option to use when creating the model's collection. */
     timeseries?: mongodb.TimeSeriesCollectionOptions;
 
+    /** The number of seconds after which a document in a timeseries collection expires. */
+    expireAfterSeconds?: number;
+
+    /** The time after which a document in a timeseries collection expires. */
+    expires?: number | string;
```
simonbrunel, line 1308: Could make this even safer by using the [StringValue  type provided by the ms lib](https://github.com/vercel/ms/blob/master/src/index.ts#L44):

```ts
import type { StringValue as MsStringValue } from 'ms';

// ...

/** The time after which a document in a timeseries collection expires. */
expires?: number | MsStringValue;
```

Uzlopak: That ist good to know. @vkarpov15 should give the Feedback how he wants to set the expiration and then I would also fix that with the last commit.

Uzlopak:
Maybe instead of naming it expiresAfterSeconds or expires calling it timeseriesExpires is better?

types/index.d.ts
Uzlopak: I checked it. ms does not provide typings...

simonbrunel: Indeed, types will be available in [version 3](https://github.com/vercel/ms#typescript-support).

lib/model.js
```diff
@@ -1375,6 +1375,12 @@ Model.createCollection = function createCollection(options, callback) {
   this.schema.options.timeseries;
   if (timeseries != null) {
     options = Object.assign({ timeseries }, options);
+    if (this.schema.options.expireAfterSeconds) {
```
vkarpov15, line 1378: Let's add a check for `if (options.expireAfterSeconds != null)`. We don't want the schema option to override user-provided options if someone calls `createCollection({ expireAfterSeconds })`.

So I think this should look like:

```javascript
if (options.expireAfterSeconds != null) {
  // ...
} else if (options.expires != null) {
  // ..
} else if (this.schema.options.expireAfterSeconds != null) {
  // ...
} else if (this.schema.options.expires != null) {
  // ...
}
```

vkarpov15 (changes requested):

lib/model.js
Uzlopak: Note to myself: I guess typings of parameter options of createCollection has also be adapted.

Uzlopak:
@vkarpov15 

Adapted changes. Added unit and typings tests. Changed typings.

lib/model.js
Uzlopak: done

vkarpov15 (approved):
Thanks :+1:

PR #11493, opened 2022-03-07
fix: length comparaison in lean.test.js

zazapeta:
**Summary**

Fixed length comparaison between POJO and Mongoose doc in lean.test.js


AbdelrahmanHafez (approved):
Welcome @zazapeta

Good catch, thanks 👍

Issue #11496, opened 2022-03-07
Typescript Populate Issue


vkarpov15:
### Discussed in https://github.com/Automattic/mongoose/discussions/11477

<div type='discussions-op-text'>

<sup>Originally posted by **supertiger1234** March  4, 2022</sup>
```ts
export interface Friend {
  blocked: boolean
}
const FriendSchema = new Schema<Friend>({
  blocked: Boolean
})
const Friends = model<Friend>("friends", FriendSchema);


export interface User {
  friends: Types.ObjectId[];
}
const UserSchema = new Schema<User>({
  friends: [{ type: Schema.Types.ObjectId, ref: 'friends' }]
})
const Users = model<User>("friends", UserSchema);


Users.findOne({}).populate<{friends: Friend[]}>('friends').then(user => {
  user?.friends[0].blocked // works
  user?.friends.map(friend => friend.blocked) // Property 'blocked' does not exist on type 'ObjectId'.ts(2339)
})
```
how do i fix this typescript issue? When using `map` I get the error `Property 'blocked' does not exist on type 'ObjectId'.ts(2339)`. 

Mongoose version: 6.2.4
</div>

Issue #11041, opened 2021-12-03
v6: populate makes single document query type non-null


aaronadamsCA:
All of these should have the same type, but they do not:

```ts
import { model } from "mongoose";

const MyModel = model("MyModel");

let maybeDoc = await MyModel.findById(null).exec();
typeof maybeDoc; // (Document<any, any, unknown> & { _id: Types.ObjectId }) | null

maybeDoc = await MyModel.findById(null).populate("someField").exec();
typeof maybeDoc; // Document<any, any, unknown> & { _id: Types.ObjectId }

maybeDoc = await MyModel.findOne({}).exec();
typeof maybeDoc; // (Document<any, any, unknown> & { _id: Types.ObjectId }) | null

maybeDoc = await MyModel.findOne({}).populate("someField").exec();
typeof maybeDoc; // Document<any, any, unknown> & { _id: Types.ObjectId }
```

This worked as expected in Mongoose 5.13.13. During migration, ESLint flagged some "unnecessary" `null` checks.

Node.js 16.13.1, Mongoose 6.0.14.

IslandRhythms:
`_id` is unknown for me and there is no `| null`.

aaronadamsCA:
I've provided a reproduction here:

https://codesandbox.io/s/mongoose-issue-11041-q74uu?file=/index.ts

PR #11503, opened 2022-03-08
fix: improve populate type narrowing 

mohammad0-0ahmad:
Should resolve https://github.com/Automattic/mongoose/issues/11496
@Uzlopak

### DIFF
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -167,7 +167,6 @@ function gh11503() {
   });
   const Users = model<User>('friends', UserSchema);
 
-
   Users.findOne({}).populate('friends').then(user => {
     expectType<Types.ObjectId | undefined>(user?.friends[0]);
     expectError(user?.friends[0].blocked);
@@ -175,14 +174,11 @@ function gh11503() {
   });
 
   Users.findOne({}).populate<{friends: Friend[]}>('friends').then(user => {
-    expectAssignable<Friend | undefined>(user?.friends[0]);
-    expectType<boolean | undefined>(user?.friends[0].blocked);
-    expectType<Types.ObjectId | undefined>(user?.friends[0]._id);
-
+    expectAssignable<Friend>(user?.friends[0]);
+    expectType<boolean>(user?.friends[0].blocked);
+    expectType<Types.ObjectId>(user?.friends[0]._id);
     const firstFriendBlockedValue = user?.friends.map(friend => friend)[0];
-
-    // FIXME: These two tests doesn't work, although it shows correct suggestions.
-    expectType<boolean | undefined>(firstFriendBlockedValue?.blocked);
+    expectType<boolean>(firstFriendBlockedValue?.blocked);
     expectType<Types.ObjectId>(firstFriendBlockedValue?._id);
   });
 }
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -176,7 +176,13 @@ function gh11496() {
 
   Users.findOne({}).populate<{friends: Friend[]}>('friends').then(user => {
     expectAssignable<Friend | undefined>(user?.friends[0]);
-    user?.friends[0].blocked;
-    user?.friends.map(friend => friend.blocked);
+    expectType<boolean | undefined>(user?.friends[0].blocked);
+    expectType<Types.ObjectId | undefined>(user?.friends[0]._id);
+
+    const firstFriendBlockedValue = user?.friends.map(friend => friend)[0];
+
+    // FIXME: These two test doesn't work, although it shows correct suggestions.
+    expectType<boolean | undefined>(firstFriendBlockedValue?.blocked);
+    expectType<Types.ObjectId>(firstFriendBlockedValue?._id);
   });
 }
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -151,7 +151,7 @@ function gh11321(): void {
   });
 }
 
-function gh11496() {
+function gh11503() {
   interface Friend {
     blocked: boolean
   }
@@ -181,7 +181,7 @@ function gh11496() {
 
     const firstFriendBlockedValue = user?.friends.map(friend => friend)[0];
 
-    // FIXME: These two test doesn't work, although it shows correct suggestions.
+    // FIXME: These two tests doesn't work, although it shows correct suggestions.
     expectType<boolean | undefined>(firstFriendBlockedValue?.blocked);
     expectType<Types.ObjectId>(firstFriendBlockedValue?._id);
   });
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -176,9 +176,7 @@ function gh11503() {
   Users.findOne({}).populate<{friends: Friend[]}>('friends').then(user => {
     expectAssignable<Friend>(user?.friends[0]);
     expectType<boolean>(user?.friends[0].blocked);
-    expectType<Types.ObjectId>(user?.friends[0]._id);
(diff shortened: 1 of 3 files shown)


Uzlopak:
This one is better.

types/index.d.ts
```diff
@@ -1813,7 +1813,8 @@ declare module 'mongoose' {
 
   type QueryWithHelpers<ResultType, DocType, THelpers = {}, RawDocType = DocType> = Query<ResultType, DocType, THelpers, RawDocType> & THelpers;
 
-  type UnpackedIntersection<T, U> = T extends (infer V)[] ? (V & U)[] : T & U;
+  type UnpackedIntersection<T, U> = T extends (infer A)[] ? (A & U)[] : keyof U extends never ? T & U: { [P in keyof U]: U[P] extends (infer B)[] ? (B & T)[] : U[P] & T };
```
vkarpov15, line 1816: Shouldn't the last bit be something like `T[P] & U[P]`? I think `T` should come first, and then we should be merging T's keys as well as opposed to merging T into U's keys.

test/types/populate.test.ts
```diff
@@ -149,4 +149,34 @@ function gh11321(): void {
...
+
+  Users.findOne({}).populate('friends').then(user => {
+    expectType<Types.ObjectId | undefined>(user?.friends[0]);
+    expectError(user?.friends[0].blocked);
+    expectError(user?.friends.map(friend => friend.blocked));
+  });
+
+  Users.findOne({}).populate<{friends: Friend[]}>('friends').then(user => {
+    expectAssignable<Friend | undefined>(user?.friends[0]);
+    user?.friends[0].blocked;
```
vkarpov15, line 179: Can we do `expectType<boolean>` here?

mohammad0-0ahmad: I've tried to write two tests for this & it doesn't success, although it shows correct intellisens even in map function.

Untitled

types/index.d.ts
mohammad0-0ahmad: Check the new commit

test/types/populate.test.ts
Uzlopak: So is it now fixed?

types/index.d.ts
```diff
@@ -1813,8 +1813,11 @@ declare module 'mongoose' {
 
   type QueryWithHelpers<ResultType, DocType, THelpers = {}, RawDocType = DocType> = Query<ResultType, DocType, THelpers, RawDocType> & THelpers;
 
-  type UnpackedIntersection<T, U> = T extends (infer V)[] ? (V & U)[] : T & U;
-  type UnpackedIntersectionWithNull<T, U> = T extends null ? UnpackedIntersection<T, U> | null : UnpackedIntersection<T, U>;
+  type UnpackedIntersection<T, U> = T extends (infer A)[]
+    ? (Omit<A, keyof U> & U)[]
+    : keyof U extends never
+    ? T
+  : Omit<T, keyof U> & { [P in keyof U]: U[P] extends (infer B)[] ? (HydratedDocument<B> & B)[] : HydratedDocument<U[P]> & U[P]};
```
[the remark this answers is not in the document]
Uzlopak: Well hmm, actually the thing is, that it is only a HydratedDocument if in populate the option lean is set to false or lean() is used on the query builder. So I suspect this is wrong :/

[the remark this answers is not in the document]
mohammad0-0ahmad: Alright, I will refact that in the next commit :)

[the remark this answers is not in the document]
Uzlopak: @mohammad0-0ahmad 
Imho, if you cant fix that than I would actually recommend to remove those expectTypes, which vkarpov15 requested and revert those HydratedDocument changes, if this is a hard case to solve for you. (Not suggesting that this is over your skills). Probably this is happening because tsd uses a different typescript version and thus resulting in some issues with latest typescript, which is used in vscode for their intellisense. So this is maybe some kind of scope creep. It was not even originally a task for you. So I actually dont want that you feel forced to solve this issue.

But if you think this is a nice challenge and you really want to supply an A-Grade solution, than feel free to invest more time and energy. And know that your work is atleast by me appreciated.

mohammad0-0ahmad:
I've refacted this & it should work with both VSC intellisense & tsd.

types/index.d.ts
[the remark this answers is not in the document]
mohammad0-0ahmad: Thanks for your words, no worries it was funny actually  at the beginning I though that as well but it's works now.

types/index.d.ts
```diff
@@ -2115,8 +2118,8 @@ declare module 'mongoose' {
     polygon(path: string, ...coordinatePairs: number[][]): this;
 
     /** Specifies paths which should be populated with other documents. */
-    populate<Paths = {}>(path: string, select?: string | any, model?: string | Model<any, THelpers>, match?: any): QueryWithHelpers<UnpackedIntersectionWithNull<ResultType, Paths>, DocType, THelpers, RawDocType>;
-    populate<Paths = {}>(options: PopulateOptions | Array<PopulateOptions>): QueryWithHelpers<UnpackedIntersectionWithNull<ResultType, Paths>, DocType, THelpers, RawDocType>;
+    populate<Paths = {}>(path: string, select?: string | any, model?: string | Model<any, THelpers>, match?: any): QueryWithHelpers<UnpackedIntersection<ResultType, Paths>, DocType, THelpers, RawDocType>;
+    populate<Paths = {}>(options: PopulateOptions | Array<PopulateOptions>): QueryWithHelpers<UnpackedIntersection<ResultType, Paths>, DocType, THelpers, RawDocType>;
```
Uzlopak, line 2122: In #11518 we got a regression reported. 

```suggestion
    populate<Paths = {}>(options: string[] | PopulateOptions | Array<PopulateOptions>): QueryWithHelpers<UnpackedIntersection<ResultType, Paths>, DocType, THelpers, RawDocType>;
```

Corresponding typing test would be:


```typescript
async function gh11518() {
  // populate should accept an Array with the fields 
  await Story.findOne().populate(["author"]);
}
```

If you want, we can do this in this PR. If not, I can create a PR. I just didnt wanted to create a PR before this is merged to avoid high probable merge conflict.

Uzlopak: @vkarpov15 Only the second populate type can handle string arrays correct?

mohammad0-0ahmad: OR maybe only the first one?

Uzlopak: lets wait for feedback by vkarpov

vkarpov15: That's an independent issue that I fixed in #11518, no need to discuss that in this PR. But yes, you can call both syntaxes with an array of paths.

types/index.d.ts
```diff
@@ -1813,8 +1813,11 @@ declare module 'mongoose' {
 
   type QueryWithHelpers<ResultType, DocType, THelpers = {}, RawDocType = DocType> = Query<ResultType, DocType, THelpers, RawDocType> & THelpers;
 
-  type UnpackedIntersection<T, U> = T extends (infer V)[] ? (V & U)[] : T & U;
-  type UnpackedIntersectionWithNull<T, U> = T extends null ? UnpackedIntersection<T, U> | null : UnpackedIntersection<T, U>;
```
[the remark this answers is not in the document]
vkarpov15: We introduced `UnpackedIntersectionWithNull` to fix #11041. Although it looks like the test for that issue still passes with this change, so looks like we don't need it after all.

vkarpov15 (approved):
Thanks!

mohammad0-0ahmad:
> Thanks!

I've to resolve conflicts

PR #11504, opened 2022-03-08
SyncIndexes refactor

IslandRhythms:
4 failing tests remain


lib/model.js
```diff
@@ -1438,29 +1440,82 @@ Model.syncIndexes = function syncIndexes(options, callback) {
...
  */
 
-Model.diffIndexes = function diffIndexes(options, callback) {
+Model.diffIndexes = function diffIndexes(callback) {
   const toDrop = [];
   const toCreate = [];
   callback = this.$handleCallbackError(callback);
   return this.db.base._promiseOrCallback(callback, cb => {
     cb = this.$wrapCallback(cb);
     this.listIndexes((err, indexes) => {
```
vkarpov15, line 1517: We're missing some logic here from `cleanIndexes()` that removes non-related indexes: https://github.com/Automattic/mongoose/blob/f5fe027d41d994bb6b11468495aae80232fcb740/lib/model.js#L1536-L1537 .

We should take the logic that determines which indexes to drop, https://github.com/Automattic/mongoose/blob/f5fe027d41d994bb6b11468495aae80232fcb740/lib/model.js#L1536-L1559, and put it behind a separate helper function `function indexesToDrop(dbIndexes, schemaIndexes) { ... }`. `diffIndexes()` and `cleanIndexes()` should both call `indexesToDrop()`.

vkarpov15 (changes requested):

lib/model.js
IslandRhythms: Should I do the same with create?

vkarpov15:
Gonna close this since it has fallen out of date and the work wasn't completed. I'll open a separate PR.

PR #11513, opened 2022-03-10
fix TypeScript declaration file

### LINKED ISSUE #11511: NX Next.JS app fails to build on v6.2.5
*bug*

When I try to build my app, it throws this error:
```
../../node_modules/mongoose/types/cursor.ts:1:1
Type error: Import assignment cannot be used when targeting ECMAScript modules. Consider using 'import * as ns from "mod"', 'import {a} 
from "mod"', 'import d from "mod"', or another module format instead.

> 1 | import stream = require('stream');
    | ^
  2 |
  3 | declare module 'mongoose' {
  4 |   type CursorFlag = 'tailable' | 'oplogReplay' | 'noCursorTimeout' | 'awaitData' | 'partial';
```

Downgrading to v6.2.4 fixes the issue.

{
  "extends": "../../tsconfig.base.json",
  "compilerOptions": {
    "rootDir": ".",
    "sourceMap": true,
    "declaration": false,
    "moduleResolution": "node",
    "esModuleInterop": true,
    "importHelpers": true,
    "target": "es2015",
    "module": "commonjs",
    "lib": ["es2017", "dom", "ES2021.String", "ESNext.String"],
    "typeRoots": ["./@types", "./node_modules/@types"],
    "skipLibCheck": true,
    "skipDefaultLibCheck": true,
    "strictNullChecks": true,
    "baseUrl": ".",
    "jsx": "preserve",
    "allowJs": true,
    "esModuleInterop": true,
    "useDefineForClassFields": true,
    "strict": false,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsxImportSource": "@emotion/react"
  },
  "include": [
    "**/*.ts",
    "**/*.tsx",
    "**/*.js",
    "**/*.jsx",
    "next-env.d.ts",
    "../../libs/services/**/*",
    "../../libs/utils/general-utils/**/*",
    "../../libs/utils/react-utils/**/*"
  ],
  "exclude": ["**/*.spec.tsx", "**/*.spec.ts", "**/*.stories.tsx"]
}

App should build without issue.

### LINKED ISSUE #11512: Does not compile with TypeScript

Bug

Unable to compile due to message `Cannot find type definition file for 'mongoose/types/cursor'.`

Try compiling with any typescript project.

Compiles without errors.

Mongoose: 6.2.5

Seems the types cursor file is of the wrong file extension. Needs to change from `.ts` to `.d.ts`

https://github.com/Automattic/mongoose/blob/master/types/cursor.ts

SimonHausdorf:
**Summary**
Seems like there was a typo in `cursor.ts` file which should be a TypeScript declaration `cursor.d.ts`. It's blocking builds and tests for some people.

closes #11511
closes #11512


sirily11:
did u try this fix with next js build?

m2vi (approved):

Uzlopak:
Ah nice didn't see this PR. 

I am currently thinking how to avoid such a petty mistake for the future. Maybe adding a .gitignore in types folder and whitelisting only .d.ts files. Would be platform independent and you could not check in the file if it was a normal .ts file.

Anyway this PR lgtm!

Uzlopak:
I added in #11516 logic to avoid again this type of regression.

vkarpov15 (approved):
Thanks, and sorry for the trouble!

PR #11516, opened 2022-03-10
chore: add a .gitignore in types folder to not allow pushing simple .ts-files, add check-types-filename.js

Uzlopak:
This is a PR following #11513 
With this platform-independent solution, it is not possible to check-in simple .ts files in the types folder. It still can be forced with git add -f ./types/example.ts, but it is a "cheap" solution.
To prevent a force add, I added a check-types-extensions.js, which checks if we did not accidently added a simple .ts file. 

This should effectively prevent again a regression like it happened in 6.2.5.

Till #11513 is merged, this should correctly fail. After #11513 is merged, it should pass.


vkarpov15:
I'm just surprised, why did tsd not catch this?

Uzlopak:
I think when tsd uses typescript under the hood it can resolve .d.ts and .ts files. But if we import typings in a typescript project, typescript goes another route and then it can only find .d.ts files.

added also a check that the filename should not contain uppercase characters to avoid also a bug like #11468

vkarpov15 (commented):
I'm not thrilled by this solution because I think it is disconnected from what we're actually trying to test. What we're trying to test here is that Mongoose compiles when used with TypeScript. This additional check catches one case that could cause an error, but doesn't catch others.

I'm leaning toward adding a minimal version of TypeScript compilation tests similar to what we removed when we switched to tsd. That approach can have some advantages:

1. Make sure we don't accidentally ship something that doesn't compile
2. Let us test that Mongoose at least compiles against different versions of TypeScript

What do you think @Uzlopak ?

vkarpov15 (approved):
I thought a bit more about this and I'll just err on the side of merging this for now. This PR does the job of preventing shipping code with broken TS references and isn't too cumbersome.

PR #11520, opened 2022-03-12
Extract mongooseoptions

Uzlopak:
I extracted mongooseoptions from index.d.ts

First commit is the cut and paste into the new file. From second commit (and after) are changes regarding the typings. Currently no change of the typings, but only documentation of the jsdoc.

### DIFF
--- a/types/mongooseoptions.d.ts
+++ b/types/mongooseoptions.d.ts
@@ -3,35 +3,59 @@ import stream = require('stream');
 declare module 'mongoose' {
 
   interface MongooseOptions {
-    /** true by default. Set to false to skip applying global plugins to child schemas */
+    /**
+     * Set to false to skip applying global plugins to child schemas.
+     *
+     * @default true
+     */
     applyPluginsToChildSchemas?: boolean;
 
     /**
-     * false by default. Set to true to apply global plugins to discriminator schemas.
+     * Set to true to apply global plugins to discriminator schemas.
      * This typically isn't necessary because plugins are applied to the base schema and
      * discriminators copy all middleware, methods, statics, and properties from the base schema.
+     *
+     * @default false
      */
     applyPluginsToDiscriminators?: boolean;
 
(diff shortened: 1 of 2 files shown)


types/mongooseoptions.d.ts
```diff
@@ -0,0 +1,178 @@
...
+    runValidators?: boolean;
+
+    /**
+     * Sanitizes query filters against [query selector injection attacks](
+     * https://thecodebarbarian.com/2014/09/04/defending-against-query-selector-injection-attacks.html
+     * ) by wrapping any nested objects that have a property whose name starts with $ in a $eq.
+     */
+    sanitizeFilter?: boolean;
+
+    sanitizeProjection?: boolean;
```
Uzlopak, line 121: There was no information for sanitizeProjection regarding default value or what it does in the documentation.

types/mongooseoptions.d.ts
```diff
@@ -0,0 +1,178 @@
...
+     * ) for all validators by default.
+     */
+    runValidators?: boolean;
+
+    /**
+     * Sanitizes query filters against [query selector injection attacks](
+     * https://thecodebarbarian.com/2014/09/04/defending-against-query-selector-injection-attacks.html
+     * ) by wrapping any nested objects that have a property whose name starts with $ in a $eq.
+     */
+    sanitizeFilter?: boolean;
```
Uzlopak, line 119: it is not documented if sanitizeFilter is by default true or false

vkarpov15: false by default

types/mongooseoptions.d.ts
vkarpov15: false by default. There's some more info on sanitizeProjection here: https://thecodebarbarian.com/whats-new-in-mongoose-5-13-sanitizeprojection.html . Should add that to the docs.

PR #11524, opened 2022-03-14
fix: Model.findOneAndReplace docs

anatolykopyl:
**Summary**
The `Model.findOneAndReplace` examples had the wrong params passed to them.
`conditions` instead of `filter` and `replacement`.

**Example**
```javascript
A.findOneAndReplace(conditions)
```

When in reality it should be:
```javascript
A.findOneAndReplace(filter, replacement)
```


AbdelrahmanHafez (approved):
Welcome @anatolykopyl 👋
LGTM, thanks 👍

Issue #11515, opened 2022-03-10
Pluralize issue: 'vertex' -> 'vertexes'


justinlottes:
**Do you want to request a *feature* or report a *bug*?**
BUG

**What is the current behavior?**
pluralize('vertex') === 'vertexes'

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
pluralize('vertex') === 'vertices'

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Uzlopak:
Relevant code:
https://github.com/Automattic/mongoose/blob/f148e629ae2c110c3797902db66f240cd4ddc771/lib/helpers/pluralize.js#L25

I think the regex is wrong. Instead of `/(matr|vert|ind)ix|ex$/gi` should be `/(matr|vert|ind)(ix|ex)$/gi` and then probably a line higher as the above line handles words ending on x. 

Tests should be something like

six => sixes
sex => sexes
vertex => vertices
Index => indizes

But I want to remind that changing this behavior can have catastrophic effects in production. It means that mongoose could try to check if a collection exists and then because the pluralization was changed does not find the production collection and then creates a new collection and creates new documents there.

I would rather recommend to manually set the collectionName and if we think pluralization should be fixed do it in mongoose v7.

vkarpov15:
We'll make this change for Mongoose 7. But, in the meantime, just do `mongoose.model('Vertex', vertexSchema, 'vertices')` to override pluralization for this case

vkarpov15:
After further review, I think we should leave the plural of "vertex" as "vertexes" to be consistent with the discussion from this issue: https://github.com/Automattic/mongoose/pull/11953 . TLDR; we argued that "indexes" is an acceptable plural form of "index" because major dictionaries like Oxford and Merriam-Webster agree. "vertexes" is sufficiently similar, and major dictionaries do list "vertexes" as an acceptable plural form of "vertex".

I'm going to close this as a won't fix. If you prefer `vertices`, changing it is a one-liner `mongoose.model('Vertex', vertexSchema, 'vertices')`.

PR #11529, opened 2022-03-15
Address gh-11515, improve in-practice runtime to insert clean boolean data, and improve source code readability

GNU-Linuxer:

**Summary**

Solves gh-11515 (https://github.com/Automattic/mongoose/issues/11515) pluralization concern by rebasing from pluralize.js module (MIT license, [commit](https://github.com/plurals/pluralize/blob/36f03cd2d573fa6d23e12e1529fa4627e2af74b4/pluralize.js)), along with test suite.

Improved in-practice runtime to insert clean boolean data by skipping Set creations if the input is a boolean-typed value.

Extracted similar operations in `Schema._clone()` function to declutter the code.

`test/model.create.test.js`: renamed variables

`lib/helpers/clone.js`: removed confusing comment

`lib/schema.js`: fixed variable name and changed function name for better readability

**Examples**

Added tests for the improved pluralization (gh-11515) code.

Added tests that validates casting values to boolean.

### DIFF
--- a/lib/error/serverSelection.js
+++ b/lib/error/serverSelection.js
@@ -12,13 +12,13 @@ const isSSLError = require('../helpers/topology/isSSLError');
 /*!
  * ignore
  */
-
-const atlasMessage = 'Could not connect to any servers in your MongoDB Atlas cluster. ' +
+// David Xie: making variable names more self-evident, and change the MongoDB Atlas URL due to a 30* redirect on old URL
+const atlasErrMsg = 'Could not connect to any servers in your MongoDB Atlas cluster. ' +
   'One common reason is that you\'re trying to access the database from ' +
   'an IP that isn\'t whitelisted. Make sure your current IP address is on your Atlas ' +
-  'cluster\'s IP whitelist: https://docs.atlas.mongodb.com/security-whitelist/';
+  'cluster\'s IP whitelist: https://docs.atlas.mongodb.com/security/ip-access-list/';
 
-const sslMessage = 'Mongoose is connecting with SSL enabled, but the server is ' +
+const sslErrMsg = 'Mongoose is connecting with SSL enabled, but the server is ' +
   'not accepting SSL connections. Please ensure that the MongoDB server you are ' +
   'connecting to is configured to accept SSL connections. Learn more: ' +
   'https://mongoosejs.com/docs/tutorials/ssl.html';
@@ -34,13 +34,13 @@ class MongooseServerSelectionError extends MongooseError {
     // Special message for a case that is likely due to IP whitelisting issues.
     const isAtlasWhitelistError = isAtlas(reason) &&
       allServersUnknown(reason) &&
-      err.message.indexOf('bad auth') === -1 &&
-      err.message.indexOf('Authentication failed') === -1;
+      !err.message.includes('bad auth') && // David Xie: using String.includes() is more self-evident (compared to String.indexOf() === -1)
+      !err.message.includes('Authentication failed');
 
     if (isAtlasWhitelistError) {
-      this.message = atlasMessage;
+      this.message = atlasErrMsg;
     } else if (isSSLError(reason)) {
-      this.message = sslMessage;
+      this.message = sslErrMsg;
     } else {
       this.message = err.message;
     }
--- a/lib/helpers/clone.js
+++ b/lib/helpers/clone.js
@@ -112,10 +112,6 @@ function clone(obj, options, isArrayChild) {
 }
 module.exports = clone;
 
-/*!
- * ignore
- */
-
 function cloneObject(obj, options, isArrayChild) {
   const minimize = options && options.minimize;
   const ret = {};
--- a/lib/helpers/pluralize.js
+++ b/lib/helpers/pluralize.js
@@ -6,31 +6,36 @@ module.exports = pluralize;
  * Pluralization rules.
  */
 
+// David Xie: rebase pluralization rules from pluralize.js committed on Oct 6, 2021
+// License of pluralize.js: MIT License
+// Commit: https://github.com/plurals/pluralize/blob/36f03cd2d573fa6d23e12e1529fa4627e2af74b4/pluralize.js
+
 exports.pluralization = [
-  [/(m)an$/gi, '$1en'],
-  [/(pe)rson$/gi, '$1ople'],
-  [/(child)$/gi, '$1ren'],
-  [/^(ox)$/gi, '$1en'],
-  [/(ax|test)is$/gi, '$1es'],
-  [/(octop|vir)us$/gi, '$1i'],
-  [/(alias|status)$/gi, '$1es'],
-  [/(bu)s$/gi, '$1ses'],
-  [/(buffal|tomat|potat)o$/gi, '$1oes'],
-  [/([ti])um$/gi, '$1a'],
-  [/sis$/gi, 'ses'],
-  [/(?:([^f])fe|([lr])f)$/gi, '$1$2ves'],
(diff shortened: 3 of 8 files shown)


lib/helpers/pluralize.js
```diff
@@ -6,31 +6,36 @@ module.exports = pluralize;
  * Pluralization rules.
  */
 
+// David Xie: rebase pluralization rules from pluralize.js committed on Oct 6, 2021
```
AbdelrahmanHafez, line 9: Please remove the `// David Xie: ...` comments that mention the commits, they don't serve much value.

lib/helpers/clone.js
```diff
@@ -112,10 +112,6 @@ function clone(obj, options, isArrayChild) {
 }
 module.exports = clone;
 
-/*!
```
AbdelrahmanHafez, line 115: This is used to avoid automatically generating docs for `cloneObject` since it's meant to be used internally and not part of the public API.

lib/schema.js
```diff
@@ -305,73 +305,88 @@ Schema.prototype.paths;
...
- *
  * @return {Schema} the cloned schema
  * @api public
  * @memberOf Schema
  * @instance
  */
 
-Schema.prototype.clone = function() {
-  const s = this._clone();
+Schema.prototype.clone = function schemaDeepCopy() {
```
AbdelrahmanHafez, line 314: Using different terminology(`schemaDeepCopy` vs `schema.clone`) can be a bit confusing in the long run.
Perhaps `function cloneSchema() {...}` would be more consistent?

lib/schema.js
```diff
@@ -305,73 +305,88 @@ Schema.prototype.paths;
...
+    s[element] = utils.clone(this[element]); // Example: s.options = utils.clone(this.options);
+  });
+
+  const arrayDeepCopy = ['callQueue', 'plugins', 'mapPaths'];
+  // ECMAScript2018 way to deep copy an array, using spread operator
+  arrayDeepCopy.forEach((element) => {
+    s[element] = [...this[element]]; // Example: s.callQueue = [...this.callQueue];
+  });
+
+  // s.base = this.base;
```
AbdelrahmanHafez, line 351: [Please don't commit commented out code](https://kentcdodds.com/blog/please-dont-commit-commented-out-code).

test/model.create.test.js
```diff
@@ -49,31 +49,31 @@ describe('model', function() {
...
 
       assert.ok(Array.isArray(docs));
       assert.equal(docs.length, 1);
     });
 
     it('returns a promise', function() {
-      const p = B.create({ title: 'returns promise' });
-      assert.ok(p instanceof mongoose.Promise);
+      const returnPromise = testModel.create({ title: 'returns promise' });
+      assert.ok(returnPromise instanceof mongoose.Promise);
```
AbdelrahmanHafez, line 76: I would expect a variable with a verb to be a function, perhaps `promise`/`value`?

test/helpers/pluralize.test.js
```diff
@@ -0,0 +1,21 @@
...
+// sex => sexes
+// vertex => vertices
+// Index => indices (David Xie's fix)
+describe('pluralize.js testings', function() {
+  it('Pluralize words', function(done) {
+    const original = ['six', 'sex', 'vertex', 'Index'];
+    const plurals = original.map((element) => pluralize(element));
+    const expected = ['sixes', 'sexes', 'vertices', 'indices'];
+    assert.deepStrictEqual(plurals, expected);
+    done();
```
AbdelrahmanHafez, line 19: We don't need to have a `done();` here anymore, since mocha now runs synchronous tests to the end and will throw an error if the assertion fails.

lib/error/serverSelection.js
```diff
@@ -34,13 +34,13 @@ class MongooseServerSelectionError extends MongooseError {
     // Special message for a case that is likely due to IP whitelisting issues.
     const isAtlasWhitelistError = isAtlas(reason) &&
       allServersUnknown(reason) &&
-      err.message.indexOf('bad auth') === -1 &&
-      err.message.indexOf('Authentication failed') === -1;
+      !err.message.includes('bad auth') && // David Xie: using String.includes() is more self-evident (compared to String.indexOf() === -1)
```
AbdelrahmanHafez, line 37: I agree that `includes` provides better readability, but after we merge this, the comment won't mean much in the new context.

Nitpick: I think we can make it even more explicit to `err.message.includes('bad auth') === false` over `!err.message.includes('bad auth')` because it's much easier to not notice the `!` than the ` === false`, what do you think?

AbdelrahmanHafez:
Welcome, @GNU-Linuxer 👋

I've posted a few comments, also some tests are failing due to the proposed changes, can you fix those?
Overall, those feel like heavy changes and potentially breaking. I'd rather ship those on v7.0 maybe.

What do you think? @GNU-Linuxer @vkarpov15

lib/helpers/pluralize.js
```diff
@@ -6,31 +6,36 @@ module.exports = pluralize;
...
-  [/(matr|vert|ind)ix|ex$/gi, '$1ices'],
-  [/([m|l])ouse$/gi, '$1ice'],
-  [/(kn|w|l)ife$/gi, '$1ives'],
-  [/(quiz)$/gi, '$1zes'],
-  [/^goose$/i, 'geese'],
-  [/s$/gi, 's'],
-  [/([^a-z])$/, '$1'],
-  [/$/gi, 's']
-];
+  [/s?$/i, 's'],
```
vkarpov15, line 14: We unfortunately cannot merge this change until Mongoose 7, because of all of the backwards breaking changes here.

lib/helpers/pluralize.js
```diff
@@ -6,31 +6,36 @@ module.exports = pluralize;
  * Pluralization rules.
  */
 
+// David Xie: rebase pluralization rules from pluralize.js committed on Oct 6, 2021
+// License of pluralize.js: MIT License
```
vkarpov15, line 10: Given that Mongoose is Apache 2 licensed, based on my understanding we can't have part of the code being MIT... Is there any way we can simply just use a fixed version of pluralize instead?

lib/schema.js
vkarpov15: I agree, this change doesn't seem necessary or helpful. Maybe `schemaClone()`, but I'd prefer to make this `function clone()`

lib/schema.js
```diff
@@ -305,73 +305,88 @@ Schema.prototype.paths;
 Schema.prototype.tree;
 
 /**
- * Returns a deep copy of the schema
- *
- * ####Example:
```
vkarpov15, line 310: Why is this removed? We still want to keep this content for the docs.

test/model.create.test.js
```diff
@@ -150,15 +150,15 @@ describe('model', function() {
       });
 
       it('and should reject promise on error', async function() {
-        const p = B.create({ title: 'optional callback 4' });
-        const doc = await p;
+        const callback = testModel.create({ title: 'optional callback 4' });
```
vkarpov15, line 153: `callback` refers to a value of type promise here, that's confusing

vkarpov15 (changes requested):
In the future, can you please split up related changes into separate PRs? This PR is a hodge podge of unrelated changes; some very minor, some significant, and some heavily backwards breaking. I'd love to merge some of these changes for immediate release, but with the pluralize changes and the unexpected removal of `Schema.prototype.clone()` docs there's no way for us to easily do so.

vkarpov15:
I'm going to close this PR. Not really much we can do with this. Feel free to re-open in smaller PRs with similar changes grouped together.

PR #11544, opened 2022-03-19
Add strictPopulate type to PopulateOptions

Kaljuk:
Added strictPopulate to PopulateOptions, otherwise given option can't be used in TS projects when enforcing types.

**Summary**

When using populate in mongoose 6, the types for populate options might've been overlooked. ```PopulateOptions``` interface is missing `strictPopulate` prop, so I added one, with description from the doc. Otherwise I can't use populate with given option.

**Examples**

```
await DocumentModel.findById('some_id')
  .populate({
    match: { somePath: { $ne: undefined } } 
    path: 'somePath',
    strictPopulate: false
  })
```
Previously this was not allowed in typescript dev environment due to enforcing of types, and a type error was thrown.
Now I added the respective type and all works.

***For future reference***
It seems that the `options?: any` seems to be unnecessary and wrongfully referring to _`options.options`_. But when we look at the current codebase with `strictPopulate` for example, then the value will be taken from `options.strictPopulate`, not _`options.options.strictPopulate`_. 

TL;DR I'd suggest removing the `options?: any` from the `PopulateOptions` entirely and replace it with the specific options (i.e. strictPopulate, like in this PR)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -709,6 +709,8 @@ declare module 'mongoose' {
     model?: string | Model<any>;
     /** optional query options like sort, limit, etc */
     options?: any;
+    /** optional boolean, set to `false` to allow populating paths that aren't in the schema */
+    strictPopulate?: boolean;
     /** deep populate */
     populate?: string | PopulateOptions | (string | PopulateOptions)[];
     /**


types/index.d.ts
```diff
@@ -709,6 +709,8 @@ declare module 'mongoose' {
     model?: string | Model<any>;
     /** optional query options like sort, limit, etc */
     options?: any;
+    /** optional boolean, set to `false` to allow populating paths that aren't in the schema */
+    strictPopulate?: boolean;
```
AbdelrahmanHafez, line 713: LGTM, Thanks 👍
Can you add a test case in `test\types\populate.test.ts` to prevent someone from accidentally removing this in the future?

AbdelrahmanHafez (changes requested):

AbdelrahmanHafez:
LGTM.
I agree that having typed options is better than `any`, and I personally think that having them all in one place would be more accurate, either top-level or all in `options`.

Sadly we can't do that change now because it'll be backward breaking, we should at least add accurate typings to represent the API.

You're more than welcome to submit PRs improving the types.

AbdelrahmanHafez:
Welcome, @Kaljuk 👋

Other than the missing test, LGTM.
I agree that having typed options is better than `any`, and I personally think that having them all in one place would be more accurate, either top-level or all in `options`.

Sadly we can't do that change now because it'll be backward breaking, we should at least add accurate typings to represent the API.

You're more than welcome to submit PRs improving the types.

types/index.d.ts
Kaljuk: Sounds good, will do 👍

PR #11549, opened 2022-03-21
fix: missing await for an async function call

Hashen110:

**Summary**

add missing `await` for an async function call


AbdelrahmanHafez (approved):
Welcome @Hashen110 

Good catch, thanks 👍

PR #11551, opened 2022-03-21
fix: continue is unnecessary as the last statement in a loop

Hashen110:

**Summary**

`continue` is unnecessary as the last statement in a loop

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -77,11 +77,8 @@ module.exports = function cast(schema, obj, options, context) {
       if (type === 'function') {
         obj[path] = val.toString();
       }
-
-      continue;
     } else if (path === '$expr') {
       val = cast$expr(val, schema);
-      continue;
     } else if (path === '$elemMatch') {
       val = cast(schema, val, options, context);
     } else if (path === '$text') {
@@ -317,7 +314,6 @@ module.exports = function cast(schema, obj, options, context) {
                     context: context
                   });
                 }
-                continue;
               }
             } else {
               val[$cond] = schematype.castForQueryWrapper({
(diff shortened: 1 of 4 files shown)


lib/cast.js
```diff
@@ -317,7 +314,6 @@ module.exports = function cast(schema, obj, options, context) {
                     context: context
                   });
                 }
-                continue;
```
AbdelrahmanHafez, line 320: This seems to be in a deeply nested condition, not sure if removing it is okay, frankly, I prefer to just keep it for safety.

I'm curoius, how did you spot those unnecessary `continue` statements? Are you using some kind of tool?

AbdelrahmanHafez, line 320: This seems to be in a deeply nested condition, not sure if removing it is okay, frankly, I prefer to just keep it for safety.

I'm curoius, how did you spot those unnecessary `continue` statements? Are you using some kind of tool?

AbdelrahmanHafez: I took a deeper look into this, it should be fine to remove this `continue` as well.
Thanks 👍

vkarpov15 (commented):
I'm ok with the `applyProjection()` and `castFilterPath()` changes, but not the other two. The changes for `cast.js` and `document.js` are a bit risky IMO because those `for` loops have very long and cumbersome `if` statements, which is our fault. Even if it is ok to remove `continue` there, we may add code in the future that will change that, so I'd rather err on the side of caution unless there's some reason (maybe performance) to prefer removing `continue`?

lib/cast.js
Hashen110: IDE (WebStorm) suggested as a warning.

vkarpov15:
I'm going to close this PR for now. Thanks for the contribution, but I don't see enough benefit to risk merging this. Granted that's mostly our fault, we should refactor our code to avoid having such long `for` loops.

PR #11553, opened 2022-03-21
fix(query): prevent sending `exec` to mquery in setOptions

### LINKED ISSUE #11416: Query was always already executed in Node 16.14.0
Report a Bug

**What is the current behaviour?**
While the query executes normally in Node 16.13, the error "Query was already executed" is always present in Node 16.14; by switching Node version with nvm or installing Node 16.14 on an EC2 (Amazon Linux)

**If the current behaviour is a bug, please provide the steps to reproduce.**

The code, while is wrapped by a custom loader, is fairly straightforward:
``` javascript
// simplified version
app.get('/resource', async (req, res, next) => {
  let query = model('Resource').find(req.query);
  if (req.query.page) query = query.limit(10).skip(req.query.page - 1)
  // several lines of query builder here, basically handling pagination and population
  // methods used: orFail, sort, select, populate, limit, skip
  ...
  return res.status(200).json({ records: await query });
})
```

**What is the expected behaviour?**
In Node 16.13, the response reads: 
`{ records: [Object], [Object] }`

In Node 16.14, the response reads, after being caught by error handler:
`Query was already executed: Resource.find({})`

This behaviour persists even with deleting node_modules and reinstalling all packages

**Edit:**
The following variations have been tested and the behaviour persists
``` javascript
await model('Resource').find(req.query) // await the find directly, disabling the query builder altogether 
await model('Resource').find(req.query).exec() // await the find with exec
await query.exec() // await the query with exec

const query = model('Resource').find(req.query);
  // build query with and without reassignment, i.e. query = query.limit(10)
await query; // await the query after building,
```

| |version|remarks
|--|--|--|
|Node|16.14| |
|Mongoose|6.2.2| |
|MongoDB|5.0.6| MongoDB Atlas Free tier (M0) |

AbdelrahmanHafez:
fixes #11416

An alternative solution to removing `exec` only, is to send only the known options to mquery, which according to the documentation of mquery are:
* tailable
* sort
* limit
* skip
* maxScan
* maxTime
* batchSize
* comment
* snapshot
* hint
* slaveOk
* safe


lib/query.js
```diff
@@ -1630,7 +1630,10 @@ Query.prototype.setOptions = function(options, overwrite) {
     }
   }
 
-  return Query.base.setOptions.call(this, options);
+  const optionsToSendToMQuery = Object.assign({}, options);
+  delete optionsToSendToMQuery.exec;
```
Uzlopak, line 1634: delete is a slow operation. Checking if exec is actually set should prevent a potential performance regression
Something like

'exec' in optionsToSendGoMQuery && delete optionsToSendToMQuery.

I prefer the solution of whitelisting valid operators.

AbdelrahmanHafez: Fixed 👍

vkarpov15:
I took a closer look and I think the solution here is to bring the `Query.prototype.setOptions()` code into Mongoose, and add the Mongoose-specific list of options to whitelist. 

In my experience, I think Mongoose should move away from using mquery as much as possible, because mquery is a bit of a messy abstraction for Mongoose. There's too much back-and-forth interplay between mquery and Mongoose. For example, in this case, mquery's list of documented options is correct if you're just using mquery. But Mongoose queries also have helpers for `allowDiskUse`, `writeConcern`, `maxTimeMS`, `collation`, and a couple of others. And mquery will call those methods if those options are set.

I'll close this and put in an alternative PR. But thanks for identifying this issue - this is a tricky problem and finding the exact problem makes the solution a lot easier.

AbdelrahmanHafez:
@vkarpov15 
Sounds good. I think we can merge this with the `exec` exclusion fix to handle #11416
Then we can work on bringing setOptions to mongoose codebase.

What do you think?

vkarpov15:
@AbdelrahmanHafez take a look at the change I just pushed up and see if it makes sense.

@Uzlopak is right that this should be a whitelist rather than a blacklist. Plenty of other options cause errors. For example, try:

```javascript
// Throws 'Error: Options must be an object, got "true"'
const users = await User.find({ name: 'Hafez' }).setOptions({ setOptions: true });
```

AbdelrahmanHafez:
@vkarpov15 LGTM 👍

vkarpov15 (approved):

PR #11554, opened 2022-03-22
feat(base): add option to set `allowDiskUse` globally

### LINKED ISSUE #11478: allowDiskUse configuration globally

feature

right now I need to set `allowDiskUse: true`  per Aggregate or Query

not a bug

I'd like to have something like this mongoose.set('allowDiskUse', true)

and let all the queries and aggregates to use allowDiskUse: true

mongoose 6.2.4
node 16
mongodb 5

AbdelrahmanHafez:
fixes #11478


lib/model.js
```diff
@@ -4214,10 +4214,14 @@ Model.aggregate = function aggregate(pipeline, options, callback) {
 
   const aggregate = new Aggregate(pipeline || []);
   aggregate.model(this);
-
   if (options != null) {
     aggregate.option(options);
   }
+  const defaultAllowDiskUse = this.base.options.allowDiskUse || false;
+  const operationAllowDiskUse = options && options.allowDiskUse;
+  if (operationAllowDiskUse == null && defaultAllowDiskUse != null) {
```
Uzlopak, line 4222: Can defaultAllowDiskUse ever be undefined or null? In line 4220 it will always be set to false if allowDiskUse is null or undefined. So I would say the second condition is always truthy. So we should maybe just set allowDiskUse if we didn't set allowDiskUse on the operation and allowDiskUse is set globally.

```suggestion
  if (operationAllowDiskUse == null && defaultAllowDiskUse === true) {
```

Or simpler

```suggestion
  if (operationAllowDiskUse == null && defaultAllowDiskUse) {
```

AbdelrahmanHafez: Good catch, I changed the logic a little bit to assign the default value as long as `operationAllowDiskUse == null`.
Ther reason I'm trying to avoid `defaultAllowDiskUse === true` is because using whatever the user sends to the driver is a good idea, for example, if a user sends `'true'`/`1`/`0`, we'd wanna send it to the driver instead of not sending anything at all.

lib/model.js
```diff
@@ -4214,10 +4214,14 @@ Model.aggregate = function aggregate(pipeline, options, callback) {
 
   const aggregate = new Aggregate(pipeline || []);
   aggregate.model(this);
-
   if (options != null) {
     aggregate.option(options);
   }
+  const defaultAllowDiskUse = this.base.options.allowDiskUse != null ? this.base.options.allowDiskUse : false;
```
vkarpov15, line 4220: For consistency, I'd prefer if we implemented this the same way we implemented global `maxTimeMS` using the [`applyGlobalMaxTimeMS()` helper](https://github.com/Automattic/mongoose/blob/80245edf5aeab304411681035f7f4153dbd2b692/lib/helpers/query/applyGlobalMaxTimeMS.js). Two reasons: 1) consistency in not applying default options until right before execution, and 2) `allowDiskUse` is also an option for queries, so we need to make global `applyDiskUse` apply to both aggregations and queries.

vkarpov15 (changes requested):

lib/model.js
AbdelrahmanHafez: Done, I initially wanted to use the option right before execution, but that made testing a little bit more challenging.
I used Sinon to add a spy on the native driver `aggregate `method to get the options that end up there. It will help with some other tests as well.

lib/query.js
```diff
@@ -11,7 +11,7 @@ const MongooseError = require('./error/mongooseError');
 const ObjectParameterError = require('./error/objectParameter');
 const QueryCursor = require('./cursor/QueryCursor');
 const ReadPreference = require('./driver').get().ReadPreference;
-const applyGlobalMaxTimeMS = require('./helpers/query/applyGlobalMaxTimeMS');
+const { applyGlobalMaxTimeMS, applyGlobalDiskUse } = require('./helpers/query/applyGlobalOption');
```
Uzlopak, line 14: Is Object destructuring ok? 
Unit tests seem to be fine, but what about Browser Support?

test/index.test.js
```diff
@@ -911,5 +913,73 @@ describe('mongoose module:', function() {
...
+        m.set('allowDiskUse', true);
+
+        const db = await m.connect(start.uri);
+
+        const userSchema = new m.Schema({
+          name: String
+        });
+
+        const User = db.model('User', userSchema);
+        const nativeAggregateSpy = sinon.spy(User.collection, 'aggregate');
```
Uzlopak, line 953: Please try to avoid this. It is good to not restore the spy in the it-Block as a failing test could mean that a spy is not restored. So restoring in an after block is recommended.

Do in the describe-block after line 916 a let nativeAggregateSpy and assign a value to it in this line (=> remove the const here)

In the after block you do just the 
nativeAggregateSpy && nativeAggregateSpy.restore()

I had cases where I used a global spy or wanted to introduce a global spy and was massively blocked because there was general sinon.restore() used.

lib/query.js
AbdelrahmanHafez: Should be fine, we're using Webpack which should handle that.

test/index.test.js
AbdelrahmanHafez: I want to restore everything to its original state after each test.
I learned the hard way that trying to control things globally introduces more problems that are harder to fix in the long run.

You can have the best of both worlds by grouping up your global fakes into a function(s) and then explicitly calling them at the beginning of the `it` block. That way we won't have to jump back and forth in the codebase trying to figure out what caused some specific behavior.

lib/query.js
vkarpov15: Yeah destructuring should be fine here

vkarpov15 (approved):
Nice :+1: I'll merge this into 6.3 branch

PR #11565, opened 2022-03-24
Incorrect Example Syntax - Dynamic References via `refPath`

chandiwalaaadhar:
"on:" is a reserved keyword which fails on execution, therefore modelId: would be the correct key and works well

**Summary**
This PR will make the documentation example codes bug free as the current syntax gave error ```Error: `on` may not be used as a schema pathname```

**Examples**

NA

### DIFF
--- a/docs/populate.md
+++ b/docs/populate.md
@@ -491,7 +491,7 @@ storing comments. A user may comment on either a blog post or a product.
 ```javascript
 const commentSchema = new Schema({
   body: { type: String, required: true },
-  on: {
+  modelId: {
     type: Schema.Types.ObjectId,
     required: true,
     // Instead of a hardcoded model name in `ref`, `refPath` means Mongoose


Uzlopak:
Hmm. I would actually think that 'on' being a "reserved keyword" should be classified as a bug.

AbdelrahmanHafez:
Reserved keywords should be removed from #10414, if you can't use `on`, that's something we should fix.

That being said, while we allow usage of keywords that were previously reserved, we recommend against it because it will break plugins that rely on using them, so the example should not mention a case where it might cause issues for the end-user.

Uzlopak:
I think the problem is, that Schema/Model is an EventEmitter. So while you can set an attribute named "on", it collides with the on property of the EventEmitter.

AbdelrahmanHafez:
@Uzlopak 
That's correct. We've come up with a workaround in https://github.com/Automattic/mongoose/issues/9010, which is giving an alias `$on` and using that alias internally, so users can use `on` if they want to.

Uzlopak:
Should we open a ticket to examine this issue?

AbdelrahmanHafez:
@Uzlopak Done.
https://github.com/Automattic/mongoose/issues/11580

PR #11579, opened 2022-03-25
chore(deps-dev): upgrade highlight.js from 9.18.3 to 11.5.0

AbdelrahmanHafez:
This PR fixes the breaking changes from upgrading to [v10](https://github.com/highlightjs/highlight.js/blob/main/VERSION_10_UPGRADE.md) and [v11](https://github.com/highlightjs/highlight.js/blob/main/VERSION_11_UPGRADE.md).

The only downside is IE11 support being dropped in v10 which I think is fine.

I also tested the rendering manually on my machine, and the docs highlighting is identical to the one in production.


vkarpov15 (approved):
Thanks! I'll manually go through the code highlighting locally as well just to be sure.

PR #11581, opened 2022-03-25
chore(eslint): include tools, website.js and benchmarks to eslint config

AbdelrahmanHafez:
This PR includes the following files/folders to be automatically linted. Is there a reason to not want to lint those?
* website.js
* benchmarks
* tools


vkarpov15 (approved):
No particular reason. I think there might have been some linter issues at some point that we didn't want to fix.

PR #11588, opened 2022-03-27
Bug/types/update query

AbdelrahmanHafez:
Currently, we can do the following in JavaScript:
```ts
const userSchema = new Schema({ name: String });

await User.updateOne({}, { name: 'Sam' });
```

But TypeScript doesn't know that the second parameter to `updateOne` can take `name` as a top level property.
It would, however, understand the types if we use `$set`:
```ts
// provides inteliisense support
await User.updateOne({}, { $set: { name: 'Sam' } });

// does not provide intellisense support
await User.updateOne({}, { name: 'Sam' });
```

I managed to make it read the properties from the interface, but another bug seems to be that it converts all types to `any` instead of the type defined on the schema.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2058,7 +2058,7 @@ declare module 'mongoose' {
    * { age: 30 }
    * ```
    */
-  export type UpdateQuery<T> = _UpdateQuery<_UpdateQueryDef<T>>['$set'] & _UpdateQuery<_UpdateQueryDef<T>> & AnyObject;
+  export type UpdateQuery<T> = Partial<T> & _UpdateQuery<_UpdateQueryDef<T>> & AnyObject;
 
   export type DocumentDefinition<T> = {
     [K in keyof Omit<T, Exclude<keyof Document, '_id' | 'id' | '__v'>>]:
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2058,7 +2058,7 @@ declare module 'mongoose' {
    * { age: 30 }
    * ```
    */
-  export type UpdateQuery<T> = _UpdateQuery<_UpdateQueryDef<T>> & AnyObject;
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -2058,7 +2058,7 @@ declare module 'mongoose' {
    * { age: 30 }
    * ```
    */
-  export type UpdateQuery<T> = _UpdateQuery<_UpdateQueryDef<T>> & AnyObject;
+  export type UpdateQuery<T> = Partial<T> & _UpdateQuery<_UpdateQueryDef<T>> & AnyObject;
```
Uzlopak, line 2061: I think the reason is the Union Type with AnyObject resulting in a Type Widening. I would usually try with Omit<AnyObject, keyof T & _UpdateQuery<_UpdateQueryDef<T>>> instead of just using AnyObjecy.

```suggestion
  export type UpdateQuery<T> = Partial<T> & _UpdateQuery<_UpdateQueryDef<T>> & Omit<AnyObject, keyof T & _UpdateQuery<_UpdateQueryDef<T>>>;
```

Didn't test it, just wrote it with my mobile.

Also look into the populate typing PR. @mohammad0-0ahmad used a different notation to achieve the correct typings

mohammad0-0ahmad: Or maybe the following, if update query should accept any type:
@vkarpov15  what do u think about that ?
```
export type UpdateQuery<T> = _UpdateQuery<_UpdateQueryDef<T>> & (T | AnyObject);
```

vkarpov15: @mohammad0-0ahmad that approach works for me, as long as it is `& (Partial<T> | AnyObject)`

vkarpov15 (approved):
Thanks. Overall, I think this approach works and is a neat improvement for intellisense. I'll try out @mohammad0-0ahmad 's suggestion after merging.

vkarpov15:
Actually, I didn't notice that this is causing test failures, sorry. I'm going to have to revert and we'll have to consider how we can do this without breaking our tests.

Issue #8934, opened 2020-05-04
Extract __v field natively when lean() used


meness:
**Do you want to request a *feature* or report a *bug*?** Feature

**What is the current behavior?**
`lean()` returns `__v` field.

**What is the expected behavior?**
`lean()` is used to prepare the document to be sent to Client-as docs say. I think Client doesn't need that field.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
`MongoDB: 4.2.5`
`Mongoose: 5.9.11`
`Node.js: 12.10.0`

vkarpov15:
That's a good point. We will look into stripping out the version key if you're using `lean()`.

vkarpov15:
I think this is better implemented as a feature:

```javascript
lean({ versionKey: false })
```

jonnyfux:
Any news on this? Or is there a workaround until the feature is realesed?

JeHwanYoo:
> I think this is better implemented as a feature:
> 
> ```js
> lean({ versionKey: false })
> ```

## Suggested Improvement

As you mentioned, it works if written that way, but the type definition of the `lean` query is a bit confusing.

How about adding a type definition for `{ versionKey: boolean }`?

## Type Defintion

https://github.com/Automattic/mongoose/blob/53362c3e2b70efcd3c4783638565fb74918440e2/types/query.d.ts#L541

```ts
    /** Sets the lean option. */
    lean<
      LeanResultType = GetLeanResultType<RawDocType, ResultType, QueryOp>
    >(
      val?: boolean | any
    ): QueryWithHelpers<
      ResultType extends null
        ? LeanResultType | null
        : LeanResultType,
      DocType,
      THelpers,
      RawDocType,
      QueryOp,
      TInstanceMethods
      >;
```

JeHwanYoo:
Are the parameter types `Document.prototype.toObject` and the parameter for `Query.prototype.lean` compatible? 

`ToObjectOptions` includes `{ versionKey: boolean }`.

https://github.com/Automattic/mongoose/blob/53362c3e2b70efcd3c4783638565fb74918440e2/types/index.d.ts#L191

PR #11593, opened 2022-03-28
gh-8934: strip version key when option set to false

IslandRhythms:


lib/query.js
```diff
@@ -2242,6 +2243,11 @@ Query.prototype._find = wrapThunk(function(callback) {
     }
 
     if (!mongooseOptions.populate) {
+      if (mongooseOptions.lean && mongooseOptions.lean.versionKey === false) {
+        docs.forEach((doc) => {
+          delete doc.__v;
```
AbdelrahmanHafez, line 2248: People can set a custom versionKey in their schemas, so we need to cover that scenario as well:
```suggestion
          delete doc.[schema.options.versionKey || '__v'];
```

test/model.querying.test.js
```diff
@@ -2399,6 +2399,19 @@ describe('model: querying:', function() {
       });
     });
 
+    it('removes the __v property if versionKey: false is set (gh-8934)', function(done) {
+      const title = 'Wooooot ' + random();
+
+      const post = new BlogPostB();
+      post.set('title', title);
+      post.save(function(err) {
+        BlogPostB.find({ title: title }).lean({ versionKey: false }).exec(function(err, docs) {
```
AbdelrahmanHafez, line 2408: This can be simplified by using async/await, and `findOne` instead of `find`
```js
const title = 'Wooooot ' + random();
const post = await BlogPostB.create({ title });
const foundPost= await BlogPostB.find({ title }).lean({  versionKey: false });
assert.strictEqual(foundPost.__v, undefined);
```

test/model.querying.test.js
```diff
@@ -2399,6 +2399,19 @@ describe('model: querying:', function() {
       });
     });
 
+    it('removes the __v property if versionKey: false is set (gh-8934)', function(done) {
```
AbdelrahmanHafez, line 2402: We need to add a test-case for the [custom `versionKey`](https://mongoosejs.com/docs/guide.html#versionKey) as well.

AbdelrahmanHafez (changes requested):

lib/query.js
vkarpov15: Better approach is to just do `delete doc[schema.options.versionKey]`.

Also need to add a `&& schema.options.versionKey` check on L2246, because setting `schema.options.versionKey = false` is how you disable versioning for the whole schema.

test/model.querying.test.js
```diff
@@ -2399,6 +2399,19 @@ describe('model: querying:', function() {
...
+      const title = 'Wooooot ' + random();
+
+      const post = new BlogPostB();
+      post.set('title', title);
+      post.save(function(err) {
```
vkarpov15, line 2407: Use async/await rather than callbacks

lib/query.js
```diff
@@ -2204,6 +2204,7 @@ function _castArrayFilters(query) {
  * @api private
  */
 Query.prototype._find = wrapThunk(function(callback) {
+
```
vkarpov15, line 2207: Also need to cover the `findOne()` and `findOneAndUpdate()` case

vkarpov15 (changes requested):

lib/query.js
```diff
@@ -2475,6 +2480,10 @@ Query.prototype._findOne = wrapThunk(function(callback) {
       return null;
     }
 
+    if (mongooseOptions.lean && mongooseOptions.lean.versionKey === false && this.schema.options.versionKey) {
```
vkarpov15, line 2483: Move this logic into the `this.completeOne()` function so it isn't repeated across multiple functions.

lib/query.js
```diff
@@ -3922,6 +3931,10 @@ Query.prototype._findAndModify = function(type, callback) {
     if (err) {
       return callback(err);
     }
+    const mongooseOptions = _this._mongooseOptions;
```
vkarpov15, line 3934: Move this logic into `_completeOne()` as well since it is the same as the logic my other comment refers to

vkarpov15 (changes requested):

lib/query.js
```diff
@@ -2240,8 +2241,12 @@ Query.prototype._find = wrapThunk(function(callback) {
     if (this.options.explain) {
       return callback(null, docs);
     }
-
     if (!mongooseOptions.populate) {
+      if (mongooseOptions.lean && mongooseOptions.lean.versionKey === false && _this.schema.options.versionKey) {
+        docs.forEach((doc) => {
+          delete doc[_this.schema.options.versionKey];
```
Uzlopak, line 2247: delete on an Object is naturally slow. Can we please first check if the key is actually set and if so then delete it?

```suggestion
          if (_this.schema.options.versionKey in doc) 
          { 
              delete doc[_this.schema.options.versionKey];
          }
```

Or better first create a local variable `cont versionKey = _this.schema.options.versionKey;` to improve the perf by avoiding the lookups.

lib/query.js
```diff
@@ -2423,6 +2428,9 @@ Query.prototype._completeOne = function(doc, res, callback) {
   }
 
   if (!mongooseOptions.populate) {
+    if (mongooseOptions.lean && mongooseOptions.lean.versionKey === false && this.schema.options.versionKey) {
+      delete doc[this.schema.options.versionKey];
```
Uzlopak, line 2432: The same as above

test/model.querying.test.js
```diff
@@ -2399,6 +2399,17 @@ describe('model: querying:', function() {
       });
     });
 
+    it('removes the __v property if versionKey: false is set (gh-8934)', async function() {
+      const title = 'Wooooot ' + random();
+      await BlogPostB.create({ title });
+      const foundPost = await BlogPostB.find({ title }).lean({ versionKey: false });
+      assert.strictEqual(foundPost.__v, undefined);
```
Uzlopak, line 2406: I think we should actually test if the attribute exists and not if the value is undefined.
Like
```suggestion
      assert.ok(!('__v' in foundPost));
```

Also the other two assertions should be modified accordingly

lib/query.js
AbdelrahmanHafez: @Uzlopak I see this is a repetitive concern, what do you think of creating a utility function that deletes and checks before deletion? Hopefully, the v8 team will address this slow deletion issue sometime in the future and we won't need such utility.

```js
function deleteFromObject(object, fieldToDelete) { 
  if (fieldToDelete in object) {
    delete object[fieldToDelete]; 
  }
}
```

Uzlopak: I think a deleteFromObject function would result in a bigger deoptimization in v8, as the First Parameter would be Always polymorph and the fieldToDelete would be a byValue Parameter, duplicating the memory usage for the String. Probably would be a heavier Impact than checking If the field exists in the object and making deleting directly faster.

Maybe i should Benchmark this further. So neglect this remark for now

Uzlopak: But we should definitely store versionKey in a temp variable to avoid the lookup.

AbdelrahmanHafez: Honestly, in my head, all synchronous code is free unless it becomes a problem.

So the performance of `delete` is something I wouldn't ever think about unless someone complains about it, there is generally much more room for performance optimization than the synchronous code.

But you do make a good point when it comes to hot paths where optimizing synchronous code has a noticeable effect on performance.

vkarpov15 (approved):

PR #11600, opened 2022-03-30
(Gh 7971) Can define discriminators in the schema

IslandRhythms:


lib/schema.js
```diff
@@ -569,6 +575,12 @@ Schema.prototype.add = function add(obj, prefix) {
           this.nested[prefix.substring(0, prefix.length - 1)] = true;
         }
         this.path(prefix + key, val);
+        if (val.discriminators) {
```
vkarpov15, line 578: There's a case we missed: `val` can be a Schema, and Schemas can have a `discriminators` property. Let's replace this with `if (utils.isPOJO(val) && utils.isPOJO(val.discriminators))`

lib/schema.js
```diff
@@ -540,6 +540,12 @@ Schema.prototype.add = function add(obj, prefix) {
         this.nested[prefix.substring(0, prefix.length - 1)] = true;
       }
       this.path(prefix + key, val);
+      if (Array.isArray(val) && val.length != 0 && val[0].discriminators) {
```
vkarpov15, line 543: There's a case we missed: `val` can be a Schema, and Schemas can have a `discriminators` property. Let's replace this with `if (utils.isPOJO(val[0]) && utils.isPOJO(val[0].discriminators))`

vkarpov15 (changes requested):

lib/schema.js
```diff
@@ -540,6 +540,12 @@ Schema.prototype.add = function add(obj, prefix) {
         this.nested[prefix.substring(0, prefix.length - 1)] = true;
       }
       this.path(prefix + key, val);
+      if (utils.isPOJO(val[0]) && utils.isPOJO(val[0].discriminators)) {
```
AbdelrahmanHafez, line 543:
```suggestion
      if (utils.isPOJO(val[0] && val[0].discriminators)) {
```

AbdelrahmanHafez: @vkarpov15 
Would it be more intention revealing to use
`if ((val[0] instanceof Schema) === false && val[0].discriminators) {`?

vkarpov15: That's fair, we can do `&& !(val[0] instanceof Schema)` instead

PR #11601, opened 2022-03-30
`continueOnError` option for `eachAsync()`

### LINKED ISSUE #6355: Allow eachAsync() to continue execution on error
http://thecodebarbarian.com/cursors-in-mongoose-45#comment-3663648177

vkarpov15:
Fix #6355

### DIFF
--- a/lib/error/eachAsyncMultiError.js
+++ b/lib/error/eachAsyncMultiError.js
@@ -8,10 +8,13 @@ const MongooseError = require('./');
 
 
 /**
- * The connection failed to reconnect and will never successfully reconnect to
- * MongoDB without manual intervention.
+ * If `eachAsync()` is called with `continueOnError: true`, there can be
+ * multiple errors. This error class contains an `errors` property, which
+ * contains an array of all errors that occurred in `eachAsync()`.
+ *
  * @api private
  */
+
 class EachAsyncMultiError extends MongooseError {
   /**
    * @param {String} connectionString
--- a/lib/error/eachAsyncMultiError.js
+++ b/lib/error/eachAsyncMultiError.js
@@ -1,0 +1,38 @@
+/*!
+ * Module dependencies.
+ */
+
+'use strict';
+
+const MongooseError = require('./');
+
+
+/**
+ * The connection failed to reconnect and will never successfully reconnect to
+ * MongoDB without manual intervention.
+ * @api private
+ */
+class EachAsyncMultiError extends MongooseError {
+  /**
+   * @param {String} connectionString
+   */
+  constructor(errors) {
+    let preview = errors.map(e => e.message).join(', ');
+    if (preview.length > 50) {
+      preview = preview.slice(0, 50) + '...';
+    }
+    super(`eachAsync() finished with ${errors.length} errors: ${preview}`);
+
+    this.errors = errors;
+  }
+}
+
+Object.defineProperty(EachAsyncMultiError.prototype, 'name', {
+  value: 'EachAsyncMultiError'
+});
+
+/*!
+ * exports
+ */
+
+module.exports = EachAsyncMultiError;
--- a/lib/helpers/cursor/eachAsync.js
+++ b/lib/helpers/cursor/eachAsync.js
@@ -26,7 +26,7 @@ module.exports = function eachAsync(next, fn, options, callback) {
   const parallel = options.parallel || 1;
   const batchSize = options.batchSize;
   const continueOnError = options.continueOnError;
-  let aggregatedErrors = null;
+  const aggregatedErrors = [];
   const enqueue = asyncQueue();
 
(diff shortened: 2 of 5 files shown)


lib/error/eachAsyncMultiError.js
```diff
@@ -0,0 +1,38 @@
...
+ * Module dependencies.
+ */
+
+'use strict';
+
+const MongooseError = require('./');
+
+
+/**
+ * The connection failed to reconnect and will never successfully reconnect to
```
Uzlopak, line 11: Is the comment correct?

lib/helpers/cursor/eachAsync.js
```diff
@@ -24,10 +25,11 @@ const promiseOrCallback = require('../promiseOrCallback');
 module.exports = function eachAsync(next, fn, options, callback) {
   const parallel = options.parallel || 1;
   const batchSize = options.batchSize;
+  const continueOnError = options.continueOnError;
+  let aggregatedErrors = null;
```
Uzlopak, line 29: The name indicates actually that this is an Array of Errors.

Uzlopak (changes requested):
I didn't tested it but only reading the code I think there is a small bug in there.

The preview in EachAsyncMultiError is only generated in the constructor with the passes Array, but not when we add an Error to the errors attribute of the Error. So I assume the Error will not have the correct Error Message.

I guess we could change the `let aggregatedErrors = null` to `const aggregatedErrors = continue error && []` and push Errors directly in there and in the last Moment when we want to pass the Error Object to the callback we construct the actual EachAsyncMultiError-Object.

types/cursor.d.ts
```diff
@@ -25,10 +31,10 @@ declare module 'mongoose' {
      * will wait for the promise to resolve before iterating on to the next one.
      * Returns a promise that resolves when done.
      */
-    eachAsync(fn: (doc: DocType[]) => any, options: { parallel?: number, batchSize: number }, callback: CallbackWithoutResult): void;
-    eachAsync(fn: (doc: DocType) => any, options: { parallel?: number }, callback: CallbackWithoutResult): void;
-    eachAsync(fn: (doc: DocType[]) => any, options: { parallel?: number, batchSize: number }): Promise<void>;
-    eachAsync(fn: (doc: DocType) => any, options?: { parallel?: number }): Promise<void>;
+    eachAsync(fn: (doc: DocType[]) => any, options: EachAsyncOptions, callback: CallbackWithoutResult): void;
```
Uzlopak, line 34: The typings changes are imho not the same. When batchSize is number then doc will be an Array. With these changes the doc can be array or a single Object.

lib/helpers/cursor/eachAsync.js
```diff
@@ -62,14 +64,19 @@ module.exports = function eachAsync(next, fn, options, callback) {
           return done();
         }
         if (err != null) {
-          error = err;
-          finalCallback(err);
-          return done();
+          if (continueOnError) {
+            aggregatedErrors = aggregatedErrors || new EachAsyncMultiError([]);
```
AbdelrahmanHafez, line 68: `aggregatedErrors` is a bit confusing, it makes me think that it's an array of errors.
Perhaps `eachMultiAsyncError` might be a more descriptive name.

test/helpers/cursor.eachAsync.test.js
```diff
@@ -134,4 +134,61 @@ describe('eachAsync()', function() {
...
+
+    return eachAsync(next, fn, { continueOnError: true }).
+      then(() => { throw new Error('Expected error'); },
+        err => {
+          assert.ok(err);
+          assert.equal(err.name, 'EachAsyncMultiError');
+          assert.equal(err.errors.length, 2);
+          assert.equal(err.errors[0].message, 'Doc number 1');
+          assert.equal(err.errors[1].message, 'Doc number 3');
+        });
```
AbdelrahmanHafez, line 162: I learned from [a useful guide](http://thecodebarbarian.com/async-functions-in-javascript.html#:~:text=Another%20common%20pattern%20is%20to%20use%20Promise%23catch().) a better way to handle errors 😄
```suggestion
    const err = await eachAsync(next, fn, { continueOnError: true }).then(()=> null, err=> err);
    assert.ok(err);
    assert.equal(err.name, 'EachAsyncMultiError');
    assert.equal(err.errors.length, 2);
    assert.equal(err.errors[0].message, 'Doc number 1');
    assert.equal(err.errors[1].message, 'Doc number 3');
```

types/cursor.d.ts
AbdelrahmanHafez: I agree with Uzlopak
Perhaps we need to split them into two different types.

AbdelrahmanHafez (changes requested):
Overall looks good to me, with only a couple of comments.

PR #11602, opened 2022-03-30
types: improve lastErrorObject typing for rawResults

simllll:
**Summary**

This improves the type for lastErrorObject, this helps to use the correct values as there is for example no modifiedCount or other things that a simple updateOne returns.

**Examples**

```ts
// findOneAndUpdate with rawResult: true
const result = Model.findOneAndUpdate(query, { $inc: { occurence: 1 } },  { upsert: true, returnDocument: 'after', rawResult: true });

// should work
updateResult.lastErrorObject?.upserted
// should throw error
updateResult.lastErrorObject?.modifiedCount
```

### DIFF
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -265,3 +265,14 @@ async function gh11306(): Promise<void> {
   expectType<any[]>(await MyModel.distinct('name'));
   expectType<string[]>(await MyModel.distinct<string>('name'));
 }
+
+async function gh11602(): Promise<void> {
+  const updateResult = await Model.findOneAndUpdate(query, { $inc: { occurence: 1 } }, {
+    upsert: true,
+    returnDocument: 'after',
+    rawResult: true
+  });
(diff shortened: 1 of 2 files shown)


Uzlopak:
Hi Simon,

Lgtm. Could you please add your Sample Code to the typings Tests in test/types? I guess models.test.ts is the fitting file.

Maybe also add an error test?

Like 
`expectError(updateResult.lastErrorObject?.invalid);`
?

Greetings from Teheran to Vienna :)

AbdelrahmanHafez (changes requested):
Thanks 👍
I second Uzlopak's comment, would be nice if we can have tests.

simllll:
I've added a test in queries.test.ts as there are also some other similar tests in it already. Let me know if I should move it to models or another file if you think it fits better somewhere else.

Uzlopak (approved):
Lgtm

AbdelrahmanHafez (approved):
LGTM, thanks 👍

Issue #11547, opened 2022-03-20
[Feature] Schema#removeIndex(...)


AbdelrahmanHafez:
I came across a case where I cloned a schema, and would like to reuse it in another model without the indexes.
The issue is that the cloned schema includes its indexes, and there is no official way to remove indexes, I had to access `schema._indexes` which doesn't work with TS.

```js
const cartSchema = new Schema({ customerId: String, products: [productSchema]  });
cartSchema.index({ customerId: 1 }, { unique: true });

const cartSchemaClone = cartSchema.clone();

cartSchemaClone._indexes = []; // there is no other way to do it.

const checkoutSchema = new Schema({ cart: cartSchemaClone });
```

AbdelrahmanHafez:
Turns out `cartSchemaClone._indexes = [];` doesn't really remove all indexes, it only removes the top-level paths indexes and ignores nested paths indexes.

The workaround that I found for removing all indexes:
```js
const cartSchemaClone = cartSchema.clone();
cartSchemaClone.set('excludeIndexes', true); // ignore _all_ indexes
const checkoutSchema = new Schema({ cart: cartSchemaClone });
```

This fixes my issue, but would become a problem once I need to add some indexes. I think we need to provide:
`Schema#clearIndexes` and `Schema#removeIndex`.

vkarpov15:
We should support both schema.removeIndex('my-index-name') and schema.removeIndex({ status: 1 }) . The latter should remove any index that is deep equal to the given spec.

AbdelrahmanHafez:
I agree that removing an index from schema shouldn't touch MongoDB.

For deep equality check, I think we should check both the index definition and the options. Also, deep equality check should treat objects with the same properties in different order as equal.

For string index names, I am not sure if they're unique or not, one other concern, AFAIK people can set a custom index name, if we wanna match with the index name, does that mean we'll need to fetch indexes from the database? Needing to fetch indexes means we'll have to tie the index removal with database connections.

I think supporting string index names, knowing that users can set custom indexes names, will introduce more issues than the feature requires, deep equality check should do the trick.

PR #11606, opened 2022-03-31
gh-11547 Can remove and clear indexes on the schema

IslandRhythms:


lib/schema.js
```diff
@@ -579,6 +580,23 @@ Schema.prototype.add = function add(obj, prefix) {
...
+    for (let j = 0; j < this._indexes[i].length; j++) {
+      if (Object.keys(this._indexes[i][j]) == index || util.isDeepStrictEqual(this._indexes[i][j], index)) {
+        this._indexes.splice(i, 1);
+      }
+    }
+  }
+}
+
+Schema.prototype.clearIndexes = function clearIndexes() {
+  return this._indexes.length = 0;
```
vkarpov15, line 597: No need to `return` here.

test/index.test.js
```diff
@@ -911,5 +912,32 @@ describe('mongoose module:', function() {
       const m = new mongoose.Mongoose();
       assert.deepEqual(await m.syncIndexes(), {});
     });
+    it('Allows for the removal of indexes via string or object (gh-11547)', async function() {
+      const m = new mongoose.Mongoose();
+
+      const db = await m.connect('mongodb://localhost:27017');
```
vkarpov15, line 918: No need to create a new `Mongoose` instance or do a separate `connect()` here.

lib/schema.js
```diff
@@ -579,6 +580,23 @@ Schema.prototype.add = function add(obj, prefix) {
   return this;
 };
 
+Schema.prototype.removeIndex = function removeIndex(index) {
+  if (arguments.length > 1 || Object.keys(arguments[0]).length > 1 && typeof arguments[0] !== 'string') {
```
vkarpov15, line 584: Not sure why there's `Object.keys(arguments[0]).length > 1` here, you should be able to call `removeIndex({ name: 1, age: 1 })` to remove a compound index on `{ name, age }`

lib/schema.js
```diff
@@ -579,6 +580,23 @@ Schema.prototype.add = function add(obj, prefix) {
...
+    throw new Error('removeIndex() takes only 1 parameter');
+  }
+  for (let i = 0; i < this._indexes.length; i++) {
+    for (let j = 0; j < this._indexes[i].length; j++) {
```
vkarpov15, line 588: You don't need to loop through the index itself. You just need to check if `this._indexes[i][0]` is deep equal to the `index` argument, if the `index` argument is an object

vkarpov15 (changes requested):

AbdelrahmanHafez:
@IslandRhythms 
I noticed that you're facing linting issues in many of the submitted PRs.
I highly recommend installing the [ESLint extension in VSCode](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) or your IDE of choice, it'll significantly increase your productivity, and you'll find out/automatically fix linting errors immediately.

test/index.test.js
```diff
@@ -911,5 +912,32 @@ describe('mongoose module:', function() {
       const m = new mongoose.Mongoose();
       assert.deepEqual(await m.syncIndexes(), {});
     });
+    it('Allows for the removal of indexes via string or object (gh-11547)', async function() {
```
AbdelrahmanHafez, line 915: @IslandRhythms 
We need to cover the case where a schema is using another schema with indexes.
For example
```js
const productSchema = new Schema({ name: String });
productSchema.index({ name: 1 });

const orderSchema = new Schema({
  products: [productSchema]
});
```

vkarpov15: At first glance, I'm opposed to having `orderSchema.clearIndexes()` clear all indexes on child schemas, at least by default. This can cause surprises if you use `productSchema` in multiple places.

vkarpov15: We currently don't clone child schemas. In your example, `orderSchema` contains a reference to `productSchema`, so if `orderSchema.clearIndexes()` cleared indexes on child schemas, it would clear `productSchema`'s indexes as well. That behavior isn't ideal, but given that we don't clone child schemas we shouldn't make `clearIndexes()` clear child schema indexes by default. We can add an option though.

lib/schema.js
```diff
@@ -579,6 +580,26 @@ Schema.prototype.add = function add(obj, prefix) {
...
+  if (arguments.length > 1) {
+    throw new Error('removeIndex() takes only 1 argument');
+  }
+
+  if (typeof index !== 'object' && typeof index !== 'string') {
+    throw new Error('removeIndex() may only take either an object or a string as an argument');
+  }
+
+  for (let i = 0; i < this._indexes.length; i++) {
+    if (Object.keys(this._indexes[i][0]) == index || util.isDeepStrictEqual(this._indexes[i][0], index)) {
```
vkarpov15, line 593: `Object.keys(this._indexes[i][0]) == index` -> `this._indexes[i][1].name === index`.

[Indexes have names in MongoDB](https://www.mongodb.com/docs/manual/tutorial/avoid-text-index-name-limit/), so `removeIndex(name)` should remove the index named 'name'.

If the schema index doesn't have a `name` option, then we shouldn't support removing it by `name`. I don't want Mongoose to have to replicate MongoDB's logic for default index names.

test/index.test.js
```diff
@@ -911,5 +911,29 @@ describe('mongoose module:', function() {
...
+      schema.index({ weight: 1 });
+      schema.index({ age: 1, name: 1 });
+      schema.index({ location: 1 });
+      assert.equal(schema._indexes.length, 4);
+      const startingIndexes = JSON.parse(JSON.stringify(schema._indexes));
+      schema.removeIndex('title');
+      schema.removeIndex({ weight: 1 });
+      schema.removeIndex({ age: 1, name: 1 });
+      assert.equal(schema._indexes.length, 1);
+      assert.notDeepStrictEqual(startingIndexes, schema._indexes);
```
vkarpov15, line 934: Add an assert that checks that `schema._indexes[0]` is deep equal `{ location: 1 }`

vkarpov15 (changes requested):

vkarpov15 (approved):

PR #11611, opened 2022-04-01
Add permissions to GitHub actions

neilnaveen:
**Summary**
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

**Examples**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,9 @@ name: Test
 on:
     pull_request:
     push:
+permissions:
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks! I guess I shouldn't be surprised that GitHub actions can modify the repo's contents by default, but I hadn't thought of that as a security issue.

PR #11616, opened 2022-04-03
Create TypeScript compilation benchmark that we run automatically on PRs

vkarpov15:

**Summary**

Issues like #10349 make me anxious about merging TS changes, because sometimes what looks like perfectly reasonable TS causes massive compiler slowdowns. This PR creates a new benchmark workflow that runs `tsc --extendedDiagnostics` several times on a simple script and parses the result. So we can at least have an automated spot check that a given TypeScript-related PR doesn't cause 5x the number of instantiations.

Also, I added some logic to store benchmark results on Mongoose's Atlas cluster, which will help provide some neat charts :)

image

**Examples**


AbdelrahmanHafez (approved):
LGTM, this is a great start for having a time series for benchmarks.
It's important that in the future we have our benchmark files as immutable, so for example, if we were to add another test case, we'd need to create a new file for it, and then the benchmarking logic should create data for each file separately, and save it in the `results` array.

That way we can compare the same test across multiple mongoose versions.
Thanks 👍

benchmarks/typescript.js
```diff
@@ -0,0 +1,82 @@
...
+    return;
+  }
+
+  await mongoose.connect(process.env.DB_URL);
+
+  const BenchmarkResult = mongoose.model('BenchmarkResult', mongoose.Schema({
+    githash: { type: String, required: true },
+    benchmarkName: { type: String, required: true },
+    results: 'Mixed'
+  }, { timestamps: true }), 'BenchmarkResult');
```
AbdelrahmanHafez, line 73: We'd need to add an index to githash, I imagine this collection will grow over time.
```suggestion
    const benchmarkResultSchema = mongoose.Schema({
       githash: { type: String, required: true },
       benchmarkName: { type: String, required: true },
       results: 'Mixed'
     }, { timestamps: true });
     
     benchmarkResultSchema.index({ githash: 1 }, { unique: true });
     
    const BenchmarkResult = mongoose.model('BenchmarkResult', benchmarkResultSchema);
    await BenchmarkResult.syncIndexes();
```

Nitpick: Naming a collection `BenchmarkResult` conflicts with the default (therefore standard in my head) behavior of all lowercase, plural, collections names. I let mongoose do the pluralization instead.

vkarpov15: TBH, I prefer keeping the collection name the same as the model name. I personally have a strong aversion to unnecessary translation layers, so in our existing Mongoose backend logic we keep collection name = model name to avoid having to mentally calculate what Monngoose pluralizes "BenchmarkResult" to.

AbdelrahmanHafez: As long as we're being consistent. 🤷
Even though I agree with your argument right now, since it's the default behavior and almost all projects use the default behavior, I'd rely on that.

But if someone were to be consistent with any pattern, it should be fine.

PR #11617, opened 2022-04-03
Code coverage workflow

### LINKED ISSUE #11308: Add a codecoverage Workflow
@vkarpov15 
Please add a codecoverage Workflow.

E.g. coveralls.io

I am Not a collaborator to activate or atleast send you an invite to activate the 3rd party tool.

If you activate it , I can supply the correct configuration for the workflow.

vkarpov15:
**Summary**

Fix #11308. We've considered adding a code coverage workflow over the years, but never have because we don't have permissions on the Automattic org to add a tool like Coveralls. However, using GitHub actions, we can upload artifacts associated with a workflow, which means we can at least create an HTML zip of code coverage output on each run. You can then download the zip output from the GitHub workflow summary page. For example, [here's the summary page for 405a7f9](https://github.com/Automattic/mongoose/actions/runs/2085815809), here's a screenshot of the output from 405a7f9 after downloading.

image

@Uzlopak 

**Examples**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,3 +71,31 @@ jobs:
 
       - name: Typings
         run: npm run test-tsd
+  
+  coverage:
+    runs-on: ubuntu-latest
+    name: Coverage
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm install
+      - name: Setup
+        run: |
+          wget -q https://downloads.mongodb.org/linux/mongodb-linux-x86_64-${{ matrix.mongo-os }}-${{ matrix.mongo }}.tgz
+          tar xf mongodb-linux-x86_64-${{ matrix.mongo-os }}-${{ matrix.mongo }}.tgz
+          mkdir -p ./data/db/27017 ./data/db/27000
+          printf "\n--timeout 8000" >> ./test/mocha.opts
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez (approved):
LGTM, this is also great for a start.
I think in order to be able to leverage its benefits, we'd need to find a way to make it more accessible.

Ideally, the end result should be a URL easily accessible that serves the HTML from coverage.zip.
Having to download the file, then extracting, then open the HTML file seems like a lot of work and I would probably not check it.

An alternative/addition might be seeing if there is a different data format (JSON?) that we can parse and send to MongoDB as well, and then serving the benchmarks/code coverage stats on a website with cool graphs.

Uzlopak:
I wonder if we can directly upload the lcov file to one of the coverage providers? 

Not tested Idea:
You own the mongoose org on github. Maybe you create a fake/surrogate repo there and create an account in coveralls and get the coveralls token for that surrogate repo. Then we could upload the lcov file directy to coveralls with 
https://www.npmjs.com/package/coveralls by setting

Badge should be also working by pointing to the surrogate project.

vkarpov15:
Re: uploading lcov directly to Coveralls, I've done so in the past: https://github.com/mongodb-js/connect-mongodb-session/blob/3a27eb1f3722334bd141d6f5abf33d958f36e5f2/.travis.yml . The problem is that we still need a coveralls token: https://www.npmjs.com/package/coveralls#running-locally . Can't upload to Coveralls without first adding the repo to Coveralls, and can't add the repo to Coveralls without your org's permission.

Could always fork to the mongoosejs org, but the problem there is maintaining the fork and causing confusion as to which is the  official repo.

I put in a request with Automattic to allow installing Coveralls GitHub app on Automattic's org, we'll see if that works. In the meantime, I'll merge this PR :+1:

image

PR #11624, opened 2022-04-05
Ts benchmark

AbdelrahmanHafez:
**Summary**

This PR allows running TS benchmarks automatically by adding new folders to the `benchmarks/typescript` directory, it also saves each directory separately.
As mentioned in #11616, it's important to make those folders immutable so that our comparison is fair among different mongoose versions.


vkarpov15 (approved):
Thanks :+1:

AbdelrahmanHafez:
Happy to help 👍

PR #11625, opened 2022-04-05
export mongodb types

### LINKED ISSUE #11605: [Types] all used types should be exported
We're currently relying on types from MongoDB types, but we're not exporting them, which means as a user, I can't have the type of options in `Model.bulkSave(documents, options)` because options has the type of `mongodb.BulkWriteOptions` which I can't access from the outside without using [TS Parameters utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype).

So we need to define a local type whenever we use a type from a third party library, use it locally and export it.

So instead of 
```ts
options?: mongodb.BulkWriteOptions
```
We should be doing:
```ts
export type MongoDBBulkWriteOptions = mongodb.BulkWriteOptions;
options?: MongoDBBulkWriteOptions;
```

That way I can import `MongoDBBulkWriteOptions| in my TS and use it.

What do you think?

Uzlopak:
Resolves #11605 

@AbdelrahmanHafez

### DIFF
--- a/test/types/mongodb.test.ts
+++ b/test/types/mongodb.test.ts
@@ -1,0 +1,6 @@
+import * as mongoose from 'mongoose';
+import * as mongodb from 'mongodb';
+import { expectType } from 'tsd';
+
+expectType<mongodb.ClientSession>({} as mongoose.MongoDBClientSession);
+expectType<mongodb.ClientSessionOptions>({} as mongoose.MongoDBClientSessionOptions);
--- a/test/types/mongodb.test.ts
+++ b/test/types/mongodb.test.ts
@@ -1,6 +1,0 @@
-import * as mongoose from 'mongoose';
-import * as mongodb from 'mongodb';
-import { expectType } from 'tsd';
-
-expectType<mongodb.ClientSession>({} as mongoose.MongoDBClientSession);
-expectType<mongodb.ClientSessionOptions>({} as mongoose.MongoDBClientSessionOptions);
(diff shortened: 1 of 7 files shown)


AbdelrahmanHafez:
@Uzlopak 
LGTM, but now that I look at it, we should avoid prefixing types with `MongoDB`, because it implies that it'll always be equal to the mongodb type, and we won't extend it.

Instead of `MongoDBClientSessionOptions` we can have it as `ClientSessionOptions`, which starts as `mongodb.ClientSessionOptions`, and is subect to change in the future if we need to, then the naming would be correct and the downstream apps that rely on our types will have names that are not tied to MongoDB types.

Uzlopak:
@AbdelrahmanHafez 

Now it is just wrapping all the used types from mongodb with MongoDB and they are not modified. 

What about the edge case of ObjectId? We already use that domain specific.

So I am actually more pro for the current status instead of renaming it to MongoDB. But now that it is encapsulated in the mongodb.d.ts renaming in vscode is easy as a cake.

Uzlopak:
I change this from draft to ready for review. 

Waiting for your feedback @AbdelrahmanHafez

vkarpov15 (commented):
I don't understand the benefit of wrapping `mongodb.MongoClientOptions`, etc. What's the advantage of doing things this way?

Uzlopak:
@vkarpov15 

Hmm. Now the remark of @AbdelrahmanHafez makes more sense. Will change it.

Issue #11594, opened 2022-03-29
Missing options attribute for aggregation type


GO-DIE:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
The type declaration file is incomplete. I can't find [Aggregate.prototype.options](https://mongoosejs.com/docs/api/aggregate.html#aggregate_Aggregate-options) in type declaration file.

Uzlopak:
@GO-DIE 

I created a PR #11628 which should fix this issue. Maybe you want to have a look

PR #11628, opened 2022-04-05
Extract aggregate type

Uzlopak:
This PR extracts aggregate, uses the PipelineStage Interfaces everywhere were possible and makes minor changes in the aggregate.js


Uzlopak:
Resolves also #11594

Uzlopak:
Resolves #11594

Uzlopak:
There seems to be an unrelated bug in the master, which makes the build pipeline red. :D

Uzlopak:
Now it exposes options attribute in types as it was requested by #11594 . I assume that you could actually argue that the main intent of the option method is to be a wrapper of the options object. So I guess that the original issue is that you can only set but not read options.

So we could theoretically set the fields of options to be readonly...

vkarpov15 (approved):
Thanks :+1:

PR #11631, opened 2022-04-05
Rename IUser to User

winklerrr:
**Summary**

Don't use prefixes in names for interfaces! See the official TS naming guidelines: <https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines>

**Examples**

Instead of `IUser` write `User`

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -17,21 +17,21 @@ To get started with Mongoose in TypeScript, you need to:
 import { Schema, model, connect } from 'mongoose';
 
 // 1. Create an interface representing a document in MongoDB.
-interface IUser {
+interface User {
   name: string;
   email: string;
   avatar?: string;
 }
 
 // 2. Create a Schema corresponding to the document interface.
-const userSchema = new Schema<IUser>({
+const userSchema = new Schema<User>({
   name: { type: String, required: true },
   email: { type: String, required: true },
   avatar: String
 });
 
 // 3. Create a Model.
-const User = model<IUser>('User', userSchema);
+const UserModel = model<User>('User', userSchema);
 
 run().catch(err => console.log(err));
 
@@ -39,7 +39,7 @@ async function run() {
   // 4. Connect to MongoDB
   await connect('mongodb://localhost:27017/test');
 
-  const user = new User({
+  const user = UserModel.create({
     name: 'Bill',
     email: 'bill@initech.com',
     avatar: 'https://i.imgur.com/dM7Thhn.png'
@@ -53,14 +53,14 @@ async function run() {
 You as the developer are responsible for ensuring that your document interface lines up with your Mongoose schema.
 For example, Mongoose won't report an error if `email` is `required` in your Mongoose schema but optional in your document interface.
 
-The `User()` constructor returns an instance of `HydratedDocument<IUser>`.
-`IUser` is a _document interface_, it represents the raw object structure that `IUser` objects look like in MongoDB.
-`HydratedDocument<IUser>` represents a hydrated Mongoose document, with methods, virtuals, and other Mongoose-specific features.
+The `UserModel.create()` constructor returns an instance of `HydratedDocument<User>`.
+`User` is a _document interface_, it represents the raw object structure that `User` objects look like in MongoDB.
+`HydratedDocument<User>` represents a hydrated Mongoose document, with methods, virtuals, and other Mongoose-specific features.
 
 ```ts
 import { HydratedDocument } from 'mongoose';
 
-const user: HydratedDocument<IUser> = new User({
+const user: HydratedDocument<User> = UserModel.create({
   name: 'Bill',
   email: 'bill@initech.com',
   avatar: 'https://i.imgur.com/dM7Thhn.png'
@@ -75,7 +75,7 @@ To define a property of type `ObjectId`, you should use `Types.ObjectId` in the 
 import { Schema, Types } from 'mongoose';
 
 // 1. Create an interface representing a document in MongoDB.
-interface IUser {
+interface User {
   name: string;
   email: string;
   // Use `Types.ObjectId` in document interface...
@@ -83,7 +83,7 @@ interface IUser {
 }
 
 // 2. Create a Schema corresponding to the document interface.
-const userSchema = new Schema<IUser>({
+const userSchema = new Schema<User>({
   name: { type: String, required: true },
   email: { type: String, required: true },
   // And `Schema.Types.ObjectId` in the schema definition.
@@ -103,7 +103,7 @@ Many Mongoose TypeScript codebases use the below approach.
 ```typescript
 import { Document, Schema, model, connect } from 'mongoose';
 
-interface IUser extends Document {
+interface User extends Document {
   name: string;
   email: string;
   avatar?: string;
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -31,7 +31,7 @@ const userSchema = new Schema<User>({
 });
 
 // 3. Create a Model.
-const UserModel = model<User>('User', userSchema);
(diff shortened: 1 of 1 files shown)


Uzlopak:
I dont care, I use I-prefix and T-prefix in my projects, but I want to emphasize that the naming guideline is not a guideline for the TS community but only relevant for the TS Project.

   AGAIN: This is NOT a prescriptive guideline for the TypeScript community

winklerrr:
But then at least stay with one programming style.
On this page the I-prefix is used: <https://mongoosejs.com/docs/typescript/statics.html>
On this next page it isn't: <https://mongoosejs.com/docs/typescript/populate.html>

Uzlopak:
My "I dont care", was not meant to be disencouraging. I just wanted to express, that I dont have a strong opinion about it. But i just wanted to clarifiy that it is not an authoritative argument. 

Your style argument is indeed changing my opinion to a strong positive.

docs/typescript.md
```diff
@@ -17,29 +17,29 @@ To get started with Mongoose in TypeScript, you need to:
...
+const UserModel = model<User>('User', userSchema);
 
 run().catch(err => console.log(err));
 
 async function run() {
   // 4. Connect to MongoDB
   await connect('mongodb://localhost:27017/test');
 
-  const user = new User({
+  const user = UserModel.create({
```
Uzlopak, line 42: Well, this is not equivalent to the previous code. Why did you change it?

vkarpov15: Please undo this change, should be `const user = new UserModel({ ...`. `UserModel.create()` is async.

vkarpov15 (changes requested):

docs/typescript.md
Uzlopak:
```suggestion
  const user = new UserModel({
```

docs/typescript.md
```diff
@@ -17,29 +17,29 @@ To get started with Mongoose in TypeScript, you need to:
...
-const userSchema = new Schema<IUser>({
+const userSchema = new Schema<User>({
   name: { type: String, required: true },
   email: { type: String, required: true },
   avatar: String
 });
 
 // 3. Create a Model.
-const User = model<IUser>('User', userSchema);
+const UserModel = model<User>('User', userSchema);
```
Uzlopak, line 34:
```suggestion
const User = model<User>('User', userSchema);
```

docs/typescript.md
```diff
@@ -53,14 +53,14 @@ async function run() {
 You as the developer are responsible for ensuring that your document interface lines up with your Mongoose schema.
 For example, Mongoose won't report an error if `email` is `required` in your Mongoose schema but optional in your document interface.
 
-The `User()` constructor returns an instance of `HydratedDocument<IUser>`.
-`IUser` is a _document interface_, it represents the raw object structure that `IUser` objects look like in MongoDB.
-`HydratedDocument<IUser>` represents a hydrated Mongoose document, with methods, virtuals, and other Mongoose-specific features.
+The `UserModel.create()` constructor returns an instance of `HydratedDocument<User>`.
```
Uzlopak, line 56:
```suggestion
The `User()` constructor returns an instance of `HydratedDocument<User>`.
```

docs/typescript.md
```diff
@@ -53,14 +53,14 @@ async function run() {
...
+`User` is a _document interface_, it represents the raw object structure that `User` objects look like in MongoDB.
+`HydratedDocument<User>` represents a hydrated Mongoose document, with methods, virtuals, and other Mongoose-specific features.
 
 ```ts
 import { HydratedDocument } from 'mongoose';
 
-const user: HydratedDocument<IUser> = new User({
+const user: HydratedDocument<User> = UserModel.create({
```
Uzlopak, line 63:
```suggestion
const user: HydratedDocument<User> = new User({
```

Uzlopak:
@vkarpov15 
I made the changes...

Uzlopak:
I think it is safe to merge this.

AbdelrahmanHafez:
I personally try to make any piece of code I read convey what it is just by reading the name, and to do so I use a widely-accepted convention:
`IUser` for the user interface, representing the type.
`User` for the model, representing a class in OOP.
`user` for the document, representing an instance in OOP.

Having to look at a `User` variable, then having to look around it to see what it is makes me do a mental effort that can be avoided.

I do agree though, it's better to remain consistent, but I do oppose those changes because they conflict with how we're defining the model as `User` in the rest of the docs, and then unnecessarily naming it as `UserModel` in the TS docs only to avoid conflicting with the `User` interface.

What do you think? @winklerrr @Uzlopak @vkarpov15

Uzlopak:
Generally speaking:
There is actually no conflict in ts when an interface is named like a variable.

On the other hand, we have in typescript types and interfaces, which are kind of the same thing, and prefixing Interfaces with I or T makes them potentially not changeable, if we later change an interrface to a type or a type to an interface. 

Here in specific:
In hindsight think I made a mistake by merging this. If we name them the same, than User variable is doing something else than the interface. Naturally you would expect that the type of the variable and the interface will have the same shape if they have the same name. Now this implicit contract is broken. 

I was just convinced, by the argument, that we have a non-I-prefix used on the populate page. But there we name it ParentModel.

So I would be feeling better, if we agree on a style in the docs and apply it everywhere in the docs. 

I am fine with your suggestion. But I worked in a company where the Models were sufixed with Model, like in the populate example.

winklerrr:
@Uzlopak my first PR was just to remove the I-prefix to be conform to the TS handbook on not using Hungarian notation.
But I also recognized the naming problem after I renamed the variables in a project of mine.

Using the list that @AbdelrahmanHafez wrote down, I would name the variables like this:

- `UserInput` for the interface, representing all inputs that are needed to create an instance of the document
- `UserDocument` for the interface, representing the whole type of the document (extending `UserInput`, see example link below). It also corresponds more to the variable name `DocType` in the implementation of `@types/mongoose`
- `userSchema` for the schema instance (`schema<UserDocument>`).
- `UserModel` for the model class (`model<UserDocument>`)
- `user` for the document instance, created by `new UserModel` (or `UserModel.create`)

For the `UserInput` and `UserDocument` interface extension, please check out this [user.model.ts][1] example.
IMHO, it shows a clean implementation of how to use Typescript's autocomplete feature, enabling the IDE to present the dev with only the required input fields when working with this model.

```ts
export interface UserInput {
  email: string;
  firstName: string;
  lastName: string;
  password: string;
}

export interface UserDocument extends UserInput {
  fullName: string;
  createdAt: Date;
  updatedAt: Date;
  comparePassword(candidatePassword: string): Promise<boolean>;
}

[...]

export default mongoose.model<UserDocument>("User", userSchema);
```

---

In the same project, the [user.service.ts][2] shows that the model class should be named `UserModel` instead of solely `User` because when writing the code, one is actually speaking to the whole `UserModel` to `findOne` or `deleteMany` documents. One is not speaking to a single `User` instance.

For example, IMHO this is more understandable when read:

```ts
export async function deleteAllUsers() {
  return UserModel.deleteMany({});
}
```

than this:

```
export async function deleteAllUsers() {
  return User.deleteMany({});
}
```

[1]: https://github.com/TomDoesTech/The-Ultimate-Guide-to-TypeScript-With-Mongoose/blob/main/src/models/user.model.ts
[2]: https://github.com/TomDoesTech/The-Ultimate-Guide-to-TypeScript-With-Mongoose/blob/main/src/services/user.service.ts

Uzlopak:
I reverted the PR.

winklerrr:
@Uzlopak what do you think about my naming proposal?

AbdelrahmanHafez:
Let's open a new issue to discuss a naming convention to follow in the documentation.

PR #11633, opened 2022-04-05
Improve GitHub actions

Uzlopak:
We should only run relevant github actions when the corresponding files are modified.


package.json
```diff
@@ -63,7 +63,9 @@
     "lib": "./lib/mongoose"
   },
   "scripts": {
-    "lint": "eslint .",
+    "lint": "npm run lint-ts && npm run lint-js",
```
vkarpov15, line 66: I'd rather avoid this change because I like to run `npm run lint -- --fix` to fix linter issues, and this would break that command. Maybe we keep `lint` script the same as it was, and add `lint-js` and `lint-ts`?

.github/workflows/test.yml
```diff
@@ -1,12 +1,42 @@
...
+      - uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846 # v3
+
+      - name: Setup node
+        uses: actions/setup-node@5b52f097d36d4b0b2f94ed6de710023fbb8b2236 # v3.1.0
+        with:
+          node-version: 14
+
+      - run: npm install
+
+      - name: Lint JS-Filess
```
vkarpov15, line 34: Typo "Filess" -> "files"

.github/workflows/benchmark.yml
```diff
@@ -1,13 +1,24 @@
...
+      - 'types/**'
+      - 'benchmarks/typescript/**'
 permissions:
   contents: read
 
 jobs:
   typescript:
     runs-on: ubuntu-20.04
-    name: TypeScript
+    name: Benchmark TypeScript-Types
```
vkarpov15, line 21: Why the "-" here?

vkarpov15 (commented):
A couple of comments, but looks good in general

Uzlopak:
@vkarpov15 
Any idea why tidelift does not work?

vkarpov15:
@Uzlopak not off the top of my head, I asked Tidelift in Slack so hopefully should be able to get that working again soon. In the meantime, feel free to ignore it.

vkarpov15 (approved):

AbdelrahmanHafez (commented):
Nice!

PR #11639, opened 2022-04-07
perf: improve perf of key order

Uzlopak:
I actually dont understand why the key ordre is so important. Nevertheless, this implementatoin is faster.

without reordering:
MONGOOSE_DEV=1 node create.js 
CheckItem x 201,161 ops/sec ±1.00% (92 runs sampled)
Board x 182 ops/sec ±0.82% (83 runs sampled)

before:
 MONGOOSE_DEV=1 node create.js 
CheckItem x 170,092 ops/sec ±1.35% (90 runs sampled)
Board x 168 ops/sec ±1.22% (80 runs sampled)

after:
MONGOOSE_DEV=1 node create.js 
CheckItem x 195,166 ops/sec ±0.41% (96 runs sampled)
Board x 174 ops/sec ±1.29% (80 runs sampled)

So we get pretty close to the non-ordered object.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1052,7 +1052,6 @@ Document.prototype.$set = function $set(path, val, type, options) {
   const merge = options.merge;
   const adhoc = type && type !== true;
   const constructing = type === true;
-  const typeKey = this.$__schema.options.typeKey;
   let adhocs;
   let keys;
   let i = 0;
--- a/lib/document.js
+++ b/lib/document.js
@@ -1154,14 +1154,15 @@ Document.prototype.$set = function $set(path, val, type, options) {
       }
     }
 
-    // Ensure all properties are in correct order by deleting and recreating every property.
-    for (const key of Object.keys(this.$__schema.tree)) {
-      if (this._doc.hasOwnProperty(key)) {
-        const val = this._doc[key];
-        delete this._doc[key];
-        this._doc[key] = val;
-      }
+    // Ensure all properties are in correct order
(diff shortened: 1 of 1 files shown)


Uzlopak:
You may ask why this is faster, and maybe it is even more correct.

So in the original solution, the keys are deleted and added back to the object, as such the keys will be added at the end of the Object and thus sorting the keys. This of course works, but if you have by any chance some additional keys, which you were not aware of, they will be at the beginning.

E.g.

you have a object with following keys

```
firstName
lastName
_id
additional
```

and we have to sort them by the schema and the schema has following keys:
```
_id
firstName
lastName
```

then with the old implementation, you would get 

```
additional
_id
firstName
lastName
```

The new way makes use of the fact, that if you assign an Object to an other Object, it will use the order of the keys of the other Object. So we first generate a new Object with keys where the value is undefined. Then we assign the original Object to the new Object and voila have all the keys in the right order. Additional keys will be at the end of the object.

Like this:

```
_id
firstName
lastName
additional
```

Uzlopak:
@AbdelrahmanHafez 

The unit test for this is already existing.

https://github.com/Automattic/mongoose/blob/dc0a332d5494501725c542a3be9762a5b9093d87/test/document.test.js#L10556

vkarpov15 (approved):
This is very cool, thanks :+1: Here's a little more about why key order matters: https://thecodebarbarian.com/introducing-mongoose-6.html#schema-enforced-key-order-when-saving

PR #11640, opened 2022-04-07
disable debug option in virtuals unit test

Uzlopak:
Unnecessary output of the mongo operations when running unit tests, because I guess somebody forgot to remove the debug option.


AbdelrahmanHafez (approved):
LGTM, thanks 👍

Issue #11598, opened 2022-03-30
[Types] `populate('path', 'selection', model)` syntax not supported in types


hasezoey:
**Do you want to request a *feature* or report a *bug*?**
types

**What is the expected behavior?**
from what i can tell the syntax `doc.populate('path', 'selection', model)` is currently not in the types

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
mongoose 6.2.9

---

Reference to the syntax i mean: https://github.com/Automattic/mongoose/blob/ec7743882791db221641798096cb6bfe20b89e79/lib/document.js#L4150 (only that from what i can tell a third parameter for a model is also supported)

Uzlopak:
@mohammad0-0ahmad 
Can you have a look at it?

mohammad0-0ahmad:
This issue is confirmed @Uzlopak, 
This is related to these types: 
https://github.com/Automattic/mongoose/blob/337d6ad2f6bbd92e186a2774fbcb8a6fb01e8ed3/types/index.d.ts#L411

PR #11649, opened 2022-04-08
Refactor doc.populate FN type.

mohammad0-0ahmad:
Related to : https://github.com/Automattic/mongoose/issues/11598

### DIFF
--- a/test/types/document.test.ts
+++ b/test/types/document.test.ts
@@ -178,6 +178,6 @@ function gh11435() {
 }
 
 async function gh11598() {
-  const doc = await Test.findOne();
-  doc?.populate('favoritDrink', undefined, model('temp', new Schema()));
+  const doc = await Test.findOne().orFail();
+  doc.populate('favoritDrink', undefined, model('temp', new Schema()));
 }
--- a/test/types/document.test.ts
+++ b/test/types/document.test.ts
@@ -1,7 +1,18 @@
-import { Schema, model, Model, Document, Error, Types } from 'mongoose';
+import { Schema, model, Model, Document, Types } from 'mongoose';
 import { expectError, expectType } from 'tsd';
 
-const schema: Schema = new Schema({ name: { type: 'String', required: true }, address: new Schema({ city: { type: String, required: true } }) });
+const Drink = model('Drink', new Schema({
+  name: String
+}));
+
+const schema: Schema = new Schema({
+  name: { type: 'String', required: true },
+  address: new Schema({ city: { type: String, required: true } }),
+  favoritDrink: {
+    type: Schema.Types.ObjectId,
+    ref: Drink
+  }
+});
 
 interface ITestBase {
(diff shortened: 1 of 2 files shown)


types/document.d.ts
```diff
@@ -186,10 +186,8 @@ declare module 'mongoose' {
     $parent(): Document | undefined;
 
     /** Populates document references. */
-    populate<Paths = {}>(path: string | PopulateOptions | (string | PopulateOptions)[], callback: Callback<this & Paths>): void;
-    populate<Paths = {}>(path: string, names: string, callback: Callback<this & Paths>): void;
-    populate<Paths = {}>(path: string, names: string): Promise<this & Paths>;
-    populate<Paths = {}>(path: string | PopulateOptions | (string | PopulateOptions)[]): Promise<this & Paths>;
+    populate<Paths = {}>(path: string | PopulateOptions | (string | PopulateOptions)[], select?: string | AnyObject, model?: Model<unknown>, match?: AnyObject, options?: PopulateOptions, callback?: Callback<this & Paths>): void;
```
mohammad0-0ahmad, line 189: Should option argument be existing?

Uzlopak: @vkarpov15

test/types/document.test.ts
```diff
@@ -164,4 +175,9 @@ function gh11435() {
   ItemSchema.pre('validate', function preValidate() {
     expectType<Model<unknown>>(this.model('Item1'));
   });
+}
+
+async function gh11598() {
+  const doc = await Test.findOne();
```
vkarpov15, line 181: You can do `await Test.findOne().orFail()` and avoid the need for `?.` below

types/document.d.ts
vkarpov15: Yes it should.

types/document.d.ts
```diff
@@ -186,10 +186,8 @@ declare module 'mongoose' {
     $parent(): Document | undefined;
 
     /** Populates document references. */
-    populate<Paths = {}>(path: string | PopulateOptions | (string | PopulateOptions)[], callback: Callback<this & Paths>): void;
-    populate<Paths = {}>(path: string, names: string, callback: Callback<this & Paths>): void;
```
vkarpov15, line 190: Don't get rid of these. The issue is that, if there's more than 1 non-callback arg, `path` must be a string. We don't support `populate(PopulateOptions, string, Model<unknown>)`, for example. So keep this syntax.

vkarpov15 (changes requested):

mohammad0-0ahmad:
@vkarpov15 
I've noticed that when rebasing this PR on top of master branch, the model loose and doc for example lose its type & become any.
Although, all tests are passed.
Can u please double-check this?
Untitled

vkarpov15 (approved):
Nice work, thanks :+1:

Issue #11644, opened 2022-04-07
What's the best way to keep track of the changes made to a document?


josegl:
Mongoose version: 6.2.2

Hi. What would be the best option to keep track of the changes that have been made to a document? I'm trying to keep an array of changes like this:
```javascript
[
  {path: '/my/path/', old_value: 'previous_value_for this path'}
]
```
And I want to achieve this the most transparent way for our developers teams. This means that I want to keep changes no matter how a document is updated. So if `myDocument` is `{prop: 2}`, then we may update `myDocument` using any of this ways:
```javascript
myDocument.prop = 1;
myDocument.set('prop', 1);
```
Or applying a JSON Patch:
```javascript
const patch = [{op: 'replace', path: '/prop', value: 1}];
fastJsonPatch.applyPatch(myDocument, patch, false, true, true); //This updates the document instead of creating a new one
```
In any of the options the target is to have an array `_changes` that in any of the examples above would be like this after applying the changes:
```javascript
[{path: '/prop', old_value: 1}]
```
In order to achieve this feature I've created a Mongoose plugin:

```javascript
const util = require('util');
const jsonpatch = require('fast-json-patch')

const proxy_handler = {
  apply: function (target, this_arg, arglist){
    const path = '/' + arglist[0].split('.').filter(p => p).join('/');
    const old_value = this_arg.get(arglist[0]);
    const change = {path, old_value};
    if(this_arg._changes){
      if(!this_arg._changes.some(change => change.path === path && util.isDeepStrictEqual(old_value, change.old_value))){//This change does not exist yet. (the same change could already exist because markModified is recursive
        this_arg._changes.unshift(change); //we insert the changes at the beggining of the array because if we have to revert the changes it is not neccesary to revert the array.
      }
    }else{
      this_arg._changes = [change];
    }
    const newtarget = target.bind(this_arg);
    newtarget(...arglist);
  }
};

const changesTracker = schema => {
  schema.post('init', function(doc){
    const $setProxy = new Proxy(doc.$set, proxy_handler);
    const setProxy = new Proxy(doc.set, proxy_handler);
    const markModifiedProxy = new Proxy(doc.markModified, proxy_handler);
    doc.$set = $setProxy;
    doc.set = setProxy;
    doc.markModified = markModifiedProxy;
  });

  schema.pre('save', function(next){
    if(this.isNew){//we do the same that in the post-init middleware because when a model is created using mongoose the init middleware does not apply
[...]

PR #11660, opened 2022-04-13
Call markModified before setting changes in Array  and in DocumentArray methods

josegl:
@vkarpov15 is there any reason why in the `array` and `DocumentArray` methods we shouldn't call `markModified` before actually write the new value for the given path?

I understand that in order to mark something as modified we must have updated it before. However I do not know how we could keep track of the previous values when updating a single value from a nested array.

This pull request is motivated by this issue: https://github.com/Automattic/mongoose/issues/11644
In this comment there is detailed info and repro: https://github.com/Automattic/mongoose/issues/11644#issuecomment-1094891011

In this pull request, calling `markModified` before writing the new value for the given path, we can keep track of the changes when updating a single value of a nested array.

Example:
```javascript
const MONGO_URL = 'mongodb://localhost/mongoose-track-changes';
const changesTracker = require('./index');  // https://github.com/josegl/mongoose-track-changes/blob/main/index.js

const notificationSchema = mongoose.Schema({
  notify_to: [String],
});

const taskSchema = new mongoose.Schema({
  notification: notificationSchema,
  description: String,
});

taskSchema.plugin(changesTracker);

test('Mongoose-track-changes', async () => {
  await mongoose.connect(MONGO_URL);
  const taskModel = mongoose.model('Task', taskSchema);
  const task = taskModel({
    notification: {
      notify_to: new Array(3).fill(0).map((_, i) => `id${i}`),
    },
    description: 'test',
  });
  await task.save();
  const saved_task = await mongoose.models.Task.findOne({});
  saved_task.notification.notify_to[2] = 'id3';
});
```
Without the changes in this pull request, after using this plugin:
https://github.com/josegl/mongoose-track-changes
In the `task._changes` array will be only a single change, and it will be this:
```javascript
{path: '/notification/notify_to/2', old_value: 'id3'}
```
When the desired value is:
```javascript
{path: '/notification/notify_to/2', old_value: 'id2'}
```
With the changes applied by this PR, the value is exactly the desired value.

I would understand your hesitations to merge this PR because as I said before we can argue that semantically you cannot mark something as modified if it has not been modified yet. If so, what would be the best solution in order to keep track of a deltas array, no matter how a document is updated, I mean that we want to keep the changes if you use this syntax:
```javascript
myModel.set('any.path.you.want', 'new_value');
```
or this one:
```javascript
myModel.any.path.you.want = 'new_value';
```
In both cases we want to have the same deltas array:
```javascript
myModel._changes
// [{path: '/any/path/you/want', old_value: 'previous_value' || undefined /*if there was not a previous value*/}]
```

### DIFF
--- a/lib/types/DocumentArray/methods/index.js
+++ b/lib/types/DocumentArray/methods/index.js
@@ -311,8 +311,8 @@ const methods = {
       return this;
     }
     const value = methods._cast.call(this, val, i);
-    arr[i] = value;
     methods._markModified.call(this, i);
+    arr[i] = value;
     return this;
   },
 
--- a/lib/types/array/methods/index.js
+++ b/lib/types/array/methods/index.js
@@ -414,9 +414,9 @@ const methods = {
       }
 
       if (!found) {
+        this._markModified();
         rawArray.push(v);
         this._registerAtomic('$addToSet', v);
-        this._markModified();
         [].push.call(added, v);
       }
     }, this);
@@ -510,9 +510,9 @@ const methods = {
 
   nonAtomicPush() {
     const values = [].map.call(arguments, this._mapCast, this);
+    this._markModified();
     const ret = [].push.apply(this, values);
     this._registerAtomic('$set', this);
-    this._markModified();
     return ret;
   },
 
@@ -530,9 +530,9 @@ const methods = {
    */
 
   pop() {
+    this._markModified();
     const ret = [].pop.call(this);
     this._registerAtomic('$set', this);
-    this._markModified();
     return ret;
   },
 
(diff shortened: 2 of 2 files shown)


vkarpov15:
I'm unable to access https://github.com/josegl/mongoose-track-changes - is that repo private?

You're right that I'm a bit hesitant to merge this PR. I think the risk of merging is minimal, but I'm not sure I see any reason to merge it. This change makes it seem like you're using a plugin which overrides `_markModified()`, which isn't supported.

Ideally, you'd create this delta in pre save middleware by looping over a document's `directModifiedPaths()`. However, that doesn't offer a way to get the previous value in an easy way.

josegl:
Your we're right, https://github.com/josegl/mongoose-track-changes was a private repo. Now it its public.

The plugin does not exactly overrides `_markModified()`, it proxies it extending  the functionality.

vkarpov15 (approved):
As much as I wish we had a better alternative for this use case, we don't right now, and I don't see any harm in merging this. Thanks :+1:

PR #11665, opened 2022-04-14
Small grammatical mistake

MMortaga:
Seems like a typo if I'm not mistaken.

**Summary**

Just a small typo.
**Examples**


Uzlopak:
It seems actually correct. It says, that you should use ObjectIds as Reference unless you know what you are doing by using e.g. strings as foreignKey.

MMortaga:
> It seems actually correct. It says, that you should use ObjectIds as Reference unless you know what you are doing by using e.g. strings as foreignKey.

This might be the case but got me confused for a second. If that's how it's intended, you're free to close the PR.

PR #11688, opened 2022-04-19
Update mongodb-dependent type parameters

### LINKED ISSUE #11687: tsconfig.json format is wrong, so strictNullChecks is ignored
Bug:

Calls to `tsc` ignore `strictNullChecks` because it's not inside the compilerOptions object:

```json
{
  "compilerOptions": {
    "paths": {
      "mongoose" : ["./types/index.d.ts"]
    }
  },
  "strictNullChecks": true
}
```

`tsc` should have strict null checks on like so:

```json
{
    "compilerOptions": {
        "strictNullChecks": true,
        "paths": {
            "mongoose" : ["./types/index.d.ts"]
        }
    }
}
```

When it does, there are 5 errors in connection.d.ts and index.d.ts [1] on the Typescript 4.7 beta. I'll take a look at the errors and try to fix them in a PR. 

[1]
Something like this:

```
15:03 $ tsc --noemit --pretty false
types/connection.d.ts(209,113): error TS2344: Type 'ResultType' does not satisfy the constraint 'Document'.
types/index.d.ts(186,58): error TS2344: Type 'T' does not satisfy the constraint 'Document'.
types/index.d.ts(320,129): error TS2344: Type 'T' does not satisfy the constraint 'Document'.
types/index.d.ts(321,145): error TS2344: Type 'T' does not satisfy the constraint 'Document'.
types/index.d.ts(470,133): error TS2344: Type 'ResultType' does not satisfy the constraint 'Document'.
```

sandersn:
mongodb.Collection and mongodb.ChangeStream now both require that their type arguments `extend mongodb.Document`. This requires adding this constraint in a few places.

Fixes #11687

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -205,7 +205,7 @@ declare module 'mongoose' {
    * http://mongoosejs.com/docs/api.html#drivers-node-mongodb-native-collection-js
    */
   let Collection: Collection;
-  interface Collection<T extends AnyObject = AnyObject> extends CollectionBase<T> {
+  interface Collection<T extends mongodb.Document = mongodb.Document> extends CollectionBase<T> {
     /**
      * Collection constructor
      * @param name name of the collection
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -183,7 +183,7 @@ declare module 'mongoose' {
    * section collection.js
    * http://mongoosejs.com/docs/api.html#collection-js
    */
-  interface CollectionBase<T> extends mongodb.Collection<T> {
+  interface CollectionBase<T extends mongodb.Document> extends mongodb.Collection<T> {
     /*
       * Abstract methods. Some of these are already defined on the
       * mongodb.Collection interface so they've been commented out.
@@ -205,7 +205,7 @@ declare module 'mongoose' {
    * http://mongoosejs.com/docs/api.html#drivers-node-mongodb-native-collection-js
    */
   let Collection: Collection;
-  interface Collection<T = AnyObject> extends CollectionBase<T> {
+  interface Collection<T extends AnyObject = AnyObject> extends CollectionBase<T> {
     /**
      * Collection constructor
      * @param name name of the collection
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -205,7 +205,7 @@ declare module 'mongoose' {
    * http://mongoosejs.com/docs/api.html#drivers-node-mongodb-native-collection-js
    */
   let Collection: Collection;
-  interface Collection<T = AnyObject> extends CollectionBase<T> {
+  interface Collection<T extends AnyObject = AnyObject> extends CollectionBase<T> {
```
sandersn, line 208: I used `AnyObject` here for readability, but `mongodb.Document` and `AnyObject` are interchangeable because they have the same definition.

jokester:
This PR may also help my https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59700#issuecomment-1103460297 to land. Thanks 👍

types/index.d.ts
Uzlopak: maybe we should use mongodb.Document, because what happens if they change their typing?

sandersn: That makes more sense, especially since CollectionBase has the same constraint on its type parameter.

Uzlopak:
Changes make sense to me. Only question  for me is, does it have some backwards breaking? Or is mongodb.Document really just an AnyObject equivalent?

sandersn:
Right now, mongodb has

```ts
interface Document {
  [key: string]: any
}
```

and mongoose has 

```ts
interface AnyObject {
  [k: string]: any
}
```

So there's definitely no break right now. They're identical from the compiler's perspective. mongodb.Document might change in the future, but (1) I assume that's unlikely (2) it makes sense at a surface level that interfaces that extend mongodb types and functions that return mongodb types should use mongodb-based types instead of mongoose types.

On (2), I don't know mongodb, I just maintain Definitely Typed, so it's a guess on my part.

Uzlopak (approved):
LGTM!

PR #11692, opened 2022-04-20
use node 18 instead of node 17 in CI/CD

Uzlopak:
node 18 is out, node 17 is history. time to change the CI/CD platform matrix for node


AbdelrahmanHafez (approved):
LGTM!

PR #11693, opened 2022-04-20
fix jsdoc for mongoose.createConnection

### LINKED ISSUE #11686: createConnection does not return a Promise, as it should
Bug

Running createConnection() gives the following error "TypeError: mongoose.createConnection(...).then is not a function"

mongoose.createConnection('mongodb://localhost:27017/db', {}).then((db) => {});

That the function would return a Promise, according to the documentation:
@return {Connection} the created Connection object. Connections are thenable, so you can do `await mongoose.createConnection()`

Mongoose 6.3.0
The problem is this line: https://github.com/Automattic/mongoose/blob/master/lib/index.js#L286
In Mongoose 5.12.3 its working, and then the code reads:
```
 if (arguments.length > 0) {
    return conn.openUri(uri, options, callback);
  }

  return conn;
```
  
  It **returns** the Promise created by openUri()

Uzlopak:
Resolves #11686


AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #11695, opened 2022-04-20
fix: make mongoose typings work without esmModuleInterop true

### LINKED ISSUE #11664: Typescript: default import does not work without esModuleInterop

Default import does not work without `esModuleInterop`.
https://github.com/Automattic/mongoose/pull/11343 was planned to fix this, but it seems like it doesn't work.

`package.json`
```json
{
  "name": "mongoose-ts",
  "version": "1.0.0",
  "dependencies": {
    "mongoose": "6.2.11"
  },
  "devDependencies": {
    "typescript": "4.6.3"
  }
}
```
`tsconfig.json`
```json
{
  "compilerOptions": {
	"target": "ES2021",
	"module": "commonjs",
	"outDir": "./dist",
	"noEmitOnError": true,
	"strict": true,
	"skipLibCheck": true,
  }
}
```
`index.ts`
```ts

mongoose.model('a')
```

The error:
```
index.ts:1:8 - error TS1192: Module '"mongoose"' has no default export.

1 import mongoose from 'mongoose'
         ~~~~~~~~
```

It should compile

Mongoose 6.2.11
Node.js v17.8.0
Typescript 4.6.3

Uzlopak:
Resolves #11664


Uzlopak:
This was kind of messy Bug.

To get it work without esmModuleInterOp I had to move the imports into the namespace and do the default export in the namespace. Exporting a default forces you to export all types, interfaces,namespaces and classes in that specific namespace to be explicitly exported.

Moved NativeError and NativeDate out of the mongoose-namespace, as that was clearly wrong as we dont export such classes. Changed two "classes" (SchemaTypeOptions, MixedSchemaTypeOptions) to interfaces, as they can not be instantiated.

Fixed also validateSync typing and modified model.validate to properly handle pathsToValidate (normalized the behaviour).

vkarpov15:
At first glance looks good. I'll do some quick tests this PR over the next couple of days.

vkarpov15 (approved):
Thanks :+1:

PR #11705, opened 2022-04-22
Updated Readme.md

mahendrap1512:
**Summary**
I just start using mongoose, and by default it was creating collection name in plural version, I scanned the docs but couldn't find the solution, So I thought it would be good to be added in the quick start doc.

**Examples**
This is just a documentation update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,13 +153,7 @@ The first argument is the _singular_ name of the collection your model is for. *
 const MyModel = mongoose.model('Ticket', mySchema);
 ```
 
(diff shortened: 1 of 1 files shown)


Uzlopak:
It is actually documented here:
https://mongoosejs.com/docs/api/mongoose.html#mongoose_Mongoose-model

mahendrap1512:
> It is actually documented here: https://mongoosejs.com/docs/api/mongoose.html#mongoose_Mongoose-model

Thanks for the information. 👍

vkarpov15 (approved):
Thanks :+1:

PR #11708, opened 2022-04-22
remove include in tsconfig.json

Uzlopak:
@mohammad0-0ahmad reported that with this [change](https://github.com/Automattic/mongoose/commit/5194756664665a6bc4ed5769d300f3a9d1e6f2c3)  the intellisense in vscode atleast in our typescript tests is broken. 

I dont understand why the include was done anyway. tsd is not using the configuration in tsconfig.json but from the tsc-configuration in package.json. 

Maybe @sandersn can give a clue why he added include instruction to tsconfig.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,6 +120,7 @@
       "allowSyntheticDefaultImports": true,
       "strictPropertyInitialization": false,
       "noImplicitAny": false,
+      "strictNullChecks": true,
       "module": "commonjs",
       "target": "ES2017"
     }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,6 +4,5 @@
     "paths": {
       "mongoose" : ["./types/index.d.ts"]
     }
-  },
-  "include": ["types/*"]
+  }
 }


vkarpov15 (approved):
I'm not entirely convinced that this `include` is what caused the issue. But I think we should be ok to merge this PR just in case.

Also, now that we're using tsd for TypeScript tests, do we even need a top-level `tsconfig.json`? If we're not using it, I want to remove it.

Uzlopak:
The reason for the tsconfig.json is actually: When you use vscode to develop the typings, then vscode looks into the root of the project and loads that tsconfig. I assume other intelliSense editors als use that tsconfig. So it is helping while working on the typings. Or else you are kind of blind while developing the typings and have to run tsd all the time to be sure it works as expected.

sandersn:
> Maybe @sandersn can give a clue why he added include instruction to tsconfig.json

I don't think I had a correct setup because I got lots of errors from the files in `test/`. I guessed that only the types in `types/` were part of the project and that  `test/` files were loose example files that shouldn't be treated as part of mongoose itself.

PR #11584, opened 2022-03-25
chore(deps-dev): bump webpack from 4.44.1 to 5.70.0

AbdelrahmanHafez:
This PR upgrades webpack from v4.44.1 to v5.70.0.

The affecting breaking change is that, in webpack v4, Webpack used to polyfill node dependencies automatically which had performance drawbacks, in v5 webpack changed this to have people opt-in the dependencies they need for their applications.

Which means we need to install the following 3 dependencies:
* assert (the browser version)
* crypto-browserify
* stream-browserify

My concern is regarding `assert`, it has a conflicting name with node's native `assert`, which means that we _might_ get a slightly different behavior from node's assert.

Not sure if having `assert` as a dev dependency will mean that in production apps using mongoose will use the native `assert` or the browserify one, but even if production apps will use the native `assert`, another concern would be that we will have different behavior in the development process and CI will happily pass and we'd know about potential issues too late.

So it seems like our options are as follows:
* Use node:assert, but that's only supported in v14 (only ESM), and v16 for CJS. So we can't do that now since we need to support Node 12.
* Create an alias to assert-browserify that refers to npm:assert and use it.
* Use a copy of npm:assert which is called [assert-browserify](https://www.npmjs.com/package/assert-browserify), but it's not an official release, so I prefer not to go in this direction.
* Avoid upgrading Webpack until we can use the [`node:` protocol](https://2ality.com/2021/12/node-protocol-imports.html).

I'm leaning towards creating an alias to [assert](https://www.npmjs.com/package/assert) and calling it assert-browserify, and only use it for webpack.

What do you think?

Also, after upgrading Webpack started show warnings it didn't previously show regarding bundle size, any idea how we can decrease the bundle size?
```
Starting browser build...
asset ./dist/browser.umd.js 922 KiB [emitted] [minimized] [big] (name: main) 1 related asset
runtime modules 740 bytes 4 modules
modules by path ./node_modules/ 1.62 MiB
  javascript modules 1.6 MiB 204 modules
  json modules 12.7 KiB
    modules by path ./node_modules/browserify-sign/ 2.23 KiB
      ./node_modules/browserify-sign/browser/algorithms.json 2.07 KiB [built] [code generated]
      ./node_modules/browserify-sign/browser/curves.json 162 bytes [built] [code generated]
    + 4 modules
modules by path ./lib/ 638 KiB 145 modules
+ 11 modules

WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
[...]

PR #11717, opened 2022-04-27
fix: fix browser build for Webpack 5

vkarpov15:
Re: #11584

**Summary**

With #11584, there were a few issues that popped up in the browser build:

1. Missing `Buffer`
2. Missing `process`
3. Looks like Babel was failing on `isAsyncFunction()`

**Examples**

### DIFF
--- a/lib/helpers/isAsyncFunction.js
+++ b/lib/helpers/isAsyncFunction.js
@@ -8,10 +8,15 @@ try {
   asyncFunctionPrototype = Object.getPrototypeOf(async function() {});
 } catch (err) {}
 
-module.exports = function isAsyncFunction(v) {
-  return (
-    asyncFunctionPrototype != null &&
-    typeof v === 'function' &&
-    Object.getPrototypeOf(v) === asyncFunctionPrototype
-  );
-};
+if (asyncFunctionPrototype == null) {
+  module.exports = function isAsyncFunction() {
+    return false;
+  };
+} else {
+  module.exports = function isAsyncFunction(v) {
+    return (
+      typeof v === 'function' &&
+      Object.getPrototypeOf(v) === asyncFunctionPrototype
+    );
+  };
+}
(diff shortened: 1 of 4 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

lib/helpers/isAsyncFunction.js
```diff
@@ -1,9 +1,16 @@
...
-const asyncFunctionPrototype = Object.getPrototypeOf(async function() {});
+let asyncFunctionPrototype = null;
+// try/catch for Babel compatibility, because Babel preset-env requires
+// regenerator-runtime for async/await and we don't want to include that
+// for a simple check.
+try {
+  asyncFunctionPrototype = Object.getPrototypeOf(async function() {});
+} catch (err) {}
 
 module.exports = function isAsyncFunction(v) {
```
Uzlopak, line 11: we could avoid a potential performance penalty by checking the value of asyncFunctionPrototype only once.

```javascript
module.exports =  asyncFunctionPrototype !== null
? function isAsyncFunction(v) {
  return (
    typeof v === 'function' &&
    Object.getPrototypeOf(v) === asyncFunctionPrototype
  );
}
: function () { return false; };
```

Uzlopak:
What is the actual purpose of the browser build?

vkarpov15:
@Uzlopak the browser build is supposed to let you run basic schema validation in the browser. The syntax is slightly different: there's no models, you need to define a schema when you create a document. But you can at least define and use Mongoose schemas in the browser. For example:

```html
<html>
  <head>
    <script src="/dist/browser.umd.js"></script>
    <script>
        quest: { type: String, match: /Holy Grail/i, required: true },
        favoriteColor: { type: String, enum: ['Red', 'Blue'], required: true }
      });

      doc = new mongoose.Document({
        name: 'Sir Lancelot of Camelot',
        quest: 'To seek the Holy Grail',
        favoriteColor: 'Blue'
      }, schema);
      console.log(doc.validateSync()); // null

      doc.set({
        name: 'Sir Galahad of Camelot',
        quest: 'To seek the Holy Grail',
        favoriteColor: 'Yellow'
      });

      console.log(doc.validateSync()); // ValidationError
    </script>
  </head>
  <body></body>
</html>
```

AbdelrahmanHafez:
I'm genuinely curious if anyone out there is relying on that.

PR #11749, opened 2022-05-01
chore: Drop Node 12 Support

Uzlopak:
Node 12 is EOL. We should remove support for it.

Do we have somewhere documented, which versions we support?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [12, 14, 16, 18]
+        node: [14, 16, 18]
         os: [ubuntu-18.04, ubuntu-20.04]
         include:
           - os: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
This is a breaking change so it needs to be in a major release.

I oppose this change for now, since maintaining Node 12 until v7.x shouldn't take any effort for us, and it'd be nice to support Node 12 for those who are still using it for whatever reason.

Uzlopak:
Yes, we should discuss this.

vkarpov15:
I agree with @AbdelrahmanHafez , we shouldn't drop support for Node 12 until Mongoose 7. That would be severely backwards breaking, and a significant departure from the way we've done things in the past. Remember, Mongoose supported Node 4 until August 2021. If maintaining Node 12 support doesn't cause us too much friction, we should keep it.

In the meantime, what do you think about creating a `7.x` branch and merging this into the `7.x` branch?

Uzlopak:
I dont think it is worth a stale branch :). We should first make plans where we want to move with mongoose :)

PR #11765, opened 2022-05-03
optimize size of browser bundle, use buffer v.5.7.1 package to match buffer package of mongodb in browser bundle

Uzlopak:
- use buffer v.5.7.1 as a devDependency so that the version matches the buffer package from mongodb.
- use alias in webpack on bn.js to reduce redundancy, saving about 260 kb

before:
Screenshot from 2022-05-03 23-10-04

after:
Screenshot from 2022-05-03 23-09-48

before:1048 kb
after: 781 kb

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "babel-loader": "8.2.5",
     "benchmark": "2.1.4",
     "bluebird": "3.7.2",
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

Issue #11761, opened 2022-05-03
LeanDocument Type doesn't include _id or id


noseworthy:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
I believe that the types defined here are incorrect:
https://github.com/Automattic/mongoose/blob/c22152c5a95c5ce9ed688d96b37b05c7df6be1ec/types/index.d.ts#L2261-L2267
This says that `LeanDocuments` (that is documents returned from `.toObject()` or when using the `lean` option don't have the `_id` or `id` fields. I don't think this is correct.

This issue only presented itself in version 6.3.2 and wasn't present in 6.3.1

**If the current behavior is a bug, please provide the steps to reproduce.**

```typescript
import mongoose from 'mongoose'

async function test() {
  const thingSchema = new mongoose.Schema({
    name: mongoose.Schema.Types.String,
  });

  const ThingModel = mongoose.model('Thing', thingSchema);

  // TypeScript will complain here that `_id` doesn't exist
  const { _id, ...thing1 } = (await ThingModel.create({name: 'Thing 1'})).toObject();

  // if I ignore the above warning and this log happens, _id is a valid ObjectId.
  console.log({_id, thing1});
}

```

**What is the expected behavior?**

`_id` and `id` are present on LeanDocuments (I think...)
**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Node.js: v16.15.0
TypeScript: v4.6.2
mongoose: v6.3.2

Uzlopak:
I think, that id is not on a lean document if you did not define id in the schema explicitly, as id is actually a virtual field. 

@taxilian

Uzlopak:
I guess you can avoid the bug for now, by explicitly defining _id: Schema.Types.ObjectId in your Schema.

noseworthy:
Yeah, I'm just `@ts-ignore`-ing the few errors we're getting for now. However, I don't feel like we should be required to define `_id` in the schema. The POJO that we get back has the `_id` on it regardless of the schema definition. I think the TypeScript type should reflect that.

That said, I don't have super deep understanding of the inner workings of all this and I could be missing something. 🤷‍♂️

It's why I raised a bug rather than open an MR. Didn't feel comfortable mucking about on this one, haha. Thanks for taking a look at it! 😁

taxilian:
[...]

PR #11767, opened 2022-05-04
Fix: Refactor LeanDocument type

mohammad0-0ahmad:
Related to https://github.com/Automattic/mongoose/issues/11761

### DIFF
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -97,7 +97,7 @@ async function gh11761() {
   {
     // make sure _id has been added to the type
     const { _id, ...thing1 } = (await ThingModel.create({ name: 'thing1' })).toObject();
-    expectType<Types.ObjectId | undefined>(_id);
+    expectType<Types.ObjectId>(_id);
 
     console.log({ _id, thing1 });
   }
@@ -131,10 +131,9 @@ async function gh11118(): Promise<void> {
   const docs = await UserModel.find().lean().exec();
 
   for (const doc of docs) {
-    const _id: Types.ObjectId | undefined = doc._id;
+    const _id: Types.ObjectId = doc._id;
   }
 }
-
 
   // Helpers to simplify checks
   type IfAny<IFTYPE, THENTYPE> = 0 extends (1 & IFTYPE) ? THENTYPE : IFTYPE;
@@ -269,9 +268,8 @@ async function _11761() {
 
   const ThingModel = model('Thing', thingSchema);
 
-  const { _id, name, id, __v } = (await ThingModel.create({ name: 'Thing 1' })).toObject();
+  const { _id, name, __v } = (await ThingModel.create({ name: 'Thing 1' })).toObject();
   expectType<string>(name);
-  expectType<string | undefined>(id);
-  expectType<Types.ObjectId | undefined>(_id);
-  expectType<number | undefined>(__v);
+  expectType<Types.ObjectId>(_id);
+  expectType<number>(__v);
 }
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -97,7 +97,7 @@ async function gh11761() {
   {
     // make sure _id has been added to the type
     const { _id, ...thing1 } = (await ThingModel.create({ name: 'thing1' })).toObject();
-    expectType<Types.ObjectId>(_id);
+    expectType<Types.ObjectId | undefined>(_id);
 
     console.log({ _id, thing1 });
   }
@@ -131,7 +131,7 @@ async function gh11118(): Promise<void> {
   const docs = await UserModel.find().lean().exec();
 
   for (const doc of docs) {
-    const _id: Types.ObjectId = doc._id;
+    const _id: Types.ObjectId | undefined = doc._id;
   }
 }
 
@@ -242,7 +242,9 @@ async function _11761() {
 
   const ThingModel = model('Thing', thingSchema);
 
-  const { _id, name, id } = (await ThingModel.create({ name: 'Thing 1' })).toObject();
+  const { _id, name, id, __v } = (await ThingModel.create({ name: 'Thing 1' })).toObject();
   expectType<string>(name);
   expectType<string | undefined>(id);
+  expectType<Types.ObjectId | undefined>(_id);
+  expectType<number | undefined>(__v);
 }
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -234,3 +234,15 @@ async function _11767() {
   // expectError(examFound2Obj.questions[0].populated);
   expectType<string[]>(examFound2Obj.questions[0].answers);
 }
+
+async function _11761() {
+  const thingSchema = new Schema<{ name: string }>({
+    name: Schema.Types.String
+  });
+
+  const ThingModel = model('Thing', thingSchema);
+
+  const { _id, name, id } = (await ThingModel.create({ name: 'Thing 1' })).toObject();
+  expectType<string>(name);
+  expectType<string | undefined>(id);
+}
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -1,5 +1,5 @@
-import { Schema, model, Document, LeanDocument, Types, BaseDocumentType, DocTypeFromUnion, DocTypeFromGeneric } from 'mongoose';
-import { expectError, expectNotType, expectType } from 'tsd';
+import { Schema, model, Document, LeanDocument, Types, DocTypeFromUnion, DocTypeFromGeneric, BaseDocumentType } from 'mongoose';
+import { expectError, expectType } from 'tsd';
 
 const schema: Schema = new Schema({ name: { type: 'String' } });
 
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -1,4 +1,4 @@
-import { Schema, model, Document, LeanDocument, Types, DocTypeFromUnion, DocTypeFromGeneric, BaseDocumentType } from 'mongoose';
+import { Schema, model, Document, LeanDocument, Types } from 'mongoose';
 import { expectError, expectType } from 'tsd';
 
 const schema: Schema = new Schema({ name: { type: 'String' } });
@@ -134,6 +134,33 @@ async function gh11118(): Promise<void> {
     const _id: Types.ObjectId | undefined = doc._id;
   }
 }
+
+
+  // Helpers to simplify checks
+  type IfAny<IFTYPE, THENTYPE> = 0 extends (1 & IFTYPE) ? THENTYPE : IFTYPE;
+  type IfUnknown<IFTYPE, THENTYPE> = unknown extends IFTYPE ? THENTYPE : IFTYPE;
+
+// tests for these two types are located in test/types/lean.test.ts
+export type DocTypeFromUnion<T> = T extends (Document<infer T1, infer T2, infer T3> & infer U) ?
+  [U] extends [Document<T1, T2, T3> & infer U] ? IfUnknown<IfAny<U, false>, false> : false : false;
+
+export type DocTypeFromGeneric<T> = T extends Document<infer IdType, infer TQueryHelpers, infer DocType> ?
+  IfUnknown<IfAny<DocType, false>, false> : false;
+
+/**
+   * Helper to choose the best option between two type helpers
+   */
+export type _pickObject<T1, T2, Fallback> = T1 extends false ? T2 extends false ? Fallback : T2 : T1;
+
+/**
+   * There may be a better way to do this, but the goal is to return the DocType if it can be infered
+   * and if not to return a type which is easily identified as "not valid" so we fall back to
+   * "strip out known things added by extending Document"
+   * There are three basic ways to mix in Document -- "Document & T", "Document<ObjId, mixins, T>",
+   * and "T extends Document". In the last case there is no type without Document mixins, so we can only
+   * strip things out. In the other two cases we can infer the type, so we should
+   */
+export type BaseDocumentType<T> = _pickObject<DocTypeFromUnion<T>, DocTypeFromGeneric<T>, false>;
 
 async function getBaseDocumentType(): Promise<void> {
   interface User {
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -5,13 +5,22 @@ declare module 'mongoose' {
   /** A list of paths to skip. If set, Mongoose will validate every modified path that is not in this list. */
   type pathsToSkip = string[] | string;
 
+  /**
+   * @summary Lean document base type.
+   */
   interface LeanDocumentBaseType {
     _id?: Types.ObjectId;
     id?: string;
     __v?: number;
   }
 
-  type ResolveLeanDocumentBaseType<T extends Document, DocType> = IfEquals<DocType, any> extends true ? T : PopulateAFromB<DocType, LeanDocumentBaseType>;
+  /**
+   * @summary Utility helps to prepare an accurate Document type.
+   * @description It merge the type of the Document and SchemaDefinition "raw document" & make sure to have the correct type of LeanDocumentBaseType.
+   * @param {T} T Instance of {@link Document} class.
+   * @param {RawDocType} RawDocType Schema's raw document type.
+   */
+  type ResolveDocumentBaseType<T, RawDocType> = IfEquals<RawDocType, any> extends true ? T : PopulateAFromB<RawDocType, LeanDocumentBaseType>;
 
   /**
    * Generic types for Document:
@@ -230,11 +239,11 @@ declare module 'mongoose' {
     set(value: any): this;
 
     /** The return value of this method is used in calls to JSON.stringify(doc). */
-    toJSON<T = LeanDocument<DocType>>(options?: ToObjectOptions & { flattenMaps?: true }): FlattenMaps<T>;
-    toJSON<T = LeanDocument<DocType>>(options: ToObjectOptions & { flattenMaps: false }): T;
+    toJSON<T = DocType>(options?: ToObjectOptions & { flattenMaps?: true }): FlattenMaps<LeanDocument<T>>;
+    toJSON<T = DocType>(options: ToObjectOptions & { flattenMaps: false }): LeanDocument<T>;
 
     /** Converts this document into a plain-old JavaScript object ([POJO](https://masteringjs.io/tutorials/fundamentals/pojo)). */
-    toObject<T = DocType, R = ResolveLeanDocumentBaseType<this, T>>(options?: ToObjectOptions): LeanDocument<R>;
+    toObject<T = DocType>(options?: ToObjectOptions): LeanDocument<ResolveDocumentBaseType<this, T>>;
 
     /** Clears the modified state on the specified path. */
     unmarkModified(path: string): void;
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -1,8 +1,17 @@
+/// <reference path="./utilities.d.ts" />
 declare module 'mongoose' {
   import mongodb = require('mongodb');
 
   /** A list of paths to skip. If set, Mongoose will validate every modified path that is not in this list. */
   type pathsToSkip = string[] | string;
+
+  interface LeanDocumentBaseType {
+    _id?: Types.ObjectId;
+    id?: string;
+    __v?: number;
+  }
+
+  type ResolveLeanDocumentBaseType<T extends Document, DocType> = IfEquals<DocType, any> extends true ? T : PopulateAFromB<DocType, LeanDocumentBaseType>;
 
   /**
    * Generic types for Document:
@@ -131,7 +140,7 @@ declare module 'mongoose' {
      */
     getChanges(): UpdateQuery<this>;
 
-    /** The string version of this documents _id. */
+    /** The string "by default" version of this documents _id. */
     id?: any;
 
     /** Signal that we desire an increment of this documents version. */
@@ -225,7 +234,7 @@ declare module 'mongoose' {
     toJSON<T = LeanDocument<DocType>>(options: ToObjectOptions & { flattenMaps: false }): T;
 
     /** Converts this document into a plain-old JavaScript object ([POJO](https://masteringjs.io/tutorials/fundamentals/pojo)). */
-    toObject<T = LeanDocument<DocType>>(options?: ToObjectOptions): Require_id<T>;
+    toObject<T = DocType, R = ResolveLeanDocumentBaseType<this, T>>(options?: ToObjectOptions): LeanDocument<R>;
 
     /** Clears the modified state on the specified path. */
     unmarkModified(path: string): void;
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -14,12 +14,27 @@ declare module 'mongoose' {
   }
 
   /**
-   * @summary Utility helps to prepare an accurate Document type.
-   * @description It merge the type of the Document and SchemaDefinition "raw document" & make sure to have the correct type of LeanDocumentBaseType.
-   * @param {T} T Instance of {@link Document} class.
-   * @param {RawDocType} RawDocType Schema's raw document type.
+   * @summary Helper that resolve the lean document returned by {@link Document.toObject} function.
+   * @param {T} Document Instance of {@link Document} class.
+   * @param {R} RawDocType Schema's raw document type.
+   * @param {O} Options of {@link Document.toObject}.
+   * @param {D} the default returned type. "Returns when {@link O Options} is not provided".
+   * @returns lean document.
    */
-  type ResolveDocumentBaseType<T, RawDocType> = IfEquals<RawDocType, any> extends true ? T : PopulateAFromB<RawDocType, LeanDocumentBaseType>;
+  type ResolveToObjectLeanDocument<T, R, O, D = LeanDocument<IfEquals<R, any> extends true ? T : R>> =/*  O extends ToObjectOptions ? ResolveLeanDocumentDependingOnToObjectOptions<D, O> : */ D;
+
+
+  /**
+   * @summary Helper that resolve the lean document returned by {@link Document.toObject} function.
+   * @param {T} LeanDocument of {@link Document} Instance.
+   * @param {O} Options is the first argument of {@link Document.toObject}.
+   * @returns lean document with some changes depending on {@link O Options} value.
+   */
+  type ResolveLeanDocumentDependingOnToObjectOptions<T, O extends ToObjectOptions> = {
+
+  };
+
+
 
   /**
    * Generic types for Document:
@@ -242,7 +257,7 @@ declare module 'mongoose' {
     toJSON<T = DocType>(options: ToObjectOptions & { flattenMaps: false }): LeanDocument<T>;
 
     /** Converts this document into a plain-old JavaScript object ([POJO](https://masteringjs.io/tutorials/fundamentals/pojo)). */
-    toObject<T = DocType>(options?: ToObjectOptions): LeanDocument<ResolveDocumentBaseType<this, T>>;
+    toObject<T = DocType, O extends ToObjectOptions = {}>(options?: O): ResolveToObjectLeanDocument<this, T, O>;
 
     /** Clears the modified state on the specified path. */
     unmarkModified(path: string): void;
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -9,9 +9,8 @@ declare module 'mongoose' {
    * @summary Lean document base type.
    */
   interface LeanDocumentBaseType {
-    _id?: Types.ObjectId;
-    id?: string;
-    __v?: number;
+    _id: Types.ObjectId;
+    __v: number;
   }
 
   /**
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -227,4 +227,5 @@ async function _11532() {
 
   if (!leanResult) return;
   expectType<string>(leanResult.child.name);
+  expectType<number>(leanResult?.__v);
 }
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -227,5 +227,4 @@ async function _11532() {
 
   if (!leanResult) return;
   expectType<string>(leanResult.child.name);
-  expectError(leanResult?.__v);
 }
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -48,6 +48,18 @@ interface ITest extends Document {
 
 const Test = model<ITest, Model<ITest, QueryHelpers>>('Test', schema);
 
+// Test subDoc id type overriding
+
+const testDoc = new Test({ docs: [{ id: 10 }] });
+
+if (testDoc?.docs?.[0]) {
+  const subDoc = testDoc?.docs[0];
+  expectType<number | undefined>(subDoc?.id);
+
+  const leanDoc = subDoc.toObject();
+  expectType<number | undefined>(leanDoc.id);
+}
+
 Test.find({}, {}, { populate: { path: 'child', model: ChildModel, match: true } }).exec().then((res: Array<ITest>) => console.log(res));
 
 Test.find().byName('test').byName('test2').orFail().exec().then(console.log);
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -55,9 +55,6 @@ const testDoc = new Test({ docs: [{ id: 10 }] });
 if (testDoc?.docs?.[0]) {
   const subDoc = testDoc?.docs[0];
   expectType<number | undefined>(subDoc?.id);
-
-  const leanDoc = subDoc.toObject();
-  expectType<number | undefined>(leanDoc.id);
 }
 
 Test.find({}, {}, { populate: { path: 'child', model: ChildModel, match: true } }).exec().then((res: Array<ITest>) => console.log(res));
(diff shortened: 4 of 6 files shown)


taxilian:
oh please please please don't do this. This does not fix leandocument -- it neuters it.

If you can give me some specific tests that you want to ensure work I will make sure they work *without* breaking everything that I spent days making sure worked. I promise there are reasons for all of the things that you just pulled out -- every single one of them. I would not have spent a week making sure they all worked otherwise.

mohammad0-0ahmad:
> 

Hello @taxilian !
Would you like to explain how does these changes "neuters" LeanDocument ?

taxilian:
So the goal of `LeanDocument` is to give you the document as it would be if you query it with `lean()` -- right? So here are the options:

1. Ideally we know the interface of the document that was placed inside -- when you grab it with `lean` then you should get that same document with no changes, unless some plugin or something has made those changes. The exceptions to that are that `_id` would be added if it wasn't overridden and `__v` would be added if it's not disabled. My personal argument is that at least `__v` should only be provided if the user wants to know about it, so they should be responsible for adding it to the interface. It has not ever been added by types before, so I don't think we should start now and set a bad pattern.  I also disagree with adding _id, but it's not my call so I'll add it in -- not that hard anyway.

2. If we can't figure out the interface of the underlying document (e.g. if they extend Document instead of using generics) then the best we can do is say "Take the interface of the full mongoose document and subtract everything that is added when it is hydrated" -- which is what the original code did by omitting everything except `_id`, `id`, and `__v`. I think `id` shouldn't have been there since AFAIK by default it wouldn't be there on a lean document, but I didn't write that code so I left it as it was.

What this code does is instead says "Hey, I have an interface that I'll strip some array types out of and otherwise leave it alone". It basically doesn't try to pull out any of the things that LeanDocument was intended to pull out and removes 90% of the utility.

I have a fix for the original issue in #11761 mostly done which *doesn't* remove all the functionality, but I get the feeling there are additional things we should consider so if I can get any other tests to fix at the same time I'm happy to look at those as well.

If there is a better place we can chat about this I'm happy to do so.

taxilian:
So honestly I wonder if the best / easiest way to address this is to just change the toObject return value like so: 

```
toObject<T = DocType>(options?: ToObjectOptions): Require_id<LeanDocument<T>>;
```

We could make LeanDocument always add a _id but in some cases it may be used on a subdocument which may not have a _id (I *always* turn them off in my documents as they add bloat and are almost never needed for my stuff, but I don't know how common that is). It seems like it might be better to only assume they will be there on the root document, so the easiest way would be to just modify lean() and toObject to use Require_id and add the _id if it isn't there.

@mohammad0-0ahmad  and/or @vkarpov15 thoughts?

Also, I don't understand why this would be new because the branch of types which I added isn't even being triggered for this test case..... but this is the test I'm using:

```
async function gh11761() {
  const thingSchema = new Schema<{ name: string }>({
    name: Schema.Types.String
  });

  const ThingModel = model('Thing', thingSchema);

  const { _id, ...thing1 } = (await ThingModel.create({ name: 'thing1' })).toObject();

  console.log({ _id, thing1 });
}
```

taxilian:
AHH! I see why it happened; because I changed the return type of `.toObject` because using `this` there broke a bunch of stuff. lean() already has support correctly... I'll have a proposed PR soon.

taxilian:
See #11769

mohammad0-0ahmad:
With all you said @taxilian, I don't really got the reason why mongoose must have the mentioned types.
IMHO, I think your implementation  increase the complexity of the code and make the TS server a bit slower without clear reason to do so.
As I said, keeping the deleted tests that you provided will still work in this PR by moving the unnecessary types to the test file, so It doesn't make sense "I think" to keep them.
Test this In my PR to double-check that.
```
// Helpers to simplify checks
type IfAny<IFTYPE, THENTYPE> = 0 extends (1 & IFTYPE) ? THENTYPE : IFTYPE;
type IfUnknown<IFTYPE, THENTYPE> = unknown extends IFTYPE ? THENTYPE : IFTYPE;

// tests for these two types are located in test/types/lean.test.ts
type DocTypeFromUnion<T> = T extends (Document<infer T1, infer T2, infer T3> & infer U) ?
  [U] extends [Document<T1, T2, T3> & infer U] ? IfUnknown<IfAny<U, false>, false> : false : false;

type DocTypeFromGeneric<T> = T extends Document<infer IdType, infer TQueryHelpers, infer DocType> ?
  IfUnknown<IfAny<DocType, false>, false> : false;

/**
 * Helper to choose the best option between two type helpers
 */
type _pickObject<T1, T2, Fallback> = T1 extends false ? T2 extends false ? Fallback : T2 : T1;

/**
 * There may be a better way to do this, but the goal is to return the DocType if it can be infered
 * and if not to return a type which is easily identified as "not valid" so we fall back to
 * "strip out known things added by extending Document"
 * There are three basic ways to mix in Document -- "Document & T", "Document<ObjId, mixins, T>",
 * and "T extends Document". In the last case there is no type without Document mixins, so we can only
 * strip things out. In the other two cases we can infer the type, so we should
 */
type BaseDocumentType<T> = _pickObject<DocTypeFromUnion<T>, DocTypeFromGeneric<T>, false>;
async function getBaseDocumentType(): Promise<void> {
  interface User {
    name: string;
    email: string;
    avatar?: string;
  }

  type UserDocUnion = User & Document<Types.ObjectId>;
  type UserDocGeneric = Document<Types.ObjectId, {}, User>;

  // DocTypeFromUnion should correctly infer the User type from our unioned type
  type fromUnion1 = DocTypeFromUnion<UserDocUnion>;
  expectType<User>({} as fromUnion1);
  // DocTypeFromUnion should give a "false" type if it isn't a unioned type
  type fromUnion2 = DocTypeFromUnion<UserDocGeneric>;
  expectType<false>({} as fromUnion2);
  // DocTypeFromUnion should give a "false" type of it's an any
  expectType<false>({} as DocTypeFromUnion<any>);

  // DocTypeFromGeneric should correctly infer the User type from our Generic constructed type
  type fromGeneric1 = DocTypeFromGeneric<UserDocGeneric>;
  expectType<User>({} as fromGeneric1);
  // DocTypeFromGeneric should give a "false" type if it's not a type made with Document<?, ?, DocType>
  type fromGeneric2 = DocTypeFromGeneric<UserDocUnion>;
  expectType<false>({} as fromGeneric2);
  // DocTypeFromGeneric should give a "false" type of it's an any
  expectType<false>({} as DocTypeFromGeneric<any>);

  type baseDocFromUnion = BaseDocumentType<UserDocUnion>;
  expectType<User>({} as baseDocFromUnion);

  type baseDocFromGeneric = BaseDocumentType<UserDocGeneric>;
  expectType<User>({} as baseDocFromGeneric);
}

async function getBaseDocumentTypeFromModel(): Promise<void> {
  interface User {
    name: string;
    email: string;
    avatar?: string;
  }
  const schema = new Schema<User>({});
  const Model = model('UserBaseDocTypeFromModel', schema);
  type UserDocType = InstanceType<typeof Model>;

  type baseFromUserDocType = BaseDocumentType<UserDocType>;

  expectType<User & { _id: Types.ObjectId }>({} as baseFromUserDocType);

  const a: UserDocType = {} as any;

  const b = a.toJSON();
}
```

mohammad0-0ahmad:
Take a look at this as well:
Capture

mohammad0-0ahmad:
I am leaving the final word for @vkarpov15  @Uzlopak to check both and merge the better one.

taxilian:
@mohammad0-0ahmad given that the tests that you pulled out all specifically test a particular aspect of LeanDocument, yes if you pull that aspect out and then test only those things then of course the tests will pass -- that's the whole point of unit tests is to break down a problem into smaller pieces so you can test that unit of code.

Just because you can test a unit of code separately does not mean that the unit of code is not needed! It also doesn't change the fact that even completely ignoring those tests this PR fundamentally breaks LeanDocument by no longer removing the things added by Document!

Since it is apparently necessary I will make some tests to help demonstrate -- though if you actually look at the type output it should be pretty clear the difference, which is significant.

(also, I apologize I'm dealing with a minor catastrophe at work right now so taking time to do this is frustrating me more than it actually should, which is definitely not your fault -- trying to compensate, but having these removed would be absolutely disasterous for my usage of mongoose, which is why I spent most of a week writing these improvements and testing them in the first place. I don't have time to do this right now, but I also can't afford to have this discarded because the purpose wasn't understood.)

mohammad0-0ahmad:
@taxilian as I said I've pulled out them because these tests needs the mentioned types.
Please provide which cases you are talking about that doesn't satisfy  this PR when you have time for that.

taxilian:
I do see that I misunderstood a few pieces of what is actually happening -- I'm still pretty sure that this will be a problem for my code, and it *definitely* results in less readable output in the IDE since it doesn't find its way back to the original types. I need to run this through all the tests with my code to see where it falls down which will take me a few hours -- but it's definitely fair for you to want me to check my assumptions, no matter how frustrated I am by other things or emotionally invested in my changes I may be =]

taxilian:
Yeah, with this change there are now 37 errors in my codebase all due to LeanDocument related types. Unfortunately none of them are easy to track down -- it may take me a few hours to distill this into an explainable case / unit test.

mohammad0-0ahmad:
@taxilian Can you just provide code snippet to one of them ?

taxilian:
The tricky bit is that I'm using an extra abstraction, so I actually define my schemas as a class with decorators -- it's a really nice system which lets me define things only once (instead of needing to define the types and schema separately) and also easily find the actual implementation of any bit of code using the browser. As nice as that is, I don't want to make compatibility with my abstraction the foundation of my argument for doing things a particular way if I can avoid it, so I want to try to reproduce this without my abstraction, but that requires a more contrived example, which is why it's going to take some time.

taxilian:
https://github.com/gradecam/mongoose-decorators-ts/tree/feature/mongoose6 - this is the project, but I am pretty confident that I can demonstrate the issues in a more specific way.

vkarpov15:
I'm still reviewing this PR and #11769, I just wanted to make a quick comment about `Require_id`:

The reason why we have `Require_id` is because `_id` is optional when creating a new document by default, `new Model({ foo: 'bar' })` is perfectly ok even though you don't specify `_id`. But Mongoose adds an `_id` by default. And, even if Mongoose didn't, the MongoDB server _would_ add an `_id`. So while you technically can create a document without an `_id` in Mongoose, any document you load from MongoDB will have an `_id` unless you explicitly project it out. Hence we typically add `Require_id` for query results.

So adding `Require_id` to the result of `toObject()` isn't a terrible workaround. I wouldn't say it lines up exactly with how we expect `Require_id` to be used, but I'm happy to go that approach if it unblocks #11761.

mohammad0-0ahmad:
> I'm still reviewing this PR and #11769, I just wanted to make a quick comment about `Require_id`:
> 
> The reason why we have `Require_id` is because `_id` is optional when creating a new document by default, `new Model({ foo: 'bar' })` is perfectly ok even though you don't specify `_id`. But Mongoose adds an `_id` by default. And, even if Mongoose didn't, the MongoDB server _would_ add an `_id`. So while you technically can create a document without an `_id` in Mongoose, any document you load from MongoDB will have an `_id` unless you explicitly project it out. Hence we typically add `Require_id` for query results.
> 
> So adding `Require_id` to the result of `toObject()` isn't a terrible workaround. I wouldn't say it lines up exactly with how we expect `Require_id` to be used, but I'm happy to go that approach if it unblocks #11761.

In this PR, I think we don't need to add Require_id to the result of toObject(),
Please compare this PR with https://github.com/Automattic/mongoose/pull/11769 and pick one of them to merge.

taxilian:
Okay, here is a simple test that shows one issue, though it's pretty contrived:

```typescript
interface ThingType {
  name: string;
}
const thingSchema = new Schema<ThingType>({
    name: Schema.Types.String
});
const ThingModel = model('Thing', thingSchema);
type ThingDocument = HydratedDocument<ThingType>;
// or
type ThingyDocument = InstanceType<typeof ThingModel>;

function AcceptLeanDoc(val: LeanDocument<ThingDocument>) {
    // This should fail but it thinks the type is valid
    val.populate('');
}

```

image

Your type isn't actually constructing a lean type, it's starting with a type that is already lean -- which is great when you can do that, but there are cases where you can't or it's impractical. At best you have an extra type you have to pass around everywhere.

types/document.d.ts
```diff
@@ -132,7 +132,7 @@ declare module 'mongoose' {
     getChanges(): UpdateQuery<this>;
 
     /** The string version of this documents _id. */
-    id?: any;
+    id?: string;
```
vkarpov15, line 135: This is a hard no. This would break anyone's code who uses `extends Document` with an `id` field that isn't a string.

taxilian:
So I guess there exists the possibility that you don't see it as a needed thing to use LeanDocument on a HydratedDocument type and have it create the result of .lean(), but this would definitely be a breaking change since previously it *did* attempt to do that and now it doesn't

types/document.d.ts
mohammad0-0ahmad: > Okay, here is a simple test that shows one issue, though it's pretty contrived:
> 
> ```ts
> interface ThingType {
>   name: string;
> }
> const thingSchema = new Schema<ThingType>({
>     name: Schema.Types.String
> });
> const ThingModel = model('Thing', thingSchema);
> type ThingDocument = HydratedDocument<ThingType>;
> // or
> type ThingyDocument = InstanceType<typeof ThingModel>;
> 
> function AcceptLeanDoc(val: LeanDocument<ThingDocument>) {
>     // This should fail but it thinks the type is valid
>     val.populate('');
> }
> ```
> 
> image
> 
> Your type isn't actually constructing a lean type, it's starting with a type that is already lean -- which is great when you can do that, but there are cases where you can't or it's impractical. At best you have an extra type you have to pass around everywhere.

I think this will be so instead:
```
async function _11762() {
  interface ThingType {
    name: string;
  }
  const thingSchema = new Schema<ThingType>({
    name: Schema.Types.String
  });
  const ThingModel = model('Thing', thingSchema);

  function AcceptLeanDoc(val: LeanDocument<ThingType>) {
    // This should fail but it thinks the type is valid
    val.populate('');
  }
}
```

mohammad0-0ahmad:
> Okay, here is a simple test that shows one issue, though it's pretty contrived:
> 
> ```ts
> interface ThingType {
>   name: string;
> }
> const thingSchema = new Schema<ThingType>({
>     name: Schema.Types.String
> });
> const ThingModel = model('Thing', thingSchema);
> type ThingDocument = HydratedDocument<ThingType>;
> // or
> type ThingyDocument = InstanceType<typeof ThingModel>;
> 
> function AcceptLeanDoc(val: LeanDocument<ThingDocument>) {
>     // This should fail but it thinks the type is valid
>     val.populate('');
> }
> ```
> 
> image
> 
> Your type isn't actually constructing a lean type, it's starting with a type that is already lean -- which is great when you can do that, but there are cases where you can't or it's impractical. At best you have an extra type you have to pass around everywhere.

I think this will be so instead:
```
async function _11762() {
  interface ThingType {
    name: string;
  }
  const thingSchema = new Schema<ThingType>({
    name: Schema.Types.String
  });
  const ThingModel = model('Thing', thingSchema);

  function AcceptLeanDoc(val: LeanDocument<ThingType>) {
    // This should fail but it thinks the type is valid
    val.populate('');
  }
}
```

types/document.d.ts
mohammad0-0ahmad: This would break anyone's code who uses extends Document with an id field that isn't a string.
But then we have to update docs because it says it will be string.
Or users can omit id type before in type overriding case.

taxilian:
Yes, you can do that -- it requires passing around an extra type, though, and it is definitely a breaking change. I don't know if anyone besides me is relying on it -- it would break all of my code, though I could probably work around it, it would just add a ton of complexity.

mohammad0-0ahmad:
I think Document third generic "DocType" is ThingType  interface in your example, which is the default value  for toObject Fn.
I might be wrong, but let's ask @vkarpov15  about that.

taxilian:
Just to make sure we are clear on what is a new change and what is not -- the parts I added which are new in 6.3.6 were the optimizations which would find the underlying type if available and then use that.  This is what was there before:

https://github.com/Automattic/mongoose/blob/6.3.0/types/index.d.ts#L2243

```
export type LeanDocument<T> = Omit<_LeanDocument<T>, Exclude<keyof Document, '_id' | 'id' | '__v'> | '$isSingleNested'>;
```

this PR says this: 
```
export type LeanDocument<T> = _LeanDocument<T> & Omit<Pick<Document, '_id' | 'id' | '__v'>, keyof T>;
```

Your PR is not just a fix -- it is fundamentally changing the way that LeanDocument works in such a way that it removes a feature which has been there for a number of versions. I do see what you are saying, and you're correct that `DocType` is `ThingType` in my example -- in fact that's exactly what the code you ripped out used to find `ThingType` -- but I want to be clear that this is actually a change.

While I disagree with it (because I rely on this behavior heavily myself) I can see your argument that you could get by without LeanDocument being able to work on a `mongoose.Document` instead of only on `DocType` -- there are definitely places where that would work and maybe an optimization could be made at the very least. That said, *this is a change* -- let's be really clear on that because you seem to be suggesting that it isn't.

taxilian:
Ahh -- and the bigger problem is subdocuments.

image

LeanDocument needs to also convert subdocuments to their underlying types, which are not known to the parent document.

taxilian:
Well, the good news is that I have an example now that shows what I'm talking about... the bad news is that my branch fails it partially as well, so I need to fix that! .toObject works, but .lean does not.

```
async function _11767() {
  interface Question {
    text: string;
    answers: Types.Array<string>;
    correct: number;
  }
  const QuestionSchema = new Schema<Question>({
    text: String,
    answers: [String],
    correct: Number
  });
  interface Exam {
    element: string;
    dateTaken: Date;
    questions: Types.DocumentArray<Question>;
  }
  const ExamSchema = new Schema<Exam>({
    element: String,
    dateTaken: Date,
    questions: [QuestionSchema]
  });

  const ExamModel = model<Exam>('Exam', ExamSchema);

  const examFound = await ExamModel.findOne().lean().exec();
  if (!examFound) return;

  // $pop shouldn't be there, because questions should no longer be a mongoose array
  expectError(examFound.questions.$pop());
  // popoulated shouldn't be on the question doc because it shouldn't
  // be a mongoose subdocument anymore
  expectError(examFound.questions[0]!.populated('answers'));

  const examFound2 = await ExamModel.findOne().exec();
  if (!examFound2) return;
  const examFound2Obj = examFound2.toObject();

  expectError(examFound2Obj.questions.$pop());
  expectError(examFound2Obj.questions[0].populated('answers'));
}
```

mohammad0-0ahmad:
@taxilian 
Capture

mohammad0-0ahmad:
> Well, the good news is that I have an example now that shows what I'm talking about... the bad news is that my branch fails it partially as well, so I need to fix that! .toObject works, but .lean does not.
> 
> ```ts
> async function _11767() {
>   interface Question {
>     text: string;
>     answers: Types.Array<string>;
>     correct: number;
>   }
>   const QuestionSchema = new Schema<Question>({
>     text: String,
>     answers: [String],
>     correct: Number
>   });
>   interface Exam {
>     element: string;
>     dateTaken: Date;
>     questions: Types.DocumentArray<Question>;
>   }
>   const ExamSchema = new Schema<Exam>({
>     element: String,
>     dateTaken: Date,
>     questions: [QuestionSchema]
>   });
> 
>   const ExamModel = model<Exam>('Exam', ExamSchema);
> 
>   const examFound = await ExamModel.findOne().lean().exec();
>   if (!examFound) return;
> 
>   // $pop shouldn't be there, because questions should no longer be a mongoose array
>   expectError(examFound.questions.$pop());
>   // popoulated shouldn't be on the question doc because it shouldn't
>   // be a mongoose subdocument anymore
>   expectError(examFound.questions[0]!.populated('answers'));
> 
>   const examFound2 = await ExamModel.findOne().exec();
>   if (!examFound2) return;
>   const examFound2Obj = examFound2.toObject();
> 
>   expectError(examFound2Obj.questions.$pop());
>   expectError(examFound2Obj.questions[0].populated('answers'));
> }
> ```

I will check that tomorrow, I don't have more energy to do that today ;)

taxilian:
Yeah I lose track of the specifics of how subdocument works, but nested arrays etc are still a real thing and it still gets converted to documents and needs to be converted back for lean to work correctly.

taxilian:
My suggestion would be to use the simple fix that I proposed which is the same as 6.3.6 but fixes this one issue and then we can rehash if your change makes sense. @vkarpov15  can of course decide =]

mohammad0-0ahmad:
> Yeah I lose track of the specifics of how subdocument works, but nested arrays etc are still a real thing and it still gets converted to documents and needs to be converted back for lean to work correctly.
Untitled

taxilian:
> Yeah I lose track of the specifics of how subdocument works, but nested arrays etc are still a real thing and it still gets converted to documents and needs to be converted back for lean to work correctly.
> Untitled

You haven't defined the type for that correctly; it would be `mongoose.Types.DocumentArray<{noise:string}>`. As you have it you don't get the $pop, etc functions. Also, if you do want to pretend it isn't ever a mongoose type (which means not having proper types for those) then it should be `Array<{noise: string}>`

mohammad0-0ahmad:
https://mongoosejs.com/docs/subdocs.html
Untitled

taxilian:
> https://mongoosejs.com/docs/subdocs.html
> 
> Untitled

Yes, that's the correct syntax for the Schema -- I'm talking about the typescript types -- the interface.

mohammad0-0ahmad:
@taxilian You are right sorry I had incorrect code example above, but your assumption still incorrect, This PR still working with Types.DocumentArray as I can see.
https://mongoosejs.com/docs/typescript/subdocuments.html#subdocument-arrays.
Capture

types/document.d.ts
vkarpov15: You can do new Schema({ id: Number }), that's perfectly valid in Mongoose. Mongoose just overrides the default id virtual.

vkarpov15:
I'm going to close this one in favor of #11769. I think that is the less breaking approach - inasmuch as we still support `LeanDocument`, we should avoid breaking it as much as possible.

In the future, I'd like to remove `LeanDocument` because there's simply too many edge cases for us to reliably pull out the raw document interface from a type that `extends Document` or `& Document`. For example, the `id: Number` case I mentioned here: https://github.com/Automattic/mongoose/pull/11767#discussion_r865464072. I think taking in the raw document interface and creating a hydrated document interface from that would be more reliable.

Thanks for the fast PR @mohammad0-0ahmad , I appreciate it given that the amount of discussion the bug in #11761 caused.

mohammad0-0ahmad:
Can you please reopen it, I've fixed the mentioned regression.

mohammad0-0ahmad:
thanks @AbdelrahmanHafez !

AbdelrahmanHafez:
Happy to help @mohammad0-0ahmad

mohammad0-0ahmad:
@vkarpov15  I am testing TS performance locally, and I got the following result,
But in GitHub it shows big difference  in check time and total time values.
Capture

Do you have any idea why I got that? I am wondering if this workflow is returns consistent benchmark results.

github

taxilian:
It's worth noting that even as relatively good as this benchmark is (and I found it super helpful when I was working on things) the results are really inconsistent across different systems with different loads. Just on my laptop I got significantly different numbers when I ran things e.g. 30 minutes later, even if it was the same thing. Presumably this is due to system load. Sometimes the difference would be negligible and sometimes significant -- I started looking at the complexity numbers more than the time because the time simply wasn't reliable enough across different runs.

mohammad0-0ahmad:
@taxilian  Would you like to review the new changes ?

taxilian:
> @taxilian Would you like to review the new changes ?
> I am working on options as well to make type more consistent.

At first glance I'm completely baffled by a number of things, so I'm going to have to dig into it and I don't know if I can do it today. The first thing that I'm really confused about, though, is why you would move all the stuff you pulled out (the stuff that my code all needed, but you say isn't needed) into a unit test -- if it's not needed then it's not needed. If it is needed then it is needed; I can't imagine any situation where having it in a unit test would do anyone any good -- though if I can find some time to dig into it more deeply I may find something I'm missing.

taxilian:
So I really don't understand what you are trying to fix with the updated changes.

You are fundamentally changing what LeanDocument does. I get that you don't think that it needs to do what it does -- that's a totally valid argument and discussion which can be had, since you obviously don't need it to do what it does. I however *do* need it to work the way that it has worked, though ideally it should work actually a little bit better than that (hence my changes)

This PR, however, *fundamentally changes what LeanDocument does* -- this isn't something that should be done in a patch release. What problem are you solving that isn't solved by the PR that he merged yesterday?

taxilian:
Also, I did actually try the branch -- and all my stuff is broken exactly as expected because nested documents, document arrays, etc are all not correctly translated. It's possible that I could restructure my types in such a way that they would be, but that's the point -- this is a breaking change.

Uzlopak:
Look guys. I think I have to make the unpopular decision and ask you to stop on this topic and end this "conflict".

LeanDocument got patched. For now we have fixed the issue. Maybe not the best way to do it, but it works now. Like the saying: "a solution is a temporary fix which becomes permanent". I assume you, @mohammad0-0ahmad are eager to simplify the typings, that is fine. But not now is not the time. We have no time pressure and we can be sure that with the next release that bug is not shipped.

I can guarantee you, that @vkarpov15 will not merge it as there is already no confidence that it wont break anything.

So, my decision is: I close this PR and propose another path to solve this: 

@taxilian 

I ask you to add more typings tests and cover all edge cases if necessary and provide a PR. Only test files for typings and if you find small bugs and you can fix them, than fine. But not more big typing changes regarding the LeanDocument topic.
Maybe we need to avoid testing the LeanDocument as expectType. This means: no "expectType<LeanDocument<T>>(bla)" but "expectType<{ name: string }>(bla). So if we remove one day LeanDocument we wont need to rewrite the tests. So I assume that there is some work to do.

@mohammad0-0ahmad 

When taxilian provided the typings tests, or says they are complete, we will review the PR. Like they say regarding TDD: "First measure and then cut." You are then free to provide a PR with refactored typings and as we already defined the tests, we would not need to touch the tests. So we can assure that there is no regression. 

@taxilian 
@mohammad0-0ahmad 
I think you are both equally good devs. And that I propose this path to solve this issue and conflict,does not mean that you can not collaborate. On the contrary: I encourage you to work together. E.g. it could happen, that @taxilian finds some big bug while writing the typings tests, which needs alot of time to fix it with the current LeanDocument style. Then we could store the failing typings test on a topic branch in this repo and then you could work together to fix it by doing the big bugfix together.

I hope you stay motivated and dont take all this personal.

Best Regards
Aras aka uzlopak

taxilian:
Very much agreed -- and please let me reiterate that I have no hard feelings towards anyone here, and I apologize for any "harsh tone" which I've probably used despite my best efforts -- it's mostly from fighting a losing battle with my ceph cluster all week.

@Uzlopak the PR which was merged by @vkarpov15 already has all of my things working, so I don't have any currently broken things that I need to fix, just things which *this* PR would have broken.

@mohammad0-0ahmad I actually (somewhat grudgingly ;-]) like the direction you're going with this, I was just concerned about trying to do it here in what I have perceived as being something intended to merge into 6.3.x. I would be happy to continue a discussion on a PR with longer term plans where we can try to hammer out some better type tests -- which I think we *all* can agree needs to happen. I really like what I've seen so far on the automatic type stuff, just trying to figure out how I can mesh it with how I'm doing things. *please* feel free to loop me in on anything you'd like reviewed for future stuff and we can go back and forth as needed in a less stressful way =]

Again, sorry my tone wasn't always great and for any disrespect that I conveyed, it was not intentional.

mohammad0-0ahmad:
That is really disappointing that you are closing this PR again in that way.
I don't say that you have to merge it right now, but I would like to try to get better code and more reliable intellisense to get better UX and don't have backward breaking changes.

Firstly, I really need you @taxilian to understand that I am not targeting your code because it is yours, but I am just make things better and more straight forward,
Yesterday before merging that ANOTHER branch, THIS branch had some break changes like rely on only schema definition because I didn't know that mongoose allows extending Document in that way without passing the schema definition type as generic.
Therefore, THIS PR had this regression.
Just to clarify, you don't really need to be so confused that way about moving the mentioned TS helper to test file.
Simply, the current tests are relying on these types, but mongoose in THIS PR require them to return lean document. But I would like to avoid touching current tests to avoid having regressions. I hope that you are understanding right now why I kept them in test file.

Secondly,  please don't state something without be sure about it. You are saying that I am fundamentally changing what Lean Document does "which might possible", but please don't state that before reviewing the code. And it works, like it worked, "I believe". I am really know that I can't just  say  "it doesn't need to do that because I've never needed to do that". And I would never like to make a breaking change in types as a bug-fix or waste someone else efforts or time.
When you are argument that something, do some breaking change and don't cover all use cases, please mention that instead of just saying that a thing make  fundamentally changes.
And I don't really agree with that should be in a patch release if It covers all current  allowed use cases.
You say that all your stuff is broken, which stuff you means if all tests in test file that you wrote by self is passed, sorry, but I don't really like that way with augmenting I would prefer to show that case that is meant.

@Uzlopak  with all my respect, but I don't take it as a conflict and as I said my only goal is better, simple code and more reliable intellisense even if @vkarpov15  would not like to merge it now, It can still as suggestion that can be discussed to get possibly better result in the feature.
Sorry If I am pressured someone here, but I just mentioned @taxilian  to review the code when he has time even if this PR will not be merged It might be helpful to get new ideas to improve the current code without need to merge this one. So please don't consider that as conflict or computation to show who is better dev,  just consider it  subjective discussion and do not take it personally.

Talking about some new unit tests, I would like to encourage you to create tests to checks if __v is persistence in lean document and think about what should happening when supplying a generic to lean function, right now it doesn't add _id by doing so.
 
Untitled

It might be also fantastic if we can get consistence lean document type when customizing toObject options argument.

At the end I do really apologize @taxilian  @vkarpov15  @Uzlopak  If you feel that I waste your time, or I waste your efforts  @taxilian that you spent days for making sure it will work. And I will not be working on this one anymore for now, like you wish.

Best regards!

taxilian:
I'll see if I can take some more time after I fix my other crisis to look over this, but I was not stating something without checking:

Previously with LeanDocument you could pass it a HydratedDocument type. With your change you cannot. I did review the code enough to be sure that was still the case, at least in every test that I did (though admittedly I only ran through things for about 10 minutes to see if I was missing anything, which was not thorough).

That is specifically the breaking change I was referring to -- you could pass in a HydratedDocument and get the base type out and in this PR you cannot.

Uzlopak:
First of all, it was not about me feeling wasting my time or so. I was under the impression, that this gets a heated discussion. I wanted to moderate and avoid getting it more heated.

Because you both confirmed that this is not a heated discussion and claim to want to work further on this I assume there is no harm to reopen it. 

Keep up the great work. :)

mohammad0-0ahmad:
> I'll see if I can take some more time after I fix my other crisis to look over this, but I was not stating something without checking:
> 
> Previously with LeanDocument you could pass it a HydratedDocument type. With your change you cannot. I did review the code enough to be sure that was still the case, at least in every test that I did (though admittedly I only ran through things for about 10 minutes to see if I was missing anything, which was not thorough).
> 
> That is specifically the breaking change I was referring to -- you could pass in a HydratedDocument and get the base type out and in this PR you cannot.

Alright, I've not noticed that, but that is great you could find that breaking change.
Send a code snippet for a test that shows this case you mentioned when you have time for that.

types/document.d.ts
```diff
@@ -1,9 +1,41 @@
...
   import mongodb = require('mongodb');
 
   /** A list of paths to skip. If set, Mongoose will validate every modified path that is not in this list. */
   type pathsToSkip = string[] | string;
 
+  /**
+   * @summary Lean document base type.
+   */
+  interface LeanDocumentBaseType {
+    _id: Types.ObjectId;
```
vkarpov15, line 12: This isn't sufficiently flexible. Not all lean documents have `_id` as an ObjectId. Maybe `interface LeanDocumentBaseType<IdType = Types.ObjectId> { _id: IdType; }` ?

types/document.d.ts
```diff
@@ -221,11 +253,11 @@ declare module 'mongoose' {
     set(value: any): this;
 
     /** The return value of this method is used in calls to JSON.stringify(doc). */
-    toJSON<T = LeanDocument<DocType>>(options?: ToObjectOptions & { flattenMaps?: true }): FlattenMaps<T>;
-    toJSON<T = LeanDocument<DocType>>(options: ToObjectOptions & { flattenMaps: false }): T;
+    toJSON<T = DocType>(options?: ToObjectOptions & { flattenMaps?: true }): FlattenMaps<LeanDocument<T>>;
```
vkarpov15, line 256: I'd rather keep `toJSON<T = LeanDocument<T>>` to allow opting out in cases where `LeanDocument` doesn't quite work with how people wrote their code

test/types/lean.test.ts
```diff
@@ -135,6 +135,32 @@ async function gh11118(): Promise<void> {
   }
 }
 
+  // Helpers to simplify checks
+  type IfAny<IFTYPE, THENTYPE> = 0 extends (1 & IFTYPE) ? THENTYPE : IFTYPE;
+  type IfUnknown<IFTYPE, THENTYPE> = unknown extends IFTYPE ? THENTYPE : IFTYPE;
+
+// tests for these two types are located in test/types/lean.test.ts
+export type DocTypeFromUnion<T> = T extends (Document<infer T1, infer T2, infer T3> & infer U) ?
```
vkarpov15, line 143: Are these helpers necessary? Don't seem used.

mohammad0-0ahmad:
> Can you please fix the merge conflicts? How is the state of this PR?

I was waiting for @taxilian  response, but never mind, I will close this PR.

PR #11769, opened 2022-05-04
Feature/ts lean missing doc fixes #11761 without killing LeanDocument

### LINKED ISSUE #11761: LeanDocument Type doesn't include _id or id

taxilian:
**Summary**

Per #11761 `.toObject` doesn't add `_id` to the document if it is missing; this fixes that.

**Examples**

See unit tests in the code

### DIFF
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -225,7 +225,7 @@ declare module 'mongoose' {
     toJSON<T = DocType>(options: ToObjectOptions & { flattenMaps: false }): LeanDocument<T>;
 
     /** Converts this document into a plain-old JavaScript object ([POJO](https://masteringjs.io/tutorials/fundamentals/pojo)). */
-    toObject<T = DocType>(options?: ToObjectOptions): LeanDocument<T>;
+    toObject<T = DocType>(options?: ToObjectOptions): Require_id<LeanDocument<T>>;
 
     /** Clears the modified state on the specified path. */
     unmarkModified(path: string): void;
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -221,11 +221,11 @@ declare module 'mongoose' {
     set(value: any): this;
 
     /** The return value of this method is used in calls to JSON.stringify(doc). */
-    toJSON<T = DocType>(options?: ToObjectOptions & { flattenMaps?: true }): FlattenMaps<LeanDocument<T>>;
-    toJSON<T = DocType>(options: ToObjectOptions & { flattenMaps: false }): LeanDocument<T>;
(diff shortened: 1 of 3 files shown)


types/document.d.ts
```diff
@@ -225,7 +225,7 @@ declare module 'mongoose' {
     toJSON<T = DocType>(options: ToObjectOptions & { flattenMaps: false }): LeanDocument<T>;
 
     /** Converts this document into a plain-old JavaScript object ([POJO](https://masteringjs.io/tutorials/fundamentals/pojo)). */
-    toObject<T = DocType>(options?: ToObjectOptions): LeanDocument<T>;
+    toObject<T = DocType>(options?: ToObjectOptions): Require_id<LeanDocument<T>>;
```
vkarpov15, line 228: In the interest of backwards compat, why not make this just `toObject<T = LeanDocument<DocType>>` and return `Require_id<T>`? That should allow people to bypass going through `LeanDocument`, which seems to be causing part of the problem.

vkarpov15 (commented):
A quick suggestion to hopefully make changes backwards compatible.

types/document.d.ts
taxilian: Definitely an option; by doing that you will not be converting any mongoose.Types.Array types to real javascript arrays and any embedded subdocuments won't get converted to POJOs.

taxilian:
@vkarpov15 I added one more quick change -- .lean was not actually applying LeanDocument, which resulted in lean not actually propagating and fixing things like the arrays, etc

vkarpov15 (approved):

PR #11778, opened 2022-05-06
Lock threads after 90 days inactivity

Uzlopak:
This workflow will lock threads after 90 days of inactivity We already autoclose threads after 19 days. So this iavoids that somebody is posting in very old issues or pull_requests.

We could actually integrate it into the stale workflow file.

### DIFF
--- a/.github/workflows/lock-threads.yml
+++ b/.github/workflows/lock-threads.yml
@@ -1,0 +1,29 @@
+name: 'Lock Threads'
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+permissions:
+  issues: write
+  pull-requests: write
+
+concurrency:
+  group: lock
+
+jobs:
+  action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v3
+        with:
+          issue-inactive-days: '90'
+          issue-comment: >
+            This issue has been automatically locked since there
+            has not been any recent activity after it was closed.
+            Please open a new issue for related bugs.
+          pr-comment: >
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez:
Thanks a lot for the PR, appreciate the effort.

I think that locking threads older than 90 days limits people who might be able to add to the topic, or even want to contribute to the repository by implementing a feature that's not a high priority for us, but is for them for some reason.

I'd say that it's quite rare that someone comments on an old issue, and if it reoccurs multiple times we can always manually lock it.

What do you think? @Uzlopak

vkarpov15:
I'm also somewhat hesitant to merge this. Admittedly it is annoying when people post "me too why this closed?" on an 8 year old issue that's marked "can't reproduce". But occasionally people post helpful comments on long-closed issues. And although I do sometimes lock old issues that get a little too many comments, I'd rather not make that something we do automatically.

What do you think @Uzlopak ?

Uzlopak:
Imho I think that peole should be encouraged to open new issues. I find it aweful that people revive old issues and write: "I have that issue too" and then you see it is in an 4 year old already closed issue. We could set the date to the date when mongoose 6 was released. Also this workflow should only affect closed issues. So active discussions are not affected. And we could set that specific labels are excemptet, like "discussion". So issues where we labeled as "help" can not hijacked.

vkarpov15:
It is definitely annoying when people make "me too" comments on long-closed issues. However, in my experience that is relatively infrequent and easily ignored. I don't want to block people from making helpful comments on old issues by default. And I'm not sure our current labeling system is able to capture the distinction between issues that are "me too" magnets vs issues that can benefit from additional activity.

Issue #11698, opened 2022-04-21
BUG - Subdocument mapping that has `select: false` results in MongoDB Server Error


Refzlund:
# Bug Report

~~Before creating an issue please make sure you are using the latest version of mongoose~~
~~Getting a `Operaion ... timed out after 10000ms` error? Read this FAQ entry: https://mongoosejs.com/docs/faq.html#operation-buffering-timed-out~~
~~Getting a `x.$__y is not a function` error? Read this FAQ entry: https://mongoosejs.com/docs/faq.html#not-a-function~~

### What is the current behavior?

When a **document** maps a **subdocument** — [Map](https://mongoosejs.com/docs/schematypes.html#maps)`({ field: 'hello', number: 0 }, { field: 'world', number: 1 } )`

The subdocument contains a property with `select: false` for ex. `number` — `Map({ field: 'hello' }, { field: 'world' } )`

Mongoose will query:
```ts
document.findOne { _id: new ObjectId("623ec001e3d4002a123561e7") } {
  projection: {
    'subdocuments.$*.number': 0
  }
}
```
Resulting in a MongoDB error response:
`FieldPath field names may not start with '$'. Consider using $getField or $setField.`

### Steps to reproduce
```ts
const subdocument = new Schema({
	selected: { type: Number }, 
	not_selected: { type: Number, select: false }
})

const document = new Schema({
	subdocument_mapping: {
		type: Map,
		of: subdocument
	}
})

const Document = mongoose.model('User', document)
mongoose.set('debug', true)
mongoose.set('debug', function (coll, method, query, doc) {
	console.log(`${coll}.${method}`, query, doc)
})

async function test() {
	await Document.find({})
}
```

#### Console log
```ts
users.find {} { projection: { 'subdocument_mapping.$*.not_selected': 0 } }
FieldPath field names may not start with '$'. Consider using $getField or $setField.
MongoServerError: FieldPath field names may not start with '$'. Consider using $getField or $setField.
    at Connection.onMessage (C:\Temp\Mongoose bug\node_modules\mongodb\lib\cmap\connection.js:203:30)
    at MessageStream.<anonymous> (C:\Temp\Mongoose bug\node_modules\mongodb\lib\cmap\connection.js:63:60)
    at MessageStream.emit (node:events:526:28)
    at processIncomingData (C:\Temp\Mongoose bug\node_modules\mongodb\lib\cmap\message_stream.js:108:16)
[...]

PR #11785, opened 2022-05-09
Gh-11698, Prevents .$* from being put into projections by avoiding drilling down into maps

IslandRhythms:


vkarpov15 (approved):
Thanks :+1:

PR #11786, opened 2022-05-09
fix vulnerability in example in redis todoRouter

IslandRhythms:


IslandRhythms:
Very strange, getting a connection refused from redis.

AbdelrahmanHafez (approved):
LGTM, thanks 👍

AbdelrahmanHafez:
@IslandRhythms 
What do you think of using [ioredis](https://www.npmjs.com/package/ioredis) in our examples?

When I needed to use redis with node, I found that ioredis (at the time) had much better capabilities with better promises support.

vkarpov15 (approved):

vkarpov15:
@AbdelrahmanHafez I've never used ioredis, but I don't use redis much :) Right now regular redis seems to support async/await just fine, so I don't think it's worth changing.

AbdelrahmanHafez:
@vkarpov15 
If node-redis support async/await, I agree that it's not worth changing. 👍
Glad that people are jumping on the async/await bandwagon, I assumed node-redis team was planning to take the [passport stance](https://github.com/jaredhanson/passport/issues/536) because node-redis only supported that recently, I remember doing that search sometime in 2021.

Issue #11580, opened 2022-03-25
Remove `on` from schema reserved keys


AbdelrahmanHafez:
We need to verify that people can use `on` as a path in their schemas.

vkarpov15:
I confirmed that you can use `on` as a schema path. You get a warning that you can suppress using the `supressReservedKeysWarning` option.

As far as Mongoose itself is concerned, we don't need to warn about `on` as a reserved key. Documents support `$on` as an alternative, and Mongoose uses `$on` internally in the [couple of places where we use documents as event emitters internally](https://github.com/Automattic/mongoose/blob/0ecae5cec34064224f57400159c51f994ce32af7/lib/types/subdocument.js#L408-L409). My only concern is plugins that rely on `.on()`.

@AbdelrahmanHafez what do you think about removing `on` from reserved keys in a minor release?

AbdelrahmanHafez:
Nice.
Sounds good, there should be no risk in releasing this in a minor release. I'm anticipating that in a couple of years plugins will break due to those changes, but then again the fix is easy on the plugins' end. 👍

PR #11804, opened 2022-05-13
on is no longer a reserved word gh-11580

IslandRhythms:


AbdelrahmanHafez:
@IslandRhythms 
Even if users can use it now, should we remove the warning functionality?
Keeping it under `reserved` allows us to log a warning when a person uses it, which as far as I understand is still the desired behavior.

AbdelrahmanHafez (approved):

vkarpov15 (approved):
Thanks :+1:

PR #11813, opened 2022-05-17
test: 💍 Add missing test for type of Connection.deleteModel()

tkowpt:
Following the advice of github user Uzlopak to add a test, that
Connection.deleteModel is accepting a RegExp as well.

**Summary**

Yesterday, I have submitted a fix for the types of Connection.deleteModel(), because the javascript code accepts a RegExp and not just a string. The fix is already merged, but a reviewer asked me to add a test and gave me an advice how to do it.

**Examples**

This was exactly the advice how to add a test:

`expectType<Connection>(conn.deleteModel(/.+/));`


Uzlopak (approved):
Thanks

PR #11814, opened 2022-05-17
Gh 10423 adds a transform option to lean queries

IslandRhythms:


lib/query.js
```diff
@@ -4007,12 +4013,83 @@ Query.prototype._findAndModify = function(type, callback) {
...
+      if (Array.isArray(doc[key])) {
+        doc[key].forEach((item) => {
+          for (const k in item) {
+            if (typeof item[k] === 'object' && item[k] != null && Object.keys(item[k]).length) {
+              _completeOneLean(item[k], res, opts);
+            }
+            opts.lean.transform({}, item);
+          }
+        });
+      } else if (typeof doc[key] === 'object' && doc[key] != null && Object.keys(doc[key]).length) {
```
vkarpov15, line 4027: This won't work, because we don't want to call `completeOneLean()` on nested paths as opposed to subdocuments.

```javascript
new Schema({
  nestedPath: { prop: String }, // <-- nested path, `nestedPath` isn't a subdocument
  subdocument: new Schema({ prop: String }) // <-- `subdocument` is a subdocument
});
```

lib/query.js
```diff
@@ -4007,12 +4013,83 @@ Query.prototype._findAndModify = function(type, callback) {
...
+        _completeOneLean(doc[key], res, opts);
+      }
+    }
+    opts.lean.transform({}, doc);
```
vkarpov15, line 4031: Should do `doc = opts.lean.transform({}, doc)`

lib/query.js
```diff
@@ -4007,12 +4013,83 @@ Query.prototype._findAndModify = function(type, callback) {
...
 function _completeOneLean(doc, res, opts, callback) {
+  if (opts.lean && opts.lean.transform) {
+    for (const key of Object.keys(doc)) {
+      if (Array.isArray(doc[key])) {
+        doc[key].forEach((item) => {
+          for (const k in item) {
+            if (typeof item[k] === 'object' && item[k] != null && Object.keys(item[k]).length) {
+              _completeOneLean(item[k], res, opts);
+            }
+            opts.lean.transform({}, item);
```
vkarpov15, line 4024: Let's change the `transform()` function so it only takes 1 parameter, `doc`. The first param is always `{}`, which isn't very useful.

lib/query.js
```diff
@@ -4007,12 +4013,83 @@ Query.prototype._findAndModify = function(type, callback) {
...
 }
 
+/*!
+ * ignore
+ */
+
+function _completeManyLean(model, docs, fields, userProvidedFields, opts, callback) {
+  if (opts.lean && opts.lean.transform) {
+    if (Array.isArray(docs)) {
+      docs.forEach(doc => {
```
vkarpov15, line 4051: I'd really recommend just lifting the logic from https://github.com/mongoosejs/mongoose-lean-getters/blob/master/index.js . That's more tested and handles nuances like subdocument vs nested path and field projections, which this logic doesn't handle. It also avoids adding extra overhead of iterating through the whole document as opposed to just subdocs.

vkarpov15 (changes requested):
Good job getting this working. I'm a little bummed that Mongoose doesn't handle calling `completeOneLean()` recursively already. But given that we need to do that, you should iterate through the schema's `childSchemas` to get subdocuments, rather than iterating through the entire object and assuming that any sub-objects you find are subdocuments.

vkarpov15 (approved):
Good work :+1:

PR #11817, opened 2022-05-18
Fix codeql sec warnings

Uzlopak:
I checked the codeql warnings. We had three false positives, which i dismissed. I hope this will fix the valid positive security warnings.

### DIFF
--- a/lib/helpers/path/setDottedPath.js
+++ b/lib/helpers/path/setDottedPath.js
@@ -1,23 +1,25 @@
 'use strict';
 
+const specialProperties = require('../specialProperties');
+
+
 module.exports = function setDottedPath(obj, path, val) {
   if (path.indexOf('.') === -1) {
+    if (specialProperties.has(path)) {
+      return;
+    }
+
     obj[path] = val;
     return;
   }
   const parts = path.split('.');
 
-  if (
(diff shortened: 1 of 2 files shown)


lib/helpers/path/setDottedPath.js
```diff
@@ -6,6 +6,15 @@ module.exports = function setDottedPath(obj, path, val) {
     return;
   }
   const parts = path.split('.');
+
+  if (
```
vkarpov15, line 10: Use `lib/helpers/specialProperties.js` instead of listing these properties out explicitly. Like how `clone()` handles it: https://github.com/Automattic/mongoose/blob/6f951f3baadd3a752d3e04bccb26b2163dd3e2a8/lib/helpers/clone.js#L141-L143

Also, we should skip special properties rather than throw an error.

vkarpov15 (commented):
One minor suggestion. Otherwise LGTM :+1:

lib/helpers/path/setDottedPath.js
Uzlopak: improved.

vkarpov15 (approved):
Thanks :+1:

Issue #11682, opened 2022-04-18
Proper setup for instance methods with Typescript


ShaneKeney:
**Do you want to request a *feature* or report a *bug*?**
*Help*.

**What is the current behavior?**
Following is specified in mongoose documentation:  "We recommend your document interface contain the properties defined in your schema and line up with what your documents look like in MongoDB. Although you can add [instance methods](https://mongoosejs.com/docs/guide.html#methods) to your document interface, we do not recommend doing so."

What is the recommended way for proper compatibility with Typescript project if `extends Document` is also discouraged.  This is specifically being asked in reference to instance methods such that TS can check properly when such methods are being used elsewhere in the project.

**What is the expected behavior?**

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
    Version of "mongoose": "^6.0.12",

AlejandroRodarte:
+1. I have an instance method that requires access to the `isModified` method, which requires my `Document` interface to extend from `mongoose.Document` in order to shut errors down. It would be nice to know what is the suggested way to implement this without extending.

Uzlopak:
The reason why it is not recommended, is because the typescript engine in vscode could get very slow. If you have no issue with that, you can ofc extend from Document. Maybe we should document that better to know why we dont recommend to extend from Document. Or maybe fix the underlying performance issue.

ShaneKeney:
@Uzlopak Understood.  That definitely makes me feel more at ease for leveraging it in our project.  This is by no means insinuating timeline or urgency because you all do great work and have bigger ticket items to consider.

I think a lot of confusion with TS and Mongoose overall can be solved with more concrete examples of how to set up a boiler plate project via common use cases.  
e.g. `Model.create` does not check field types or give any TS safety
e.g. Schema's that use `type: String, required: true, default: "Hello"` for supplying default values are confusing to deal with when creating the related TS interface.

Uzlopak:
@mohammad0-0ahmad is currently working on more typesafety.

Uzlopak:
#11673

vkarpov15:
We're adding more info on this to our docs. Below is an example of how you can define instance methods in a type safe way.

```typescript
import { Model, Schema, model } from 'mongoose';

interface IUser {
  firstName: string;
  lastName: string;
}

// Put all user instance methods in this interface:
interface IUserMethods {
  fullName(): string;
}

// Create a new Model type that knows about IUserMethods...
[...]

PR #11830, opened 2022-05-25
Fix a broken link introduced by #11682

lchenay:
**Summary**

#11682 have rename a doc file and by mistake forget a reference.


Uzlopak:
thx

Issue #11709, opened 2022-04-22
Typescript doc exemple for query helpers does not seem to work


MathieuFouillet:
**Do you want to request a *feature* or report a *bug*?**

It's either a bug or an issue with the doc.

**What is the current behavior?**

I can't create the [simple query helper from the doc](https://mongoosejs.com/docs/typescript/query-helpers.html). I have these 2 errors :

- TS2339: Property 'byName' does not exist on type '{}'.
- TS2339: Property 'find' does not exist on type '{}'.

````

import { Document, Model, Query, Schema, connect, model } from 'mongoose';

interface Project {
  name: string;
  stars: number;
}

const schema = new Schema<Project>({
  name: { type: String, required: true },
  stars: { type: Number, required: true }
});
// Query helpers should return `Query<any, Document<DocType>> & ProjectQueryHelpers`
// to enable chaining.
interface ProjectQueryHelpers {
  byName(name: string): Query<any, Document<Project>> & ProjectQueryHelpers;
}
schema.query.byName = function(name): Query<any, Document<Project>> & ProjectQueryHelpers {
  return this.find({ name: name });
};

// 2nd param to `model()` is the Model class to return.
const ProjectModel = model<Project, Model<Project, ProjectQueryHelpers>>('Project', schema);

run().catch(err => console.log(err));

async function run(): Promise<void> {
  await connect('mongodb://localhost:27017/test');
  
  // Equivalent to `ProjectModel.find({ stars: { $gt: 1000 }, name: 'mongoose' })`
  await ProjectModel.find().where('stars').gt(1000).byName('mongoose');
}

````

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

- node v14.15.4
- "typescript": "^4.6.3"
- "mongoose": "^6.3.1"
- mongodb: 4.4.6
- tsconfig.json
```
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "display": "Node 14",

  "compilerOptions": {
    "lib": ["es2020"],
    "module": "commonjs",
    "target": "es2020",

    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true
  }
}
```

IslandRhythms:
Does not happen on 6.3.0

PR #11832, opened 2022-05-25
Another fix for #11709 Typescript doc exemple for query helpers does not seem to work

theo97490:
The fix provided in #11709 didn't solve `TS2339: Property 'find' does not exist on type '{}'.`

When i copy paste the code raw the problem still happens.

I modifed by adding the third class type `THelpers` in the Query template. I hope the problem isn't me.

### DIFF
--- a/docs/typescript/query-helpers.md
+++ b/docs/typescript/query-helpers.md
@@ -31,8 +31,9 @@ interface Project {
   stars: number;
 }
 
+type ProjectDocumentType = Document<any, any, Project> & Project;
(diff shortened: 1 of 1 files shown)


vkarpov15 (commented):
Thanks for catching this. We just put in a commit to fix this issue before we saw this PR :( See https://github.com/Automattic/mongoose/commit/6f951f3baadd3a752d3e04bccb26b2163dd3e2a8 . But thanks for pointing this out, we appreciate it :+1:

PR #11843, opened 2022-05-28
feat(types+document): add `$assertPopulated()` for working with manually populated paths in TypeScript

### LINKED ISSUE #11758: Typescript: How to represent populated document fields on newly-constructed documents
feature? or question

When constructing a new document which has a foreign key field that could be populated, sometimes I will construct the document using the full foreign document rather than just _id. I would then like to use the fields from that "populated" document from the resulting document I've constructed. However, the type system thinks that field is just an "_id". The only way I know how to get the document in its "populated" form would be to go and query for it using the `populate` method, and provide the populated "paths" object there. I obviously don't want to go and re-fetch the document I've just made, so is there any way to define that this field is populated on creation? 

```typescript

interface NestedChild {
    name: string
    _id: Types.ObjectId
}

interface Parent {
    nestedChild: Types.ObjectId
    name?: string
}

    nestedChild: { type: Schema.Types.ObjectId, ref: 'NestedChild' },
    name: String
}))

const NestedChildModel = model<NestedChild>('NestedChild', nestedChildSchema)

async function run() {
    const parent = new ParentModel({
        nestedChild: new NestedChildModel({ name: 'test' }),
        name: 'Parent'
    })

    await parent.save()

    // doesn't work, nestedChild is expected to be an object id
    console.log(parent.nestedChild.name)
}

run()
```

I would like there to be a way to specify the "paths" on model construction similarly to "populate" so that documents I'm creating for the first time can be represented properly.

Mongoose 6.3.1

vkarpov15:
Fix #11758

**Summary**

Right now, if you instantiate a document with manually populated paths, there's no way to indicate that a path is manually populated as opposed to just an ObjectId.

```ts
const parent = new ParentModel({
    nestedChild: new NestedChildModel({ name: 'test' }),
    name: 'Parent'
})

parent.nestedChild; // always ObjectId, even if manually populated
```

With this PR, you can at least do:

```
const parent = new ParentModel({
    nestedChild: new NestedChildModel({ name: 'test' }),
    name: 'Parent'
}).$assertPopulated<{ nestedChild: NestedChild }>('nestedChild')

doc.nestedChild.name; // string
```

which will make TypeScript work correctly, and also correctly throw a runtime error if 'nestedChild' isn't populated. What do you think @AbdelrahmanHafez @Uzlopak ?

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -4315,8 +4315,8 @@ Document.prototype.$populated = Document.prototype.populated;
  *
  *     const doc = await Model.findOne().populate('author');
  *
- *     doc.$assertPopulated('author'); // ok
- *     doc.$assertPopulated('other path');
+ *     doc.$assertPopulated('author'); // does not throw
+ *     doc.$assertPopulated('other path'); // throws an error
  *
  *
  * @param {String} path
--- a/lib/document.js
+++ b/lib/document.js
@@ -4319,7 +4319,7 @@ Document.prototype.$populated = Document.prototype.populated;
  *     doc.$assertPopulated('other path'); // throws an error
  *
  *
- * @param {String} path
+ * @param {String | Array<String>} path
  * @return {Document} this
  * @memberOf Document
  * @instance
--- a/lib/document.js
+++ b/lib/document.js
@@ -4251,10 +4251,10 @@ Document.prototype.$getPopulatedDocs = function $getPopulatedDocs() {
  *
  * #### Example:
  *
- *     Model.findOne().populate('author').exec(function (err, doc) {
- *       console.log(doc.author.name)         // Dr.Seuss
- *       console.log(doc.populated('author')) // '5144cf8050f071d979c118a7'
- *     })
+ *     const doc = await Model.findOne().populate('author');
+ *
+ *     console.log(doc.author.name); // Dr.Seuss
+ *     console.log(doc.populated('author')); // '5144cf8050f071d979c118a7'
  *
  * If the path was not populated, returns `undefined`.
  *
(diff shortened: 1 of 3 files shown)


lib/document.js
```diff
@@ -4308,6 +4308,37 @@ Document.prototype.populated = function(path, val, options) {
...
 
+/**
+ * Throws an error if a given path is not populated
+ *
+ * #### Example:
+ *
+ *     const doc = await Model.findOne().populate('author');
+ *
+ *     doc.$assertPopulated('author'); // ok
+ *     doc.$assertPopulated('other path');
```
AbdelrahmanHafez, line 4319:
```suggestion
 *     doc.$assertPopulated('other path'); // throws an error
```

lib/document.js
```diff
@@ -4308,6 +4308,37 @@ Document.prototype.populated = function(path, val, options) {
...
 Document.prototype.$populated = Document.prototype.populated;
 
+/**
+ * Throws an error if a given path is not populated
+ *
+ * #### Example:
+ *
+ *     const doc = await Model.findOne().populate('author');
+ *
+ *     doc.$assertPopulated('author'); // ok
```
AbdelrahmanHafez, line 4318:
```suggestion
 *     doc.$assertPopulated('author'); // does not throw
```

lib/document.js
```diff
@@ -4308,6 +4308,37 @@ Document.prototype.populated = function(path, val, options) {
...
+ *
+ * #### Example:
+ *
+ *     const doc = await Model.findOne().populate('author');
+ *
+ *     doc.$assertPopulated('author'); // does not throw
+ *     doc.$assertPopulated('other path'); // throws an error
+ *
+ *
+ * @param {String} path
```
AbdelrahmanHafez, line 4322:
```suggestion
 * @param {String | Array<string>} path
```

AbdelrahmanHafez:
LGTM, thanks! 👍

One minor note though, it might be a good idea to give this label of _typescript_ as well, because I first assumed it was a JS fix, not a _mostly_ TS one.

AbdelrahmanHafez (approved):

PR #11844, opened 2022-05-28
fix documention of error handling

### LINKED ISSUE #11838: Typescript instanceof mongoose.Error

bug

Following the example in https://mongoosejs.com/docs/api/error.html#error_Error

```
const Model = mongoose.model('Test', new mongoose.Schema({ answer: Number }))
const doc = new Model({ answer: 'not a number' })
const err = doc.validateSync()

err instanceof mongoose.Error // true
```

I got the error in `err instanceof mongoose.Error `

```
namespace Error
The right-hand side of an 'instanceof' expression must be of type 'any' or of a type assignable to the 'Function' interface type.ts(2359)
```

node 16.14.2
mongoose 6.3.4

Uzlopak:
Resolves #11838

### DIFF
--- a/lib/error/index.js
+++ b/lib/error/index.js
@@ -5,11 +5,11 @@
  * Mongoose-specific errors.
  *
  * #### Example:
- *     const Model = mongoose.model('Test', new Schema({ answer: Number }));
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #11846, opened 2022-05-28
perf: improve isBsonType

Uzlopak:
Following the PR by @coyotte508 this PR improves the isBsonType function by avoiding the use of the actually heavy get-function.

### DIFF
--- a/lib/helpers/isBsonType.js
+++ b/lib/helpers/isBsonType.js
@@ -1,13 +1,15 @@
 'use strict';
 
-const get = require('./get');
-
 /*!
  * Get the bson type, if it exists
  */
 
 function isBsonType(obj, typename) {
-  return get(obj, '_bsontype', void 0) === typename;
+  return (
+    typeof obj === 'object' &&
+    obj !== null &&
+    obj._bsontype === typename
+  );
 }
 
 module.exports = isBsonType;
--- a/test/helpers/isBsonType.test.js
+++ b/test/helpers/isBsonType.test.js
@@ -3,6 +3,9 @@
 const assert = require('assert');
 const isBsonType = require('../../lib/helpers/isBsonType');
 
+const Decimal128 = require('mongodb').Decimal128;
+const ObjectId = require('mongodb').ObjectId;
+
 describe('isBsonType', () => {
   it('true for any object with _bsontype property equal typename', () => {
     assert.ok(isBsonType({ _bsontype: 'MyType' }, 'MyType'));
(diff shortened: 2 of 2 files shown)


lib/helpers/isBsonType.js
```diff
@@ -1,13 +1,15 @@
...
 /*!
  * Get the bson type, if it exists
  */
 
 function isBsonType(obj, typename) {
-  return get(obj, '_bsontype', void 0) === typename;
+  return (
+    typeof obj === 'object' &&
+    obj !== null &&
+    obj._bsontype === typename
```
coyotte508, line 11: During my previous PR I was wondering whether or not to add an additional check, eg ` && obj.constructor.name !== 'Object'` or ` && obj.constructor.name.localeCompare(typename, undefined, { sensitivity: 'accent' })`, if someone ever decides to store json data with `{_bsonType: xxx}` inside

Uzlopak: I dont think that this is necessary. The code change here is functionally equivalent to the previous code. If somebody really starts to create objects with their own _bsonType attribute, than it is their issue. ;)

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

lib/helpers/isBsonType.js
```diff
@@ -1,13 +1,15 @@
 'use strict';
 
-const get = require('./get');
-
 /*!
  * Get the bson type, if it exists
  */
 
 function isBsonType(obj, typename) {
-  return get(obj, '_bsontype', void 0) === typename;
```
AbdelrahmanHafez, line 10: Do we need the `typeof obj` check here?
How about `return (obj && obj._bsontype) === typename;`?

Uzlopak: Yes. If it was not an object, we would get an error. It is safer this way.

AbdelrahmanHafez: Can you give me an example of a value that throws an error with the suggested solution but not with the current one?

Uzlopak: @AbdelrahmanHafez  If I remove the null-check we get 36 breaking unit tests.

AbdelrahmanHafez: That shouldn't be the case, since the `&&` in `return (obj && obj._bsontype) === typename` does the `null` check. I tested locally and things are working for me. 🤨 

Anyway, either solution works well. Thanks for the performance improvement. 👍

PR #11855, opened 2022-05-30
implement Schema.discriminator

### LINKED ISSUE #11837: `Schema.discriminator()` not defined in TypeScript

Report a bug

The [newly added `Schema.discriminator()`](https://github.com/Automattic/mongoose/pull/11657/files#diff-ffadab0ba7a844f70b864530afe444d3684ca96212ca5b3e3bcd72291e535123R476-R480) function is not defined in the TypeScript file. This means that I am unable to use that function in TypeScript without overriding the compiler (which I don't like doing), and it wont be suggested by Intellisense, even when using Javascript.

Make a new Schema in a typescript file and try to call `.discriminator()`.

It should compile.

Mongoose 6.3.3, but I can still see it's not present even in the latest commit to `master`.

Uzlopak:
Resolves #11837

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -473,9 +473,39 @@ Schema.prototype.defaultOptions = function(options) {
   return options;
 };
 
+/**
+ * Inherit a Schema by applying a discriminator on an existing Schema.
+ *
+ *
+ * ####Example:
+ *
+ *     const options = { discriminatorKey: 'kind' };
+ *
+ *     const eventSchema = new mongoose.Schema({ time: Date }, options);
+ *     const Event = mongoose.model('Event', eventSchema);
+ *
+ *     // ClickedLinkEvent is a special type of Event that has
+ *     // a URL.
+ *     const ClickedLinkEvent = Event.discriminator('ClickedLink',
+ *     new mongoose.Schema({ url: String }, options));
+ *
+ *     // When you create a generic event, it can't have a URL field...
+ *     const genericEvent = new Event({ time: Date.now(), url: 'google.com' });
+ *     assert.ok(!genericEvent.url);
+ *     // But a ClickedLinkEvent can
+ *     const clickedEvent = new ClickedLinkEvent({ time: Date.now(), url: 'google.com' });
+ *     assert.ok(clickedEvent.url);
+ *
+ * @param {String} name the name of the discriminator
+ * @param {Schema} schema the Schema of the discriminated Schema
+ * @return {Schema} the Schema instance
+ * @api public
+ */
+
 Schema.prototype.discriminator = function(name, schema) {
-  this._applyDiscriminators = {};
-  this._applyDiscriminators[name] = schema;
+  this._applyDiscriminators = { [name]: schema };
+
+  return this;
 };
 
 /**
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -120,6 +120,17 @@ declare module 'mongoose' {
     useProjection?: boolean;
   }
 
+  export type DiscriminatorModel<M, T> = T extends Model<infer T1, infer T2, infer T3, infer T4>
+    ?
+    M extends Model<infer M1, infer M2, infer M3, infer M4>
+      ? Model<M1 | T1, M2 | T2, M3 | T3, M4 | T4>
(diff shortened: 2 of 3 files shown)


Uzlopak:
@simon-abbott
@mohammad0-0ahmad

lib/schema.js
```diff
@@ -473,9 +473,39 @@ Schema.prototype.defaultOptions = function(options) {
...
+ * @param {String} name the name of the discriminator
+ * @param {Schema} schema the Schema of the discriminated Schema
+ * @return {Schema} the Schema instance
+ * @api public
+ */
+
 Schema.prototype.discriminator = function(name, schema) {
-  this._applyDiscriminators = {};
-  this._applyDiscriminators[name] = schema;
+  this._applyDiscriminators = { [name]: schema };
```
Uzlopak, line 506: @vkarpov15 
Why do we always overwrite _applyDiscriminators?

simon-abbott: Yeah, I would expect multiple calls to `discriminator` to add multiple discriminators. This appears to only keep the last one.

Uzlopak:
Regarding your comment returning void. Actually by using discriminator we mutate the schema. Maybe it would ne better to actually instantiate a new Schema instance and apply the discriminator on it. Idk. Is it possible to type mutation in typescript?

lib/schema.js
vkarpov15: @Uzlopak @simon-abbott you're right, we should keep the previous `_applyDiscriminators`, this should be `Object.assign(this._applyDiscriminators, { [name]: schema })`

types/index.d.ts
```diff
@@ -120,6 +120,17 @@ declare module 'mongoose' {
...
 
+  export type DiscriminatorModel<M, T> = T extends Model<infer T1, infer T2, infer T3, infer T4>
+    ?
+    M extends Model<infer M1, infer M2, infer M3, infer M4>
+      ? Model<M1 | T1, M2 | T2, M3 | T3, M4 | T4>
+      : M
+    : M;
+
+  export type DiscriminatorSchema<DocType, M, TInstanceMethods, TQueryHelpers, TVirtuals, T> = T extends Schema<infer T1, infer T2, infer T3, infer T4, infer T5>
+    ? Schema<T1 | DocType, DiscriminatorModel<T2, M>, T3 | TInstanceMethods, T4 | TQueryHelpers, T5 | TVirtuals>
```
vkarpov15, line 131: Shouldn't this be `Schema<DocType & T1>`? Or `Schema<Omit<DocType, keyof T1> & T1>` ? The returned schema should have every field from `DocType` and `T1`, with fields from `T1` overriding conflicting fields in `DocType`.

Uzlopak:
applied changes

lib/schema.js
```diff
@@ -473,9 +473,39 @@ Schema.prototype.defaultOptions = function(options) {
...
+ * @param {String} name the name of the discriminator
+ * @param {Schema} schema the Schema of the discriminated Schema
+ * @return {Schema} the Schema instance
+ * @api public
+ */
+
 Schema.prototype.discriminator = function(name, schema) {
-  this._applyDiscriminators = {};
-  this._applyDiscriminators[name] = schema;
+  this._applyDiscriminators = Object.assign({}, this._applyDiscriminators, { [name]: schema });
```
vkarpov15, line 506: There's no need to create a new object every single time. Just `Object.assign(this._applyDiscriminators, { [name]: schema });` should be enough.

vkarpov15 (commented):
One quick tweak, otherwise LGTM :+1:

mohammad0-0ahmad:
I've not used this before actually, but according to the docs or "JS-docs" the example shows that discriminator will return a model, and it is a model FN.
But the ts-tests in this PR calls discriminator  by schema instead, and it returns a schema as well.
So I think either the docs should be fixed or these tests.
Sorry if I've missed or misunderstand the issue here.

lib/schema.js
Uzlopak: Unit Tests fail if i apply your Suggestion

vkarpov15 (approved):

PR #11885, opened 2022-06-02
make arg of Query.sort optional

ghost91-:
**Summary**

This change adjusts the type definitions to make the `arg` of `Query.sort` optional. This is useful in situations where the the sorting argument might be `undefined` in user code, because it avoids having to call `.sort(...)` only conditionally.

From an implementation perspective, the argument really is optional (see the implementation in mquery: https://github.com/aheckmann/mquery/blob/master/lib/mquery.js#L1239). In theory, we could also allow `null`, and explicitly add `undefined`, to make it easier for people who use the `exactOptionalPropertyTypes` TypeScript compiler option. If that is desired, I can adjust the PR accordingly.

**Examples**

```ts
function find(sortOrder?: { [key: string]: SortOrder }) {
  return SomeModel.find().sort(sortOrder); // this would be an error if the argument is not optional
}
```


Uzlopak:
Please add a test in test/types. Also if you like do the change regarding exactOptionalPropertyTypes

ghost91-:
Done, I added both `null` and `undefined`, and a couple of tests.

Uzlopak:
Thanks

PR #11890, opened 2022-06-03
InsertMany returns lean documents if lean is set to true

### LINKED ISSUE #11887: insertMany - lean: false returns hydrated documents
### Prerequisites

### Mongoose version

6.3.4

### Node.js version

16

### MongoDB server version

5.0.3

### Description

While working on the autotyping PR I realized, that lean: true does still return hydrated documents. 

### Steps to Reproduce

```javascript
    it('insertMany should not return hydrated Documents if lean is set to true', async function() {
      const eventSchema = new Schema({
        _id: Number,
        title: String,
      });
      const EventModel = db.model('Test', eventSchema);
      const test = await EventModel.insertMany(
        [{ _id: 3, title: 'Event 2' }],
        { lean: true}
      );

      assert.equal(typeof test[0].isNew, 'undefined')
    });
```

### Expected Behavior

Should return lean documents

Uzlopak:
Resolves #11887

Probably backwards breaking.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3463,17 +3463,16 @@ Model.$__insertMany = function(arr, options, callback) {
             return callback(err);
           }
 
-          callback(null, docs);
+          if (lean) {
+            return callback(null, docs.map(doc => doc.toObject()));
+          }
+
+          return callback(null, docs);
         });
       }
 
       if (lean) {
-        if (options.populate != null) {
-          callback(null, docAttributes.map(doc => doc.toObject()));
-        } else {
-          callback(null, docObjects);
-        }
-        return;
+        return callback(null, docObjects);
       }
 
       callback(null, docAttributes);
--- a/lib/model.js
+++ b/lib/model.js
@@ -3467,6 +3467,15 @@ Model.$__insertMany = function(arr, options, callback) {
         });
       }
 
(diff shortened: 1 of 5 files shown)


lib/model.js
```diff
@@ -3463,10 +3463,18 @@ Model.$__insertMany = function(arr, options, callback) {
             return callback(err);
           }
 
-          callback(null, docs);
+          if (lean) {
```
vkarpov15, line 3466: The problem is that in queries, `lean()` avoids hydrating documents. As opposed to here in `insertMany()`, as implemented, we're hydrating the document and then converting it back to a [POJO](https://masteringjs.io/tutorials/fundamentals/pojo). So there's no performance benefit here.

I'd say the better approach would be to avoid calling `doc = new _this(doc);`. It might be better to use `Model.validate()` to validate docs if `lean` is set. Or perhaps just skip validation entirely with `insertMany()` with `lean`.

vkarpov15 (commented):
I'm not sure that this implementation offers the same benefits that `lean` does. What do you think about skipping `new _this(doc)` if `lean` is set?

vkarpov15:
I'm gonna close this issue for now since it's been open for a while and we're not going to merge it without significant changes.

PR #11841, opened 2022-05-27
fix(document): fix ObjectId conversion for external schemas

coyotte508:
**Summary**

When dealing with external schemas, i.e. a mongoose schema created inside another module, `instanceOf` may not work.

This results in the value of ObjectIds in the schema being converted to strings when saving to the database.

**Examples**

I have a monorepo project with pnpm. One module `lib` contains a schema, which I use to create a model in my program. 

Both my program and the imported `lib` module have the same `bson`, `mongoose` and `mongodb` versions.

PR #11891, opened 2022-06-03
Fix ObjectId `instanceof` checks

noseworthy:
**Summary**

PR #11841 changed a whole bunch of checks for `instanceof ObjectId` to
use the `isBsonType()` function instead. Unfortunately some were missed
in the `./lib/schema/objectid.js` file.

Convert the remaining `instanceof` checks in `objectid.js` to use the
`isBsonType()` check instead so that all `ObjectId`s created from any
`bson` modules should validate/cast properly.

This was found because my nested document with required ObjectIds
started failing validation when I loaded the documents from JSON files
in my tests using `bson.EJSON` after the release of 6.3.4 -> 6.3.5.

**Examples**

This test when run in a project that depends on `mongoose` and `bson`
will fail before this fix and work afterward.

```js
const { EJSON } = require('bson');

describe('validation of required ObjectId type', () => {
  it('should be able to insert documents that correspond to the schema', async () => {
    const TestDocumentSchema = new mongoose.Schema({
      name: {
        required: true,
      },
      owner: {
        default: () => mongoose.Types.ObjectId(),
        required: true,
      },
    });

    const TestDocumentModel = mongoose.model(
      'TestDocument',
      TestDocumentSchema,
    );

    const result = await TestDocumentModel.insertMany([
      EJSON.parse(
        EJSON.stringify({
          name: 'Document 1',
          owner: new mongoose.Types.ObjectId(),
        }),
      ),
      EJSON.parse(EJSON.stringify({ name: 'document 2' })),
    ]);

    expect(result).not.toBeNull();
  });
});
```

### DIFF
--- a/lib/schema/objectid.js
+++ b/lib/schema/objectid.js
@@ -9,6 +9,7 @@ const SchemaType = require('../schematype');
 const castObjectId = require('../cast/objectid');
 const getConstructorName = require('../helpers/getConstructorName');
 const oid = require('../types/objectid');
+const isBsonType = require('../helpers/isBsonType');
 const utils = require('../utils');
 
 const CastError = SchemaType.CastError;
@@ -113,7 +114,7 @@ ObjectId.prototype.auto = function(turnOn) {
  * ignore
  */
 
-ObjectId._checkRequired = v => v instanceof oid;
+ObjectId._checkRequired = v => isBsonType(v, 'ObjectID');
 
 /*!
  * ignore
@@ -161,7 +162,7 @@ ObjectId.cast = function cast(caster) {
  */
 
 ObjectId._defaultCaster = v => {
-  if (!(v instanceof oid)) {
+  if (!(isBsonType(v, 'ObjectID'))) {
     throw new Error(v + ' is not an instance of ObjectId');
   }
   return v;
@@ -221,7 +222,7 @@ ObjectId.prototype.checkRequired = function checkRequired(value, doc) {
  */
 
 ObjectId.prototype.cast = function(value, doc, init) {
-  if (!(value instanceof oid) && SchemaType._isRef(this, value, doc, init)) {
+  if (!(isBsonType(value, 'ObjectID')) && SchemaType._isRef(this, value, doc, init)) {
     // wait! we may need to cast this to a document
     if ((getConstructorName(value) || '').toLowerCase() === 'objectid') {
       return new oid(value.toHexString());


Uzlopak:
Can you add the unit test to avoid again an regression?

noseworthy:
> Can you add the unit test to avoid again an regression?

Sure thing, @Uzlopak. Where should I put it? I was going to add that example test to this MR, and started putting it in `test/schema.validation.test.js`, but that didn't feel like the right place

Uzlopak:
Just create a schema.objectid.test.js. there is currently not a very strict Organisation. It is more important to have a test

noseworthy:
> Just create a schema.objectid.test.js. there is currently not a very strict Organisation. It is more important to have a test

Good thing you asked for a test. I can't recreate this issue in this repo, but I keep getting this error in my repository with the exact same test code:

```js
const result = await TestModel.insertMany([
  EJSON.deserialize({ name: 'Document 1', owner: { $oid: '6005c0bb52e4d267b5d96009' } })
]);
```

```
TestDocument validation failed: owner: Path `owner` is required.
ValidationError: TestDocument validation failed: owner: Path `owner` is required.
    at model.Object.<anonymous>.Document.invalidate (/Users/noseworthy/Development/X/backend/node_modules/mongoose/lib/document.js:2970:32)
    at /Users/noseworthy/Development/X/backend/node_modules/mongoose/lib/document.js:2759:17
    at /Users/noseworthy/Development/X/backend/node_modules/mongoose/lib/schematype.js:1333:9
    at processTicksAndRejections (node:internal/process/task_queues:78:11)
```

So there's something weird going on in my repository where the `v => v instanceof oid` check is returning false for me, but not in the mongoose repo. I'll try and dig a bit deeper.

noseworthy:
Sorry, @Uzlopak. I can't build a test in this repo that fails before my fix. As @coyotte508 said in the original MR #11841 , this issue is caused by having different instances of `bson` installed. `instanceof` is not validating that `ObjectId`s created by one instance of the dependency match objects created by `mongoose`s instance of `bson`.

I tried to hack together a test where I tried to artificially break `instanceof` but wasn't successful. I don't think there's much point in adding a test that doesn't fail before the fix. Any ideas on how I can replicate this?

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

Uzlopak:
@AbdelrahmanHafez why did you merge this?

AbdelrahmanHafez:
@Uzlopak 
Because it's an extension of #11841 which didn't include any tests, coming up with test cases for these scenarios is quite difficult, and OP couldn't come up with test cases.

Issue #11867, opened 2022-06-01
[Types] sort by $meta textScore is not allowed


Jule-:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.3.5

### Node.js version

16.14.2

### MongoDB server version

4.2.3

### Description

It is not possible to handle a text index search with sorting on the meta score like described in the MongoDB documentation: https://www.mongodb.com/docs/manual/core/text-search-operators/#query-framework

```
db.stores.find(
   { $text: { $search: "coffee shop cake" } },
   { score: { $meta: "textScore" } }
).sort( { score: { $meta: "textScore" } } )
```

Typescript is complaining:

```
Type '{ score: { $meta: string; }; }' is not assignable to type 'string | { [key: string]: SortOrder; }'.
  Types of property 'score' are incompatible.
    Type '{ $meta: string; }' is not assignable to type 'SortOrder | undefined'.ts(2345)
```

### Steps to Reproduce

Mongoose typings should allow doing:

```ts
const stores = await Store.find(
   { $text: { $search: "coffee shop cake" } },
   { score: { $meta: "textScore" } }
).sort( { score: { $meta: "textScore" } } )
```

### Expected Behavior

No Typescript error.

Uzlopak:
Should be fixed with PR #11370

PR #11893, opened 2022-06-04
fix(types): Allow sorting by text score

coyotte508:
**Summary**

Sorting by text score is a special case. See https://www.mongodb.com/docs/manual/reference/method/cursor.sort/#text-score-metadata-sort

This fixes a type error caused by 6.3.5

**Examples**

This code fails with 6.3.5:

```ts
const cities = await InseeCity.find({ $text: { $search: req.query.query as string } })
    .sort({ score: { $meta: "textScore" } })
```

It used to work

### DIFF
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -579,7 +579,7 @@ declare module 'mongoose' {
     snapshot(val?: boolean): this;
 
     /** Sets the sort order. If an object is passed, values allowed are `asc`, `desc`, `ascending`, `descending`, `1`, and `-1`. */
-    sort(arg?: string | { [key: string]: SortOrder } | { score: { $meta: 'textScore' } } | undefined | null): this;
+    sort(arg?: string | { [key: string]: SortOrder } | { [key: string]: { $meta: 'textScore' } } | undefined | null): this;
 
     /** Sets the tailable option (for use with capped collections). */
     tailable(bool?: boolean, opts?: {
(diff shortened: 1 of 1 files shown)


coyotte508:
Note: should fix #11867 and apparently https://github.com/Automattic/mongoose/pull/11370 would fix it as well.

types/query.d.ts
```diff
@@ -579,7 +579,7 @@ declare module 'mongoose' {
     snapshot(val?: boolean): this;
 
     /** Sets the sort order. If an object is passed, values allowed are `asc`, `desc`, `ascending`, `descending`, `1`, and `-1`. */
-    sort(arg?: string | { [key: string]: SortOrder } | undefined | null): this;
+    sort(arg?: string | { [key: string]: SortOrder } | { score: { $meta: 'textScore' } } | undefined | null): this;
```
vkarpov15, line 582: The sort key doesn't necessarily have to be called `score`. You can also do `{ textScore: { $meta: 'textScore' } }`, see https://thecodebarbarian.com/2014/04/10/a-nodejs-perspective-on-whats-new-in-mongodb-2-6-part-i-text-search.html .

According to https://www.mongodb.com/docs/manual/reference/operator/aggregation/meta/, `{ $meta: 'indexKey' }` is also valid.

Maybe something like `{ [key: string]: { $meta: 'textScore' | 'indexKey' } }` ?

vkarpov15 (changes requested):

types/query.d.ts
coyotte508: Good point about the `score` key. `{ $meta: 'indexKey' }` should be valid but not for sorting.

vkarpov15 (approved):
Nice, thanks :+1:

PR #11898, opened 2022-06-05
fix: Discriminator generic type not being passed to schema

GCastilho:
-- Fix instance created from model returned from discriminator having type 'any'
-- Fix model created from discriminator not inheriting the properties from the base model

**Summary**

A model created from a discriminator does not have types for the "document" (save, find, etc) as the original model does and does not inherit the properties from the base model

**Examples**

```typescript
type EventSchema = {
    time: Date
}
const eventSchema = new mongoose.Schema<EventSchema>({ time: Date }, options);
const Event = mongoose.model('Event', eventSchema);

type ClickedLinkEventSchema = {
    url: string
}
const clickedLinkEventSchema = new mongoose.Schema<ClickedLinkEventSchema>({ url: String }, options)
const ClickedLinkEvent = Event.discriminator('ClickedLink', clickedLinkEventSchema);
const clickedLinkEvent = new ClickedLinkEvent({
    time: new Date(),
    url: 'https://mongoosejs.com/',
});

# Old
clickedLinkEvent.time  // time has type any
clickedLinkEvent.save() // save has type any

# New
clickedLinkEvent.time  // time has type Date
clickedLinkEvent.save() // save has the type from Document['save']
```

This is a remake of #11897 with fixed branch name and commits

### DIFF
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -3,7 +3,7 @@ declare module 'mongoose' {
 
   export interface AcceptsDiscriminator<B> {
     /** Adds a discriminator type. */
-    discriminator<D>(name: string | number, schema: Schema<D>, value?: string | number | ObjectId): Model<B & D>;
+    discriminator<D>(name: string | number, schema: Schema<D>, value?: string | number | ObjectId): Model<Omit<B, keyof D> & D>;
     discriminator<T, U>(name: string | number, schema: Schema<T, U>, value?: string | number | ObjectId): U;
   }
 
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -1,9 +1,9 @@
 declare module 'mongoose' {
   import mongodb = require('mongodb');
 
-  export interface AcceptsDiscriminator {
+  export interface AcceptsDiscriminator<B> {
     /** Adds a discriminator type. */
-    discriminator<D>(name: string | number, schema: Schema, value?: string | number | ObjectId): Model<D>;
+    discriminator<D>(name: string | number, schema: Schema<D>, value?: string | number | ObjectId): Model<B & D>;
     discriminator<T, U>(name: string | number, schema: Schema<T, U>, value?: string | number | ObjectId): U;
   }
 
@@ -116,7 +116,7 @@ declare module 'mongoose' {
   const Model: Model<any>;
   interface Model<T, TQueryHelpers = {}, TMethodsAndOverrides = {}, TVirtuals = {}> extends
     NodeJS.EventEmitter,
-    AcceptsDiscriminator,
+    AcceptsDiscriminator<T>,
     IndexManager,
     SessionStarter {
     new <DocType = AnyKeys<T> & AnyObject>(doc?: DocType, fields?: any | null, options?: boolean | AnyObject): HydratedDocument<T, TMethodsAndOverrides, TVirtuals>;
--- a/types/schematypes.d.ts
+++ b/types/schematypes.d.ts
@@ -279,7 +279,7 @@ declare module 'mongoose' {
         discriminator<D>(name: string | number, schema: Schema<D>, value?: string | number | ObjectId): Model<Omit<B, keyof D> & D>;
 
         /** The schematype embedded in this array */
-        caster?: SchemaType;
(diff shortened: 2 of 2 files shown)


types/models.d.ts
```diff
@@ -1,9 +1,9 @@
 declare module 'mongoose' {
   import mongodb = require('mongodb');
 
-  export interface AcceptsDiscriminator {
+  export interface AcceptsDiscriminator<B> {
     /** Adds a discriminator type. */
-    discriminator<D>(name: string | number, schema: Schema, value?: string | number | ObjectId): Model<D>;
+    discriminator<D>(name: string | number, schema: Schema<D>, value?: string | number | ObjectId): Model<B & D>;
```
vkarpov15, line 6: Probably better to use `Omit<B, keyof D> & D` instead of `B & D`. Discriminator schemas can overwrite parent schema properties.

vkarpov15 (changes requested):
Tests are failing because you also need to modify `types/schematypes.d.ts`, there's a couple places in that file where we use `AcceptsDiscriminator`: https://github.com/Automattic/mongoose/blob/4cf2920ba4ce9910a12ef6885fa19d17049db211/types/schematypes.d.ts#L272 for example.

types/models.d.ts
GCastilho: This is indeed a bug. Fixed!

GCastilho:
I've updated the classes in `types/schematypes.d.ts` that implemented the `AcceptsDiscriminator` to be generics

Their 'B' type is passed to `SchemaType` and `Schema` so I believe it is "the type of the class", so "an Array of numbers", using your example

It defaults to `any` so no change in behavior is expected

types/schematypes.d.ts
```diff
@@ -269,17 +269,17 @@ declare module 'mongoose' {
...
-      class Array extends SchemaType implements AcceptsDiscriminator {
+      class Array<B = any> extends SchemaType<B> implements AcceptsDiscriminator<B> {
         /** This schema type's name, to defend against minifiers that mangle function names. */
         static schemaName: 'Array';
 
         static options: { castNonArrays: boolean; };
 
         discriminator<T, U>(name: string | number, schema: Schema<T, U>, value?: string): U;
-        discriminator<D>(name: string | number, schema: Schema, value?: string): Model<D>;
+        discriminator<D>(name: string | number, schema: Schema<D>, value?: string | number | ObjectId): Model<Omit<B, keyof D> & D>;
```
GCastilho, line 279: I've just noticed

Is the `value` here supposed to also allow number and ObjectId? Because I've copied this line from the AcceptsDiscriminator interface and didn't noticed that it didn't allow for these types here

I _suppose_ they should because they implements AcceptsDiscriminator that allows it, but I'm not sure

vkarpov15 (approved):
Thanks :+1:

vkarpov15:
In hindsight, I think I'm going to have to revert this PR. This really slows down our TS performance benchmark.

Before:

```
Run node ./benchmarks/typescript
simple instantiations: 42294
simple memory used: 1[6](https://github.com/Automattic/mongoose/runs/6776724234?check_suite_focus=true#step:5:7)5158.9
simple check time: 2.[7](https://github.com/Automattic/mongoose/runs/6776724234?check_suite_focus=true#step:5:8)21[9](https://github.com/Automattic/mongoose/runs/6776724234?check_suite_focus=true#step:5:10)999999999995
simple total time: 4.175999999999999
```

After:

```
Run node ./benchmarks/typescript
simple instantiations: 2557[1](https://github.com/Automattic/mongoose/runs/6747560604?check_suite_focus=true#step:5:1)4
simple memory used: 19440[7](https://github.com/Automattic/mongoose/runs/6747560604?check_suite_focus=true#step:5:8).2
simple check time: 3.99[8](https://github.com/Automattic/mongoose/runs/6747560604?check_suite_focus=true#step:5:9)
simple total time: 5.448
```

GCastilho:
Well, it _does_ make sense that performance would slow down a little since the types were being defaulted to any and with this PR some generics were added for the types to be passed down

It is a new "feature", per se, it make sense for it to impact _something_

PR #11900, opened 2022-06-05
feat(mongoose): add `setDriver()` function to allow overwriting `driver` in a more consistent way

vkarpov15:
**Summary**

Right now overwriting the Mongoose driver is a bit messy, because 1) `Connection` and `Collection` aren't changed when you do `mongoose.driver.set()`, and 2) driver is always global, so no way to have 2 Mongoose instances with the same driver in the same process.

**Examples**

You can now do `mongoose.setDriver()` without any extra work to use a different driver.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -507,7 +507,6 @@ Mongoose.prototype.pluralize = function(fn) {
  */
 
 Mongoose.prototype.model = function(name, schema, collection, options) {
-
   const _mongoose = this instanceof Mongoose ? this : mongoose;
 
   if (typeof schema === 'string') {
@@ -786,7 +785,6 @@ Object.defineProperty(Mongoose.prototype, 'Collection', {
  * @method Connection
  * @api public
  */
-
 
 Object.defineProperty(Mongoose.prototype, 'Connection', {
   get: function() {
--- a/lib/index.js
+++ b/lib/index.js
@@ -35,6 +35,7 @@ const shardingPlugin = require('./plugins/sharding');
 const trusted = require('./helpers/query/trusted').trusted;
 const sanitizeFilter = require('./helpers/query/sanitizeFilter');
 const isBsonType = require('./helpers/isBsonType');
+const MongooseError = require('./error/mongooseError');
 
 const defaultMongooseSymbol = Symbol.for('mongoose:default');
 
@@ -62,6 +63,7 @@ function Mongoose(options) {
   this.connections = [];
   this.models = {};
   this.events = new EventEmitter();
+  this.__driver = driver.get();
   // default global options
   this.options = Object.assign({
     pluralization: true,
@@ -70,7 +72,6 @@ function Mongoose(options) {
   }, options);
   const conn = this.createConnection(); // default connection
   conn.models = this.models;
-  driver._mongooseInstances.push(this);
 
   if (this.options.pluralization) {
     this._pluralize = legacyPluralize;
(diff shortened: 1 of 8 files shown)


lib/index.js
```diff
@@ -135,13 +137,43 @@ Mongoose.prototype.ConnectionStates = STATES;
...
+ */
+
+Mongoose.prototype.setDriver = function setDriver(driver) {
+  const _mongoose = this instanceof Mongoose ? this : mongoose;
+
+  if (_mongoose.__driver === driver) {
+    return;
+  }
+
+  if (_mongoose.connections && _mongoose.connections.find(conn => conn.readyState !== STATES.disconnected)) {
```
AbdelrahmanHafez, line 164:
```suggestion
  const openConnection = _mongoose.connections && _mongoose.connections.find(conn => conn.readyState !== STATES.disconnected);
  if (openConnection) {
```

AbdelrahmanHafez (commented):
Can you think of a way we can mimic having different drivers and setting them to verify that this works?

Would deep cloning the driver, modify some arbitrary property, then setting it, and asserting that `get`ting the driver, we'll have the arbitrary property be enough testing?

Otherwise, LGTM.

test/types/populate.test.ts
```diff
@@ -228,4 +228,29 @@ async function _11532() {
   if (!leanResult) return;
   expectType<string>(leanResult.child.name);
   expectError(leanResult?.__v);
+}
+
+async function gh11710() {
```
Uzlopak, line 233: Is this accidently added?

lib/index.js
```diff
@@ -135,13 +137,44 @@ Mongoose.prototype.ConnectionStates = STATES;
...
+Mongoose.prototype.setDriver = function setDriver(driver) {
+  const _mongoose = this instanceof Mongoose ? this : mongoose;
+
+  if (_mongoose.__driver === driver) {
+    return;
+  }
+
+  const openConnection = _mongoose.connections && _mongoose.connections.find(conn => conn.readyState !== STATES.disconnected);
+  if (openConnection) {
+    const msg = 'Cannot modify Mongoose driver if a connection is already open. ' +
```
Uzlopak, line 166: nitpick: why do we need a one time variable? We could directly pass it to MongooseError

lib/index.js
```diff
@@ -135,13 +137,44 @@ Mongoose.prototype.ConnectionStates = STATES;
...
+ * @memberOf Mongoose
+ * @method setDriver
+ * @api public
+ */
+
+Mongoose.prototype.setDriver = function setDriver(driver) {
+  const _mongoose = this instanceof Mongoose ? this : mongoose;
+
+  if (_mongoose.__driver === driver) {
+    return;
```
Uzlopak, line 161: According to the typings we expect always to return Mongoose.

```suggestion
    return _mongoose;
```

Uzlopak (changes requested):
I dont know how to test this. Maybe creating a mock with a Proxy around the actual mongodb driver, and then checking if the Proxy is set after using setDriver. And when we are successful we set back the original driver.

lib/index.js
vkarpov15: Better readability IMO. Easier to break up the long message if there's a temporary variable rather than putting it as a param.

test/types/populate.test.ts
vkarpov15: No, just wrong base. We should merge this into 6.4.

Uzlopak (approved):
LGTM

Issue #6818, opened 2018-08-06
Validator bug when calls save() function repeatedly


lutherman-buildit:
Hi,

When I test my validator in mocha, I find that the validators will be wrong if call save() function repeatedly.

Here is my part of schema:
```javascript
const ProjectSchema   = new Schema({
    name: String
    // ... 
    roles: {
        type: [{
            name: String, 
            folders: {
                type: [{
                    folderId: {
                        type: Schema.Types.ObjectId,
                        validate: {
                            validator: MongoValidator.rolesFoldersFolderId
                        },
                    },
                   // ...
                }],
                validate: {
                    validator: MongoValidator.rolesFolders
                }
            }
        }],
        validate: {
            validator: MongoValidator.roles
        }
    },
    // ...
    active: { type: Boolean, default: true, select: false},
});
```

And here is my part of mocha code:
```javascript
// ...
it('Edit folder id in role to a duplicate id', () => {
       return expect(Project.findById('5b67ac70bdb7915e3041feab')
            .then(projectDoc => {
               // First save(): add a folder object
               projectDoc.roles.find(role => role.name === 'role1').folders.push({ folderId: '5b67aadf3405b05e09304b83' });
               return projectDoc.save();
            })
           .then(projectDoc => {
               // Second save(): edit the folder id
               projectDoc.roles.find(role => role.name === 'role1').folders.find(roleFolder => roleFolder.folderId.toString() === '5b67aadf3405b05e09304b83').folderId = '5b67aadf3405b05e09304b82';
               return projectDoc.save();
            }))
            .be.rejected.then(error => {
                  expect(error).to.have.property('name').equal('ValidationError');
 expect(error).to.have.property('errors').to.have.property('roles.1.folders').to.have.property('message').equal('Folder ID in role must be unique');
             });
});
// ...
```
After I run the code with debugger, I find two problems happened in second save().
___
### Cannot Trigger the Validator
I find that the validator 'rolesFolders' will not be triggered to validate the 'folders' array of 'role1' object, but can be triggered by other role objects ('Administrator' and 'role2').

Here is the code of validator 'rolesFolders':
```javascript
function rolesFolders(folders) {
  if ((this.isNew || this.isModified('folders')) && folders.find((folder, index) => folders.find((folderRef, indexRef) => folder.folderId.toString() === folderRef.folderId.toString() && index !== indexRef))) {
    throw new Error('Folder ID in role must be unique');
  } else {
    return true;
  }
}
```
___
[...]

PR #11902, opened 2022-06-06
fix: make `doValidate()` on document array elements run validation on the whole subdoc

### LINKED ISSUE #11770: Make `$__reset()` not mark paths as `init`

Potential bug/issue

It is very strange that `$__reset()`, which should reset a document to a "nothing modified" state, relies on marking subdocuments as `init`. This popped up in #11672, and this behavior may be hiding other bugs. No concrete issues right now, more of a note to self to investigate further.

vkarpov15:
Fix #11770

**Summary**

In #11770 I made a note that we leave subdocuments in `init` state after `save()` to work around some issues like #6818. That is technically incorrect, because `init` should only be used for paths that are loaded from the database. The underlying issue comes down to the fact that array subdocument paths' `doValidate()` function doesn't run validation on paths in the document.

In this PR, I made it so that the embedded schema type in document arrays have a `doValidate()` that is the same as `SubdocumentPath`'s. In order to support this change, needed to fix a couple of places where we relied on the old behavior in update validators and `Model.validate()`.

Future work: make array subdocument paths a separate class, and make primitive arrays also validate all their subpaths when calling `doValidate()`.

We should probably put this change in 6.4, or hold off until after 6.4 It is a fairly sizable change of Mongoose internals, and while our tests do all pass, I'd like to be cautious with this one.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2484,7 +2484,7 @@ function _getPathsToValidate(doc) {
       const fullPathToSubdoc = subdoc.$__fullPathWithIndexes();
 
       for (const p of paths) {
-        if (p === null || p.startsWith(fullPathToSubdoc + '.')) {
+        if (p == null || p.startsWith(fullPathToSubdoc + '.')) {
           paths.delete(p);
         }
       }
@@ -2507,7 +2507,7 @@ function _getPathsToValidate(doc) {
 
     if (_pathType.$isMongooseDocumentArray) {
       for (const p of paths) {
-        if (p === null || p.startsWith(_pathType.path + '.')) {
+        if (p == null || p.startsWith(_pathType.path + '.')) {
           paths.delete(p);
         }
       }
--- a/lib/document.js
+++ b/lib/document.js
@@ -2505,6 +2505,14 @@ function _getPathsToValidate(doc) {
       continue;
     }
 
+    if (_pathType.$isMongooseDocumentArray) {
+      for (const p of paths) {
+        if (p === null || p.startsWith(_pathType.path + '.')) {
+          paths.delete(p);
+        }
+      }
+    }
+
     // Optimization: if primitive path with no validators, or array of primitives
     // with no validators, skip validating this path entirely.
     if (!_pathType.caster && _pathType.validators.length === 0) {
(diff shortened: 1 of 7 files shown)


lib/document.js
```diff
@@ -2505,6 +2505,14 @@ function _getPathsToValidate(doc) {
       continue;
     }
 
+    if (_pathType.$isMongooseDocumentArray) {
+      for (const p of paths) {
+        if (p === null || p.startsWith(_pathType.path + '.')) {
```
AbdelrahmanHafez, line 2510: Is strict equality in `p === null` intended, or did you mean `p == null`?

AbdelrahmanHafez:
LGTM, thanks. 👍

Frankly, I've always found it difficult to tell the difference between Subdocument, embedded documents, and arrays of documents.
If we were to design mongoose with the knowledge we have today, is there a way we can have logic apply to all cases?

AbdelrahmanHafez (approved):

lib/document.js
Uzlopak: In ine 2487 we also have strill null check

lib/helpers/updateValidators.js
```diff
@@ -125,26 +125,20 @@ module.exports = function(query, schema, castedDoc, options, callback) {
...
+                for (const key of Object.keys(err.errors)) {
+                  const _err = err.errors[key];
+                  _err.path = updates[i] + '.' + key;
+                  validationErrors.push(_err);
+                }
+              } else {
+                err.path = updates[i];
+                validationErrors.push(err);
+              }
               return callback(null);
```
Uzlopak, line 138: I guess we can remove this line, as we anyway go to line 141?

Uzlopak (changes requested):
can we have some tests please :)?

vkarpov15:
@AbdelrahmanHafez we did some work to consolidate array subdocuments and single nested subdocuments for 6.0. Now, `ArraySubdocument` extends from `Subdocument`. Bringing nested paths into that mix would be very tricky, because nested paths are by design not subdocuments.

PR #11921, opened 2022-06-09
fix(types): improve helpers typings

10kc-emilyk:
**Summary**
A simple typescript definition fix.

**Rationale**

`isObjectIdOrHexString` in the `any` position is used by typescript, resulting in a `false` when passed arguments of type `mongodb.ObjectId | string` - and typescript or typescript-eslint will complain in certain circumstances, such unreachable code. Similar with isValidObjectId

**Examples**

### DIFF
--- a/types/helpers.d.ts
+++ b/types/helpers.d.ts
@@ -20,14 +20,13 @@ declare module 'mongoose' {
    * of an ObjectId.
    */
   function isObjectIdOrHexString(v: mongodb.ObjectId): true;
-  function isObjectIdOrHexString(v: string): boolean;
+  function isObjectIdOrHexString(v: mongodb.ObjectId | string): boolean;
   function isObjectIdOrHexString(v: any): false;
 
   /**
    * Returns true if Mongoose can cast the given value to an ObjectId, or
    * false otherwise.
    */
-  function isValidObjectId(v: mongodb.ObjectId): true;
-  function isValidObjectId(v: Types.ObjectId): true;
+  function isValidObjectId(v: mongodb.ObjectId | Types.ObjectId): true;
   function isValidObjectId(v: any): boolean;
 }


types/helpers.d.ts
```diff
@@ -20,14 +20,13 @@ declare module 'mongoose' {
    * of an ObjectId.
    */
   function isObjectIdOrHexString(v: mongodb.ObjectId): true;
-  function isObjectIdOrHexString(v: string): boolean;
+  function isObjectIdOrHexString(v: mongodb.ObjectId | string): boolean;
```
AbdelrahmanHafez, line 23: Welcome to mongoose @10kc-emilyk 

Shouldn't `mongodb.ObjectId` always be true, since it is an object id?

Also, I wonder what effect will that have on the typing, if we have one signature with `ObjectId`, and another with `ObjectId | string`. Which one will TS pick when I use `isObjectIdOrHexString(someObjectId);`?

10kc-emilyk: TS will pick the first overload signature, so when the type is just `ObjectId`, the return type will resolve as `true`. The secondary one will only be hit if you're passing a value that could be an objectId _or_ a string. Currently, if you pass a value that is `ObjectId | string`, the return type will be `false`.

AbdelrahmanHafez: Thanks for the explanation.

Does that mean the proposed change below for `isValidObjectId` will make it so that all the following cases will return `true`?
* mongodb.ObjectId | Types.ObjectId
* mongodb.ObjectId
* Types.ObjectId

Or will it only return true if the type is an exact match (e.g. `mongodb.ObjectId | Types.ObjectId`)?

Can we add some type tests to /tests/types to verify that things work well and avoid regressions in the future?

Uzlopak:
Yeah makes sense.

types/helpers.d.ts
Uzlopak: We have already typings tests.

AbdelrahmanHafez: Today I learned: you can see references for types from PRs directly:
Screenshot 2022-06-10 at 5 00 04 PM
Thanks @Uzlopak

PR #11922, opened 2022-06-10
fix: toJSON with undefined path, add unit test

kerryChen95:
**Summary**

For `doc = model.findById(id, projection)`, `doc` may not contain some paths because of `projection`.

`doc.toJSON()` should not contain those paths even work with transform.

**Examples**

See unit test in the changes.


Uzlopak:
Please run one time `npm run lint -- --fix` to solve the linting issues and commit it. Thx

PR #11924, opened 2022-06-11
fix: post hook 'res' param with type any

GCastilho:
-- Fix post hook using PostMiddlewareFunction generic type without passing T to ResType

**Summary**

The type for the 'post hook' middleware is passing the `T` only to the first argument of the PostMiddlewareFunction (the `this` type), not to the second, resulting on the `ResType` defaulting to `any`

**Examples**
```typescript
type User = {
  name: string
}
const userSchema = new Schema<User>({ ... })

# Before
userSchema.post('validate', (doc, next) => {
  // doc has type 'any'
})

# After
userSchema.post('validate', (doc, next) => {
  // doc has type 'User'
})

```

### DIFF
--- a/test/types/middleware.test.ts
+++ b/test/types/middleware.test.ts
@@ -32,6 +32,16 @@ schema.post<Aggregate<any>>('aggregate', async function(res: Array<any>) {
   console.log('Pipeline', this.pipeline(), res[0]);
 });
 
+schema.post<Aggregate<ITest>>('aggregate', function(res, next) {
+  expectType<ITest[]>(res);
+  next();
+});
+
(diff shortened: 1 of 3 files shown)


Uzlopak:
Can you add some typings tests please?

GCastilho:
@Uzlopak I've added some tests for the types. Please let me know if you think it needs more or if there is a problem with them

Uzlopak (approved):
LGTM

vkarpov15:
In principle this looks ok, but the TS benchmark results don't look so good. Although I suspect they may be better after https://github.com/Automattic/mongoose/pull/11925 .

GCastilho:
Thank you :D

PR #11927, opened 2022-06-13
feat: add `skipMiddlewareFunction()` and `overwriteMiddlewareResult()` for skipping and modifying middleware results

### LINKED ISSUE #8393: Modify `find()` results using `Array#filter()` in `post('find')`
Is it possible somehow modify the `find*` results in post hooks?

**Example:** 
I have a schema of tokens, they have expiration time and i want expired tokens to be deleted automatically and not being presented in `find*` results.

Aggregation hooks solves the results presence issue but dow not solve the deletion.
Query hooks solves the deletion issue (which is the most important of two) mut lacks of result modification.

```javascript
const tokensSchema = new mongoose.Schema({
  exp: { type: Number, immutable:true }
});

tokensSchema.post('find', async (items)=>{
    const now = Date.now();

    const toDelete: IAuthToken[] = [];
    // is there a way to store filtered items?
    items = items.filter(item => {
      if (item.exp <= now) {
        toDelete.push(item);

        return false;
      }

      return true;
    });

    await Promise.all(toDelete.map(async (item) => await item.remove()));
  });
});

tokensSchema.post( /find(?:.+)/i,
  async function(item: IAuthToken) {
    const now = Date.now() + 3600000;

    item.exp <= now && await item.remove();
  });
```

I know that returned documents can be checked for deletion with `.$isDeleted()` method, but it is absolutely uncomfortable to check it everywhere right after the fetch.

I assume that i may not see/missed the right solution due to i'm new to the mongoose (and mongo at all tbh).

### LINKED ISSUE #11426: Unable to filter response from 'find / findOne' within plugin
Feature

Unable to filter response from a `.find()` or `.findOne()` call.

Given a plugin:

```javascript
(schema: Schema) => {
    schema.post('findOne', async function(doc) {
      if (await shouldRemove(doc)) return undefined
    
      return doc
    })
  }
```

I should be able to selectively return either nothing, or a document, from `.find()` and `.findOne()` calls.

If `shouldRemove` returns true, for no document to be returned from `undefined`.  

Mongoose v6.1.6, Node v16.13.2,

vkarpov15:
Fix #11426
Fix #8393

**Summary**

Right now there's no way to overwrite the result of a function in a `post()` hook. For example, there's no way to use `filter()` to modify a `find()` result in a `post('find')` hook.

There's also no way to "cancel" an operation in a pre hook without throwing an error. Which is inconvenient, because there are some use cases where it would be great to skip executing the underlying function (local caching, memorizing results, etc.)

This PR adds `overwriteMiddlewareResult()` for the former, and `skipMiddlewareFunction()` for the latter, from https://github.com/vkarpov15/kareem/pull/30 .

@AbdelrahmanHafez @Uzlopak what do you think about the naming and overall idea?

**Examples**

See tests

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,6 +8,7 @@ require('./driver').set(require('./drivers/node-mongodb-native'));
 
 const Document = require('./document');
 const EventEmitter = require('events').EventEmitter;
+const Kareem = require('kareem');
 const Schema = require('./schema');
 const SchemaType = require('./schematype');
 const SchemaTypes = require('./schema/index');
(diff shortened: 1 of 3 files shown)


Uzlopak:
I think skipMiddleware is better than skipMiddlewareFunction 
But it raises the question: are you skipping the middleware or are you skipping the hook? skipMiddleware would tell me that all the  potential following hooked functions are also skipped. skipHook would indicate that only this function gets skipped.

So it would be imho good to show in a unit test what happens with following hooks. Are they also skipped or not.

junmiu:
@vkarpov15 `skipMiddlewareFunction` only works for `Query` not `QueryCursor`, here is an example
``` js
// setup a pre hook for find
schema.pre('find', function (next) {
  next(skipMiddlewareFunction([]));
});
```
``` js
// err is { "args": [[]] }, an instance of skipWrappedFunction
mongooseModel.find().cursor().on('error', function (err) { console.log(err); });
```
The reason is in `lib/cursor/QueryCursor.js`, it called like this
``` js
// here may want to call query._find()
// or wrap this with kareem.createWrapper()
model.hooks.execPre('find', query, (err) => {
  if (err != null) {
    _this._markError(err);
    _this.listeners('error').length > 0 && _this.emit('error', err);
    return;
  }
  // do other things
});
```

Issue #11944, opened 2022-06-17
Export node-mongodb-native types


LvChengbin:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Allow users to provide `node-mongodb-native` package manually.

### Motivation

Although mongoose provide a static `mongo` property reference to `node-mongodb-native` but it's hard to import some `class` from `node-mongodb-native` as `Type` in typescript.

I am going to implement a class similar to the code bellow:

```ts
class BucketService {
    bucket: typeof mongoose.mongo.GridFSBucket;
    constructor() {
        this.bucket = new mongoose.mongo.GridFSBucket( ... );
    }
}
```

In this case, error message is emitted from typescript compiler.

```
error TS2740: Type 'GridFSBucket' is missing the following properties from type 'typeof GridFSBucket': prototype, INDEX, getEventListeners, errorMonitor, and 4 more.
```

If I install `node-mongodb-native` in my package just for importing the `GridFSBucket` type, I have to keep it has same version with which in mongoose dependencies.

Otherwise, exporting types from `node-mongodb-native` to support syntax like `import { GridFSBucket } from 'mongoose/mongo` might be also make sense.

vkarpov15:
We'll consider exporting all the types. But the better approach really is to keep the same version of node-mongodb-native in your `package.json` dependencies.

We're not likely to put node-mongodb-native as a peer dependency in the near future, because right now we think that supporting exactly 1 version of node-mongodb-native per version of Mongoose is better for maintainability.

AbdelrahmanHafez:
From a usability point of view, I find exporting all MongoDB types as `mongoose/mongodb-driver` or `mongoose/mongodb-driver-types` seems like the best solution in terms of clear separation of types, and least maintenance cost.

@vkarpov15 Having to manually check for what version of the native driver mongoose is using, and adding it to package.json is not a very reliable way. Since we already have the driver as a dependency, I think we should be able to export types, and can't see a downside for anyone here.

What do you think?

vkarpov15:
Yeah i think exporting the MongoDB driver types is a good option as long as we don't have to duplicate the types. If there's a 1-5 line solution, I'm all for it

vkarpov15:
Fixed in #12948

AbdelrahmanHafez:
@vkarpov15 Seems like this is not completely fixed.
[...]

PR #11954, opened 2022-06-19
Export mongo types

### LINKED ISSUE #11605: [Types] all used types should be exported

AbdelrahmanHafez:
fixes #11605 #11944

### DIFF
--- a/test/types/mongodb.test.ts
+++ b/test/types/mongodb.test.ts
@@ -1,0 +1,7 @@
+import { AggregateOptions } from 'mongoose/mongodb-driver';
+import { expectType } from 'tsd';
+
+
+const options: AggregateOptions = { };
+expectType<boolean | undefined>(options.allowDiskUse);
+expectType<string | undefined>(options.dbName);
--- a/types/mongodb-driver.d.ts
+++ b/types/mongodb-driver.d.ts
@@ -1,0 +1,3 @@
+declare module 'mongoose/mongodb-driver' {
+  export * from 'mongodb';
+}


Uzlopak:
Please dont merge it yet. I have to think about it a little.

Uzlopak:
I think this could result in some issues, because your declaration you export the whole mongodb definitions and active code. But this would be wrong, as you can not instantiate anything.

AbdelrahmanHafez:
@Uzlopak 
Can you elaborate on what the issue is? Can you think of a potential fix for the issue?

vkarpov15 (approved):
I took a look and this works for me. @Uzlopak do you have any concrete objections? If not, I'd like to merge this over the next couple of days.

Uzlopak:
I think, the main issue here is, that it is a export * from 'mongo'. So not only types are exported but also classes and stuff. So this is possible:

```typescript
import { Collection } from 'mongoose/mongodb-driver';

const conn = new Collection();
```

but we dont expose mongodb-driver imho

AbdelrahmanHafez:
I haven't tested this, but I assumed because this was a `.d.ts` file, TS would only export types.
Does this actually work? Do you happen to know if there is a way to _only_ export types?

vkarpov15 (changes requested):
@AbdelrahmanHafez I'm unfortunately unable to get this PR working. The tests do pass, but when I compile the below script locally:

```ts
import { AggregateOptions } from 'mongoose/mongodb-driver';

const options: AggregateOptions = { };
```

I get the below error:

```
$ ./node_modules/.bin/tsc gh-11954.ts 
gh-11954.ts:1:34 - error TS2307: Cannot find module 'mongoose/mongodb-driver' or its corresponding type declarations.

1 import { AggregateOptions } from 'mongoose/mongodb-driver';
                                   ~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in gh-11954.ts:1
```

`typescript@4.7.4` . Can you try compiling this script locally and see if you get something different? :confused:

AbdelrahmanHafez:
@vkarpov15 
Sorry for the delayed response. I've been occupied the past couple of months.

So, yes, I get the same error while trying to compile the script above, but I also get the same error when I try the following
## gh-11954.ts
```ts
import { Schema } from 'mongoose';
```

## Output
```bash
$ ~/dev/mongoose  ./node_modules/.bin/tsc gh-11954.ts
gh-11954.ts:1:24 - error TS2307: Cannot find module 'mongoose' or its corresponding type declarations.

1 import { Schema } from 'mongoose';
                         ~~~~~~~~~~


Found 1 error in gh-11954.ts:1
```


Within the context of the _root of the package_, ts can not resolve `mongoose`, let alone `mongoose/mongo-driver`.
I'm not sure how we can test this as if it were released without actually releasing a new version.

I believe that testing within the `types/test` context should be enough. What do you think? @vkarpov15

vkarpov15:
I'm going to close this one for now. Haven't been able to get it to work, and this PR is pretty stale. We'll reconsider this in the future.

PR #11956, opened 2022-06-20
types: Update Expression Typings

### LINKED ISSUE #11952: aggregate $match Error: Type 'ObjectId' is not assignable to type 'Expression'.
### Prerequisites

### Mongoose version

6.4.0

### Node.js version

16.14.0

### MongoDB server version

5.0.9

### Description

update to latest version and get error on aggregate
```javascript
  await Model.aggregate([
                  {
                      $match: { _id: new mongoose.Types.ObjectId('stringObjecId') },
                  },
  ]);
```
get Error `Type 'ObjectId' is not assignable to type 'Expression'.`
how to solve this?

### Steps to Reproduce

```javascript
  await Model.aggregate([
                  {
                      $match: { _id: new mongoose.Types.ObjectId('stringObjecId') },
                  },
  ]);
```

### Expected Behavior

_No response_

andreialecu:
**Summary**

Fixes #11952

### DIFF
--- a/types/expressions.d.ts
+++ b/types/expressions.d.ts
@@ -2391,7 +2391,7 @@ declare module 'mongoose' {
         *
         * @see https://www.mongodb.com/docs/manual/reference/operator/query/exists/
         */
-      $exists: boolean;
+      $exists: boolean | NullExpression;
     }
   }
 
--- a/types/expressions.d.ts
+++ b/types/expressions.d.ts
@@ -1,3 +1,5 @@
+import type { ObjectIdLike } from 'bson';
+
 declare module 'mongoose' {
 
   /**
@@ -1034,7 +1036,7 @@ declare module 'mongoose' {
        *
        * @see https://docs.mongodb.com/manual/reference/operator/aggregation/ne/#mongodb-expression-exp.-ne
        */
-      $ne: Expression | [Expression, Expression];
+      $ne: Expression | [Expression, Expression] | null;
     }
 
     export interface Cond {
@@ -2381,9 +2383,20 @@ declare module 'mongoose' {
         lang: 'js'
       };
     }
+
+    export interface Exists {
+      /**
+        * When <boolean> is true, $exists matches the documents that contain the field, including documents where the field value is null.
+        * If <boolean> is false, the query returns only the documents that do not contain the field.
+        *
+        * @see https://www.mongodb.com/docs/manual/reference/operator/query/exists/
+        */
+      $exists: boolean;
+    }
   }
 
-  type Path = string;
+  type Path = string | ObjectIdLike;
+
 
   export type Expression =
     Path |
@@ -2428,7 +2441,8 @@ declare module 'mongoose' {
     DateExpression |
     BinaryExpression |
     FunctionExpression |
-    ObjectIdExpression;
+    ObjectIdExpression |
+    ConditionalExpressionOperator;
 
   export type ObjectIdExpression =
     TypeExpressionOperatorReturningObjectId;
@@ -2594,7 +2608,8 @@ declare module 'mongoose' {
   export type ConditionalExpressionOperator =
     Expression.Cond |
     Expression.IfNull |
-    Expression.Switch;
+    Expression.Switch |
+    Expression.Exists;
 
   export type StringExpressionOperator =
     StringExpressionOperatorReturningArray |
--- a/test/types/PipelineStage.test.ts
+++ b/test/types/PipelineStage.test.ts
@@ -1,3 +1,4 @@
+import { ObjectId } from 'mongodb';
 import type { PipelineStage } from 'mongoose';
 import { expectError, expectType } from 'tsd';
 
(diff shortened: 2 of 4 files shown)


types/expressions.d.ts
```diff
@@ -2381,31 +2383,41 @@ declare module 'mongoose' {
...
-    TrigonometryExpressionOperator |
-    TypeExpressionOperator |
-    AccumulatorOperator |
-    VariableExpressionOperator |
-    WindowOperator;
+  type Path = string | ObjectIdLike;
+
+
+  export type Expression = Path |
+  ArithmeticExpressionOperator |
```
andreialecu, line 2402: Screenshot 2022-06-20 at 11 48 38

Not sure why this got reformatted, but I'm getting this otherwise.

types/expressions.d.ts
```diff
@@ -2381,9 +2383,20 @@ declare module 'mongoose' {
...
     }
+
+    export interface Exists {
+      /**
+        * When <boolean> is true, $exists matches the documents that contain the field, including documents where the field value is null.
+        * If <boolean> is false, the query returns only the documents that do not contain the field.
+        *
+        * @see https://www.mongodb.com/docs/manual/reference/operator/query/exists/
+        */
+      $exists: boolean;
```
Uzlopak, line 2394:
```suggestion
      $exists: boolean | NullExpression;
```

types/expressions.d.ts
```diff
@@ -2381,9 +2383,20 @@ declare module 'mongoose' {
...
+    }
   }
 
-  type Path = string;
+  type Path = string | ObjectIdLike;
```
Uzlopak, line 2398: No. Path is just a Path. Not ObjectIdLike.

```suggestion
  type Path = string;
```

Uzlopak: I created a separate type, so that one day if we have good way to generate valid paths by mongoose Schema we would already have it properly encapsulated it.

types/expressions.d.ts
```diff
@@ -2594,7 +2608,8 @@ declare module 'mongoose' {
   export type ConditionalExpressionOperator =
     Expression.Cond |
     Expression.IfNull |
-    Expression.Switch;
+    Expression.Switch |
+    Expression.Exists;
```
Uzlopak, line 2612: No this is not a Conditional Expression Operator
see https://www.mongodb.com/docs/manual/reference/operator/aggregation/#conditional-expression-operators

Uzlopak:
```suggestion
```

types/expressions.d.ts
```diff
@@ -2594,7 +2608,8 @@ declare module 'mongoose' {
   export type ConditionalExpressionOperator =
     Expression.Cond |
     Expression.IfNull |
-    Expression.Switch;
+    Expression.Switch |
```
Uzlopak, line 2611:
```suggestion
    Expression.Switch;
```

types/expressions.d.ts
```diff
@@ -2428,7 +2441,8 @@ declare module 'mongoose' {
     DateExpression |
     BinaryExpression |
     FunctionExpression |
-    ObjectIdExpression;
+    ObjectIdExpression |
+    ConditionalExpressionOperator;
```
Uzlopak, line 2445: I think this is kind of wrong.

Uzlopak (commented):
We have to think about this more

andreialecu:
There's also a problem with booleans, like `{$ne: true}`, and I'd assume numbers as well.

```
Type '{ $ne: true; }' is not assignable to type 'Expression | undefined'.
              Types of property '$ne' are incompatible.
                Type 'true' is not assignable to type 'Expression | [Expression, Expression] | undefined'.
```

Uzlopak:
Well it is wrong because in an aggregation pipeline $ne is an array with two values, which are compared.

Uzlopak:
https://www.mongodb.com/docs/manual/reference/operator/aggregation/ne/

andreialecu:
I think `$match` is defined incorrectly now, and it needs to be something like this instead:

```ts
    export interface Match {
      /** [`$match` reference](https://docs.mongodb.com/manual/reference/operator/aggregation/match/) */
      $match: FilterQuery<any> & { $expr?: Expression };
    }
```

types/expressions.d.ts
Jule-: I am not aware of all changes but it seems that `$exists` is not accepted anymore by the typings and despite the fact that it is not shown in the MongoDB documentation (oversight?) I think it is still a valid operator for `$match` query in the aggregation pipeline.

Wherever you put it, I think this should work:
```ts
const result = await Model.aggregate([{
  $match: { optional: { $exists: true } }
}]);
```
mongoose@6.4.0:
```
Type '{ $exists: true; }' is not assignable to type 'Expression'.
```

vkarpov15 (approved):
I'll merge this for now to unblock @andreialecu .

However, there are still some issues with the expression typings. For one thing @Uzlopak , I think we'll have to add `AnyObject`, because the following code will still fail:

```javascript
  {
    $match: {
      _id: new ObjectId('stringObjecId'),
      a: { $exists: true },
      b: null,
      c: 'test',
      d: { foo: 42 } // <-- valid, but current typings will fail
    }
  }
```

Uzlopak:
I'll be honest:
I think we should revert the expressions PR. It needs more work to work properly.

lucamodica:
Does $nin operator still exists? For the moment I have to workaround with

`{ $not: { $in: [<expression>, <arrayExpression>] } }`

vkarpov15:
@lucamodica yes `$nin` still is an operator on the MongoDB server. Does `$nin` not work for you in Mongoose 6.4?

lucamodica:
Nope, or at least both `$nin` and `$exists` are not recognized as `Expression` .

andreialecu:
https://github.com/Automattic/mongoose/pull/11969 should fix this.

PR #11965, opened 2022-06-21
[Types] Re-add the possibility to pass undefined for projection in Model.find

ghost91-:
**Summary**

This re-adds the possibility to pass `undefined` as projection to the `find` method of models.

**Examples**

```ts
  Project.find({}, undefined, { limit: 5 });
```
(see the added tests for more details)

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -1,5 +1,5 @@
 import { ObjectId } from 'bson';
-import { Schema, Document, Model, connection, model, Types } from 'mongoose';
+import { Schema, Document, Model, connection, model, Types, CallbackError } from 'mongoose';
 import { expectError, expectType } from 'tsd';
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

AbdelrahmanHafez:
@ghost91- 
Should we also do the same for .findOne/findOneAndUpdate/findById?

ghost91-:
For those, it's already done, if I am not mistaken. `undefined` is not listed explicitly, but the parameters marked optional. Unless you are using the `exactOptionalPropertyTypes` option, it results in the same thing. I just tend to specify `undefined` explicitly most of the time because I sometimes use that option.

Granted, there also are no tests for the other methods, so it may be worth adding some.

Issue #11977, opened 2022-06-22
6.4 has compile errors on Typescript 4.8 beta


sandersn:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.4

### Node.js version

16.15

### MongoDB server version

4.7

### Description

Mongoose 6.4 uses the new type `MergeType` throughout models.d.ts. This type requires its type arguments to `extend Record<string | number, any>`, but none of them are! It's used 16 times in models.d.ts and all of its arguments are unconstrained type parameters, like `T`, which could possibly allow `null` or `undefined`.

For more background, see: https://github.com/microsoft/TypeScript/issues/49489

### Steps to Reproduce

Clone mongoose
Open types/models.d.ts in an editor running Typescript 4.8.

You can get the same errors when compiling a package that depends on mongoose, like `@types/mongoose-unique-validator`. (I discovered this error by reviewing failures on Definitely Typed.)

### Expected Behavior

No compile errors on Typescript 4.8.

rongem:
I also have compiling errors after upgrading to mongoose 6.4 under typescript 7.4. It seems that mongoose changed the way, complex parameters (i. e. in find method) are interpreted. Return types for many functions are now no longer the type, but Omit<Merge<somthing>>.

sandersn:
I don't know mongoose, but its new `MergeType` is a better reflection of object spread semantics or `Object.assign` semantics than before, since when you spread `{ ...A, ...B}`, B's properties overwrite any of A's that conflict.

PR #11978, opened 2022-06-22
Fix: simplify MergeType constraints

sandersn:
Fixes 16 errors in models.d.ts on Typescript 4.8 which is currently in beta. Typescript 4.8 forbids assignment of an unconstrained type
parameter to one that disallows `null` or `undefined`: https://github.com/microsoft/TypeScript/issues/49489

This is the minimal bug fix: `MergeType` now allows `null | undefined`, just like it does on shipping versions of Typescript. The result of `MergeType` can be unexpected when used with `undefined` or `SomeObjectType | undefined`, but this might not be a problem in practise.

There are two other possible fixes:

1. Constrain *lots* of types in models.d.ts to `Record<string | number, any>` -- that is, to forbid `null | undefined`. I don't know mongoose -- this might be a huge API break.
2. Make MergeType correctly handle `null` and `undefined`: it's pretty clear that `MergeType<A, null> === A` but I'm not sure whether `MergeType<A | undefined, B>` should be something like `(Omit<A, keyof B> & B) | undefined`.

**Summary**

Fixes Automattic/mongoose#11977 in a way that doesn't improve correctness, but doesn't break the API either.

### DIFF
--- a/types/utility.d.ts
+++ b/types/utility.d.ts
@@ -10,6 +10,6 @@ declare module 'mongoose' {
       ? T
       : Omit<T, keyof U> & U;
 
-  type MergeType<A extends Record<number | string, any>, B extends Record<number | string, any>> = Omit<A, keyof B> & B;
+  type MergeType<A, B> = Omit<A, keyof B> & B;
 
 }


vkarpov15 (approved):
Thanks :+1:

PR #11984, opened 2022-06-24
Fixes to connection typings

### LINKED ISSUE #11983: types: wrong types related to mongoose.connection
### Prerequisites

### Mongoose version

6.4.0

### Node.js version

16

### MongoDB server version

5

### Description

After upgrading a few mongoose versions, some of our typings broke.

See the repro, add those lines to `connection.test.ts`

### Steps to Reproduce

Add these lines to `test/types/connection.test.ts`. We use this in unit tests and the model isn't important.

```ts
expectType<Promise<void>>(conn.model('test').ensureIndexes()); // Object is of type 'unknown'.ts(2571)

expectType<Promise<mongodb.InsertManyResult<AnyObject>>>(
  conn.collection('test').insertMany([{ _id: 'string' }]) // Type 'string' is not assignable to type 'ObjectId'.ts(2322)
);
```

### Expected Behavior

Both should be allowed.

andreialecu:
**Summary**

Fixes #11983

### DIFF
--- a/test/types/connection.test.ts
+++ b/test/types/connection.test.ts
@@ -1,4 +1,4 @@
-import { createConnection, Schema, Collection, Connection, ConnectionSyncIndexesResult, Model, AnyObject, AnyObjectAnyId } from 'mongoose';
+import { createConnection, Schema, Collection, Connection, ConnectionSyncIndexesResult, Model, AnyObjectAnyId } from 'mongoose';
 import * as mongodb from 'mongodb';
 import { expectError, expectType } from 'tsd';
 
--- a/test/types/connection.test.ts
+++ b/test/types/connection.test.ts
@@ -1,4 +1,4 @@
-import { createConnection, Schema, Collection, Connection, ConnectionSyncIndexesResult, Model } from 'mongoose';
+import { createConnection, Schema, Collection, Connection, ConnectionSyncIndexesResult, Model, AnyObject, AnyObjectAnyId } from 'mongoose';
 import * as mongodb from 'mongodb';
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -108,6 +108,8 @@ declare module 'mongoose' {
     [k: string]: any
   }
 
+  export type AnyObjectAnyId = AnyObject & { _id?: string | number | boolean | object };
```
vkarpov15, line 111: Why is this necessary? Can't we do `_id?: any` ?

andreialecu: `{ _id?: any }` doesn't work; it was what I initially tried. There's a conditional type somewhere that doesn't agree with that. :)

And `AnyObject` should already include `{ _id?: any }` implicitly by how it's defined:
```
  export interface AnyObject {
    [k: string]: any
  }
```

You can try replacing it with `_id?: any`; you can see that the error comes back.

test/types/connection.test.ts
```diff
@@ -1,4 +1,4 @@
-import { createConnection, Schema, Collection, Connection, ConnectionSyncIndexesResult, Model } from 'mongoose';
+import { createConnection, Schema, Collection, Connection, ConnectionSyncIndexesResult, Model, AnyObject, AnyObjectAnyId } from 'mongoose';
```
andreialecu, line 1:
```suggestion
import { createConnection, Schema, Collection, Connection, ConnectionSyncIndexesResult, Model, AnyObjectAnyId } from 'mongoose';
```

PR #11563, opened 2022-03-24
Feature: Infer schema type automatically.

mohammad0-0ahmad:
# Tasks to be implemented :
- [x] Infer schema type automatically from document definition.
- [x] Provide TS utility to be able to extract schema type from schema instance to be able to reuse the type as separated type "InferSchemaType".
- [x] Improve Model.create function type to show schema properties in suggestions dialog.
- [x]  Adding "statics"  property in schema options  to support auto type inference for model statics methods.
- [x]  Adding "methods"  property in schema options  to support auto type inference for document instance methods.
- [x] Support custom type key "typeKey" to inference schema type.
- [x] Improve Model new function "document constructor" type to show schema properties in suggestions dialog.
- [x] Improve Model.insertMany function type to show schema properties in suggestions dialog.
- [x]  Adding "query"  property in schema options  to support auto type inference for query helpers methods.

# Status:
- [x]  Ready to review.
- [x]  Docs is updated.
- [x]  Ready to merge.

Do you have any suggestions or ideas to be implemented in this PR, please leave a comment :slightly_smiling_face:

### Get it early:
You can install mongoose including this feature by running:
```shell
 npm i Automattic/mongoose#pull/11563/head
```
# Screenshots :
- Auto schema type inference & create function:

Untitled

- InferSchemaType utility & supporting custom type key:

Untitled

PR #11986, opened 2022-06-24
fix(types): avoid adding non-existent properties from model constructor for typegoose

### LINKED ISSUE #11960: [Types] mongoose 6.4.0 (PR 11563) broke explicitly set property types
### Prerequisites

### Mongoose version

6.4.0

### Node.js version

any

### MongoDB server version

any

### Description

With mongoose 6.4.0 (i think because of #11563) the explicitly set property types are ignored and so break subdocuments, subdocument arrays and maps (and probably other types that require more than explicit objects or get transformed)

### Steps to Reproduce

1. clone https://github.com/typegoose/typegoose-testing/
2. checkout branch `mongoose6.4.0typebreak`
3. run `yarn install` (or equivalent)
4. inspect the types either via something like vscode or run `yarn run build` (or equivalent)

this branch also contains commits for previous versions:
- [`ef07ac6df32938f30cadbbb955b5f0af0301b536`](https://github.com/typegoose/typegoose-testing/tree/ef07ac6df32938f30cadbbb955b5f0af0301b536) for a typegoose inital example (mongoose 6.3.9) (`HEAD~2`)
- [`23d7787e5101cdee81c267cda0f4745c70b6a46f`](https://github.com/typegoose/typegoose-testing/tree/23d7787e5101cdee81c267cda0f4745c70b6a46f) for a mongosoe native (no typegoose) example at 6.3.9 (`HEAD~1`)
- [`fc31453a5463f03d434fb1459dfb26ce4a770bbb`](https://github.com/typegoose/typegoose-testing/tree/fc31453a5463f03d434fb1459dfb26ce4a770bbb) for a mongoose native (no typegoose) example at 6.4.0 (`HEAD`)

### Expected Behavior

All types to work like in the previous minor version / explicitly set types to overwrite inferred types

vkarpov15:
Fix #11960

Please take a look @mohammad0-0ahmad

**Summary**

In #11563, it looks like the model constructor pulls types from the parameter to the model constructor, which is very wrong. @hasezoey reported that this also causes issues for Typegoose.

The tests in `document.test.ts#L195-L196` are also wrong, `unExistProperty` shouldn't be defined on the document instance. See the below script:

```javascript

void async function main() {
  await mongoose.connect('mongodb://localhost:27017');

  const Test = mongoose.model('Test', mongoose.Schema({ name: String }));
  const doc = new Test({ name: 'test', unExistProperty: 1 });
  console.log(doc);
  console.log(doc.name); // 'test'
  console.log(doc.unExistProperty); // undefined

  await mongoose.disconnect();
}();
```

**Examples**

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,6 +24,7 @@
       ],
       "rules": {
         "@typescript-eslint/triple-slash-reference": "off",
+        "@typescript-eslint/no-non-null-assertion": "off",
         "spaced-comment": [
           "error",
           "always",
--- a/test/types/document.test.ts
+++ b/test/types/document.test.ts
@@ -197,3 +197,44 @@ function autoTypedDocument() {
   expectType<ReturnType<AutoTypedSchemaType['methods']['instanceFn']>>(new AutoTypedModel().instanceFn());
 
 }
+
+async function gh11960() {
+  type DocumentType<T> = Document<any> & T;
+  type SubDocumentType<T> = DocumentType<T> & Types.Subdocument;
+  type ArraySubDocumentType<T> = DocumentType<T> & Types.ArraySubdocument;
+
+  interface Nested {
+    dummy?: string;
+  }
+
+  interface Parent {
+    username?: string;
+    map?: Map<string, string>;
+    nested?: SubDocumentType<Nested>;
+    nestedArray?: ArraySubDocumentType<Nested>[];
+  }
+
+  const NestedSchema = new Schema({
+    dummy: { type: String }
+  });
+
+  const ParentSchema = new Schema({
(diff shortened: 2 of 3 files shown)


mohammad0-0ahmad:
Hello @vkarpov15! 
The changes look good to cover the regression.
Because of the requirements, you described "arbitrary objects in new and create and insert".
I would like to say this changes make it better, and I would prefer as will to prevent adding "arbitrary objects" when strict mode is enabled. 
I will consider making that available.
Thanks for fixing this regression @vkarpov15, and sorry for the trouble @hasezoey, and thanks for  reporting the issue. 
Best Regards!

PR #11997, opened 2022-06-28
fix(types): change index "weights" to be more explicit

hasezoey:
**Summary**

Currently the types for index option `weights` was defined as `AnyObject` which allows only string keys and any value type, but with this change `string | number` keys are allowed (i dont know if indexes support this, please review) and explicitly set the value type to `number`, because that is the only type mongodb accepts (from what i know)

**Examples**

```ts
sch.index({ field1: 1 }, { weights: { field1: 10 } });
```

### DIFF
--- a/types/indexes.d.ts
+++ b/types/indexes.d.ts
@@ -93,6 +93,6 @@ declare module 'mongoose' {
      * ```
      */
     expires?: number | string;
-    weights?: Record<string | number, number>;
+    weights?: Record<string, number>;
   }
 }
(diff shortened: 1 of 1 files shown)


types/indexes.d.ts
```diff
@@ -93,6 +93,6 @@ declare module 'mongoose' {
      * ```
      */
     expires?: number | string;
-    weights?: AnyObject;
+    weights?: Record<string | number, number>;
```
hasezoey, line 96: i am unsure if `string | number` is actually supported, should this maybe be changed to just `string`?

Uzlopak: I would say string only.

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

PR #11998, opened 2022-06-28
add eslint rule `eol-last` to change EOL to be consistent

hasezoey:
**Summary**

This PR adds eslint rule `eol-last` (on level `warn`) to change files to be consistent because before some files had a empty last line and some did not (it is just warn to not interrupt development but also easily fixed with a `lint --fix`)

PS: i dont know if this rule has to be added somewhere else too, like a prettier or vscode config


hasezoey:
the tests just failed, but i dont think the failure is related to this PR's changes, could someone with permissions maybe re-run failed jobs?

AbdelrahmanHafez (approved):
LGTM, thanks 👍

Uzlopak:
Thanks.

PR #12001, opened 2022-06-28
Extend types test for #11960

hasezoey:
**Summary**

This PR extends the types test that was initially added for #11960, with more tests and another case

**WARNING**: this PR is currently set to fail because of the new case, so i dont think it is good to merge it yet

---

Edit:
the first commit extends the initial test case to actually test what the reproduction repository of the mentioned issue had problematic
the second commit duplicates the test for `new` to also test `create` (which was not updated yet)


vkarpov15 (approved):
Thanks :+1: I'll put in a fix for this in master

PR #12009, opened 2022-06-29
Fix broken h4 markdown tags for example in documentation 

AshikNesin:
**Summary**
For some of the methods which are mentioned below we didn't  give space between markdown tag as a result it's not being rendered properly in the documentation.

I've searched the entire codebase, it seems like we've issues in the following methods which are addressed in this PR.
- skipMiddlewareFunction
- overwriteMiddlewareResult
- discriminator
- removeIndex
- clearIndexes

**Examples**
For example, this is how the api for `skipMiddlewareFunction` is rendered right now

CleanShot 2022-06-29 at 10 12 43@2x

And after the fix, it'll be

CleanShot 2022-06-29 at 10 11 10@2x

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1223,7 +1223,7 @@ Mongoose.prototype._promiseOrCallback = function(callback, fn, ee) {
 /**
  * Use this function in `pre()` middleware to skip calling the wrapped function.
  *
- * ####Example:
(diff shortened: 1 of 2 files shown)


Uzlopak:
Thanks

PR #12022, opened 2022-06-30
Add ESLint rule `no-multiple-empty-lines` and apply it

hasezoey:
**Summary**

This change is to make coding style more consistent by only allowing up to `2` empty lines after each other

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -187,6 +187,7 @@
       "error"
     ],
     "no-empty": "off",
-    "eol-last": "warn"
+    "eol-last": "warn",
+    "no-multiple-empty-lines": ["warn", { "max": 2 }]
   }
 }
--- a/test/model.querying.test.js
+++ b/test/model.querying.test.js
@@ -1838,7 +1838,6 @@ describe('model: querying:', function() {
     const [a, b, c, d] = await Test.create([docA, docB, docC, docD]);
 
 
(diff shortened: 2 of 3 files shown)


.eslintrc.json
```diff
@@ -187,6 +187,7 @@
       "error"
     ],
     "no-empty": "off",
-    "eol-last": "warn"
+    "eol-last": "warn",
+    "no-multiple-empty-lines": ["warn", { "max": 2 }]
```
hasezoey, line 191: should `max` maybe be set to `1` instead of `2`?

i initially had tried this, but there were so many cases found where `2` were used and so i thought it is part of the style, but wanted to ask before merge

hasezoey:
i think the failed test has nothing to do with this PR, so maybe a re-run of the test would be needed

.eslintrc.json
AbdelrahmanHafez: I disagree with this `max` value, I think that _strictly_ limiting to 2 lines is a bit too much. I usually allow up to 5 lines in my ESLint config to prevent people from going crazy with more lines than unnecessary. But I've found situations where I want to add up to 4~5 empty lines to make it visually clear that pieces of code are separated in the same file.

vkarpov15 (approved):
Thanks :+1:

Issue #8925, opened 2020-05-03
Use something other than dox to parse documentation


vkarpov15:
Re: #8857, dox is long out of date and has a few upstream security vulnerabilities. Shouldn't impact users since dox is a dev dependency, but it causes confusion and reflects badly on us.

hasezoey:
once https://github.com/tj/dox/pull/196 lands, would this issue still be needed?

the only "active" alternative i see is to use "typedoc", but for that to properly function, some typescript errors would have to be resolved

Twipped:
https://github.com/tj/dox/releases/tag/v1.0.0

vkarpov15:
I'll close this one for now, given #12024. Thanks @hasezoey !

PR #12024, opened 2022-06-30
Update `dox` to latest version and fix some documentation issues

hasezoey:
**Summary**

This PR updates dependency `dox` from `0.3.1`(from 2015) to `0.9.1`(from 2022), which should address #8925 at least a little

re #8925

---

This PR: 
- upgrades dox through the versions, a commit made each time a "minor" version was updated or when a change was required.
- does **not** guarantee that all anchors or stuff are the same, i had modified some things to be more consistent so some could have changed.
- should make debugging easier when dox fails with a error (5de8b9aff41816bfdae5711a788a3ab190c543f3)
- adds `yarn.lock` got `gitignore` (because `package-lock.json`was already in it)
- fixes empty `@returns` tags to be properly set (meaning to have types)
- fixes wrong array usage (things like `[Type]` were used when jsdoc only allows `Type[]` or `Array<Type>`)
- removes function definitions from `VirtualType.prototype.set` and `VirtualType.prototype.get` because they were not proper jsdoc (also dox did not want to correctly register things like `@callback` or `@typedef` or `@type` with `@name`) (7155089f33267dd0c468c35a6dff61176c9b04ed)
- adds many `@memberOf` tags to many jsdoc comments (because they were missing and resulting in wrong documentation)
- fixes `SchemaType.prototype.castFunction` being listed as `cast`
- removes `@static` and `@function` from `SchemaType.prototype.castFunction`
- moves the jsdoc for `DocumentArrayPath.set` and `SubdocumentPath.set` to the correct location
- replaces all (that i found) occurrences of `- ` in `@param` descriptions, because they were making the documentation formatting weird (e57594c7a72dc21f91d3fbda2742cf0132e03042)
- replaces all found occurrences of `@param {*}` with `@param {Any}` to be more consistent and also because `*` was making the documentation formatting weird (953853f46647cba203f576e08883cc44da61f844)
- removes all text after `@static` because in later versions of `dox` nothing was given to the processing anymore and was also not proper jsdoc (but things are still defined thanks to other things) (1719bf6dc4f1ab77640b8f8d3e12ad94004e9461)
- adds a new property to `ctx` in `docs/source/api.js` named `isFunction` that after processing adds `()` to the end of the `ctx.string`, this was done because things like `@function` and `@static` could be in a different sequence than expected
- changes to use `ctx.constructor` over `data.name` for static properties (to be consistent with non-static properties) (cfe72488bd28b2c818ef87e86a87240523687b4b)
- removes a debug log of the `ctx` (39a02a18d1840d68868b54acd91d392e9640088c)

PS: this PR also fixes some documentation issues that are visible in current `master`

***Note: please only someone with knowledge of how the documentation works merge this PR***


hasezoey:
Note: there are currently conflicts with the base branch, but i will resolve them once it is ready for merge (after having had reviews)

lib/connection.js
```diff
@@ -1452,7 +1452,7 @@ Connection.prototype.setClient = function setClient(client) {
  *
  * @param {Object} options
  * @param {Boolean} options.continueOnError `false` by default. If set to `true`, mongoose will not throw an error if one model syncing failed, and will return an object where the keys are the names of the models, and the values are the results/errors for each model.
- * @returns
+ * @return {Promise|undefined} Returns `undefined` if callback is specified, returns a promise if no callback, when the Promise resolves the value is a list of the dropped indexes.
```
Uzlopak, line 1455: it does not contain a callback?! Also it is defined as async function, so it is always a Promise?!

lib/error/index.js
```diff
@@ -73,7 +73,7 @@ MongooseError.Messages = MongooseError.messages;
  *
  * @api public
  * @memberOf Error
- * @static DocumentNotFoundError
+ * @static
```
Uzlopak, line 76: shouldnt we then add @name DocumentNotFoundError?

lib/connection.js
hasezoey: i dont really know about that, i just copied it from `lib/model.js`, where i also had slightly extended the description, and it just seemed to return everything from what `lib/model.js` returned

lib/index.js
```diff
@@ -1044,7 +1046,7 @@ Mongoose.prototype.isObjectIdOrHexString = function(v) {
  *
  * @param {Object} options
  * @param {Boolean} options.continueOnError `false` by default. If set to `true`, mongoose will not throw an error if one model syncing failed, and will return an object where the keys are the names of the models, and the values are the results/errors for each model.
- * @returns
+ * @return {Promise|undefined} Returns `undefined` if callback is specified, returns a promise if no callback, when the Promise resolves the value is a list of the dropped indexes.
```
Uzlopak, line 1049: We dont pass callback here

lib/error/index.js
hasezoey: it is already added by `jsdocparsecomments` / `dox` automatically, i dont quite know where it is fetching that information from, but it is there (probably the next line after the comment?)

also by jsdoc standard, `@static` does not have any parameters or really a description

lib/connection.js
hasezoey: updated in 407fe6194e4f81aba170b6b88dd4b987c8bdb789, is this more correct?

lib/index.js
hasezoey: updated in 407fe6194e4f81aba170b6b88dd4b987c8bdb789, is this more correct?

docs/source/api.js
```diff
@@ -147,26 +193,49 @@ function parse() {
...
+            ctx.string = tag.string;
             ctx.name = tag.string;
+            ctx.isFunction = true;
+            break;
+          case 'instance':
+            ctx.isInstance = true;
+            break;
+          case 'deprecated':
+            ctx.deprecated = true;
+            break;
```
hasezoey, line 211: i have now included basic handling for `@deprecated` tag, but i am not sure where or in what style it should appear in the documentation

docs/source/api.js
```diff
@@ -147,26 +193,49 @@ function parse() {
...
+
+      // add "()" to the end of the string if function
+      if ((ctx.isFunction || ctx.type === "method") && !ctx.string.endsWith("()")) {
+        ctx.string = ctx.string + "()";
+      }
+
+      // fixing up "something.prototypemissingdot" to "something.prototype.missingdot"
       if (/\.prototype[^.]/.test(ctx.string)) {
         ctx.string = `${ctx.constructor}.prototype.${ctx.name}`;
       }
```
hasezoey, line 236: could this maybe be removed?

hasezoey (commented):
I have now updated the branch to include the latest master again (because documentation changes were made), and i also have touched-up some of the documentation again, which includes:
- adding some comments to signal what is doing what
- move most string manipulation handling to after the "tags" loop
- add a jsdoc "typedef" to remember all properties
- add a proper type for `@returns {void}` or `{undefined}` as to not result in empty fields

Also some other discussion questions:
- should tags and handling for `@receiver` be removed and be replaced with the proper tags everywhere?
- should `{Any}` maybe be replaced again everywhere with `{*}`? because by jsdoc definition `{*}` is the proper "any" type, but currently otherwise causing some documentation style problems

hasezoey:
Made another small update which does:
- fix `Aggregate.prototype.lookup` not having a proper next line for a jsdoc tag
- re-adding `null` to listed types when `nullable`
- fixing `Query.prototype.box` jsdoc to be proper jsdoc and extending the comment for the parameters (please check if correct, i am not familiar with that function)
- fixing `@type` tags not being properly escaped with `{` and `}`

hasezoey:
I would say this PR is finished for changes, unless some are requested / merging latest master, some changes could probably be put in their own PR / cherry-picked to land sooner.

---

i would like to request @vkarpov15 to review this PR

vkarpov15 (approved):
Code is LGTM. I'll review this more after I merge, it's easier to test after merging to master because we don't have deployment previews.

hasezoey:
> we don't have deployment previews.

isnt the script `docs:view` a deployment preview or do you maybe mean compiled documentation diff?

docs/source/api.js
hasezoey: continued in https://github.com/Automattic/mongoose/pull/12071

Issue #12030, opened 2022-07-01
schema type auto infer bugs


imranbarbhuiya:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.4.2

### Node.js version

18

### MongoDB server version

5.0

### Description

I've found a few more bugs in auto infer type

1. Adding an object directly to the array doesn't infer type correctly

```ts
const userSchema = new Schema({
	users: [
		{
			username: { type: String }
		}
	]
});
```

2. passing function in default makes the key optional

```ts
const userSchema = new Schema({
	createdAt: { type: Date, default: Date.now }
});
```

3. Array type becomes `Array` instead of  `Types.DocumentArray` which makes pushing fields difficult

```ts
const userSchema = new Schema({
	users: [
		new Schema({
			username: { type: String },
			credit: { type: Number, default: 0 }
		})
	]
});

const userModel = model('user', userSchema);

const data = await userModel.findOne();
data!.users.push({ username: 'parbez' });
// ts warns to add createdAt which can be solved using Types.DocumentArray
```
4. Adding empty object as default doesn't makes the key required
```ts
const userSchema = new Schema({
	data: { type: { role: Object }, default: {} }
});
```

5. Assigning an object directly to a key doesn't infer type correctly

```ts
const userSchema = new Schema({
	data: {
		username: String
	}
});
```

### Steps to Reproduce

added code blocks above

mohammad0-0ahmad:
@imranbarbhuiya  Thanks for reporting this one as well, I will check the new changes and see what we can do 👍

imranbarbhuiya:
I've updated code block 4 and added the 5th point which is the same as 1 but for an object.

shiva-prasad-reddy:
```
const schema = new Schema({
  track: {
    backupCount: {
      type: Number,
      default: 0,
    },
    count: {
      type: Number,
      default: 0,
    },
  },
});
type ISchema = InferSchemaType<typeof schema>;
```

**resulting to a type of**
```
type ISchema = {
    track: {
        backupCount: {
            type: NumberConstructor;
            default: number;
        };
        count: {
            type: NumberConstructor;
            default: number;
        };
    };
}
```

**but it should be**
```
type ISchema = {
    track: {
        backupCount: NumberConstructor;
        count: NumberConstructor;
    };
}
```

mohammad0-0ahmad:
> ```
[...]

PR #12042, opened 2022-07-02
Improve Auto typed schema.

mohammad0-0ahmad:
Related to: [12030](https://github.com/Automattic/mongoose/issues/12030)

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -540,7 +540,7 @@ export type AutoTypedSchemaType = {
     favoritDrink?: 'Tea' | 'Coffee',
     favoritColorMode: 'dark' | 'light'
     friendID?: Types.ObjectId;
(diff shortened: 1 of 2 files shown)


test/types/schema.test.ts
```diff
@@ -630,3 +630,75 @@ function gh11987() {
...
+      },
+      count: {
+        type: Number,
+        default: 0
+      }
+    }
+  });
+
+  expectType<{
+    track?: {
```
vkarpov15, line 698: As per discussion in #12030, this should ideally be `track: { backupCount: number; count: number; }`. But I don't think that's significant enough to block merging and shipping this PR, because it's a meaningful improvement. Feel free to put in a separate PR.

vkarpov15 (approved):
Thanks :+1:

PR #12048, opened 2022-07-04
fix benchmark

lpizzinidev:
- [x] Fixed update.js benchmark formatting and implementation


Uzlopak:
Are now all benchmarks refactored and fixed?

lpizzinidev:
The `/benchjs` folder is done.
I will check for bottlenecks and other possible improvements in the next few days if it is ok.

lpizzinidev:
@Uzlopak After analyzing the benchmarks I found that `mongoose` reading ops/sec is actually worse than the driver's one (file `read.js`).
Looks to me like the `find` operation is slower.

PR #12054, opened 2022-07-05
Update schematypes.md

MontgomeryWatts:
Expand description of `min` and `max` and include `expires` option for Date schema type.

**Summary**

The documentation for [SchemaType Options](https://mongoosejs.com/docs/schematypes.html#schematype-options) does not include the `expires` option for the `Date` type. It _can_ be found [here](https://mongoosejs.com/docs/api.html#schemadateoptions_SchemaDateOptions-expires) but it can be difficult to find.

### DIFF
--- a/docs/schematypes.md
+++ b/docs/schematypes.md
@@ -279,8 +279,9 @@ const schema2 = new Schema({
 
 <h5>Date</h5>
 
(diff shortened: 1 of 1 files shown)


Uzlopak (approved):
LGTM

Issue #11945, opened 2022-06-17
`Model.applyDefaults()` and `Model.castObject()`


ajwootto:
### Prerequisites

- [X] I have written a descriptive issue title

### Mongoose version

6.0.11

### Node.js version

16

### MongoDB version

4.6

### Operating system

macOS

### Issue

I have a use case where I want to be able to apply the Mongoose schema I have defined for a model to some plain data, without constructing the Mongoose document wrapper around it by doing `new Model`. The reason for this is to apply all the schema's defaults, datatype casting etc. without the CPU intensive task of constructing a full document with all the getters / setters and change tracking.

Is there a way to do this in Mongoose? Essentially to provide it some JSON data and get back a lean version of the document I would otherwise get by doing `new Model`?

Edit: I guess this question presumes that a "lean" query actually applies the Mongoose schema to the returned data, which I'm not actually sure is true. A better question then is basically "can I apply the schema defined by Mongoose to a piece of plain data without constructing a full document"?

vkarpov15:
I'm not sure I understand what you're trying to do. If you have some plain data, you already have a lean document - a lean document is a [POJO](https://masteringjs.io/tutorials/fundamentals/pojo).

I'm guessing what you're trying to do is apply defaults and type casting without hydrating a full document?

ajwootto:
Yeah exactly, I basically just want to apply the schema to my POJO and get back a POJO with all the defaults and type casting applied

vkarpov15:
This is a good idea. It would be nice to have helper functions on models that apply defaults to a POJO and apply type casting to a POJO. Good for both internal refactoring as well as providing a way to pick and choose which parts of Mongoose you want to use. Similar to `Model.validate()`.

PR #12057, opened 2022-07-06
feat(model): add `applyDefaults()` helper that allows applying defaults to document or POJO

vkarpov15:
Re: #11945

**Summary**

`Model.applyDefaults(doc)` applies the defaults specified in the model's schema to `doc`. `doc` can be either a document or a POJO/lean document.

Couple of use cases:

1. Create a document with `defaults: false`, because you want to apply defaults later after some other logic.
2. Making it easier to write plugins like mongoose-lean-defaults, etc.

**Examples**

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -187,6 +187,7 @@
       "error"
     ],
     "no-empty": "off",
-    "eol-last": "warn"
+    "eol-last": "warn",
+    "no-multiple-empty-lines": ["warn", { "max": 2 }]
   }
 }
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,5 +1,5 @@
 # These are supported funding model platforms
 
-github: vkarpov15
+github: mongoosejs
 open_collective: mongoose
 tidelift: npm/mongoose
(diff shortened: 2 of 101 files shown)


AbdelrahmanHafez (approved):
This is awesome, thanks!

Is there a way where we can use `applyDefaultToPOJO` inside `applyDefaults`? I'm concerned that on the long run we'll add some rules to `applyDefaults` that won't be applied to `applyDefaultsToPojo` and they'll go out of sync. This won't happen if `applyDefaults` relies on `applyDefaultsToPojo`

What do you think? @vkarpov15

vkarpov15:
@AbdelrahmanHafez I would love to make it so that we follow the same code path for both documents and POJOs. But unfortunately there's a lot of little edge cases. For example, for documents we need to set paths to `default` state, but not for POJOs. For POJOs, we need to recursively call `applyDefaultsToPOJOs()` for subdocs, but not for documents, because subdocuments call `applyDefaults()` recursively internally within the constructor. There probably is a way to unify the code paths without introducing any performance issues, but I'd like to get better benchmarking in place first.

PR #12062, opened 2022-07-07
fix(types): support findOneAndReplace with rawResult

lantw44:
**Summary**

Both findOneAndReplace and findOneAndUpdate support the rawResult
option, which changes the return type to include more information from
the MongoDB driver. Add an overload for findOneAndReplace to support
rawResult in a way similar to findOneAndUpdate.

**Examples**

Allow code like this to compile:
```typescript
const result = await models.OCPISession.findOneAndReplace(
  { uid },
  { uid, field: value },
  { session, rawResult: true, upsert: true, new: true },
).exec();
console.log(result.lastErrorObject?.upserted);
```


vkarpov15 (approved):
Thanks :+1: I'll double check whether our benchmarks show performance regressions or not in master, because this is branched off of a commit that had a performance issue.

PR #12067, opened 2022-07-07
added global `id` option to disable id on schemas

IslandRhythms:


AbdelrahmanHafez (changes requested):
This is great, thanks!

Can you add TS types+test to the PR as well?

Also, is there an option to set this on schema definition or only as a global option for now? It'd be nice if we can have an option to do that on the schema, but we can do it on a future PR.

vkarpov15 (changes requested):
Also add `id` to the TypeScript definitions for global Mongoose options here: https://github.com/Automattic/mongoose/blob/d742f4e71a026bc26acae0b4a797108af134f319/types/mongooseoptions.d.ts

AbdelrahmanHafez:
Thanks, I didn't know that.
@vkarpov15

AbdelrahmanHafez (approved):

vkarpov15 (approved):

PR #12071, opened 2022-07-08
Documentation-Generation remove `.prototype.` fix-up code

hasezoey:
**Summary**

This should no longer be required, because ".prototype." is now set in one place only so if a error occurs, it should be easy to find and fix

re https://github.com/Automattic/mongoose/pull/12024#discussion_r912474117


vkarpov15 (approved):
Thanks :+1:

PR #12072, opened 2022-07-08
Refactor all `@receiver` tags to proper jsdoc tags

hasezoey:
**Summary**

- refactor all `@receiver` tags into proper jsdoc tags where needed
- changes case `receiver` in `docs/source/api` to be a warning and no-op
- also change handling of case `property` to properly use types instead of trying to parse them from the string (now properly uses arrays and other symbols)

re https://github.com/Automattic/mongoose/pull/12024#pullrequestreview-1026843166


vkarpov15 (approved):
Thanks :+1:

PR #12081, opened 2022-07-08
added schema level lean option

IslandRhythms:


IslandRhythms:
It's complaining about one line in model.js that
- I did not touch
- is the comments for the `buildBultWriteOperations` where the line is the indented `@param`

AbdelrahmanHafez (changes requested):
Nice, thanks! 👍
Can you also add the option to the JSDoc and TS type definitions + TS tests?

Also, what do you think of calling it `defaultLean` instead of `lean`?

How would we go about overriding that for a specific query? I think we can `.setOptions({ lean: false });`, I also think it might be useful to have a `.notLean()` method that does the reverse of `.lean()`.

lib/query.js
```diff
@@ -1632,6 +1632,9 @@ Query.prototype.setOptions = function(options, overwrite) {
     this._mongooseOptions.defaults = options.defaults;
     // deleting options.defaults will cause 7287 to fail
   }
+  if (this.schema && 'lean' in this.schema.options) {
```
AbdelrahmanHafez, line 1635: Do we also need to check if the user is passing a lean value?
```suggestion
  if (options.lean == null && this.schema && 'lean' in this.schema.options) {
```

test/schema.test.js
```diff
@@ -2792,4 +2792,16 @@ describe('schema', function() {
...
+  it('allows a lean option on schemas so that all documents are lean when running a query (gh-10090)', async function() {
+    const testSchema = new mongoose.Schema({
+      name: String
+    }, { lean: true });
+    const Test = db.model('gh10090', testSchema);
+    await Test.create({
+      name: 'I am a lean doc, fast and small'
+    });
+    const entry = await Test.findOne();
+    assert.equal(entry instanceof mongoose.Document, false);
```
vkarpov15, line 2805: Add an extra check in this test that verifies you can still overwrite `lean()`.

So `const doc = await Test.findOne().lean(false); assert.ok(entry instanceof mongoose.Document)`

Also, no need to do  `assert.equal(entry instanceof mongoose.Document, false);`. Just `assert.ok(!entry instanceof mongoose.Document);`

PR #12082, opened 2022-07-08
feat(model): allow passing `timestamps` option to `Model.bulkSave(...)`

### LINKED ISSUE #12059: Feat: add `timestamps` option to `bulkSave`
### Prerequisites

### 🚀 Feature Proposal

Currently, we're able to set `timestamps` to `false` on bulkWrite for a specific operation, or for a `Document#save` by passing it as an option, and for `updateOne` as an option, but there is no way to write a `bulkSave` without updatings timestamps.

### Example

```js
const user = new User({ name: 'John Doe' });
await User.bulkSave([user], { timestamps: false });
```

AbdelrahmanHafez:
fixes #12059
I also refactored `bulkSave` to use async/await instead of the ugly promise chain, even though I wrote it, I find it difficult to read.
This comes with the downside of shorter stack traces, hopefully, this is not an issue.

@vkarpov15 
There is a test failing, `bulkSave` calls `bulkWrite` with `timestamps` false, so mongoose sends the writeOperations to MongoDB without `updatedAt` or `createdAt`, but the documents in memory still have `updatedAt` and `createdAt` that is inconsistent with the database.

This happens when we `document.schema.s.hooks.execPre('save',...)`. Do you know if there is a way to make pre-save hook not update timestamps? or pass an option to it? pre-post hooks seem like a blackbox to me and for years I couldn't figure out how they work.

### DIFF
--- a/lib/helpers/model/castBulkWrite.js
+++ b/lib/helpers/model/castBulkWrite.js
@@ -20,7 +20,7 @@ module.exports = function castBulkWrite(originalModel, op, options) {
       const model = decideModelByObject(originalModel, op['insertOne']['document']);
 
       const doc = new model(op['insertOne']['document']);
-      if (model.schema.options.timestamps) {
+      if (model.schema.options.timestamps && options.timestamps !== false) {
         doc.initializeTimestamps();
       }
       if (options.session != null) {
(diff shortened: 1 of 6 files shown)


vkarpov15:
@AbdelrahmanHafez I think the issue is this line: https://github.com/Automattic/mongoose/blob/c48461a744caac60bc832700368f8057a927cbc0/lib/helpers/timestamps/setupTimestamps.js#L42-L45 . `setupTimestamps()` registers a `pre('save')` hook that sets timestamps unless `this.$__.saveOptions.timestamps === false`.

I think that `bulkSave()` needs to set `doc.$__.saveOptions = options` on every doc being saved.

AbdelrahmanHafez:
Thanks, @vkarpov15, that fixes it. t👍

vkarpov15 (approved):
Nice, thanks :+1:

PR #12087, opened 2022-07-11
fix(types): add $let as a possible expression to $addFields

AbdelrahmanHafez:
Aggregation pipeline types currently don't accept the valid 
`$addFields: { myFieldName: { $let: { /* $let expression goes here */ } } }`

This PR fixes that, although I'm not sure if adding `Expression.Let` follows the desired structure for types.


vkarpov15 (approved):
Thanks :+1:

PR #12088, opened 2022-07-11
fix(types): fix $switch expression type

AbdelrahmanHafez:
[$switch](https://www.mongodb.com/docs/manual/reference/operator/aggregation/switch/#mongodb-expression-exp.-switch) statements accept `branches` instead of the current `$branches`, `case` instead of `$case`, and `default` instead of `$default`.


vkarpov15 (approved):
Thanks :+1:

PR #12090, opened 2022-07-11
feat(array): add property option "castNonArrays"

### LINKED ISSUE #10398: castNonArrays SchemaType option
bug

when `castNonArrays` is set to `false`, it still casts stuff

```ts
// NodeJS: 16.3.0
// MongoDB: 4.2-bionic (Docker)
import * as mongoose from "mongoose"; // mongoose@5.12.14

const schema1 = new mongoose.Schema({
  someArray: [{
    type: String,
    castNonArrays: false
  }]
});

const model1 = mongoose.model("model1", schema1);

(async () => {
  await mongoose.connect(`mongodb://localhost:44361/`, { useNewUrlParser: true, dbName: "verifyMASTER", useCreateIndex: true, useUnifiedTopology: true });

  const newdoc = await model1.create({ someArray: "hello" }); // no validation error

  const found = await model1.findById(newdoc);

  console.log(found);

  await mongoose.disconnect();
})();
```

yes typescript is used, but shouldnt matter for this example

Full Reproduction Repository: https://github.com/typegoose/typegoose-testing/tree/mongoosecastNonArrays

to fail with an cast error

nodejs: 16.4.0
mongoose: 5.12.14
mongodb(server): 4.0.25

---

my guess is that because the following line: 

https://github.com/Automattic/mongoose/blob/012808cd356ed542828bc8f39776b8dca496bf5d/lib/schema/array.js#L397

is `||` instead of `&&` (and no error)

hasezoey:
fixes #10398

**Summary**

Adds Property option `castNonArrays` to disable casting non-array values per-property instead of globally

**Examples**

```js
const Model = db.model('Test', new Schema({ x1: { castNonArrays: false, type: [String] } }));
const doc = new Model({ x1: "some non-array value" });
await doc.validate(); // Errors with "CastError"
```

PS: this is my first actual code contribution, please check if i changes everything necessary

Note on performance:
This may affect performance by doing additional checks, but it should not be too much

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -138,7 +138,7 @@ SchemaArray.schemaName = 'Array';
  *
  * - `castNonArrays`: `true` by default. If `false`, Mongoose will throw a CastError when a value isn't an array. If `true`, Mongoose will wrap the provided value in an array before casting.
  *
- * @static options
+ * @static
  * @api public
  */
 
@@ -161,8 +161,8 @@ SchemaArray.defaultOptions = {};
  *     const User = mongoose.model('User', new Schema({ test: Array }));
  *     new User({ }).validateSync().errors.test.message; // Path `test` is required.
  *
- * @param {String} option - The option you'd like to set the value for
- * @param {*} value - value for option
+ * @param {String} option The option you'd like to set the value for
+ * @param {Any} value value for option
  * @return {undefined}
  * @function set
  * @api public
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -393,7 +393,7 @@ SchemaArray.prototype.cast = function(value, doc, init, prev, options) {
     return value;
   }
 
-  if (init || SchemaArray.options.castNonArrays) {
+  if ((init || SchemaArray.options.castNonArrays) && (typeof this.options.castNonArrays === 'boolean' ? this.options.castNonArrays : true)) {
     // gh-2442: if we're loading this from the db and its not an array, mark
     // the whole array as modified.
     if (!!doc && !!init) {
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -393,7 +393,8 @@ SchemaArray.prototype.cast = function(value, doc, init, prev, options) {
     return value;
   }
 
-  if ((init || SchemaArray.options.castNonArrays) && (typeof this.options.castNonArrays === 'boolean' ? this.options.castNonArrays : true)) {
+  const castNonArraysOption = this.options.castNonArrays != null ? this.options.castNonArrays : SchemaArray.options.castNonArrays;
+  if (init || castNonArraysOption) {
     // gh-2442: if we're loading this from the db and its not an array, mark
     // the whole array as modified.
     if (!!doc && !!init) {
--- a/types/schemaoptions.d.ts
+++ b/types/schemaoptions.d.ts
@@ -204,7 +204,7 @@ declare module 'mongoose' {
     query?: Record<any, <T extends QueryWithHelpers<unknown, DocType>>(this: T, ...args: any) => T> | QueryHelpers,
 
     /**
-     * Set wheter to cast non-array values to arrays
+     * Set whether to cast non-array values to arrays.
      * @default true
      */
     castNonArrays?: boolean;
--- a/types/schemaoptions.d.ts
+++ b/types/schemaoptions.d.ts
@@ -202,5 +202,11 @@ declare module 'mongoose' {
      * Query helper functions
      */
     query?: Record<any, <T extends QueryWithHelpers<unknown, DocType>>(this: T, ...args: any) => T> | QueryHelpers,
+
+    /**
+     * Set wheter to cast non-array values to arrays
+     * @default true
+     */
+    castNonArrays?: boolean;
   }
 }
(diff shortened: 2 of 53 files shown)


types/schemaoptions.d.ts
```diff
@@ -202,5 +202,11 @@ declare module 'mongoose' {
      * Query helper functions
      */
     query?: Record<any, <T extends QueryWithHelpers<unknown, DocType>>(this: T, ...args: any) => T> | QueryHelpers,
+
+    /**
+     * Set wheter to cast non-array values to arrays
```
AbdelrahmanHafez, line 207:
```suggestion
     * Set whether to cast non-array values to arrays.
```

hasezoey: thanks

AbdelrahmanHafez (approved):
One minor change.
Otherwise LGTM, thanks! 👍

lib/schema/array.js
```diff
@@ -393,7 +393,7 @@ SchemaArray.prototype.cast = function(value, doc, init, prev, options) {
     return value;
   }
 
-  if (init || SchemaArray.options.castNonArrays) {
+  if ((init || SchemaArray.options.castNonArrays) && (typeof this.options.castNonArrays === 'boolean' ? this.options.castNonArrays : true)) {
```
AbdelrahmanHafez, line 396: I think the current state wouldn't work if we have `SchemaArray.options.castNonArrays` set to false, but on a specific path we want to override it to `true`. The following should work.
```suggestion
  const castNonArraysOption = this.options.castNonArrays != null ? this.options.castNonArrays : SchemaArray.options.castNonArrays;
  if (init || castNonArraysOption) {
```

What do you think of writing a test verifying that we can set `SchemaArray.options.castNonArrays` to `false` and then override it on a specific path?

hasezoey: currently i have documented that it required both `SchemaArray.options.castNonArrays` set to `true` or `init` AND `castNonArrays` (Property option) unset or set to `true`

hasezoey: added a case in b4715b5cb75d3a110e8394b0835798d618327ae9

AbdelrahmanHafez: I'm not sure https://github.com/Automattic/mongoose/commit/b4715b5cb75d3a110e8394b0835798d618327ae9 implements what I'm trying to describe, that would look more like that:
```js
it('`castNonArrays` on specific paths takes precedence over global option', function() {
  // Arrange
  const m = new Mongoose();
  m.Schema.Types.Array.options.castNonArrays = false;

  const userSchema = new Schema({ friendsNames: { type: [String], castNonArrays: true } });
  const User = m.model('User', userSchema);

  // Act
  const user = new User({ friendsNames: 'Sam' });

  // Assert
  const err = user.validateSync();
  assert.ok(err == null);
});
```

hasezoey: > I'm not sure https://github.com/Automattic/mongoose/commit/b4715b5cb75d3a110e8394b0835798d618327ae9 implements what I'm trying to describe, that would look more like that:

thanks for clarifying, currently b4715b5cb75d3a110e8394b0835798d618327ae9 tests that `SchemaArray.options.castNonArrays` takes priority over the property option

AbdelrahmanHafez: I'm a bit lost. Can you please point me to the specific line/column where we set `Schema.Types.Array.options.castNonArrays` to false?
The case that I'd like to cover by a test-case is: globalCastNonArrays: false, propertyCastNonArrays: true.
Does this test-case cover that? 🤨

AbdelrahmanHafez: I can see that `beforeEach` and `afterEach` we're re-assigning the values, but I can't see where we _change_ them.

Instead of using beforeEach/afterEach and trying to keep the values unchanged, can we instead instantiate new mongoose instances like https://github.com/Automattic/mongoose/pull/12090#discussion_r917988447? That way the test is self-contained and more explicit.

What do you think?

hasezoey: yes this could be done, but before adding i would like to wait for someone like @vkarpov15 to say what direct this change should go:
- let the property option overwrite the global, or
- let the property option only apply when the global is also `true`

AbdelrahmanHafez: Going with "path-option overrides global option" is consistent with how we do things in mongoose.
Check the following for example: https://github.com/Automattic/mongoose/blob/db1ba6e2a55a8134b99fc9074597ba2d5219e8b9/lib/helpers/setDefaultsOnInsert.js#L19-L22 
and
https://github.com/Automattic/mongoose/blob/db1ba6e2a55a8134b99fc9074597ba2d5219e8b9/lib/query.js#L3387-L3394

If we go with "letting the property apply _only_ when the global is also true", why would a person specify it if does nothing?

What do you think? @hasezoey @vkarpov15

hasezoey: for now i have applied the suggested changes (and the new test), because before i was not sure how mongoose handles such cases, thanks for the examples of where such cases exist already

vkarpov15: @hasezoey thanks for making the suggested change. The modified behavior that says "use the global value if there's no value set on the current schematype instance" is more consistent with the rest of Mongoose.

vkarpov15 (approved):
LGTM. I'm merging this into 6.5 branch.

PR #12114, opened 2022-07-16
fix: mongoose schema default should not mutate original object

### LINKED ISSUE #12102: Mongoose schema default mutating original object nested properties provided to model constructor
### Prerequisites

### Mongoose version

6.4.3

### Node.js version

16.15.1

### MongoDB server version

4

### Description

Nested object are mutated in the original object that is provided as document for model constructor when in model schema defaults are declared and the original objects do not include values for this properties.

### Steps to Reproduce

```
const Model = mongoose.model("TestModel", {
    topLevelField1: {
        nestedField1: String,
        nestedField2: {type: String, default: 'defValue'}
    },
    topLevelField2: {type: String, default: 'defValue'}
 });

 const originalData = {
    topLevelField1: {
        nestedField1: 'provided'
    }
 }
 const model = new Model(originalData);
 console.log(originalData)
 ```

### Expected Behavior

Expected result (works in mongoose 5.7.0):
`originalData === { topLevelField1: { nestedField1: 'provided' } }`

Current result (mongoose 6.4.3): 
`originalData === { topLevelField1: { nestedField1: 'provided', nestedField2: 'defValue' } }`

Expect that original object nested fields are not mutated by mongoose schema defaults.

### PR DESCRIPTION
**Pull request by lpizzinidev**
Closes #12102

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1150,7 +1150,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
 
       if (utils.isNonBuiltinObject(valForKey) && pathtype === 'nested') {
         // Create a copy of the object to avoid editing the original
-        const pathKey = Object.assign({}, path[key]);
+        const pathKey = utils.clone(path[key]);
         $applyDefaultsToNested(pathKey, prefix + key, this);
         this.$set(prefix + key, pathKey, constructing, Object.assign({}, options, { _skipMarkModified: true }));
         continue;
--- a/lib/document.js
+++ b/lib/document.js
@@ -1150,7 +1150,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
 
       if (utils.isNonBuiltinObject(valForKey) && pathtype === 'nested') {
         // Create a copy of the object to avoid editing the original
-        const pathKey = { ...path[key] }
+        const pathKey = { ...path[key] };
         $applyDefaultsToNested(pathKey, prefix + key, this);
         this.$set(prefix + key, pathKey, constructing, Object.assign({}, options, { _skipMarkModified: true }));
         continue;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Uzlopak** [comment]
I dont like spread operator here. Use clone if you need to create an immutable operation.

**Review by Uzlopak — CHANGES_REQUESTED** [review]
Use our internal clone method.

**Comment by lpizzinidev** [comment]
@Uzlopak With `utils.clone()` it's failing the tests, but with `Object.assign(...)` is working fine. 
Is it necessary to use the utils function?

**Comment by Uzlopak** [comment]
If you claim that you dont want to mutate the schema then you have to deepclone. Or else it is still mutating something.

**Comment by lpizzinidev** [comment]
The schema should be mutated, the problem is that the current implementation is mutating the original object by adding the default nested properties.
This is why I should use the standard JS method in my opinion. 
I'm not trying to make a copy of the document in my code change, I'm trying to make a copy of the object so it doesn't get mutated by the `$applyDefaultsToNested` call.

File: lib/document.js [hunk 1]
```diff
@@ -1149,8 +1149,10 @@ Document.prototype.$set = function $set(path, val, type, options) {
       }
 
       if (utils.isNonBuiltinObject(valForKey) && pathtype === 'nested') {
-        $applyDefaultsToNested(path[key], prefix + key, this);
-        this.$set(prefix + key, path[key], constructing, Object.assign({}, options, { _skipMarkModified: true }));
+        // Create a copy of the object to avoid editing the original
+        const pathKey = utils.clone(path[key]);
```
vkarpov15, line 1153: I don't like this change, it seems unnecessarily heavy from a performance perspective. I think we'd just need to refactor and be smarter about setting defaults on the nested path without setting them on the input parameter. We'll take a look and see if we can come up with an alternative solution. I agree this is a bug, I just don't think the solution is ideal.

PR #12115, opened 2022-07-18
feat(document): add `$inc()` helper that increments numeric paths

### LINKED ISSUE #11915: can I do $inc while using .save?
### Prerequisites

### Mongoose version

5.13.6

### Node.js version

14.19.1

### MongoDB version

4.0.5

### Operating system

Linux

20.04

### Issue

I have a operation where an object is pushed into an array and a counter is increased.
```javascript
let foo = await Foo.findOne({ <some-filter> });
foo.bars.push({ bar: 2 });
foo.bar_count++;
await foo.save();
```
when I check the query being executed in debug mode, I see it is doing $push on array but $set on the count. In my use case I want to use $inc due to some concurrency issues. I searched and found `.$inc()` but that seems to be deprecated.
Is there any way by which I can tell `save` how to make the query? or maybe change the query which will be executed?

vkarpov15:
Fix #11915

**Summary**

See #11915. Right now, there's no way to use [`$inc`](https://www.mongodb.com/docs/manual/reference/operator/update/inc/) with `save()`. `doc.num += 2` becomes a `$set`.

With this change, you can explicitly increment a given path, and when you `save()`, Mongoose will send a `$inc`.

**Examples**

```javascript
const schema = new Schema({
  counter: Number
});
Test = db.model('Test', schema);

await Test.create({ counter: 0 });
const doc = await Test.findOne();
assert.strictEqual(doc.counter, 0);
const doc2 = await Test.findOne();

doc2.counter = 1;
await doc2.save();

doc.$inc('counter');
await doc.save();

const res = await Test.findById(doc);
assert.equal(res.counter, 2);
```

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -11572,7 +11572,7 @@ describe('document', function() {
         doc2.counter = 1;
         await doc2.save();
 
-        doc.$inc('counter');
+        doc.$inc('counter', 1);
         await doc.save();
 
         const res = await Test.findById(doc);
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -11551,6 +11551,202 @@ describe('document', function() {
     assert.ok(err);
     assert.ok(err.errors['testProp.testSubProp.nested.from']);
   });
+
+  describe('$inc (gh-11915)', function() {
+    describe('top-level path', function() {
+      let Test;
+
+      beforeEach(function() {
+        const schema = new Schema({
+          counter: Number
+        });
+        Test = db.model('Test', schema);
+      });
+
+      it('sends a $inc command for a given path', async function() {
+        await Test.create({ counter: 0 });
+        const doc = await Test.findOne();
+        assert.strictEqual(doc.counter, 0);
+        const doc2 = await Test.findOne();
+
+        doc2.counter = 1;
+        await doc2.save();
+
+        doc.$inc('counter');
+        await doc.save();
+
+        const res = await Test.findById(doc);
+        assert.equal(res.counter, 2);
+      });
+
+      it('works as a $set if the document is new', async function() {
+        const doc = new Test({ counter: 0 });
+        doc.$inc('counter', 2);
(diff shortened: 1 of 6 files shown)


test/document.test.js
```diff
@@ -11551,6 +11551,202 @@ describe('document', function() {
...
+      it('sends a $inc command for a given path', async function() {
+        await Test.create({ counter: 0 });
+        const doc = await Test.findOne();
+        assert.strictEqual(doc.counter, 0);
+        const doc2 = await Test.findOne();
+
+        doc2.counter = 1;
+        await doc2.save();
+
+        doc.$inc('counter');
```
AbdelrahmanHafez, line 11575: Would be nice to have the test explicitly increment specific value, to have some idea how it works when looking at the tests as an example:
```suggestion
        doc.$inc('counter', 1);
```

AbdelrahmanHafez (approved):
One minor nitpick, otherwise LGTM. 👍

Kind of surprised by the number of scenarios we have to take into account for this to work properly.

AbdelrahmanHafez:
On a second thought, what do you think about the following syntax? How easy would it be to support it?
```js
document.counter.$inc(1);
document.friends[3].age.$inc(1);
```

vkarpov15:
@AbdelrahmanHafez `document.counter.$inc(1);` is unfortunately not viable. [You can't define properties on primitives, it throws an error in strict mode](https://masteringjs.io/tutorials/fundamentals/strict#no-setting-properties-on-primitives).

vkarpov15:
Re: number of cases, it is mostly just because I'm trying to be more defensive about shipping new features. In the past, we've often had edge cases in new features like "oh this doesn't work with subdocuments". So this time I'm trying to cover every potential bug I can think of with tests.

Uzlopak (changes requested):
LGTM

Uzlopak (approved):
LGTM

PR #12121, opened 2022-07-19
feat(model): add `hydrate` option to `Model.watch()` to automatically hydrate `fullDocument`

### LINKED ISSUE #11936: Change streams: return mongoose Document if `fullDocument` is present in the change document
### Prerequisites

### 🚀 Feature Proposal

MongoDB change events may contain the whole document that was modified under the key `fullDocument`. Would it make sense to have mongoose automatically wrap this in an appropriate mongoose Document class?

### Motivation

Given that I already use mongoose Documents most of the time, it might make sense to have them here. This would enable me to use instance methods, getters, setters, virtual fields, etc.

### Example

_No response_

vkarpov15:
Fix #11936

**Summary**

Automatically hydrate `fullDocument` if `hydrate: true` is set, but only if you call `Model.watch()`. `Connection.prototype.watch()` doesn't support `hydrate: true`.

Helpful because 1) it might be hard to figure out what function you need to call to init a Mongoose document without setting anything modified or firing setters, 2) this makes it easier for us to add a global `hydrateFullDocument` option so we can do this on all `Model.watch()` calls by default.

**Examples**

### DIFF
--- a/lib/cursor/ChangeStream.js
+++ b/lib/cursor/ChangeStream.js
@@ -20,6 +20,13 @@ class ChangeStream extends EventEmitter {
     this.pipeline = pipeline;
     this.options = options;
 
+    if (options && options.hydrate && !options.model) {
+      throw new Error(
+        'Cannot create change stream with `hydrate: true` ' +
+        'unless calling `Model.watch()`'
+      );
+    }
+
     // This wrapper is necessary because of buffering.
     changeStreamThunk((err, driverChangeStream) => {
       if (err != null) {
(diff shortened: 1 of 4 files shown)


lib/cursor/ChangeStream.js
```diff
@@ -69,6 +81,32 @@ class ChangeStream extends EventEmitter {
...
+          if (data.fullDocument != null) {
+            data.fullDocument = this.options.model.hydrate(data.fullDocument);
+          }
+          return originalCb(null, data);
+        };
+      }
+
+      let maybePromise = this.driverChangeStream.next(cb);
+      if (maybePromise && typeof maybePromise.then === 'function') {
+        maybePromise = maybePromise.then(data => {
```
AbdelrahmanHafez, line 100: Smart usage of promises right there.

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

Issue #12070, opened 2022-07-08
When the `_id` field is specified in the schema definition, the type of `_id` field cannot be correctly inferred in the HydratedDocument, perhaps due to incorrect implemention of `Require_id`


Starrah:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.4.3

### Node.js version

16.15.1

### MongoDB server version

This issue have no relevance to MongoDB server.

### Description

When the `_id` field is specified in the schema definition, eg, `new Schema({_id: String})`, the type of `_id` field cannot be correctly inferred in the `HydratedDocument` (the type returned by `new Model` or `await Model.findOne`). 

Please see the "Steps to Reproduce" section for reproducing code. 

I think that the bug is perhaps due to incorrect implemention of `Require_id`: https://github.com/Automattic/mongoose/blob/3445d240886e508962d7dd6a8481715a3cb373da/types/index.d.ts#L111-L115. 

Sorry that I'm not excellent in Typescript, but what does `U extends any` here mean?
As far as I know, everything extends any? so the contiditional clause (T & { _id: Types.ObjectId }) is always used, and the clause T & Required<{ _id: U }> is never used.

Thus, `_id` will be always infered with type ObjectId, even if I have defined another type such as String in the Schema definition.

The issue have been previously discussed at https://github.com/Automattic/mongoose/pull/11563/commits/e0d299ff545caed9dc194c3c5ce5bfdfd1fa1b4a#r915551078 with @mohammad0-0ahmad @Uzlopak, please see the link for details.

### Steps to Reproduce

The following test case demonstrates the bug:
```ts
// index.test-d.ts
import {model, Schema} from "mongoose";
import {expectType} from "tsd";

const schema_with_string_id = new Schema({_id: String, nickname: String})
const theModel = model('test', schema_with_string_id)
const obj = new theModel()

expectType<string>(obj._id)
```
when testing it with `tsd`, gives the following error: 
```
 index.test-d.ts:8:0
  ✖  8:0  Parameter type string is declared too wide for argument type string & ObjectId.
```

vkarpov15:
Thanks for the concise bug report, fix will be in 6.4.5 next week :+1:

Issue #12069, opened 2022-07-07
TypeScript: timestamps not inferred from schema


janickvwcotiss:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.4.0

### Node.js version

18.2.0

### MongoDB server version

5.0.9

### Description

Hi there, my issue is to do with typescript.

The issue I am having is that when creating a document schema with timestamps enabled the inferred document type doesn't have the corresponding `createdAt` and `updatedAt` fields present.

This means that when querying documents through the schema typescript will throw the property doesn't exist on type issue for the `createdAt` and `updatedAt` fields.

Hope this is pretty simple to understand, I've done my best to proved a good code example to best explain the issue I am having.

I realised mongoose has just recently migrated to typescript and isn't fully feature complete yet so appreciate any help.

P.s. I realise I haven't manually typed the interface for the schema e.g. `new Schema<IDocument>` and `export const Document = model<IDocument>('Document', DocumentSchema);` this is because we are migrating our whole code base to typescript and would ideally like to save some time by using `InferSchemaType`. But obviously open to any ideas even if it'll end up requiring typing the interfaces manually.

Thanks.

### Steps to Reproduce

```typescript
import { Schema, model, Model, InferSchemaType } from 'mongoose';

const DocumentSchema = new Schema(
  {
    url: { type: String, required: true, unique: true },
    fileSize: { type: Number, required: true },
    mimeType: { type: String, required: true },
    fileName: { type: String, required: true },
  },
  {
    timestamps: true,
  }
);

// Document comes out as
// const Document: Model<{
//   url: string;
//   fileSize: number;
//   mimeType: string;
//   fileName: string;
// }, {}, {}, {}, any>
export const Document: Model<InferSchemaType<typeof DocumentSchema>> = model(
  'Document',
  DocumentSchema
);

async function findDocument(id: string) {
  // Document without timestamp in query
  const doc1 = await Document.findById(id);
  if (doc1) {
    console.log(doc1.createdAt); // Error: "Property 'createdAt' does not exist on type 'Document...'"
    console.log(doc1.updatedAt);// Error: "Property 'updatedAt' does not exist on type 'Document...'"
  }

  // With timestamp in query
  const doc2 = await Document.findById(id, undefined, { timestamps: true });
  if (doc2) {
[...]

Issue #11967, opened 2022-06-21
[help] InferSchemaType, accessing _id field


philiparvidsson:
### Prerequisites

- [X] I have written a descriptive issue title

### Mongoose version

6.4.0

### Node.js version

18.3

### MongoDB version

Unsure

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

arch-5.18.5 something

### Issue

Using `InferSchemaType` to create the schema interface type, how do I access the `_id` field? If I put it in the schema definition, I have to also set it explicitly when inserting new documents. If I don't put it there, `InferSchemaType` does not include it in the interface.

What do I do? I'd like to know the same for `.save()` and so on. Am I supposed to do a type union?

Code:

```javascript
import { Document, InferSchemaType, Schema } from "mongoose"

const mySchema = new Schema({
  foo: { type: String, required: true }
})

type MySchemaType_Problematic = InferSchemaType<typeof mySchema>
// MySchemaType_Problematic is missing the _id field along with the save() method etc.

// Workaround:
type MySchemaType_Workaround = InferSchemaType<typeof mySchema> & Document
// MySchemaType_Workaround can now be used properly, with _id and save() etc.
```

vkarpov15:
@philiparvidsson can you please provide some code samples? That would be helpful for me to understand what you're trying to do.

philiparvidsson:
I added some example code, providing a brief example of the issue as well as the workaround I'm using.

mohammad0-0ahmad:
@philiparvidsson  InferSchemaType meant to be a helper to infer schema defenition type.
So It will not return a document function but it will obtain paths type.
As I descriped _id property it is not added yet because there it a default path in the schema.
Do you need to use these types within mongoose code or you obtain the type of schema for another usecase?

Can you please provide more details to be able to help maybe?

philiparvidsson:
In this particular case, I'm passing a MySchemaType_Problematic to a factory function. If the factory cannot generate the needed type, it needs to log that it did not understand the data it received. So I tried to do: log.error(`Factory was unable to create instance for {mySchemaType._id}`)

But it doesn't work since the type doesn't have an id. I guess I could cast it to any because I know for a fact that _id is there, but that defeats the purpose of having any type on it to begin with.

philiparvidsson:
[...]

Issue #11910, opened 2022-06-07
LeanDocument, the third...


Uzlopak:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

@mohammad0-0ahmad 
@taxilian 
@vkarpov15 
@AbdelrahmanHafez 

Tbh, I really dislike the current state of LeanDocument typings. In my opinion, the automatic infer schema PR was the right starting point. 

The shape of the Lean and Hydrated Documents is defined by the Schema. Also the options of a Schema change the shape of the LeanDocument. You store a __v field in the document if you use versioning. And if you use a different fieldname for versioning than this effects also the actual shape of the Document. So omitting fields like this is imho just plain wrong:

https://github.com/Automattic/mongoose/blob/c714d52236fe8ea8531c2ca4b2f10ae11f625080/types/index.d.ts#L552

I think we have to write first all the unit tests for the typings and agree what we actually expect. I think this is something I would start as next.

vkarpov15:
LeanDocument is gone as of 7.x

PR #12122, opened 2022-07-19
[⏳] Enhanced auto typed schema.

mohammad0-0ahmad:
Enhance the auto typed schema by covering the effects related to the schema options value.
Issues to be covered or enhance the already solved ones:
- [x] https://github.com/Automattic/mongoose/issues/12069
- [x] Coverage for ```__v``` path
- [x] Coverage for ```id``` virtual path
- [x] https://github.com/Automattic/mongoose/issues/11967
- [ ] Optimize ts-benchmark result.
- [ ] https://github.com/Automattic/mongoose/issues/12070
- [ ] https://github.com/Automattic/mongoose/issues/11910
- [ ] https://github.com/Automattic/mongoose/issues/11787
- [ ] https://github.com/Automattic/mongoose/issues/12126

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -705,5 +705,5 @@ function gh12122() {
   expectType<DefaultSchemaOptions>({} as ObtainSchemaGeneric<typeof Test2, 'TSchemaOptions'>);
 
   const Test3 = new Schema({ test: String }, { timestamps: true });
-  expectType<{ test?: string; createdAt: Date; updatedAt: Date; }>({} as InferSchemaType<typeof Test3>);
+  expectType<{ __v: string; test?: string; createdAt: Date; updatedAt: Date; }>({} as InferSchemaType<typeof Test3>);
 }
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -12,7 +12,8 @@ import {
   SchemaOptions,
   ObtainSchemaGeneric,
   DefaultSchemaOptions,
-  FlatRecord
+  FlatRecord,
+  ApplySchemaOptions
 } from 'mongoose';
 import { expectType, expectError, expectAssignable } from 'tsd';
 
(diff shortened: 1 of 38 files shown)


test/types/schema.test.ts
```diff
@@ -599,20 +600,6 @@ function gh11997() {
   userSchema.index({ name: 1 }, { weights: { name: 1 } });
 }
 
-function gh12003() {
```
mohammad0-0ahmad, line 602: This test is incorrect, Auto typed schema will not work if typeKey has been changed in this way, because it will be "type" and will not reflect the correct type of the actual value.
@Uzlopak  @vkarpov15  do you have any thoughts?

mohammad0-0ahmad: I don't think we can support this way to supply schema options with auto typed schema.
It will work just if the end user use "const" assertion and not use predefined type, then the type will reflect the actual value.

hasezoey:
@mohammad0-0ahmad is this PR still planned on being updated? because it looks rather stale to me - if this should not get a update within ~30 days, i would close it as "stale"

mohammad0-0ahmad:
> @mohammad0-0ahmad is this PR still planned on being updated? because it looks rather stale to me - if this should not get a update within ~30 days, i would close it as "stale"

Hello @hasezoey !

Sorry for being inactive for a while.
I don't think as well that I will have time to continue with this one during the next month.
You can close it right now and when I have time to continue with it, then we can reopen it.
@vkarpov15 @Uzlopak might can check this PR and maybe complete it if they have time for that and  if it something  good to have soon.

Best regards everyone :)

vkarpov15:
We're going to close this PR for now. We implemented part of this PR in #12731, but ran into some issues like #12871. Given those issues, we want to sort out getting this working for just `timestamps` option before expanding to other similar tasks. We need to be very cautious with big TypeScript changes, because there's always some obscure issue that pops up.

francescov1:
@vkarpov15 Was there any progress made on https://github.com/Automattic/mongoose/issues/11787? Or any relevant findings? I tried tackling this about 6 months ago but didn't have any luck, would like to give it another shot

PR #12125, opened 2022-07-19
fix(types): map correct generics from model to schema

emiljanitzek:
**Summary**

It is not possible to define a model with both generic `TInstanceMethods` and `TQueryHelpers`. There is a mismatch between generics for the `model` and the the `Schema`.

**Examples**

See type test for example

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -7,7 +7,9 @@ import {
   model,
   Types,
   UpdateQuery,
-  CallbackError
+  CallbackError,
+  HydratedDocument,
+  Query
 } from 'mongoose';
 import { expectAssignable, expectError, expectType } from 'tsd';
 import { AutoTypedSchemaType, autoTypedSchema } from './schema.test';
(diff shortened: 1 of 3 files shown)


test/types/models.test.ts
```diff
@@ -329,6 +331,37 @@ function gh11911() {
...
+  interface UserQueryHelpers {
+    byName(this: UserModelQuery, name: string): this
+  }
+  interface User {
+    name: string;
+  }
+  interface UserInstanceMethods {
+    findByName(name: string): Promise<HydratedDocument<User>>;
+  }
+  type UserModel = Model<User, UserQueryHelpers> & UserInstanceMethods;
```
vkarpov15, line 345: Seems a little odd that `UserInstanceMethods` is part of the `UserModel` type. You shouldn't be able to do `UserModel.findByName()`...

vkarpov15 (commented):
In general LGTM, just had a question about one line in the tests that looks potentially wrong.

test/types/models.test.ts
emiljanitzek: You are correct, I think the confusion came from `TInstanceMethod` and `TStaticMethods` on Schema and `TMethodsAndOverrides` on Model which is the same as `TInstanceMethod`. I've renamed the interfaces in the test and added statics too.

vkarpov15 (approved):
LGTM. I'm going to get rid of the `U` constraint, because that seems to ding performance, but otherwise looks good.

Issue #11916, opened 2022-06-09
I have a problem when I switch database before creating timeseries.


satasuk03:
I have a problem when I switch database before creating timeseries.  
it will creates a collection instead o f timeseries.  

Here's the code

```typescript
// db.ts
export const useDB = (guidId: string) =>
  mongoose.connection.useDb(guidId, {
    // ensures connections to the same databases are cached
    useCache: true,
    // remove event listeners from the main connection
    noListener: true,
  });

// =====================================

// schema file
import { useDB } from 'db';
import mongoose, { Schema } from 'mongoose';

export interface ActiveUserDoc {
  timestamp: Date;
  messages: Number;
  metadata: { discordId: string };
}

const ActiveUserSchema = new Schema(
  {
    timestamp: Date,
    messages: Number,
    metadata: {
      discordId: String,
    },
  },
  {
    timeseries: {
      timeField: 'timestamp',
      metaField: 'metadata',
      granularity: 'hours',
    },
    expireAfterSeconds: 60 * 24 * 7, // 7 days
  },
);

export const getModel = (guildId: string) =>
  useDB(guildId).model<ActiveUserDoc, mongoose.Model<ActiveUserDoc>>(
    'ActiveUser',
    ActiveUserSchema,
    'active_user',
  );

```

<img width="445" alt="image" src="https://user-images.githubusercontent.com/96050763/172794966-5417c08b-c163-4f2a-a70f-eec8412907c9.png">

_Originally posted by @satasuk03 in https://github.com/Automattic/mongoose/issues/10611#issuecomment-1150794186_

satasuk03:
This is the way I temporary fix the problem

```typescript
import { useDB } from 'db';
import mongoose, { Schema } from 'mongoose';

export interface ActiveUserDoc {
  timestamp: Date;
  messages: Number;
  metadata: { discordId: string };
}

const ActiveUserSchema = new Schema(
  {
    timestamp: Date,
    messages: Number,
    metadata: {
      discordId: String,
    },
  },
  {
    autoCreate: false,
    autoIndex: false,
  },
);

export const getModel = (guildId: string) => {
  const conn = useDB(guildId);
  const model = conn.model<ActiveUserDoc, mongoose.Model<ActiveUserDoc>>(
    'ActiveUser',
    ActiveUserSchema,
    'active_user',
  );
  model.createCollection({
    timeseries: {
      timeField: 'timestamp',
      metaField: 'metadata',
      granularity: 'hours',
[...]

PR #12130, opened 2022-07-20
fix(connection): re-run `Model.init()` if re-connecting after explicitly closing a connection

### LINKED ISSUE #12047: Mongoose does not automatically recreate index after reconnecting to different instance (calling .connection.close() then .connect())
### Prerequisites

### Mongoose version

6.1.4

### Node.js version

16.14.2

### MongoDB server version

5.3

### Description

I have a test suite for my personal project where I use mongoose to connect to a new instance of MongoDB for each test case. The actual setup is complex, but following is the idea:

```ts
describe('suite', () => {
     beforeEach(() => {
        const uri = getNewMongoUri();
        mongoose.connect(uri);
     });

     afterEach(() => {
        mongoose.connection.close();
     });


     it('case1'); //...
     it('case2'); //...
});
```

For one of the suites, I need to have text indexes built. However it seems like mongoose only automatically builds the text index when `.connect` is first called. It doesn't rebuild indexes for subsequent calls even if I am trying to connect to a completely different server.

I suggest adding this caveat to `autoIndex` option documentation. Or make index automatically rebuild for each `.connect` call prehaps?

**_There is a solution to this, which is to simply call `ensureIndexes` manually. But this behaviour just seemed weird and thought it could be a bug._**

### Steps to Reproduce

You just have to build a test suite or a script that does something similar to this (and make sure you use the index in `case1` and `case2`:

```ts
describe('suite', () => {
     beforeEach(() => {
        const uri = getNewMongoUri();
        mongoose.connect(uri);
     });

     afterEach(() => {
        mongoose.connection.close();
     });


     it('case1'); //...
     it('case2'); //...
});
```

I used `mongodb-memory-server`, but also reproduceable using multiple installs directly on the OS.

### Expected Behavior

Mongoose should automatically rebuild indexes after each `.connect` call if using different servers.

vkarpov15:
Fix #12047

**Summary**

The [`Model.init()` function](https://mongoosejs.com/docs/api/model.html#model_Model-init) is responsible for:

1. Calling `createCollection()` to ensure the Model's underlying collection exists, unless `autoCreate = false`
2. Calling `createIndexes()` to ensure the indexes defined in the Model's schema exist, unless `autoIndex = false`

Mongoose calls `Model.init()` when you create a model using `mongoose.model()` or `Collection.prototype.model()`. The `createCollection()` and `createIndexes()` calls are queued up until Mongoose connects to MongoDB.

The issue is that, if you close a connection and then reconnect to a different database or cluster, Mongoose won't re-run `init()`. Because 1) Mongoose only runs `init()` when you call `model()`, and 2) Mongoose stores a `$init` property on a model that contains a promise wrapping the `init()` call [here](https://github.com/Automattic/mongoose/blob/422f9da02d2e8c0c227887265bce25cecaecf403/lib/model.js#L1315-L1327). If `$init` is set, then calling `init()` just returns `$init`. This is to avoid duplicate `init()` calls.

This fix deletes `$init` when the user calls `Connection.prototype.close()`, and ensures that Mongoose calls `init()` when the user opens a connection using `openUri()` or `setClient()`. So `close()` clears `$init` caches, and re-opening the connection means Mongoose re-runs `init()`.

Related issue: #11916

**Examples**

See tests

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -550,8 +550,8 @@ Connection.prototype.dropDatabase = _wrapConnHelper(function dropDatabase(cb) {
   // init-ed. It is sufficiently common to call `dropDatabase()` after
   // `mongoose.connect()` but before creating models that we want to
   // support this. See gh-6796
-  for (const name of Object.keys(this.models)) {
-    delete this.models[name].$init;
+  for (const model of Object.values(this.models)) {
+    delete model.$init;
   }
   this.db.dropDatabase(cb);
 });
@@ -838,6 +838,11 @@ Connection.prototype.openUri = function(uri, options, callback) {
     );
   }
 
+  for (const model of Object.values(this.models)) {
+    // Errors handled internally, so safe to ignore error
+    model.init(function $modelInitNoop() {});
+  }
+
   return this.$initialConnection;
 };
 
@@ -949,6 +954,13 @@ Connection.prototype.close = function(force, callback) {
     this.$wasForceClosed = !!force.force;
   } else {
     this.$wasForceClosed = !!force;
+  }
+
+  for (const model of Object.values(this.models)) {
+    // If manually disconnecting, make sure to clear each model's `$init`
+    // promise, so Mongoose knows to re-run `init()` in case the
+    // connection is re-opened. See gh-12047.
+    delete model.$init;
   }
 
   return promiseOrCallback(callback, cb => {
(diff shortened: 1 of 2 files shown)


IslandRhythms (approved):

AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #12139, opened 2022-07-22
fix(types): better typescript support for schema plugins

emiljanitzek:
**Summary**

Found more problems when working with typescript and plugins. Since the Schema type sets default values (empty object) for `TInstanceMethods`, `TQueryHelpers` and `TStaticMethods` we need to be able to pass along these types when trying to create query helpers/virtuals/methods/statics.

This PR also contains the fix for Schema and model mismatch. Will rebase this if the other PR gets accepted.

**Examples**

Added `plugin.tests.ts` for examples

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -157,9 +157,20 @@ declare module 'mongoose' {
 
   type QueryResultType<T> = T extends Query<infer ResultType, any> ? ResultType : never;
 
-  type PluginFunction<DocType> = (schema: Schema<DocType>, opts?: any) => void;
+  type PluginFunction<
+    DocType,
+    M = Model<DocType, any, any, any>,
+    TInstanceMethods = {},
+    TQueryHelpers = {},
+    TVirtuals = {},
+    TStaticMethods = {}> = (schema: Schema<DocType, M, TInstanceMethods, TQueryHelpers, TVirtuals, TStaticMethods>, opts?: any) => void;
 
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -291,15 +296,15 @@ declare module 'mongoose' {
...
+    static(obj: { [F in keyof TStaticMethods]: TStaticMethods[F] } & { [name: string]: (this: M, ...args: any[]) => any }): this;
 
     /** Object of currently defined statics on this schema. */
-    statics: { [name: string]: (this: M, ...args: any[]) => any };
+    statics: { [F in keyof TStaticMethods]: TStaticMethods[F] } & { [name: string]: (this: M, ...args: any[]) => any };
 
     /** Creates a virtual type with the given name. */
     virtual<T = HydratedDocument<DocType, TInstanceMethods>>(
-      name: string,
+      name: keyof TVirtuals,
```
vkarpov15, line 307: Same with above. This change is a bit risky.

vkarpov15 (changes requested):

emiljanitzek:
Thanks for the feedback @vkarpov15. I've updated the types and added test for this as well.

vkarpov15 (approved):

vkarpov15:
Looks like there were some merge conflict issues, because tests are failing now. Can you please patch that up @emiljanitzek ?

I took a look and it looks like this change doesn't cause any significant performance degradation.

emiljanitzek:
@vkarpov15 rebased and fixed the conflicts.

PR #12140, opened 2022-07-22
Touch-up JSDOC of some more files

hasezoey:
**Summary**

This PR does:
- touch-up some more file for JSDOC, for:
  - `lib/aggregate.js`
  - `lib/cast.js`
  - `lib/collection.js`
  - `lib/connection.js`
  - `lib/document.js`
- replace all old (broken) links `#\w+_\w+\.` to be `#\w+_\w+-` (from `#model_Model.find` to `#model_Model-find`)

PR #12141, opened 2022-07-22
fix(schematype): actually always return "this" where specified

hasezoey:
**Summary**

while going through the code for #12140, i noticed that 2 functions were saying they always return `this`, which was not actually true for 1 case (in each function), this PR fixes that


vkarpov15 (approved):
Good catch, thanks!

Issue #12111, opened 2022-07-15
[Docs] Populate's behavior is not greatly documented


hasezoey:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Currently [`populate`](https://mongoosejs.com/docs/api/document.html#document_Document-populate)'s behavior is not greatly documented for cases in which it fails, the documentation currently just says `not populated`, which could mean many things, so this issue proposes that the documentation would need to be updated.

Current docs:
https://github.com/Automattic/mongoose/blob/115f9220d6fb9da89e21a2a3f5895b6b31129afe/lib/document.js#L4181-L4200

Things that should be changed / documented:
- change `not populated` to actually say what type is expected in case it fails (like is it `null | undefined`~~, or the raw id~~)
- what happens for paths that dont actually exist (does populate throw or try to do it anyway or just ignore it)
- document what the actual return type of `.populate` call is (currently it is just `Promise`)
- document that `Document.prototype.populate` is (at least a bit) different from `Query.prototype.populate`

PS: some of those *may* be fixed by just using the `@see` tags from the jsdoc in the documentation

PR #12144, opened 2022-07-23
Add "@see" handling and output to documentation

hasezoey:
**Summary**

This PR adds JSDOC `@see` handling (in mongoose specific way) and output to the documentation

the following matches apply for the custom `@see` handling:
```txt
# The following is "raw tag" to (->) "text output as listed in the documentation"
"@see External Links http://someurl.com/" -> "External Links"
"@see External https://someurl.com/" -> "External"
"@see Id href #some_Some-method" -> "Id href"
"@see Local Absolute /docs/somewhere" -> "Local Absolute"
"@see No Href" -> "No Href"
"@see https://someurl.com" -> "" (uses fallback "No text")
"@see Some#Method #something" -> "Some#Method"
```

This PR also does some things i noticed while going through:
- add `:` to `Parameters` list to be consistent with other `h5`
- add proper link to `Schema.prototype.plugin`

As a side-effect, this also addresses some things mentioned in #12111

### DIFF
--- a/docs/api_split.pug
+++ b/docs/api_split.pug
@@ -75,5 +75,10 @@ block content
         h5 Inherits:
         ul
           li <span class="method-type">&laquo;#{prop.inherits}&raquo;</span>
+      if prop.see != null 
+        h5 See:
+        ul
+          each see in prop.see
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #12145, opened 2022-07-23
docs(api_split.pug): add "code" to parameter name

hasezoey:
**Summary**

This changes Documentation Parameter names from "someOption" to `someOption`, which in my opinion is better to read

before:
before img

after:
after img

<details>
<summary>before/after with darkreader</summary>
before:

before img

after:
after img

</details>

### DIFF
--- a/docs/api_split.pug
+++ b/docs/api_split.pug
@@ -57,11 +57,11 @@ block content
             - if (param.nested)
               ul(style="margin-top: 0.5em")
                 li
-                  | #{param.name}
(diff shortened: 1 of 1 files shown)


hasezoey:
By the way, should the headers maybe also be wrapped in code-blocks too?

vkarpov15 (approved):
You're right, much easier to read. Thanks!

vkarpov15:
@hasezoey headers would be a good idea too. As long as that doesn't screw up header ids.

PR #12147, opened 2022-07-24
ts-benchmark

mohammad0-0ahmad:
Use the latest release of ts-benchmark.
Reasons:
- Fix some issues related to GitHub integration.
- Add ability to benchmark the target branch of the pull request. "Helpful if target branch is not the master branch is our case."

Screenshots:
This show how the result will be look like when creating a PR against 6.5 branch for instance instead of master.

Capture


vkarpov15 (approved):
Thanks :+1:

PR #12151, opened 2022-07-26
fix(timestamps): set timestamps on child schema when child schema has timestamps: true but parent schema does not

### LINKED ISSUE #12119: Different timestamp behaviour for different operators on nested subdocuments
### Prerequisites

### Mongoose version

6.3.3

### Node.js version

12

### MongoDB server version

4.5

### Description

When updating a nested document array, the timestamps of the nested documents are only updated in certain cases.

When using `findOneAndUpdate` and updating the nested document via `$set` operator, the timestamps are not updated in certain cases even if the `timestamps` option is enabled on the child schema.

Automatic update of the timestamps on the deepest schema only works if either all nodes in-between have option `timestamp: true` OR when using the `$push` operator.

This is due to function `applyTimestampsToSingleNested` (and respectively `applyTimestampsToDocumentArray`) exiting too early when the timestamp option on the current node is not set, ignoring all child nodes from there.

```

  // `timestamps` option set to true on deepest sub document
  const ConditionSchema = new mongoose.Schema(
    {
      kind: String,
      amount: Number,
    },
    { timestamps: true },
  );

  // no `timestamps` option defined
  const ProfileSchema = new mongoose.Schema({
    conditions: {
      type: [ConditionSchema],
    },
  });

  const UserSchema = new mongoose.Schema(
    {
      name: String,
      profile: {
        type: ProfileSchema,
      },
    },
    { timestamps: true },
  );

  const User = mongoose.model("User", UserSchema);
```

### Steps to Reproduce

```
// $set does not update timestamps when inserting / replacing whole node
User.findOneAndUpdate(
          { name: 'Xyz' },
          { $set: { profile: {conditions: [{ kind: 'price', amount: 10 }] } } },
);
```

```
// $push does update timestamps when inserting to node
User.findOneAndUpdate(
          { name: 'Xyz' },
          { $push: { 'profile.conditions': { kind: 'price', amount: 10 } } },
);
```

### Expected Behavior

Timestamps should also update on nested subdocuments.

vkarpov15:
Fix #12119

**Summary**

#12119 is due to the fact that `applyTimestampsToChildren()` stops when it finds a schema that doesn't have `timestamps: true`. With this PR, `applyTimestampsToChildren()` will keep going to find any child schemas that have `timestamps: true`.

**Examples**

### DIFF
--- a/lib/helpers/update/applyTimestampsToChildren.js
+++ b/lib/helpers/update/applyTimestampsToChildren.js
@@ -82,11 +82,14 @@ function applyTimestampsToChildren(now, update, schema) {
 function applyTimestampsToDocumentArray(arr, schematype, now) {
   const timestamps = schematype.schema.options.timestamps;
 
+  const len = arr.length;
+
   if (!timestamps) {
+    for (let i = 0; i < len; ++i) {
+      applyTimestampsToChildren(now, arr[i], schematype.schema);
+    }
     return;
   }
-
-  const len = arr.length;
 
   const createdAt = handleTimestampOption(timestamps, 'createdAt');
   const updatedAt = handleTimestampOption(timestamps, 'updatedAt');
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

Issue #12157, opened 2022-07-27
Document.prototype.$set doesn't work per the documenation when passing path as an object of key/value pairs


randy5235:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.3.4

### Node.js version

16.x.x

### MongoDB server version

4.2

### Description

The document.set documentation allows for passing either (path: string, value: string, type: string, options: object)
or
(path: object, type: string, option: object)
or 
(path: object, options: object)

with type always optional.

It currently ignores passing path as an object.

### Steps to Reproduce

```
const mongoose = require('mongoose');
const connection = mongoose.connect('connectionString');

const testSchema = new mongoose.Schema({
  email: String,
  name: String,
});

const Test = mongoose.model('Test', testSchema);

// insert a couple of records doesn't really matter what

const myUser = Test.find({_id: <id from generated record>});
// This doesn't work as expect
myUser.set( { path1 : 'Value', path2: 'Value'}, {strict: false});
// setting them like this does
myUser.set('path1', 'Value', {strict: false});
myUser.set('path2', 'Value', {strict: false});



### Expected Behavior

Expected behavior:

Expectation is that the documented behavior is honored.

**Comment by randy5235** [comment]
I opened this PR to resolve this behavior: https://github.com/Automattic/mongoose/pull/12158

**Comment by IslandRhythms** [comment]
```javascript
const mongoose = require('mongoose');
mongoose.set('debug', true);
const testSchema = new mongoose.Schema({
  email: String,
  name: String,
});

const Test = mongoose.model('Test', testSchema);


async function run() {
    // insert a couple of records doesn't really matter what
    await mongoose.connect('mongodb://localhost:27017');
    await mongoose.connection.dropDatabase();

    const user = await Test.create({ name: 'Test Testerson', email: 'test@localhost.com'})
    const myUser = Test.find({ _id: user._id });
    // This doesn't work as expect
    myUser.set( { path1 : 'Value', path2: 'Value'}, {strict: false});
    // setting them like this does
    myUser.set('path1', 'Value', {strict: false});
    await myUser.set('path2', 'Value', {strict: false});
    console.log(await Test.findOne());
}

run();

```

IslandRhythms:
Could also be a duplicate of #12155

randy5235:
I'm closing this. After rereading the code that is in place. I suspect that my interpretation of the signature:
[...]

PR #12158, opened 2022-07-27
Update document.js

randy5235:
Fix for Issue #12157

**Summary**
Allow document.set to take an object of key/value pairs for path value, per the documentation.

**Examples**

Please see issue #12157 for example code.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -993,6 +993,12 @@ Document.prototype.overwrite = function overwrite(obj) {
  */
 
 Document.prototype.$set = function $set(path, val, type, options) {
+  if(utils.isPOJO(path) {
+     options = type;
+     type = val;
+     val = undefined;
+   }
+  
   if (utils.isPOJO(type)) {
     options = type;
     type = undefined;
--- a/lib/document.js
+++ b/lib/document.js
@@ -993,7 +993,7 @@ Document.prototype.overwrite = function overwrite(obj) {
  */
 
 Document.prototype.$set = function $set(path, val, type, options) {
-  if(utils.isPOJO(path) {
+  if (utils.isPOJO(path) {
      options = type;
      type = val;
      val = undefined;
--- a/lib/document.js
+++ b/lib/document.js
@@ -993,7 +993,7 @@ Document.prototype.overwrite = function overwrite(obj) {
  */
 
 Document.prototype.$set = function $set(path, val, type, options) {
-  if (utils.isPOJO(path) {
+  if (utils.isPOJO(path)) {
      options = type;
      type = val;
      val = undefined;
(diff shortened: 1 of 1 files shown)


lib/document.js
```diff
@@ -993,6 +993,12 @@ Document.prototype.overwrite = function overwrite(obj) {
  */
 
 Document.prototype.$set = function $set(path, val, type, options) {
+  if (utils.isPOJO(path) {
```
hasezoey, line 996:
```suggestion
  if (utils.isPOJO(path)) {
```

you are currently missing a ending `)`, so the tests fail

hasezoey (changes requested):
if possible, it would also be good to have tests for this

Uzlopak (commented):
fix linting issues

Uzlopak:
Well the code is totally broken.

I assume the right way would be something like

```js
  if (utils.isPOJO(path)) {
    for (const pojoPath of Object.keys(path)) {
      this.$set(pojoPath, path[pojoPath], options)
    }
    return this
  }
```

Uzlopak (changes requested):
This is not fixing the issue. I would actually prefer to fix the typings.

lib/document.js
randy5235: Thanks, thats what I get for using the online editor.

randy5235:
I missed a parenthesis, Sorry I created this via the github UI. I'll get a proper update in shortly.  Regarding the fix, the function already handles this case (as stated per the documentation), it just doesn't switch the parameters as needed to apply it. If you look at the existing code below mine, you see how it handles the types, My code checks if the path parameter is an object,  if so, it moves the rest of the parameters forward essentially since val is no longer needed (as the values should be contained within the object of key/value pairs.  The existing code the currently handles the types. The issue I see here is that if passing path as an object, it will apply any given type to all keys(paths). @Uzlopak what are your thoughts on allowing a Type object of key:value pairs (Path: Type) to assign per path?

lib/document.js
```diff
@@ -993,6 +993,12 @@ Document.prototype.overwrite = function overwrite(obj) {
  */
 
 Document.prototype.$set = function $set(path, val, type, options) {
+  if (utils.isPOJO(path)) {
```
vkarpov15, line 996: Make this check `if (utils.isPOJO(path) && arguments.length === 2)`

vkarpov15 (changes requested):
The following syntax currently works:

```
myUser.set( { path1 : 'Value', path2: 'Value'}, null, {strict: false});
```

I can see the benefit of supporting `myUser.set(obj, options)`, but I'd rather not break the current syntax.

PR #12160, opened 2022-07-27
style(buffer): change ".split" to "utils.each" for defining methods

hasezoey:
**Summary**

This PR changes in `buffer.js` a `.split` call into a `utils.each` call for defining functions, to be consistent


Uzlopak:
Thanks

PR #12166, opened 2022-07-28
Update populate.md

amanfrinati:
Fix Populating Maps example

**Summary**

The "librarySchema" example in the Populating Maps chapter contains a syntax error


Uzlopak:
Thanks

PR #12167, opened 2022-07-28
fix(types): pass type to mongodb bulk write operation

emiljanitzek:
**Summary**

Found issue when trying to perform a `bulkWrite` using `$addToSet`. The mongodb definition of `AnyBulkWriteOperation`requires the type to be passed to accept any input to $addToSet

**Examples**

See test

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -282,6 +282,33 @@ function bulkWrite() {
   M.bulkWrite(ops);
 }
 
+function bulkWriteAddToSet() {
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1: I took a look and this strangely enough seems to help performance

Before:

```
Instantiations:             364045
Memory used:               205244K
```

After:

```
Instantiations:             329480
Memory used:               197814K
```

Why? Who knows. But I'll take it.

PR #12168, opened 2022-07-29
🚑️ hotfix: ObjectId infererence

iammola:
## Summary

> **Note**
> I initially spoke about this issue [here](https://github.com/mohammad0-0ahmad-forks/mongoose/commit/d092fdcf06cf3063e2d35530676c00f2af4edbfb#r79603415). I was trying to get @mohammad0-0ahmad's opinion on it.

A required `ObjectId` field in a schema caused the inferred type to be deeply checked for other schema properties.

## Examples

```ts
const MySchema = new Schema({ 
  class: { type: Schema.Types.ObjectId, required: true },
  dob: Date
});

type MySchemaType = InferSchemaType<typeof MySchema>;

// When hovered over in VS Code
// type MySchemaType = {
//   dob?: unknown;
//   class?: {
//       _id?: ... | undefined;
//       equals?: {} | undefined;
//       id?: {
//           [x: number]: unknown;
//           equals?: {} | undefined;
//           set?: {} | undefined;
//           toJSON?: {} | undefined;
//           [Symbol.iterator]?: {} | undefined;
//           ... 99 more ...;
//           some?: {} | undefined;
//       } | undefined;
//       ... 6 more ...;
//       toString: {};
//   } | undefined;
// }
```

It only affects schemas with at least one required `ObjectId` field. This means that a schema like ⬇️ works fine.
```ts
const MySchema = new Schema({
  class: Schema.Types.ObjectId,
  dob: Date
});

type MySchemaType = InferSchemaType<typeof MySchema>;

// When Hovered in VS Code
type MySchemaType = {
    dob?: Date | undefined;
    class?: Types.ObjectId | undefined;
}
```

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -429,7 +429,7 @@ export function autoTypedSchema() {
     mixed3: Schema.Types.Mixed,
     objectId1: Schema.Types.ObjectId,
     objectId2: 'ObjectId',
-    objectId3: 'ObjectID',
+    objectId3: 'objectId',
     customSchema: Int8,
     map1: { type: Map, of: String },
     map2: { type: Map, of: Number },
--- a/types/inferschematype.d.ts
+++ b/types/inferschematype.d.ts
@@ -11,10 +11,11 @@ import {
   DateSchemaDefinition,
   ObtainDocumentType,
   DefaultTypeKey,
-  IfEquals,
-} from "mongoose";
+  ObjectIdSchemaDefinition,
+  IfEquals
+} from 'mongoose';
 
-declare module "mongoose" {
+declare module 'mongoose' {
   /**
    * @summary Obtains document schema type.
    * @description Obtains document schema type from document Definition OR returns enforced schema type if it's provided.
@@ -22,19 +23,11 @@ declare module "mongoose" {
    * @param {EnforcedDocType} EnforcedDocType A generic type enforced by user "provided before schema constructor".
    * @param {TypeKey} TypeKey A generic of literal string type."Refers to the property used for path type definition".
    */
-  type ObtainDocumentType<
-    DocDefinition,
-    EnforcedDocType = any,
-    TypeKey extends TypeKeyBaseType = DefaultTypeKey
-  > = IsItRecordAndNotAny<EnforcedDocType> extends true
-    ? EnforcedDocType
-    : {
-        [K in keyof (RequiredPaths<DocDefinition, TypeKey> &
-          OptionalPaths<DocDefinition, TypeKey>)]: ObtainDocumentPathType<
-          DocDefinition[K],
-          TypeKey
-        >;
-      };
+   type ObtainDocumentType<DocDefinition, EnforcedDocType = any, TypeKey extends TypeKeyBaseType = DefaultTypeKey> =
+   IsItRecordAndNotAny<EnforcedDocType> extends true ? EnforcedDocType : {
+     [K in keyof (RequiredPaths<DocDefinition, TypeKey> &
+     OptionalPaths<DocDefinition, TypeKey>)]: ObtainDocumentPathType<DocDefinition[K], TypeKey>;
+   };
 
   /**
    * @summary Obtains document schema type from Schema instance.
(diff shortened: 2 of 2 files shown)


Uzlopak:
@mohammad0-0ahmad

mohammad0-0ahmad:
> @mohammad0-0ahmad 

Thanks @Uzlopak  for mention, It might be helpful if you check this disscution

types/inferschematype.d.ts
```diff
@@ -160,7 +159,7 @@ type ResolvePathType<PathValueType, Options extends SchemaTypeOptions<PathValueT
           PathValueType extends DateSchemaDefinition ? Date :
             PathValueType extends typeof Buffer | 'buffer' | 'Buffer' | typeof Schema.Types.Buffer ? Buffer :
               PathValueType extends BooleanSchemaDefinition ? boolean :
-                PathValueType extends ObjectIdSchemaDefinition ? Types.ObjectId :
+                PathValueType extends 'objectId' | 'ObjectId' | typeof Schema.Types.ObjectId ? Types.ObjectId :
```
vkarpov15, line 162: Why not just add 'objectid' to `ObjectIdSchemaDefinition` ?

mohammad0-0ahmad: @vkarpov15 
Which values string values that should be resolved to object id ?

test/types/schema.test.ts
```diff
@@ -429,7 +429,7 @@ export function autoTypedSchema() {
     mixed3: Schema.Types.Mixed,
     objectId1: Schema.Types.ObjectId,
     objectId2: 'ObjectId',
-    objectId3: 'ObjectID',
+    objectId3: 'objectId',
```
mohammad0-0ahmad, line 432: @vkarpov15  
It is a bit confusing about the supported strings that should be resolved to ```Types.ObjectId```,
Please clarify which ones that are supported.

mohammad0-0ahmad: @iammola 
Please add this test to the end of models.test.ts
```ts
function gh12168() {
  const TermSchema = new Schema({
    session: { type: Schema.Types.ObjectId, required: true, ref: '...' },
    start: { type: Date, required: true, unique: true },
    end: { type: Date, required: true, unique: true }
  });

  const TermModel = model('test', TermSchema);

  expectType<{ session: Types.ObjectId; start: Date; end: Date; }>({} as InferSchemaType<typeof TermSchema>);
}
```

types/inferschematype.d.ts
mohammad0-0ahmad: @iammola 
You might need to as well to make this change as well:
```diff
 - PathValueType extends Record<string, any> ? ObtainDocumentType<PathValueType, any, TypeKey> :
+ PathValueType extends Record<string, any> ? IfEquals<PathValueType, Types.ObjectId, Types.ObjectId, ObtainDocumentType<PathValueType, any, TypeKey>> :

```

mohammad0-0ahmad: @vkarpov15  Please tell me when you feel that we can't improve the performance because of auto typed schema, so I can the entire feature, I wouldn't like to make mongoose worse.

mohammad0-0ahmad: I ran benchmark test after deleting the entire types folder, I got this result:
We just might consider that, I think.
Capture

Uzlopak: Deleting all the types is a quite radical approach, imho

mohammad0-0ahmad: I was just wondering what I will get if it has no types.

mohammad0-0ahmad: I think it is weird to get this result even if the package have no type definition files. 
```
Instantiations               15723
Memory⠀used             129661K
Check⠀time                  1.36s
Total⠀time                    2.31s
```

Uzlopak: I guess tsc loads also typings from node_modules and maybe global Modules

mohammad0-0ahmad: Sure, but  It might be more helpful to be able as well to diagnose only specific package type definitions.

test/types/schema.test.ts
vkarpov15: @mohammad0-0ahmad I took a look and it doesn't look like we support 'objectid', the below JavaScript throws an error:

```
'use strict';

  fooId: 'objectid' // Throws "TypeError: Invalid schema configuration: `Objectid` is not a valid type at path `fooId`."
});

console.log('Done');
```

So the 'objectid' change is unnecessary and incorrect

And I would rather avoid reverting the auto typed schema feature. It does make TypeScript a little slower, but on the other hand it also removes the need for document interfaces. We'll figure out some other way to make things faster. I've learned that there's always something you can do to come up with functionally equivalent types that don't cause tsc to grind to a halt, it's just a guessing game to figure out what that is :)

vkarpov15:
I'm going to close this, since it doesn't look like using 'objectid' instead of 'ObjectId' is supported

test/types/schema.test.ts
mohammad0-0ahmad: Thanks, @vkarpov15, That makes me feel better. Without doubt, we will find a solution for that :)

mohammad0-0ahmad:
@vkarpov15 
we need actually to have the following changes because the current types don't work correctly with this test, which was provided by @iammola in the mention discussion above.

```ts 
// in models.test.ts
function gh12168() {
  const TermSchema = new Schema({
    session: { type: Schema.Types.ObjectId, required: true, ref: '...' },
    start: { type: Date, required: true, unique: true },
    end: { type: Date, required: true, unique: true }
  });

  const TermModel = model('test', TermSchema);

  expectType<{ session: Types.ObjectId; start: Date; end: Date; }>({} as InferSchemaType<typeof TermSchema>);
}
```
```diff
// inferschematype.d.ts line 171
-PathValueType extends 'objectId' | 'ObjectId' | typeof Schema.Types.ObjectId ? Types.ObjectId :
+PathValueType extends ObjectIdSchemaDefinition ? Types.ObjectId :
```

iammola:
#12205

mohammad0-0ahmad:
Hello @vkarpov15 !
Would you like to recheck the mentioned issue by @iammola and reopen this PR?
So @iammola  can refactor it and maybe apply the suggested changes.

vkarpov15:
@mohammad0-0ahmad yeah I made a note to check out #12205.

PR #12196, opened 2022-08-02
fix(types): add schema plugin option inference

hasezoey:
**Summary**

This PR adds types for plugin options to be automatically inferred, if defined, with possibility to overwrite.

This inferring is currently strict, because if it would not be then on the first definition it would just overwrite the inferred type and only intellisense suggestions would remain.


hasezoey:
force pushed to fix lints in one commit

vkarpov15 (approved):
Works for me :+1: I took a look and it doesn't look like this causes any meaningful performance degradation.

PR #12197, opened 2022-08-02
chore(tidelift-alignment): run only in main repository, not in forks

hasezoey:
**Summary**

This PR disabled the workflow `tidelift-alignment` on forks, because it would otherwise always run but error because of missing secrets.


AbdelrahmanHafez (approved):

AbdelrahmanHafez:
Thanks @hasezoey 👍

PR #12207, opened 2022-08-03
Chore: Add test for auto typed approach.

mohammad0-0ahmad:
Related to https://github.com/Automattic/mongoose/issues/10349#issuecomment-1201805483
Note: the new added test fails because the test folder "benchmarks\typescript\auto-typed" doesn't exist in master branch yet.
It should work as it should when target master and targeted branch does have the mentioned test folder.

### DIFF
--- a/.github/workflows/benchmark-auto-typed.yml
+++ b/.github/workflows/benchmark-auto-typed.yml
@@ -1,0 +1,37 @@
+name: TypeScript Benchmark "auto-typed-mongoose"
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/benchmark.yml"
+      - "package.json"
+      - "types/**"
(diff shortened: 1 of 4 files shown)


.github/workflows/benchmark-auto-typed.yml
```diff
@@ -0,0 +1,37 @@
+name: TypeScript Benchmark "auto-typed-mongoose"
```
vkarpov15, line 1: Let's keep this in the `benchmark.yml` file to minimize overhead. Just add 2 extra steps to `npm install` and run ts-benchmark in auto-typed directory.

vkarpov15 (changes requested):

vkarpov15:
I'm going to close this PR for now, it is long out of date and I'm not sure it is necessary.

PR #12212, opened 2022-08-04
Model's type parameter must be defined

### LINKED ISSUE #12213: Compile error with Typescript 4.8 beta on type from mongodb@4.8.1
### Prerequisites

### Mongoose version

6.5.1

### Node.js version

16.15.1

### MongoDB server version

4.8.1

### Description

In Typescript 4.8 and mongodb [since at least 4.8.1](https://github.com/Automattic/mongoose/commit/547e5fc92051ca602c7c7d16b2518a5745711711), `mongodb.AnyBulkWriteOperation` requires its type argument to be defined. It forbids null and undefined. Previous versions of Typescript didn't check this properly for unconstrained type parameters (lone type parameters like T). Typescript 4.8, which is in beta and will release in a week or two, does.

### Steps to Reproduce

1. Open types/models.d.ts in an editor running Typescript 4.8
2. Compile errors at Model.bulkWrite on lines 147-149.

Found by Definitely Typed's nightly run: https://dev.azure.com/definitelytyped/DefinitelyTyped/_build/results?buildId=136307&view=logs&j=6399d89d-3d30-5fff-3e92-07bbaf7ebe2f&t=482eae7c-85d2-5e8a-ea70-a5ab41e9bc13

### Expected Behavior

No compile error.

sandersn:
In Typescript 4.8 and mongodb [since at least 4.8.1](https://github.com/Automattic/mongoose/commit/547e5fc92051ca602c7c7d16b2518a5745711711), `mongodb.AnyBulkWriteOperation` requires its type argument to be defined. It forbids `null` and `undefined`. Previous versions of Typescript didn't check this properly for unconstrained type parameters (lone type parameters like `T`). Typescript 4.8, which is in beta and will release in a week or two, does.

This PR adds a constraint to Model that requires its type parameter to be defined. It can no longer include `null` or `undefined`. In particular, that means `Model<unknown>` is no longer allowed; you have to write `Model<{}>` instead, which is the same type excluding `null` and `undefined` [1]. You can see this change in test/types/document.test.ts:179:

```ts
    expectType<Model<unknown>>(this.$model('Item1'));
// becomes
    expectType<Model<{}>>(this.$model('Item1'));
```

This change ripples out to other types as well, most notably the type parameter for `Schema`, which now also disallows `null` and `undefined`.

#### Alternative Fix

This is a pretty comprehensive PR. It should be possible to isolate the change to `Model.bulkWrite` by adding a conditional type:

```ts
mongodb.AnyBulkWriteOperation<T>
// would become
mongodb.AnyBulkWriteOperation<T extends {} ? T : any>
```

But this is unsafe since you can pass in writes that don't match Model's `T`, and because presumably AnyBulkWriteOperation really doesn't expect `null` or `undefined` and might crash at runtime.

I wouldn't recommend this smaller alternative except temporarily.

[1] `unknown` means "any value in Javascript". `{}` means "any value except null or undefined". `object` means "any non-primitive" (and does *not* allow null)

#### Recommendation

I'd recommend testing with typescript@beta during the Typescript beta period or, even better, typescript@next.

Fixes #12213

### DIFF
--- a/test/types/document.test.ts
+++ b/test/types/document.test.ts
@@ -176,7 +176,7 @@ function gh11435() {
   const ItemSchema = new Schema<Item>({ name: String });
 
   ItemSchema.pre('validate', function preValidate() {
-    expectType<Model<unknown>>(this.$model('Item1'));
+    expectType<Model<{}>>(this.$model('Item1'));
   });
 }
 
--- a/types/connection.d.ts
+++ b/types/connection.d.ts
@@ -142,13 +142,13 @@ declare module 'mongoose' {
     readonly models: Readonly<{ [index: string]: Model<any> }>;
 
     /** Defines or retrieves a model. */
-    model<T, U, TQueryHelpers = {}>(
+    model<T extends {}, U, TQueryHelpers = {}>(
       name: string,
       schema?: Schema<T, any, any, TQueryHelpers, any, any>,
       collection?: string,
       options?: CompileModelOptions
     ): U;
-    model<T>(name: string, schema?: Schema<T>, collection?: string, options?: CompileModelOptions): Model<T>;
+    model<T extends {}>(name: string, schema?: Schema<T>, collection?: string, options?: CompileModelOptions): Model<T>;
 
     /** Returns an array of model names created on this connection. */
     modelNames(): Array<string>;
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -71,7 +71,7 @@ declare module 'mongoose' {
     $markValid(path: string): void;
 
     /** Returns the model with the given name on this document's associated connection. */
-    $model<ModelType = Model<unknown>>(name: string): ModelType;
+    $model<ModelType = Model<{}>>(name: string): ModelType;
 
     /**
      * A string containing the current operation that Mongoose is executing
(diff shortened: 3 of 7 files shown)


sandersn:
I'm stuck trying to figure out how the type tests break, so I created #12221 for the small temporary fix, which will unblock people using Typescript 4.8 with mongoose.

vkarpov15 (commented):
@sandersn sorry for the late reply. Why is this change necessary? It looks like Mongoose's types tests work fine with `typescript@4.8.1-rc`. Can you come up with a script that fails to compile on 4.8.1-rc1 but should succeed?

sandersn:
I believe that the repro steps in #12213 still fail if you revert #12221 .

The problem is that #12221 makes typing less safe when the type parameter includes `undefined`. I believe it's not *supposed* to, but it would be better if all of the types correctly forbid `undefined` as with this PR. Unfortunately, I never got it to pass tests and ran out of time to work on it. You can close it if you don't want to pass it on to somebody else.

vkarpov15:
I'll close for now. This doesn't seem like a huge problem, but I'm open to opposing viewpoints on that.

PR #12216, opened 2022-08-05
Remove Unused PUG helpers

hasezoey:
**Summary**

This PR removes unused PUG helpers:
- `klass.js`
- `linktype.js`
- `href.js`

I removed them because i could not find them used anywhere in any pug files or pug itself

before merge, someone who knows about those helpers, please say if they were used anywhere that i didnt notice


hasezoey:
Force pushed to rebase onto latest master

Uzlopak (approved):
those should be artifacts, which were left when I moved the old docs into gh-pages

PR #12223, opened 2022-08-06
docs(api.pug): change to redirect to mongoose API page

### LINKED ISSUE #12217: [Docs] API documentation is generated multiple times
### Prerequisites

### Issue

I have just noticed that API content (at least for `Mongoose`) is generated 2 times, once via `docs/splitApiDocs.js` and `api.pug` and once with `docs/source/api.js` and `docs/api_split.pug` and so resulting in multiple paths, with "should be same content" at `https://mongoosejs.com/docs/api.html` and `https://mongoosejs.com/docs/api/mongoose.html`.

This issue is about discussing / tracking of combining this, this is also a question on if there is a reason to keep it this way.

PS: i noticed this while doing #12218

hasezoey:
**Summary**

This PR addresses #12217, with the approach of using a script to redirect based on the anchor

fixes #12217

**Note: This description was changed later, so the earlier comments might reference older information**

### DIFF
--- a/docs/api.pug
+++ b/docs/api.pug
@@ -6,4 +6,4 @@ append style
 
 block content
   p Redirecting to proper API page, please wait
-  noscript This Page required JavaScript to Redirect old links properly
+  noscript This Page requires JavaScript to Redirect old links properly
--- a/docs/api.pug
+++ b/docs/api.pug
@@ -2,66 +2,8 @@ extends layout
 
 append style
   link(rel="stylesheet", href="/docs/css/api.css")
-  script(src="/docs/js/api-bold-current-nav.js")
+  script(src="/docs/js/redirect-old-api.js")
 
 block content
-  h1 API Docs
-
-  include includes/native
-
-  div.api-nav
-    div.api-nav-content
-      each item in docs
-        - if (!item.hideFromNav)
-          div.nav-item(id='nav-' + item.name)
-            div.nav-item-title
-              a(href='./api/' + item.name.toLowerCase() + '.html')
-                | #{item.name}
-            ul.nav-item-sub
-              each prop in item.props
-                li
-                  a(href='./api/' + item.name.toLowerCase() + '.html#' + prop.anchorId)
-                    | #{prop.string}
-
-  each item in docs
-    hr.separate-api
-    div.item-header-wrap
-      <a class="edit-docs-link" href="#{item.editLink}" target="_blank">
-        <img src="/docs/images/pencil.svg" />
-      </a>
-      h2(id=item.name, class="item-header")
-        a(href='#' + item.name)
-          | #{item.name}
-    ul
-      each prop in item.props
-        li
-          a(href='#' + prop.anchorId)
-            | #{prop.string}
-    each prop in item.props
-      hr.separate-api-elements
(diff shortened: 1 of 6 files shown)


hasezoey:
Merged latest master to resolve conflicts

vkarpov15 (commented):
I'm a little concerned about this breaking existing links to the api page. Any anchor that links to a header that isn't in `/api/mongoose` will break.

We'd like to make this change eventually, that's why we added `api_split.pug`, but still wary of broken links. What do you think @hasezoey ?

hasezoey:
> I'm a little concerned about this breaking existing links to the api page. Any anchor that links to a header that isn't in /api/mongoose will break.

like i said, with this current way, it would break all anchor links to the old page

the other workaround would be to redirect via script, with anchors

hasezoey:
@vkarpov15 should i change it to a js script that redirects to the correct page *with* anchor instead of a meta-refresh?

hasezoey:
i just noticed that because `api.html` includes *all* pages, just redirecting to `mongoose.html` (even with proper anchor) would not be correct or working, so what should we do about that?
options would be:
- dont redirect with anchor at all
- only redirect to `mongoose.html` and have only those anchors that exist there working
- implement a custom script to redirect properly everywhere

@vkarpov15

hasezoey:
- Updated Implementation to be a script that tries to use the anchor to select the file (base case is `api/mongoose.html`)
- change all links that used `api.html` to link to the proper place
- add proper title to all generated API outputs (previously pug-variable `title` was unset)
- change `api.js` to only export the `docs` array

The only thing that still needs to be updated is `search.js`, but i dont know enough about it so i would like to ask @vkarpov15 to take care of that

hasezoey:
added to the 6.7 milestone to be considered for merge, because i think this PR is ready and handles all (that i know of) cases

vkarpov15:
We'll hold off on this one until 7.0, I'm still a little uncomfortable with merging this one due to SEO concerns. But should be fine to merge with 7.0 release.

hasezoey:
> We'll hold off on this one until 7.0, I'm still a little uncomfortable with merging this one due to SEO concerns. But should be fine to merge with 7.0 release.

sure, would it be alright if some changes (like links to `api.html` in code & guides) are cherry-picked and put in another PR for direct merging?

vkarpov15:
@hasezoey sure

hasezoey:
Rebased on latest 6.8 branch, fixes because of #12644 and #12645

PS: because this was rebased on 6.8 instead of the base (which is `master`), there are a lot of unrelated commits showing, they should be gone again once changing the base to `7.0` and having 6.8 merged in to `7.0`

hasezoey:
force-push to rebase on latest master (clean-up the PR commits)

hasezoey:
rebased onto latest `7.0` to lessen commit history (making it easier to see changes), also to check if everything works

vkarpov15 (approved):
Thanks :+1:

hasezoey:
@vkarpov15 

just asking to not forget it:
could you update the search to use the actual pages, i did not update it because i am not quite sure how it works (the last time i tested and the current live version still **only** links to `api.html`)

PR #12201, opened 2022-08-02
fix: Update ChangeStream hydration condition

ramos-ph:
**Summary**

There is an issue when using a `ChangeStream` in a replica set, you'll get an `undefined` error when you try to call `mongoose.disconnect()`:

```js
await mongoose.connect(`mongodb://${host}:${port}/?replicaSet=${rset}`, {
  autoIndex: false,
});

const changeStream = mongoose.connection.watch(pipeline, {
  fullDocument: "updateLookup",
});
const stream = changeStream.on("change", (change) => {
  console.log(change);
});

await stream.close();
await mongoose.disconnect(); // `data.fullDocument` is undefined
```

The issue was introduced in one of the commits from versions 6.4.7 to 6.5.0, as stated in #12169. This PR fixes this by correctly addressing where the options are coming from.

**Examples**

Here's the result for the current version:
image

Here's the result for the fix:
image

Closes #12169

PR #12224, opened 2022-08-07
fix(ChangeStream): try to fix "cannot read property fullDocument of undefined"

hasezoey:
**Summary**

This PR tries to fix the errors `Uncaught TypeError: Cannot read properties of undefined (reading 'fullDocument')` that were often thrown in ReplicaSet tests, see [This Run](https://github.com/Automattic/mongoose/runs/7662155766?check_suite_focus=true) as a example.

PS: i dont know if this fully fixes it, but it should at least help


vkarpov15 (approved):
Thanks :+1:

hasezoey:
just noticed that this was a Duplicate of #12201

Issue #11999, opened 2022-06-28
[Test] `clearTestData` may fail with `MongoServerError: all indexes on collection dropped`


hasezoey:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Currently `test/utils.js` function `clearTestData` currently may fail with `MongoServerError: all indexes on collection dropped` because all documents in the collection are deleted with `deleteMany` (instead of collection drop) and at the same time all indexes are dropped, which does not work well

Proposal to fix:
only have one delete at a time happening, meaning only delete all documents then delete all indexes or vise-versa (or just use `dropCollection` if performance has increased)

---

Log showing the error: https://github.com/Automattic/mongoose/runs/7088863488?check_suite_focus=true#step:6:4066

PR #12236, opened 2022-08-08
Drop collections between tests, fix a couple of issues with alternative drivers

vkarpov15:
**Summary**

Re: #11999, we get some issues with dropping indexes and deleting documents at the same time in tests. The reason why we do `dropIndexes()` and `deleteMany()` in between test cases is that MongoDB's WiredTiger storage engine used to be incredibly slow when dropping databases and collections - it used to take over 60 seconds to `dropDatabase()`. These days the difference isn't anywhere near that bad. On my local, it takes about 4 seconds to run document tests with WT using `dropIndex()` and `deleteMany()`, and 11 seconds with `collection.drop()`. We also typically use the in-memory storage engine for tests anyway.

Also fixed a couple of issues that we ran into: `bulkWrite()` and other collection methods can throw synchronous errors, so we need to be careful to catch those.

**Examples**


lib/model.js
```diff
@@ -3575,13 +3575,17 @@ Model.bulkWrite = function(ops, options, callback) {
...
+          if (error) {
+            return cb(error);
+          }
 
-        cb(null, res);
-      });
+          cb(null, res);
+        });
+      } catch (err) {
+        return cb(err);
```
AbdelrahmanHafez, line 3587: I've always didn't like the fact that bulkWrite could throw both sync and async errors, but I accepted it as a fact.
Glad this is changing now.

AbdelrahmanHafez (changes requested):
We have a few failing tests, otherwise LGTM. Thanks. 👍

IslandRhythms (approved):

test/model.test.js
```diff
@@ -4229,7 +4247,12 @@ describe('Model', function() {
       });
 
 
-      await Person.collection.drop();
+      await Person.collection.drop().catch(err => {
+        if (err.codeName === 'NamespaceNotFound') {
```
AbdelrahmanHafez, line 4251: Should mongoose, generally, ignore this error and return a fake result?
If I drop a collection and that connection does not exist, I wouldn't mind if I just receive a response saying that it's dropped.

What do you think? @vkarpov15

vkarpov15: @AbdelrahmanHafez we may consider that for a future release, but it's a pretty low priority.

AbdelrahmanHafez: I agree it's a low priority.

PR #12237, opened 2022-08-08
Extend JSDOC Return type for `Aggregate.prototype.exec`

hasezoey:
**Summary**

This PR does:
- extend JSDOC Return type for `Aggregate.prototype.exec`
- add a extra check in a aggregate test for sanity ([see this run](https://github.com/Automattic/mongoose/runs/7729209058))

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -963,7 +963,7 @@ Aggregate.prototype.pipeline = function() {
  *     const result = await aggregate.exec();
  *
  * @param {Function} [callback]
- * @return {Promise} Returns a Promise if no "callback" is given.
+ * @return {Promise<Document[]>} Returns a Promise if no "callback" is given.
  * @api public
  */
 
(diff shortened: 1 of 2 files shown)


lib/aggregate.js
```diff
@@ -963,7 +963,7 @@ Aggregate.prototype.pipeline = function() {
  *     const result = await aggregate.exec();
  *
  * @param {Function} [callback]
- * @return {Promise} Returns a Promise if no "callback" is given.
+ * @return {Promise<Document[]>} Returns a Promise if no "callback" is given.
```
AbdelrahmanHafez, line 966: I don't think that's correct.

Aggregate execution results:
1- Plain JS objects, mongoose doesn't do any parsing on them, so they do not extend the `Document` class.
2- Can be of any type, for example, you can `$addFields` or `$group`, which changes the shape of the return type from the defined schema to anything, that's why it's a design decision for mongoose not to do anything for aggregation results.

hasezoey: thanks for explaining, i will close this PR then, i had just guessed from some tests and the surrounding code (i dont know much about mongodb aggregate)

PR #12244, opened 2022-08-09
Fix some flakey tests from #11999

vkarpov15:
**Summary**

The change from #12236 introduced some flakey tests, this PR fixes those

**Examples**

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -7014,61 +7014,61 @@ describe('Model', function() {
       await Test.insertMany([
         {
           metadata: { sensorId: 5578, type: 'temperature' },
-          timestamp: new Date('2021-05-18T00:00:00.000Z'),
+          timestamp: Date.now() + 500,
           temp: 12
         },
         {
           metadata: { sensorId: 5578, type: 'temperature' },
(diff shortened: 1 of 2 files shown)


test/model.test.js
```diff
@@ -7006,79 +7006,16 @@ describe('Model', function() {
         autoCreate: false
       });
 
-      const Test = db.model('TestGH11229Var2', schema);
+      const Test = db.model('TestGH11229Var2', schema, 'TestGH11229Var2');
 
       await Test.collection.drop().catch(() => {});
       await Test.createCollection({ expires: '5 seconds' });
 
-      await Test.insertMany([
```
Uzlopak, line 7014: Didnt i write this unit test? But i understand that on expiration test is enough and this one just checks the conversion of ms to a number.

Uzlopak (approved):

test/model.test.js
vkarpov15: @Uzlopak I think that, in hindsight, testing that MongoDB actually expires the documents is a bit too brittle. The MongoDB server doesn't always expire the documents in time, and that can cause the tests to unexpectedly fail. So I think it is better to just make sure the server is configured to expire documents.

Issue #12254, opened 2022-08-10
Using Schema.loadClass with a class which inherits from Document will hang up any find operations.


WolverinDEV:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.5.2

### Node.js version

v16.13.0

### MongoDB server version

docker latest (c8b57c4bf7e3)

### Description

When using `Schema.loadClass` with an class inheriting from `Document`, loading such entries will stall forever.  
A simple fix would be:  
```ts
import { Document } from "mongoose";
import * as mongoose from "mongoose";

mongoose.Schema.prototype.loadClass = function(original: Function) {
    return function(this: mongoose.Schema, model: any, ...args: any[]) {
        if(model === Document) {
            /* Don't load mongoose Document class. */
            return;
        }

        return original.call(this, model, ...args);
    }
}(mongoose.Schema.prototype.loadClass);
```

### Steps to Reproduce

```ts
const DownloadJobSchema = new mongoose.Schema();

class B extends Document {}
DownloadJobSchema.loadClass(B);

/* Create at least one entry which could be found. */
this.jobsModel.create({ someKey: "with some data" });

/* Lets find that one entry... */
console.log("Finding one entry");
this.jobsModel.findOne({}).then(() => console.log("You'll never see me!"));
```

### Expected Behavior

`findOne` finding one entry in the database and return it.

PR #12255, opened 2022-08-10
Fixing issue #12254

WolverinDEV:
**Summary**
Lookup issue #12254 for more details.  
This PR fixes that issue.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -23,6 +23,7 @@ const setupTimestamps = require('./helpers/timestamps/setupTimestamps');
 const utils = require('./utils');
 const validateRef = require('./helpers/populate/validateRef');
 const util = require('util');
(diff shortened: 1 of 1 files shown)


lib/schema.js
```diff
@@ -23,6 +23,7 @@ const setupTimestamps = require('./helpers/timestamps/setupTimestamps');
 const utils = require('./utils');
 const validateRef = require('./helpers/populate/validateRef');
 const util = require('util');
+const Document = require('./document');
```
vkarpov15, line 26: This is unfortunately a circular import, `Document` is `{}` here.

vkarpov15 (changes requested):
As implemented, this PR doesn't work. I'm working on some alternative fixes. I'll close this for now.

PR #12256, opened 2022-08-11
feat: re-export default Mongoose instance properties for ESM named imports support

### LINKED ISSUE #12148: add esm support
### Prerequisites

### 🚀 Feature Proposal

issue #10513 was closed without fixing it. The issue is we still can't name import `Schema, model`.

To fix this, we can add a new file `index.mjs`. Name export all the structures like `Schema, model`,  default exports `mongoose`, and add export map in package.json.

**Alternative**

Export all the structures in `index.js` file like 
```ts
exports.Schema = Schema
```
and it works for both cjs and esm

### Motivation

_No response_

### Example

_No response_

vkarpov15:
Fix #12148

**Summary**

I did some reading in https://github.com/nodejs/node/pull/35249/files and https://simonplend.com/node-js-now-supports-named-imports-from-commonjs-modules-but-what-does-that-mean/, it looks like all we need to do to support ESM named imports in Node is to explicitly re-export all the properties on the Mongoose global.

This does come with some caveats. Most notably that `this` will no longer be an instance of `Mongoose` in all exported functions by default. This shouldn't break anything - we added support for destructured imports in https://github.com/Automattic/mongoose/commit/a73c2ab8cbc3472ee8f8545437533f456ea0eaa3. But it is still a risk.

I'm also not exporting `Connection` and `Collection`, because those are custom getters and overwriting those can break support for custom drivers.

**Examples**


index.js
```diff
@@ -11,3 +10,52 @@ const mongoose = require('./lib/');
...
+module.exports.startSession = mongoose.startSession;
+module.exports.pluralize = mongoose.pluralize;
+module.exports.model = mongoose.model;
+module.exports.deleteModel = mongoose.deleteModel;
+module.exports.modelNames = mongoose.modelNames;
+module.exports.plugin = mongoose.plugin;
+// module.exports.connection = mongoose.connection;
+module.exports.connections = mongoose.connections;
+// module.exports.Collection = mongoose.Collection;
+// module.exports.Connection = mongoose.Connection;
```
imranbarbhuiya, line 32: imo add a comment in code about why these things aren't exported. So in the future, when anyone checks these codes, they'll be able to understand why these things aren't exported.

imranbarbhuiya (changes requested):

Uzlopak (approved):
LGTM

index.js
Uzlopak: Do we export these by our typings to directly be instantiated? If so, we should not directly export them.

imranbarbhuiya: yeah. Typings exports them

AbdelrahmanHafez: I second @imranbarbhuiya 
A comment specifying why those are not exported would be nice.

Otherwise, LGTM. 👍

index.js
```diff
@@ -11,3 +10,52 @@ const mongoose = require('./lib/');
...
+module.exports.createConnection = mongoose.createConnection;
+module.exports.connect = mongoose.connect;
+module.exports.disconnect = mongoose.disconnect;
+module.exports.startSession = mongoose.startSession;
+module.exports.pluralize = mongoose.pluralize;
+module.exports.model = mongoose.model;
+module.exports.deleteModel = mongoose.deleteModel;
+module.exports.modelNames = mongoose.modelNames;
+module.exports.plugin = mongoose.plugin;
+// module.exports.connection = mongoose.connection;
```
hasezoey, line 29: why not export the default connection, is it also because of "custom driver may overwrite this"?

vkarpov15: Yes exactly. `setDriver()` can recreate the default connection.

index.js
vkarpov15: The weird quirk is that you can do `import { connection } from 'mongoose'` in TypeScript even without this change. This change is just for Node's particular ESM implementation.

I have an alternative idea: just use `Object.defineProperty()` to create a getter for these particular properties. That should work for Node's ESM implementation.

vkarpov15: Actually, looks like `defineProperty()` only works if you set `value`, not `get`:

```javascript
// Works
Object.defineProperty(module.exports, 'connection', {
  value: mongoose.connection
});

// Doesn't work
Object.defineProperty(module.exports, 'connection', {
  get: () => mongoose.connection
});
```

So it looks like this idea won't work.

2 options:

1. Leave `connection`, `Collection`, `Connection` commented out. You'll still be able to use them in TypeScript, both as named imports and as `const connection = mongoose.connection`. Just won't be able to use them as named imports with Node ESM.
2. Uncomment them. However, this is a potentially breaking change because any TypeScript code that uses a custom driver and `Connection` as a named import will (probably) no longer work.

I'm leaning towards (1) because that's the least risky approach. Will do some more tinkering and see if there are other considerations.

hasezoey: i think you could also just:
- define it on another object (or its own value) and then export it
- use a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) object (or similar)

though i did not try this, so these are just thoughts

vkarpov15: The problem is that Node ESM finds named imports by parsing the source, not by actually running it. So proxy wouldn't help. Defining `connection` etc. on another object would be better, but still a more significant change than I'd like.

Issue #2724, opened 2015-02-28
Model.remove() should allow passing options


vkarpov15:

hasezoey:
From what i can tell, the `remove` function accepts options, since ~5.7.4 402db1a8f455306b1e0146d3fa3c776e3ee50da0

https://github.com/Automattic/mongoose/blob/master/lib/model.js#L963-L976

so i think this can be closed
cc @vkarpov15 @Uzlopak @AbdelrahmanHafez

AbdelrahmanHafez:
Thanks @hasezoey

hasezoey:
@AbdelrahmanHafez this should maybe also be added to the proper milestone

PR #12258, opened 2022-08-11
fix(types): add parameter "options" to "Model.remove"

hasezoey:
**Summary**

This PR adds parameter `options` for `Model.remove`, which i noticed was missing when searching through #2724

---

currently `RemoveOptions` also extends from `mongodb.DeleteOptions` because the options are passed through to `deleteOne` (from what i could tell), but i dont know if mongoose wants to actually expose all of these options


AbdelrahmanHafez (approved):
LGTM, thanks. 👍

Issue #12257, opened 2022-08-11
Does mongoose fully support mongodb 6.0?


ZeroCho:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

I know that mongodb 6.0 is out and it needs mongodb nodejs driver 4.8 or newer.
As I can see in package.json of mongoose, mongoose uses mongodb@4.8, so I guess mongoose can support mongodb 6.0.
I opened this issue to know whether mongoose is tested on mongodb 6.0 and can be used along with mongodb 6.0. Can anybody tell me? Thank you in advance.

Uzlopak:
@vkarpov15 
@hasezoey 
Should we add mongo6 to our test pipeline?

Uzlopak:
There is no official mongo 6 image on docker hub...

hasezoey:
>  Should we add mongo6 to our test pipeline?

we could, at least from a MMS perspective, because 6.0.0 binaries are up at https://dl.mongodb.org/dl/linux, though MMS does not have any 6.0.0 tests yet - but i dont see anything currently preventing its usage

Uzlopak:
@hasezoey 

Well i think vkarpov wanted to use also mongodb memory for the cicd pipeline. Maybe you can provide a PR for using mms in the pipeline?

hasezoey:
> Maybe you can provide a PR for using mms in the pipeline?

is MMS not currently used in the Pipeline? if not, then what is it currently installed for?

Uzlopak:
@hasezoey 
In the cicd pipeline we spawn a mongo server without the use of mms. I added mms into mongoose codebase to easier test the mongoose unit tests locally. When i tried to implement it for the pipeline it was actually not that trivial. But maybe you can implement it?

Issue #12052, opened 2022-07-04
In-memory driver


vkarpov15:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

We did some work to in 6.4 to improve support for alternative drivers: using something other than the MongoDB Node driver to store data and execute queries. One potential application is an in-memory driver: when you call `create()`, the in-memory driver will store the document in memory rather than storing it in MongoDB, and `find()` will search through the in-memory data structure rather than executing a MongoDB query.

### Motivation

There's a couple of use cases for this:

1. Much faster test suites, and easier to set up CI workflows.
2. Easier to create a playground on mongoosejs.com (#10970), because in that case we don't have to set up a MongoDB instance or Node.js server to execute Mongoose code. We can just run the code directly in the browser.

### Example

`mongoose.setDriver(require('@mongoosejs/in-memory-driver'))`

Nicholaiii:
The Mongo memory server is a lot of overhead for testing, this would be a great feature

vkarpov15:
We can close this with https://github.com/mongoosejs/in-memory-driver/, which powers https://playground.mongoosejs.io/

PR #12262, opened 2022-08-11
Change Tests to use MMS over manually downloading & installing

hasezoey:
**Summary**

This PR changes the Tests to use MMS instead of requiring a active instance and "manually" downloading in CI tests

This PR also does:
- reset color after "No shard Testing" message
- change from "mocha.opts" file to a ".mocharc.yml"
- change "common.js"(test utils) to use URL generation based on `MONGOOSE_TEST_URI` for both `uri` and `uri2`
- change some tests from `assert.ok(number === 0)` to `assert.equal(number, 0)` to get better error messages (expire after seconds)
- change tests to always use `start.uri` over alias and static coding
- update used 5.0.x version used in test.yml from 5.0.2 to 5.0.8

This PR is a DRAFT because it is still WIP, it is WIP because:
- old code is commented out instead of removed (like setting up the replica sets & "test.yml" "manual" downloading)
- tests fail because some tests expect journaling, which is not available with `ephermeralForTest`, and `wiredTiger` yet to speed up tests (current count of failing tests is 2)

re https://github.com/Automattic/mongoose/issues/12257#issuecomment-1211883438
re https://github.com/Automattic/mongoose/issues/12052 (somewhat)


hasezoey:
merged latest master, updated the used mongodb version to their latest patch version (4.0.2 -> 4.0.28, 5.0.2 -> 5.0.8)

though i am still unsure what to do with the tests that require journaling

hasezoey:
Updated to latest master and applied changes for #12356

Issues still to resolved:
- there are tests which check something with journaling, but in-memory engine is used, what should be done about those tests?

.github/workflows/test.yml
```diff
@@ -69,16 +71,16 @@ jobs:
 
       - run: npm install
 
-      - name: Setup
```
Uzlopak, line 72: maybe it makes sense to cache the downloaded mongo server to avoid downloading the mongo server again and again?

https://github.com/actions/cache

hasezoey: > maybe it makes sense to cache the downloaded mongo server to avoid downloading the mongo server again and again?

true, will do once i figure out what the current CI problem is

hasezoey: done in 518ee0e46ee655880b031f2e8fa5e3fc5427a27c, though it will not take effect until at least one CI passes

hasezoey:
Updated to fix CI issues

test/mocha-fixtures.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  if (startMemoryReplset) {
+    mongorreplset = await mms.MongoMemoryReplSet.create({ replSet: { count: 3, args: ['--setParameter', 'ttlMonitorSleepSecs=1'] } }); // using 3 because even numbers can lead to vote problems
+    replseturi = mongorreplset.getUri();
+  } else {
+    replseturi = '';
+  }
+
+  process.env.MONGOOSE_TEST_URI = instanceuri;
```
vkarpov15, line 46: Can we set these as `global` variables rather than overwriting `process.env`? I get why setting these variables is necessary, I just really don't like overwriting `process.env`.

test/mocha-fixtures.js
```diff
@@ -0,0 +1,61 @@
...
+  // set some options when running in a CI
+  if (process.env.CI) {
+    process.env.MONGOMS_PREFER_GLOBAL_PATH = '1'; // set MMS to use "~/.cache/mongodb-binaries" even when the path does not yet exist
+  }
+
+  if (startMemoryInstance) { // Config to decided if an mongodb-memory-server instance should be used
+    // it's needed in global space, because we don't want to create a new instance every test-suite
+    mongoinstance = await mms.MongoMemoryServer.create({ instance: { args: ['--setParameter', 'ttlMonitorSleepSecs=1'] } });
+    const uri = mongoinstance.getUri();
+    instanceuri = uri.slice(0, uri.lastIndexOf('/'));
```
vkarpov15, line 34: Why `lastIndexOf('/')`? I think we should keep the db name in the connection string...

test/mocha-fixtures.js
```diff
@@ -0,0 +1,61 @@
...
+  } else {
+    instanceuri = process.env.MONGOOSE_TEST_URI;
+  }
+
+  if (startMemoryReplset) {
+    mongorreplset = await mms.MongoMemoryReplSet.create({ replSet: { count: 3, args: ['--setParameter', 'ttlMonitorSleepSecs=1'] } }); // using 3 because even numbers can lead to vote problems
```
vkarpov15, line 40: Can we do `count: 1`? `count: 3` takes longer to start up and we don't do things that would require `count: 3`, like testing replica set failover.

vkarpov15 (commented):
Looking good so far, I like this idea :+1:

test/mocha-fixtures.js
hasezoey: i overwrite a `process.env` variable, because that is the exact variable that was already used in `test/common.js`, and so was the easiest way of implementing it (i dont know if `global` carries over, at least in jest it does not)

test/mocha-fixtures.js
hasezoey: > Why lastIndexOf('/')? I think we should keep the db name in the connection string...

we can keep it, but it would get overwritten anyway again by `test/common.js`, do you know of a better way to overwrite the dbname while keeping everything else while it is still a string (ie in `test/common.js`)?

test/mocha-fixtures.js
hasezoey: i changed it to `3` because the previous value was `2`, and mongodb recommends to not use EVEN numbers, so instead of going down to 1 i had gone up to 2, see https://github.com/Automattic/mongoose/blob/97c3bdb2260addb0d5a25f12bc58ded10aefdbe0/test/common.js#L234 for previous code

test/mocha-fixtures.js
Uzlopak: split by / 
pop last element to nirvana
push dbName 
Join by /

hasezoey: >  pop last element to nirvana

isnt that what is currently done?

because the original URI would be something like `mongodb://localhost:9999/dbname?queryoptions` in this case `dbname?queryoptions` would be dropped (what currently is done)
though i guess yes we could slice by `/` and `?` (if available)

Uzlopak: Yeah. 
Or you do a more complex replace with regex.

I think the complexity with regex is not worth it.

hasezoey: removed with efaf89b2be02f540c54dac1795b2fcb595a21999
proper handling added in 48973effae9ccea2f62bfe2c7bde3e1fe52620ad

hasezoey:
Update:
- merged latest master
- updated all MMS configurations to use `storageEngine: 'wiredTiger'`

i changed the storage-engine to be `wiredTiger` because this is the only engine that has journaling which is needed for some tests and it seems like mongodb is removing (or possibly renaming?) `ephermerForTest` in 6.1, but does not have yet any changelog entry (and it is also not mentioned in the mongodb documentation anymore from what i can tell)
tests still pass at ~40-60 seconds (normal instance + replset count 3)

hasezoey:
actually convert it back to draft, because i am seeing some MMS startup errors, will try to fix those before un-drafting again

hasezoey:
Updated MMS to fix the problems i had found, i think the test now failing is not related with changing to MMS

also as for performance:
locally, with now storage-engine `wiredTiger` and the directories used on a `tmpfs`, the times range from 40s to 60s
in the github CI it seems like it is not having a `tmpfs`, so it is slower, but not by much and the time is between `60s and 120s` (mocha does not give more accurate times it seems once hitting a minute)

test/mocha-fixtures.js
vkarpov15: `global` does carry over in Mocha. That is one of many reasons why we advise people against using Jest: https://mongoosejs.com/docs/jest.html#globalsetup-and-globalteardown . Overwriting env var is more of an antipattern than overwriting global IMO, although I think it would be better to just export these.

PR #12266, opened 2022-08-11
fix(populate): handle specifying recursive populate as a string with discriminators

vkarpov15:
**Summary**

A bug that I ran into earlier today. If you `populate()` on a discriminator model and set the `populate` option to a string as follows:

```javascript
return Log.find({}).
              populate({
                path: 'activity',
                populate: 'postedBy'
              }).
              sort({ seq: -1 });
```

You'll get the following error:

```
TypeError: Cannot create property 'strictPopulate' on string 'postedBy'
```

This fixes that case

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -4704,7 +4704,7 @@ function populate(model, docs, options, callback) {
  */
 
 function _execPopulateQuery(mod, match, select, assignmentOpts, callback) {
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #12268, opened 2022-08-12
Add UUID Schema Type (BSON Buffer SubType 4)

### LINKED ISSUE #3208: Make UUID a formal Schema.Type
Hi,

I was wondering.. why is UUID not supported as a formal Schema Type? Mongo has support for storing it as BinData(), which seems to require about 50% of the space as compared to storing them as Strings.

The mongo shell has a built in function UUID() for the conversion... It seems like this would be a nice addition to mongoose.

hasezoey:
**Summary**

This PR tries to add Schema Type UUIDv1 (and later UUIDv4), but is a Draft (WIP), because i think i need some help with this, because i am not familiar enough with mongoose

continuation of #3242 (which was closed by the author without comment)
fixes #3208

---

Current Blockages:
- first UUIDv1 test fails because it times-out at `doc.save`, and i can figure out why `doc.save` times-out at that point

Questions / Notes:
- i am referencing [BSON Types: Binary Data](https://www.mongodb.com/docs/manual/reference/bson-types/#binary-data) for sub-types, but does anyone know what the difference between `3`(UUID (old)) and `4`(UUID) is?
- should maybe just one UUID type be added, with options to optionally set which version, because the bson types (and seemingly mongodb) support all UUID versions?

Notes on the state of the added code:
- JSDOC comments are not updated yet (have been commented-out / marked with `//`)
- "old" code has been commented-out but not removed yet
- UUIDv4 has not been added yet
- more tests would be needed


Uzlopak:
May I suggest something regarding UUID. I would actually prefer if you would implement a generic UUID-type. mongo is not differentiating imho between UUID v1 or v4. What we could do then, would be to assign a uuid generator to the default method of the schematype.So if somebody wants for some reason to use uuid v5 or the maybe more interesting v6, than dev can override the generator.

hasezoey:
@Uzlopak thats basically what i had also suggested in `Questions / Notes:` section, so i would also be in favor of that

though i would like some implementation details, like should the version be required / defaulted to something like 4 or default to none, or also no version support at all, and if version support, what option name to set the version, like `uuidVersion: 4`?

Uzlopak:
I apologize @hasezoey 

I also wanted to implement at one time uuid for mongoose, so i also had once worked on that branch. So it is nice to be on the same page ;).

Regarding subtype 3 and 4 is nothing uuid itself specific, but how mongodb was storing uuids internally. See:
https://studio3t.com/knowledge-base/articles/mongodb-best-practices-uuid-data/#binary-subtypes-0x03-and-0x04

The thing is, that nodejs has builtin uuid v4. But probably v6 would be better regarding indexing, but v6 is not even implemented in the uuid package. 

I expect actually no good by setting v4 as default. It could be that actually v6 is better regarding indexing (never benchmarked). But because we would decide that we ues v4 because it is builtin in node, I expect actually bike shedding. So instead to actively  decide one of the uuid-versions by the implementers, they assume that v4 is probably the best for a mongo collectoin because mongoose people had decided to use it so they know more than we do.

hasezoey:
> The thing is, that nodejs has builtin uuid v4. But probably v6 would be better regarding indexing, but v6 is not even implemented in the uuid package.

thanks for clarifying

> I expect actually no good by setting v4 as default. It could be that actually v6 is better regarding indexing (never benchmarked). But because we would decide that we ues v4 because it is builtin in node, I expect actually bike shedding. So instead to actively decide one of the uuid-versions by the implementers, they assume that v4 is probably the best for a mongo collectoin because mongoose people had decided to use it so they know more than we do.

i take from this the following:
- dont set a default and just allow any uuid by default (if no specific version is set)
- try to optionally support version 1-6 (ie validation, i dont know about generation)

hasezoey:
with 4c50837c6574c39485b88272394bea7f8797284c i have renamed it to `SchemaUUID`, but have not yet added extra functionality for specific versions, because for that i would like to have the tests working for now, but i still cant figure out why "doc.save" hangs

hasezoey:
i think i have now resolved the errors i was encountering in the tests, it came down to using the default connection instead of the one created

hasezoey:
update to the current state of the PR:
- the schema type is working, both saving & retrieving
- the mongodb type in the database is correct (checked with MongoDB Compass), i would like to add a test for that, but not sure how to do this
- the tests pass
- the jsdoc exists and is updated to the UUID type
- extra options (like setting specific UUID version) are not implemented yet

Uzlopak:
regarding the unit test. I think you can use directly the mongo driver and check if findOne has a raw options. Then you would get the bson Buffer. bson should be always the same structure, so you can theoretically slice the range of the field and its value and check if it has the right data.

lib/schema/uuid.js
```diff
@@ -0,0 +1,300 @@
...
+  this._cast = caster;
+
+  return this._cast;
+};
+
+/*!
+ * ignore
+ */
+
+SchemaUUID._checkRequired = v => UUID_FORMAT.test(v);
```
vkarpov15, line 183: Not necessary to do `UUID_FORMAT.test(v)`, just do `v != null`. This implementation will lead to Mongoose checking against UUID_FORMAT twice.

lib/schema/uuid.js
```diff
@@ -0,0 +1,300 @@
...
+  $all: handleArray,
+  $gt: handleSingle,
+  $gte: handleSingle,
+  $in: handleArray,
+  $lt: handleSingle,
+  $lte: handleSingle,
+  $ne: handleSingle,
+  $nin: handleArray,
+  $options: handleSingle,
+  $regex: handleSingle
```
vkarpov15, line 273: UUID shouldn't support `$regex` or `$options` because it looks like UUID is stored as a buffer in MongoDB, and MongoDB doesn't let you do `$regex` on buffers

test/schema.uuid.test.js
```diff
@@ -0,0 +1,56 @@
...
+  afterEach(() => util.stopRemainingOps(db));
+
+  it('basic functionality should work', async function() {
+    const doc = new Model({ x: '09190f70-3d30-11e5-8814-0f4df9a59c41' });
+    assert.ifError(doc.validateSync());
+    assert.ok(typeof doc.x === 'string');
+    assert.strictEqual(doc.x, '09190f70-3d30-11e5-8814-0f4df9a59c41');
+    await doc.save();
+
+    const query = Model.findOne({ x: '09190f70-3d30-11e5-8814-0f4df9a59c41' });
```
vkarpov15, line 39: I'd like to see a test using `$lte` and `$nin`

vkarpov15 (commented):
Couple of quick suggestions but otherwise looks good!

Any thoughts on keeping this in Mongoose itself versus putting uuid type in a plugin? Right now I'm leaning toward putting this in the Mongoose repo, because it doesn't add too much code weight or complexity, and because I'm starting to like adding more functionality to Mongoose vs asking people to install a bunch of plugins.

lib/schema/uuid.js
hasezoey: i had originally used Decimal128 as a template, but i did not find why `_checkRequired` as set when `checkRequired` was used, if it is the same as `_cast` and `cast`, then it should have not much effect on the current implementation because `checkRequired` does not call `_checkRequired`

lib/schema/uuid.js
hasezoey: thanks, i just copy-pasted that from the old PR, because i have no clue what this actually does or what works and what does not

test/schema.uuid.test.js
hasezoey: will try to add one

hasezoey:
> Any thoughts on keeping this in Mongoose itself versus putting uuid type in a plugin?

well i just did this because the old PR was just closed without comment (by the author) and it is a BSON type so i though mongoose itself should maybe support it

test/schema.uuid.test.js
hasezoey: i just looked into `$lte`, though i dont quite understand how that is supposed to work with UUID, because from what i can tell there is not "less than" or "higher than" in UUID (at least in the generic implementation and not having a specific version)

also mongodb does not list anything that `$lte` (or `$lt`, `$gt`, `$gte`) would work on UUID, do they work on buffers?

lib/schema/uuid.js
hasezoey: resolved in 47a0d078b1a5bc77767068683043c2a389333eb4

lib/schema/uuid.js
hasezoey: resolved in c76d09784cc95b61f79554218712fa25e818cb43

test/schema.uuid.test.js
AbdelrahmanHafez: @hasezoey 
Assuming that we store UUIDs as strings in MongoDB, $lt/$gt in mongo generally works with strings by sorting alphabetically. Check this [mongo playground](https://mongoplayground.net/p/UjrfQpq9mqa), for example.

I fail to see how that can be useful for UUIDs, though. AFAIK, UUIDs don't have any guarantees for alphabetical sorting.

hasezoey: > Assuming that we store UUIDs as strings in MongoDB

they are not, we store them as BSON UUID Buffers (bin-type 4)

AbdelrahmanHafez: Interesting. I don't know what the difference from a developer experience would be like. Care to explain to me what some pros and cons of storing them as UUID buffers instead of strings are?

hasezoey: > Interesting. I don't know what the difference from a developer experience would be like. Care to explain to me what some pros and cons of storing them as UUID buffers instead of strings are?

i dont know, but mongodb supports this Buffer type and so is probably more efficient in storing that, and using "it raw" (with the mongo client instead of mongoose), it is kinda cumbersome to use it this way, but with mongoose we can provide a easy back-end by converting it to string (like the current implementation) and the users using mongoose dont have to worry about the back-end buffer

hasezoey:
> regarding the unit test. I think you can use directly the mongo driver and check if findOne has a raw option

added with a187b01d67262cefab29bd186e37c21f79788458

test/schema.uuid.test.js
Uzlopak: It is a decision made by mongodb that bson stores uuids as buffers. The same as ObjectIds are stored as buffers. Only issue is that uuid version 1 to 5 are not incrementing sequentially. But there is the proposal  of uuid v6, which should do increment sequentially. 

http://gh.peabody.io/uuidv6/

Thats why I like that we can provide a uuid instead of forcing the use of v1 or v4. So if somebody wants to use v6, it can be done.

hasezoey: > instead of forcing the use of v1 or v4.

we currently dont force a type, mongodb allows version 0-9 (read like a regex), and what i discussed earlier was custom validators on mongoose's side

hasezoey: tests for `$in` and `$nin` added in fd17d9bb26d5124c288098a24f0eaed647ffc09e

though i still dont know about `$lt` (and derivatives)

hasezoey:
Update to the current state of this PR:
- UUID is implemented and working
- there are currently no options to set a specific uuid version to be validated against (so all version 0-9 are allowed also to be mixed)
- there are no tests for `$lt` (and derivatives) because i dont quite know how to use this on buffers
- there are no tests for `$bits*` operators
- UUID is not added to the types yet because i have no clue where i need to add it everywhere (and auto-inference to work with string)

test/schema.uuid.test.js
vkarpov15: Does MongoDB support `$gt`, `$lt`, etc. for buffers? There are times when I want to sort by `_id` in tests to ensure consistent sort order, regardless of what that sort order is. So it would be nice to support `$gt` and `$lt`, but not strictly necessary. If MongoDB doesn't support those with buffers, let's remove support for `$gt` and `$lt`.

I think it is fine to just support uuidv4 for now. If we want to support different uuid versions, that makes me more inclined to make uuid support a plugin.

lib/schema/uuid.js
```diff
@@ -0,0 +1,329 @@
...
+
+  if (typeof value === 'string') {
+    if (UUID_FORMAT.test(value)) {
+      return stringToBinary(value);
+    } else {
+      throw new CastError(SchemaUUID.schemaName, value, this.path);
+    }
+  }
+
+  if (Buffer.isBuffer(value)) {
```
vkarpov15, line 131: Should we check for the buffer's subtype here? Any reason why we might want to avoid casting buffers and `Binary` instances with different subtypes as UUID?

test/schema.uuid.test.js
Uzlopak: There is a plugin for mongoose 6 and uuid. But uuid is a native bson type. If it was some non-bson type then i would expect to be a plugin

hasezoey: > Does MongoDB support $gt, $lt, etc. for buffers?

i dont quite know, but it seems at least mongoose lists it as supported for normal buffers

> But uuid is a native bson type. If it was some non-bson type then i would expect to be a plugin

that is also my reasoning

> I think it is fine to just support uuidv4 for now

currently we dont check for a specific version, though i can add it as a optional option like `enforceVersion: 4` (or something like that)

lib/schema/uuid.js
hasezoey: at least at that point it cannot be checked, because those are nodejs native buffers which dont have a bson sub-type, but in the `instanceof Binary` check it can be done, should it be added as another nested check (`if () { if() { throw Error } }`) or directly in the same if (`if (here) {}`)? if nested, what error should be used?

test/schema.uuid.test.js
Uzlopak: We could just set the default field with crypto.randomUUID from node. If you want to have a different uuid version a implementer can just pass something like 

```js
const { v4, v1 } = require('uuid')

new Schema({ 
  x: { type: mongoose.Schema.Types.UUID, default: v4 },
  y: [{ type: mongoose.Schema.Types.UUID, default: v1 }]
})
```

hasezoey: > We could just set the default field with crypto.randomUUID from node

would it be a good idea to provide a default for a specific version? when i should add it, how is this done in mongoose? (it probably is not as simple as setting `.defaultOptions = { default: generateV4 }`, is `.auto` the correct function?)

vkarpov15 (commented):
Overall looks good. I'll do a little more review later, but I'd like to merge this into 6.7.

hasezoey:
updated to add tests as "todo"s that still need to be, also merged latest master

vkarpov15:
Merging into 6.7 :+1:

L14MJ4M3S0N:
@vkarpov15 Looks like this got merged without types. Created an issue for it.

PR #12271, opened 2022-08-13
Change Local Documentation serving to use `express`

hasezoey:
**Summary**

This PR changes local documentation serving from `serve-handler` to `express`.

This change is because:
- `serve-handler` did not have a update in a long time ([2 years](https://www.npmjs.com/package/serve-handler))
- `serve-handler` had non-flexible options for re-writing paths, which resulted in `index.html` (in project root, which is the landing-page) to not be served anymore locally

This change also has the effect that `http` module is not used directly anymore


Uzlopak:
Why not something super small. Like polka.

hasezoey:
> Why not something super small. Like polka.

because i only really know the one alternative to `serve-handler`, which is `express`, it is already relatively small, easy to use (low code) and "just works" (and is also still maintained / has updates)

PR #12272, opened 2022-08-13
Fix search url

BLooDBRothER:
## Bug description
Clicking the search button to search anything under the API section that contains the URL like **mongoosejs.com/docs/api/<name.html>** will redirect to the 404 page not found.
mongoose_1
mongoose_2

## Solution for the bug
In the [navbar js code](https://github.com/Automattic/mongoose/blob/master/docs/js/navbar-search.js) that has onClick function which redirects to **'search.html?q=' + encodeURIComponent(q); -> https://mongoosejs.com/docs/api/search.html?q=find** but it should redirect to **https://mongoosejs.com/docs/search.html?q=find**
mongoose_3
Changing ```location.href = 'search.html?q=' + encodeURIComponent(q);``` **to** ```location.href = '/docs/search.html?q=' + encodeURIComponent(q);``` will resolve the issue.

### DIFF
--- a/docs/js/navbar-search.js
+++ b/docs/js/navbar-search.js
@@ -1,6 +1,6 @@
 document.getElementById('search-button-nav').onclick = function() {
   var q = document.getElementById('search-input-nav').value;
-  window.location.href = 'search.html?q=' + encodeURIComponent(q);
+  window.location.href = '/docs/search.html?q=' + encodeURIComponent(q);
 };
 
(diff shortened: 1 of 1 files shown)


Uzlopak:
@hasezoey do you approve?

hasezoey (approved):
Seems like before this only works when being in `/docs/anywhere`, but not when deeper, like `docs/api/somewhere`

Uzlopak:
Well probably i would have set the baseurl but this should solve it atleast for now

PR #12275, opened 2022-08-13
Fix: Fixed typos in the repository

Muhammed-Rajab:
# Summary

 - Changed all 'atomically' to 'automatically'

I've been using mongoose and found out that 'automatically' was spelled incorrectly throughout the codebase.

# Explanation

- Changed every occurrence of  'atomically' to 'automatically'


Uzlopak (changes requested):
This PR does not only wrongfully renamed occurences of atomically to automatically but also changes single quotes to double quotes .

Atomically is correct and is derived from atomicity https://en.wikipedia.org/wiki/Atomicity_(database_systems)

PR #12280, opened 2022-08-15
refactor(model): allow optionally passing indexes to createIndexes and cleanIndexes

### LINKED ISSUE #12250: mongoose.syncIndexes() creates indexes even if they exist
### Prerequisites

### Mongoose version

6.5.2

### Node.js version

16.x

### MongoDB server version

4.x

### Description

mongoose.syncIndexes(...) currently sends a `createIndex` command to the database even if an index already exists.
It should only create indexes for indexes that do not exist on the MongoDB server.

### 12247.js
```js
'use strict';
const sinon = require('sinon');
const mongoose = require('./');
const { Schema } = mongoose;
const assert = require('node:assert');


run().catch(console.error);

async function run() {
  // Arrange

    name: { type: String }
  }, { autoIndex: false });

  userSchema.index({ name: 1 });

  const createIndexSpy = sinon.spy(User.collection, 'createIndex');
  const listIndexesSpy = sinon.spy(User.collection, 'listIndexes');

  // Act
  await mongoose.syncIndexes();
  assert.equal(createIndexSpy.callCount, 1);
  assert.equal(listIndexesSpy.callCount, 1);

  await mongoose.syncIndexes();

  // Assert
  assert.equal(listIndexesSpy.callCount, 2);
  assert.equal(createIndexSpy.callCount, 1);
}
```

### Expected Behavior

mongoose should only create indexes for indexes that do not exist on the MongoDB server.

AbdelrahmanHafez:
This refactors `diffIndexes` to smaller functions. 
Also, refactors `cleanIndexes` and `createIndexes` to accept receiving optional parameters for the indexes to drop/create. This reduces the number of `listIndexes` required for `syncIndexes` to work.

I'd like to merge this since it's in a functional state, then I'll work on fixing #12250 once I can dedicate more time to it.

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -6765,18 +6765,20 @@ describe('Model', function() {
 
         const User = db.model('User', userSchema);
 
+        await User.init();
+
         const createIndexSpy = sinon.spy(User.collection, 'createIndex');
         const listIndexesSpy = sinon.spy(User.collection, 'listIndexes');
 
         // Act
         await User.syncIndexes();
         assert.equal(createIndexSpy.callCount, 1);
-        assert.equal(listIndexesSpy.callCount, 1);
+        assert.equal(listIndexesSpy.callCount, 2);
 
         await User.syncIndexes();
 
         // Assert
-        assert.equal(listIndexesSpy.callCount, 2);
+        assert.equal(listIndexesSpy.callCount, 4);
         assert.equal(createIndexSpy.callCount, 1);
       });
     });
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -6713,7 +6713,6 @@ describe('Model', function() {
         Event.discriminator('BuyEvent', buyEventSchema);
 
         // Act
-
         const droppedByClickEvent = await ClickEvent.syncIndexes();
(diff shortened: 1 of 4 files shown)


test/model.test.js
```diff
@@ -6756,6 +6755,31 @@ describe('Model', function() {
         );
 
       });
+      xit('creates indexes only when they do not exist on the mongodb server (gh-12250)', async() => {
```
AbdelrahmanHafez, line 6758: This is marked as pending because this is not fixed yet, the PR was initially meant to fix the issue, but it's taking longer than anticipated. So for now let's stick to the refactoring, and I'll fix it later

vkarpov15 (commented):
This is looking great so far, no concerns. Let me know when you think this is ready :+1:

AbdelrahmanHafez:
@vkarpov15 
I think we can merge this as a refactor/perf improvement for now, and once I can dedicate more time to fix #12250, I'll open a new PR to fix it.

I remember one challenging part was understanding why `createIndexes` under `syncIndexes` is using `ensureIndexes`. I assume that since we already have the indexes `toCreate` from the previous step, we should utilize that explicitly, without having to rely on `ensureIndexes`.

What do you think?

vkarpov15 (approved):

PR #12290, opened 2022-08-17
fix: Update isDirectModified typing

gabrielDonnantuoni:
As the current code already implements it, just a type declaration fix:
 - Change arg type from `string` to `string | Array<string>`.


AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #12299, opened 2022-08-19
fix(types): sync single-generic-no-constraint "model" between "index.d.ts" and "connection.d.ts"

hasezoey:
**Summary**

This PR syncs the types for `.model` for `index.d.ts` and `connection.d.ts` for the single-no-constraint-generic overload


vkarpov15 (approved):
Thanks :+1: confirmed no impact on perf

PR #12310, opened 2022-08-22
language_override option mention

IslandRhythms:


AbdelrahmanHafez (changes requested):
Thanks, @IslandRhythms 

Can you please add TS types in `types\indexes.d.ts` and the tests to verify that it's working correctly?

vkarpov15 (approved):
It isn't necessary to add language_override to `indexes.d.ts` because MongoDB driver already defines `language_override` here:

https://github.com/mongodb/node-mongodb-native/blob/846365a5ef58acb53480fbd5a7e0e83cbfed2a3e/src/operations/indexes.ts#L96-L118

PR #12311, opened 2022-08-22
add support for $densify stage pipeline ¯\_(ツ)_/¯

IslandRhythms:


Uzlopak (approved):
(╯°□°)╯︵ ┻━┻

AbdelrahmanHafez:
I added a few JS/TS tests to verify that it's working.  ¯\_(ツ)_/¯

I also can see that you've initially defined `densify` as an `aggregateOption`, which I don't think is the correct place for it, it should be a pipeline stage instead.

Please let me know what you think. @IslandRhythms @Uzlopak @vkarpov15

vkarpov15 (approved):
Nice, thanks @IslandRhythms and @AbdelrahmanHafez :+1:

Issue #9056, opened 2020-05-26
Any plan to be compatible with Deno?


ghost:
This is a feature request.
Do you have any plan to become typescript compatible with Deno?

Using pika.dev version of Mongoose will work?
pkg.fyi/npm:mongoose

Thanks!

wenq1:
Any plan to make smarter judgements?

vkarpov15:
Haven't tried with either Deno or Pika. Will test it out and see if there's anything blocking us.

MeisterTea:
Did anybody managed to use mongoose with deno?

I get a `mongoose.connect` is not a function

vkarpov15:
Haven't tried it yet. Will try out Deno over the weekend and evaluate how much work it would be for us to support it.

vkarpov15:
I took a look at Deno and unfortunately we're currently blocked. Here's what's happening:

1) Do **not** use pika.dev's distribution of Mongoose. For some reason, they're pulling the [browser version of Mongoose](https://mongoosejs.com/docs/browser.html). So you won't be able to connect to MongoDB with Pika's version.

2) Deno currently bugs out with the [require_optional npm module](https://www.npmjs.com/package/require_optional), which is something that the MongoDB driver uses internally.

3) Even without `require_optional`, Deno currently doesn't have a polyfill for Node's crypto module: https://deno.land/std/node/README.md . That's going to be hard for the MongoDB driver to work around.

Here's the Deno script I was working with:

```
import { createRequire } from "https://deno.land/std/node/module.ts";

const require = createRequire(import.meta.url);

const mongoose = require('mongoose');

console.log(mongoose.version);
```

Anybody know of a good workaround for using crypto in deno?

luckycreationsindia:
Anyone tried god crypto module?

https://github.com/invisal/god_crypto

Also MongoDB is available on deno platform:

https://deno.land/x/mongo@v0.12.1

ebebbington:
@luckycreationsindia but deno_mongo is a deno plugin, thus very unstable

KotadiyaHiral:
> Anyone tried god crypto module?
> 
> https://github.com/invisal/god_crypto
> 
> Also MongoDB is available on deno platform:
> 
> https://deno.land/x/mongo@v0.12.1

Yes, MongoDB is available on deno platform, but there is very limited database features.

oceaann:
> I took a look at Deno and unfortunately we're currently blocked. Here's what's happening:

[...]

PR #12313, opened 2022-08-22
Use setPrototypeOf() instead of __proto__ to allow running on Deno

vkarpov15:
Re: #9056

**Summary**

With this change and `--unstable`, I've been able to successfully execute crud operations on Deno v1.24.3.

```javascript
import { createRequire } from "https://deno.land/std/node/module.ts";

const require = createRequire(import.meta.url);

run().catch(err => {
  console.error(err);
  process.exit(-1);
});

async function run() {

  const Test = mongoose.model('Test', mongoose.Schema({ name: String }));
  await Test.deleteMany({});
  await Test.create({ name: 'foo' });

  console.log(await Test.findOne());
  await mongoose.disconnect();
}
```

```
$ deno run --allow-read --allow-env --allow-net --unstable ./test.js 
6.5.2
model {
  "$__": InternalCache {
    activePaths: StateMachine { paths: { _id: "init", name: "init", __v: "init" }, states: { init: [Object] } },
    skipId: true
  },
  "$isNew": false,
  _doc: {
    _id: ObjectId {
      [Symbol(id)]: Buffer(12) [
         99,   3, 232, 161, 119,
         74, 216, 211,  11,  47,
        135,  14
      ]
    },
    name: "foo",
    __v: 0
  }
}
$ 
```

The `console.log()` output isn't great, looks like Deno doesn't support Node's `inspect()` output. But that's fixable. Turns out the big issue was that Deno doesn't support `__proto__`: https://github.com/denoland/deno/issues/4324.

**Examples**

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -81,7 +81,7 @@ function Connection(base) {
  * Inherit from EventEmitter
  */
 
-Connection.prototype.__proto__ = EventEmitter.prototype;
+Object.setPrototypeOf(Connection.prototype, EventEmitter.prototype);
 
 /**
  * Connection ready state
--- a/lib/drivers/node-mongodb-native/collection.js
+++ b/lib/drivers/node-mongodb-native/collection.js
@@ -34,7 +34,7 @@ function NativeCollection(name, conn, options) {
  * Inherit from abstract Collection.
  */
 
-NativeCollection.prototype.__proto__ = MongooseCollection.prototype;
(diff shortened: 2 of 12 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

This is great, can't wait for the day I can get rid of all the weird hacks for my build steps to run TS conveniently.

PR #12314, opened 2022-08-23
Revert "Use setPrototypeOf() instead of __proto__ to allow running on Deno"

AbdelrahmanHafez:
Reverts Automattic/mongoose#12313


hasezoey:
what was the reason that this change was reverted?

AbdelrahmanHafez:
We're merging this in 6.6 branch instead of master, see #12315

PR #12321, opened 2022-08-23
Make ArraySubdocuments apply `_id` defaults on init

### LINKED ISSUE #12264: _id field for subdocuments array items no longer being set on retrieval
### Prerequisites

### Mongoose version

6.5.2

### Node.js version

16.x

### MongoDB server version

5.x

### Description

In the version of Mongoose I was using before ( 6.0.11 ), if you had data inserted in Mongo that contained a subdocument array containing items without _id fields, Mongoose would fill in those fields with new _ids when the document was retrieved.

On the latest version of Mongoose, that seems to no longer be the case. I'm not sure if this is a deliberate change of behaviour or a regression. See the sample code below:

### Steps to Reproduce

```typescript
import * as mongoose from 'mongoose'

interface NestedChild {
    name: string
    _id: Types.ObjectId
}

interface Parent {
    nestedChild: Types.DocumentArray<NestedChild>
    name?: string
}

    nestedChild: [nestedChildSchema],
    name: String
}))

export async function run() {
    console.log('Running test')
    const dbConnection = mongoose.connections[0]
    const { insertedId } = await dbConnection.collection('parents').insertOne({
        nestedChild: [{ name: 'foo' }],
        name: 'Parent'
    })

    const parentRaw = await dbConnection.collection('parents').findOne({ _id: insertedId })
    const parentFound = await ParentModel.findOne({ _id: insertedId })

    // this is always undefined
    console.log('raw data', parentRaw?.nestedChild[0]._id)
    // on mongoose 6.0.11 this is an object id
    // on mongoose 6.5.2 this is undefined
    console.log('after retrieval', parentFound?.nestedChild[0]._id)
}

run()

```

### Expected Behavior

Unless this was deliberate, I would expect this behaviour to be unchanged and for mongoose to fill in the _id fields of the subdocuments upon retrieval

vkarpov15:
Fix #12264

**Summary**

@AbdelrahmanHafez you're right that https://github.com/Automattic/mongoose/commit/4c85c7008677db89ee9eb142785a09076a548a14 is the root cause of #12264. I think the `skipId` change is a bug, because that change makes it so that init-ing an array subdocument skips the `_id` default, but not any other defaults.

**Examples**

### DIFF
--- a/test/types.documentarray.test.js
+++ b/test/types.documentarray.test.js
@@ -748,4 +748,18 @@ describe('types.documentarray', function() {
     doc.arr.push(subdoc);
     await doc.validate();
   });
+
+  it('applies _id default (gh-12264)', function() {
+    mongoose.deleteModel(/Test/);
+    const nestedArraySchema = Schema({
+      subDocArray: [{ name: String }]
+    });
+
+    const Model = db.model('Test', nestedArraySchema);
+    const doc = new Model().init({
+      subDocArray: [{ name: 'foo' }]
+    });
+
+    assert.ok(doc.subDocArray[0]._id instanceof mongoose.Types.ObjectId);
+  });
 });
--- a/lib/schema/documentarray.js
+++ b/lib/schema/documentarray.js
@@ -392,7 +392,7 @@ DocumentArrayPath.prototype.getDefault = function(scope, init, options) {
 };
 
 const _toObjectOptions = Object.freeze({ transform: false, virtuals: false });
-const initDocumentOptions = Object.freeze({ skipId: true, willInit: true });
+const initDocumentOptions = Object.freeze({ skipId: false, willInit: true });
 
 /**
  * Casts contents


test/types.documentarray.test.js
```diff
@@ -748,4 +748,18 @@ describe('types.documentarray', function() {
...
   });
+
+  it('applies _id default (gh-12264)', function() {
+    mongoose.deleteModel(/Test/);
+    const nestedArraySchema = Schema({
+      subDocArray: [{ name: String }]
+    });
+
+    const Model = db.model('Test', nestedArraySchema);
+    const doc = new Model().init({
```
AbdelrahmanHafez, line 759: Is `new Model().init(...)` the equivalent of creating a document in the DB by bypassing mongoose and then fetching it?

vkarpov15: `init()` is what Mongoose calls to set a document's properties when loading it from the database. So yeah.

PR #12323, opened 2022-08-24
Add support for AbortSignal on eachAsync()

### LINKED ISSUE #12173: Make eachAsync AbortSignal-compatible
### Prerequisites

### 🚀 Feature Proposal

Allow users to pass an `AbortSignal` to `QueryCursor.prototype.eachAsync()`. When the signal is aborted, Mongoose will close the cursor and halt without any errors.

### Motivation

Passing an `AbortSignal` to `eachAsync()` would make it easy to process a stream of documents using a timeout.

### Example

```js
const signal = AbortSignal.timeout(ms('30s'));

await Post.find()
  .cursor()
  .eachAsync(processPost, { parallel: 50, signal })
```

vkarpov15:
Fix #12173

**Summary**

Node has support for AbortController and AbortSignal now: https://www.nearform.com/blog/using-abortsignal-in-node-js/ . So we can reasonably easily add support for AbortSignal in a few places, like `eachAsync()`.

The only potential problem with this implementation is that it doesn't actually kill the underlying MongoDB cursor (see https://www.mongodb.com/docs/manual/reference/command/killCursors/ ). However, MongoDB does recommend not killing cursors directly from applications, so this omission is justifiable.

**Examples**

### DIFF
--- a/lib/helpers/cursor/eachAsync.js
+++ b/lib/helpers/cursor/eachAsync.js
@@ -16,6 +16,9 @@ const promiseOrCallback = require('../promiseOrCallback');
  * @param {Function} next the thunk to call to get the next document
  * @param {Function} fn
  * @param {Object} options
+ * @param {Number} [options.batchSize=null] if set, Mongoose will call `fn` with an array of at most `batchSize` documents, instead of a single document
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
Nice!
LGTM, thanks. 👍

PR #12332, opened 2022-08-25
Gh 12156 ignore cast errors option

IslandRhythms:


lib/model.js
```diff
@@ -3715,7 +3715,8 @@ Model.applyDefaults = function applyDefaults(doc) {
  * @api public
  */
 
-Model.castObject = function castObject(obj) {
+Model.castObject = function castObject(obj, options) {
+  const opts = options || {};
```
AbdelrahmanHafez, line 3719: Creating another variable that points at the same object makes it a bit confusing with no value added. 
```suggestion
  options = options || {};
```

AbdelrahmanHafez (changes requested):
Please add JSDoc, TS types, and TS tests.

test/model.test.js
```diff
@@ -8859,6 +8859,35 @@ describe('Model', function() {
...
+      }));
+
+      const obj = {
+        num: 'not a number',
+        nested: { num: '2' },
+        subdoc: { num: '3' },
+        docArr: [{ num: '4' }]
+      };
+      const ret = Test.castObject(obj, { ignoreCastErrors: true });
+      assert.ok(ret);
```
AbdelrahmanHafez, line 8889: Would be nice if we assert that castObject returns the castable paths, and ignores the invalid ones.
```suggestion
      const obj = {
        num: 'not a number',
        nested: { num: '2' },
        subdoc: 'not a subdoc',
        docArr: [{ num: '4' }]
      };
      const ret = Test.castObject(obj, { ignoreCastErrors: true });
      assert.deepStrictEqual(ret, { nested: { num: 2 }, docArr: [{ num: 4 }] });
```

AbdelrahmanHafez:
A few minor comments, but overall LGTM. Thanks 👍

types/models.d.ts
```diff
@@ -137,7 +137,7 @@ declare module 'mongoose' {
     baseModelName: string | undefined;
 
     /* Cast the given POJO to the model's schema */
-    castObject(obj: AnyObject): T;
+    castObject(obj: AnyObject, options: AnyObject): T;
```
AbdelrahmanHafez, line 140:
```suggestion
    castObject(obj: AnyObject, options?: { ignoreCastErrors?: boolean }): T;
```

AbdelrahmanHafez (approved):
LGTM, thanks 👍

lib/model.js
```diff
@@ -3710,12 +3710,15 @@ Model.applyDefaults = function applyDefaults(doc) {
  *     Test.castObject({ num: 'not a number' }); // Throws a ValidationError
  *
  * @param {Object} obj object or document to cast
+ * @param {Object} options options passed to castObject
+ * @param {Boolean} options.ignoreCastErrors If set to true, will not throw a ValidationError and only return values that were successfully casted.
```
AbdelrahmanHafez, line 3714:
```suggestion
 * @param {Boolean} options.ignoreCastErrors If set to `true` will not throw a ValidationError and only return values that were successfully cast.
```

PR #12336, opened 2022-08-26
test(schema.validation): fix tests to not exit early

### LINKED ISSUE #5891: Multiple number of callback-style test written incorrectly
bug

I've haven't looked through all the tests, but look at `schema.validation.test.js`. The `done` function is called synchronously and doesn't wait for any of the `asserts` to be done.

```
      it('array required', function(done) {
        var Loki = new Schema({
          likes: {type: Array, required: true}
        });

        Loki.path('likes').doValidate(null, function(err) {
          assert.ok(err instanceof ValidatorError);
        });

        Loki.path('likes').doValidate(undefined, function(err) {
          assert.ok(err instanceof ValidatorError);
        });

        Loki.path('likes').doValidate([], function(err) {
          assert.ok(err instanceof ValidatorError);
        });
        done();
      });
```

This causes just about anything to pass...

```
      it('array required', function(done) {
        var Loki = new Schema({
          likes: {type: Array, required: true}
        });

        Loki.path('likes').doValidate(['this test should throw an error'], function(err) {
          assert.ok(err instanceof ValidatorError);
        });
        done();
      });
```

I'll try to clean up some of these where I can, but I don't have a full understanding of the test suite to know where these issues may be.

hasezoey:
**Summary**

This PR changes some tests in `schema.validation.test` to not potentially exit early.

fixes #5891 (at least partially but all example cases given)

Note: the approach used in this PR is also used already later in the file, so i had gone for that

### DIFF
--- a/test/schema.validation.test.js
+++ b/test/schema.validation.test.js
@@ -106,6 +106,7 @@ describe('schema', function() {
     });
 
     it('string regexp', function(done) {
+      let remaining = 10;
       const Test = new Schema({
         simple: { type: String, match: /[a-z]/ }
       });
@@ -114,6 +115,7 @@ describe('schema', function() {
 
       Test.path('simple').doValidate('az', function(err) {
         assert.ifError(err);
+        --remaining || done();
       });
 
       Test.path('simple').match(/[0-9]/);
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez:
This is great, would be nice if we can make `doValidate()` return a promise if no callback is passed, that would simplify the process a lot, and we won't have to do `--remaining || done()`.

vkarpov15 (commented):
Thanks :+1: although I think what we should do long term is make `doValidate()` optionally return a promise and just use async/await. The `--remaining` pattern works well, but is a bit anachronistic.

PR #12337, opened 2022-08-26
Change Comments from `/*!` to `/**` to make possible jsdoc available

hasezoey:
**Summary**

This PR changes most `/*!` to `/**` to make jsdoc available for those that have descriptions / tags defined, also does:
- change some casing for `* ignore`
- change some `@ignore` to `ignore`
- change some comments from `/*!` to `//`
- add / change tags to be more in line with current implementations (for those that were `/*!` before)
- change all `@private` to `@api private`
- changes `api public` to `api private` for `const mongoose = module.exports = exports = new Mongoose({`, because this had no effect anyway (from what i could tell)
- `Model.prototype.$remove` is now public in the documentation
- `Model.db;` now properly appears as a static property in the documentation
- `ArraySubdocument.prototype.$parent` is now `api public` (because other aliases are)
- removes `@api` tags from jsdoc in `test/*` files

Things that were ignored:
- all that were just `ignore`
- all that said `module exports`
- all that said `inherits from` (assigning `.prototype` / `._proto_`)

### DIFF
--- a/lib/types/ArraySubdocument.js
+++ b/lib/types/ArraySubdocument.js
@@ -16,7 +16,7 @@ const documentArrayParent = require('../helpers/symbols').documentArrayParent;
  * @param {Object} obj js object returned from the db
  * @param {MongooseDocumentArray} parentArr the parent array of this document
  * @param {Boolean} skipId
- * @param {Object} Fields
+ * @param {Object} fields
  * @param {Number} index
  * @inherits Document
  * @api private
--- a/lib/types/ArraySubdocument.js
+++ b/lib/types/ArraySubdocument.js
@@ -10,12 +10,14 @@ const utils = require('../utils');
 
 const documentArrayParent = require('../helpers/symbols').documentArrayParent;
 
-/*!
+/**
  * A constructor.
  *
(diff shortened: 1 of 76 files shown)


lib/types/ArraySubdocument.js
```diff
@@ -10,12 +10,14 @@ const utils = require('../utils');
...
 const documentArrayParent = require('../helpers/symbols').documentArrayParent;
 
-/*!
+/**
  * A constructor.
  *
  * @param {Object} obj js object returned from the db
  * @param {MongooseDocumentArray} parentArr the parent array of this document
  * @param {Boolean} skipId
+ * @param {Object} Fields
```
vkarpov15, line 19: Nit: lowercase F

vkarpov15 (approved):
Thanks!

lib/types/ArraySubdocument.js
hasezoey, line 19:
```suggestion
 * @param {Object} fields
```

hasezoey:
fixed with proper merge

PR #12338, opened 2022-08-26
fix(cast): Update cast to allow correctly querying for nested discriminators

### LINKED ISSUE #12318: Find on nested discriminators is broken after upgrade from mongoose 5 to mongoose 6
### Prerequisites

### Mongoose version

6.5.2

### Node.js version

16.15.0

### MongoDB server version

4.4

### Description

We are currently migrating from mongoose 5 to mongoose 6, and it seems like nested discriminators are broken with mongoose 6. When we `find` by a nested discriminated field, the query will return all documents instead of matching ones. This has created quite a few bugs for us while testing our code with mongoose 6, where completely unrelated data is being returned from queries. The culprit seems to be the new `strictQuery` option, and setting it to `false` patches the issue. However, having to set this option for nested fields seems like a broken behavior to use.

### Steps to Reproduce

The following script shows the error. The find call should return only the second document. However, both are returned.
```js

const BaseDocumentSourceSchema = new mongoose.Schema({
  source: { type: String },
}, {
  discriminatorKey: 'type'
});

const InternalDocumentSchema = new mongoose.Schema({
  source: { type: String },
});

const ExternalDocumentSchema = new mongoose.Schema({
  source: { type: String },
  url: { type: String },
});

const DocumentSchema = new mongoose.Schema({
  source: BaseDocumentSourceSchema,
});
DocumentSchema.path('source').discriminator('internal', InternalDocumentSchema);
DocumentSchema.path('source').discriminator('external', ExternalDocumentSchema);

const DocumentModel = mongoose.model('Document', DocumentSchema);

(async () => {
  await mongoose.connect(`mongodb://localTestUser:HhPQ5rEUvbDgqrkE3c@localhost:27017/frankhr_test`);

  await mongoose.connection.db.dropDatabase();

  const doc1 = await DocumentModel.create({ source: { type: 'internal' } });
  const doc2 = await DocumentModel.create({ source: { type: 'external', url: 'url' } });

  console.log(doc1, doc2);

  // Expected result is document 2. However, it will return both documents.
  const results = await DocumentModel.find({ 'source.url': 'url' });
  console.log(results);

  // Query correctly returns document 2
  const strictResults = await DocumentModel.find({ 'source.url': 'url' }, undefined, { strictQuery: false });
  console.log(strictResults);

  await mongoose.disconnect();
})();

```

### Expected Behavior

Queries using nested discriminated fields should return the correct data.

ramos-ph:
**Summary**

This PR is an attempt to fix #12318. Due to the `strictQuery` changes from version 5 to 6 from Mongoose, `Model.find` doesn't work how it used to.

I tracked this issue down to `lib/cast.js`. It was apparently introduced in https://github.com/Automattic/mongoose/commit/d1d8492fe2ecb120a39af1a8ab0a2624b0bdb834.

Just some observations, I also tried using `schema.options.discriminatorType`, but it didn't work.
Instead I opted by checking the amount of discriminators in the schema.

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -260,6 +260,11 @@ module.exports = function cast(schema, obj, options, context) {
           continue;
         }
 
+        // allows finding nested discriminators (gh-12318)
+        if (schema.discriminators && Object.keys(schema.discriminators).length > 0) {
+          continue;
+        }
+
         const strict = 'strict' in options ? options.strict : schema.options.strict;
         const strictQuery = 'strictQuery' in options ?
           options.strictQuery :
@@ -273,7 +278,9 @@ module.exports = function cast(schema, obj, options, context) {
           }
           throw new StrictModeError(path, 'Path "' + path + '" is not in ' +
             'schema, strict mode is `true`, and upsert is `true`.');
-        } if (strictQuery === 'throw') {
+        }
+
+        if (strictQuery === 'throw') {
           throw new StrictModeError(path, 'Path "' + path + '" is not in ' +
             'schema and strictQuery is \'throw\'.');
         } else if (strictQuery) {
--- a/test/model.discriminator.querying.test.js
+++ b/test/model.discriminator.querying.test.js
@@ -315,6 +315,29 @@ describe('model', function() {
       it('discriminator model only finds documents of its type when fields selection set as empty object', function(done) {
         checkDiscriminatorModelsFindDocumentsOfItsType({}, done);
       });
+
+      describe('using nested discriminators with `strictQuery = true`', function() {
+        it('only finds the related documents (gh-12318)', async function() {
+          const BaseDocumentSchema = new BaseSchema({ source: { type: String } }, { discriminatorKey: 'type' });
+          const InternalDocumentSchema = new BaseSchema({ source: { type: String } });
+          const ExternalDocumentSchema = new BaseSchema({ source: { type: String }, url: { type: String } });
+          const DocumentSchema = new BaseSchema({ source: BaseDocumentSchema });
+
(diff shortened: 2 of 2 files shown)


lib/cast.js
```diff
@@ -260,6 +260,11 @@ module.exports = function cast(schema, obj, options, context) {
           continue;
         }
 
+        // allows finding nested discriminators (gh-12318)
+        if (schema.discriminators && Object.keys(schema.discriminators).length > 0) {
```
vkarpov15, line 264: As far as I can tell, this check basically tells Mongoose to ignore strict mode if there are discriminators. I don't think that's the problem we're trying to solve here.

vkarpov15 (changes requested):
I appreciate the test case, but I don't think the solution you proposed will work. I think the correct answer here is `mongoose.set('strictQuery', false)`

lib/cast.js
ramos-ph: I see, thanks for the feedback. I also tried putting this a little below, wrapping L283.
https://github.com/Automattic/mongoose/blob/6f8d35b40285a83c6646e2b39e3df601a7912c57/lib/cast.js#L282-L284
This also fixes the querying issue, but as far as I remember, it breaks a few tests. And it seemed wrong to put it there.

ramos-ph: I saw Fabian's reply on the mentioned issue. Should we do this on schema level or just find another way to solve it?

vkarpov15: Yeah I think the more correct approach is setting the `strictQuery` option. Ignoring strict mode for discriminators isn't something we're going to do. I'm going to close this PR, but thanks for your hard work on this :+1:

PR #12341, opened 2022-08-27
fix(document): allow calling `$assertPopulated()` with values to better support manual population

### LINKED ISSUE #12233: Typescript: Can't represent newly-constructed document as "populated"
### Prerequisites

### 🚀 Feature Proposal

related to #11758
When constructing a new document, there currently isn't a way (that I can find) to tell Mongoose that I want the document to be represented as "populated". This presents a problem when constructing a new one using related documents:
```typescript

interface NestedChild {
    name: string
    _id: Types.ObjectId
}

interface Parent {
    nestedChild: Types.ObjectId
    name?: string
}

    nestedChild: { type: Schema.Types.ObjectId, ref: 'NestedChild' },
    name: String
}))

const NestedChildModel = model<NestedChild>('NestedChild', nestedChildSchema)

async function run() {
    const newParent = new ParentModel({
        name: 'Parent'
    })

    // doesnt work, model instance requires assignment to object id
    newParent.nestedChild = new NestedChildModel({ name: 'NestedChild' })

    // can't use assertPopulated because the internal $populated field isn't set
    // this will throw
    const asserted = newParent.$assertPopulated<{nestedChild: NestedChild}>('nestedChild')

    // instead, sketchy no-op populate call to make type system happy
    const populated = await newParent.populate<{nestedChild: NestedChild}>('nestedChild')

    // now I can assign the instance
    populated.nestedChild = new NestedChildModel({ name: 'NestedChild' })
}

run()
```
 Basically the $assertPopulated method that was recently added doesn't help in this case, because the document isn't actually "populated" technically so the internal $populated field is false. 
 
 A workaround I found is to basically perform a no-op "populate" call to receive the new type that contains the populated field.

Should there be some kind of built-in way to represent this case without workarounds? Could `$assertPopulated` just proceed if the document is "new" and the field being asserted on hasn't been modified yet? (since it then has the "potential" to be populated)

### Motivation

I'm writing an API that creates new documents and returns them in populated form. Currently I don't have a way of constructing a populated document from scratch and having it be typed correctly.

### Example

The easiest way I see would just be to make `$assertPopulated` not throw when the document is new:
```typescript
import { Schema, model, connect, Types, MergeType } from 'mongoose'

interface NestedChild {
    name: string
    _id: Types.ObjectId
}

interface Parent {
    nestedChild: Types.ObjectId
    name?: string
}

    nestedChild: { type: Schema.Types.ObjectId, ref: 'NestedChild' },
    name: String
}))

const NestedChildModel = model<NestedChild>('NestedChild', nestedChildSchema)

async function run() {
    const newParent = new ParentModel({
        name: 'Parent'
    })

    // doesnt work, model instance requires assignment to object id
    newParent.nestedChild = new NestedChildModel({ name: 'NestedChild' })

    // this call succeeds because nothing has been set on the populated field yet, it's still possible for it to be "populated" by assignment later on
    const asserted = newParent.$assertPopulated<{nestedChild: NestedChild}>('nestedChild')

    // now I can assign the instance
    asserted.nestedChild = new NestedChildModel({ name: 'NestedChild' })
}

run()

```

vkarpov15:
Fix #12233

**Summary**

#12233 highlighted a gap in `$assertPopulated()`: there's no way to manually populate using `=`.

```ts
// TypeScript complains about incorrect type
newParent.nestedChild = new NestedChildModel({ name: 'NestedChild' })

newParent.$assertPopulated<{ nestedChild: NestedChildModel }>('nestedChild')
```

```ts
// Throws at runtime because `nestedChild` isn't populated yet
newParent.$assertPopulated<{ nestedChild: NestedChildModel }>('nestedChild')

newParent.nestedChild = new NestedChildModel({ name: 'NestedChild' })
```

This PR combines the two steps into 1 for TypeScript's benefit:

```ts
// Does `$assertPopulated()` and `$set()` in one operation
const doc = newParent.$assertPopulated<{nestedChild: NestedChild}>('nestedChild', {
  nestedChild: new NestedChildModel({ name: 'NestedChild' })
})
```

What do you think of this @ajwootto ?

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -4418,7 +4418,7 @@ Document.prototype.$populated = Document.prototype.populated;
  *     doc.$assertPopulated('other path'); // throws an error
  *
  *
- * @param {String|String[]} paths
+ * @param {String|String[]} path
  * @return {Document} this
  * @memberOf Document
  * @method $assertPopulated
@@ -4426,14 +4426,18 @@ Document.prototype.$populated = Document.prototype.populated;
  * @api public
  */
 
-Document.prototype.$assertPopulated = function $assertPopulated(paths) {
-  if (Array.isArray(paths)) {
-    paths.forEach(path => this.$assertPopulated(path));
+Document.prototype.$assertPopulated = function $assertPopulated(path, values) {
+  if (Array.isArray(path)) {
+    path.forEach(p => this.$assertPopulated(p, values));
     return this;
   }
 
-  if (!this.$populated(paths)) {
-    throw new MongooseError(`Expected path "${paths}" to be populated`);
+  if (arguments.length > 1) {
+    this.$set(values);
+  }
+
+  if (!this.$populated(path)) {
+    throw new MongooseError(`Expected path "${path}" to be populated`);
   }
 
   return this;
--- a/lib/document.js
+++ b/lib/document.js
@@ -4417,8 +4417,13 @@ Document.prototype.$populated = Document.prototype.populated;
  *     doc.$assertPopulated('author'); // does not throw
  *     doc.$assertPopulated('other path'); // throws an error
  *
+ *     // Manually populate and assert in one call. The following does
+ *     // `doc.$set({ likes })` before asserting.
+ *     doc.$assertPopulated('likes', { likes });
  *
- * @param {String|String[]} path
+ *
+ * @param {String|String[]} path path or array of paths to check. `$assertPopulated` throws if any of the given paths is not populated.
+ * @param {Object} [values] optional values to `$set()`. Convenient if you want to manually populate a path and assert that the path was populated in 1 call.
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks. 👍

lib/document.js
```diff
@@ -4418,22 +4418,26 @@ Document.prototype.$populated = Document.prototype.populated;
  *     doc.$assertPopulated('other path'); // throws an error
  *
  *
- * @param {String|String[]} paths
+ * @param {String|String[]} path
  * @return {Document} this
```
hasezoey, line 4422: maybe consider adding `values` to the jsdoc (and maybe also to the typescript types) to explain what `values` is expected to be / used for

PR #12343, opened 2022-08-28
test(common): change mongoose require to be explicit "index" file

hasezoey:
**Summary**

This PR changes the `test/common.js` import for `mongoose` to be a explicit file.

Before typescript's language server could not infer the types for mongoose, but now it has actual proper types, which should make writing (and following) tests easier by having intellisense / auto-completion available.

<details>
<summary>Screenshot showcasing problem</summary>

Before change:
before

After Change:
after

</details>

PS: weirdly this "typescript cannot infer the index file" only happens when trying to import the root `index.js`, from `index.js` itself importing `lib/` actually has proper intellisense

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -4,7 +4,7 @@
  * Module dependencies.
  */
 
-const mongoose = require('../');
+const mongoose = require('../index');
 const Collection = mongoose.Collection;
 const assert = require('assert');
 


AbdelrahmanHafez (approved):
This has always been bugging me. Thanks a lot!

PR #12345, opened 2022-08-28
Use npm pack to npm install Mongoose for TypeScript benchmarks

### LINKED ISSUE #12320: Make benchmarks use Mongoose in `package.json` and `npm pack`
### Prerequisites

### Issue

Our TypeScript benchmarks missed a perf regression (see #10349 later discussions) that popped up around 6.3.2. Because I suspect something strange is going on with using `paths` and importing Mongoose's `index.d.ts` from top level, as opposed to installing with npm: https://github.com/Automattic/mongoose/blob/8c0800af9c0deb9216ebf15ee0232dd9ebd347d2/benchmarks/typescript/simple/tsconfig.json#L6 .

We should be able to work around this in GitHub Workflows using `npm pack`

vkarpov15:
Fix #12320

**Summary**

I managed to make ts-benchmark pick up type regressions:

image

Haven't been able to get GitHub to run the tests though. Suspect it is because ts-benchmark hijacks `git config --global user.name`, which is something we should fix.

**Examples**


vkarpov15:
Yes this does provide us with automatic detection. In theory. I'll keep running TS benchmarks manually for now to make sure.

We did store historic stats, but not currently. That's something I want to add back.

PR #12348, opened 2022-08-29
Call _markModified in splice array method

josegl:
In the implementation of the other array methods that mutate the array,
the _markModified function is called, but in the splice method it
wasn't.

This leads to do not detect changes that happen in arrays when the splice method is used.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -16,18 +16,18 @@ You can refer to [this table](https://docs.mongodb.com/drivers/node/current/comp
 
 Below are the [semver](http://semver.org/) ranges representing which versions of mongoose are compatible with the listed versions of MongoDB server.
 
-| MongoDB Sever |            Mongoose            |
-| :-----------: | :----------------------------: |
-|     `6.x`     |            `^6.5.0`            |
-|     `5.x`     |            `^6.0.0`            |
-|    `4.4.x`    |      `^5.10.0 \| ^6.0.0`       |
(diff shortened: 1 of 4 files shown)


hasezoey (changes requested):
A test would be good so that in the future this does not silently break again

AbdelrahmanHafez (changes requested):
Please add a test to demonstrate the issue you're facing and prevent future regression.

josegl:
@hasezoey, @AbdelrahmanHafez How could I test this and where. It seems that there are a bunch of places where tests are placed, but I cannot find the most suitable place to put it. And to be honest I'm not very sure how could I test this. Could you please help me in that regard?

AbdelrahmanHafez:
@josegl 
You can try `test\types.array.test.js` to write a unit test that tests the array class directly, or it might actually be better to test how you'd run it in your apps in `test\document.test.js` by creating a schema, a model, and a document, mark the document as not modified, and then `splice`, and then assert that it's marked as modified.

vkarpov15 (approved):
Nice catch, thanks. I'll wait until you've added the tests to merge.

josegl:
I've added the test as you guys suggested.

hasezoey:
> I've added the test as you guys suggested.

looks good, though maybe run a `npm run lint --fix` and push that too

hasezoey (approved):
LGTM

Uzlopak (approved):
LGTM

PR #12350, opened 2022-08-29
style(cast): update "cast" JSDOC to include more options

hasezoey:
**Summary**

This PR adds some options for `cast.js`'s `cast` function, this should make types better (while in the js code, not in typescript types)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #12352, opened 2022-08-29
feat(types+schema): allow defining schema paths using mongoose.Types.* to work around TS type inference issues

### LINKED ISSUE #12205: InterSchemaType does not properly infer "required" field for Schema.Types.ObjectId type
### Prerequisites

### Mongoose version

^6.5.0

### Node.js version

16.13.1

### MongoDB server version

6.x

### Description

A schema field  with `{type: Schema.Types.ObjectId, required: true}` cause faulty inference of Schema type. But,  `{type: Schema.Types.ObjectId, required: false}` ,  `{type: String, required: true}` seem to work

### Steps to Reproduce

The following **works**:

```js
const campaignSchema = new Schema(
  {
    client: {
      type: Schema.Types.ObjectId,
      trim: true,
      ref: "User",
      required: false,
    },
  },
  { timestamps: true }
);

type ICampaign = InterSchemaType<ICampaignSchema>; // {client?: Types.ObjectId | undefined }
```

The following **doesn't work:**

```js
const campaignSchema = new Schema(
  {
    client: {
      type: Schema.Types.ObjectId, // works when the type is changed to "String"
      trim: true,
      ref: "User",
      required: true, // fails to interpret schema
    },
  },
  { timestamps: true }
);

type ICampaign = InterSchemaType<ICampaignSchema>; // DOES NOT return {client:  Types.ObjectId }
```
Screenshot-from-2022-08-03-09-31-43

### Expected Behavior

```js
type ICampaign = InterSchemaType<ICampaignSchema>;  // {client:  Types.ObjectId }
```

vkarpov15:
Fix #12205

**Summary**

I've spent quite some time debugging #12205 and haven't come up with quite a complete picture of why this is happening, but I got a decent enough workaround.

The issue comes down to the fact that, for some reason, `InferSchemaType<>` runs the `ObtainDocumentType` calculation twice: once with the schema definition, and once with the returned value from the first run. I suspect that's because of the [self-referential `DocType = ObtainDocumentType<DocType>` line here](https://github.com/Automattic/mongoose/blob/0af4cb815ec36dd50d18755f5238506ef9297a5a/types/index.d.ts#L175), but I haven't been able to figure out how to remove that. Any thoughts on this issue @mohammad0-0ahmad ?

Also had to add some fixes to `ResolvePathType<>` because, apparently, `Schema.Types.ObjectId extends typeof Schema.Types.ObjectId` is false :man_shrugging: .

The workaround I came up with is to allow defining ObjectId and Decimal128 paths using `mongoose.Types.ObjectId` in addition to `mongoose.Schema.Types.ObjectId` (and same for Decimal128). This is likely a decent DX improvement anyway, since the difference between `mongoose.Types.ObjectId` and `mongoose.Schema.Types.ObjectId` can be confusing.

**Examples**


mohammad0-0ahmad:
Yes, that seems to be correct, I've not checked the changes you made in this PR, but I've noticed that we can solve it as well by making this change "Infer doc type directly without using ```ObtainSchemaGeneric``` helper":
```diff
 -type InferSchemaType<SchemaType> = ObtainSchemaGeneric<SchemaType, 'DocType'>;
+ type InferSchemaType<SchemaType> = SchemaType extends Schema<any, any, any, any, any, any, any, infer T> ? T : unknown;
```
OR by making the following change which I mentioned before in https://github.com/Automattic/mongoose/pull/12168:
```diff
// inferschematype.d.ts
//line 171
-PathValueType extends Record<string, any> ? ObtainDocumentType<PathValueType, any, TypeKey> :
+PathValueType extends Record<string, any> ? IfEquals<PathValueType, Types.ObjectId, Types.ObjectId, ObtainDocumentType<PathValueType, any, TypeKey>>:
```
After using TS for a while, sometimes it feels that TS can have unexpected behavior, for instance in the first solution ```ObtainSchemaGeneric``` it is literally making the same as the provided solution but gives two different results.
I wrote that https://github.com/Automattic/mongoose/blob/0af4cb815ec36dd50d18755f5238506ef9297a5a/types/index.d.ts#L175 for obtaining doc type for one time instead of calling ```ObtainDocumentType``` multiple time in the code and for be able to show the schema paths when user hovers on schema itself.
I don't really know right now how TS utilities works in details, "I've not read TS source code yet 😄" But I think it might be helpful if we ask someone of TS team about this specific case, It might be helpful to understand the TS behavior more.
@vkarpov15 can you mention someone here of TS team that could be able to help here quickly, maybe without needing to dig into TS source code, because I don't think we can find the answer in its docs. "I might be wrong"

vkarpov15:
I don't know of anyone on the TS team that might help with this.

I spent a long time wrangling with the `type InferSchemaType<SchemaType> = SchemaType extends Schema<any, any, any, any, any, any, any, infer T> ? T : unknown;` solution, it fixes this particular issue but introduces another issue with the #12030 test [here](https://github.com/Automattic/mongoose/blob/caa9a69ebb4b7abb37dc69b798fd8db29439e39b/test/types/schema.test.ts#L635-L647). And I wasn't able to figure out any workaround for that issue.

IslandRhythms (approved):
LGTM

mohammad0-0ahmad:
> I don't know of anyone on the TS team that might help with this.
> 
> I spent a long time wrangling with the `type InferSchemaType<SchemaType> = SchemaType extends Schema<any, any, any, any, any, any, any, infer T> ? T : unknown;` solution, it fixes this particular issue but introduces another issue with the #12030 test [here](https://github.com/Automattic/mongoose/blob/caa9a69ebb4b7abb37dc69b798fd8db29439e39b/test/types/schema.test.ts#L635-L647). And I wasn't able to figure out any workaround for that issue.

I will check that at the weekend.

test/types/schema.test.ts
```diff
@@ -640,6 +644,28 @@ function gh12030() {
     ]
   });
 
+  type A = ResolvePathType<[
```
mohammad0-0ahmad, line 647: these tests are a bit wrong, I think they should be:
```TS
  const a = [
    {
      username: { type: String }
    }
  ];

  expectType<{
    username?: string
  }[]>({} as ResolvePathType<typeof a>);

  const b = {
    users: [
      {
        username: { type: String }
      }
    ]
  };

  expectType<{
    users: {
      username?: string
    }[];
  }>({} as ResolvePathType<typeof b>);

  const Schema1 = new Schema({
    users: [
      {
        username: { type: String }
      }
    ]
  });
```

test/types/schema.test.ts
```diff
@@ -737,3 +763,36 @@ function pluginOptions() {
   schema.plugin<any, SomePluginOptions>(pluginFunction2, { option2: 0 });
   expectError(schema.plugin<any, SomePluginOptions>(pluginFunction2, {})); // should error because "option2" is not optional
 }
+
+function gh12205() {
+  const campaignSchema = new Schema(
+    {
+      client: {
+        type: new Types.ObjectId(),
```
mohammad0-0ahmad, line 771: Is ObjectId instance an acceptable value to pass ?

HiiiiD:
Guys, I think I figured out the solution: `String` doesn't extend `typeof String`, but it extends `String`, same thing for `ObjectId` and for `Decimal128`
And this is for every class in TS

HiiiiD:
BTW the fix that I proposed also solves #12373

HiiiiD:
Any news? Do I have to create another PR to address this?

mohammad0-0ahmad:
> Guys, I think I figured out the solution: `String` doesn't extend `typeof String`, but it extends `String`, same thing for `ObjectId` and for `Decimal128` And this is for every class in TS @vkarpov15 @mohammad0-0ahmad

I think that is incorrect, I think the entire issue as  @vkarpov15 mentioned that InferSchemaType reinfer the already inferred type. I might misunderstand you, but check this example first: 
https://www.typescriptlang.org/play?ssl=5&ssc=26&pln=1&pc=1#code/MYewdgzgLgBAHjAvDAylATgSzAcwNwBQBUAngA4CmMAKhdADzUB8y1MFcUFYAJhDKUogAZqgzYcMAPwD0AVyoAuGMICGAGwgVCxclTjokNOlHqCKI+E0JA

mohammad0-0ahmad:
@vkarpov15 
Would you like to revert changes inside inferschematype.d.ts and test to refactor ```InferSchemaType``` helper with the following?

```diff
- type InferSchemaType<SchemaType> = ObtainSchemaGeneric<SchemaType, 'DocType'>;
+ type InferSchemaType<TSchema> = TSchema extends Schema<any, any, any, any, any, any, string, infer DocType> ? DocType : unknown;
```

types/inferschematype.d.ts
```diff
@@ -31,14 +31,14 @@ declare module 'mongoose' {
...
-   * @param {SchemaType} SchemaType A generic of schema type instance.
+   * @param {Schema} TSchema `typeof` a schema instance.
    * @example
    * const userSchema = new Schema({userName:String});
    * type UserType = InferSchemaType<typeof userSchema>;
    * // result
    * type UserType = {userName?: string}
    */
-  type InferSchemaType<SchemaType> = ObtainSchemaGeneric<SchemaType, 'DocType'>;
+  type InferSchemaType<TSchema> = ObtainSchemaGeneric<TSchema, 'DocType'>;
```
mohammad0-0ahmad, line 41: Would you like to test this ?
```ts
   type InferSchemaType<TSchema> = TSchema extends Schema<any, any, any, any, any, any, string, infer DocType> ? DocType : unknown;
   ```

File: test/types/schema.test.ts [hunk 4]
```diff
@@ -737,3 +763,36 @@ function pluginOptions() {
...
+  const Campaign = model('Campaign', campaignSchema);
+  const doc = new Campaign();
+  expectType<Types.ObjectId>(doc.client);
+
+  type ICampaign = InferSchemaType<typeof campaignSchema>;
+  expectType<{ client: Types.ObjectId }>({} as ICampaign);
+
+  expectType<'type'>({} as ObtainSchemaGeneric<typeof campaignSchema, 'TPathTypeKey'>);
+
+  type A = ObtainDocumentType<{ client: { type: Schema.Types.ObjectId, required: true } }>;
```
mohammad0-0ahmad, line 787: You use ```Schema.Types.ObjectId``` as type, I think this is incorrect.
Same thing for the following tests.

PR #12354, opened 2022-08-30
documentation: add info about connect logged in

AntonyOnScript:
**Summary**
I was studying about how mongoose works and i felt it was missing an important information for a getting started page, how to make a "connect" logged in. I think it can make learn path easier for beginners.


hasezoey:
note that user and password can also be provided via options `user` and `pass`, see [`.createConnection`](https://mongoosejs.com/docs/api.html#mongoose_Mongoose-createConnection) (some for `.connect`)

vkarpov15 (approved):
Thanks :+1:

PR #12367, opened 2022-08-31
FIX: Added fix if mongodb://localhost fails

drsimplegraffiti:
**Summary**

**Examples**

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Once connected, the `open` event is fired on the `Connection` instance. If you'r
 
 **Note:** _If the local connection fails then try using 127.0.0.1 instead of localhost. Sometimes issues may arise when the local hostname has been changed._
 
-If `localhost` use
+If `localhost` fails, use
 
 ```js
 await mongoose.connect('mongodb://0.0.0.0:27017/my_database');
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Mongoose is a [MongoDB](https://www.mongodb.org/) object modeling tool designed to work in an asynchronous environment. Mongoose supports both promises and callbacks.
 
-[![Slack Status](https://img.shields.io/badge/slack-mongoosejsteam-34D058.svg?logo=slack )](https://mongoosejsteam.slack.com)
+[![Slack Status](https://img.shields.io/badge/slack-mongoosejsteam-34D058.svg?logo=slack)](https://mongoosejsteam.slack.com)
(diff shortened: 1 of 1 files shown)


Uzlopak:
That is imho quite radical approach when the issue is actually, that you should not use localhost but the IPv4 127.0.0.1 on node18

Uzlopak:
I dont agree with this change. Also above your remark it says to use 127.0.0.1

README.md
```diff
@@ -2,7 +2,7 @@
 
 Mongoose is a [MongoDB](https://www.mongodb.org/) object modeling tool designed to work in an asynchronous environment. Mongoose supports both promises and callbacks.
 
-[![Slack Status](https://img.shields.io/badge/slack-mongoosejsteam-34D058.svg?logo=slack )](https://mongoosejsteam.slack.com)
+[![Slack Status](https://img.shields.io/badge/slack-mongoosejsteam-34D058.svg?logo=slack)](https://mongoosejsteam.slack.com)
```
vkarpov15, line 5: Why all these unrelated changes?

README.md
```diff
@@ -91,22 +97,22 @@ const BlogPost = new Schema({
   author: ObjectId,
   title: String,
   body: String,
-  date: Date
+  date: Date,
```
vkarpov15, line 100: :-1:

README.md
```diff
@@ -77,6 +77,12 @@ Once connected, the `open` event is fired on the `Connection` instance. If you'r
 
 **Note:** _If the local connection fails then try using 127.0.0.1 instead of localhost. Sometimes issues may arise when the local hostname has been changed._
 
+If `localhost` fails, use
```
vkarpov15, line 80: The remark above says to use `127.0.0.1`, which is better IMO.

README.md
```diff
@@ -77,6 +77,12 @@ Once connected, the `open` event is fired on the `Connection` instance. If you'r
...
+
+```js
+await mongoose.connect('mongodb://0.0.0.0:27017/my_database');
```
vkarpov15, line 83: I'm pleasantly surprised that connecting to `0.0.0.0` actually works. However, I don't want to rely on that or recommend that because [0.0.0.0 is "used to designate an invalid, unknown or non-applicable target"](https://en.wikipedia.org/wiki/0.0.0.0).

vkarpov15 (commented):
I appreciate the PR, but I don't think the suggestion is correct or helpful. Plus all the unnecessary and unrelated style changes.

Issue #8703, opened 2020-03-21
How to run error handling middleware on `validateSync()`?


vkarpov15:
Right now we skip async validators on `validateSync()`, but the way we determine if a validator is async is based on the number of arguments. Since error handling middleware must take 3 arguments, `validateSync()` will never run error handling middleware #4885 

Re: #4885

hasezoey:
is this still a issue because `doValidate` uses `isAsyncFunction` which checks the function prototype instead of argument length?

Uzlopak:
I think a unit test, which proves that the error handling middleware is called should prove that this issue is done.

hasezoey:
i have opened up #12371 which address *some* of this issue

i have noticed that `validateSync` does **not** trigger any hooks, even if they are normal sync functions, so i made #12372 as a draft for now - if it is wanted that this function to not execute hooks, please close the PR (stating that)

PR #12371, opened 2022-09-01
Add Test that `doValidateSync` skips async functions and ignores promises

hasezoey:
**Summary**

This PR adds a test that tests that `doValidateSync` ignores `async function()` and ignores returned promises

re #8703

### DIFF
--- a/test/schema.validation.test.js
+++ b/test/schema.validation.test.js
@@ -564,6 +564,44 @@ describe('schema', function() {
           done();
         }, { a: 'b' });
       });
+
+      it('doValidateSync should ignore async function and script waiting for promises (gh-4885)', function(done) {
+        let asyncCalled = false;
+        let normalCalled = false;
+        let promiseCalled = false;
+        let promiseCompleted = false;
+
+        const schema = new Schema({
+          prop: {
+            type: Boolean,
+            validate: [
+              {
+                validator: async() => { asyncCalled = true; }
+              },
+              {
+                validator: () => { normalCalled = true; }
+              },
+              {
+                validator: () => {
+                  promiseCalled = true;
+                  return new Promise((res) => setTimeout(() => {
+                    promiseCompleted = true;
+                    return res();
+                  }, 1000));
+                }
+              }
+            ]
+          }
+        });
+
+        schema.path('prop').doValidateSync(true);
+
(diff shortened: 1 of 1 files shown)


test/schema.validation.test.js
```diff
@@ -564,6 +564,44 @@ describe('schema', function() {
...
+              {
+                validator: async() => { asyncCalled = true; }
+              },
+              {
+                validator: () => { normalCalled = true; }
+              },
+              {
+                validator: () => {
+                  promiseCalled = true;
+                  return new Promise((res) => setTimeout(() => {
```
AbdelrahmanHafez, line 587: I think we can convert this whole promise to `Promise.resolve();`, no need for a setTimeout.
Also, if async validators don't work, we should throw an error if we receive a validator that returns a promise, similar to https://github.com/Automattic/mongoose/pull/9176.
What do you think? @hasezoey

hasezoey: > Also, if async validators don't work, we should throw an error if we receive a validator that returns a promise, similar to

from what i know async validators are allowed and in the documentation and also executed in `validate`, but just not in `validateSync` (stated in the documentation)

AbdelrahmanHafez: Interesting, I didn't know that.
So they would work with `await Document#validate()`, but not with `Document#validateSync()` nor implicitly when running `await user.save();`, is that about right? @hasezoey

hasezoey: > I think we can convert this whole promise to Promise.resolve();, no need for a setTimeout.

i made it this way because promises may be executed earlier when no "blocking" is found, and i wanted to check that promise function get called but "not finish executing" and `Promise.resolve` would "maybe immediately" set this value

hasezoey: > Interesting, I didn't know that. So they would work with `await Document#validate()`, but not with `Document#validateSync()` nor implicitly when running `await user.save();`, is that about right? @hasezoey

from what i know, in both `validate` and `save` async validators are run (and promises "awaited") (from what i can tell, `save` calls `validate`)

AbdelrahmanHafez: Makes sense. Thanks @hasezoey

vkarpov15 (approved):
Thanks :+1: this is a helpful test to have

PR #12372, opened 2022-09-01
Add test for `validateSync` to test ignore async functions but still run error hooks

### LINKED ISSUE #8703: How to run error handling middleware on `validateSync()`?

hasezoey:
**Summary**

This PR adds tests that test `validateSync` to execute sync functions and error hooks but ignore `async function()` and ignores returned Promises

fixes #8703

Blocked currently because `validateSync` does not seem to execute any hooks

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -12262,6 +12262,74 @@ describe('document', function() {
     const test2 = {};
     assert.strictEqual(test2.constructor.polluted, undefined);
     assert.strictEqual(Object.polluted, undefined);
+
+  });
+
+  it('validateSync executes error hook and ignores async function (gh-4885)', function(done) {
+    let asyncCalled = false;
+    let errorHookCalled = false;
+    let normalHookCalled = false;
+
+    const schema = new Schema({
+      prop: Boolean
+    });
+
+    schema.post('validate', async function() { asyncCalled = true; });
+    schema.post('validate', () => { normalHookCalled = true; throw new Error('test error for error hook'); });
+    schema.post('validate', function(err, doc, next) {
+      errorHookCalled = true;
+      next();
+    });
+
+    const model = db.model('test', schema);
+
+    const doc = new model({ prop: true });
+    doc.validateSync();
+
+    // schema.path('prop').doValidateSync(true);
+
+    assert.strictEqual(normalHookCalled, true);
+    assert.strictEqual(asyncCalled, false);
+    assert.strictEqual(errorHookCalled, true);
+    done();
+  });
+
+  it('validateSync executes error hook and ignores returned Promises (gh-4885)', function(done) {
+    let asyncCalled = false;
+    let asyncCompleted = false;
+    let errorHookCalled = false;
+    let normalHookCalled = false;
+
+    const schema = new Schema({
+      prop: Boolean
+    });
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez (commented):
If async validators don't work, I think we should throw an error instead of silently ignoring them. Similar to how we handle `transform` option, see https://github.com/Automattic/mongoose/pull/9176.
What do you think @hasezoey

hasezoey:
> If async validators don't work, I think we should throw an error instead of silently ignoring them. Similar to how we handle `transform` option, see #9176.

i dont know what the convention for this is here, though currently in #12371 `doValidateSync` silently ignores the async function (as stated in the documentation)

this PR here (#12372) focuses on hooks, which do not seem to be executed in `validateSync` currently

AbdelrahmanHafez:
Interesting, I didn't know that.

So async validators would run with `Document#validate()`, but not with `Document#validateSync()`, nor implicitly with `await user.save();`. Is that correct? @hasezoey

AbdelrahmanHafez:
Thanks @hasezoey 
I just realized they are two different PRs, thought there was an issue with GitHub and my comments didn't go through, apparently I just duplicated them. 😄

hasezoey:
rebased on latest master to resolve conflicts

vkarpov15:
@hasezoey is this PR still valid? It's still in draft state and has been for a while, so I wanted to check if we can close this PR.

hasezoey:
> @hasezoey is this PR still valid? It's still in draft state and has been for a while, so I wanted to check if we can close this PR.

i have been waiting for feedback about the pr and the referenced issue, simply said this PR adds a test for a issue, but the functionality is not quite what this test is meant to test and there has not been feedback about what to do (is this wanted behavior or should be changed)

if i remember correctly, the current issue is about that `validateSync` states that it does not call any `async` hooks but should still call (sync) error handling hooks, which it seemingly does not

i am not sure what the exact problem is / was because the logs have expired and i am currently not on my dev machine

vkarpov15:
Ok, I'll take a closer look tomorrow

hasezoey:
rebased onto latest master to resolve conflicts

current issue is that both tests fail because none of the 3 `validate` hooks are called

vkarpov15:
So I think the confusion is due to the fact that I mixed up async hooks and async validators when I wrote up #8703. Right now it is expected behavior that `validateSync()` skips all middleware, and I'm not convinced it is worthwhile to make `validateSync()` execute sync middleware.

Right now we only have one function that supports sync middleware, `init()`, and that code feels a bit hacky. There's also the concern about `validateSync()` firing `pre('validate')` and `post('validate')` hooks as opposed to `pre('validateSync')` and `post('validateSync')`, which may be a better API.

Thanks for the work you put in on this PR @hasezoey , and I'm very sorry it took me so long to take a more thorough look. However, I'm going to close this PR and #8703 because I'm not convinced that executing validation hooks on `validateSync()` is worth doing ATM.

PR #12391, opened 2022-09-04
fix(model): make `applyHooks()` and `applyMethods()` handle case where custom method is set to Mongoose implementation

### LINKED ISSUE #12254: Using Schema.loadClass with a class which inherits from Document will hang up any find operations.

vkarpov15:
Fix #12254

**Summary**

#12254 was a tricky issue, but comes down to the fact that `applyHooks()` will apply hooks twice if you have a custom method named `init`, which leads to infinite recursion. I added a workaround in `applyHooks()`.

I also made `applyMethods()` ignore methods that are strictly equal to the Mongoose implementation. That is, calling `schema.method('save', Model.prototype.save)` will be a no-op now. Which is better behavior IMO, that better supports the case of calling `loadClass()` with a class that extends Document, since in that case you don't want `save()` to be a custom method anyway.

**Examples**

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -8864,9 +8864,9 @@ describe('Model', function() {
   it('works if passing class that extends Document to `loadClass()` (gh-12254)', async function() {
     const DownloadJobSchema = new mongoose.Schema({ test: String });
 
-    class B extends mongoose.Document {}
-
-    B.prototype.foo = 'bar';
+    class B extends mongoose.Document {
+      get foo() { return 'bar'; }
+    }
 
     DownloadJobSchema.loadClass(B);
 
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -8875,6 +8875,7 @@ describe('Model', function() {
     const { _id } = await Test.create({ test: 'value' });
     const doc = await Test.findById(_id);
     assert.ok(doc);
+    assert.equal(doc.foo, 'bar');
     assert.equal(doc.test, 'value');
   });
 });
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -8860,6 +8860,23 @@ describe('Model', function() {
       assert.equal(error.errors['docArr.0.num'].name, 'CastError');
     });
   });
+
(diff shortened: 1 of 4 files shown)


test/model.test.js
```diff
@@ -8860,6 +8860,23 @@ describe('Model', function() {
...
+    B.prototype.foo = 'bar';
+
+    DownloadJobSchema.loadClass(B);
+
+    const Test = db.model('Test', DownloadJobSchema);
+
+    const { _id } = await Test.create({ test: 'value' });
+    const doc = await Test.findById(_id);
+    assert.ok(doc);
+    assert.equal(doc.test, 'value');
```
AbdelrahmanHafez, line 8878: Perhaps we need to assert that the value from the loaded class reflects on the documents.

```suggestion
    assert.equal(doc.foo, 'bar');
    assert.equal(doc.test, 'value');
```

AbdelrahmanHafez (commented):
One minor comment, otherwise LGTM. Thanks. 👍

Issue #11725, opened 2022-04-29
Old documentation, delete them?


Uzlopak:
We should think about deleting all the documentation regarding mongoose up to version 4. It makes no sense to keep that old stuff. Also nobody requests currently Support for those versions. And if... Nobody would provide support for them. End of life.

Uzlopak:
After examining the code, removing it from the master could result that the are also removed in the gh-pages branch.

Uzlopak:
But how can remove the old documentation without deleting it in the gh-pages branch? :/

hasezoey:
currently the documentation code does not reference anything before 4.x, but that documentation still exists in `gh-pages` and i dont see a way to remove that without modifying `gh-pages` directly

Uzlopak:
Ah yes. back in the days i moved all old docs into gh-pages.

ThePaulin:
There is no need to remove old documentation since it doesn't cause any issues to people looking for documentation for newer releases

Uzlopak:
I removed like a long time ago old documentation by adding them to gh-pages. So this is done already. :)

hasezoey:
> I removed like a long time ago old documentation by adding them to gh-pages

not really removed, just moved (and is currently harder to access by not being in the version selector, but still accessable)

Uzlopak:
Its accessable yes. But not promoted by the dropdown. If you want you can add them. I actually would remove the 4.x docs from the dropdown to avoid the impression that it is still supported. But anyhow Nobody reported anyway a bug for 4.x so we could actually remove 4.x from the dropdown or add all the old docs. Or leave it as is.

PR #12396, opened 2022-09-05
docs(scripts/website): remove setting "latest38x" which is not used anywhere

hasezoey:
**Summary**

This PR removes setting `pkg.latest38x` which is not used anywhere anymore

re #11725


AbdelrahmanHafez (approved):
LGTM, thanks 👍

vkarpov15 (approved):
Thanks :+1:

PR #12405, opened 2022-09-08
Update connection.js

iso-1986:
The previous test fails, even if "client" is instance of "mongodb.MongoClient".

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -1472,7 +1472,7 @@ Connection.prototype.getClient = function getClient() {
  */
 
 Connection.prototype.setClient = function setClient(client) {
-  if (!(client instanceof mongodb.MongoClient)) {
+  if ((client instanceof mongodb.MongoClient)!==true) {
     throw new MongooseError('Must call `setClient()` with an instance of MongoClient');
   }
   if (this.readyState !== STATES.disconnected) {


Uzlopak:
Is it already october?

Uzlopak:
This change makes no sense. There is no failing test.

And to be honest I think you have some malicious intent. Maybe getting trying to get some workflow secrets or so. Or you are trying to get experience on how to pass low quality PRs into popular repositories for hacktober.

Anyhow: Without a better explaination why this change is useful i am going close this PR now.

iso-1986:
Thanks for the reply @Uzlopak.
I understand why you think what you think. It's Ok.

Seriously, I tried many times to make this work and I understand that this change make no sense, but Javascript in those cases, sometimes make no sense...
I don't know why, but even if I give an instance of MongoClient to the setClient function, it throw the error : 'Must be call with an instance of MongoClient'. And this happened on node 18.7.0, mongodb node driver 4.8.1 with mongodb 6.0.1.

By the way I don't know much about code, git and Javascript, but since I change this line everything works for me.

Uzlopak:
Do you use jest or typescript?
Open please an issue if it is a serious problem

iso-1986:
No, I don't use jest or type script.

I don't know if this is a serious problem, because as I said I don't know much about code... but I'll do as you said, and we gona see where it lead ;).

Thanks for your time. Peace.

PR #12418, opened 2022-09-11
fix(query): use correct Query constructor when cloning query

### LINKED ISSUE #12365: Clone Query ignores hooks
### Prerequisites

### Mongoose version

6.5.0

### Node.js version

16.15.0

### MongoDB server version

5.0

### Description

The [clone function](https://mongoosejs.com/docs/api.html#query_Query-clone) from a Query doesn't trigger the hooks defined in its schema.

It may leads to an error as it doesn't rexecute the exact same query.

### Steps to Reproduce

```node
const schema = new Schema().pre("find", () => console.log("pre hook function"))
const myModel = model("Test", schema)
const query = myModel.find()

await query.clone() // Do nothing
await query // Log "pre hook function"
```

The only workaround is a "deep copy":
```node
const schema = new Schema().pre("find", () => console.log("pre hook function"))
const myModel = model("Test", schema)
const initialQuery = myModel.find()

// Magic is here
const queryCloned = initialQuery.model.find(initialQuery.getFilter())
await queryCloned() // Log "pre hook function"
```

### Expected Behavior

Execute the schema hooks on a cloned Query.

vkarpov15:
Fix #12365

**Summary**

#12365 pointed out that we're not cloning hooks correctly. That's because `this.constructor` doesn't seem to pick up the correct Query class - it picks up the default `Query`, not the model-specific `Query`.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -5069,8 +5069,9 @@ Model.compile = function compile(name, schema, collectionName, connection, base)
   model.Query = function() {
     Query.apply(this, arguments);
   };
-  model.Query.prototype = Object.create(Query.prototype);
+  Object.setPrototypeOf(model.Query.prototype, Query.prototype);
   model.Query.base = Query.base;
+  model.Query.prototype.constructor = Query;
   applyQueryMiddleware(model.Query, model);
   applyQueryMethods(model, schema.query);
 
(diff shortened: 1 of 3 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #12419, opened 2022-09-11
Improve handling of defaults when using `pull()`

vkarpov15:
**Summary**

The problem here is that `pull()` adds defaults to the query filter that is sent to MongoDB on `save()`, so it won't match documents that don't have the default set. This behavior is problematic because this may cause `save()` to start not pulling documents if the user adds a new property with a default to the schema.

We fix this by explicitly removing any fields that are set to `default` in the document.

```javascript
const catschema = new Schema({
  name: String,
  colors: [{
    _id: false,
    hex: { type: String, default: '#ffffff' },
    name: String
  }]
});
const Cat = db.model('Test', catschema);

const cat = new Cat({});
cat.init({
  name: 'Garfield',
  colors: [{ name: 'Orange' }]
});

cat.colors.pull({ name: 'Orange' });
// Sends `{ $pull: { colors: { $or: [{ name: 'Orange', hex: '#ffffff' }] } } }
await cat.save();
```

**Examples**

### DIFF
--- a/lib/types/array/methods/index.js
+++ b/lib/types/array/methods/index.js
@@ -603,7 +603,11 @@ const methods = {
 
     if (values[0] instanceof ArraySubdocument) {
       this._registerAtomic('$pullDocs', values.map(function(v) {
-        return v.$__getValue('_id') || v;
+        const _id = v.$__getValue('_id');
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #12437, opened 2022-09-16
Set $inc default value in case field has not been specified on the document

### LINKED ISSUE #12435: $inc on non-existing field results in NaN
### Prerequisites

### Mongoose version

6.6.1

### Node.js version

16.15.0

### MongoDB server version

4.2.1

### Description

According to [MongoDB document on `$inc`](https://www.mongodb.com/docs/manual/reference/operator/update/inc/),

> If the field does not exist, [$inc](https://www.mongodb.com/docs/manual/reference/operator/update/inc/#mongodb-update-up.-inc) creates the field and sets the field to the specified value.

In Mongoose, $inc operations on non-existing fields result in correct data in the database, however, the field in Mongoose document will become `NaN` because `currentValue` is undefined.

https://github.com/Automattic/mongoose/blob/586c86f8356906c3541c9f6b83d5b1eeedda4361/lib/document.js#L1738-L1740

### Steps to Reproduce

```javascript
    count: Number,
    // no default value, not required
});
const Model = mongoose.model('IncTest', schema);
await Model.collection.drop();
const doc = new Model({
    // optional 'count' field no set
});
await doc.save();
doc.$inc('count', 1);
await doc.save();

// doc.count is expected to be 1.
// But instead, it is NaN
```

### Expected Behavior

_No response_

lpizzinidev:
fixes #12435

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -11864,6 +11864,18 @@ describe('document', function() {
     assert.strictEqual(doc.sub.propertyB, 'foo');
     assert.strictEqual(called, 1);
   });
+
+  it('If the field does not exist, $inc should create it and set is value to the specified one (gh-12435)', async function() {
+    const schema = new mongoose.Schema({
+      count: Number
+    });
+    const Model = db.model('IncTest', schema);
+    const doc = new Model({ });
+    await doc.save();
+    doc.$inc('count', 1);
+    await doc.save();
+    assert.strictEqual(doc.count, 1);
+  });
 });
 
 describe('Check if instance function that is supplied in schema option is availabe', function() {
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

test/document.test.js
```diff
@@ -11864,6 +11864,18 @@ describe('document', function() {
...
+  it('If the field does not exist, $inc should create it and set is value to the specified one (gh-12435)', async function() {
+    const schema = new mongoose.Schema({
+      count: Number
+    });
+    const Model = db.model('IncTest', schema);
+    const doc = new Model({ });
+    await doc.save();
+    doc.$inc('count', 1);
+    await doc.save();
+    assert.strictEqual(doc.count, 1);
```
Uzlopak, line 11877: can you please also read the document from the database and check if it is actually also stored properly on the database?

lpizzinidev: Sure 👍

Uzlopak (approved):

Issue #12066, opened 2022-07-07
dynamic populate with refPath in sub-document


iammola:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.4.3

### Node.js version

18.4.0

### MongoDB server version

5.0

### Description

This is most likely a bug, but I'm not sure.

I have a path in my model that is a sub-document, and I want to dynamically populate another path in that sub-document using the `refPath` option.

```ts
import mongoose from "mongoose";

const subSchema = new mongoose.Schema({
  by: {
    type: mongoose.Schema.Types.ObjectId,
    refPath: "user_type", // path to model
    required: true,
  },
  user_type: {
    type: String,
    required: true,
    // the `by` field should be populated on one of these models
    enum: ["Parent", "Child"],
  },
});

const RecordSchema = new mongoose.Schema({
  title: {
    type: String,
    required: true,
  },
  created: {
    type: subSchema,
    required: true
  },
});

const ParentSchema = new mongoose.Schema({
  name: {
    type: String,
    required: true,
  },
  job: { type: String }
});

const ChildSchema = new mongoose.Schema({
  name: {
    type: String,
    required: true
  },
  siblings: [mongoose.Schema.Types.ObjectId]
});
```

I then query the database for a document and populate the `created.by` field, but the path isn't replaced with `null` or the foreign document.

```ts
import mongoose from "mongoose";

async function run() {
  const [parent, child] = await Promise.all([
    ParentModel.create({ name:"Lionel Messi", job: "Footballer" }),
    ChildModel.create({ name: "Thiago Messi" })
  ]);

  await RecordModel.create([
    { title: "Heart of a Lio", created: { by: parent._id, user_type: "Parent" } },
    { title: "My day with Hulk", created: { by: child._id, user_type: "Child" } },
    { title: "How to be the GOAT", created: { by: parent._id, user_type: "Parent" } },
  ]);

  return await RecordModel.find({}).populate("created.by").lean().exec();
}

const RecordModel = mongoose.model('Record', RecordSchema);
const ParentModel = mongoose.model('Parent', ParentSchema);
const ChildModel = mongoose.model('Child', ChildSchema);

run().then(console.log);

// [{
//   title: "Heart of a Lio",
//   created: {
//     by: "ParentID",
//     user_type: "Parent"
//   }
// }, { ... }, { ... }]
```

### Steps to Reproduce

[...]

PR #12440, opened 2022-09-17
Fix dynamic virtual populate on sub-documents

### LINKED ISSUE #12363: Dynamic virtual populate not working on subdocument with separate schema
### Prerequisites

### Mongoose version

6.5.4

### Node.js version

16.15.0

### MongoDB server version

4.2.1

### Description

Seems to be related to #8277 and #8742

When using dynamic ref on child schema `{ ref: (doc) => doc.refType }`, the passed in `doc` is the parent document, not the subdocument, causing dynamic ref path not correctly returned.

I tried both `virtuals` option in schema constructor options and schema `virtual()` method, tried both `ref` and `refPath` in virtual options, none works.

### Steps to Reproduce

```typescript

const nestedSchema = new mongoose.Schema({
    targetType: String,
    targetId: mongoose.Schema.Types.ObjectId,
}, {
    virtuals: {
        target: {
            options: {
                ref: (doc: any) => {
                    // doc should be the sub document
                    // but instead, it is the parent document
                    console.debug(doc);
                    return doc.targetType;
                },
                localField: 'targetId',
                foreignField: '_id',
                justOne: true,
            }
        }
    }
});

    nested: {
        type: nestedSchema
    },
});

const NestedDynamicVirtualPopulateTest = mongoose.model('NestedDynamicVirtualPopulateTest', parentSchema);

await NestedDynamicVirtualPopulateTest.collection.drop();
const target = new NestedDynamicVirtualPopulateTest({});
await target.save();

const parent = new NestedDynamicVirtualPopulateTest({
    nested: {
        targetType: (target.constructor as typeof NestedDynamicVirtualPopulateTest).modelName,
        targetId: target._id,
    }
});
await parent.save();
await parent.populate('nested.target');
// @ts-ignore
expect(parent.nested.target).to.exist;
```

### Expected Behavior

Since the virtual is defined on the child schema, the dynamic ref path should be relative to the subdocument, not the parent document.

Hybrid-Force:
Fix #12363

**Summary**
Apply dynamic ref getter of virtual population on sub-documents instead of the parent document.

Mongoose supports dynamic virtual populate via `ref` and `refPath` options. However, the `ref` function is called on the parent document, and `refPath` is relative to the parent document. It does not make much sense for dynamic virtual populates defined on nested schemas, because the child schema has no knowledge of the parent schema.

After this patch, `ref` is called on sub-document, and `refPath` is relative to sub-document instead of the parent document.

### DIFF
--- a/lib/helpers/populate/getModelsMapForPopulate.js
+++ b/lib/helpers/populate/getModelsMapForPopulate.js
@@ -355,114 +355,118 @@ function _virtualPopulate(model, docs, options, _virtualRes) {
   const virtual = _virtualRes.virtual;
 
   for (const doc of docs) {
-    let modelNames = null;
-    const data = {};
+    // Use the correct target doc/sub-doc for dynamic ref on nested schema. See gh-12363
+    let subDocs = _virtualRes.nestedSchemaPath ?
+      utils.getValue(_virtualRes.nestedSchemaPath, doc) : doc;
+    subDocs = Array.isArray(subDocs) ? subDocs : subDocs ? [subDocs] : [];
+    for (const subDoc of subDocs) {
+      let modelNames = null;
+      const data = {};
 
-    // localField and foreignField
-    let localField;
-    const virtualPrefix = _virtualRes.nestedSchemaPath ?
-      _virtualRes.nestedSchemaPath + '.' : '';
-    if (typeof virtual.options.localField === 'function') {
-      localField = virtualPrefix + virtual.options.localField.call(doc, doc);
-    } else if (Array.isArray(virtual.options.localField)) {
-      localField = virtual.options.localField.map(field => virtualPrefix + field);
-    } else {
-      localField = virtualPrefix + virtual.options.localField;
-    }
-    data.count = virtual.options.count;
-
-    if (virtual.options.skip != null && !options.hasOwnProperty('skip')) {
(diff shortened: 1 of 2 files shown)


vkarpov15 (changes requested):
I took a closer look and the distinction that `refPath` is relative to the top-level document rather than the subdocument is intentional. Paraphrasing from [this comment](https://github.com/Automattic/mongoose/issues/12066#issuecomment-1186070762): if `refPath` is relative to the subdocument, it is impossible to make `refPath` reference a field outside of the subdocument (like a top-level field) without introducing new syntax. Whereas with `ref()`, you can always call `this.ownerDocument()` to get the top-level document.

I do like the fix that `ref()` on virtuals is now called on the subdocument rather than the top-level document. But `refPath` should still be relative to top-level document. Is that possible?

Also, it looks like these changes only affect virtual populate. So it looks like `refPath` is still relative to top-level document for conventional populate.

Hybrid-Force:
> I do like the fix that `ref()` on virtuals is now called on the subdocument rather than the top-level document. But `refPath` should still be relative to top-level document. Is that possible?

Sure, I will update the patch.

vkarpov15:
Closing this in favor of #14652, which fixes OP's problem with fewer changes. We will still need to release this in the next major release though, because #14652 is backwards breaking

PR #12441, opened 2022-09-17
Fixed InferSchemaType nullable arrays type inference

### LINKED ISSUE #12420: `mongoose.InferSchemaType` does not correctly infer nullable arrays
### Prerequisites

### Mongoose version

6.5.5

### Node.js version

16.15.1

### MongoDB server version

5.0.12

### Description

When working with Mongoose schemas involving array fields that are nullable, `mongoose.InferSchemaType` does not seem to correctly infer the schema type. The notion of optionality does not seem to be reflected.

### Steps to Reproduce

Let's suppose we create a Mongoose schema with 2 fields of 2 different types, one being an array of strings, the other an array of objects, both optional.

```ts
  {
    comments: { type: [String], default: () => undefined, required: false },
    reference: {
      type: [
        {
          type: { type: String, required: false },
          value: { type: String, required: false },
        },
      ],
      default: () => undefined,
      required: false,
    },
  },
  {
    collection: 'tests',
    timestamps: false,
  },
);

export type TestType = mongoose.InferSchemaType<typeof TestSchema>;
```

In `TestType`, the notion of optionality of our fields seems to be lost:
inferred_type

The problem is the same if we replace `default: () => undefined` by `default: () => null`.

If we just try `default: undefined` or `default: null`, another problem appears:
inferred_type_2

Perhaps there is a misunderstanding/error on my side in defining the schema. If this is the case, I would be happy to receive your advices. Otherwise, it must be a bug? 🐛

### Expected Behavior

```
{
    comments?: string[];
    reference?: {
        type?: string;
        value?: string;
    }[];
}
```

lpizzinidev:
fixes #12420

### DIFF
--- a/types/inferschematype.d.ts
+++ b/types/inferschematype.d.ts
@@ -69,7 +69,7 @@ type IsPathRequired<P, TypeKey extends TypeKeyBaseType> =
   P extends { required: true | [true, string | undefined] } | ArrayConstructor | any[]
     ? true
     : P extends (Record<TypeKey, ArrayConstructor | any[]>)
-      ? P extends { default: undefined }
+      ? P extends { default: undefined } | { required: false }
         ? false
         : true
       : P extends (Record<TypeKey, any>)
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -741,3 +741,32 @@ function pluginOptions() {
   schema.plugin<any, SomePluginOptions>(pluginFunction2, { option2: 0 });
   expectError(schema.plugin<any, SomePluginOptions>(pluginFunction2, {})); // should error because "option2" is not optional
 }
+
+function gh12420() {
+  const TestSchema = new Schema(
+    {
+      comments: { type: [String], required: false },
+      reference: {
+        type: [
+          {
+            type: { type: String, required: false },
+            value: { type: String, required: false }
+          }
+        ],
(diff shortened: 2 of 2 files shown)


hasezoey (approved):
LGTM

types/inferschematype.d.ts
```diff
@@ -69,7 +69,7 @@ type IsPathRequired<P, TypeKey extends TypeKeyBaseType> =
   P extends { required: true | [true, string | undefined] } | ArrayConstructor | any[]
     ? true
     : P extends (Record<TypeKey, ArrayConstructor | any[]>)
-      ? P extends { default: undefined }
+      ? P extends { default: undefined } | { required: false }
```
vkarpov15, line 72: See my comments here: https://github.com/Automattic/mongoose/issues/12420#issuecomment-1258236649 . We shouldn't rely on `required: false`, because `required` is false by default anyway. We instead need to check if `default` is a function that returns undefined.

vkarpov15 (changes requested):

types/inferschematype.d.ts
lpizzinidev: 👍 I will work on this

vkarpov15:
I'm going to close this since the PR isn't going in the right direction, and now has merge conflicts.

PR #12447, opened 2022-09-20
Add clarification about overwrite flag in model.js

Tzvika-m:
Add calrification for using the `overwrite` flag in model.findByIdAndUpdate and model.findOneAndUpdate

**Summary**

The way mongoose changes the update behaviour (adds `$set` to the update command) and how to prevent it is not intirely clear.
I believe this note could help save some time and effort for future users.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2558,7 +2558,7 @@ Model.$where = function $where() {
  *     Model.findOneAndUpdate(query, { $set: { name: 'jason bourne' }}, options, callback)
  *
  * This helps prevent accidentally overwriting your document with `{ name: 'jason bourne' }`.
- * To prevent this behaviour, see the `overwrite` flag
+ * To prevent this behaviour, see the `overwrite` option
  *
  * #### Note:
  *
--- a/lib/model.js
+++ b/lib/model.js
@@ -2691,7 +2691,7 @@ function _decorateUpdateWithVersionKey(update, options, versionKey) {
  *     Model.findByIdAndUpdate(id, { $set: { name: 'jason bourne' }}, options, callback)
  *
  * This helps prevent accidentally overwriting your document with `{ name: 'jason bourne' }`.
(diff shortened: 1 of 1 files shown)


Uzlopak (changes requested):
For me this change has no additional value as it just says see overwrite. It should have more than "see overwrite".

Tzvika-m:
@Uzlopak What did you have in mind?

lib/model.js
```diff
@@ -2558,6 +2558,7 @@ Model.$where = function $where() {
  *     Model.findOneAndUpdate(query, { $set: { name: 'jason bourne' }}, options, callback)
  *
  * This helps prevent accidentally overwriting your document with `{ name: 'jason bourne' }`.
+ * To prevent this behaviour, see the `overwrite` flag
```
hasezoey, line 2561:
```suggestion
 * To prevent this behaviour, see the `overwrite` option
```

lib/model.js
```diff
@@ -2690,6 +2691,7 @@ function _decorateUpdateWithVersionKey(update, options, versionKey) {
  *     Model.findByIdAndUpdate(id, { $set: { name: 'jason bourne' }}, options, callback)
  *
  * This helps prevent accidentally overwriting your document with `{ name: 'jason bourne' }`.
+ * To prevent this behaviour, see the `overwrite` flag
```
hasezoey, line 2694:
```suggestion
 * To prevent this behaviour, see the `overwrite` option
```

hasezoey (approved):
LGTM, i think it helps having a note to the option to prevent the automatic extra behavior form happening (if the option exists) in the example itself

some small things to suggest to keep the naming more consistent

Uzlopak (approved):
I guess I follow in this case @hasezoey

Tzvika-m:
Great, thanks
I've committed your suggestions @hasezoey

Issue #12431, opened 2022-09-14
Issues with Decimal128 type


HiiiiD:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.6.1

### Node.js version

18.2.0

### MongoDB server version

5.x

### Description

Adding a field to a Mongoose Schema with the type `Schema.Types.Decimal128` breaks all the typings of `model`, including the ones of the other fields of the Mongoose Schema.
Make sure to use a version of Typescript < 4.8.x due to the fact that otherwise the build breaks.

### Steps to Reproduce

For reference this is the `tsconfig.json` that I used to compile
```json
{
    "compilerOptions": {
        "declaration": true,
        "outDir": "./build",
        "module": "CommonJS"
    },
    "exclude": [
        "node_modules",
        "build"
    ]
}
```

```ts
import { model, Schema } from 'mongoose';

const testSchema = new Schema({
    testDate: { type: Date }
});

const TestModel = model('Test', testSchema);
```
As expected `TestModel` has the following type
```ts
declare const TestModel: import("mongoose").Model<{
    testDate?: Date;
}, {}, {}, {}, Schema<any, import("mongoose").Model<any, any, any, any, any>, {}, {}, {}, {}, "type", {
    testDate?: Date;
}>>;
```

If we change our schema just a little bit, introducing `Schema.Types.Decimal128` it breaks every type.

```ts
import { model, Schema } from 'mongoose';

const testSchema = new Schema({
    testDate: { type: Date },
    testDecimal: { type: Schema.Types.Decimal128 },
});

const TestModel = model('Test', testSchema);
```

This results in
```ts
declare const TestModel: import("mongoose").Model<{
    testDate?: {
        [Symbol.toPrimitive]?: {};
        toString: {};
        valueOf: {};
        toLocaleString: {};
        toJSON?: {};
        toDateString?: {};
        toTimeString?: {};
        toLocaleDateString?: {};
        toLocaleTimeString?: {};
        getTime?: {};
        getFullYear?: {};
        getUTCFullYear?: {};
        getMonth?: {};
        getUTCMonth?: {};
        getDate?: {};
        getUTCDate?: {};
        getDay?: {};
        getUTCDay?: {};
        getHours?: {};
        getUTCHours?: {};
        getMinutes?: {};
        getUTCMinutes?: {};
        getSeconds?: {};
        getUTCSeconds?: {};
[...]

PR #12450, opened 2022-09-20
Fix types resolved from InferSchemaType

HiiiiD:
**Summary**

Currently `InferSchemaType<>` returns weird types when using `Schema.Types.*`, check #12431 for better explanation

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -721,19 +721,22 @@ function gh12450() {
     decimalValue: { type: Schema.Types.Decimal128, required: true }
   });
 
-  type A = InferSchemaType<typeof Schema2>;
-
-  // Understand why required breaks the type
+  // Understand why required on Schema.Types.Decimal128 breaks the type
   expectType<{ createdAt: Date, decimalValue: Types.Decimal128 }>({} as InferSchemaType<typeof Schema2>);
 
   const Schema3 = new Schema({
+    createdAt: { type: Date, required: true },
+    decimalValue: { type: Schema.Types.Decimal128 }
+  });
+
+  expectType<{ createdAt: Date, decimalValue?: Types.Decimal128 }>({} as InferSchemaType<typeof Schema3>);
+
+  const Schema4 = new Schema({
     createdAt: { type: Date },
     decimalValue: { type: Schema.Types.Decimal128 }
   });
 
-  type U = InferSchemaType<typeof Schema3>;
-
-  expectType<{ createdAt?: Date, decimalValue?: Types.Decimal128 }>({} as InferSchemaType<typeof Schema3>);
+  expectType<{ createdAt?: Date, decimalValue?: Types.Decimal128 }>({} as InferSchemaType<typeof Schema4>);
 
 }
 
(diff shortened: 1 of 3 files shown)


hasezoey (changes requested):
This change seems to break some existing types tests (see test runs), could you maybe fix those and add new ones that this change is meant to address?

Note: in case you dont know how to run those types tests locally:
- install everything with `npm install`
- and run the types test with `npm run test-tsd`

HiiiiD:
Yes, I'm testing right now

HiiiiD:
I've rebased on top of master, but I have an issue that I don't know how to fix
I've tried for an hour to find out why `required: true` with `Schema.Types.Decimal128` breaks the types but I haven't found a solution.
Any clues @hasezoey?

hasezoey:
The following would fix it for now as mentioned in https://github.com/Automattic/mongoose/issues/12431#issuecomment-1256026048, though i dont know why it happens (probably a typescript bug?)

```diff
index 7c0038a869..ff9914217c 100644
--- a/types/inferschematype.d.ts
+++ b/types/inferschematype.d.ts
@@ -38,7 +38,16 @@ declare module 'mongoose' {
    * // result
    * type UserType = {userName?: string}
    */
-  type InferSchemaType<SchemaType> = ObtainSchemaGeneric<SchemaType, 'DocType'>;
+  type InferSchemaType<SchemaType> = ObtainDocType<SchemaType>; // "ObtainDocType" is used over "ObtainSchemaGeneric", because "ObtainSchemaGeneric" would return a modified result
+
+  /**
+   * @summary Obtains the "DocType" generic of a Schema, has to be used because with "ObtainSchemaGeneric" it would return a modified type
+   * @param {TSchema} TSchema A generic of schema type instance
+   */
+  type ObtainDocType<TSchema> =
+    TSchema extends Schema<any, any, any, any, any, any, any, infer DocType>
+      ? DocType
+      : unknown;
 
   /**
    * @summary Obtains schema Generic type by using generic alias.
```

[inferschematype.patch.txt](https://github.com/Automattic/mongoose/files/9632766/inferschematype.patch.txt)

HiiiiD:
I've applied your changes @hasezoey but have failing tests

hasezoey:
i have updated my comment at https://github.com/Automattic/mongoose/issues/12431#issuecomment-1256026048 because it (after some more testing) contained "wrong" information.

it seems like if we dont do `infer TPathTypeKey` but instead `any`, then the `DocType` for the thing this PR is meant to solve, solves it but creates other issues in other places, but as soon as `infer SomeName` is done again for the `TPathTypeKey` generic, then the type this PR is trying to solve, breaks again
but i have no clue what that generic is there for or what it affects or why typescript does not give the correct type

vkarpov15 (commented):
I suspect this will be fixed by #12352, I'll check a little later.

vkarpov15:
Confirmed this is fixed by #12352, without any test failures. I'm going to add the test to #12352 and close this issue. Thanks for pointing out this issue :+1:

PR #12459, opened 2022-09-22
fix(types): remove unused defaults in "PluginFunction"

hasezoey:
**Summary**

Remove default from helper type `PluginFunction`, these default are unnecessary because all values are passed-through from where the helper type is used and so they go always unused

i already did a ts-benchmark, but nothing changed (`Instantiations` are the same as master and this branch, and time is within margin of error)


vkarpov15 (approved):
Thanks :+1:

Issue #12457, opened 2022-09-22
Return-Types for `Schema.prototype.discriminator` do not align with runtime behavior


hasezoey:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Currently `Schema.prototype.discriminator` takes the discriminator schema as a argument (argument number 2 of 2) and `this` is the parent schema the discriminator schema is applied on and returns `this` at the end.
But the types currently imply that this function returns the discriminator schema (the schema from the argument):

Runtime:
https://github.com/Automattic/mongoose/blob/4495cf3a407a1fd6443b28a437800ed6a1b3c133/lib/schema.js#L528-L532

Types:
https://github.com/Automattic/mongoose/blob/4495cf3a407a1fd6443b28a437800ed6a1b3c133/types/index.d.ts#L198

PR #12460, opened 2022-09-22
Update `DiscriminatorSchema` to have better names and combine statics

hasezoey:
**Summary**

This PR updates `DiscriminatorSchema` to:
- have better and more reasonable names
- also transfer & combine all statics from both schemas

also see #12457 (which could make this PR obsolete when resolved)


vkarpov15 (approved):
Thanks :+1:

PR #12462, opened 2022-09-22
feat(index): add ability to set multiple options at once in ".set"

### LINKED ISSUE #10438: Allow passing options object to `Mongoose#set`
Enhancement proposal.

Currently, `Mongoose#set` only supports the `mongoose.set(key, value)` signature.
```javascript
mongoose.set('useCreateIndex', true);
mongoose.set('useNewUrlParser', true);
mongoose.set('useUnifiedTopology', true);
```

It would be nice if we could pass an object to set multiple options in a single function call. 
```javascript
mongoose.set({
  useCreateIndex: true,
  useFindAndModify: false,
  useNewUrlParser: true,
  useUnifiedTopology: true,
});
```

Mongoose: 5.13.2

hasezoey:
**Summary**

This PR adds the ability to set multiple options at once for `mongoose.set`

behavior is:
- if `key` is not a `object` and no `value` is given, return the option (like before)
- if `key` is not a `object` and `value` is given, convert them to a `{ key: value }` object
- if `key` is a `object`, ignore `value` and use that object
- if there is a error for a specific option, collect the error and evaluate all other options
- if there is only `1` error, throw that error (to keep backwards compat)
- if there are multiple errors, throw the whole array of errors

fixes #10438

### DIFF
--- a/lib/error/setOptionError.js
+++ b/lib/error/setOptionError.js
@@ -5,8 +5,8 @@
 'use strict';
 
 const MongooseError = require('./mongooseError');
-// const getConstructorName = require('../helpers/getConstructorName');
 const util = require('util');
+const combinePathErrors = require('../helpers/error/combinePathErrors');
 
 class SetOptionError extends MongooseError {
   /**
@@ -25,7 +25,7 @@ class SetOptionError extends MongooseError {
    * Console.log helper
    */
   toString() {
-    return _generateMessage(this);
+    return combinePathErrors(this);
   }
 
   /**
@@ -53,7 +53,7 @@ class SetOptionError extends MongooseError {
     }
 
     this.errors[key] = error;
-    this.message = _generateMessage(this);
+    this.message = combinePathErrors(this);
   }
 }
 
@@ -81,27 +81,6 @@ Object.defineProperty(SetOptionError.prototype, 'toJSON', {
 Object.defineProperty(SetOptionError.prototype, 'name', {
   value: 'SetOptionError'
 });
-
-/*!
- * ignore
- */
-
-function _generateMessage(err) {
-  const keys = Object.keys(err.errors || {});
-  const len = keys.length;
-  const msgs = [];
-  let key;
-
-  for (let i = 0; i < len; ++i) {
-    key = keys[i];
-    if (err === err.errors[key]) {
-      continue;
-    }
-    msgs.push(key + ': ' + err.errors[key].message);
-  }
-
-  return msgs.join(', ');
-}
 
 class SetOptionInnerError extends MongooseError {
   /**
--- a/lib/error/setOptionError.js
+++ b/lib/error/setOptionError.js
@@ -1,0 +1,122 @@
+/*!
+ * Module requirements
+ */
+
+'use strict';
+
+const MongooseError = require('./mongooseError');
+// const getConstructorName = require('../helpers/getConstructorName');
+const util = require('util');
+
+class SetOptionError extends MongooseError {
+  /**
+   * Mongoose.set Error
+   *
+   * @api private
+   * @inherits MongooseError
+   */
+  constructor() {
+    super('');
+
+    this.errors = {};
+  }
+
+  /**
+   * Console.log helper
+   */
+  toString() {
+    return _generateMessage(this);
+  }
+
+  /**
+   * inspect helper
+   * @api private
+   */
+  inspect() {
+    return Object.assign(new Error(this.message), this);
+  }
+
+  /**
+  * add message
+  * @param {String} key
+  * @param {String|Error} error
+  * @api private
+  */
+  addError(key, error) {
+    if (error instanceof SetOptionError) {
+      const { errors } = error;
+      for (const optionKey of Object.keys(errors)) {
+        this.addError(optionKey, errors[optionKey]);
+      }
+
+      return;
+    }
+
+    this.errors[key] = error;
+    this.message = _generateMessage(this);
+  }
+}
+
+
+if (util.inspect.custom) {
+  // Avoid Node deprecation warning DEP0079
+  SetOptionError.prototype[util.inspect.custom] = SetOptionError.prototype.inspect;
+}
+
+/**
+ * Helper for JSON.stringify
+ * Ensure `name` and `message` show up in toJSON output re: gh-9847
+ * @api private
+ */
+Object.defineProperty(SetOptionError.prototype, 'toJSON', {
+  enumerable: false,
+  writable: false,
+  configurable: true,
+  value: function() {
+    return Object.assign({}, this, { name: this.name, message: this.message });
+  }
+});
+
+
+Object.defineProperty(SetOptionError.prototype, 'name', {
+  value: 'SetOptionError'
+});
+
+/*!
+ * ignore
+ */
+
+function _generateMessage(err) {
+  const keys = Object.keys(err.errors || {});
+  const len = keys.length;
+  const msgs = [];
+  let key;
+
+  for (let i = 0; i < len; ++i) {
+    key = keys[i];
+    if (err === err.errors[key]) {
+      continue;
+    }
+    msgs.push(key + ': ' + err.errors[key].message);
+  }
+
+  return msgs.join(', ');
+}
+
+class SetOptionInnerError extends MongooseError {
+  /**
+   * Error for the "errors" array in "SetOptionError" with consistent message
+   * @param {String} key
+   */
+  constructor(key) {
+    super(`"${key}" is not a valid option to set`);
+  }
+}
+
+SetOptionError.SetOptionInnerError = SetOptionInnerError;
+
+/*!
+ * Module exports
+ */
+
+module.exports = SetOptionError;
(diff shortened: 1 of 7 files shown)


lib/index.js
```diff
@@ -213,36 +218,61 @@ Mongoose.prototype.setDriver = function setDriver(driver) {
...
  * @api public
  */
 
 Mongoose.prototype.set = function(key, value) {
   const _mongoose = this instanceof Mongoose ? this : mongoose;
 
-  if (VALID_OPTIONS.indexOf(key) === -1) {
-    throw new Error(`\`${key}\` is an invalid option.`);
+  if (arguments.length === 1 && typeof key !== 'object') {
+    return _mongoose.options[key];
```
AbdelrahmanHafez, line 231: Previously we were validating the key even for getters, we should do the same here as well.

lib/index.js
```diff
@@ -213,36 +218,61 @@ Mongoose.prototype.setDriver = function setDriver(driver) {
...
-        get: function() {
-          return this;
-        }
-      });
-    } else {
-      delete mongoose.Types.ObjectId.prototype._id;
+  // array for errors to collect all errors for all key-value pairs, like ".validate"
+  const errors = [];
+
+  for (const [k, v] of Object.entries(options)) {
```
AbdelrahmanHafez, line 247: nitpick: `k`, `v` aren't very descriptive, I'd go for something like `key`, `value` which would be more readable.
```suggestion
  for (const [key, value] of Object.entries(options)) {
```

hasezoey: > nitpick: k, v aren't very descriptive, I'd go for something like key, value which would be more readable.

the problem with the suggested are that those names are already taken, without renaming the function parameters

lib/index.js
hasezoey: done with 72bd56cb4022af40413c4f19614c8746d215cac3

lib/index.js
```diff
@@ -213,36 +218,61 @@ Mongoose.prototype.setDriver = function setDriver(driver) {
...
+      }
+    }
+  }
+
+  if (errors.length) {
+    // this is to keep backwards-compat
+    if (errors.length === 1) {
+      throw errors[0];
     }
+    throw errors;
```
AbdelrahmanHafez, line 279: [We should not throw an array](https://eslint.org/docs/latest/rules/no-throw-literal), because that does not include the stack trace, for example, even though each individual error has it.

For that, we'd have to create a new error that aggregates all the errors and adds them as to a single property. Look at `lib/error/validation.js` to see how we aggregate multiple validation errors into a single error object with a concatenated message and an `errors` property, including the individual errors.

test/index.test.js
```diff
@@ -1102,4 +1102,71 @@ describe('mongoose module:', function() {
...
+
+    it('should throw a a array of errors when using multiple invalid keys', function() {
+      try {
+        m.set({
+          invalid1: true,
+          invalid2: true
+        });
+        assert.fail('Expected .set to throw');
+      } catch (err) {
+        assert.ok(err instanceof Array);
```
AbdelrahmanHafez, line 1148: Same comment as above, this is an anti-pattern.

hasezoey: didnt know about that, will look to add a new error

lib/index.js
hasezoey: if you mention this rule and want to actually enforce it, maybe consider adding it to the current linting rules, because it seems like it is not there yet or by default

test/index.test.js
```diff
@@ -1102,4 +1102,71 @@ describe('mongoose module:', function() {
...
+        m.set({
+          invalid: true,
+          debug: true
+        });
+        assert.fail('Expected .set to throw');
+      } catch (err) {
+        assert.ok(err instanceof Error);
+        assert.strictEqual(err.message, '`invalid` is an invalid option.');
+        assert.strictEqual(m.options['debug'], true);
+      }
```
AbdelrahmanHafez, line 1169: This test and the one above would pass even if `m.set(...)` did not throw an error because it would not get to the `catch` block, and there would be no assertions.

Search for `threw = false;` in the codebase to see how we solve this problem.

test/types/base.test.ts
```diff
@@ -51,3 +51,12 @@ function gh12100() {
...
 }
+
+function setAsObject() {
+  mongoose.set({
+    debug: true,
+    autoIndex: false
+  });
+
+  expectError(mongoose.set({ invalid: true }));
+}
```
AbdelrahmanHafez, line 62: Nice!

test/index.test.js
hasezoey: i dont quite understand what you mean, the successful path has a `assert.fail()` after it, and this check is explicitly checking for a specific error message

lib/index.js
AbdelrahmanHafez: I agree that we shouldn't shadow the original key/value.

We can use special identifiers for them and avoid conflict. What do you think of `optionKey` and `optionValue`?

hasezoey: sure

hasezoey: done in 28b12e26670571088f1b634fbd19adbfebb8f7db

lib/index.js
AbdelrahmanHafez: I don't know if ESLint will enforce it against arrays or not, but feel free to add it to this PR as well if it's not part of the `eslint/recommended` rules.

test/index.test.js
AbdelrahmanHafez: My bad, I didn't notice the `assert.fail(...)` line.
This is fine then, thanks. 👍

lib/index.js
hasezoey: actually, i just looked, it is already in the eslint config, but it seems like it did not catch that (it is set to `error`)

AbdelrahmanHafez: Yep, I don't think that ESLint is smart enough to pick up the types from variables.
It protects against `throw "Something bad happened.";` for example, but not other more complicated cases.

hasezoey: fixed with 4ab2e442c4eeb50c1cd11407a10a5017da16ba05

test/index.test.js
hasezoey: fixed with 4ab2e442c4eeb50c1cd11407a10a5017da16ba05

test/index.test.js
```diff
@@ -500,16 +501,13 @@ describe('mongoose module:', function() {
   });
 
   it('throws an error on setting invalid options (gh-6899)', function() {
-    let threw = false;
     try {
       mongoose.set('someInvalidOption', true);
+      assert.fail('Expected mongoose.set to throw');
```
vkarpov15, line 506: Can you replace this with `assert.throws()` instead please?

lib/error/setOptionError.js
```diff
@@ -0,0 +1,122 @@
...
+
+Object.defineProperty(SetOptionError.prototype, 'name', {
+  value: 'SetOptionError'
+});
+
+/*!
+ * ignore
+ */
+
+function _generateMessage(err) {
```
vkarpov15, line 89: It looks like this is copied from `error/validation.js`. Maybe we move this function into a `lib/helpers/error/combinePathErrors.js` file?

test/index.test.js
```diff
@@ -1102,4 +1100,85 @@ describe('mongoose module:', function() {
...
+        assert.ok(err.errors['invalid1'] instanceof SetOptionError.SetOptionInnerError);
+        assert.strictEqual(err.errors['invalid1'].message, '"invalid1" is not a valid option to set');
+        assert.ok(err.errors['invalid2'] instanceof SetOptionError.SetOptionInnerError);
+        assert.strictEqual(err.errors['invalid2'].message, '"invalid2" is not a valid option to set');
+      }
+    });
+
+    it('should apply all values, even if there are errors', function() {
+      assert.strictEqual(m.options['debug'], undefined);
+      try {
```
vkarpov15, line 1159: Same, just use `assert.throws()` here

vkarpov15 (commented):
Couple of minor comments. But this looks great!

test/index.test.js
hasezoey: sure, should other tests also be re-written to use this? (i based my code style on the other tests (just removed the unnecessary `let threw = false`)

hasezoey: i just tried to do this, but it seems like `assert.throws` only supports checking that it is a instance of a error and the message, but does not return the error anywhere to check additional properties, which is used in the current version of the new tests to check each error (and the `errors` object) individually

lib/error/setOptionError.js
hasezoey: done in 6596b2ceae5e317a9b6e22939b32bb28206b165e

test/index.test.js
vkarpov15: That's a good point, we can't use `assert.throws()`. Please disregard this comment.

vkarpov15 (approved):
LGTM. Any objection to merging this into the 6.7 branch?

AbdelrahmanHafez (approved):
LGTM, thanks 👍

hasezoey:
> LGTM. Any objection to merging this into the 6.7 branch?

no, sounds good to me

PR #12463, opened 2022-09-22
fix(types): infer number enum types from schema if using `enum: [0, 1] as const`

### LINKED ISSUE #12242: Missing inferred `enum` property when type is Number
### Prerequisites

### Mongoose version

6.5.1

### Node.js version

17.3.0

### MongoDB server version

5.0.5

### Description

When you precise an `enum` property on a Schema, and when type is `Number`, the type doesn't take in account the enum array, but `InferSchemaType` resolves to `number`.

### Steps to Reproduce

```ts
const dbExample = new Schema(
  {
    active: { type: Number, enum: [0, 1], required: true }
  }
);

type Example = InferSchemaType<typeof dbExample>;
```
Here, in type `Example`, `active` will be `number`.

### Expected Behavior

Following the previous example, `active` should be `0 | 1`.

vkarpov15:
Fix #12242

**Summary**

#12242 shows that we can do a better job of inferring enum types for numbers, like how we infer enums from strings. You do need to use `as const` as @hasezoey pointed out, but otherwise this approach seems to work.

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -741,3 +741,14 @@ function pluginOptions() {
   schema.plugin<any, SomePluginOptions>(pluginFunction2, { option2: 0 });
   expectError(schema.plugin<any, SomePluginOptions>(pluginFunction2, {})); // should error because "option2" is not optional
 }
+
+function gh12242() {
+  const dbExample = new Schema(
+    {
+      active: { type: Number, enum: [0, 1] as const, required: true }
+    }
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM <sub>though complex types and typescript syntax look like magic to me sometimes</sub>

AbdelrahmanHafez (approved):
LGTM, thanks. 👍

PR #12470, opened 2022-09-26
style: add names to previously anonymous plugin functions

hasezoey:
**Summary**

This PR adds names to previously `anonymous` plugin functions, which could be logged with `schema.plugins`, this should make it easier to debug if there should ever be the need (and for users to actually see what it is doing instead of a unhelpful `anonymous`)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #12471, opened 2022-09-26
Add more names to previously `anonymous` functions

hasezoey:
**Summary**

This PR adds names (again) to previously unnamed functions, this time focusing on hook functions.
This will help when having to debug how / if any hooks were applied (for example having to debug why a hook as been applied 2 times)

continuating of #12470


AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #12483, opened 2022-09-29
fix(types): add "array of array key-value pairs" as a argument option for "query.sort()"

### LINKED ISSUE #12434: find.sort() params no longer accept an array
### Prerequisites

### Mongoose version

6.6.1

### Node.js version

16

### MongoDB server version

5.0.12

### Description

Hi 

I am just upgrading from 6.3.1 to 6.6.1 and I noticed that the sort() funciton params have changed
I can not really find any explenaiton why or even any mention of this change being expected i.e its not in the release notes

It appears to have been done in this commit 8d917a98e1a2f491d0e9888625b435b760050d39, feels like a regression but maybe it was intentional?

I used to be able to do

```
        .sort([
            ['year', 1],
            ['quarter', 1],
            ['setNumber', 1],
        ])
``` 

But now it appears it has to be

```
        .sort({
            year: 1,
            quarter: 1,
            setNumber: 1,
        })
```

### Steps to Reproduce

Attempt to pass an array of arrays to sort()

```
        .sort([
            ['year', 1],
            ['quarter', 1],
            ['setNumber', 1],
        ])
```

### Expected Behavior

_No response_

hasezoey:
**Summary**

This PR adds a type for the "array of array with key-value pairs" as a possible type
also adds some more documentation about this way to the jsdoc

fixes #12434
was broken by #11852

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -2952,7 +2952,7 @@ Query.prototype.distinct = function(field, conditions, callback) {
  *     query.sort('field -test');
  *
  *     // also possible is to use a array with array key-value pairs
- *     query.sort([["field", "asc"]]);
+ *     query.sort([['field', 'asc']]);
  *
  * #### Note:
  *
--- a/lib/query.js
+++ b/lib/query.js
@@ -2951,11 +2951,14 @@ Query.prototype.distinct = function(field, conditions, callback) {
  *     // equivalent
  *     query.sort('field -test');
  *
+ *     // also possible is to use a array with array key-value pairs
+ *     query.sort([["field", "asc"]]);
+ *
  * #### Note:
  *
  * Cannot be used with `distinct()`
  *
- * @param {Object|String} arg
+ * @param {Object|String|Array.<Array<(string | number)>>} arg
  * @return {Query} this
  * @see cursor.sort https://docs.mongodb.org/manual/reference/method/cursor.sort/
  * @api public
--- a/lib/query.js
+++ b/lib/query.js
@@ -2958,7 +2958,7 @@ Query.prototype.distinct = function(field, conditions, callback) {
  *
  * Cannot be used with `distinct()`
  *
- * @param {Object|String|Array.<Array<(string | number)>>} arg
+ * @param {Object|String|Array<Array<(string | number)>>} arg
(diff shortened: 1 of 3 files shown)


lib/query.js
```diff
@@ -2951,11 +2951,14 @@ Query.prototype.distinct = function(field, conditions, callback) {
...
  *
+ *     // also possible is to use a array with array key-value pairs
+ *     query.sort([["field", "asc"]]);
+ *
  * #### Note:
  *
  * Cannot be used with `distinct()`
  *
- * @param {Object|String} arg
+ * @param {Object|String|Array.<Array<(string | number)>>} arg
```
AbdelrahmanHafez, line 2961: Is the extra dot a typo?
```suggestion
 * @param {Object|String|Array<Array<(string | number)>>} arg
```

lib/query.js
```diff
@@ -2951,11 +2951,14 @@ Query.prototype.distinct = function(field, conditions, callback) {
  *     // equivalent
  *     query.sort('field -test');
  *
+ *     // also possible is to use a array with array key-value pairs
+ *     query.sort([["field", "asc"]]);
```
AbdelrahmanHafez, line 2955: Nitpick: single quotes instead of double quotes.
```suggestion
 *     query.sort([['field', 'asc']]);
```

lib/query.js
hasezoey: not that i know of, though https://jsdoc.app/tags-type.html , though i should either remove it, or add it to the second array too

AbdelrahmanHafez (approved):
One potential typo, but overall LGTM.
Thanks 👍

hasezoey:
Force pushed to fix a commit message and a oversight in the test types

lib/query.js
hasezoey: fixed with 06b3c6e6445c1d89a3b88060f98788fe64ecd535

lib/query.js
hasezoey: changed to be consistent with 8bbcd93f715e13c656691e4224287bbbd74edc7a

AbdelrahmanHafez (approved):

PR #12485, opened 2022-09-29
fix(query): treat `findOne(_id)` as equivalent to `findOne({ _id })`

### LINKED ISSUE #12325: Breaking change in findOne(ObjectId) method from mongoose 5 to mongoose 6
### Prerequisites

### Mongoose version

6.5.2

### Node.js version

14

### MongoDB server version

4.8.1

### Description

Hello 👋 

The query `findOne(ObjectId(id)` is not working as expected anymore.  **Even more it return not the document we are looking for.** 

I looked in Changelog, and Documation about migrating from 5 to 6, I don't see this documented. 

**The native driver of mongo handles this correctly** and I think mongoose should handle this too.
When I looked a  [#8268](https://github.com/Automattic/mongoose/issues/8268), I understood that was a bug and was fixed on >5.7.

We had a critical bug because a completely different document were returned, please let me knows. 
Thank you.

PS: 
_I understand this is not the best solution to find a document with an ObjectId given, we have findById for that.
But this should throw an error at least, or return a null value but never another document._

### Steps to Reproduce

```javascript
const connection = mongoose.createConnection("mongodb://localhost:27017/test");

const userSchema = new mongoose.Schema(
  {
    name: String,
  },
  {
    collection: "usersFid",
    versionKey: false,
  }
);

const testModel = connection.model("User", userSchema);

let testId = null;

async function populateCollection() {
  await Promise.all(
    [...Array(100).keys()].map((idx) =>
      new testModel({ name: `name${idx}` }).save()
    )
  );
}

async function cleanInsertTest() {
  await testModel.deleteMany({});

  await populateCollection();

  testId = (await new testModel({ name: "User To Get" }).save())._id;

  await populateCollection();
}

async function findTest() {
  const result = await testModel.findOne(testId).lean().exec();
  console.log("Search Id", testId.toString()); // NO THE SAME USER !
  console.log("Doc Found:", result); // NO THE SAME USER !

  //in 5.X.X it prints the inserted test user
  //in 6.X.X it prints random user
}

async function run() {
  await cleanInsertTest();
  await findTest();
  connection.close();
}

run();
```

### Expected Behavior

_No response_

vkarpov15:
Fix #12325

**Summary**

#12325 pointed out that, in Mongoose 5, `findOne(ObjectId)` worked the same as `findById(ObjectId)`. I don't see any reason why we shouldn't continue to support this behavior.

**Examples**

### DIFF
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -4132,4 +4132,15 @@ describe('Query', function() {
     assert.equal(item.get('select.key.some'), 'value');
     assert.equal(item.doNotSelect, undefined);
   });
+
+  it('treats ObjectId as object with `_id` for `merge()` (gh-12325)', async function() {
+    const testSchema = new mongoose.Schema({ name: String });
+    const Test = db.model('Test', testSchema);
+    const _id = new mongoose.Types.ObjectId();
+
+    const q = Test.find(_id);
+
+    assert.ok(q.getFilter()._id instanceof mongoose.Types.ObjectId);
+    assert.equal(q.getFilter()._id.toHexString(), _id.toHexString());
+  });
 });
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -4138,7 +4138,12 @@ describe('Query', function() {
     const Test = db.model('Test', testSchema);
     const _id = new mongoose.Types.ObjectId();
 
-    const q = Test.find(_id);
+    let q = Test.find(_id);
+
+    assert.ok(q.getFilter()._id instanceof mongoose.Types.ObjectId);
+    assert.equal(q.getFilter()._id.toHexString(), _id.toHexString());
+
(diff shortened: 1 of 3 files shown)


test/query.test.js
```diff
@@ -4132,4 +4132,15 @@ describe('Query', function() {
     assert.equal(item.get('select.key.some'), 'value');
     assert.equal(item.doNotSelect, undefined);
   });
+
+  it('treats ObjectId as object with `_id` for `merge()` (gh-12325)', async function() {
+    const testSchema = new mongoose.Schema({ name: String });
+    const Test = db.model('Test', testSchema);
+    const _id = new mongoose.Types.ObjectId();
+
+    const q = Test.find(_id);
```
Uzlopak, line 4141: can we have also a test for findOne please?

PR #12458, opened 2022-09-22
test(types-plugin): reproduce #12454

hasezoey:
**Summary**

Add a types test case for #12454

Error:
```txt
  test/types/plugin.test.ts:72:18
  ✖  72:18  Argument of type (schema: Schema<any, Model<any, any, any, any, any>, {}, {}, {}, {}, "type", { [x: string]: any; }>) => void is not assignable to parameter of type PluginFunction<Test, TestModel, TestInstanceMethods, TestQueryHelpers, TestVirtuals, TestStaticMethods>.
  Types of parameters schema and schema are incompatible.
    Type Schema<Test, TestModel, TestInstanceMethods, TestQueryHelpers, TestVirtuals, TestStaticMethods, "type", Test> is not assignable to type Schema<any, Model<any, any, any, any, any>, {}, {}, {}, {}, "type", { [x: string]: any; }>.
      The types returned by discriminator(...) are incompatible between these types.
        Type Schema<Omit<Test, never>, unknown, unknown, unknown, unknown, {}, "type", Omit<Test, never>> | Schema<Test, TestModel, TestInstanceMethods, ... 4 more ..., Test> is not assignable to type Schema<{ [x: string]: any; }, Model<any, any, any, any, any>, {}, {}, {}, {}, "type", { [x: string]: any; }> | Schema<Omit<{ [x: string]: any; }, never>, unknown, unknown, unknown, unknown, {}, "type", Omit<...>>.
          Type Schema<Test, TestModel, TestInstanceMethods, TestQueryHelpers, TestVirtuals, {}, "type", Test> is not assignable to type Schema<{ [x: string]: any; }, Model<any, any, any, any, any>, {}, {}, {}, {}, "type", { [x: string]: any; }> | Schema<Omit<{ [x: string]: any; }, never>, unknown, unknown, unknown, unknown, {}, "type", Omit<...>>.
            Type Schema<Test, TestModel, TestInstanceMethods, TestQueryHelpers, TestVirtuals, {}, "type", Test> is not assignable to type Schema<{ [x: string]: any; }, Model<any, any, any, any, any>, {}, {}, {}, {}, "type", { [x: string]: any; }>.
              Types of property statics are incompatible.
                Type {} & { [name: string]: (this: TestModel, ...args: any[]) => any; } is not assignable to type {} & { [name: string]: (this: Model<any, any, any, any, any>, ...args: any[]) => any; }.
[...]

PR #12486, opened 2022-09-29
fix(types): make schema.plugin() more flexible for schemas that don't define any generics

### LINKED ISSUE #12454: How to correctly type custom plugins? Also unable to register previously working plugins since v6.5.2
### Prerequisites

### Mongoose version

6.6.1

### Node.js version

16.14.0

### MongoDB version

5.0.5

### Operating system

Windows

11

### Issue

My custom plugin used to be typed correctly up to version `6.5.1` where I'm basically following [this guide](https://mongoosejs.com/docs/typescript/statics-and-methods.html#statics). However, starting from v6.5.2 to the current v6.6.1, I get the following:
image

I was also unable to register the [mongoose-unique-validator](https://github.com/mongoose-unique-validator/mongoose-unique-validator#readme) plugin and being shown the following error in my IDE:
image

I can't seem to find any info in the docs regarding how to correctly type my custom plugin nor am I sure if this is a bug or a breaking change.

vkarpov15:
Fix #12454
Re: #12139

**Summary**

I think just switching to `any` should be enough to fix #12454, it doesn't break any of the tests from #12139. Any thoughts @emiljanitzek ?

**Examples**

### DIFF
--- a/test/types/plugin.test.ts
+++ b/test/types/plugin.test.ts
@@ -59,10 +59,17 @@ const testSchema = new Schema<Test, TestModel, TestInstanceMethods, TestQueryHel
   lastName: { type: String, required: true }
 });
 
+function pluginGeneric(schema: Schema): void {
+  schema.static('test', function() {
+    return 0;
+  });
+}
+
 testSchema.plugin(pluginVirtuals);
 testSchema.plugin(pluginQueryHelpers);
 testSchema.plugin(pluginMethods);
 testSchema.plugin(pluginStatics);
+testSchema.plugin(pluginGeneric);
 
 const Foo = connection.model<Test, TestModel, TestQueryHelpers>('Test', testSchema);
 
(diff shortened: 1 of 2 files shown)


emiljanitzek:
This would make the plugin function type a lot more loose and you would not get the same type safety. So I would prefer it not to change. I would say that the problem is that the Schema class is too strictly typed by default. I haven't tested this but I think changing the Schema class generics should have the same effect without changing the plugin interface.

```
export class Schema<
    EnforcedDocType = any,
    M = Model<EnforcedDocType, any, any, any>,
    TInstanceMethods = any,
    TQueryHelpers = any,
    TVirtuals = any,
    TStaticMethods = any,
    TPathTypeKey extends TypeKeyBaseType = DefaultTypeKey,
    DocType extends ObtainDocumentType<DocType, EnforcedDocType, TPathTypeKey> = ObtainDocumentType<any, EnforcedDocType, TPathTypeKey>>
    extends events.EventEmitter {
```

hasezoey (approved):
LGTM, somehow i missed this as a possible solution when making #12458

PR #12488, opened 2022-09-30
fix(types): avoid relying on `typeof this`, which isn't supported in TypeScript < 4.4

### LINKED ISSUE #12375: This line broke typings in a project I help maintain:
This line broke typings in a project I help maintain:

```
toConstructor(): typeof this;
```

That project uses `typescript` `4.1.3`, which is quite dated. When dependabot bumped `mongoose` from `6.2.0` to `6.5.4`, CI broke with this error:

```
node_modules/mongoose/types/query.d.ts:619:29 - error TS2304: Cannot find name 'this'.

619     toConstructor(): typeof this;
                                ~~~~


Found 1 error.

error Command failed with exit code 2
```

I tried bumping `typescript`, but it was not very successful. Versions below `4.4` showed the same error and `v4.4+` generated a lot of other errors, unrelated to the library. Most of them were to do with `error` becoming `unknown` instead of `any` in `catch`.

Not sure if this is actionable, but flagging the problem just in case.

_Originally posted by @kachkaev in https://github.com/Automattic/mongoose/issues/12284#issuecomment-1234662198_

vkarpov15:
Fix #12375

**Summary**

It looks like we can avoid relying on `typeof this` just by redefining the `Query` type. No perf issues, so I don't see any reason why we shouldn't do this. What do you think @Uzlopak ?

**Examples**

### DIFF
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -616,7 +616,7 @@ declare module 'mongoose' {
     then: Promise<ResultType>['then'];
 
     /** Converts this query to a customized, reusable query constructor with all arguments and options retained. */
-    toConstructor(): typeof this;
+    toConstructor(): typeof Query<ResultType, DocType, THelpers, RawDocType>;
 
     /** Declare and/or execute this query as an update() operation. */
     update(filter?: FilterQuery<DocType>, update?: UpdateQuery<DocType> | UpdateWithAggregationPipeline, options?: QueryOptions<DocType> | null, callback?: Callback<UpdateWriteOpResult>): QueryWithHelpers<UpdateWriteOpResult, DocType, THelpers, RawDocType>;


hasezoey:
maybe it would also help to explicitly define what typescript version is supported, currently i assume(d) that mongoose only supports typescript 4.8, because that is the version that is used in the tests (and set in the package.json)

vkarpov15:
Because `toConstructor()` returns a constructor you can use to create new instances of the same query. Admittedly it isn't used much, and `clone()` serves a similar purpose. But `typeof this` or `typeof Query<...>` should be interchangeable AFAIK, unless someone is using a class that `extends Query`.

AbdelrahmanHafez (approved):
LGTM, thanks! 👍 

I don't think we should entertain all backward compatibility requests for older TS versions. Upgrading TS to newer versions is reasonably straightforward, and using the latest versions means we can utilize the new features that the TS team keeps shipping.

vkarpov15:
Yeah I don't want to add a separate CI matrix for TypeScript versions. Supporting one version of TypeScript is hard enough, given how immature and unstable TypeScript is. But we can add small fixes like this to unblock compiling on older TS versions, if someone requests it and the fix doesn't break anything else.

PR #12491, opened 2022-09-30
feat(schema): add `alias()` method that makes it easier to define multiple aliases for a given path

### LINKED ISSUE #12368: Add `Schema#alias()` function that defines a new alias
### Prerequisites

### 🚀 Feature Proposal

When creating documents from api data, users will sometimes have their schemas be identical to the keys coming in. Therefore, I propose we create a function that does the mapping and returns the resultant object.

### Motivation

It is a bit tedious to have to manually map the result of the api call to your documents when the keys are identical. We could possibly take this one step further and implement an option that converts the keys of the apiResult to match the ones defined in mongoose. So if an api is in snake case but your models are in camel case, the function would convert the keys to camel, then map.

### Example

```javascript
function apiMap(apiResult, schema) {
  const mongooseDoc = {};
  schema.keyName === apiResult.keyName
  mongooseDoc[schema.keyName] = apiResult.keyValue

  return mongooseDoc
}

const guts = apiMap(apiResult, schema)
await Model.create(guts);
```

vkarpov15:
Fix #12368

**Summary**

One approach we recommend for renaming fields from API requests is using aliases. Here's an example of [using an alias to convert the GitHub API's `avatar_url` response to `avatar` in Mongoose](https://github.com/mongoosejs/backend/blob/7a40a525291dc2e284b78829b453b1ac59926385/src/db/subscriber.js#L8). But right now there's no way to define multiple aliases for a single field. This PR adds a new schema method, so you can do something like `schema.alias('avatar', ['avatar_url', 'photo'])`.

**Examples**


hasezoey (changes requested):
LGTM, with some minor details

changes to the types still need to be done

AbdelrahmanHafez (changes requested):
Nice!

Would also be great if we make the API support arrays of strings as follows:
```js
const schema = new Schema({
  avatar: { type: String, alias: ['avatarURL', 'avatarImageURL'] }
});
```

Also, would be nice if we could have TS types and tests for those changes.

lib/schema.js
```diff
@@ -659,9 +692,50 @@ Schema.prototype.add = function add(obj, prefix) {
...
+ * @param {String|String[]} alias the path(s) to use as an alias for `path`
+ * @return {Schema} the Schema instance
+ * @api public
+ */
+
+Schema.prototype.alias = function alias(path, alias) {
+  if (Array.isArray(alias)) {
+    for (const a of alias) {
+      this.alias(path, a);
+    }
```
hasezoey, line 735: with the updates, couldnt this be changed to just `aliasFields(this, { [path]: alias })`? (or combine array and non-array paths)?

hasezoey:
also just to be sure, where is checked that a `alias` cannot be a special name like `__proto__`/`prototype`?

AbdelrahmanHafez (approved):
LGTM, thanks 👍

hasezoey (approved):
LGTM

PR #12506, opened 2022-10-02
fix(model): avoid saving applied defaults if path is deselected

### LINKED ISSUE #12414: Saving a document that was fetched with .select() causes not selected array to be saved empty
### Prerequisites

### Mongoose version

6.5.2

### Node.js version

16.x

### MongoDB server version

5.0.6

### Description

This seems to have been changed from version 6.2.2 and 6.2.3. Still an issue in 6.6.0.

```js
const dividerSchema = new mongoose.Schema({
    dividerKey: { type: String, default: "" }
}, { _id: false });

const dataSchema = new mongoose.Schema({
    sectionA: {
        include: { type: Boolean, default: true },
        dividers: [dividerSchema]
    },
    sectionB: {
        include: { type: Boolean, default: true },
        dividers: [dividerSchema],
    }
});
```

For the given schema if we try to retrieve a document 
```
DataModel.findOne({ "_id": createdDocument._id }).select("sectionA.include sectionA.carrierId sectionB").exec();
``` 
and then imminently save it the following update query is created

```
data_xes.updateOne({ _id: new ObjectId("631ba9fbadbfff8b4ab1f888") }, { '$set': { 'sectionA.dividers': [] } }, { session: null })
```

This is weird since I never selected or have touch that value.

### Steps to Reproduce

```js

const connectToMongo = async () => {
    mongoose.set("debug", true);
    await mongoose.connect("", {
        autoIndex: false,
        autoCreate: false
    });
}

const dividerSchema = new mongoose.Schema({
    dividerKey: { type: String, default: "" }
}, { _id: false });

const dataSchema = new mongoose.Schema({
    sectionA: {
        include: { type: Boolean, default: true },
        dividers: [dividerSchema]
    },
    sectionB: {
        include: { type: Boolean, default: true },
        dividers: [dividerSchema],
    }
});

export const DataModel = mongoose.model("data_x", dataSchema);

const doWork = async () => {
    const createdDocument = await DataModel.create({
        sectionA: {
            include: true,
            dividers: [
                {
                    dividerKey: "SOME_DIVIDER_KEY"
                },
                {
                    dividerKey: "SOME_DIVIDER_KEY"
                }
            ]
        },
        sectionB: {
            include: true,
            dividers: [
                {
                    dividerKey: "SOME_DIVIDER_KEY"
                },
                {
                    dividerKey: "SOME_DIVIDER_KEY"
                }
            ]
        }
    });

    const doc = await DataModel.findOne({ "_id": createdDocument._id }).select("sectionA.include sectionA.carrierId sectionB").exec();

    await doc.save();
};

connectToMongo()
.then(async () => await doWork())
.then(async () => {
    await mongoose.disconnect();
    process.exit(0);
})
.catch((err) => {
    console.error(err);
    process.exit(-1);
});
```

### Expected Behavior

I would expect non selected sub array not to update.

vkarpov15:
Fix #12414

**Summary**

If you load a document with an array deselected `const doc = await Model.findOne().select({ arr: 0 })` and then save with no changes, Mongoose will save an empty array. This looks like it applies to all deselected defaults.

This PR makes it so that Mongoose skips deselected defaults when calculating what changed in the document.

**Examples**

### DIFF
--- a/lib/helpers/projection/isPathExcluded.js
+++ b/lib/helpers/projection/isPathExcluded.js
@@ -12,6 +12,10 @@ const isDefiningProjection = require('./isDefiningProjection');
  */
 
 module.exports = function isPathExcluded(projection, path) {
+  if (projection == null) {
+    return false;
+  }
+
   if (path === '_id') {
     return projection._id === 0;
   }
--- a/lib/model.js
+++ b/lib/model.js
@@ -51,11 +51,13 @@ const {
   getRelatedDBIndexes,
   getRelatedSchemaIndexes
 } = require('./helpers/indexes/getRelatedIndexes');
+const isPathExcluded = require('./helpers/projection/isPathExcluded');
 const decorateDiscriminatorIndexOptions = require('./helpers/indexes/decorateDiscriminatorIndexOptions');
 const isPathSelectedInclusive = require('./helpers/projection/isPathSelectedInclusive');
 const leanPopulateMap = require('./helpers/populate/leanPopulateMap');
 const modifiedPaths = require('./helpers/update/modifiedPaths');
 const parallelLimit = require('./helpers/parallelLimit');
+const parentPaths = require('./helpers/path/parentPaths');
 const prepareDiscriminatorPipeline = require('./helpers/aggregate/prepareDiscriminatorPipeline');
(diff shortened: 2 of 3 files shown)


hasezoey (changes requested):
LGTM, though tests would be good to have

Issue #12342, opened 2022-08-28
TypeScript QueryHelpers return any instead of model type


florianwalther-private:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.5.3

### Node.js version

16.3.2

### MongoDB server version

5.0.11

### Description

I implemented a query helper function following the [tutorial documentation in the documentation](https://mongoosejs.com/docs/typescript/query-helpers.html).

As soon as I add this method, the return value of that query turns into `any`. It should return the proper model type instead.

### Steps to Reproduce

Implement a query helper as outlined in the documentation: https://mongoosejs.com/docs/typescript/query-helpers.html

### Expected Behavior

The return type after calling the query helper function should be the appropriate model type.

hasezoey:
i think the mentioned documentation is at least slightly out-of-date, because a proper type for query-helpers should be `QueryWithHelpers` instead of just `Query`?

florianwalther-private:
I'm not sure what the correct return type is because I find it a bit confusing, but it should certainly not be any. After calling `exec()`/`await`, I want the model back like it does without the query helper chained to it.

hasezoey:
re https://github.com/Automattic/mongoose/issues/12347#issuecomment-1230463627

> but the referenced (my) issue is not about the documentation

from what i can tell, this issue is about the documentation because "you followed the documentation, but its not working" and after having taken a look at the documentation and the current code in mongoose's actual types, i came to the conclusion that the documentation is probably outdated

---

reason for the documentation being outdated:
since the documentation has been written `QueryWithHelpers` has been added, which should make the types easier & cleaner, see

https://github.com/Automattic/mongoose/blob/0af4cb815ec36dd50d18755f5238506ef9297a5a/types/query.d.ts#L24

by taking care of the `Query<...> & Type`

also the documentation lists the first parameter as a static `any` to `Query<any, ...>`, which is the result type, and so ending up in the wrong type

TL;DR: change your type to `QueryWithHelpers` for less code, and also change the first parameter from `any` to the actual return type

currently the documentation is basically guiding to be like (like in issue #12347)
https://github.com/Automattic/mongoose/blob/6cb1284821405c71a4d75e9857ef45d95c367972/types/models.d.ts#L405

here it should actually be `ResultDoc` or `{}` and definitely not `any`, see 

[...]

PR #12507, opened 2022-10-03
TypeScript query helpers quick fix and docs improvements

vkarpov15:
**Summary**

I cleaned up the TS query helper docs based on feedback from #12342, and also found an issue with auto typed query helpers. Looks like they defaulted to using raw DocType rather than a `HydratedDocument`, which meant TS thought that every query helper returned a lean result.

**Examples**


hasezoey (approved):
LGTM

PR #12511, opened 2022-10-03
chore(mocha-fixtures): fix spelling and remove "example" comments

hasezoey:
**Summary**

This PR is a small follow-up of #12262, where i noticed some spelling mistakes and example comments left-overs, also using `MONGOOSE_REPLSET_URI` in the false case of `startMemoryReplset`

### DIFF
--- a/test/mocha-fixtures.js
+++ b/test/mocha-fixtures.js
@@ -12,7 +12,7 @@ let mongoinstance;
 // a replset-instance for running repl-set tests
 let mongorreplset;
 
-// decide wheter to start a in-memory or not
+// decide wheter to start a instance or not
 const startMemoryInstance = !process.env.MONGOOSE_TEST_URI;
 const startMemoryReplset = !process.env.MONGOOSE_REPLSET_URI;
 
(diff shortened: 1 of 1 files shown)


test/mocha-fixtures.js
```diff
@@ -12,7 +12,7 @@ let mongoinstance;
 // a replset-instance for running repl-set tests
 let mongorreplset;
 
-// decide wheter to start a in-memory or not
+// decide wheter to start a instance or not
```
AbdelrahmanHafez, line 15:
```suggestion
// decide whether to start an instance or not
```

test/mocha-fixtures.js
```diff
@@ -12,7 +12,7 @@ let mongoinstance;
 // a replset-instance for running repl-set tests
 let mongorreplset;
 
-// decide wheter to start a in-memory or not
+// decide whether to start an instance or not
```
AbdelrahmanHafez, line 15:
```suggestion
// decide whether to start an instance or use an in-memory database
```

AbdelrahmanHafez (approved):
One minor suggestion, otherwise LGTM. Thanks! 👍

test/mocha-fixtures.js
hasezoey: it is actually more correct the way it is, because MMS is set to starts a `wiredTiger` (disk stored) and not a `ephermeralForTest` (inmemory) database engine

AbdelrahmanHafez: "not" does not give so much information. Can we make that comment a little bit more elaborate?

hasezoey: it sounds descriptive to me, but sure, to me it reads `whether to (do thing) or not`

hasezoey: should be resolved with fddf23ed4903b7115d29e689553cebc517af5c84, is that better?

AbdelrahmanHafez (approved):

test/mocha-fixtures.js
AbdelrahmanHafez: LGTM, thanks. 👍 
It probably is clear to you and would be clear to me if I dig a bit more and understand more about the context, but I currently don't, and the comment's purpose is to make giving me context easier.

So I'm assuming that if we don't start an instance, we'll start something else. and "not" does not tell much about that thing.

PR #12522, opened 2022-10-04
fix(types): indicate that Schema.prototype.discriminator() returns `this`

### LINKED ISSUE #12457: Return-Types for `Schema.prototype.discriminator` do not align with runtime behavior

vkarpov15:
Fix #12457

**Summary**

#12457 pointed out that our types are slightly incorrect: `Schema.prototype.discriminator()` returns `this`, not the discriminated schema. Needed to change some tests which relied on the incorrect behavior. In cases like this where the TS types and runtime behavior disagree, we should assume the runtime behavior is more correct and fix the types, including any incorrect typing tests.

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -570,9 +570,7 @@ export type AutoTypedSchemaType = {
 // discriminator
 const eventSchema = new Schema<{ message: string }>({ message: String }, { discriminatorKey: 'kind' });
 const batchSchema = new Schema<{ name: string }>({ name: String }, { discriminatorKey: 'kind' });
-const discriminatedSchema = batchSchema.discriminator('event', eventSchema);
-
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM

Issue #12514, opened 2022-10-03
Jest mock timer warning without using fake timers


sspenst:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.6.0

### Node.js version

18.7.0

### MongoDB server version

6.5.4

### Description

Using `jest.spyOn` shows the following warning in the console:

(node:35252) [MONGOOSE] Warning: Mongoose: looks like you're trying to test a Mongoose app with Jest's mock timers enabled. Please make sure you read Mongoose's docs on configuring Jest to test Node.js apps: https://mongoosejs.com/docs/jest.html

### Steps to Reproduce

Create a test with the following:

`jest.spyOn(logger, 'error').mockImplementation(() => ({} as Logger));`

For additional context: https://github.com/sspenst/pathology/blob/main/tests/pages/api/avatar/avatar.test.ts#L23

### Expected Behavior

No warning should appear in the console.

hasezoey:
cannot reproduce, this must mean that mock timers are somehow enabled in your tests OR you are not using NodeJS (like deno) and that function somehow does not include `nextTick`

the line checking for `nextTick` is defined like:
https://github.com/Automattic/mongoose/blob/94b308ec89fce0feaa21454337e15351b6a71c01/lib/helpers/printJestWarning.js#L12

which i cant get to output anything other than 9 (`process.nextTick.toString().indexOf('nextTick')`) when both running in NodeJS REPL or in jest test without `enableMockTimers` set

> MongoDB server version 6.5.4

from what i can tell that version does not exist yet

sspenst:
Sorry I was looking at the wrong MongoDB version, I'm using `"node_modules/mongodb": { "version": "4.9.1", ...`

But thanks for the code reference, I was able to narrow down the root cause:

`const newrelicWinstonFormatter = newrelicFormatter(winston);`

Using [winston logger](https://github.com/winstonjs/winston) with this option causes the output of `process.nextTick.toString()` to be:

```
function wrappedProcess() {
  const segment = shim.getActiveSegment()
  if (!segment) {
    return fn.apply(this, arguments)
  }

  // Manual copy because helper methods add significant overhead in some usages
  const len = arguments.length
  const args = new Array(len)
  for (let i = 0; i < len; ++i) {
    args[i] = arguments[i]
  }

  shim.bindSegment(args, shim.FIRST, segment)

  return fn.apply(this, args)
[...]

PR #12524, opened 2022-10-05
Check if nextTick is set to fakeTimer function #12514

sspenst:
**Summary**

Jest mock timer warning is printed when `fakeTimers` are not enabled.

**Examples**

As per this comment https://github.com/Automattic/mongoose/issues/12514#issuecomment-1267965749, `newrelicWinstonFormatter` will no longer cause the Jest fake timer warning to print.

### DIFF
--- a/lib/helpers/printJestWarning.js
+++ b/lib/helpers/printJestWarning.js
@@ -9,7 +9,10 @@ if (typeof jest !== 'undefined' && typeof window !== 'undefined') {
     'https://mongoosejs.com/docs/jest.html');
 }
 
-if (typeof jest !== 'undefined' && process.nextTick.toString().indexOf('nextTick') === -1) {
+if (typeof jest !== 'undefined' && process.nextTick.toString() ===
+  'function () {\n' +
+  '                return clock[method].apply(clock, arguments);\n' +
+  '            }') {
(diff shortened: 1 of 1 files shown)


sspenst:
> the problem with checking against a specific definition is that it could change at any time without notice

Yea I wasn't sure how else to handle this. Would be nice if the fake nextTick function had a name that could be checked against eg `process.nextTick.toString().indexOf('fakeTimerNextTick') === -1)`

lib/helpers/printJestWarning.js
```diff
@@ -9,7 +9,10 @@ if (typeof jest !== 'undefined' && typeof window !== 'undefined') {
     'https://mongoosejs.com/docs/jest.html');
 }
 
-if (typeof jest !== 'undefined' && process.nextTick.toString().indexOf('nextTick') === -1) {
+if (typeof jest !== 'undefined' && process.nextTick.toString() ===
+  'function () {\n' +
```
vkarpov15, line 13: Unfortunately this check is way too brittle for us to merge. This would break even if Jest changes their internal code formatting a little bit.

We'll see if there is a way we can pull Jest's fake timer implementation without adding an extra `require()`.

vkarpov15:
Closing in favor of #12527

PR #12526, opened 2022-10-05
docs(models): add section on MongoDB Views

### LINKED ISSUE #5694: Is there a way we can handle views with mongoose !?
***feature***

currently there is a plugin, but that only selects few fields, and doesn't work on multiple models

https://docs.mongodb.com/manual/reference/method/db.createView/#db.createView

using mongoDB 3.4

vkarpov15:
Fix #5694

**Summary**

Right now there isn't any info on the docs on how to use Mongoose with MongoDB views. Admittedly I haven't used views in production, so I can't say I'm an expert on the subject. But this new docs section at least demonstrates how to get started with creating a new View in Mongoose.

I also replaced h2 with h1 and h3 with h2 in the models docs, because it doesn't make much sense for the main title of the page to be an h2.

**Examples**

### DIFF
--- a/docs/models.md
+++ b/docs/models.md
@@ -13,7 +13,7 @@ reading documents from the underlying MongoDB database.
 * [Change Streams](#change-streams)
 * [Views](#views)
 
-<h3 id="compiling"><a href="#compiling">Compiling your first model</a></h3>
+<h2 id="compiling"><a href="#compiling">Compiling your first model</a></h3>
 
 When you call `mongoose.model()` on a schema, Mongoose _compiles_ a model
 for you.
--- a/docs/models.md
+++ b/docs/models.md
@@ -1,4 +1,4 @@
-## Models
+# Models
 
 [Models](./api.html#model-js) are fancy constructors compiled from
 `Schema` definitions. An instance of a model is called a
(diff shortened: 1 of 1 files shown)


docs/models.md
```diff
@@ -11,6 +11,7 @@ reading documents from the underlying MongoDB database.
 * [Deleting](#deleting)
 * [Updating](#updating)
 * [Change Streams](#change-streams)
+* [Views](#views)
 
 <h3 id="compiling"><a href="#compiling">Compiling your first model</a></h3>
```
hasezoey, line 16:
```suggestion
<h2 id="compiling"><a href="#compiling">Compiling your first model</a></h3>
```

i think this was missed when changing to h2

hasezoey (approved):
LGTM

docs/models.md
AbdelrahmanHafez, line 16:
```suggestion
<h2 id="compiling"><a href="#compiling">Compiling your first model</a></h2>
```

vkarpov15: already fixed by @hasezoey . Thanks for catching this though.

hasezoey: actually, i just noticed that in my suggestions i did not change the closing tag

PR #12534, opened 2022-10-06
(fix): Throws error when updating a key name that match the discriminator key name on nested object

### LINKED ISSUE #12517: Throw when setting a key name that match the discriminator key name
### Prerequisites

### Mongoose version

6.6.3

### Node.js version

16

### MongoDB server version

4.9.1

### Description

Hello, I have a discriminator key to kind in the root doc  but I also have a subSchema which contains a field name kind.
When I am setting the sub field Kind, it throw an error.

### Steps to Reproduce

```javascript

mongoose.set("strict", "throw");

const options = { discriminatorKey: "kind"};

const EventSchema = new mongoose.Schema(
  { name: String, kind: String, animals: { kind: String, world: String } },
  options
);

const Event = mongoose.model("Event", EventSchema);

const ClickedLinkEvent = Event.discriminator(
  "ClickedLink",
  new mongoose.Schema({ url: String }, options)
);

async function main() {

  const eventDoc = await Event.findByIdAndUpdate(
    { _id: mongoose.Types.ObjectId() },
    {
      $set: {
        name: "name",
        animals: { kind: "kind", world: "world" },
      },
    },
    {
    }
  );

  console.log(eventDoc);
}

main();

```

### Expected Behavior

_No response_

lpizzinidev:
closes #12517

### DIFF
--- a/lib/helpers/query/castUpdate.js
+++ b/lib/helpers/query/castUpdate.js
@@ -196,7 +196,12 @@ function walkUpdatePath(schema, obj, op, options, context, filter, pref) {
       }
     }
 
-    if (schema.discriminatorMapping != null && key === schema.options.discriminatorKey && !options.overwriteDiscriminatorKey) {
+    if (
+      schema.discriminatorMapping != null &&
+      key === schema.options.discriminatorKey &&
+      !options.overwriteDiscriminatorKey &&
+      prefix === ''
+    ) {
       if (strictMode === 'throw') {
         const err = new Error('Can\'t modify discriminator key "' + key + '" on discriminator model');
         aggregatedError = _appendError(err, context, key, aggregatedError);
--- a/lib/helpers/query/castUpdate.js
+++ b/lib/helpers/query/castUpdate.js
@@ -196,15 +196,15 @@ function walkUpdatePath(schema, obj, op, options, context, filter, pref) {
       }
     }
 
+    const discriminatorKey = (prefix ? prefix + key : key);
     if (
       schema.discriminatorMapping != null &&
-      key === schema.options.discriminatorKey &&
-      !options.overwriteDiscriminatorKey &&
-      prefix === ''
+      discriminatorKey === schema.options.discriminatorKey &&
+      !options.overwriteDiscriminatorKey
     ) {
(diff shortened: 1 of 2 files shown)


abarriel:
Can a discriminator be set deeper  and not only on root? if it's the case we should handle that
`discriminatorKey: 'obj.objkind',`

lpizzinidev:
> Can a discriminator be set deeper and not only on root?

@abarriel This is the question.
I've been conservative and assumed that the discriminator may be set only on root properties.
The other approach, as you pointed out, would be to allow deeper discriminators, something like `discriminatorKey: 'obj.objkind'`.

I'll wait for more feedback and apply changes if needed.

lib/helpers/query/castUpdate.js
```diff
@@ -196,7 +196,12 @@ function walkUpdatePath(schema, obj, op, options, context, filter, pref) {
       }
     }
 
-    if (schema.discriminatorMapping != null && key === schema.options.discriminatorKey && !options.overwriteDiscriminatorKey) {
+    if (
+      schema.discriminatorMapping != null &&
+      key === schema.options.discriminatorKey &&
```
vkarpov15, line 201: Let's make this `(prefix ? prefix + '.' + key : key) === schema.options.discriminatorKey &&`

Although we don't have test coverage for nested discriminator keys, it seems to work at first glance. For example, the following script works as expected.

```javascript
'use strict';

mongoose.set('debug', true);

const { Schema } = mongoose;

run().catch(err => console.err(err));

async function run() {

  const schema = new Schema({ name: String, meta: { test: String } }, { discriminatorKey: 'meta.kind' });
  const Disc = Test.discriminator('Disc', Schema({ test: String }));

  const disc = new Disc({ name: 'test' });
  await disc.save();

  console.log(await Disc.findById(disc));
}
```

Given that, it would be great if this check didn't break on if discriminator key has a `.`.

vkarpov15 (changes requested):

lib/helpers/query/castUpdate.js
lpizzinidev: 👍 I'll work on this

PR #12540, opened 2022-10-08
feat(document): add `$timestamps()` method to set timestamps for `save()`, `bulkSave()`, and `insertMany()`

### LINKED ISSUE #12117: bulkSave should take document.$__.saveOptions.timestamps into account
### Prerequisites

### 🚀 Feature Proposal

re #12059 
Now we enabled the possibility to set `timestamps` option on all documents, it'd be nice if we can allow people to set `timestamps` option on each individual document, and send all the write operations at once.

### Motivation

To enable people to have more control over which documents get timestamps and which ones don't.

### Example

_No response_

vkarpov15:
Fix #12117

**Summary**

Quick feature for 6.7 to make working with timestamps and `bulkSave()` + `insertMany()` easier. Right now, there's no way to configure timestamps for individual documents when using `bulkSave()` and `insertMany()`. This PR adds a `$timestamps()` method that sets the timestamps option for all future `save()`, `bulkSave()`, and `insertMany()`.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -963,7 +963,7 @@ Document.prototype.$session = function $session(session) {
  *     await doc.save(); // Does **not** apply timestamps
  *
  * @param {Boolean} [value] overwrite the current session
- * @return {Document} this
+ * @return {Document|boolean|undefined} When used as a getter (no argument), a boolean will be returned indicating the timestamps option state or if unset "undefined" will be used, otherwise will return "this"
  * @method $timestamps
  * @api public
  * @memberOf Document
--- a/lib/document.js
+++ b/lib/document.js
@@ -949,6 +949,48 @@ Document.prototype.$session = function $session(session) {
 };
 
 /**
+ * Getter/setter around whether this document will apply timestamps by
+ * default when using `save()` and `bulkSave()`.
+ *
+ * #### Example:
+ *
+ *     const TestModel = mongoose.model('Test', new Schema({ name: String }, { timestamps: true }));
+ *     const doc = new TestModel({ name: 'John Smith' });
+ *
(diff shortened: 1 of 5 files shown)


lib/document.js
```diff
@@ -948,6 +948,48 @@ Document.prototype.$session = function $session(session) {
...
+ *     const TestModel = mongoose.model('Test', new Schema({ name: String }, { timestamps: true }));
+ *     const doc = new TestModel({ name: 'John Smith' });
+ *
+ *     doc.$timestamps(); // true
+ *
+ *     doc.$timestamps(false);
+ *     await doc.save(); // Does **not** apply timestamps
+ *
+ * @param {Boolean} [value] overwrite the current session
+ * @return {Document} this
```
hasezoey, line 966:
```suggestion
 * @return {Document|boolean|undefined} When used as a getter (no argument), a boolean will be returned indicating the timestamps option state or if unset "undefined" will be used, otherwise will return "this"
```

test/timestamps.test.js
```diff
@@ -626,6 +626,20 @@ describe('timestamps', function() {
       assert.strictEqual(cat.updatedAt, old);
     });
 
+    it('can skip with `$timestamps(false)` (gh-12117)', async function() {
+      const cat = await Cat.findOne();
+      const old = cat.updatedAt;
+
+      await delay(10);
```
hasezoey, line 633: is this 10ms? just asking to be sure that the test does not time-out

hasezoey (approved):
LGTM, just one documentation thing and a question

test/timestamps.test.js
vkarpov15: Yeah ms

PR #12542, opened 2022-10-08
feat(model): add `mergeHooks` option to `Model.discriminator()` to avoid duplicate hooks

### LINKED ISSUE #12472: Plugin hooks are cloned on `Model.discriminator`
### Prerequisites

### Mongoose version

6.6.1

### Node.js version

any

### MongoDB server version

any

### Description

Currently when having a plugin which adds hooks and using `discriminator` the plugin only gets run twice where as the hooks do not get merged correctly unless they are global functions

### Steps to Reproduce

```ts
// NodeJS: 18.8.0
// MongoDB: 5.0 (Docker)
// Typescript 4.8.3
import * as mongoose from 'mongoose'; // mongoose@6.6.1

function globalPlugin(schema: any) {
  console.log('PLUGIN');
  schema.pre('save', function nonGlobalHook() {});
}

const baseSchema = new mongoose.Schema({});
baseSchema.plugin(globalPlugin); // plugin gets once run here

const BaseModel = mongoose.model('Base', baseSchema);

const disSchema = new mongoose.Schema({}); // schema with no inheritance from "baseSchema"
baseSchema.plugin(globalPlugin); // and plugin gets once run here too

const DisModel = BaseModel.discriminator('Dis', disSchema);

console.log('TEST1', (BaseModel.schema as any).s.hooks._pres.get('save')); // only has 1 "nonGlobalHook"
console.log('TEST2', (DisModel.schema as any).s.hooks._pres.get('save')); // has 2 "nonGlobalHook"
```

when moving `nonGlobalHook` to the more global scope (outside of `globalPlugin`), it works correctly and only gets applied once (deduplicated?)

### Expected Behavior

The Plugin to only run twice (current behavior) and the hooks to either get merged correctly (only one occurrence) or not merged at all when not extend from original

---

i know this seems like it is intended behavior, but it seems that should be behavior that could be controlled by the user
(i noticed this because typegoose currently uses this approach of constructing a new schema for each class and not using a schema that is already in a model)

vkarpov15:
Fix #12472

**Summary**

There isn't a good way for Mongoose to dedupe hooks other than checking if you have two functions that are `===`. Deduping based on `toString()` would prone to false positives with certain design patterns.

So for cases where deduping is not going to work, like #12472, we now have a `mergeHooks` option to `Model.discriminator()`. `Model.discriminator(name, discriminatorSchema, { mergeHooks: false })` makes Mongoose take the discriminator schema's hooks.

**Examples**

### DIFF
--- a/lib/helpers/model/discriminator.js
+++ b/lib/helpers/model/discriminator.js
@@ -19,10 +19,12 @@ const CUSTOMIZABLE_DISCRIMINATOR_OPTIONS = {
  * ignore
  */
 
-module.exports = function discriminator(model, name, schema, tiedValue, applyPlugins) {
+module.exports = function discriminator(model, name, schema, tiedValue, applyPlugins, mergeHooks) {
   if (!(schema && schema.instanceOfSchema)) {
     throw new Error('You must pass a valid discriminator Schema');
   }
+
+  mergeHooks = mergeHooks == null ? true : mergeHooks;
 
   if (model.schema.discriminatorMapping &&
       !model.schema.discriminatorMapping.isRoot) {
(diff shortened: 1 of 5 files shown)


hasezoey (changes requested):
LGTM, but the new option is missing from the types

hasezoey (approved):
LGTM

PR #12549, opened 2022-10-13
feat(document): add $clone method

### LINKED ISSUE #11849: `Document.prototype.$clone()`
### Prerequisites

### 🚀 Feature Proposal

`doc.$clone()` should return a copy of `doc` with:

1. Deep clone of `_doc`
2. Deep clone of `$__`

So return a new doc with the same state (same `isNew`, populated state, change tracking state, etc.)

### Motivation

This has popped up a couple of times in user questions, for example: https://github.com/Automattic/mongoose/discussions/10788 . This has also popped up in internal work. Most recently, in #11442 `$clone()` would have been helpful because we could just create a clone of the document to handle that particular manually populated edge case, so we can ensure that we don't overwrite the document's existing `$__.wasPopulated`.

### Example

_No response_

lpizzinidev:
closes #11849 

Not sure if this is the best approach.
Anyway, open to feedback.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -4582,6 +4582,27 @@ Document.prototype.getChanges = function() {
   return changes;
 };
 
+/**
+ * Returns a copy of this document with a deep clone of `_doc` and `$__`.
+ *
+ * @return {Document} a copy of this document
+ * @api private
+ * @method $clone
+ * @memberOf Document
+ * @instance
+ */
+
+Document.prototype.$clone = function() {
+  const clonedDoc = Object.assign(Object.create(Object.getPrototypeOf(this)), this);
+  if (this._doc) {
+    clonedDoc._doc = clone(this._doc);
+  }
+  if (this.$__) {
+    clonedDoc.$__ = Object.assign(Object.create(Object.getPrototypeOf(this.$__)), this.$__);
+  }
+  return clonedDoc;
+};
+
 /*!
  * Module exports.
  */
--- a/lib/document.js
+++ b/lib/document.js
@@ -4595,9 +4595,6 @@ Document.prototype.getChanges = function() {
 Document.prototype.$clone = function() {
   const Model = this.constructor;
   const clonedDoc = new Model();
-  for (const p of Object.getOwnPropertyNames(this)) {
-    clonedDoc[p] = clone(this[p]);
-  }
   if (this._doc) {
     clonedDoc._doc = clone(this._doc);
   }
--- a/lib/document.js
+++ b/lib/document.js
@@ -4595,15 +4595,20 @@ Document.prototype.getChanges = function() {
 Document.prototype.$clone = function() {
   const Model = this.constructor;
   const clonedDoc = new Model();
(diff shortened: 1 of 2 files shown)


Uzlopak:
Well I dont know if we want this feature. So we need feedback from @vkarpov15 

But I know for sure, that we have to keep in mind that Object.assign does only a shallow Copy. So what happens with deep nested objects? We need tests for that case.

test/document.test.js
```diff
@@ -11928,6 +11928,20 @@ describe('document', function() {
...
+    const Test = db.model('Test', schema);
+
+    const item = await Test.create({ name: 'Mongoose' });
+
+    const doc = await Test.findById(item._id);
+    const clonedDoc = doc.$clone();
+
+    assert.deepEqual(doc, clonedDoc);
+    assert.deepEqual(doc._doc, clonedDoc._doc);
+    assert.deepEqual(doc.$__, clonedDoc.$__);
```
vkarpov15, line 11943: I'd like to add some extra test cases to make sure this `$clone()` method works. For example:

1. Modify a field in `clonedDoc`. Make sure that the field is modified in `clonedDoc`, but **not** in `doc`.
2. Save `clonedDoc`. Make sure that doesn't affect modified paths on `doc`.
3. Add a value that fails validation to `doc`. Run `await doc.validate()`. Then create a new cloned doc `clonedDoc = doc.$clone()`, fix the invalid value, run `await clonedDoc.validate()`. Make sure the path is still invalid in `clonedDoc`.
4. Set a session on `clonedDoc`, make sure that doesn't affect the original doc.

lib/document.js
```diff
@@ -4582,6 +4582,27 @@ Document.prototype.getChanges = function() {
...
+ * @instance
+ */
+
+Document.prototype.$clone = function() {
+  const clonedDoc = Object.assign(Object.create(Object.getPrototypeOf(this)), this);
+  if (this._doc) {
+    clonedDoc._doc = clone(this._doc);
+  }
+  if (this.$__) {
+    clonedDoc.$__ = Object.assign(Object.create(Object.getPrototypeOf(this.$__)), this.$__);
```
vkarpov15, line 4601: We would need to do a deep clone on `$__`, that may contain deeply nested properties.

lib/document.js
```diff
@@ -4582,6 +4582,27 @@ Document.prototype.getChanges = function() {
...
+ *
+ * @return {Document} a copy of this document
+ * @api private
+ * @method $clone
+ * @memberOf Document
+ * @instance
+ */
+
+Document.prototype.$clone = function() {
+  const clonedDoc = Object.assign(Object.create(Object.getPrototypeOf(this)), this);
```
vkarpov15, line 4596: I really don't like using `Object.create()` like this, because `Object.create()` skips running the `Document` constructor. I'd try just using `const Model = this.constructor; const clonedDoc = new Model();`

vkarpov15 (changes requested):

vkarpov15 (approved):
Thanks :+1: I'll merge this into 6.8

PR #12568, opened 2022-10-19
fix(aggregate): avoid adding extra `$match` stage if user manually set discriminator key to correct value in first pipeline stage

### LINKED ISSUE #12478: Model with discriminatorKey throws error in aggregation pipeline when performing a search
### Prerequisites

### Mongoose version

6.3.1

### Node.js version

14

### MongoDB server version

4.4.0

### Description

If you are trying to use the aggregation pipeline to search a Model that has a discriminatorKey, this error is thrown because the discriminatorKey's match phase is automatically injected into the pipeline before the search's match phase.

`$match with $text is only allowed as the first pipeline stage`

### Steps to Reproduce

```js
import { Schema, model } from 'mongoose';

// setup a parent schema with a discriminatorKey
const ParentSchema: Schema = new Schema(
	{
		title: String,
		description: String,
	},
	{
		discriminatorKey: 'partition',
	}
);

// create search index on the title and description fields
ParentSchema.index({
	partition: 1,
	title: 'text',
	description: 'text',
});

// create child schema to inherit from parent
const ChildSchema: Schema = new Schema(
	{
		type: String,
	},
	{
		discriminatorKey: 'partition',
	}
);

// create models
const Parent = model('Parent', ParentSchema);
Parent.discriminator('Child', ChildSchema);
const Child = model('Child');

(async function () {

	// insert a child document to search for
	await Child.create({
		title: 'Hello World',
		description: 'This is a test',
		type: 'Test',
	});

	// search for child document using the aggregation pipeline
	await Child.aggregate([
		{
			$match: {
				partition: 'Child',
				$text: {
					$search: 'test',
				},
			},
		},
	]);

	// MongoServerError: $match with $text is only allowed as the first pipeline stage
})();
```
Clearly I put $text in the first pipeline stage but mongoose injects its own discriminatorKey match into the pipeline ahead of mine and produces this in mongo:
```js
db.parents.aggregate([
	{ $match: { partition: 'Child' } },
	{ $match: { partition: 'Child', $text: { $search: 'test' } } },
]);
```
If I remove the partition param from my $match, then mongoose's $match is merged into mine and it works:
```js
Child.aggregate([
	{
		$match: {
			$text: {
				$search: 'test',
			},
		},
	},
])
```
Produces this in mongo...
```js
db.parents.aggregate([
	{ $match: { $text: { $search: 'test' }, partition: 'Child' } }
]);
```

### Expected Behavior

If the first pipeline stage is a $match, Mongoose should always merge its discriminatorKey params into the first stage instead of inserting an additional $match above it. If the first stage already contains a discriminatorKey param then the auto-generated params from Mongoose should be dropped.

vkarpov15:
Fix #12478

**Summary**

#12478 points out that `prepareDiscriminatorPipeline()` unnecessarily adds a `$match` stage with the discriminator key if the user already explicitly set the correct discriminator key in the initial `$match`.

**Examples**

### DIFF
--- a/lib/helpers/aggregate/prepareDiscriminatorPipeline.js
+++ b/lib/helpers/aggregate/prepareDiscriminatorPipeline.js
@@ -12,7 +12,9 @@ module.exports = function prepareDiscriminatorPipeline(pipeline, schema, prefix)
     // If the first pipeline stage is a match and it doesn't specify a `__t`
     // key, add the discriminator key to it. This allows for potential
     // aggregation query optimizations not to be disturbed by this feature.
-    if (originalPipeline[0] != null && originalPipeline[0].$match && !originalPipeline[0].$match[filterKey]) {
+    if (originalPipeline[0] != null &&
+        originalPipeline[0].$match &&
+        (originalPipeline[0].$match[filterKey] === undefined || originalPipeline[0].$match[filterKey] === discriminatorValue)) {
       originalPipeline[0].$match[filterKey] = discriminatorValue;
       // `originalPipeline` is a ref, so there's no need for
       // aggregate._pipeline = originalPipeline
--- a/test/helpers/aggregate.test.js
+++ b/test/helpers/aggregate.test.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const assert = require('assert');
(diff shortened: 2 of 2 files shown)


AbdelrahmanHafez (approved):
Nice.
LGTM, thanks! 👍

hasezoey (approved):
LGTM

PR #12569, opened 2022-10-19
fix(document): allow creating document with document array and top-level key named `schema`

### LINKED ISSUE #12480: Array validation getting failed because of the "schema" key inside the Schema
### Prerequisites

### Mongoose version

6.2.2

### Node.js version

16.14.2

### MongoDB server version

4.4.6

### Description

Hi Team,
I have observed a bug in the schema class/validator provided by mongoose. When a schema definition contains a key named: `schema` and another key containing an array of documents, the validation gets failed even with completely valid documents.

#### Step 1. Defining a schema that contains a key named `schema` and another key that embeds an array.
```javascript
const AuthorSchema = new Schema({
  fullName: { type: "String", required: true },
});

const BookSchema = new Schema({
  schema: { type: "String", required: true },
  title: { type: "String", required: true },
  authors: [AuthorSchema],
});

const Book = model("book", BookSchema);
```

#### Step 2. Insert a document in the collection
```javascript
  const book = await Book.create({
    schema: "design",
    authors: [{ fullName: "Sourabh Bagrecha" }],
    title: "The power of JavaScript",
  });
```

#### Step 3. Which throws the following error:
```bash
mongoose/node_modules/mongoose/lib/document.js:3055
    this.$__.validationError = new ValidationError(this);
                               ^

ValidationError: book validation failed: authors: Cast to Array failed for value "[ { fullName: 'Sourabh Bagrecha' } ]" (type Array) at path "authors" because of "TypeError"
    at model.Document.invalidate (/Users/sourabh/Work/repro/mongoose/node_modules/mongoose/lib/document.js:3055:32)
    ...call stack clipped...
    {
  errors: {
    authors: CastError: Cast to Array failed for value "[ { fullName: 'Sourabh Bagrecha' } ]" (type Array) at path "authors" because of "TypeError"
        at model.$set (/Users/sourabh/Work/repro/mongoose/node_modules/mongoose/lib/document.js:1417:9)
        ...call stack clipped...  
      {
      stringValue: `"[ { fullName: 'Sourabh Bagrecha' } ]"`,
      messageFormat: undefined,
      kind: 'Array',
      value: [ { fullName: 'Sourabh Bagrecha' } ],
      path: 'authors',
      reason: TypeError: doc.schema.path is not a function
          at new MongooseDocumentArray (/Users/sourabh/Work/repro/mongoose/node_modules/mongoose/lib/types/DocumentArray/index.js:60:47)
          ...call stack clipped...
      valueType: 'Array'
    }
  },
  _message: 'book validation failed'
}
```
Note that the validation error does not occur when:
- The field schema is removed
- The field schema is renamed to Schema (uppercase S)
- The array field is removed

The validation error occurs only when both the schema field AND the array field exist in the schema

#### Step 4. Now if we update the schema by uppercasing the key `Schema` from lowercased `schema`
```javascript
  const BookSchema = new Schema({
    Schema: { type: "String", required: true },
// ^^^^^^^
    title: { type: "String", required: true },
    authors: [AuthorSchema],
  });

  const book = await Book.create({
    Schema: "design",
// ^^^^^^^
    authors: [{ fullName: "Sourabh Bagrecha" }],
    title: "The power of JavaScript",
  });
```

#### Step 5. It will work completely fine if schema was spelled differently (e.g. with an uppercase S, i.e.: `Schema`)
```javascript
{
  "Schema": "design",
  "title": "The power of JavaScript",
  "authors": [
    {
      "fullName": "Sourabh Bagrecha",
      "_id": new ObjectId("633343a2396c4534d5a70e19")
    }
  ],
  "_id": new ObjectId("633343a2396c4534d5a70e18"),
  "__v": 0
}
```

### Expected Behavior

I understand that `schema` is a reserved keyword that can be used to fetch the document's schema.
image

But in that case, it should throw a warning(compile-time) while we are specifying the `schema` key in our schema definition. Also, the validation is only throwing the error, when we are specifying an embedded array of documents in our definition. In all the other cases, having the `schema` key in our definition doesn't throw any validation error at the time of insert/update.

The expected behavior IMO would be a consistent error irrespective of the embedded array.

vkarpov15:
Fix #12480

**Summary**

Looks like we missed a spot when we allowed using `schema` as a top-level key: `DocumentArray/index.js` still uses `doc.schema` rather than `doc.$__schema`

**Examples**

### DIFF
--- a/lib/types/DocumentArray/index.js
+++ b/lib/types/DocumentArray/index.js
@@ -57,7 +57,7 @@ function MongooseDocumentArray(values, path, doc) {
   // to make more proof against unusual node environments
   if (doc && doc instanceof Document) {
     internals[arrayParentSymbol] = doc;
-    internals[arraySchemaSymbol] = doc.schema.path(path);
+    internals[arraySchemaSymbol] = doc.$__schema.path(path);
 
     // `schema.path()` doesn't drill into nested arrays properly yet, see
     // gh-6398, gh-6602. This is a workaround because nested arrays are
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -11928,6 +11928,26 @@ describe('document', function() {
     assert.ok(rawDoc);
     assert.deepStrictEqual(rawDoc.tags, ['mongodb']);
   });
+
+  it('can create document with document array and top-level key named `schema` (gh-12480)', async function() {
+    const AuthorSchema = new Schema({
+      fullName: { type: 'String', required: true }
+    });
+
+    const BookSchema = new Schema({
+      schema: { type: 'String', required: true },
+      title: { type: 'String', required: true },
+      authors: [AuthorSchema]
+    }, { supressReservedKeysWarning: true });
+
+    const Book = db.model('Book', BookSchema);
+
+    await Book.create({
+      schema: 'design',
+      authors: [{ fullName: 'Sourabh Bagrecha' }],
+      title: 'The power of JavaScript'
+    });
+  });
 });
 
 describe('Check if instance function that is supplied in schema option is availabe', function() {


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

PR #12570, opened 2022-10-19
fix(cast): make schema-level strictQuery override schema-level strict for query filters

### LINKED ISSUE #12508: Option `strict` is used before Schema and global level `strictQuery`
### Prerequisites

### Mongoose version

6.63

### Node.js version

16

### MongoDB server version

4.9.1

### Description

Hello, when strict is set to throw, I cannot set strictQuery to anything else then throw.
On docs, it says
`However, strictQuery is tied to strict by default. This means that, by default, Mongoose will filter out query filter properties that are not in the schema.`
I cannot  untied strictQuery and Query globally on in shcema option, I have to use QueryOption.
This is the expected Behavior ? the doc is unclear.

### Steps to Reproduce

```js

mongoose.set("debug", true);
mongoose.set("strictQuery", false);
const options = {
  strict: "throw",
  strictQuery: false
};

const eventSchema = new mongoose.Schema({ time: Date }, options);
const Event = mongoose.model("Event", eventSchema);

async function run() {

  // 1. Not working
  await Event.find({ url: "google.com" },
}

run();

```

### Expected Behavior

Should be able to set strictQuery to anything else then throw in schema option or globally not only in query opt

vkarpov15:
Fix #12508

**Summary**

Looks like schema-level `strict` overrides schema-level `strictQuery`, which is definitely not something we want given that we want to switch `strictQuery` back to `false` by default.

**Examples**

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -268,8 +268,11 @@ module.exports = function cast(schema, obj, options, context) {
           options.strictQuery :
           'strict' in options ?
             options.strict :
-            'strict' in schema._userProvidedOptions ? schema._userProvidedOptions.strict :
-              schema.options.strictQuery;
+            'strictQuery' in schema._userProvidedOptions ?
+              schema._userProvidedOptions.strictQuery :
+              'strict' in schema._userProvidedOptions ?
+                schema._userProvidedOptions.strict :
+                schema.options.strictQuery;
         if (options.upsert && strict) {
           if (strict === 'throw') {
             throw new StrictModeError(path);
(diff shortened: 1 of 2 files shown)


lib/cast.js
```diff
@@ -268,8 +268,11 @@ module.exports = function cast(schema, obj, options, context) {
           options.strictQuery :
           'strict' in options ?
             options.strict :
-            'strict' in schema._userProvidedOptions ? schema._userProvidedOptions.strict :
-              schema.options.strictQuery;
+            'strictQuery' in schema._userProvidedOptions ?
```
AbdelrahmanHafez, line 271: Nitpick: This seems to be growing so much that it's not easy to understand it at all, I think we should refactor this to an external function that builds the strictQuery value.

AbdelrahmanHafez (changes requested):
One refactor suggestion, but otherwise LGTM. Thanks 👍

IslandRhythms (approved):
LGTM

PR #12578, opened 2022-10-21
fix(query): allow overwriting discriminator key with `overwriteDiscriminatorKey` if `strict: 'throw'`

### LINKED ISSUE #12513: StrictModeError: Throw when setting discriminator field
### Prerequisites

### Mongoose version

6.6.3

### Node.js version

16

### MongoDB server version

4.9.1

### Description

Hello, I have a throw error on kind, this is a working exemple, event when I set `overwriteDiscriminatorKey` to true

`StrictModeError: Field kind is not in schema and strict mode is set to throw.`

I think I found the issue [castUpdate!L47](https://github.com/Automattic/mongoose/blob/master/lib/helpers/query/castUpdate.js#L47) where overwriteDiscriminatorKey is check only after strict throw. 

**it only happens when you pass the updateObj directly as a second params, otherwise when we pass with `$set: updateData` it go through castUpdate.js!L47**

```js
if (schema.options.strict === 'throw' && obj.hasOwnProperty(schema.options.discriminatorKey)) {
    throw new StrictModeError(schema.options.discriminatorKey);
  } else if (!options.overwriteDiscriminatorKey) {
    delete obj[schema.options.discriminatorKey];
  }
```

### Steps to Reproduce

```js

mongoose.set("strict", "throw");

const options = { discriminatorKey: "kind" };
const EventSchema = new mongoose.Schema({ name: String }, options);
const Event = mongoose.model("Event", EventSchema);

async function main() {

  const eventDoc = await Event.findByIdAndUpdate(
    { _id: mongoose.Types.ObjectId() },
    {
        name: "name",
        kind: "kind",
    },
    {
      overwriteDiscriminatorKey: true,
    },
  );

  console.log(eventDoc);
}

main();

```

### Expected Behavior

_No response_

vkarpov15:
Fix #12513

**Summary**

A couple of quick fixes for how we handle `overwriteDiscriminatorKey`. `overwriteDiscriminatorKey` should overwrite `strict: 'throw'`.

**Examples**


IslandRhythms (approved):
LGTM

Issue #11943, opened 2022-06-17
Browser bundle contains NodeJS code that throws "Uncaught TypeError: Cannot read properties of undefined (reading 'split')"


bitflower:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.3.8

### Node.js version

16.14.0

### MongoDB server version

4.2.19

### Description

When bundling up `mongoose` for the browser this code is included from the file `dist/browser.umd.js`:

```javascript
e.nodeMajorVersion=parseInt(t.versions.node.split(".")[0],10);
```

It throws this error in Chrome: 

```
Uncaught TypeError: Cannot read properties of undefined (reading 'split')
```

My guess is that the code line above was never supposed to end up in the browser bundle because it is server-side node code.

### Steps to Reproduce

See the `browser.umd.js` file in the `dist` folder of `mongoose`.

### Expected Behavior

The `browser.umd.js` contains only browser-compatible code.

vkarpov15:
What are you using to bundle Mongoose? I confirmed this code doesn't appear using Mongoose's top-level `webpack.config.js`: https://github.com/Automattic/mongoose/blob/b66cdc62feb0ac735158ea4aee71cd9236481758/webpack.config.js .

It looks like the code you're referencing comes from [this file in js-bson](https://github.com/mongodb/js-bson/blob/96130212f1e675015c717d98e1d8470de3120375/etc/prepare.js), which is a post-install script for the MongoDB driver's bson parser.

bitflower:
I use StencilJS which uses Rollup in the background. So you say it depends on my bundler? I just thtought it is Mongoose specific because I found the problematic line in your bundle. But maybe rollup doesn‘t take the bundled browser bundle?

vkarpov15:
That's strange, we'll check to see if that code is in the v6.3.8 bundle. Running Webpack on current master branch doesn't result in that code in the bundle...

vkarpov15:
I'm unable to repro, below shows my terminal output from running `npm install mongoose@6.3.8` and searching for the code in question.

```
$ npm install mongoose@6.3.8

added 28 packages, and audited 29 packages in 4s

4 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
$ grep -r "nodeMajor" node_modules/mongoose/*
$ 

```

If this issue is still affecting you, please provide more detailed repro instructions.

[...]

PR #12584, opened 2022-10-24
fix: (5.x) make access to `process.versions` lazy

maciasello:
**Summary**

This PR (commit: 9396dbbd5962532eeebc7aa18840f1f4a5af66bf) fixes gh-11943 in 5.x branch. It makes access to `process.versions` lazy (in a function, not at module level) and thus does not crash mongoose loaded in a browser.

This PR (commit: f9146a304339283bbb45aca63ad82baf1ad5c5f5) also contains an *attempt* to implement a test for this issue. Unfortunately it does not achieve the main goal. I haven't been able to mock/replace value of `process.versions` to mimic what is available in the browser. Whatever is written to it (or its properties) is lost and the original values are always returned.
I am open for any suggestion on how to achieve that and can work on the PR until it is good enough. I have checked that the main change fixes the issue in our project.

Also it includes a fix (commit: c290eea2d6dc1e9ad5e7ae1b04fde926065e6d9d) to enable running `browser.test.js` in isolation (`npx mocha --exit ./test/browser.test.js`). Without the change the driver is not set before being accessed.
It doesn't fail when running all tests in batch, so it raises a question if driver objects set by tests do not interfere with results of other tests.

**Examples**

No examples needed. For the failure, just load the original `dist/browser.umd.js` into a browser.

### DIFF
--- a/lib/cursor/QueryCursor.js
+++ b/lib/cursor/QueryCursor.js
@@ -37,7 +37,7 @@ const utils = require('../../lib/utils');
 function QueryCursor(query, options) {
   const streamOpts = { objectMode: true };
   // for node < 12 we will emit 'close' event after 'end'
-  if (utils.nodeMajorVersion >= 12) {
+  if (utils.nodeMajorVersion() >= 12) {
     // set autoDestroy=true because on node 12 it's by default false
     // gh-10902 need autoDestroy to destroy correctly and emit 'close' event for node >= 12
     streamOpts.autoDestroy = true;
@@ -104,7 +104,7 @@ QueryCursor.prototype._read = function() {
           return _this.emit('error', error);
         }
         // for node >= 12 the autoDestroy will emit the 'close' event
-        if (utils.nodeMajorVersion < 12) {
+        if (utils.nodeMajorVersion() < 12) {
           _this.on('end', () => _this.emit('close'));
         }
       });
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -936,4 +936,6 @@ exports.errorToPOJO = function errorToPOJO(error) {
   return ret;
 };
 
-exports.nodeMajorVersion = parseInt(process.versions.node.split('.')[0], 10);
+exports.nodeMajorVersion = function nodeMajorVersion() {
+  return parseInt(process.versions.node.split('.')[0], 10);
+};
--- a/test/browser.test.js
+++ b/test/browser.test.js
@@ -20,6 +20,10 @@ describe('browser', function() {
     exec('node --eval "require(\'./lib/browser\')"', done);
   });
 
+  it('require() works in an environment where process.versions is an empty object (gh-5842)', function(done) {
+    exec('node --eval "Object.defineProperty(process, \'versions\', { value: {} }); require(\'./dist/browser.umd\')"', done);
+  });
+
   it('using schema (gh-7170)', function(done) {
     exec('node --eval "const mongoose = require(\'./lib/browser\'); new mongoose.Schema();"', done);
   });
--- a/test/browser.test.js
+++ b/test/browser.test.js
@@ -4,6 +4,8 @@
  * Module dependencies.
  */
 
+require('../lib/browser');
+
 const Document = require('../lib/browserDocument');
 const Schema = require('../lib/schema');
 const assert = require('assert');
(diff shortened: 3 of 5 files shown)


maciasello:
Not sure what is the workflow. Can I get a pair of eyes on this even though in draft still? I would greatly appreciate some hints on:
 - if/how to test the change (I tried a couple approaches, but failed -> see the description)
 - is 5.x still maintained (hopefully)
 CC @vkarpov15

lib/utils.js
```diff
@@ -936,4 +936,6 @@ exports.errorToPOJO = function errorToPOJO(error) {
   return ret;
 };
 
-exports.nodeMajorVersion = parseInt(process.versions.node.split('.')[0], 10);
+exports.nodeMajorVersion = function nodeMajorVersion() {
```
vkarpov15, line 939: Maybe we make `nodeMajorVersion()` return `null` if `process.versions` is undefined?

lib/cursor/QueryCursor.js
```diff
@@ -104,7 +104,7 @@ QueryCursor.prototype._read = function() {
           return _this.emit('error', error);
         }
         // for node >= 12 the autoDestroy will emit the 'close' event
-        if (utils.nodeMajorVersion < 12) {
+        if (utils.nodeMajorVersion() < 12) {
```
vkarpov15, line 107: If we make `nodeMajorVersion()` return `null` if `process.versions` is undefined, add a check: `const nodeMajorVersion = utils.nodeMajorVersion(); if (nodeMajorVersion != null && nodeMajorVersion < 12) { ... }`

test/browser.test.js
```diff
@@ -18,6 +19,10 @@ describe('browser', function() {
     exec('node --eval "require(\'./lib/browser\')"', done);
   });
 
+  it('require() works in an environment where process.versions is an empty object (gh-5842)', function(done) {
+    exec('node --eval "const process = require(\\"node:process\\"); process.versions = {}; require(\'./lib/browser\')"', done);
```
vkarpov15, line 23: Make this `require('process')`. Older node versions don't support `require('node:process')`

vkarpov15 (changes requested):
A couple of minor suggestions. Most important is to fix the tests, but would also be nice to handle the case when `process.versions.node` is undefined

lib/utils.js
maciasello: Sure! Can do

lib/cursor/QueryCursor.js
maciasello: Are we sure that in the browser scenario any of the variants (entering or not entering the branch) is correct? Shouldn't it be an error if the execution path reached this point in the browser? I understand we don't to crash too much, but this check may create a false appearance that it can still work.
However, to fully assess that I would need to understand if there is any scenario where QueryCursor can work in the browser.

test/browser.test.js
maciasello: Sure, will do. This test was just a placeholder as it didn't properly verify the issue.
I found that `Object.defineProperty(process, 'versions', { value: {} })` works for replacing the versions object. I pushed an updated version (still without the `null` handling described in the other comment).

maciasello: There is another issue connected to it.
If you do `require('./lib.browser.js')` with `process.versions` replaced with `{}` there is another piece of code that fails. It is located in one of the dependent libraries (`bluebird` required by `mquery`).
It seems to be removed by tree shaking done by webpack as it is not present in the final browser bundle.
Thus, I made the test load the bundle file (`./dist/browser.umd.js`) not `./lib/browser`. It may lead to additional quirks when testing locally as the file must be built before running tests.

lib/cursor/QueryCursor.js
maciasello: @vkarpov15 What do you think about that? :arrow_up:

lib/utils.js
maciasello: Waiting with this change until the convo below is resolved. :arrow_down:

maciasello:
@vkarpov15 I pushed a fix for the tests. Linter and tests should not fail anymore. Can you look at https://github.com/Automattic/mongoose/pull/12584#discussion_r1004286844 ?

lib/cursor/QueryCursor.js
vkarpov15: "Shouldn't it be an error if the execution path reached this point in the browser?" <-- sure, but we can say the same thing about a lot of the codebase. I'd prefer to avoid having individual snippets of code being aware about browser vs not browser, and instead rely on the browser build to determine what should or shouldn't be in the browser build. Ideally this code wouldn't be in the browser build at all, but that's a little harder to fix.

TLDR; let's do the simple, low-risk fix of checking if `nodeMajorVersion` is `null` here.

vkarpov15 (approved):
Got some test failures that we'll fix in 5.x branch, but otherwise good to go :+1:

PR #12585, opened 2022-10-24
feat(types/models): add missing create constructor signature override type

naorpeled:
**Summary**

While looking into the constructor signature overrides of the Mongoose's model `create` function, I've noticed that one of its overrides is missing. 
This PR adds it.


vkarpov15 (approved):
Thanks :+1:

Issue #12494, opened 2022-10-01
Using populate on schema with combined Map of Array<Schema> will not return Array in v6


lattam:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.6.3

### Node.js version

14.16.1

### MongoDB server version

4.4.3

### Description

Using populate on schema with "Map of Array<Schema>" will ignore the Array part and will return only one Schema object after upgrading from mongoose 5 to 6. 

### Steps to Reproduce

```js
const mongoose = require('mongoose');

const connection = mongoose.createConnection('mongodb://localhost:27037/test');

const User = new mongoose.Schema({
  name: String,
  addresses: {
    type: Map,
    of: [{
      type: mongoose.Schema.Types.ObjectId,
      ref:  'Address',
    }],
    default: {},
  },
});

const Address = new mongoose.Schema({
  city: String,
});

const UserModel = connection.model('User', User);
const AddressModel = connection.model('Address', Address);

async function init() {
  await UserModel.deleteMany({});
  await AddressModel.deleteMany({});

  const address = await new AddressModel({ city: 'London' }).save();

  await new UserModel({
    name: 'Name',
    addresses: {
      home: [address._id]
    }
  }).save();
}

async function test() {
  const query = UserModel.find({});

  query.populate({
    path: 'addresses.home',
  });

  const documents = await query.exec();

  console.log(JSON.stringify(documents));
  // v5.X.X outputs:
  // [{"_id":"63383ac599e8f5243db0cc8f","name":"Name", "addresses":{"home":[{"_id":"63383ac599e8f5243db0cc8d","city":"London"}]}}]

  // v6.X.X outputs:
  // [{"_id":"63383b503110bec5287a2ed7","name":"Name","addresses":{"home":{"_id":"63383b503110bec5287a2ed5","city":"London"}}}]
}

async function run() {
  await init();
  await test();
  connection.close();
}

run();
```

### Expected Behavior

Return the Array of sub elements, not just the first object.

chinmaykumbhare:
hi @lattam can you assign me this issue? I can fix it.

IslandRhythms:
```javascript
const mongoose = require('mongoose');

const connection = mongoose.createConnection('mongodb://localhost:27017/test');

const User = new mongoose.Schema({
  name: String,
  addresses: {
    type: Map,
    of: [{
[...]

PR #12601, opened 2022-10-26
Better support for populating maps of arrays of refs

vkarpov15:
**Summary**

#12494 shows that, in 5.x, you could use maps of arrays of refs. We don't currently have any test coverage for that case, so in 5.x it worked more by accident than by design. In 6.x we unexpectedly broke that. This PR gives us better support for populating maps of arrays of refs, including populating just one entry in the map.

**Examples**

### DIFF
--- a/test/types.map.test.js
+++ b/test/types.map.test.js
@@ -1071,7 +1071,7 @@ describe('Map', function() {
     const UserModel = db.model('User', User);
     const AddressModel = db.model('Address', Address);
 
-    const address = await new AddressModel({ city: 'London' }).save();
+    const address = await AddressModel.create({ city: 'London' });
 
     const { _id } = await new UserModel({
       name: 'Name',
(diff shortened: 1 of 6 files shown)


test/types.map.test.js
```diff
@@ -1050,4 +1050,56 @@ describe('Map', function() {
...
+    });
+
+    const Address = new mongoose.Schema({
+      city: String
+    });
+
+    const UserModel = db.model('User', User);
+    const AddressModel = db.model('Address', Address);
+
+    const address = await new AddressModel({ city: 'London' }).save();
```
hasezoey, line 1074:
```suggestion
    const address = await AddressModel.create({ city: 'London' });
```

test/types.map.test.js
```diff
@@ -1050,4 +1050,56 @@ describe('Map', function() {
...
+
+    const { _id } = await new UserModel({
+      name: 'Name',
+      addresses: {
+        home: [address._id]
+      }
+    }).save();
```
hasezoey, line 1081:
```suggestion
    const { _id } = await UserModel.create({
      name: 'Name',
      addresses: {
        home: [address._id]
      }
    });
```

hasezoey (approved):
LGTM, aside from just some style

PR #12613, opened 2022-10-29
feat(model): add `mergePlugins` option to `discriminator`

### LINKED ISSUE #12604: Option to disable `applyPlugins` for `discriminator` function
### Prerequisites

### Issue

(Continuation of #12472)

I had now tried #12542 in the project (typegoose) where i had first noticed the problem of hooks being duplicated (from plugins) which got resolved, but now i had noticed that the plugins are still force-fully copied over from the base schema instead of merging or being able to use the plugins applied to the discriminator model.

This is a proposal to add a option similar to `mergeHooks` to disable applying plugins from the base model (ie disabling the always `true` of `applyPlugins`), for example this new options could be named `mergePlugins` or `applyPlugins` (like the parameter name) and default to `true`

https://github.com/Automattic/mongoose/blob/bd034fa5395dba1c83f152913810a0394131a95e/lib/model.js#L1245

## Example

Current Behavior:

```ts
it('supports `mergeHooks` option to use the discriminator schema\'s hooks over the base schema\'s (gh-12472)', function() {
  let pluginTimes = 0;
  const shapeDef = { name: String };
  const shapeSchema = Schema(shapeDef, { discriminatorKey: 'kind' });
  shapeSchema.plugin(myPlugin, { opts1: true });

  const Shape = db.model('Test', shapeSchema);

  const triangleSchema = Schema({ ...shapeDef, sides: { type: Number, enum: [3] } });
  triangleSchema.plugin(myPlugin, { opts2: true });
  const Triangle = Shape.discriminator(
    'Triangle',
    triangleSchema
  );
  const squareSchema = Schema({ ...shapeDef, sides: { type: Number, enum: [4] } });
  squareSchema.plugin(myPlugin, { opts3: true });
  const Square = Shape.discriminator(
    'Square',
    squareSchema,
    { mergeHooks: false }
  );

  assert.equal(Triangle.schema.s.hooks._pres.get('save').filter(hook => hook.fn.name === 'testHook12472').length, 2); // correct
  assert.equal(Square.schema.s.hooks._pres.get('save').filter(hook => hook.fn.name === 'testHook12472').length, 1); // correct

  const squareFilteredPlugins = Square.schema.plugins.filter((obj) => obj.fn.name === 'myPlugin');
  assert.equal(squareFilteredPlugins.length, 1); // correct, should only have been applied once
  assert.equal(squareFilteredPlugins[0].opts['opts3'], true); // FAIL: because it force-applies plugins from the base, instead of the re-applied

  assert.equal(pluginTimes, 3); // correct, 3 times the plugin function is called

  function myPlugin(schema, opts) {
    pluginTimes += 1;
    schema.pre('save', function testHook12472() {});
  }
});
```

Example new Behavior:

```ts
it('supports `mergeHooks` option to use the discriminator schema\'s hooks over the base schema\'s (gh-12472)', function() {
  let pluginTimes = 0;
  const shapeDef = { name: String };
  const shapeSchema = Schema(shapeDef, { discriminatorKey: 'kind' });
  shapeSchema.plugin(myPlugin, { opts1: true });

  const Shape = db.model('Test', shapeSchema);

  const triangleSchema = Schema({ ...shapeDef, sides: { type: Number, enum: [3] } });
  triangleSchema.plugin(myPlugin, { opts2: true });
  const Triangle = Shape.discriminator(
    'Triangle',
    triangleSchema
  );
  const squareSchema = Schema({ ...shapeDef, sides: { type: Number, enum: [4] } });
  squareSchema.plugin(myPlugin, { opts3: true });
  const Square = Shape.discriminator(
    'Square',
    squareSchema,
    { mergeHooks: false, mergePlugins: false } // new option "mergePlugins", could also be named as "applyPlugins"
  );

  assert.equal(Triangle.schema.s.hooks._pres.get('save').filter(hook => hook.fn.name === 'testHook12472').length, 2); // correct
  assert.equal(Square.schema.s.hooks._pres.get('save').filter(hook => hook.fn.name === 'testHook12472').length, 1); // correct

  const squareFilteredPlugins = Square.schema.plugins.filter((obj) => obj.fn.name === 'myPlugin');
  assert.equal(squareFilteredPlugins.length, 1); // correct, should only have been applied once
  assert.equal(squareFilteredPlugins[0].opts['opts3'], true); // pass, because of using the discriminator's plugins instead of force-applied base plugins

  assert.equal(pluginTimes, 3); // correct, 3 times the plugin function is called

  function myPlugin(schema, opts) {
    pluginTimes += 1;
    schema.pre('save', function testHook12472() {});
  }
});
```

re #12473

lpizzinidev:
Fix #12604

### DIFF
--- a/lib/helpers/model/discriminator.js
+++ b/lib/helpers/model/discriminator.js
@@ -19,12 +19,13 @@ const CUSTOMIZABLE_DISCRIMINATOR_OPTIONS = {
  * ignore
  */
 
-module.exports = function discriminator(model, name, schema, tiedValue, applyPlugins, mergeHooks) {
+module.exports = function discriminator(model, name, schema, tiedValue, applyPlugins, mergeHooks, mergePlugins) {
   if (!(schema && schema.instanceOfSchema)) {
     throw new Error('You must pass a valid discriminator Schema');
   }
 
   mergeHooks = mergeHooks == null ? true : mergeHooks;
+  mergePlugins = mergePlugins == null ? true : mergePlugins;
 
   if (model.schema.discriminatorMapping &&
       !model.schema.discriminatorMapping.isRoot) {
@@ -34,7 +35,7 @@ module.exports = function discriminator(model, name, schema, tiedValue, applyPlu
 
   if (applyPlugins) {
     const applyPluginsToDiscriminators = get(model.base,
-      'options.applyPluginsToDiscriminators', false) || !mergeHooks;
+      'options.applyPluginsToDiscriminators', false) || !mergeHooks || !mergePlugins;
     // Even if `applyPluginsToDiscriminators` isn't set, we should still apply
     // global plugins to schemas embedded in the discriminator schema (gh-7370)
     model.base._applyPlugins(schema, {
@@ -185,7 +186,9 @@ module.exports = function discriminator(model, name, schema, tiedValue, applyPlu
       schema.s.hooks = model.schema.s.hooks.merge(schema.s.hooks);
     }
 
-    schema.plugins = Array.prototype.slice.call(baseSchema.plugins);
+    if (mergePlugins) {
+      schema.plugins = Array.prototype.slice.call(baseSchema.plugins);
+    }
     schema.callQueue = baseSchema.callQueue.concat(schema.callQueue);
     delete schema._requiredpaths; // reset just in case Schema#requiredPaths() was called on either schema
   }
--- a/lib/helpers/model/discriminator.js
+++ b/lib/helpers/model/discriminator.js
@@ -19,13 +19,12 @@ const CUSTOMIZABLE_DISCRIMINATOR_OPTIONS = {
  * ignore
  */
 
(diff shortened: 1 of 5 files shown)


lib/helpers/model/discriminator.js
```diff
@@ -34,7 +35,7 @@ module.exports = function discriminator(model, name, schema, tiedValue, applyPlu
 
   if (applyPlugins) {
     const applyPluginsToDiscriminators = get(model.base,
-      'options.applyPluginsToDiscriminators', false) || !mergeHooks;
+      'options.applyPluginsToDiscriminators', false) || !mergeHooks || !mergePlugins;
```
vkarpov15, line 38: I'm not a fan of how `mergePlugins` applies here. Seems a little strange that `mergePlugins` also effectively disables `applyPlugins`.

Maybe instead we reuse `applyPlugins` instead of adding a new option?

vkarpov15 (commented):
A minor comment. Also we'll have to merge this into 6.8, this is a new option

vkarpov15 (approved):
Thanks :+1:

PR #12616, opened 2022-10-31
fix(query): select Map field with `select: false` when explicitly requested

### LINKED ISSUE #12603: Select of a Map type with `select:false` not working as expected
### Prerequisites

### Mongoose version

6.6.2

### Node.js version

16.16.0

### MongoDB server version

5.0.3

_No response_

### Description

After version 6.6.2 the select with a `+` on a field of type `Map` with `select: false` does not seem to work as expected, it just does not return the element, i'm not sure but it may be related with this change:
- fix(query): handle select: false on map paths in query results [#12467](https://github.com/Automattic/mongoose/pull/12467)

### Steps to Reproduce

```js

async function main(){
  // Connect to a mongodb server
  await mongoose.connect((await MongoMemoryServer.create({ binary: { version: '5.0.3' } })).getUri());

  // Create the model
  const MyModel = mongoose.model('ModelName', new mongoose.Schema({
    title: String,
   // Create the type we are testing, a Map with `select:false` 
   body: {
      type: Map,
      of: { en: String, pt: String },
      select: false,
    }
  }));
  // Set the body
  await MyModel.create({
    title: "teste",
    body: {
      'A':{'en': 'en test A value', 'pt': 'pt test A value'},
      'B':{'en': 'en test B value', 'pt': 'pt test B value'}
    }
  })
  // Get the all items with the body field.
  console.log(await MyModel.find({}).select('+body').lean());
}

main();

```

### Expected Behavior

We expect the body to be returned:
```js
[
  {
    "_id": new ObjectId("635a471c7cf95eff198f4b2f"),
    title: 'teste',
    "body": {
      "A": {
        "en": "en test A value",
        "pt": "pt test A value"
      },
      "B": {
        "en": "en test B value",
        "pt": "pt test B value"
      }
    }
    __v: 0
  }
]
```

But on version >=6.6.2 it just returns the title:
```js
[
  {
    _id: new ObjectId("635a471c7cf95eff198f4b2f"),
    title: 'teste',
    __v: 0
  }
]
```

It seems that the issue requires a type `Map`, a select with a `+` and mongoose version >= 6.6.2
we can get the body by changing:
- mongoose to version 6.6.1 or lower
- use `type: Object`, instead of `Map`
- remove the `+` from the select (but  not what we want)

lpizzinidev:
fix #12603

**Summary**
Selecting a Map field with the flag `select` to `false` in the schema declaration using the `.select('+field')` syntax now correctly returns the field.

### DIFF
--- a/lib/queryhelpers.js
+++ b/lib/queryhelpers.js
@@ -232,7 +232,9 @@ exports.applyPaths = function applyPaths(fields, schema) {
     schema.eachPath(function(path, type) {
       if (prefix) path = prefix + '.' + path;
       if (type.$isSchemaMap || path.endsWith('.$*')) {
-        if (type.options && type.options.select === false) {
+        const plusPath = '+' + path;
+        const hasPlusPath = fields && plusPath in fields;
+        if (type.options && type.options.select === false && !hasPlusPath) {
           excluded.push(path);
         }
         return;
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -4133,6 +4133,34 @@ describe('Query', function() {
     assert.equal(item.doNotSelect, undefined);
   });
 
+  it('Map field with select: false is selected when explicitly requested (gh-12603)', async function() {
+    const testSchema = new mongoose.Schema({
+      title: String,
+      body: {
+        type: Map,
+        of: { en: String, pt: String },
+        select: false
+      }
+    });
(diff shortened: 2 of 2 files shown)


vkarpov15 (approved):
Great, thanks :+1:

Issue #12582, opened 2022-10-23
Documentation website does not respect selected mongoose version when clicking on links that go the api docs


kevo1ution:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

5.x.x

### Node.js version

n/a

### MongoDB server version

n/a

### Description

When navigating on the documentation site for mongoose v5, clicking on links that lead to `docs/api` ignore the fact that you have v5 selected.

This problem I think is similar to the issue here https://github.com/Automattic/mongoose/issues/12548

### Steps to Reproduce

1. navigate to https://mongoosejs.com/docs/5.x/docs/guide.html
2. click `Queries` under `Guides`.
3. Click any of the query links: for example, `Model.deleteMany()`

![mongoose_docs_bug_10_23_2022](https://user-images.githubusercontent.com/8290559/197419973-cd733b9e-9661-4a40-94e0-a4f8208b6232.gif)

### Expected Behavior

I expected v5 to stay selected. The workaround I had to do is change the url it links to, to have the base url `https://mongoosejs.com/docs/5.x/docs/`.
So `https://mongoosejs.com/docs/api.html#model_Model.deleteMany` becomes `https://mongoosejs.com/docs/5.x/docs/api.html#model_Model.deleteMany`. (This is mildly annoying)

PR #12623, opened 2022-11-01
fix(docs): links starting with "/docs/" on markdown doc files are relative links

kevo1ution:
**Summary**
- modifies various doc markdown files to use relative links prepended with `../`. This means that if the current link is `/docs/5.x/docs/` then resulting url will continue to be prepended with `/docs/5.x`. This prevents confusion when users click around in the docs for an older version.
- it makes a strong assumption that when a user is able to click on a `/docs/` link they are on a page that is prepended by either `/docs/<major-version>.x/docs` or just `/docs` when on the latest. From initially clicking around, I believe this is correct.

**Examples**
fixes: https://github.com/Automattic/mongoose/issues/12582


kevo1ution:
@vkarpov15 This is ready for review, is there a CI action that serves the changes?

hasezoey (changes requested):
most of it looks good to me, though some links just got `../` prepended which makes them relative, but some (before absolute) links in directories would now lead to wrong links

for example all (previously absolute) links in `docs/typescript/*` would not link to `../docs/typescript/*`, which would be wrong, it should be `../../` (or just directly relative link to the page instead of going all the way back)

kevo1ution:
> most of it looks good to me, though some links just got `../` prepended which makes them relative, but some (before absolute) links in directories would now lead to wrong links
> 
> for example all (previously absolute) links in `docs/typescript/*` would now link to `../docs/typescript/*`, which would be wrong, it should be `../../` (or just directly relative link to the page instead of going all the way back)

Good catch, I think this only relates to nested markdown files in the `/docs` folder which is just `/docs/typescript/*.md` and `/docs/tutorials/*.md`. My last two commits should fix that.

hasezoey (approved):
LGTM

@vkarpov15 should this change be backported to 5.x?

docs/change-streams.md
```diff
@@ -1,7 +1,7 @@
 ## Change Streams
 
 [Change streams](https://www.mongodb.com/developer/quickstart/nodejs-change-streams-triggers/) let you listen for updates to documents in a given model's collection, or even documents in an entire database.
-Unlike [middleware](/docs/middleware.html), change streams are a MongoDB server construct, which means they pick up changes from anywhere.
+Unlike [middleware](../docs/middleware.html), change streams are a MongoDB server construct, which means they pick up changes from anywhere.
```
vkarpov15, line 4: I may be missing something, but why not just do `./middleware.html` ? The `../docs` doesn't seem to be necessary.

docs/tutorials/dates.md
```diff
@@ -13,15 +13,15 @@ const userSchema = new mongoose.Schema({
 const User = mongoose.model('User', userSchema);
 ```
 
-When you create a user [document](/docs/documents.html), Mongoose will cast
+When you create a user [document](../../docs/documents.html), Mongoose will cast
```
vkarpov15, line 16: Ditto here, maybe just `../documents.html` instead of adding the extra `../docs`?

vkarpov15 (commented):
I like the idea of making links relative, and definitely like the idea of backporting this to 5.x. Just had some questions.

docs/change-streams.md
hasezoey: there were many places where this would apply, and i "overlooked" it because this PR is basically not modifying the link but making them relative from how they were before, maybe with #12645 we could change them to be more relative?

kevo1ution: Let me know what y'all want, I can make the related change

kevo1ution:
@vkarpov15 @hasezoey any conclusion on:
> I may be missing something, but why not just do ./middleware.html ? The ../docs doesn't seem to be necessary.

> there were many places where this would apply, and i "overlooked" it because this PR is basically not modifying the link but making them relative from how they were before, maybe with https://github.com/Automattic/mongoose/issues/12645 we could change them to be more relative?

I can make the related changes in this PR if you want (e.g. `../docs/middleware.html` => `./middleware.html`), just need to know what y'all want

hasezoey:
> I can make the related changes in this PR (e.g. ../docs/middleware.html => ./middleware.html), just need to know what y'all want

as for me, i dont "care" if it is this PR or a next one (this PR already has a focus and has done it already), and i am not sure on what style is wanted to be used for consistency, though i personally would be for the smallest relative instead of the long way

kevo1ution:
> as for me, i dont "care" if it is this PR or a next one (this PR already has a focus and has done it already), and i am not sure on what style is wanted to be used for consistency, though i personally would be for the smallest relative instead of the long way

Just added the relevant changes for the smallest relative links. I committed temporarily to the script that I used if you are curious. Let me know if there is anything else I should do; thanks!

docs/typescript/virtuals.md
```diff
@@ -1,6 +1,6 @@
 # Virtuals in TypeScript
 
-[Virtuals](/docs/tutorials/virtuals.html) are computed properties: you can access virtuals on hydrated Mongoose documents, but virtuals are **not** stored in MongoDB.
+[Virtuals](tutorials/virtuals.html) are computed properties: you can access virtuals on hydrated Mongoose documents, but virtuals are **not** stored in MongoDB.
```
hasezoey, line 3: i think that link should be with `../`

hasezoey (commented):
LGTM, though one link seems wrong

also, i dont know yet if this is the link style that is wanted (because #12645 did not yet get any comments from other maintainers)

kevo1ution:
> LGTM, though one link seems wrong
> 
> also, i dont know yet if this is the link style that is wanted (because #12645 did not yet get any comments from other maintainers)

Good catch; I decided to just go ahead and run my local mongoose docs site through a broken links checker to catch any other diffs I made that broke things. Using https://www.npmjs.com/package/broken-link-checker:
```bash
$ blc http://localhost:8089/ -roe | grep -v '/docs/5.x' | grep -v '/docs/4.x' | grep -v '/docs/jobs' | less
```

Fixed the last few links in previous commits. Also, I noticed there are many broken links under `docs/api/*` and `docs/api.html` paths in the production site that `blc` caught, I will comment on the issue https://github.com/Automattic/mongoose/issues/12645

docs/tutorials/virtuals.md
```diff
@@ -3,13 +3,15 @@
...
-* [Further Reading](#further-reading)
+- [Mongoose Virtuals](#mongoose-virtuals)
+  - [Your First Virtual](#your-first-virtual)
+  - [Virtual Setters](#virtual-setters)
+  - [Virtuals in JSON](#virtuals-in-json)
+  - [Virtuals in `console.log()`](#virtuals-in-consolelog)
+  - [Virtuals with Lean](#virtuals-with-lean)
+  - [Limitations](#limitations)
+  - [Populate](#populate)
+  - [Further Reading](#further-reading)
```
hasezoey, line 14: i dont think this TOC style change is wanted

docs/typescript/schemas.md
```diff
@@ -95,7 +95,7 @@ The second generic param, `M`, is the model used with the schema. Mongoose uses
 
 The third generic param, `TInstanceMethods` is used to add types for instance methods defined in the schema.
 
-The 4th param, `TQueryHelpers`, is used to add types for [chainable query helpers](/docs/typescript/query-helpers.html).
+The 4th param, `TQueryHelpers`, is `used to add types for [chainable query helpers](query-helpers.html).
```
hasezoey, line 98: why was this changed?

hasezoey (commented):
This time there seems to be some other changes than just link updates

docs/typescript/schemas.md
kevo1ution: this is the simplest relative link:
1. origin: `/docs/typescript/schemas.html`
2. relative link: `query-helpers.html`
3. becomes: `/docs/typescript/query-helpers.html`

OH accidentally fat-fingered a back tick here if that is what you are referring to

docs/tutorials/virtuals.md
kevo1ution: I don't know how this got in there, fixing right now

kevo1ution:
> This time there seems to be some other changes than just link updates

Apologies, I scanned through the diffs this time, and they all seem sensible to me

hasezoey (approved):
now it looks good to me again

@vkarpov15 @Uzlopak @AbdelrahmanHafez is this the URL style that we should go with: "no prefix when not needed"?

vkarpov15:
@hasezoey I agree that "no prefix when not needed" is a reasonable URL style. Certainly better than our `/docs/` links. I'll review this PR some more to make sure all the changes are ok, but at a high level this looks good.

vkarpov15 (approved):

kevo1ution:
Thanks for reviewing this!

PR #12649, opened 2022-11-02
fix(types): make array paths optional in inferred type of array default returns undefined

### LINKED ISSUE #12420: `mongoose.InferSchemaType` does not correctly infer nullable arrays

vkarpov15:
Fix #12420

**Summary**

Our types should treat setting `default` to a function that always returns `undefined` in the same way we treat setting `default: undefined`.

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -412,6 +412,7 @@ export function autoTypedSchema() {
     array5: any[];
     array6: string[];
     array7?: string[];
+    array8?: string[];
     decimal1?: Types.Decimal128;
     decimal2?: Types.Decimal128;
     decimal3?: Types.Decimal128;
@@ -458,6 +459,7 @@ export function autoTypedSchema() {
     array5: [],
     array6: { type: [String] },
     array7: { type: [String], default: undefined },
+    array8: { type: [String], default: () => undefined },
     decimal1: Schema.Types.Decimal128,
     decimal2: 'Decimal128',
     decimal3: 'decimal128'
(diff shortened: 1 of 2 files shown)


IslandRhythms (approved):
LGTM

hasezoey (commented):
is there already a types-test for `default: undefined`? if not, maybe one should also be added in addition to `default: () => undefined`

PR #12651, opened 2022-11-02
Add doc as second params to validation message function

IslandRhythms:


lib/error/validator.js
```diff
@@ -12,15 +12,16 @@ class ValidatorError extends MongooseError {
    * Schema validator error
    *
    * @param {Object} properties
+   * @param {Object} doc
```
vkarpov15, line 15: Object -> Document

test/schema.validation.test.js
```diff
@@ -1330,6 +1330,26 @@ describe('schema', function() {
       });
     });
 
+    it('Allows for doc to be passed as another parameter gh-12564', function(done) {
+      const s = mongoose.Schema({
+        n: {
+          type: String,
+          // required: true,
+          required: [true, function(properties, doc) {
```
vkarpov15, line 1338: Use a custom validator instead of `required`, so something like:

```
validate: {
  validator: function(v) { ... },
  message: function(properties, doc) { ... }
}
```

test/schema.validation.test.js
```diff
@@ -1330,6 +1330,26 @@ describe('schema', function() {
...
+            return 'fail ' + properties.path + ' on doc ' + doc;
+          }]
+        },
+        field: String
+      });
+      const M = mongoose.model('gh-12564', s);
+      const m = new M({ field: 'Yo' });
+
+      m.validate(function(error) {
+        assert.notEqual('fail n on doc undefined', error.errors['n'].message);
```
vkarpov15, line 1348: Why `undefined`?

vkarpov15 (changes requested):
Unfortunately it looks like your test asserts that your fix doesn't work :( let's take a look at this and see what's wrong.

vkarpov15 (approved):

PR #12657, opened 2022-11-04
Island rhythms/gh 6963 Can overwrite foreignField

IslandRhythms:


lib/model.js
```diff
@@ -4665,6 +4664,10 @@ function populate(model, docs, options, callback) {
       populateOptions.strictPopulate = model.base.options.strictPopulate;
     }
   }
+  const originalLocalField = model.schema.virtuals[populateOptions.path].options.localField;
```
hasezoey, line 4667: test failing: `cannot read "options" of undefined`, add a check that that field exists before trying to access it

lib/model.js
```diff
@@ -4675,12 +4678,13 @@ function populate(model, docs, options, callback) {
...
 
   const modelsMap = getModelsMapForPopulate(model, docs, populateOptions);
+  
   if (modelsMap instanceof MongooseError) {
     return immediate(function() {
       callback(modelsMap);
     });
   }
-
+  model.schema.virtuals[populateOptions.path].options.localField = originalLocalField;
```
hasezoey, line 4687: i think the same would be true here: add a check against "options" being undefined

hasezoey (changes requested):
Some things still need to be done:
- fix the tests (they seem to be consistently failing), the added checks dont check if the previous field is actually defined (see `options`)
- from what i can tell the options `localField` and `foreignField` are new to `populate`-options and have not been added to the jsdoc or types yet
- a `npm run lint --fix` would be good

hasezoey (approved):
LGTM now, though from what i can tell, the "new" options are not in the jsdoc or types

lib/model.js
```diff
@@ -4667,6 +4666,13 @@ function populate(model, docs, options, callback) {
       populateOptions.strictPopulate = model.base.options.strictPopulate;
     }
   }
+  let originalLocalField = '';
+  if (model.schema.virtuals[populateOptions.path]) {
+    originalLocalField = model.schema.virtuals[populateOptions.path].options.localField;
+  }
+  if (populateOptions.localField) {
+    model.schema.virtuals[populateOptions.path].options.localField = populateOptions.localField;
```
vkarpov15, line 4674: Modifying the virtual options is a bit of a red flag. This will likely mean that future `populate()` calls will use the new `localField` by default. We'll need to avoid doing this somehow.

vkarpov15 (commented):
Looks like this works, but has a few issues that we need to patch up. I'll take a look.

PR #12663, opened 2022-11-07
fix(types): set `this` to doc type in `SchemaType.prototype.validate()`

### LINKED ISSUE #12590: the key of this in Schema.path occur error with typecsrcipt
### Discussed in https://github.com/Automattic/mongoose/discussions/12557

<sup>Originally posted by **huge689** October 14, 2022</sup>
code:
```
UserSchema.path('hashed_password').validate(function (v) {
  if (this._password && this._password.length < 8) {
    this.invalidate('password', 'Password must be at least 8 characters.');
  }
});
```
error:
any
'this' implicitly has type 'any' because it does not have a type annotation.ts(2683)
user.model.ts(57, 45): An outer value of 'this' is shadowed by this container.
'this' implicitly has type 'any' because it does not have a type annotation.ts(2683)
user.model.ts(57, 45): An outer value of 'this' is shadowed by this container.
'this' implicitly has type 'any' because it does not have a type annotation.ts(2683)

</div>

vkarpov15:
Fix #12590

**Summary**

`SchemaType` should have a reference to the `DocType` so it knows that type to set `this` to in when the user does `schema.path('mypath').validate(function(v) { ... })`. With this PR, the following script compiles fine

```ts
import { Schema, SchemaType, InferSchemaType } from 'mongoose';
  
  _password: String
});

type User = InferSchemaType<typeof UserSchema>;

UserSchema.path('hashed_password').validate(function (v) {
  if (this._password && this._password.length < 8) {
    this.invalidate('password', 'Password must be at least 8 characters.');
  }
});

```

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -907,7 +907,7 @@ function gh12590() {
 
   expectType<SchemaType<User>>(UserSchema.path('hashed_password'));
 
-  UserSchema.path('hashed_password').validate(function(v: any) {
+  UserSchema.path('hashed_password').validate(function(v) {
     expectType<HydratedDocument<User>>(this);
     if (this._password && this._password.length < 8) {
       this.invalidate('password', 'Password must be at least 8 characters.');
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -897,3 +897,21 @@ function gh12562() {
     }
   );
 }
+
+function gh12590() {
+  const UserSchema = new Schema({
+    _password: String
+  });
+
+  type User = InferSchemaType<typeof UserSchema>;
+
+  expectType<SchemaType<User>>(UserSchema.path('hashed_password'));
+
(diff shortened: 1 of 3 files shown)


test/types/schema.test.ts
```diff
@@ -897,3 +897,21 @@ function gh12562() {
...
+function gh12590() {
+  const UserSchema = new Schema({
+    _password: String
+  });
+
+  type User = InferSchemaType<typeof UserSchema>;
+
+  expectType<SchemaType<User>>(UserSchema.path('hashed_password'));
+
+  UserSchema.path('hashed_password').validate(function(v: any) {
```
hasezoey, line 910: is this `v: any` required instead of just `v`?

hasezoey (commented):
LGTM, though with a question

Uzlopak (approved):

Issue #10773, opened 2021-09-23
v6: Change in subdocument not tracked using Object.assign / spread operator


saveman71:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

In some cases, change tracking in nested objects doesn't work.

**If the current behavior is a bug, please provide the steps to reproduce.**

```js
const mongoose = require('mongoose');
const Schema = mongoose.Schema;

const PaymentSchema = new Schema({
  status: String,
}, {
  _id: false,
});

const OrderSchema = new Schema({
  status: String,
  payments: {
    payout: PaymentSchema,
  },
})

async function run() {
  await mongoose.connect('mongodb://localhost:27017/test-repro');
  await mongoose.connection.dropDatabase();
  const Order = mongoose.model('Order', OrderSchema);

  const order = new Order({
    status: 'unpaid',
    payments: {
      payout: {
        status: 'unpaid',
      },
    }
  });

  await order.save();

  // It works with:
  // const newPaymentsStatus = {payout: {status: 'unpaid'}};

  // Object.assign or spread operator will break it
  const newPaymentsStatus = Object.assign({}, order.payments);

  order.status = 'paid';
  newPaymentsStatus.payout.status = 'paid';

  order.payments = newPaymentsStatus;
  console.log(order.isModified('payments'));

  await order.save().then(order => {
    // Pretends that the new object has the correct info
    console.log(order);
  });

  const notUpdated = await Order.findById({_id: order._id});

  console.log(notUpdated);
}

run();
```

**What is the expected behavior?**

Change is registered like it used to be in mongoose 5.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

MongoDB 3.4
Mongoose 6.0.6
NodeJS v14.17.6

saveman71:
So, as it used to be working in mongoose 5, I'm submitting this as a bug, but maybe this is too difficult to fix / is expected behaviour.

However, it's not so hard to end up with this kind of code and it's definitely unexpected behavior.

vkarpov15:
Thanks for reporting this issue and providing a detailed repro. This is an unfortunate bug in how Mongoose tracks saved state, and is related to our work on #9396 for 5.11 and #10596 for 6.0.

The idea of _saved state_ is to avoid unnecessary saves when you set a property to a different value, and then back to the initial value.

```javascript
const doc = await Model.findOne();

const originalValue = doc.myStr;
[...]

Issue #9396, opened 2020-09-07
isModified of a field is set to `true` even if value hasn't actually changed


adamreisnz:
**Do you want to request a *feature* or report a *bug*?**
Depends on whether this is expected behaviour or not. Possibly related to https://github.com/Automattic/mongoose/issues/4487

**Repro script**
```js
const mongoose = require('mongoose');
const {Schema} = mongoose;
const uri = `mongodb://localhost:27017/test`;

//Schema
const FooSchema = new Schema({
  bar: {
    type: String,
  },
});

//Pre-save hook 1
FooSchema.pre('save', function(next) {
  console.log('HOOK 1: BAR MODIFIED:', this.isModified('bar'), this.bar);
  this.bar = 'bar';
  next();
});

//Pre-save hook 1
FooSchema.pre('save', function(next) {
  console.log('HOOK 2: BAR MODIFIED:', this.isModified('bar'), this.bar);
  next();
});

//Model
const Foo = mongoose.model('Foo', FooSchema);

//Test function
async function test() {

  //Connect to database
  await mongoose.connect(uri);

  //Create new item
  const foo = new Foo({bar: 'bar'});
  await foo.save();

  foo.bar = 'notbar';
  await foo.save();
}

test();
```

**What is the current behavior?**
At the second save of the document, mongoose says `bar` is modified in the second hook, even though the first hook restored the value to what it was pre-save. So while the value was modified, it has not actually changed.

Output:

```
HOOK 1: BAR MODIFIED: true bar
HOOK 2: BAR MODIFIED: true bar
HOOK 1: BAR MODIFIED: true notbar
HOOK 2: BAR MODIFIED: true bar
```

**What is the expected behavior?**
Expect the following output:

```
HOOK 1: BAR MODIFIED: true bar
HOOK 2: BAR MODIFIED: true bar
HOOK 1: BAR MODIFIED: true notbar
HOOK 2: BAR MODIFIED: false bar <-- expect this to be false because between saved versions, bar has not changed
```

[...]

PR #12672, opened 2022-11-10
fix(document): handle setting array to itself after saving and pushing a new value

### LINKED ISSUE #12656: Data loss pushing items to list when sorting list and saving multiple times
### Prerequisites

### Mongoose version

6.7.0

### Node.js version

16.13.1

### MongoDB server version

4.2.8

4.8.4

### Description

The provided test case shows data loss pushing items to list/subcollection when sorting the subcollection AND saving multiple times.

If the `orderBy()` lines are omitted, the result is correct OR if the first `doc.save()` is omitted, the result is correct. 

This test case shows data loss, but a similar flow causes duplicate items to be inserted into the array on the 2nd save.

### Steps to Reproduce

```ts
import * as assert from 'assert'
import { orderBy } from 'lodash'
import { Schema } from 'mongoose'
import * as mongoose from 'mongoose'

run().catch(err => console.log(err))

/**
 * npx ts-node scripts/mongoose.ts
 */
async function run (): Promise<void> {
  await mongoose.connect('mongodb://localhost:27017/test-mongoose', {})
  await mongoose.connection.dropDatabase()
  mongoose.set('debug', true)

  const Test = mongoose.model('Test', new Schema({
    list: [{
      // _id: false,
      a: Number
    }]
  }))

  await Test.create({ list: [{ a: 1, b: 11 }] })

  let doc = await Test.findOne()
  doc.list.push({ a: 2 })
  doc.list = orderBy(doc.list, 'a', 'asc') // Same behavior if line replaced with: doc.list = [...doc.list]
  await doc.save()

  doc.list.push({ a: 3 })
  doc.list = orderBy(doc.list, 'a', 'asc') // Same behavior if line replaced with: doc.list = [...doc.list]
  await doc.save()

  doc = await Test.findOne()
  assert(doc.list.length === 3, `Got ${JSON.stringify(
    // @ts-expect-error toObject() not typed
    doc.list.toObject()
  )}`)

  process.exit()
}
```

Actual output. Note that the `{a: 3}` element is missing. If the first `doc.save()` is omitted the output is correct.
```
% npx ts-node scripts/mongoose.ts
Mongoose: tests.insertOne({ list: [ { a: 1, _id: new ObjectId("6365530081dd159228ce5c94") } ], _id: new ObjectId("6365530081dd159228ce5c93"), __v: 0}, { session: null })
Mongoose: tests.findOne({}, { projection: {} })
Mongoose: tests.updateOne({ _id: new ObjectId("6365530081dd159228ce5c93") }, { '$push': { list: { '$each': [ { a: 2, _id: new ObjectId("6365530081dd159228ce5c98") } ] } }, '$inc': { __v: 1 }}, { session: null })
Mongoose: tests.findOne({ _id: new ObjectId("6365530081dd159228ce5c93") }, { session: null, projection: { _id: 1 } })
Mongoose: tests.findOne({}, { projection: {} })
AssertionError [ERR_ASSERTION]: Got [{"a":1,"_id":"6365530081dd159228ce5c94"},{"a":2,"_id":"6365530081dd159228ce5c98"}]
    at /Users/johnpeberdy/Documents/Code/phenix/scripts/mongoose.ts:47:3
    at Generator.next (<anonymous>)
    at fulfilled (/Users/johnpeberdy/Documents/Code/phenix/scripts/mongoose.ts:5:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}

```

### Expected Behavior

The result should not change if the first `doc.save()` is omitted. If this use case is not supported, then it should fail loudly.

vkarpov15:
Fix #12656

**Summary**

We added some tracking of what the "saved state" of a document is in #9396 and #10773, but #12656 pointed out that we're not actually storing this saved state correctly when the first operation is a `push()`. Thankfully, #11644 makes this a bit easier: we can copy the state in `markModified()`.

Hopefully this is all we need, but we may want to consider tracking `savedState` by just cloning the document when the document is saved or loaded. That is easier from an implementation perspective, but bad from a performance perspective. I would prefer to not have to do that.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1468,13 +1468,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
       const doc = this.$isSubdocument ? this.ownerDocument() : this;
       savedState = doc.$__.savedState;
       savedStatePath = this.$isSubdocument ? this.$__.fullPath + '.' + path : path;
-      if (savedState != null) {
-        const firstDot = savedStatePath.indexOf('.');
-        const topLevelPath = firstDot === -1 ? savedStatePath : savedStatePath.slice(0, firstDot);
-        if (!savedState.hasOwnProperty(topLevelPath)) {
-          savedState[topLevelPath] = utils.clone(doc.$__getValue(topLevelPath));
-        }
-      }
+      doc.$__saveInitialState(savedStatePath);
     }
 
     this.$__set(pathToMark, path, options, constructing, parts, schema, val, priorVal);
@@ -1581,6 +1575,10 @@ Document.prototype.$__shouldModify = function(pathToMark, path, options, constru
     return false;
   }
   if (this.$isNew) {
+    return true;
+  }
+  // Is path already modified? If so, always modify. We may unmark modified later.
+  if (path in this.$__.activePaths.getStatePaths('modify')) {
     return true;
   }
 
@@ -1780,11 +1778,10 @@ Document.prototype.$inc = function $inc(path, val) {
 
   const currentValue = this.$__getValue(path) || 0;
 
-  this.$__setValue(path, currentValue + val);
-
   this.$__.primitiveAtomics = this.$__.primitiveAtomics || {};
   this.$__.primitiveAtomics[path] = { $inc: val };
   this.markModified(path);
+  this.$__setValue(path, currentValue + val);
 
   return this;
 };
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM

PR #12675, opened 2022-11-10
added deprecation warning for default `strictQuery`

### LINKED ISSUE #12666: Add deprecation warning for default `strictQuery`
### Prerequisites

### Issue

We're planning on switching `strictQuery` back to `false` in Mongoose 7, see #11861. We should add a deprecation warning to encourage devs to add `mongoose.set('strictQuery', false)` in their projects to prepare for this change.

lpizzinidev:
fix #12666

**Summary**
`strictQuery` will be switched back to `false` by default in Mongoose 7 so a warning message is shown on the database connection and on query execution if the `strictQuery` option is undefined and forced to true by the user.

Let me know if I understood the suggested logic correctly for the query warning message.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -32,6 +32,7 @@ const validateBeforeSave = require('./plugins/validateBeforeSave');
 
 const Aggregate = require('./aggregate');
 const PromiseProvider = require('./promise_provider');
+const printStrictQueryWarning = require('./helpers/printStrictQueryWarning');
 const shardingPlugin = require('./plugins/sharding');
 const trusted = require('./helpers/query/trusted').trusted;
 const sanitizeFilter = require('./helpers/query/sanitizeFilter');
@@ -409,9 +410,7 @@ Mongoose.prototype.connect = function(uri, options, callback) {
   const conn = _mongoose.connection;
 
   if (_mongoose.options.strictQuery === undefined) {
-    utils.warn('Mongoose: the `strictQuery` option will be switched back to `false` by default ' +
-              'in Mongoose 7. Use `mongoose.set(\'strictQuery\', false);` if you want to prepare ' +
-              'for this change. Or use `mongoose.set(\'strictQuery\', true);` to suppress this warning.');
+    printStrictQueryWarning();
   }
 
   return _mongoose._promiseOrCallback(callback, cb => {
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -409,6 +409,12 @@ Mongoose.prototype.connect = function(uri, options, callback) {
   const _mongoose = this instanceof Mongoose ? this : mongoose;
   const conn = _mongoose.connection;
 
+  if (_mongoose.options.strictQuery === undefined) {
+    utils.warn('Mongoose: the `strictQuery` option will be switched back to `false` by default ' +
+              'in Mongoose 7. Use `mongoose.set(\'strictQuery\', false);` if you want to prepare ' +
+              'for this change.');
```
vkarpov15, line 415: Add "Or use `mongoose.set('strictQuery', true)` to suppress this warning." so we can allow people to suppress this warning if necessary.

vkarpov15 (changes requested):

lpizzinidev:
@vkarpov15 
Some tests are failing due to timeout errors, probably because of the many warnings which are now displayed.
Should we add a temporary env variable to the test scripts (something like `NODE_ENV=test`) to prevent the warnings from being displayed during tests (adding a check in the `utils.warn` function)?

lib/index.js
```diff
@@ -409,6 +409,12 @@ Mongoose.prototype.connect = function(uri, options, callback) {
   const _mongoose = this instanceof Mongoose ? this : mongoose;
   const conn = _mongoose.connection;
 
+  if (_mongoose.options.strictQuery === undefined) {
+    utils.warn('Mongoose: the `strictQuery` option will be switched back to `false` by default ' +
```
vkarpov15, line 413: Let's add a `lib/helpers/printStrictQueryWarning.js` file, and make that file export a function that is wrapped in `[`util.deprecate()`](https://nodejs.org/api/util.html#utildeprecatefn-msg-code). Instead of manually printing using `util.warn()`. That will tie better into existing workflows for tracking deprecation warnings, and also make it so this warning isn't printed multiple times in tests.

vkarpov15 (changes requested):
Quick suggestion re: `util.warn()`.

lpizzinidev:
@vkarpov15 The PR is still not passing a test but I don't think it is a problem related to the implemented changes

lib/query.js
```diff
@@ -4889,6 +4890,15 @@ Query.prototype.exec = function exec(op, callback) {
   // Ensure that `exec()` is the first thing that shows up in
   // the stack when cast errors happen.
   const castError = new CastError();
+  const collection = this.mongooseCollection;
+
+  if (
```
vkarpov15, line 4895: Is this necessary? The `strictQuery` warning will print whenever you instantiate a connection, which (1) should always happen, and (2) is better behavior IMO - prints when process starts for most apps, as opposed to at runtime.

lpizzinidev: I agree that it is better to display the warning only on the connection.
It looks a bit redundant here unless there are cases to handle that I'm unaware of.
If it's ok for you I'll submit a commit with the `printStrictQueryWarning` only in `index.js`'s `connect` method.

vkarpov15 (approved):
One minor nit but otherwise LGTM, will fix that in 6.8 branch

lib/index.js
poovarasan11: working fine use mongoose.set() method

vvkyadav: yes, it works use mongoose.set() method

erikmstg: many thanks, its work

Issue #12676, opened 2022-11-10
diffIndexes() with a model with no documents in collection results in TypeError


rdeavilafloqast:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

5.13.15

### Node.js version

16.18.1

### MongoDB server version

4.4.0

### Use case:
We have a process to create indexes for our schemas instead of relying on `autoIndex`, and occasionally indexes need to be created for newly introduced schemas. As a result, the underlying collections don't actually exist yet. When running `diffIndexes()`( occasionally occurs for dry runs) with a model whose collection does not exist, the following error occurs:

>      Uncaught TypeError: indexes is not iterable
>       at /Users/richarddea/development/projects/mongoose/lib/model.js:1433:27
>       at /Users/richarddea/development/projects/mongoose/lib/model.js:5074:18
>       at processTicksAndRejections (internal/process/task_queues.js:75:11)

This is the result of [this line](https://github.com/Automattic/mongoose/blob/5.x/lib/model.js#L1433), with `indexes` being undefined. The `err` IS defined, and is of the following:

> MongoError: ns does not exist: mongoose_test.my-model
>     at MessageStream.messageHandler (/Users/richarddea/development/projects/mongoose/node_modules/mongodb/lib/cmap/connection.js:299:20)
>     at MessageStream.emit (events.js:315:20)
>     at processIncomingData (/Users/richarddea/development/projects/mongoose/node_modules/mongodb/lib/cmap/message_stream.js:144:12)
>     at MessageStream._write (/Users/richarddea/development/projects/mongoose/node_modules/mongodb/lib/cmap/message_stream.js:42:5)
>     at writeOrBuffer (internal/streams/writable.js:358:12)
>     at MessageStream.Writable.write (internal/streams/writable.js:303:10)
>     at Socket.ondata (internal/streams/readable.js:719:22)
>     at Socket.emit (events.js:315:20)
>     at addChunk (internal/streams/readable.js:309:12)
>     at readableAddChunk (internal/streams/readable.js:284:9)
>     at Socket.Readable.push (internal/streams/readable.js:223:10)
>     at TCP.onStreamRead (internal/stream_base_commons.js:188:23) {
>   ok: 0,
>   code: 26,
>   codeName: 'NamespaceNotFound'

### Fix
[...]

PR #12677, opened 2022-11-10
Fixing diffIndexes to cleanly return error instead of resulting in ty…

rdeavilafloqast:
**Summary**

Current behavior: Calling `diffIndexes()` on model whose collection doesn't exist results in a TypeError. See https://github.com/Automattic/mongoose/issues/12676#issue-1444579141

Use Case:
Dry runs of creating indexes. Occasionally calls from a newly introduced model whose collection doesn't yet exist because no documents haven been inserted. We need to return an error, if it exists, to the callback so invokers are able to gracefully handle the error.
**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1427,6 +1427,14 @@ Model.diffIndexes = function diffIndexes(options, callback) {
   return this.db.base._promiseOrCallback(callback, cb => {
     cb = this.$wrapCallback(cb);
     this.listIndexes((err, indexes) => {
+      if (err) {
+        // If the collection does not exist, it does not have indexes; safely proceed
+        if (err.codeName === 'NamespaceNotFound') {
+          indexes = [];
+        } else {
+          return cb(err, { toDrop, toCreate });
+        }
+      }
       const schemaIndexes = this.schema.indexes();
       // Iterate through the indexes created in mongodb and
       // compare against the indexes in the schema.
--- a/test/model.indexes.test.js
+++ b/test/model.indexes.test.js
@@ -70,6 +70,15 @@ describe('model', function() {
       });
     });
 
+    it('should have schema-defined indexes in toCreate when calling diffIndexes on model with no documents (gh-12676)', () => {
+      const MySchema = new Schema({
+        name: { type: String, index: true }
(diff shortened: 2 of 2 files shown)


test/model.indexes.test.js
```diff
@@ -70,6 +70,15 @@ describe('model', function() {
       });
     });
 
+    it('should result in a clean error when calling diffIndexes() against a collection that does not exist', async () => {
```
vkarpov15, line 73: Mongoose 5.x supports back to Node 4, so you can't use async functions in tests.

test/model.indexes.test.js
```diff
@@ -70,6 +70,15 @@ describe('model', function() {
...
+      const MySchema = new Schema({
+        name: { type: String, index: true }
+      });
+
+      const MyModel = db.model('my-model', MySchema); // This collection should not exist in DB
+      await assert.rejects(() => MyModel.diffIndexes());
```
vkarpov15, line 79: You should be able to remove the async function by making this `return MyModel.diffIndexes().then(() => assert.ok(false), () => {})`

lib/model.js
```diff
@@ -1427,6 +1427,9 @@ Model.diffIndexes = function diffIndexes(options, callback) {
   return this.db.base._promiseOrCallback(callback, cb => {
     cb = this.$wrapCallback(cb);
     this.listIndexes((err, indexes) => {
+      if (err) {
```
vkarpov15, line 1430: We definitely need an error check here, thanks for adding. But in the case of `NamespaceNotFound`, we know that there are no indexes in the db because the collection doesn't exist. So I think we should add a check `if (err.codeName === 'NamespaceNotFound') { indexes = []; } else { return cb(err); }`

vkarpov15 (changes requested):

lib/model.js
rdeavilafloqast: Perfect, that's exactly the behavior I'd expect when catching that error, makes sense to just implement it in the library itself. Will do!

vkarpov15 (approved):

PR #12680, opened 2022-11-11
added overloads for Schema.pre/post with different values for SchemaPreOptions

jpilgrim:
**Summary**

You can call the `updateOne` method with either the model class or a model instance (i.e. document) as receiver:

Depending on how you call it, different pre-hooks are called and `this` is set differently:
```
// never called:
userSchema.pre("updateOne", { query: false, document: false }, async function () {
    console.log(`"updateOne",  query: false, document: false, this: ${this.constructor.name}`);
});

// called with this = model
userSchema.pre("updateOne", { query: false, document: true }, async function () {
    console.log(`"updateOne",  query: false, document: true, this: ${this.constructor.name}`);
});

// called with this = Query
userSchema.pre("updateOne", { query: true, document: false }, async function () {
    console.log(`"updateOne",  query: true, document: false, this: ${this.constructor.name}`);
});

// called with this = Query or Model, depending on receiver
userSchema.pre("updateOne", { query: true, document: true }, async function () {
    console.log(`"updateOne",  query: true, document: true, this: ${this.constructor.name}`);
});
```

In mongoose version 6.7.2 a *breaking change* has been introduced (compared with 6.7.1. !) with commit [74af5f5b34cc7fcf631e1e323f18df36c74f6201](https://github.com/Automattic/mongoose/commit/74af5f5b34cc7fcf631e1e323f18df36c74f6201) (by Luca Pizzini) in types/index.d.ts, by moving the overloads for this=Query ahead of the overloads for this=Document (HydratedDocument).

**Examples**

The following code worked in 6.7.1
```
userSchema.pre("updateOne", { query: false, document: true }, async function () {
    if (this.isModified("password")) {
        // I know, this will never happen
    }
});
```

but it produces an error in 6.7.2: Property 'isModified' does not exist on type 'Query<any, any, {}, any>'.ts(2339)

In order to fix that problem, I would suggest to add more overloads, which do not use the general SchemaPreOptions type alias, but concrete values for document and query.


types/index.d.ts
```diff
@@ -295,8 +295,16 @@ declare module 'mongoose' {
     /** Defines a pre hook for the model. */
     pre<T = HydratedDocument<DocType, TInstanceMethods>>(method: 'save', fn: PreSaveMiddlewareFunction<T>): this;
     pre<T = HydratedDocument<DocType, TInstanceMethods>>(method: 'save', options: SchemaPreOptions, fn: PreSaveMiddlewareFunction<T>): this;
+    /**
+     * pre<T = Query<any, any>>(method: MongooseQueryMiddleware | MongooseQueryMiddleware[] | RegExp, options: SchemaPreOptions, fn: PreMiddlewareFunction<T>): this;
+     * with different combinations of SchemaPreOptions.
+     */
+    pre<T = HydratedDocument<DocType, TInstanceMethods>>(method: MongooseQueryMiddleware | MongooseQueryMiddleware[] | RegExp, options:  { document: true, query: false }, fn: PreMiddlewareFunction<T>): this;
```
vkarpov15, line 302: Make `method` explicitly list out the methods where `{ document: true, query: false }` works. So `'remove' | 'updateOne' | 'deleteOne' | 'validate'`. `{ document: true, query: false }` does nothing for `findOneAndUpdate` middleware, for instance.

types/index.d.ts
```diff
@@ -295,8 +295,16 @@ declare module 'mongoose' {
...
+    pre<T = Query<any, any>>(method: MongooseQueryMiddleware | MongooseQueryMiddleware[] | RegExp, options:  { document: false, query: true }, fn: PreMiddlewareFunction<T>): this;
+    pre<T = never>(method: MongooseQueryMiddleware | MongooseQueryMiddleware[] | RegExp, options:  { document: false, query: false }, fn: PreMiddlewareFunction<T>): this;
+    pre<T = Query<any, any>|HydratedDocument<DocType, TInstanceMethods>>(method: MongooseQueryMiddleware | MongooseQueryMiddleware[] | RegExp, options:  { document: true, query: true }, fn: PreMiddlewareFunction<T>): this;
```
vkarpov15, line 305: Ditto here, make `method: 'remove' | 'updateOne' | 'deleteOne' | 'validate'`

vkarpov15 (changes requested):
Couple of fixes. Also, please run `npm run lint -- --fix`

types/index.d.ts
jpilgrim: I have added new definition for the middleware method names in order to be able to better distinguish between document only methods, query only methods or ambiguous methods. This led to a long list of cases, and I hope I've got them right (and in the right order).

Note that I had to fix one failing test which wrongly assumed "deleteOnly" to be a query-only middleware method.

We also need to change post accordingly, but we should first clarify the situation with pre only (and copy the results to post later on. That said, I'm afraid that this won't be the last commit.

types/index.d.ts
jpilgrim: See above.

vkarpov15 (approved):
Thanks :+1:

test/types/models.test.ts
```diff
@@ -193,7 +193,9 @@ projectSchema.post('save', function() {
 });
 
 projectSchema.pre('deleteOne', function() {
```
vkarpov15, line 195: Actually, I took a closer look and this test is correct. `pre('deleteOne')` registers query middleware by default, not document middleware. `pre('deleteOne', { document: true })` would register document and query middleware. Can you please undo this change and see if you can get your changes to work without this change?

types/index.d.ts
vkarpov15: `pre('deleteOne')` registers query middleware, you need to explicitly call `pre('deleteOne', { document: true })` to opt in to getting document middleware for deleteOne. So the test is correct. See the following script:

```javascript
'use strict';

async function run() {

  const schema = mongoose.Schema({ name: String });
  schema.pre('deleteOne', function() {
    console.log('Is query', this instanceof mongoose.Query);
  });

  const doc = await Test.create({ name: 'foo' });

  console.log('Calling Document.prototype.deleteOne()');
  await doc.deleteOne(); // doesn't print

  console.log('Calling Model.deleteOne()');
  await Test.deleteOne(); // Prints "Is query true"
}

```

vkarpov15 (changes requested):
Took a closer look and have a couple of comments

types/index.d.ts
jpilgrim: Of course, you're right. It's so easy to mix things up here... I will fix that, add better annotations for post and, most importantly, will add some tests. Will need some days for that though.

jpilgrim: @vkarpov15 Are there any TypeScript tests which are actually executed? The tsd-tests are not really executed (but only type checked) -- they only check whether the **inferred** type matches an expected type.
I want to add TypeScript tests which check whether the **static** type matches the **dynamic** type. This is only possible when the test code is executed. I could write plain js tests, but then I won't see the type expectation.

jpilgrim:
Completely refactored my changes.
I added a tests with registers all kind of hooks and checks at runtime the this type. The result of this test is basically how I ended up with the definitions. Added pre and post definitions.

types/index.d.ts
jpilgrim: I added a plain JS test

jpilgrim:
Will have a look at the reported issues

jpilgrim:
I hopefully fixed the linter issues (I have a different formatter setting...).

I also added a new test, generated from the existing one, which actually tests the type annotations (and found some problems, which I fixed). Now we can sure that the type annotations is in sync with the real thing.

I assume that the tests which failed on Linux failed due to an older Node version? I removed the optional chaining construct.

I'm not quite sure about the ts-benchmark. What is actually measured and what can I do there? How can I check my specific parts? Also I'm not quite sure if the benchmark simply takes longer because there are more lines to parse?

vkarpov15:
@jpilgrim ts-benchmark measures the output of `tsc --extendedDiagnostics` on [our TypeScript benchmark projects](https://github.com/Automattic/mongoose/tree/master/benchmarks/typescript/simple). Right now we don't have very sophisticated TypeScript benchmarking, but we use this to help us avoid severe regressions because TypeScript compiler performance is extremely unpredictable.

jpilgrim:
@vkarpov15 I have (again) merged the latest master. There were conflicts I have solved (by deleting the changed parts since they were exactly the parts I have newly defined.

Are there any changes required now from my side?

I have no idea what to do about the failing ts benchmark test except increasing the expected time (or whatever is measured). Of course my changes will increase the time, as significantly more overloads are defined now -- which is necessary if you want to have correct definitions.

Github says "1 change requested" -- I think that has been resolved.

jpilgrim:
@vkarpov15 Any updates?

types/middlewares.d.ts
```diff
@@ -1,8 +1,16 @@
 declare module 'mongoose' {
 
-  type MongooseDocumentMiddleware = 'validate' | 'save' | 'remove' | 'updateOne' | 'deleteOne' | 'init';
-  type MongooseQueryMiddleware = 'count' | 'estimatedDocumentCount' | 'countDocuments' | 'deleteMany' | 'deleteOne' | 'distinct' | 'find' | 'findOne' | 'findOneAndDelete' | 'findOneAndRemove' | 'findOneAndReplace' | 'findOneAndUpdate' | 'remove' | 'replaceOne' | 'update' | 'updateOne' | 'updateMany';
-  type DocumentOrQueryMiddleware = 'updateOne' | 'deleteOne' | 'remove';
+  type MongooseQueryAndDocumentMiddleware = 'remove' | 'updateOne' | 'deleteOne';
```
vkarpov15, line 3: `validate` is another one that can be both Query and Document middleware

types/middlewares.d.ts
```diff
@@ -1,8 +1,16 @@
...
+
+  type MongooseDistinctDocumentMiddleware = 'validate' | 'save' | 'init';
+  type MongooseDefaultDocumentMiddleware = MongooseDistinctDocumentMiddleware | 'remove';
```
vkarpov15, line 6: | 'validate'

vkarpov15 (changes requested):
Sorry for taking so long to review this, we've been busy working on Mongoose 7. Have a couple of comments, but otherwise LGTM

types/middlewares.d.ts
jpilgrim: According to https://mongoosejs.com/docs/middleware.html, validate is only a document middleware.

types/middlewares.d.ts
jpilgrim: I assume this is the same issue again?

types/middlewares.d.ts
vkarpov15: We added a `Query.validate()` method. Mongoose calls this method when `runValidators` is set on a query, for example:

```javascript
'use strict';

schema.pre('validate', () => {
  console.log('Regular pre validate');
});
schema.pre('validate', { query: true, document: false }, () => {
  console.log('Pre query validate');
});

async function run() {

  const doc = new Test();
  await doc.validate();

  console.log('----');
  await Test.updateOne({}, { name: 'bar' }).setOptions({ runValidators: true });
}
```

types/middlewares.d.ts
vkarpov15: Yep same issue

jpilgrim:
@vkarpov15 I have merged the (your) latest master. There seems to be some methods to be removed:

- Model.update
- Document.remove

After removing these calls in my test, I got test failures since all pre/post-middlewares for `'remove'` and `'update'` are not called anymore. 

Does that mean that these middleware-strings (`'remove'` and `'update'`) are to be removed from types/middleware.d.ts as well, as they cannot be triggered anymore? And from https://mongoosejs.com/docs/middleware.html as well?

jpilgrim:
@vkarpov15 I assume update and remove have been removed. Thus I updated the tests and type definitions accordingly. As you might note, I had to fix another test.

From my point of view the PR should be ready to be merged. Tests are all passing.

However, I cannot run `npm run ts-benchmark`: "tarball tarball data for mongoose@file:../../../mongoose.tgz (null) seems to be corrupted."

vkarpov15 (approved):
LGTM, going to merge this into master to release with 7.1. I will make a couple of related changes in master: 1) bumping up the ts-benchmark limit slightly so ts-benchmark passes, 2) undo adding 'validate' to `MongooseQueryAndDocumentMiddleware`, because of the `if (!(this instanceof Query))` change in `document.test.ts`.

PR #12682, opened 2022-11-11
fix(document): Set Document.modelName property in constructor

### LINKED ISSUE #12669: Document.modelName type definition is string, but get `undefined` at runtime
### Prerequisites

### Mongoose version

6.7.0

### Node.js version

18.12.0

### MongoDB server version

4.4.17

4.8.4

### Description

Document.modelName type definition is string, but get `undefined` at runtime

### Steps to Reproduce

# Typescript Code
```typescript
const Model = mongoose.model('Test', new Schema());

const doc = new Model();
console.log('doc.modelName', doc.modelName);
```
# Output

```log
doc.modelName undefined
```

# Type Definition

```
class Document<T = any, TQueryHelpers = any, DocType = any> {
    /** The name of the model */
    modelName: string;
}
```

### Expected Behavior

Remove Document.modelName type definition

lpizzinidev:
fix #12669

### DIFF
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -190,9 +190,6 @@ declare module 'mongoose' {
     /** Returns the list of paths that have been modified. */
     modifiedPaths(options?: { includeChildren?: boolean }): Array<string>;
 
-    /** The name of the model */
-    modelName: string;
-
     /**
      * Overwrite all values in this document with the values of `obj`, except
      * for immutable properties. Behaves similarly to `set()`, except for it


vkarpov15 (commented):
I think the more correct fix here would be to remove `modelName` from the `Document` class in the TS definitions. This is a case where the TypeScript type defs are incorrect, and it would require a little more work to make our runtime behavior match the defined types.

vkarpov15 (approved):
Thanks :+1:

PR #12683, opened 2022-11-11
fix(schema): fix setupTimestamps for browser.umd

raphael-papazikas:
With the current browser.umd it‘s no longer possible to create schemas in the browser with `timestamps: true`. The following error is thrown: `Cannot read properties of null (reading 'get')`

Example to reproduce:
```

  { a: String },
  { timestamps: true }
);
```


vkarpov15 (approved):
Thanks :+1:

Issue #12529, opened 2022-10-05
Add `typesVersion` with minimum TypeScript version


elszczepano:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.6.5

### Node.js version

16.16.0

### MongoDB server version

6.0.2 (not required)

### Description

I've experienced an issue with mongoose above version `6.6.3`. When I try to build a project that uses mongoose in version `6.6.4` or higher I receive:

![Screenshot 2022-10-05 at 20 58 43](https://user-images.githubusercontent.com/23155008/194141352-b51afc3c-cff6-4978-be7c-e913693ec3f7.png)

When I switched to `6.6.3` the issue does not occur. My TS config:
```
{
  "compilerOptions": {
    "outDir": "./dist",
    "rootDir": "./src",
    "esModuleInterop": true,
    "lib": [
      "es2021",
      "dom"
    ],
    "module": "commonjs",
    "target": "es2021",
    "moduleResolution": "node",
    "allowJs": true,
    "noImplicitAny": true,
    "composite": true,
    "skipLibCheck": true,
    "sourceMap": true,
    "strictNullChecks": true,
    "experimentalDecorators": true,
    "forceConsistentCasingInFileNames": true,
    "include": [
        "src/**/*"
    ]
  }
}
```

As I briefly checked, the pointed places seem to be syntactically correct so the issue is a bit awkward for me. I'd be grateful for a hint if an issue is on my side.

### Steps to Reproduce

Just run `tsc --build` for a project that uses `mongoose`.

### Expected Behavior

The project is built correctly.

vkarpov15:
@elszczepano what version of TypeScript? I'm able to run `tsc --build` with a slightly fixed version of your `tsconfig.json` with TypeScript 4.8.4.

Maybe because your `tsconfig.json` is invalid? You're missing a `,` after `"esModuleInterop": true` and `include` is not a valid `compilerOptions` property.

ultrarunner:
I have the exact issue as @elszczepano with typescript 4.3.4, node 14.19.

elszczepano:
> Maybe because your tsconfig.json is invalid? You're missing a , after "esModuleInterop": true and include is not a valid compilerOptions property.

Even after fixing the tsconfig the issue still occurs in my case. Regarding the TS version, I use `4.6.3`.

kYem:
[...]

PR #12689, opened 2022-11-14
types: typeof Query results into invalid syntax of query.d.ts file

### LINKED ISSUE #12688: ts bug: toConstructor(): typeof Query<ResultType, DocType, THelpers, RawDocType>;
### Prerequisites

### Mongoose version

6.7.2

### Node.js version

16.x

### MongoDB server version

5.x

4.8.4

### Description

somehow this fails now suddenly:
../../node_modules/mongoose/types/query.d.ts(619,34): error TS1144: '{' or ';' expected.
../../node_modules/mongoose/types/query.d.ts(619,45): error TS1005: '>' expected.
../../node_modules/mongoose/types/query.d.ts(619,77): error TS1109: Expression expected.
../../node_modules/mongoose/types/query.d.ts(622,19): error TS1109: Expression expected.
../../node_modules/mongoose/types/query.d.ts(622,41): error TS1109: Expression expected.
../../node_modules/mongoose/types/query.d.ts(622,50): error TS1109: Expression expected.
../../node_modules/mongoose/types/query.d.ts(622,73): error TS1005: '(' expected.
../../node_modules/mongoose/types/query.d.ts(622,114): error TS1109: Expression expected.
../../node_modules/mongoose/types/query.d.ts(622,138): error TS1005: '(' expected.

the change was made already some weeks ago, but the line is definitly wrong:

toConstructor(): typeof Query<ResultType, DocType, THelpers, RawDocType>;

Query is already a Class and therefore typeof should just get removed.

image

change line 619 to 
`toConstructor(): Query<ResultType, DocType, THelpers, RawDocType>;`

fixes it

### Steps to Reproduce

just open the query.d.ts file in your favourite IDE :)

### Expected Behavior

_No response_

simllll:
**Summary**

right now I cannot compile my proecjt, due to https://github.com/Automattic/mongoose/issues/12688

image

### DIFF
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -616,7 +616,7 @@ declare module 'mongoose' {
     then: Promise<ResultType>['then'];
 
     /** Converts this query to a customized, reusable query constructor with all arguments and options retained. */
-    toConstructor(): typeof Query<ResultType, DocType, THelpers, RawDocType>;
+    toConstructor(): Query<ResultType, DocType, THelpers, RawDocType>;
 
     /** Declare and/or execute this query as an update() operation. */
     update(filter?: FilterQuery<DocType>, update?: UpdateQuery<DocType> | UpdateWithAggregationPipeline, options?: QueryOptions<DocType> | null, callback?: Callback<UpdateWriteOpResult>): QueryWithHelpers<UpdateWriteOpResult, DocType, THelpers, RawDocType>;


simllll:
@vkarpov15 we are experienceing this issue on more and more projects... please consider this fix

hasezoey:
as i had already noted in https://github.com/Automattic/mongoose/issues/12529#issuecomment-1272516908, `typeof Query` is the correct type, where as just `Query` would be the wrong type.
This is because by default in typescript a type of a class without `typeof` refers to *a instance of the class*, where as with `typeof` it refers to the *constructor / static of a class*

vkarpov15 (changes requested):
It looks like in #12688 the issue is that you're using an older version of TypeScript, v4.6.x. Furthermore, the original issue mentions that this issue shows up in your IDE, not when you actually compile with TypeScript.

Also, as @hasezoey mentioned, the change you proposed is incorrect. We'll investigate and see if there is a better workaround for "this function returns this class" rather than "this function returns an instance of this class" that is compatible with TypeScript 4.6.

`typeof this` works in TypeScript 4.6, but it doesn't work in TypeScript < 4.4. On the other hand, `typeof Query<ResultType, DocType, THelpers, RawDocType>;` works fine in TypeScript < 4.4 but not TypeScript 4.6 :facepalm:

vkarpov15:
Related https://github.com/Automattic/mongoose/commit/86ca62894f5e5a528f924910acabc3d85483ba9c

simllll:
I will investigate on my site, as we actually use TS 4.8... something seems to install an old TS Version.. anyways, the issue is like you said and my fix is wrong, you can close my PR.

Issue #12684, opened 2022-11-11
Move lean({ virtuals: true }) support from mongoose-lean-virtuals into core Mongoose


Jokero:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.7.2

### Node.js version

18.12.0

### MongoDB server version

N/A

### Typescript version (if applicable)

4.8.4

### Description

`InferSchemaType` does not infer types for virtual fields when schema has any. According to this PR https://github.com/Automattic/mongoose/pull/11908, it should work.

### Steps to Reproduce

1) Copy example from mentioned PR https://github.com/Automattic/mongoose/pull/11908/files#diff-fc5a8155d58dc5904b33fcc871f11ee08cb6f6e7adbe69e7bfbd7d173e82d8d2 (`Automatically Inferred Types`), create a type with `InferSchemaType` and try to access `fullName`:
```ts
import { Schema, InferSchemaType } from 'mongoose';

const schema = new Schema(
    {
        firstName: String,
        lastName: String,
    },
    {
        virtuals:{
            fullName:{
                get(){
                    return `${this.firstName} ${this.lastName}`;
                }
                // virtual setter and options can be defined here as well.
            }
        }
    }
);

type User = InferSchemaType<typeof schema>;

const user: User = {};
user.fullName
```
2) You will get the following error: 
```
TS2339: Property 'fullName' does not exist on type '{ firstName?: string | undefined; lastName?: string | undefined; }'.
```

### Expected Behavior

Type `User` should contain `fullName: string` field, and `user.fullName` shouldn't fail TS compilation.

carlosingles:
As a work around, you can use the `ObtainSchemaGeneric` type to add the virtual types:

```ts
import { Schema, InferSchemaType, ObtainSchemaGeneric } from 'mongoose';
...
type User = InferSchemaType<typeof schema> & ObtainSchemaGeneric<typeof schema, 'TVirtuals'>;

const user = {} as User;
user.fullName // works
```

Jokero:
@carlosingles, thanks for the workaround, currently I hardcoded all virtuals in an interface. 

@vkarpov15, I've tested the example from my initial message in 6.7.4, and the issue still persists (`fullName` is not available in a type returned by`InferSchemaType`).

If speak generally, I think it makes sense to have a way to get both interfaces (with and without virtuals). Based on first message example:

[...]

PR #11908, opened 2022-06-07
[✔️] Auto-typed-virtuals

mohammad0-0ahmad:
Related to auto-typed mongoose.
https://github.com/Automattic/mongoose/pull/11563

PR #12702, opened 2022-11-18
Feature/enhance auto typed virtuals

carlosingles:
**Summary**

Further enhances existing virtual type inference originally from: https://github.com/Automattic/mongoose/pull/11908

Virtual properties are not available on model methods such as `findOne`, this PR adds that functionality across all model methods.

Also resolves: https://github.com/Automattic/mongoose/issues/12684

### DIFF
--- a/test/types/virtuals.test.ts
+++ b/test/types/virtuals.test.ts
@@ -126,5 +126,10 @@ function autoTypedVirtuals() {
     }
   });
 
+  type TestDoc = InferSchemaType<typeof testSchema>;
+
+  const testDoc = { email: 'some email' } as TestDoc;
+  expectType<string>(testDoc.domain);
+
   expectType<FlatRecord<AutoTypedSchemaType & VirtualsType >>({} as InferredDocType);
 }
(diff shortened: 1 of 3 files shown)


carlosingles:
apologies for the ping, but i updated the branch to fix the failing test. if there's any further feedback i'll get to it right away 🙂

test/types/virtuals.test.ts
```diff
@@ -118,5 +118,18 @@ function autoTypedVirtuals() {
   const doc = new TestModel();
   expectType<string>(doc.domain);
 
-  expectType<FlatRecord<AutoTypedSchemaType & VirtualsType >>({} as InferredDocType);
+  TestModel.findOne({}).then((doc) => {
+    if (doc) {
+      expectType<string>(doc.domain);
+      const json = doc.toJSON();
+      expectType<string>(json.domain);
```
vkarpov15, line 125: This is incorrect. `domain` shouldn't be in `json` unless you have `toJSON: { virtuals: true }` in your schema or globally.

vkarpov15 (commented):
We've confirmed that the reported issue is a bug, but there are some test failures and some issues with the test. In general, I'd like to avoid changing TypeScript tests to make them pass, because that may be hiding breaking changes.

We'll take a look and see if there's an alternative fix we can come up with.

vkarpov15:
I'll close this in favor of #12720 for now. We can re-open if it turns out that #12720 is incorrect.

carlosingles:
Thanks for the response and thanks for the fix, makes sense 🙂

PR #12704, opened 2022-11-19
feat(debug): allow setting debug on a per-connection basis

### LINKED ISSUE #12700: Allow setting `debug` on a per-connection basis
### Prerequisites

### Issue

Right now you can only enable/disable debugging globally. Would be great to do `conn.set('debug', true)` to just enable debugging for a certain connection.

lpizzinidev:
fix #12700

### DIFF
--- a/lib/drivers/node-mongodb-native/collection.js
+++ b/lib/drivers/node-mongodb-native/collection.js
@@ -77,11 +77,16 @@ function iter(i) {
     const collection = this.collection;
     const args = Array.from(arguments);
     const _this = this;
-    const debug = _this &&
+    const hasGlobalDebug = _this &&
       _this.conn &&
       _this.conn.base &&
       _this.conn.base.options &&
       _this.conn.base.options.debug;
+    const hasConnectionDebug = _this &&
+      _this.conn &&
+      _this.conn.options &&
+      _this.conn.options.debug;
+    const debug = hasGlobalDebug || hasConnectionDebug;
     const lastArg = arguments[arguments.length - 1];
     const opId = new ObjectId();
 
--- a/lib/drivers/node-mongodb-native/collection.js
+++ b/lib/drivers/node-mongodb-native/collection.js
@@ -77,16 +77,16 @@ function iter(i) {
     const collection = this.collection;
     const args = Array.from(arguments);
     const _this = this;
-    const hasGlobalDebug = _this &&
+    const globalDebug = _this &&
       _this.conn &&
       _this.conn.base &&
       _this.conn.base.options &&
       _this.conn.base.options.debug;
-    const hasConnectionDebug = _this &&
+    const connectionDebug = _this &&
       _this.conn &&
       _this.conn.options &&
       _this.conn.options.debug;
-    const debug = hasGlobalDebug || hasConnectionDebug;
+    const debug = connectionDebug == null ? globalDebug : connectionDebug;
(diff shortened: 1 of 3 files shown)


hasezoey (approved):
LGTM

lib/drivers/node-mongodb-native/collection.js
```diff
@@ -77,11 +77,16 @@ function iter(i) {
     const collection = this.collection;
     const args = Array.from(arguments);
     const _this = this;
-    const debug = _this &&
+    const hasGlobalDebug = _this &&
```
IslandRhythms, line 80: @vkarpov15 Can we use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here or should it be done like this for backwards compatibility?

test/docs/debug.test.js
```diff
@@ -67,4 +71,20 @@ describe('debug: shell', function() {
...
   });
 
+  it('should allow to set the `debug` option on a per-connection basis (gh-12700)', async function() {
+    // Disable global debug
+    mongoose.set('debug', false);
+    // `conn1` with active debug
+    const conn1 = mongoose.createConnection(start.uri);
+    conn1.set('debug', true);
+    const testModel1 = conn1.model('Test', testSchema);
+    await testModel1.create({ dob: new Date(), title: 'Connection 1' });
```
IslandRhythms, line 81: We usually use the issue number, so in this case `gh-12700`, to create the collection name but `date` is a nifty way of doing it too.

IslandRhythms (approved):
LGTM

lib/drivers/node-mongodb-native/collection.js
vkarpov15: For backwards compatibility. Optional chaining is supported natively in Node 14 and higher, but Mongoose 6 supports Node 12.

lib/drivers/node-mongodb-native/collection.js
```diff
@@ -77,11 +77,16 @@ function iter(i) {
...
       _this.conn &&
       _this.conn.base &&
       _this.conn.base.options &&
       _this.conn.base.options.debug;
+    const hasConnectionDebug = _this &&
+      _this.conn &&
+      _this.conn.options &&
+      _this.conn.options.debug;
+    const debug = hasGlobalDebug || hasConnectionDebug;
```
vkarpov15, line 89: Check for nullish rather than falsy. So `connectionDebug == null ? globalDebug : connectionDebug` . That's because setting debug to `false` on a connection should override the global debug config.

test/docs/debug.test.js
```diff
@@ -67,4 +71,20 @@ describe('debug: shell', function() {
     assert.equal(true, lastLog.includes('ISODate'));
   });
 
+  it('should allow to set the `debug` option on a per-connection basis (gh-12700)', async function() {
+    // Disable global debug
+    mongoose.set('debug', false);
```
vkarpov15, line 76: I'd recommend creating a new mongoose instance with `const m = new mongoose.Mongoose()`  and `m.createConnection()` instead of modifying the Mongoose singleton. Modifying global state in tests often causes confusing issues.

vkarpov15 (changes requested):

hasezoey (approved):

vkarpov15 (approved):
Nice, thanks :+1: merging this into 6.8

PR #12707, opened 2022-11-19
Replace used before defined User with this.

sgpinkus:
Changed line gave "'User' was used before it was defined" linting error. Will compile if you ignore the lint error but I think it's more informative to show that `this` can be used - as it is an instance of `UserModel` here.


vkarpov15 (approved):
Thanks :+1:

PR #12715, opened 2022-11-22
fix(types): Update replaceWith pipeline stage

coyotte508:
See doc: https://www.mongodb.com/docs/manual/reference/operator/aggregation/replaceWith/

image

Not sure if `ObjectExpressionOperator` should be modified instead?


vkarpov15 (approved):
Thanks :+1:

PR #12716, opened 2022-11-22
fix(cursor): make `eachAsync()` avoid modifying batch when mixing `parallel` and `batchSize`

### LINKED ISSUE #12652: eachAsync passes more elements than the specified batchSize in doc array to fn, when both parallel and batchSize are used
### Prerequisites

### Mongoose version

6.7.1

### Node.js version

16.18.0

### MongoDB server version

6.0.2

4.8.4

### Description

In a call to `eachAsync`, when both `parallel` and `batchSize` are set to numbers higher than 1, it is expecting to call the `fn` function for each batch of element(s) returned by the cursor, simultaneously, up to `parallel` times, and continue to do so till the exhaustion of the cursor.
As it turns out, the process of reading next elements from the cursor continues to fill the same `doc` array instance, which is already being processed. This can be reproduced by having a time consuming process in the `fn` function and checking the number of elements in `doc` array **before** and **after** such processing.

### Steps to Reproduce

The second and third assertion fail in the fallowing test case:

```js
'use strict';

run().catch(console.error);

async function run() {
    await mongoose.connect('mongodb://localhost:27017/test', {});

    const User = mongoose.model('user', {});

    const usersCount = 10000;

    const users = Array.from({ length: usersCount }).map(() => ({}));
    await User.create(users);

    const cursorBatchSize = 50;
    const batchSize = 100;

    let counter = 0;

    await User.find({})
        .lean()
        .cursor({ batchSize: cursorBatchSize })
        .eachAsync(async (documents) => {
            assert.equal(documents.length, batchSize);

            counter += documents.length;
            await new Promise((f) => setTimeout(f, 1));

            assert.equal(documents.length, batchSize);
        }, { parallel: 10, batchSize });

    assert.equal(counter, usersCount);

}

```

### Expected Behavior

The number of elements in `doc` array, which is passed to `fn`, should be consistent and equal to the `batchSize` passed as `options` to `eachAsync`. Executing other async functions, which might be time consuming, in `fn` function should not effect the number of elements in `doc` array passed to the `fn`.

vkarpov15:
Fix #12652

**Summary**

#12652 points out that there is a race condition in `eachAsync()` when `parallel > 1` and `batchSize` is set. Given that we run a bunch of parallel queues, `documentsBatch` can change while the `eachAsync` callback is running.

This fixes part of the issue: makes it so that at least every document gets processed, and each `eachAsync()` callback gets a consistent batch of documents that doesn't change.

However, there is still the potential issue that the individual documents in the batch can be in any order when `parallel > 1`. Because we run `next()` in parallel, you can end up with surprising behavior like document 9613 in the same batch as documents 9900-9999. Or an `eachAsync()` callback running with a batch of size 67 followed by a batch of 33 even though you have `batchSize = 100` and there's 10000 documents in the result set.

We should consider running `next()` in a queue, getting documents in order and building up the batch, and then building the next batch while `eachAsync()` runs. Rather than running the whole "build up a batch and then call the `eachAsync()` callback" process in parallel. What do you think @hasezoey ?

**Examples**

### DIFF
--- a/lib/helpers/cursor/eachAsync.js
+++ b/lib/helpers/cursor/eachAsync.js
@@ -33,7 +33,7 @@ module.exports = function eachAsync(next, fn, options, callback) {
   const aggregatedErrors = [];
   const enqueue = asyncQueue();
 
-  let drained = false;
+  let aborted = false;
 
   return promiseOrCallback(callback, cb => {
     if (signal != null) {
@@ -42,7 +42,7 @@ module.exports = function eachAsync(next, fn, options, callback) {
       }
 
       signal.addEventListener('abort', () => {
-        drained = true;
+        aborted = true;
         return cb(null);
       }, { once: true });
     }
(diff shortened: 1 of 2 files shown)


vkarpov15:
@hasezoey currently it kicks off a separate "build the batch and call the callback" function in parallel. The problem is that "build the batch" can lead to inconsistent batches because of race conditions - every document will be processed once, but you can end up with batches with out-of-order documents. What I'm suggesting is instead run 1 "build the batch" function, regardless of `parallel`, and keep running "build the batch" while we still have fewer than `parallel` callbacks running.

MongoDB doesn't support multiple `next()` calls on a cursor in parallel anyway, so there's no performance benefit to having multiple "build the batch" functions running at the same time. We implemented it this way purely for expediency.

PR #12717, opened 2022-11-23
fix: setting global `strictQuery` after Schema creation wasn't working

### LINKED ISSUE #12703: Setting global strictQuery doesn't work after schemas are created
### Prerequisites

### Mongoose version

6.7.2

### Node.js version

16.x

### MongoDB server version

6.x

4.8.0

### Description

```mongoose.set('strictQuery', false);``` only correctly disables strict query behavior if you run it before the schemas are created.  After the schemas are created, changes to this property is ignored.  Similarly, if you change it in the middle of schema creation, it affects some schemas but not others.  The schemas should respect the current state of this global if it's set, not what it was at schema creation time.

### Steps to Reproduce

```#!/usr/bin/env node
'use strict';

mongoose.connect('mongodb://localhost:27017/gh7008', { useNewUrlParser: true });
const conn = mongoose.connection;

const schema = new Schema({
  title: String,
});

const Test = mongoose.model('test', schema);

const test0 = new Test({
  title: 'chimichanga'
});
const test1 = new Test({
  title: 'burrito bowl'
});
const test2 = new Test({
  title: 'taco supreme'
});

async function run() {
  assert.strictEqual(mongoose.version, '6.7.2');
  await conn.dropDatabase();
  await test0.save();
  await test1.save();
  await test2.save();

  let cond = {
    $or: [
      {
        title: {
          $regex: 'urri',
          $options: 'i'
        }
      },
      {
        name: {
          $regex: 'urri',
          $options: 'i'
        }
      }
    ]
  }
  let found = await Test.find(cond);
  assert.strictEqual(found.length, 1);
  assert.strictEqual(found[0].title, 'burrito bowl');

  return conn.close();
}

run();
```

### Expected Behavior

The current state of ```mongoose.set('strictQuery', false);``` should be used if no per-query/schema overrides are specified.

lpizzinidev:
fix #12703

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -264,7 +264,7 @@ module.exports = function cast(schema, obj, options, context) {
         }
 
         const strict = 'strict' in options ? options.strict : schema.options.strict;
-        const strictQuery = getStrictQuery(options, schema._userProvidedOptions, schema.options);
+        const strictQuery = getStrictQuery(options, schema._userProvidedOptions, schema.options, context);
         if (options.upsert && strict) {
           if (strict === 'throw') {
             throw new StrictModeError(path);
@@ -374,7 +374,7 @@ function _cast(val, numbertype, context) {
   }
 }
 
-function getStrictQuery(queryOptions, schemaUserProvidedOptions, schemaOptions) {
+function getStrictQuery(queryOptions, schemaUserProvidedOptions, schemaOptions, context) {
   if ('strictQuery' in queryOptions) {
     return queryOptions.strictQuery;
   }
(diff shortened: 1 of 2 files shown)


lib/cast.js
```diff
@@ -387,5 +387,18 @@ function getStrictQuery(queryOptions, schemaUserProvidedOptions, schemaOptions)
   if ('strict' in schemaUserProvidedOptions) {
     return schemaUserProvidedOptions.strict;
   }
+  if (
+    context.mongooseCollection &&
+    context.mongooseCollection.conn &&
+    context.mongooseCollection.conn.base &&
+    context.mongooseCollection.conn.base.options) {
+    const mongooseOptions = context.mongooseCollection.conn.base.options;
```
Uzlopak, line 395: I guess you can simplify this
```javascript
const mongooseOptions = context.mongooseCollection &&
    context.mongooseCollection.conn &&
    context.mongooseCollection.conn.base &&
    context.mongooseCollection.conn.base.options;

if (mongooseOptions) {
    if ('strictQuery' in mongooseOptions) {
      return mongooseOptions.strictQuery;
    }
    if ('strict' in mongooseOptions) {
      return mongooseOptions.strict;
    }
}
```

Uzlopak (approved):
LGTM

vkarpov15 (approved):
Thanks :+1:

PR #12719, opened 2022-11-23
make `modifiedPaths` safer.

zzztttkkk:
**Summary**

`helpers.common.modifiedPaths` is a recursion function.

when user call `db.updateOne({key: badValue})`, the `badValue` will cause a endless recursion call. but it is impossible to check user's input. so I think we can check the recursion level , it will be helpful for debugging.

**Examples**

```javascript

async function updateDoc(){
    const count = collectionA.countDocuments({}); // miss an `await`, will cause a endless recursion call.
    await collectionB.updateOne({key: count}, {upsert: true});
}

```

### DIFF
--- a/lib/helpers/common.js
+++ b/lib/helpers/common.js
@@ -7,6 +7,8 @@
 const Binary = require('../driver').get().Binary;
 const isBsonType = require('./isBsonType');
 const isMongooseObject = require('./isMongooseObject');
+const MongooseError = require('../error');
+const util = require('util');
 
 exports.flatten = flatten;
 exports.modifiedPaths = modifiedPaths;
@@ -67,18 +69,21 @@ function flatten(update, path, options, schema) {
  * ignore
  */
 
-function modifiedPaths(update, path, result, recursion=null) {
-  if(recursion == null){
+function modifiedPaths(update, path, result, recursion = null) {
+  if (recursion == null) {
     recursion = {
-      count: 0,
-      raw: {update, path}
+      raw: { update, path },
+      trace: new WeakSet()
     };
   }
 
-  recursion.count++;
-  if(recursion.count >= 1024){
-    throw new Error(`Mongoose: bad update value, ${recursion.raw.update}, ${recursion.raw.path}`);
+  if (recursion.trace.has(update)) {
+    throw new MongooseError(`a circular reference in the update value, updateValue:
+${util.inspect(recursion.raw.update, { showHidden: false, depth: 1 })}
+updatePath: ${recursion.raw.path}
+-------`);
   }
+  recursion.trace.add(update);
 
   const keys = Object.keys(update || {});
   const numKeys = keys.length;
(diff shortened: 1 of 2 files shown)


lib/helpers/common.js
```diff
@@ -67,7 +67,19 @@ function flatten(update, path, options, schema) {
...
+  if(recursion == null){
+    recursion = {
+      count: 0,
+      raw: {update, path}
+    };
+  }
+
+  recursion.count++;
+  if(recursion.count >= 1024){
+    throw new Error(`Mongoose: bad update value, ${recursion.raw.update}, ${recursion.raw.path}`);
```
Uzlopak, line 80: Maybe create a new MongooseError?

zzztttkkk: I was just worried that it would cause a circular reference.

I will make a new commit tomorrow.

Uzlopak: Dont worry, I just think a new MongooseRecursionError or something like that would be nicer.

lib/helpers/common.js
```diff
@@ -67,7 +67,19 @@ function flatten(update, path, options, schema) {
...
+function modifiedPaths(update, path, result, recursion=null) {
+  if(recursion == null){
+    recursion = {
+      count: 0,
+      raw: {update, path}
+    };
+  }
+
+  recursion.count++;
+  if(recursion.count >= 1024){
```
vkarpov15, line 79: I think this is a bit brittle. Admittedly we're unlikely to have depth >= 1024 unless there's an infinite recursion. But I think a better approach would be to use a `WeakSet` or `WeakMap` to track what objects we've seen in the recursion, similar to how [`getIndexes()` does it](https://github.com/Automattic/mongoose/blob/master/lib/helpers/schema/getIndexes.js#L16-L29).

vkarpov15 (changes requested):
Couple of suggestions. Also, can you please add a test case for this?

zzztttkkk:
@vkarpov15 
Hi,the new commit is waiting a review.

lib/helpers/common.js
```diff
@@ -67,7 +69,22 @@ function flatten(update, path, options, schema) {
...
+      raw: { update, path },
+      trace: new WeakSet()
+    };
+  }
+
+  if (recursion.trace.has(update)) {
+    throw new MongooseError(`a circular reference in the update value, updateValue:
+${util.inspect(recursion.raw.update, { showHidden: false, depth: 1 })}
+updatePath: ${recursion.raw.path}
+-------`);
```
vkarpov15, line 84: Minor nitpick: this `----` is a bit messy. But I'll clean that up after merging, no need for that to be a blocker.

lib/helpers/common.js
```diff
@@ -67,7 +69,22 @@ function flatten(update, path, options, schema) {
...
+  }
+  recursion.trace.add(update);
```
vkarpov15, line 86: Add a check for `if (update == null) { return }` before this. Adding `null` or `undefined` to a `WeakSet()` throws an error, so we should avoid this case.

vkarpov15 (changes requested):

lib/helpers/common.js
zzztttkkk: Sorry, I don't know this. I added a null check.

vkarpov15 (approved):

PR #12720, opened 2022-11-23
fix(types): correctly infer ReadonlyArray types in schema definitions

### LINKED ISSUE #12611: Reusing array schema definitions
### Prerequisites

### Mongoose version

6.7.0

### Node.js version

16.18.0

### MongoDB version

5.x

### Operating system

macOS

_No response_

### Issue

I'm trying to migrate a project to use official typescript inferred schema definitions. One of the things I do is reuse common schema definitions. For example:

```ts
export const reusableFields = {
  description: { type: String, required: true },
  skills: { type: [Schema.Types.ObjectId], ref: "Skill", default: [] }
} as const;

export const firstSchema = new Schema({
  ...reusableFields,
  anotherField: String
})

export type Props = InferSchemaType<typeof firstSchema>;
```

Props typ is:

```ts
type Props = {
    description: string;
    skills: {
        1: unknown;
        0?: Types.ObjectId;
    };
    anotherField?: string;
}
```

Non-array types work as expected (`description`), but array types (`skills`) cannot be inferred. I haven't figured out how to make it work yet. Any suggestions?

vkarpov15:
Fix #12611

**Summary**

Bitten yet again by the fact that ReadonlyArrays and Arrays are separate classes in TypeScript :sob: 

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -942,14 +942,14 @@ function gh12590() {
 function gh12611() {
   const reusableFields = {
     description: { type: String, required: true },
-    skills: { type: [Schema.Types.ObjectId], ref: "Skill", default: [] }
+    skills: { type: [Schema.Types.ObjectId], ref: 'Skill', default: [] }
   } as const;
-  
+
   const firstSchema = new Schema({
     ...reusableFields,
     anotherField: String
   });
-  
+
   type Props = InferSchemaType<typeof firstSchema>;
   expectType<{
     description: string;
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -938,3 +938,22 @@ function gh12590() {
   });
 
 }
+
+function gh12611() {
+  const reusableFields = {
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM

though i think the old declaration can be replaced, because `ReadonlyArray` is compatible with `Array`, but not the other way around

```ts
const arr1 = ['test1'];
const arr2 = ['test2'] as const;

type TestR<T extends ReadonlyArray<any>> = T;

const t1: TestR<typeof arr1> = arr1; // works
const t2: TestR<typeof arr2> = arr2; // works

type TestW<T extends Array<any>> = T;

const t3: TestW<typeof arr1> = arr1; // works
const t4: TestW<typeof arr2> = arr2; // error "trying to assign readonly to mutable type"
```

vkarpov15:
@hasezoey I tried removing `extends (infer Item)[]`, tests fail. So it doesn't look like we can remove the declaration after all.

PR #12731, opened 2022-11-27
Infer timestamps option from schema

### LINKED ISSUE #12069: TypeScript: timestamps not inferred from schema

vkarpov15:
**Summary**

Fix #12069

Adds `{ createdAt: Date, updatedAt: Date }` if `{ timestamps: true }` in schema options. Works great, the only potential problem is the `expectType<{ createdAt: Date, updatedAt: Date } & { name?: string }>({} as WithTimestamps);` weirdness - not sure if that is a problem, but it only seems to be an issue in tsd :man_shrugging: 

**Examples**


test/types/schema.test.ts
```diff
@@ -863,6 +862,19 @@ function gh12242() {
   expectType<0 | 1>({} as Example['active']);
 }
 
+function testInferTimestamps() {
+  const schema = new Schema({
+    name: String
+  }, { timestamps: true });
+
+  type WithTimestamps = InferSchemaType<typeof schema>;
+  // For some reason, expectType<{ createdAt: Date, updatedAt: Date, name?: string }> throws
```
hasezoey, line 871: that is because the type literally is a AND between multiple things and the internal representation (as you can see with a hover in vscode) and `expectType` is strict and only matches the actual internal representation, if you want the simpler notion use `expectAssignable`

hasezoey (approved):
LGTM

test/types/schema.test.ts
vkarpov15: Thanks for clarifying, I spent an hour or two trying to figure out WTF was going on :face_with_spiral_eyes:

PR #12736, opened 2022-11-29
fix(query): handle deselecting _id when another field has schema-level `select: false`

### LINKED ISSUE #12670: 'Cannot do exclusion on field in inclusion projection' error when field1 excluded in schema & query.select('field2 -_id')
### Prerequisites

### Mongoose version

6.7.2

### Node.js version

18.x

### MongoDB server version

5.x

_No response_

### Description

I would like to report a long standing bug regarding simultaneous field exclusion in schema and select() query, with -_id not specified as first field.

### Steps to Reproduce

Suppose we have field1 excluded from selection/projection in schema:
```js
    field1: {
        type: String,
        select: false
    },
    field2: String
});

User.find().select('field2 -_id')
    .then(console.log)
    .catch(err => console.log(err.message));
```
The error returned is `Cannot do exclusion on field field1 in inclusion projection`.
However, if I first exclude the _id in the select() then the query executes just fine, no error in that case:
```js
User.find().select('-_id field2')
```
Not sure when this regression was introduced or whether it has ever worked in the first place. @vkarpov15 could you please double check? Thanks for looking into this!

### Expected Behavior

Excluding -_id from select() should work irrespective of its position in the list of fields. At the moment it only works when -_id is the first field, then the rest of the included fields.
This issue only occurs when there's another field excluded from select in the schema.

vkarpov15:
Fix #12670

**Summary**

#12670 is a fairly clear bug, thankfully with a quick fix.

**Examples**

### DIFF
--- a/lib/queryhelpers.js
+++ b/lib/queryhelpers.js
@@ -163,6 +163,9 @@ exports.applyPaths = function applyPaths(fields, schema) {
       if (!isDefiningProjection(field)) {
         continue;
       }
+      if (keys[keyIndex] === '_id') {
+        continue;
+      }
       exclude = !field;
       break;
     }
--- a/test/model.field.selection.test.js
+++ b/test/model.field.selection.test.js
@@ -475,7 +475,6 @@ describe('model field selection', function() {
     db.deleteModel(/BlogPost/);
     const BlogPost = db.model('BlogPost', BlogPostSchema);
 
-
     await BlogPost.create({
       author: 'me',
       settings: {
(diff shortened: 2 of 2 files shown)


IslandRhythms:
LGTM

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

hasezoey (approved):
LGTM, is there already a test for mongodb `< 5` to check for that error that it still gets triggered?

PR #12737, opened 2022-11-29
fix(schema): copy indexes when calling `add()` with schema instance

### LINKED ISSUE #12654: Schema.add method not copy indexes
### Prerequisites

### Mongoose version

6.6.5

### Node.js version

14.18.0

### MongoDB server version

4.2.21

4.8.4

### Description

When i use Schema.prototype.add() method for copy schema to an other schema, the indexes don't  be copied.

### Steps to Reproduce

I have a `test` schema

```ts
const test = new Schema({
   name: String
}):

...someMethods, statics...

test.index({name: 1});
```

And other schema called 'extended' and i want to add `test` schema in `extended` with [add() method](https://mongoosejs.com/docs/api/schema.html#schema_Schema-add) like the documentation say "You can also `add()` another schema and copy over all paths, virtuals, getters, setters, **indexes**, methods, and statics."

```ts
const extended = new Schema({
   property: String
});

extended.add(test);

```

But extended schema haven't the index `name`

### Expected Behavior

When i call `extended.indexes()` the returned array has the index `{name: 1}`

vkarpov15:
Fix #12654

**Summary**

Docs say that `Schema.prototype.add()` copies indexes if you call `add()` with a schema instance, but looks like that doesn't actually happen. I can see the case for putting this change in a minor release, but I think patch is fine since the docs indicate that this is the correct behavior.

**Examples**

### DIFF
--- a/lib/helpers/schema/merge.js
+++ b/lib/helpers/schema/merge.js
@@ -23,5 +23,6 @@ module.exports = function merge(s1, s2, skipConflictingPaths) {
     s1.virtuals[virtual] = s2.virtuals[virtual].clone();
   }
 
+  s1._indexes = s1._indexes.concat(s2._indexes || []);
   s1.s.hooks.merge(s2.s.hooks, false);
 };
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

hasezoey (approved):
LGTM

PR #12738, opened 2022-11-29
doc fixes

IslandRhythms:


test/docs/validation.test.js
```diff
@@ -502,6 +502,8 @@ describe('validation docs', function() {
       // When running in `validate()` or `validateSync()`, the
       // validator can access the document using `this`.
       // Does **not** work with update validators.
+      // use this.get('name') to prevent this error
+      // if (this.name === undefined) this.name = this.get('name')
```
vkarpov15, line 506: No need for the line 506 comment. Just make line 507 use `this.get('name').toLowerCase()`

vkarpov15 (changes requested):

test/docs/validation.test.js
IslandRhythms: I'll need to remove the assertion at the bottom then with the comments since it won't happen anymore if thats good with you.

PR #12771, opened 2022-12-06
style(types/aggregate): rename generic to match Query definition

hasezoey:
**Summary**

This PR renames the generic `Aggregate` uses to make it more clear what the generic does and match the definition of `Query`

**Examples**


vkarpov15 (approved):
Thanks :+1:

Issue #12536, opened 2022-10-06
Add support for $fill stage pipeline


emilskovmand:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Add the $fill stage pipeline to aggregations in mongoose.
Mongodb docs:
https://www.mongodb.com/docs/manual/reference/operator/aggregation/fill/

I would add this feature in a PR myself, if I had more insight for the module and I might even do it in the near future if no one wants to before me.

### Motivation

Why not?
And also I would love to have this for my time-series collections.

### Example

```js
myModel.aggregate([
   {
      $fill:
         {
            output:
               {
                  "emptyValueInt": { value: 0 }
               }
         }
   }
] )
```

PR #12778, opened 2022-12-08
fix(types): correctly infer `this` when using `pre('updateOne')` with `{ document: true, query: false }`

### LINKED ISSUE #11257: schema.pre('save'  for document not working since 6.1.5

I can't decide, bug or not?

```
schema.pre('save', { document: true, query: false }, function (next) {
    const user = this;  // query since 6.1.5, not a document itself
```
is not working since 6.1.5

```
{
  "compileOnSave": true,
  "compilerOptions": {
    "baseUrl": "./",
    "outDir": "./",
    "module": "commonjs",
    "moduleResolution": "node",
    "target": "es2020",
    "lib": [ "es2020", "dom" ],
    "resolveJsonModule": true,
    "sourceMap": true
  },
  "exclude": [
    "node_modules"
  ]
}
```

"this" should be the document before being saved

nodejs 16.13.2 x64
mongoose 6.1.4(working) 6.1.5~7(not working)
mongodb 5.0.5 x64

```
> mongoose.version
'6.1.4'
```

6.1.4 since I rollbacked

vkarpov15:
Fix #11257

**Summary**

`pre('updateOne')` currently infers `this` is a document, rather than a query by default.

**Examples**

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -37,7 +37,7 @@ jobs:
       - run: npm install
         working-directory: benchmarks/typescript/simple
 
-      - run: npx ts-benchmark -p ./ -f 17/100000 18 29 32 -b master -g -t --colors
+      - run: npx ts-benchmark -p ./ -f 17/110000 18 29 32 -b master -g -t --colors
         working-directory: benchmarks/typescript/simple
         env:
           DB_URL: ${{ secrets.DB_URL }}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,7 +58,7 @@ If you'd like to preview your documentation changes, first commit your changes t
 * `npm install`
 * `npm run docs:view`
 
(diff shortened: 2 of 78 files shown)


hasezoey:
ts-benchmark is failing with this PR, is this something that can be fixed without removing the types again?

vkarpov15:
@hasezoey that's just due to TS slowdowns creeping up on us, and this change is just the straw that broke the camel's back. This PR increased instantiations from around 99k to just over 100k.

I think the PR that started the TS performance degradation creep was this one: https://github.com/Automattic/mongoose/commit/76eaadd13d11bb140b911fba9103756ce5950a80 . Here's TS benchmark after: https://github.com/Automattic/mongoose/actions/runs/3231069336/jobs/5329705376 and previous TS benchmark run: https://github.com/Automattic/mongoose/actions/runs/3212307397/jobs/5251060253 . We'll investigate and see if we can get rid of some of the perf overhead of #12536

hasezoey:
i guess we just need / can only increase the threshold now, the types are just getting more complex to actually represent what is going on at runtime

PR #12781, opened 2022-12-08
typescript: Allow null for optional document fields

### LINKED ISSUE #12748: InferSchemaType: Allow null for optional fields
### Prerequisites

### Mongoose version

6.7.5

### Node.js version

16.17.1

### MongoDB server version

4.4

_No response_

### Description

Currently optional fields in schemas accept either the field type or undefined. Can we change this to also accept `null`? AFAIK `null` should be treated the same as `undefined` in this case, and it makes the types more usable. In our codebase, we generally treat `null` and `undefined` the same in our types, unless we're forced to do otherwise due to explicit behavior difference such as https://mongoosejs.com/docs/migrating_to_6.html#removed-omitundefined.

i.e. we are using

```
type Maybe<T> = T | null | undefined
```

### Steps to Reproduce

If I define the following schema:
```
const ExampleSchema = new Schema({
  example: String
});
type ExampleSchemaFields = InferSchemaType<typeof ExampleSchema>;
```
It infers the following type:
```
// type ExampleSchemaFields = {
//   example?: string | undefined;
// }
```

Can we change this to `example?: string | null | undefined`?

### Expected Behavior

_No response_

JavaScriptBach:
**Summary**

When inferring types, allow `null` as well as `undefined` if the field is optional.

Fixes #12748

**Examples**


Uzlopak:
Accidently closed.

Null has a special meaning in mongo. I doubt if this is a desired behavior.

JavaScriptBach:
We can insert `null` into a collection, so why should that not be allowed for optional Mongoose fields?

Uzlopak:
Nobody says that null is not a valid input. But while JSON.stringify removes keys with value null, the bson serializer serializes null values. 
So the expectations can be different.

So we should keep this in mind when merging this.

JavaScriptBach:
I think it is the opposite actually. JSON.stringify() removes undefined keys, but keeps null keys: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description

vkarpov15 (approved):
This works, thank you for making this change. You are correct that you can set any non-required property to `null`.

My only concern is that this change might be a bit heavy for a patch release, so we'll save it for our next minor release.

JavaScriptBach:
It turns out what I really need for my use case is for optional strings to infer as `foo: string | null | undefined` instead of `foo ?: string | undefined`. But that seems a different direction than what was taken in this project, and I don't suppose you're willing to change the inference to a required field unioned with null and undefined?

So feel free to merge or not merge this PR, but I no longer have a stake in this issue now. Thanks for reviewing either way.

vkarpov15 (approved):
Merging into 6.9

PR #12784, opened 2022-12-09
InferSchemaType: consider { required: boolean } required if it isn't explicitly false

### LINKED ISSUE #12782: InferSchemaType: Improve type safety by inferring `{required: boolean}` as a required field instead of optional
### Prerequisites

### 🚀 Feature Proposal

If I write the following code:
```
const definition = { test: { type: String, required: true } };
const schema = new Schema(definition);
type SchemaType = InferSchemaType<typeof schema>;
```

`SchemaType` will think that `test` is an optional string. This is because TypeScript infers the type of `definition` as `{ test: { type: StringConstructor, required: boolean } }` instead of `{ test: { type: StringConstructor, required: true } }`, so Mongoose can't infer whether the field is required in https://github.com/Automattic/mongoose/blob/master/types/inferschematype.d.ts#L83, and defaults to optional.

The user can workaround this by writing `required: true as const` or defining schema definition inline into the `new Schema()` constructor, but if they don't realize something is wrong, they will end up with incorrectly optional types.

I propose that we infer `required: boolean` as a required field for the following reasons:
1. `required` defaults to `false`. If the user specified `required`, it's usually `true`.
2. It's better that we type an optional field as required (makes types unnecessarily strict, but won't break) instead of vice versa (incorrectly allows optional fields which will break at runtime).

### Motivation

Improve type safety by defaulting ambiguity to required instead of optional.

### Example

See above.

JavaScriptBach:
**Summary**

Fixes #12782

**Examples**

Added a test

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -969,3 +969,12 @@ function gh12611() {
     anotherField?: string;
   }>({} as Props);
 }
+
+function gh12782() {
+  const schemaObj = { test: { type: String, required: true } };
+  const schema = new Schema(schemaObj);
+  type Props = InferSchemaType<typeof schema>;
+  expectType<{
+    test: string
+  }>({} as Props);
+}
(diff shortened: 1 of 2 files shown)


Uzlopak:
You forgot the typings tests.

vkarpov15 (approved):
Thanks :+1:

PR #12785, opened 2022-12-09
fix(model): prevent index creation on syncIndexes if not necessary

lpizzinidev:
fix #12250

**Summary**
The `model.collection.createIndex` is now executed only if there are indexes in the `toCreate` array.
`getIndexesToCreate` now includes a check to evaluate if the index was dropped. In that case, it will proceed to recreate it.


AbdelrahmanHafez (approved):
If the tests pass, LGTM. I'm kinda surprised that the fix is that simple.
Thanks @lpizzinidev :)

@vkarpov15 @Uzlopak Any idea why the tests didn't run here? 🤨

AbdelrahmanHafez:
fixes #12250

AbdelrahmanHafez (approved):

vkarpov15 (approved):
Thanks :+1:

PR #12803, opened 2022-12-15
docs: make `Document.prototype.$clone()` public

vkarpov15:
**Summary**

For @lpizzinidev . We intended to make `$clone()` public for 6.8 in #12549, but looks like we didn't add it to TypeScript or the docs :cry: 

**Examples**


IslandRhythms (approved):
LGTM

PR #12741, opened 2022-11-30
Not Working: Throwing error with "localhost:27017"

SadiqOnGithub:
by using `mongoose.connect('mongodb://localhost:27017/test')`  by node application, running on windows, is throwing this error : 

```txt
MongooseServerSelectionError: connect ECONNREFUSED ::1:27017
    at Connection.openUri (C:\Users\lenovo\Documents\webistes projects\Node\node prac\mongo\node_modules\mongoose\lib\connection.js:824:32)
    at C:\Users\lenovo\Documents\webistes projects\Node\node prac\mongo\node_modules\mongoose\lib\index.js:412:10
    at C:\Users\lenovo\Documents\webistes projects\Node\node prac\mongo\node_modules\mongoose\lib\helpers\promiseOrCallback.js:41:5
    at new Promise (<anonymous>)
    at promiseOrCallback (C:\Users\lenovo\Documents\webistes projects\Node\node prac\mongo\node_modules\mongoose\lib\helpers\promiseOrCallback.js:40:10)
    at Mongoose._promiseOrCallback (C:\Users\lenovo\Documents\webistes projects\Node\node prac\mongo\node_modules\mongoose\lib\index.js:1265:10)    
    at Mongoose.connect (C:\Users\lenovo\Documents\webistes projects\Node\node prac\mongo\node_modules\mongoose\lib\index.js:411:20)
    at main (C:\Users\lenovo\Documents\webistes projects\Node\node prac\mongo\server.js:6:17)
    at Object.<anonymous> (C:\Users\lenovo\Documents\webistes projects\Node\node prac\mongo\server.js:3:1)
    at Module._compile (node:internal/modules/cjs/loader:1205:14) {
  reason: TopologyDescription {
    type: 'Unknown',
    servers: Map(1) { 'localhost:27017' => [ServerDescription] },
    stale: false,
    compatible: true,
    heartbeatFrequencyMS: 10000,
    localThresholdMS: 15,
    setName: null,
    maxElectionId: null,
    maxSetVersion: null,
    commonWireVersion: 0,
    logicalSessionTimeoutMinutes: null
  },
  code: undefined
}
```

==================================X==================================

I apologize if this proposed change is not up to the mark or is violating some community guidelines.

but I wasted a lot of time on this so I can't resist but propose this change.

I think other people also have struggled with this
read more on this stack overflow answer  :   https://stackoverflow.com/a/70517348/14075519

**Summary**

**Examples**

PR #12811, opened 2022-12-17
chore: replace many occurrences of "localhost" with "127.0.0.1"

hasezoey:
**Summary**

This PR replaces many occurrences of `localhost` with `127.0.0.1`
Exceptions:
- everything in `CHANGELOG.md`
- some exceptions in `connections.md` where it is talked about using `localhost` as a hostname

re https://github.com/Automattic/mongoose/pull/12741

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,7 +58,7 @@ If you'd like to preview your documentation changes, first commit your changes t
 * `npm install`
 * `npm run docs:view`
 
-Visit `http://localhost:8089` and you should see the docs with your local changes. Make sure you `npm run docs:clean` before committing, because automated generated files to `docs/*` should **not** be in PRs.
+Visit `http://127.0.0.1:8089` and you should see the docs with your local changes. Make sure you `npm run docs:clean` before committing, because automated generated files to `docs/*` should **not** be in PRs.
 
 #### Documentation Style Guidelines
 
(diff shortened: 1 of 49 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

Uzlopak (approved):

CONTRIBUTING.md
```diff
@@ -58,7 +58,7 @@ If you'd like to preview your documentation changes, first commit your changes t
 * `npm install`
 * `npm run docs:view`
 
-Visit `http://localhost:8089` and you should see the docs with your local changes. Make sure you `npm run docs:clean` before committing, because automated generated files to `docs/*` should **not** be in PRs.
+Visit `http://127.0.0.1:8089` and you should see the docs with your local changes. Make sure you `npm run docs:clean` before committing, because automated generated files to `docs/*` should **not** be in PRs.
```
vkarpov15, line 61: This one isn't strictly necessary because Node !== browser. But shouldn't be a problem.

vkarpov15 (approved):
Thanks :+1:

CONTRIBUTING.md
hasezoey: i saw this in the mass replace and wanted to keep it just in case a browser also decides to change that meaning and to keep the docs more consistent (or in case it is used as a argument to some node process, like a link checker)

PR #12818, opened 2022-12-20
fix(schematypes): add property "defaultOptions"

hasezoey:
**Summary**

This PR adds the property `defaultOptions` for `SchemaType` and its extending classes
for `SchemaType` itself, property `defaultOptions` is optional, see: (and also search through the file, there is no explicit assignment)

https://github.com/Automattic/mongoose/blob/5528a6428bb08091c03d868e249c2e5a30144a71/lib/schematype.js#L55

but all mongoose provided extending classes explicitly define it:

https://github.com/Automattic/mongoose/blob/5528a6428bb08091c03d868e249c2e5a30144a71/lib/schema/SubdocumentPath.js#L326


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

vkarpov15 (approved):
Thanks :+1:

Issue #12813, opened 2022-12-18
Docs :: Add guidance on using loadClass with Typescript


sgpinkus:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I'm trying to port a JS project that uses `loadClass` to TS (it's doing my head in trying to get types working completely ..). The [documentation on TS](https://mongoosejs.com/docs/typescript.html) is good enough, except doesn't mention anything about `loadClass`. A basic example showing how to use it with Mongoose's typings, or alternatively, a statement recommending against its use, and briefly why would be helpful.

Aside: the fundamental class structure implied by loadClass seems to be at odds with Mongoose's actual class structure. Maybe this wasn't a problem in JS work, but it's definitely, at the least, a bit confusing in TS world. Maybe loadClass should be deprecated.

### Motivation

Help users trying to decide whether to try shoehorn loadClass into Mongoose typings or not.

hasezoey:
as a note for those wanting information now instead of waiting for the documentation:

from what i can tell, `loadClass` currently does not affect the typescript type, so you would have to use the class like a interface and `&` it with the actual property interface

example:

```ts
class MyClass {
  myMethod() { return 42; }
  static myStatic() { return 42; }
  get myVirtual() { return 42; }
}

const schema = new mongoose.Schema({ property1: String });
schema.loadClass(MyClass);

type MySchema {
  property1: string;
}

type MyCombined = MySchema & MyClass;

const model = mongoose.model<MyCombined>('MyClass', schema);
```

sgpinkus:
@hasezoey thanks.

Just to be clear, as far as I understand, there are "lean" caveats to above. Example: `doc.toObject().myMethod()` will compile and result in runtime error.

Main thing I wanted to do was type MyClass. The problem is "methods" and "statics" both have different sub classes in Mongoose / Mongoose type system. So the only way I've found to make it work is split the class in two: MyStatics, MyMethods, and call loadClass twice. All the methods in MyStatics have to be static for no reason which is a bit hacky, so maybe there should be `loadMethods()`, loadStatics()` to be consistent with how Mongoose actually works.

hasezoey:
[...]

PR #12820, opened 2022-12-21
Add guard to stop loadClass copying Document if Document is used as base of loaded class (same hack as implemented for Model already).

sgpinkus:
**Summary**
I want to set `Document` as the base class of a `loadClass` class so I can inherit its typings on `this` conveniently. There is a [guard condition hack in the code](https://github.com/Automattic/mongoose/blob/c77dd1bfbec6b551f0ba602b318bdd98f68c4aeb/lib/schema.js#L2291) to allow this for `Model` already. This PR just extends to guard condition to account for possibility Document is the base class. If  that didn't already exist I wouldn't PR this, but since it exists may as well account for Document too, in case people do that, and for consistency.

**Examples**
Something like this:

```
export class MyResourceMethods extends Document
{
  delete() {
    return this.remove();
  }
  async ping() {
    await this.updateOne({ $inc: { pingCount: 1 } });
    return { modified: true };
  }
 ...
}
```

If you want Typescript to know types for Document methods (updateOne, remove in example) you can either type `this` in every method or do the above. The above is more convenient.

Refs #12813.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -198,6 +198,8 @@ function Document(obj, fields, skipId, options) {
   applyQueue(this);
 }
 
+Document.prototype.$isMongooseDocumentPrototype = true;
+
 /**
  * Boolean flag specifying if the document is new. If you create a document
  * using `new`, this document will be considered "new". `$isNew` is how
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #12823, opened 2022-12-22
fix(schema): `Schema.add` should add properties to schema's `.obj`

### LINKED ISSUE #12799: Unable to add properties to a schema after it has been initialized (trying to use Schema.add method as per docs)
### Prerequisites

### Mongoose version

6.8.0

### Node.js version

18.12.1

### MongoDB server version

6.0.3

### Description

Hi,
(likely) since the latest version I am having hard time merging two Schemas. Basically an example like this no longer works for me:

```
    const testSchema = mongoose.Schema();
    testSchema.add({ test: String });
```

The Schema's obj is undefined if I do this:
```
testSchema Schema {
   obj: undefined,
   ...
```

But it did work for me before and it is still mentioned here in the docs:
https://mongoosejs.com/docs/api.html#schema_Schema-add

Can someone please confirm this is really the case?

I need to edit the original schema which is initialized right upon my app launch and then extend it based on a config file once that gets loaded.

Btw, dirty way doesn't work either - I can in theory do `testSchema.obj = { ...testSchema.obj, test: String }`, but even if I can then see it in the Schema it doesn't seem to work (I can't save anything with that key to the DB, it is just skipped).

Thank you very much.

### Steps to Reproduce

Run this code:
```
    const testSchema = mongoose.Schema();
    testSchema.add({ test: String });
```
and if you print it's `.obj` to the console it will be undefined (= empty).

lpizzinidev:
fix #12799 

**Summary**
Adding a property with the `Schema.add` method was not adding it to the `.obj` object of the schema.

**Examples**
```
const schema = new Schema();
schema.add({ test: String });
console.log(schema.obj); // Was undefined, now contains the `test` property
```

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -98,7 +98,7 @@ function Schema(obj, options) {
     return new Schema(obj, options);
   }
 
-  this.obj = obj;
+  this.obj = obj || {};
   this.paths = {};
   this.aliases = {};
   this.subpaths = {};
@@ -606,6 +606,15 @@ Schema.prototype.add = function add(obj, prefix) {
 
     const fullPath = prefix + key;
     const val = obj[key];
+
+    if (
+      this.obj &&
+      key !== '__v' &&
+      key !== '_id' &&
+      !(key in this.obj)
+    ) {
+      this.obj[key] = val;
+    }
 
     if (val == null) {
       throw new TypeError('Invalid value for schema path `' + fullPath +
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -2913,4 +2913,12 @@ describe('schema', function() {
 
     assert.equal(schema._getSchema('child.testMap.foo.bar').instance, 'Mixed');
   });
+
+  it('Schema.add method should add properties to schema.obj object (gh-12799)', () => {
+    const schema = new Schema();
+
+    schema.add({ test: String });
+
+    assert.ok('test' in schema.obj === true);
+  });
 });


lib/schema.js
```diff
@@ -607,6 +607,15 @@ Schema.prototype.add = function add(obj, prefix) {
     const fullPath = prefix + key;
     const val = obj[key];
 
+    if (
+      this.obj &&
+      key !== '__v' &&
+      key !== '_id' &&
```
vkarpov15, line 613: Why ignore `_id` and `__v`? `schema.add({ _id: Number })` should be valid.

lib/schema.js
```diff
@@ -607,6 +607,15 @@ Schema.prototype.add = function add(obj, prefix) {
...
+      !(key in this.obj)
```
vkarpov15, line 614: `.add()` overwrites keys, so I don't think this check is necessary

vkarpov15 (changes requested):
A couple of problems with the PR. I'd also like to hear why OP is using `obj`. Because `obj`  is documented as "The original object passed to the schema constructor", so according to the docs the current behavior is correct:

image

lpizzinidev:
@vkarpov15 
I missed the documentation about `obj`.
If that is the definition then the issue is not a bug (just expected behavior) and should be closed, as well as this PR.
If you think that the new behavior is an improvement I will update the docs and apply the suggested changes.

vkarpov15:
In that case, I'll close this PR for now. We may re-open it in the future if it turns out we have reason to improve `.obj`.

PR #12824, opened 2022-12-22
fix(model): respect discriminators with `Model.validate()`

### LINKED ISSUE #12621: Make `Model.validate()` respect discriminators when discriminator key is set
### Prerequisites

### Mongoose version

5.12.8

### Node.js version

v16.18.0

### MongoDB version

5.0.13

### Operating system

Linux

22.04

### Issue

Hello. I have a simple question, and please don't consider this a feature request or something similar, I am merely asking for help. I've searched the docs and online, but I couldn't find an example of this, so I thought I'd ask here.

Is it possible to have discriminators while only using a single model. I think this is best illustrated with an example. Let's say we have 3 schemas, `AnimalSchema` - the main schema, and two types of animals `CatSchema` and `DogSchema` which all have distinct properties, and the discriminator key in `AnimalSchema` is `kind`. Finally, we have an `Animal` model.

```js

const CatSchema = new Schema(
    { meows: { type: Boolean, required: true } }
);

const DogSchema = new Schema(
    { barks: { type: Boolean, required: true } }
);

const AnimalSchema = new Schema(
    {
        _id: { type: Schema.Types.ObjectId, auto: true },
        id: { type: String, required: true },
    },
    {
        discriminatorKey: 'kind',
    }
);

AnimalSchema.discriminator('cat', CatSchema);
AnimalSchema.discriminator('dog', DogSchema);

const Animal = mongoose.model('animal', AnimalSchema);
```

What I want to do is to instantiate `new Animal`s with their `kind`, and be able to check their validity like this:

```js
(async () => {
    const validPet = new Animal({
        id: '123',
        kind: 'cat',
        meows: true
    });

    const invalidPet1 = new Animal({
        id: '123',
        kind: 'dog',
        meows: true // <= a dog can't meow
    });

    const invalidPet2 = new Animal({
        id: '123',
        kind: 'INVALID',
        meows: 'INVALID'
    });

    await Animal.validate(validPet); // ✅ valid
    await Animal.validate(invalidPet1); // ❌ invalid
    await Animal.validate(invalidPet2); // ❌ invalid
})();
```

I am currently using v`5.12.8` but the code example provided in this post was made with v`6.latest` as I thought that might work and I noticed some new syntax and features, but sadly mongoose didn't pick up any errors with `invalidPet1` and `invalidPet2` when I tried to `.validate()` against the `Animals` model.

So yep, don't want to make this any longer than necessary, just wanted to ask if it's possible to do this on v5, v6 or at all. I have done all of this a few years ago with nested paths, but it's not an option in this particular use case.

Thanks.

vkarpov15:
Fix #12621

**Summary**

`Model.validate()` just uses the model's schema, doesn't look for a discriminator key in `obj`. This PR fixes that issue. Also separated out `Model.validate` tests into a separate file.

**Examples**


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

IslandRhythms (approved):
LGTM

test/model.validate.test.js
```diff
@@ -0,0 +1,146 @@
...
+
+const start = require('./common');
+
+const assert = require('assert');
+
+const mongoose = start.mongoose;
+const Schema = mongoose.Schema;
+
+describe('model: validate: ', function() {
+  afterEach(() => mongoose.deleteModel(/.*/));
```
hasezoey, line 11:
```suggestion
  beforeEach(() => mongoose.deleteModel(/.*/));
```

change it to `before` because from what i know mocha re-uses the same environment, which means it has the possibility of having models transfer from other tests here and conflicting
also it is what is used in other tests

hasezoey:
well, after the change the error changed, but i cannot quite find the reason why it still complains about overwriting a model (all the files involved have a `beforeEach` to delete models)

maybe it is related that `types.array.test.js` sets `mongoose` like `const mongoose = require('./common').mongoose;` while the others do `const mongoose = start.mongoose;`

vkarpov15:
@hasezoey `afterEach()` is intentional here. Not a big deal either way, `afterEach()` makes the tests pass so we'll use that.

hasezoey:
> afterEach() is intentional here.

what is different compared to `beforeEach` that would require a `afterEach` here unlike other test files? (just curious)

> afterEach() makes the tests pass so we'll use that.

from what i can tell, it does not make it pass - that is why i had changed it to match the other test files (because of mocha's environment bleeding) ... maybe add a `before` to clean up before any of the tests in this file run?

PR #12825, opened 2022-12-22
docs(typescript): make note about recommending `strict` mode when using auto typed schemas

vkarpov15:
Re: #12420

**Summary**

So strange. The following script fails to compile without `--strict`.

```ts
import { Schema, InferSchemaType } from 'mongoose';

const sightSchema = new Schema({
    location: { type: [Number], required: false } // all ways to define "location" give the same result
    // location: { type: [Number] }
    // location: [Number]
});

type Sight = InferSchemaType<typeof sightSchema>;

// Property 'location' is missing in type '{ name: string; }' but required in type '{ name: string; location: number[]; }'
const sight: Sight = {
    name: 'Red Square',
};
```

```
$ ./node_modules/.bin/tsc  gh-12420.ts 
gh-12420.ts:13:7 - error TS2741: Property 'location' is missing in type '{ name: string; }' but required in type '{ location: unknown[]; name?: unknown; }'.

13 const sight: Sight = {
         ~~~~~


Found 1 error in gh-12420.ts:13
```

From #12420. This PR makes a note in the docs.

**Examples**


hasezoey:
i think the problem more specifically is `strictNullChecks`, quote:

> When strictNullChecks is false, null and undefined are effectively ignored by the language. This can lead to unexpected errors at runtime.
> When strictNullChecks is true, null and undefined have their own distinct types and you’ll get a type error if you try to use them where a concrete value is expected.

[Typescript: TSConfig Reference](https://www.typescriptlang.org/tsconfig#strictNullChecks)

AbdelrahmanHafez (approved):

hasezoey (approved):

PR #12827, opened 2022-12-23
fix(schema): propagate strictQuery to implicitly created schemas for embedded discriminators

### LINKED ISSUE #12796: Cannot set strictQuery: false globally and have an embedded discriminator
### Prerequisites

### Mongoose version

6.8.0

### Node.js version

16.17.1

### MongoDB server version

4.4

_No response_

### Description

Running the following code:
```javascript

mongoose.set("strictQuery", false);

  test: [{foo: String}]
});

testSchema.path("test").discriminator("Bar", new mongoose.Schema({ bar: String }));

console.log('done');
```

Fails with:
```
Error: Can't customize discriminator option strictQuery (can only modify toJSON, toObject, _id, id, virtuals, methods)
    at merge (/.../node_modules/mongoose/lib/helpers/model/discriminator.js:174:17)
    at discriminator (/.../node_modules/mongoose/lib/helpers/model/discriminator.js:199:3)
    at DocumentArrayPath.discriminator (/.../node_modules/mongoose/lib/schema/documentarray.js:205:12)
...irrelevant stacks omitted
```

### Steps to Reproduce

See above

### Expected Behavior

I should be able to set `strictQuery: false` globally, but I can't because something about this implementation is broken with schemas that create embedded array discriminators.

vkarpov15:
Fix #12796

**Summary**

The root cause of #12796 is that we don't propagate `strictQuery` to implicitly created schemas like we do `strict`. So if global `strictQuery` is false, but the implicitly created schema's `strictQuery` is true. That gets even more confusing because embedded discriminator schemas do pick up the global `strictQuery` default, so with global `strictQuery = false` you get an error every time you create an embedded discriminator.

**Examples**

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1250,6 +1250,9 @@ Schema.prototype.interpretAsType = function(path, obj, options) {
         if (options.hasOwnProperty('strict')) {
           childSchemaOptions.strict = options.strict;
         }
+        if (options.hasOwnProperty('strictQuery')) {
+          childSchemaOptions.strictQuery = options.strictQuery;
+        }
 
         if (this._userProvidedOptions.hasOwnProperty('_id')) {
           childSchemaOptions._id = this._userProvidedOptions._id;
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM

PR #12829, opened 2022-12-23
test(query.test): add write-concern option

hasezoey:
**Summary**

This PR adds a write-concern to a create, in hopes it fixes some inconsistent replica set tests

for example: https://github.com/Automattic/mongoose/actions/runs/3762716559/jobs/6395685798


vkarpov15 (approved):
Thanks :+1:

PR #12830, opened 2022-12-23
Search content by version

vkarpov15:
**Summary**

Will need to backport this change for 5.x for it to take effect. But the idea is that the search page currently hits [this project](https://github.com/mongoosejs/backend) to search a `Content` collection, which is generated by `docs/search.js`. With this change, `Content` will have a `version` parameter, and we'll filter search results based on `version`.

**Examples**

### DIFF
--- a/docs/js/search.js
+++ b/docs/js/search.js
@@ -9,9 +9,12 @@ for (var i = 0; i < pairs.length; ++i) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM

though preferably the fallback version should not be hardcoded (would add more places that would need it replaced on a version change)

PR #12839, opened 2022-12-27
7.0 driver cleanup

vkarpov15:
**Summary**

Make drivers easier to implement, and minimize how hard-coded the MongoDB driver is in our code. Also remove the `require('../driver').get()` calls that tend to confuse bundlers and cause hard-to-debug issues when implementing custom drivers. Drivers are now fully scoped to a single Mongoose instance, minus the `driver.js` file which we still need to remove.

**Examples**


hasezoey:
if i understand correctly, this means a custom driver cannot provide types like ObjectId (etc) anymore and mongoose is now always using directly from `bson`? (from what it seems like, it was not consistent to begin with)

lib/query.js
```diff
@@ -276,7 +276,7 @@ Query.prototype.toConstructor = function toConstructor() {
  * @api public
  */
 
-Query.prototype.clone = function clone() {
+Query.prototype.clone = function() {
```
hasezoey, line 279: consider re-adding a name, will be better for debugging

lib/utils.js
```diff
@@ -904,7 +893,7 @@
       continue;
     }
     if (typeof to[key] === 'undefined') {
-      to[key] = exports.clone(fromObj[key], {
+      to[key] = clone(fromObj[key], {
```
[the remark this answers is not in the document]
hasezoey: i think this alert can be dismissed, ~5 lines above in the loop `specialProperties` are already skipped

lib/helpers/query/handleReadPreferenceAliases.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+module.exports = function handleReadPreferenceAliases(pref) {
+  switch (pref) {
+    case 'p':
```
hasezoey, line 5: are those readPref aliases documented (and in the types)?

hasezoey (approved):
overall looks good to me, though some small notes

lib/query.js
vkarpov15: The problem there is naming this `clone` conflicts with the `clone` import. So not ideal, but not a blocker.

PR #12861, opened 2023-01-02
docs(discriminators): add section about changing discriminator key

### LINKED ISSUE #6087: Add discriminator key to findOneAndUpdate
Feature

If I have the following discriminators:

```
const messageSchema = new Schema({ text: String });
const Message = mongoose.model('Message', messageSchema);

/* ============ */

const attachmentMessageSchema = new Schema({
  type: { type: String, default: 'ATTACHMENT' },
  attachment: String,
});

const AttachmentMessage = Message.discriminator(
  'AttachmentMessage',
  attachmentMessageSchema,
);
```
In order to update an instance of `AttachmentMessage` I would have to write `AttachmentMessage.findOneAndUpdate`. If I write `Message.findOneAndupdate` no changes will occur. 

I think this is because the discriminator key is only added to `find()`, `count()`, and `aggregate`. As the docs state:

> Discriminator models are special; they attach the discriminator key to queries. In other words, find(), count(), aggregate(), etc. are smart enough to account for discriminators.

However, there are cases where the server might not be aware of the discriminator type of the model I want to update. For example, the client might pass a generic message payload:
```
{
  text: 'some_text',
  attachment: <payload>
}
```

Ideally, it would be nice to be able to pass that payload to the following function:

```
updateMessage(id, data) {
  return Message.findByIdAndUpdate(id, data, {
    new: true,
  });
}
```
Could we automatically add the discriminator key to functions like `findOneAndUpdate` to address these cases? I saw a PR in Mongoose Schema extended that seems to model the desired behavior: https://github.com/briankircho/mongoose-schema-extend/pull/53/files

Node: 8.4.0
Mongo: 3.0.7
Mongoose: 4.4.6

vkarpov15:
Fix #6087

**Summary**

Add `overwriteDiscriminatorKey` to discriminator docs, because that option is a bit hard to find.

**Examples**

### DIFF
--- a/docs/discriminators.md
+++ b/docs/discriminators.md
@@ -31,12 +31,25 @@ instances.
 
 ### Discriminator keys
 
-The way mongoose tells the difference between the different
-discriminator models is by the 'discriminator key', which is
-`__t` by default. Mongoose adds a String path called `__t`
-to your schemas that it uses to track which discriminator
-this document is an instance of.
+The way Mongoose tells the difference between the different discriminator models is by the 'discriminator key', which is `__t` by default.
+Mongoose adds a String path called `__t` to your schemas that it uses to track which discriminator this document is an instance of.
 
 ```javascript
 [require:Discriminator keys]
+```
+
+### Updating the discriminator key
+
+By default, Mongoose doesn't let you update the discriminator key.
+`save()` will throw an error if you attempt to update the discriminator key.
+And `findOneAndUpdate()`, `updateOne()`, etc. will strip out discriminator key updates.
+
+```javascript
+[require:Update discriminator key]
+```
+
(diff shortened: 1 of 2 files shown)


docs/discriminators.md
```diff
@@ -31,12 +31,25 @@ instances.
 
 ### Discriminator keys
 
-The way mongoose tells the difference between the different
-discriminator models is by the 'discriminator key', which is
-`__t` by default. Mongoose adds a String path called `__t`
-to your schemas that it uses to track which discriminator
-this document is an instance of.
+The way Mongoose tells the difference between the different discriminator models is by the 'discriminator key', which is `__t` by default.
+Mongoose adds a String path called `__t` to your schemas that it uses to track which discriminator this document is an instance of.
```
hasezoey, line 35: is this meant to be on the same line? (previous line is missing double spaces to make it a new line)

hasezoey (approved):
LGTM

PR #12864, opened 2023-01-03
add type def for Aggreaget#model without arguments

dmshvetsov:
**Summary**

add missing type definition for `Aggregate.prototype.model` function that can be used for set or get Model from an Aggregate object.

According to mongoose documentation `Aggregate.prototype.model` can be used in two ways:

```js
      const aggregate = MyModel.aggregate([{ $match: { answer: 42 } }]);
      aggregate.model() === MyModel; // true
 
      // Change the model. There's rarely any reason to do this.
      aggregate.model(SomeOtherModel);
      aggregate.model() === SomeOtherModel; // true

```

...to get the aggregation object model and to set it.

But `types/aggregate.d.ts` is missing type definition for the get case of `Aggregate.prototype.model` function

This PR aims to fix this.

**Examples**

### DIFF
--- a/test/types/aggregate.test.ts
+++ b/test/types/aggregate.test.ts
@@ -1,4 +1,4 @@
-import { Schema, model, Document, Expression, PipelineStage, Types } from 'mongoose';
+import { Schema, model, Document, Expression, PipelineStage, Types, Model, Aggregate } from 'mongoose';
 import { expectType } from 'tsd';
 
 const schema: Schema = new Schema({ name: { type: 'String' } });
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (commented):
Welcome @dmshvetsov 👋 

Can you please add a test to `tests/typescript` to prevent future regressions?

Thanks!

dmshvetsov:
> Welcome @dmshvetsov 👋
> 
> Can you please add a test to `test/types` to prevent future regressions?
> 
> Thanks!

sure, will do

dmshvetsov:
@AbdelrahmanHafez could you take another look? I've added type tests for `Aggregate.prototype.model`

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

Issue #12791, opened 2022-12-11
Mongoose bulk insert misses critical information


samislam:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

![image](https://user-images.githubusercontent.com/59496975/206900816-9fad19f6-8656-4124-bbf4-33506e669ecf.png)

The picture above shows an example of a request getting sent to the following route:

![image](https://user-images.githubusercontent.com/59496975/206900873-d0a7419a-3264-48b3-a055-6ecfdb5ca476.png)

and the following picture shows what have been inserted to the database in compass (notice how there are three entries):

![image](https://user-images.githubusercontent.com/59496975/206901057-14b74732-6890-4158-ac38-ef1798fe7841.png)

As we know, `Model.create()` accepts an array of objects, or an object. 

In this example, I am sending an array of objects, to insert them.

`Model.create([])` will insert the documents one by one to the database, it doesn't skip the validation part, which is why I chose it. 
and when it finds a document with a validation error, it skips it, and moves to the next one. 
until it finishes, then it reports the errors it encounters.

That's what it **should be**, However it's not exactly working like that.

Note that I have two documents which holds validation errors:

![image](https://user-images.githubusercontent.com/59496975/206901173-ea629de7-159e-4ebb-a672-709ed3ac43e1.png)

However, mongoose is only reporting the first one, it's not reporting the second one, even though it passes by it, and it sees it.

### Motivation

Why this information is critical?

Because on the client side, I have to know which documents got inserted, and which did not. 

In this case, (when I will know which are the ones got inserted and the ones that did not), I can show the client for example that the documents x, y, z has been inserted, while the documents f, g, h did not. So the user can correct his mistake and send the request again.

The current error report is useless, because it only tells _**"there was a validation error"**_, but it doesn't tell you the _"where"_

### Example

The error report should include all the documents which refused to be written to the database in an array.

samislam:
I realized that 
```js
const data = await User.insertMany(req.body)
```
Has exactly the same behavior.
It doesn't only apply to `Model.create()`.
`Model.insertMany()` has the same problem as well.

samislam:
This question is on stackoverflow as well:
https://stackoverflow.com/questions/74760543/mongoose-bulk-insert-misses-critical-information

vkarpov15:
[...]

PR #12866, opened 2023-01-03
fix(model): ensure consistent ordering of validation errors in `insertMany()` with `ordered: false` and `rawResult: true`

vkarpov15:
Re: #12791

**Summary**

#12791 points out a few issues with our error reporting with `create()` and `insertMany()`. We do have a workaround using `insertMany()`, but the workaround does have a potential flaw: currently, `res.mongoose.validationErrors` can report validation errors in any order.

Specifically, the validation errors are in the order that the errors occur, not in the order of the original documents passed in. So if `doc1` and `doc2` both fail validation, `const res = await Model.insertMany([doc1, doc2], { ordered: false, rawResult: true })` can end up with doc2's validation error before doc1's in `res.mongoose.validationErrors` if doc1's validation takes longer than doc2's.

With this change, `res.mongoose.validationErrors` will always be in the same order as the documents passed in. So doc1's validation error will always be before doc2's. So to check whether a document at index `i` was actually inserted, you need to check if `i in res.insertedIds`. To check if there was a MongoDB-specific write error, check if `!(i in res.insertedIds)` 

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3411,7 +3411,8 @@ Model.$__insertMany = function(arr, options, callback) {
   }
 
   const validationErrors = [];
-  const toExecute = arr.map(doc =>
+  const validationErrorsToOriginalOrder = new Map();
+  const toExecute = arr.map((doc, index) =>
     callback => {
       if (!(doc instanceof _this)) {
         try {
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM

PR #12870, opened 2023-01-03
Correctly infer string enums on const arrays

### LINKED ISSUE #12869: InferSchemaType: string enums not being inferred for const arrays
### Prerequisites

### Mongoose version

6.8.2

### Node.js version

16.7.1

### MongoDB server version

4.4.6

_No response_

### Description

String enums are being inferred as type `string` instead of a more specific union type containing the enum's values.

### Steps to Reproduce

```typescript
import { Schema } from "mongoose";
import type { InferSchemaType } from "mongoose";

const test = new Schema({
  foo: {
    type: String,
    required: true,
    enum: ["foo", "bar"] as const,
  }
});

export type testSchema = InferSchemaType<typeof test>;
```

This results in the following type:
```typescript
type testSchema = {
    foo: string;
}
```

When it should be:
```typescript
type testSchema = {
    foo: "foo"|"bar";
}
```

### Expected Behavior

_No response_

JavaScriptBach:
**Summary**

Fixes #12869

Not a TypeScript expert, but I followed the example in #12463 and added a test case demonstrating this change works. I think what is happening here is that a readonly array cannot extend a regular array, so our types here need to use ReadonlyArray to handle both const and regular arrays.

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -982,3 +982,23 @@ function gh12782() {
 function gh12816() {
   const schema = new Schema({}, { overwriteModels: true });
 }
+
+function gh12869() {
+  const dbExampleConst = new Schema(
+    {
+      active: { type: String, enum: ["foo", "bar"] as const, required: true }
+    }
+  );
+
+  type ExampleConst = InferSchemaType<typeof dbExampleConst>;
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

Issue #12807, opened 2022-12-16
unknown properties in schema when set timestams to true


Code-Mythos:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.8.0

### Node.js version

16.15.1

### MongoDB server version

6.0

### Typescript version (if applicable)

4.9.4

### Description

In version 6.8.0, when I set the timestamps option to true in a schema, the created model has properties with unknown types that will cause a lot of problems in implementing the application. However, when I switch back to version 6.7.5 the problem disappears.
.
.
**Version 6.8.0:**
```
(property) Users: Model<{
    createdAt: NativeDate;
    updatedAt: NativeDate;
} & {
    email?: unknown;
    password?: unknown;
    name?: unknown;
}, {}, {
    // This is a custom method in the schema
    comparePassword: (this: Document<unknown, any, FlatRecord<{
        email: string;
        password: string;
        name: string;
...
```
.
.
**When switched back to version 6.7.5 without any modification:**
```
(property) Users: Model<{
    email: string;
    password: string;
    name: string;
}, {}, {
    comparePassword: (this: Document<unknown, any, FlatRecord<{
        email: string;
        password: string;
        name: string;
...
```

### Steps to Reproduce

Considering you are using typescript:
1. Install version 6.8.0 : npm i mongoose@6.8.0
2. Create a schema with some properties and set the timestamps option to true.
3. Create a connection using mongoose.createConnection.
4. Create a model for the schema using the connection.model.
5. Hover over the created model to see the type of the model.

### Expected Behavior

**Expected behavior:**
```
(property) Users: Model<{
    createdAt: NativeDate;
    updatedAt: NativeDate;
} & {
    email: string;
    password: string;
    name: string;
}, {}, {
    // This is a custom method in the schema
    comparePassword: (this: Document<unknown, any, FlatRecord<{
        email: string;
        password: string;
        name: string;
...
```
Or:
```
(property) Users: Model<{
    email: string;
    password: string;
    name: string;
}, {}, {
    // This is a custom method in the schema
    comparePassword: (this: Document<unknown, any, FlatRecord<{
        email: string;
        password: string;
        name: string;
...
```

IslandRhythms:
![image](https://user-images.githubusercontent.com/39510674/208138370-40f6c475-af76-431f-b4a7-9f70ebe4b0f8.png)
```typescirpt
[...]

PR #12871, opened 2023-01-04
fix(types): avoid inferring timestamps if `methods`, `virtuals`, or `statics` set

vkarpov15:
Re: #12807

**Summary**

Really not my favorite PR. I haven't been able to figure out exactly why this issue happens, I suspect because there's some self-referential issues that pop up because methods, virtuals, and statics use the inferred document type, and the inferred document type relies on schema options. But this is at least sufficient to unblock anyone using timestamp type inference.

We should consider deprecating `timestamps` type inference for 7.0 if we don't come up with a better solution. I'm going to take some time to come up with a minimal repro script for the TypeScript team. I don't expect this to end up being a bug that TS is willing to fix. But might as well so we can at least be certain that we're trying to do something TypeScript doesn't support.

**Examples**


hasezoey (approved):
LGTM

i personally think it is somehow (again) related to https://github.com/Automattic/mongoose/blob/e9b5eed230af467af37a9c3ad290d1fda2b3dfb7/types/inferschematype.d.ts#L49-L61 
but i havent been able to debug this further because typescript does not show the whole type or allows debugging type stages (like a break-point)

vkarpov15:
I think so as well. There's some circular reference stuff that's probably confusing TypeScript. In the meantime, I'll merge this to unblock.

PR #12878, opened 2023-01-06
chore(promises): removed support for custom promise libraries

### LINKED ISSUE #12872: Remove support for custom promise libraries
### Prerequisites

### Issue

MongoDB Node driver is dropping support for `promiseLibrary` option, so we should remove support for `mongoose.Promise`. Worth dropping on our end. You can always switch to bluebird by doing something like `global.Promise = require('bluebird')`

lpizzinidev:
closes #12872


lib/helpers/promiseOrCallback.js
```diff
@@ -35,7 +34,7 @@ module.exports = function promiseOrCallback(callback, fn, ee, Promise) {
     }
   }
 
-  Promise = Promise || PromiseProvider.get();
+  Promise = Promise || global.Promise || require('bluebird');
```
Uzlopak, line 37: What happens if bluebird is not installed? Crash gracefully?

Uzlopak (commented):
In test/connection.test.js we use the package q and bluebird. I wonder if we can remove those tests too? @vkarpov15

test/docs/promises.test.js
```diff
@@ -21,8 +21,6 @@ describe('promises docs', function () {
   });
 
   after(async function () {
-    mongoose.Promise = global.Promise;
```
hasezoey, line 24: i personally think the default one should be cached outside of the individual tests and reset in a hook, because tests may fail in-between and so dont reset properly if the reset is after the failed point

hasezoey (changes requested):
mostly looks good to me, though there are some test things that should maybe be changed

also i think this change should target the next major version instead of `master` (/ 6.x)

test/docs/promises.test.js
```diff
@@ -21,6 +23,8 @@ describe('promises docs', function () {
   });
 
   after(async function () {
+    global.Promise = defaultPromise;
```
hasezoey, line 26: this should work in its current way, though personally i would put this in a `afterEach` instead of just a `after`

hasezoey (approved):
LGTM

docs/promises.md
```diff
@@ -54,7 +54,7 @@ better stack traces.
 
 If you're an advanced user, you may want to plug in your own promise
 library like [bluebird](https://www.npmjs.com/package/bluebird). Just set
-`mongoose.Promise` to your favorite
+`global.Promise` to your favorite
```
vkarpov15, line 57: I'd recommend just removing this section of the docs entirely. We don't want to support or encourage overwriting `global.Promise`. If user does that, it's up to them.

lib/helpers/promiseOrCallback.js
vkarpov15, line 37: Please get rid of `require('bluebird')` here. In general, just get rid of the `Promise` param.

There's no need to import bluebird here because Mongoose doesn't support any JavaScript runtimes that don't have native promises.

test/docs/promises.test.js
```diff
@@ -168,12 +172,12 @@ describe('promises docs', function () {
     }
     // acquit:ignore:end
     // Use bluebird
-    mongoose.Promise = require('bluebird');
+    global.Promise = require('bluebird');
```
vkarpov15, line 175: Remove this section entirely. We don't want to encourage setting `global.Promise`. Plus this will help us get rid of bluebird as a dep.

vkarpov15 (changes requested):

hasezoey (commented):
LGTM, i think that `bluebird` and `q` can now be removed from the `package.json`

vkarpov15 (approved):

CutiePi:
Would've been a neat inclusion to the migration DOC 😅 

I went from v5 -> v7 and found out the hard way lol

hasezoey:
just as a note, if you still want to use custom promises, you should be able to do:
```ts
global.Promise = YourCustomPromise;
```

PR #12879, opened 2023-01-06
fix(types): add "options" parameter for single "create" call

### LINKED ISSUE #12877: Model.create() not allowing me to pass SaveOptions as the second argument
### Prerequisites

### Mongoose version

6.7.2

### Node.js version

16.18.1

### MongoDB server version

5.0.14

4.9.3

### Description

Similar to this: https://github.com/Automattic/mongoose/issues/10588 but this person's issue was due to typescript version being 3.x.

While trying  insert a new mongoose model into a collection, I'm receiving the following error:

```
The last overload gave the following error.
    Argument of type 'SaveOptions' is not assignable to parameter of type 'Callback<Customer & Document<any, any, any> & { _id: ObjectId; }>'.
      Type 'SaveOptions' provides no match for the signature '(error: NativeError, result: Customer & Document<any, any, any> & { _id: ObjectId; }): void'.

168       options,
          ~~~~~~~

  node_modules/mongoose/types/models.d.ts:185:5
    185     create<DocContents = AnyKeys<T>>(doc: T | DocContents, callback: Callback<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>): void;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The last overload is declared here.
```

From what I understand, I'm trying to call the create function with this declaration: `create<DocContents = AnyKeys<T>>(docs: Array<T | DocContents>, options?: SaveOptions): Promise<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>[]>;`
 since i am passing the optional saveOptions function alongside but for some reason mongoose is matching it with the last overload: `create<DocContents = AnyKeys<T>>(doc: T | DocContents, callback: Callback<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>): void;` and the types are not matching for obvious reasons. 

### Steps to Reproduce

Create a schema and a model. Try to use the `create` function to insert a model into the schema. Sample code:

```
import { SaveOptions } from 'mongoose';
const objToInsert = {}
const options: SaveOptions = {}
model.create(objToInsert, options)
```

Output will be something like this (also mentioned above):
```
The last overload gave the following error.
    Argument of type 'SaveOptions' is not assignable to parameter of type 'Callback<Customer & Document<any, any, any> & { _id: ObjectId; }>'.
      Type 'SaveOptions' provides no match for the signature '(error: NativeError, result: Customer & Document<any, any, any> & { _id: ObjectId; }): void'.

168       options,
          ~~~~~~~

  node_modules/mongoose/types/models.d.ts:185:5
    185     create<DocContents = AnyKeys<T>>(doc: T | DocContents, callback: Callback<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>): void;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The last overload is declared here.
```

### Expected Behavior

For mongoose to allow me to pass SaveOptions to the model.create(obj, options) function.

hasezoey:
**Summary**

This PR adds the `options` parameter to single `Model.create()` calls
This PR also moves the `...docs: Array<T | DocContents>` overload as the last option

fixes #12877

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -11,7 +11,8 @@ import {
   HydratedDocument,
   HydratedDocumentFromSchema,
   Query,
-  UpdateWriteOpResult
+  UpdateWriteOpResult,
+  SaveOptions
 } from 'mongoose';
 import { expectAssignable, expectError, expectType } from 'tsd';
 import { AutoTypedSchemaType, autoTypedSchema } from './schema.test';
@@ -568,4 +569,12 @@ function gh12573ModelAny() {
   expectType<any>(doc);
   const { fieldA } = doc;
   expectType<any>(fieldA);
+}
+
+function gh12877() {
+  const TestModel = model('Test', new Schema({}));
+
+  const objToInsert = {};
+  const options: SaveOptions = {};
+  TestModel.create(objToInsert, options);
 }
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -188,9 +188,10 @@ declare module 'mongoose' {
     create<DocContents = AnyKeys<T>>(docs: Array<T | DocContents>, options?: SaveOptions): Promise<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>[]>;
     create<DocContents = AnyKeys<T>>(docs: Array<T | DocContents>, options?: SaveOptions, callback?: Callback<Array<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>>): Promise<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>[]>;
     create<DocContents = AnyKeys<T>>(docs: Array<T | DocContents>, callback: Callback<Array<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>>): void;
-    create<DocContents = AnyKeys<T>>(doc: DocContents | T): Promise<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>;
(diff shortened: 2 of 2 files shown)


vkarpov15:
This syntax isn't supported. Because there's no way to tell the difference between `Model.create(obj1, obj2)` and `Model.create(obj1, options)`. So calling `Model.create(obj1, options)` treats `options` as a document to create, which isn't what people typically expect.

From [these docs](https://mongoosejs.com/docs/api/model.html#model_Model-create):

image


"To specify options, docs must be an array, not a spread. See [Model.save](https://mongoosejs.com/docs/api/model.html#model_Model-save) for available options."

Only potential workaround would be to export `SaveOptions` and make `create()` explicitly check if 2nd arg is `instanceof SaveOptions`. But we don't currently do that. So in order to support this, we'd need to do some extra work. But so far, hasn't been worth it because `Model.create([obj], options)` is more concise than `Model.create(obj, new SaveOptions(options))`.

hasezoey:
didnt know that and didnt check it (which i should have done), thanks for checking

PR #12884, opened 2023-01-07
Correctly infer types on document arrays

### LINKED ISSUE #12882: InferSchemaType: array of subdocuments with a "type" key is not correctly inferred
### Prerequisites

### Mongoose version

6.8.2

### Node.js version

16.17.1

### MongoDB server version

4.4

_No response_

### Description

If I have an array of objects containing a key named "type": InferSchemaType does not correctly infer it.

### Steps to Reproduce

```typescript
const schema = new Schema({
  fooArray: {
    type: [{
      type: {
        type: String,
        required: true,
      },
      foo: {
        type: Number,
        required: true,
      }
    }],
    required: true,
  }
})
type schemaT = InferSchemaType<typeof schema>;
```

### Expected Behavior

I should get the following type inferred:
```typescript
type schemaT = {
    fooArray: {
        type: string;
        foo: number;
    }[];
}
```

Instead I get:
```typescript
type schemaT = {
    fooArray: {
        type?: string | undefined;
        required?: unknown;
    }[];
}
```

JavaScriptBach:
**Summary**

Fix #12882.

The key insight here is that arrays of scalars and arrays of objects have different type inference logic. For example:

1. `type: [{type: String, required: true}]` is an array of strings.
2. `type: [{type: "String", required: true}]` is also an array of strings.
3. `type: [{type: {foo: {type: String}, required: true}]` is an array of objects that contains a field named `type`. Note here that `type` must be interpreted as a real key, not a type key. 

So if we have an array, we need to check if it's an array of scalars or objects. This is done as follows:
1. If the object has a type key that isn't callable and isn't a string, it's an object, so we need to call `ObtainDocumentType` to make sure that its keys are interpreted correctly. (i.e. if the object has a key called `type`, this is a real key, not a type key)
2. In all other cases, we continue to assume that the array is a scalar.

**Examples**

Adds tests.

### DIFF
--- a/types/inferschematype.d.ts
+++ b/types/inferschematype.d.ts
@@ -1,4 +1,3 @@
-import { OperationTime } from 'mongodb';
 import {
   Schema,
   InferSchemaType,
--- a/types/inferschematype.d.ts
+++ b/types/inferschematype.d.ts
@@ -1,6 +1,8 @@
+import { OperationTime } from 'mongodb';
 import {
   Schema,
   InferSchemaType,
+  ScalarSchemaDefinition,
   SchemaType,
   SchemaTypeOptions,
   TypeKeyBaseType,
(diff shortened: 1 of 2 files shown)


types/inferschematype.d.ts
```diff
@@ -179,8 +177,29 @@ type PathEnumOrString<T extends SchemaTypeOptions<string>['enum']> = T extends R
...
 type ResolvePathType<PathValueType, Options extends SchemaTypeOptions<PathValueType> = {}, TypeKey extends string = DefaultSchemaOptions['typeKey']> =
   PathValueType extends Schema ? InferSchemaType<PathValueType> :
-    PathValueType extends (infer Item)[] ? IfEquals<Item, never, any[], Item extends Schema ? Types.DocumentArray<ObtainDocumentPathType<Item, TypeKey>> : ObtainDocumentPathType<Item, TypeKey>[]> :
-      PathValueType extends ReadonlyArray<infer Item> ? IfEquals<Item, never, any[], Item extends Schema ? Types.DocumentArray<ObtainDocumentPathType<Item, TypeKey>> : ObtainDocumentPathType<Item, TypeKey>[]> :
+    PathValueType extends (infer Item)[] ?
+      IfEquals<Item, never, any[], Item extends Schema ?
+        // If Item is a schema, infer its type.
+        Types.DocumentArray<InferSchemaType<Item>> :
+        Item extends Record<TypeKey, any>?
+          Item[TypeKey] extends Function ?
```
vkarpov15, line 185: This looks like it also needs to handle the case where `Item[TypeKey]` is a string. `type: 'String'` is equivalent to `type: String`.

vkarpov15 (approved):
LGTM, just one minor question/suggestion.

PR #12889, opened 2023-01-08
fix(collection): handle creating model when connection disconnected with bufferCommands = false

vkarpov15:
**Summary**

If you create a new model while Mongoose is disconnected, and `bufferCommands` is false, you'll get the following error:

```
MongooseError: Cannot call `tests.findOne()` before initial connection is complete if `bufferCommands = false`. Make sure you `await mongoose.connect()` if you have `bufferCommands = false`.
```

That is unexpected. You should never get that error after calling `mongoose.connect()`, even with `bufferCommands` disabled.

**Examples**

### DIFF
--- a/lib/drivers/node-mongodb-native/collection.js
+++ b/lib/drivers/node-mongodb-native/collection.js
@@ -43,9 +43,7 @@ Object.setPrototypeOf(NativeCollection.prototype, MongooseCollection.prototype);
  */
 
 NativeCollection.prototype.onOpen = function() {
-  if (this.collection == null) {
-    this.collection = this.conn.db.collection(this.name);
-  }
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
LGTM, though i am not quite sure what this is solving

i am currently understating it as "creating a model (mongoose.model) after the connection was connected once but is not at the point of model creation while `bufferCommands` is false"?

vkarpov15:
Here's a quick overview of when this issue happens:

1. `mongoose.connect()` succeeds, `bufferCommands` is set to false
2. Mongoose loses connectivity to the database. Perhaps the MongoDB server went down, or there's a network outage.
3. Create a new model using `mongoose.model()`
4. Run `await Model.findOne()`, get error about initial connection

PR #12890, opened 2023-01-08
Close all connections in all tests, also change some tests to be working stand-alone

hasezoey:
**Summary**

This PR updates some tests / test files to close all connections that were opened, also changes some test files to be working stand-alone

though i am not really proud of some of the connection caching done with `connectionsToClose`

some more detailed notes:
- all test files *except* `connection.test.js` and `index.test.js` have been updated with `connectionsToClose` or a simple `mongoose.disconnect` instead of not closing the connections
- some spaces have been added between some tests
- many `test/helpers` test files have been updated to work when tested stand-alone (only testing a single test file)
- change some import order in `test/helpers` to be consistent across `test/helpers` test files
- some tests have been updated to use the existing connection instead of creating a new one (when it was not about using a new connection)

these changes are required for a follow-up PR i am doing to integrate MMS to the deno tests too (instead of manually downloading and starting mongodb), because otherwise it would fail with the following when the single-instance is stopped:

```txt
error: Uncaught Error: connect ECONNREFUSED 127.0.0.1:36683 - Local (undefined:undefined)
    Error.captureStackTrace(err);
          ^
    at __node_internal_captureLargerStackTrace (https://deno.land/std@0.171.0/node/internal/errors.ts:113:11)
    at __node_internal_exceptionWithHostPort (https://deno.land/std@0.171.0/node/internal/errors.ts:295:12)
    at TCPConnectWrap._afterConnect [as oncomplete] (https://deno.land/std@0.171.0/node/net.ts:369:16)
    at TCP.afterConnect (https://deno.land/std@0.171.0/node/internal_binding/connection_wrap.ts:71:11)
    at https://deno.land/std@0.171.0/node/internal_binding/tcp_wrap.ts:375:16
error Command failed with exit code 1.
```

### DIFF
--- a/test/collection.test.js
+++ b/test/collection.test.js
@@ -8,8 +8,15 @@ const assert = require('assert');
 const mongoose = start.mongoose;
 
 describe('collections:', function() {
+  const connectionsToClose = [];
+
+  after(async function() {
+    await Promise.all(connectionsToClose.map((v) => v.close()));
+  });
+
   it('should buffer commands until connection is established', function(done) {
     const db = mongoose.createConnection();
+    connectionsToClose.push(db);
     const collection = db.collection('test-buffering-collection');
     let connected = false;
     let insertedId = undefined;
@@ -43,6 +50,7 @@ describe('collections:', function() {
 
   it('returns a promise if buffering and no callback (gh-7676)', function(done) {
     const db = mongoose.createConnection();
+    connectionsToClose.push(db);
     const collection = db.collection('gh7676');
 
     const promise = collection.insertOne({ foo: 'bar' }, {})
@@ -145,6 +153,7 @@ describe('collections:', function() {
 
   it('buffers for sync methods (gh-10610)', function(done) {
     const db = mongoose.createConnection();
+    connectionsToClose.push(db);
     const collection = db.collection('gh10610');
 
     collection.find({}, {}, function(err, res) {
--- a/test/connection.test.js
+++ b/test/connection.test.js
@@ -56,6 +56,7 @@ describe('connections:', function() {
       }).asPromise();
 
       assert.strictEqual(conn.config.autoIndex, false);
+      await conn.close();
     });
 
     it('with autoCreate (gh-6489)', async function() {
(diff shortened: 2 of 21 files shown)


test/connection.test.js
```diff
@@ -764,29 +773,29 @@ describe('connections:', function() {
...
 
-    it('cache connections to the same db', function() {
+    it('cache connections to the same db', function(done) {
       const db = start();
       const db2 = db.useDb(start.databases[1], { useCache: true });
       const db3 = db.useDb(start.databases[1], { useCache: true });
 
       assert.strictEqual(db2, db3);
-      db.close();
+      db.close(done);
```
vkarpov15, line 782: Let's not add extra `done()` usage, we should avoid using `done()`. `return db.close()` should be fine because `db.close()` returns promise if no callback specified.

test/connection.test.js
```diff
@@ -764,29 +773,29 @@ describe('connections:', function() {
...
     describe('when using standard authentication', function() {
       describe('when username and password are undefined', function() {
-        it('should return false', function() {
+        it('should return false', function(done) {
           const db = mongoose.createConnection(start.uri, {});
 
           assert.equal(db.shouldAuthenticate(), false);
 
-          db.close();
+          db.close(done);
```
vkarpov15, line 794: Same, avoid adding done.

test/model.test.js
```diff
@@ -25,6 +25,8 @@ describe('Model', function() {
   let Comments;
   let BlogPost;
 
+  const connectionsToClose = [];
```
vkarpov15, line 28: Let's do `let db = null;` instead and `close()` in an `afterEach()`.

vkarpov15 (changes requested):

test/collection.test.js
```diff
@@ -8,8 +8,15 @@ const assert = require('assert');
 const mongoose = start.mongoose;
 
 describe('collections:', function() {
+  const connectionsToClose = [];
+
+  after(async function() {
```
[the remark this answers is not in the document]
hasezoey: i did this to not modify the test too much, also some tests use multiple connections

hasezoey:
> I'm just going to merge this one and make suggested changes in separate PR. Thanks +1

also good, just wanted to know before how to handle multiple connections in a test with the suggested way, because a single `let db = connection` would not be sufficient and multiple `let db1 = connection; let db2 = connection;` would with even more connections become redundant

PR #12891, opened 2023-01-08
fix(populate): merge instead of overwrite when `match` is on `_id`

### LINKED ISSUE #12834: Unexpected behavior with the populate method leading to memory issues
### Prerequisites

### Mongoose version

6.8.1

### Node.js version

18.12.1

### MongoDB server version

irrelevant to this issue

_No response_

### Description

When trying to use the populate static method (and i suspect the instance method as well) with a match on _id, mongoose completely ignores the ids it is trying to match on in its query, which can lead to a huge and unexpected query results size and issues with the garbage collector.

we've came across this problem when our backend had a sudden problem with memory usage and the garbage collector constantly blocking the event loop.

The code causing this issue is in https://github.com/Automattic/mongoose/blob/master/lib/helpers/populate/createPopulateQueryFilter.js line 13, it doesn't  add the internal match on `_id` if  one is already provided instead of combining the two

### Steps to Reproduce

```typescript

mongoose.set('debug', true);

const personSchema = new Schema({
  name: String,
  age: Number,
  stories: [{ type: Schema.Types.ObjectId, ref: 'Story' }],
});

const storySchema = new Schema({
  author: { type: Schema.Types.ObjectId, ref: 'Person' },
  title: String,
  fans: [{ type: Schema.Types.ObjectId, ref: 'Person' }],
});

const Story = mongoose.model('Story', storySchema);

function objectIdWithTimestamp(timestamp) {
    if (typeof(timestamp) == 'string') {
        timestamp = new Date(timestamp);
    }
    var hexSeconds = Math.floor(timestamp/1000).toString(16);
   return ObjectId(hexSeconds + "0000000000000000");
}

async function test() {
  const person = await Person.findOne({ });

  await Story.populate(person, {
    path: 'stories',
    match: {
      _id: { $gte: objectIdWithTimestamp(new Date('2020-01-01')) },
    },
  });
}

mongoose
  .connect('mongodb://127.0.0.1:27017/db1')
  .then(test)
  .then(() => mongoose.disconnect());  
```
In this example, the query that would be sent to the database would have only the match on the Id that was provided by the code above, which could return a huge portion of the documents in the affected collection

Here are the printed queries to the console:
```javascript
people.findOne({}, { projection: {} })
stories.find({ _id: { '$gte': new ObjectId("5e0be1000000000000000000") } }, { skip: undefined, limit: undefined, perDocumentLimit: undefined, projection: {}})
```

### Expected Behavior

The expected behavior would be that whatever match conditions are provided on the `_id`, would be combined using `$and`  with the internal match that the populate method would normally produce  `{ _id: {$in: [ /* Array of object ids*/]}}` and spit out a query like this:
```javascript
stories.find(
  {
    $and: [
      { _id: { $gte: new ObjectId('5e0be1000000000000000000') } },
      { _id: { $in: [/* Array of object ids*/] } },
    ],
  },
  {
    skip: undefined,
    limit: undefined,
    perDocumentLimit: undefined,
    projection: {},
  }
);
```

vkarpov15:
Fix #12834

**Summary**

#12834 points out that `match` on `_id` overwrites rather than merges. Which is not a very good API design, because any custom `match` that specifies `_id` will overwrite the `_id` filter specified by `populate()`. This PR fixes that.

**Examples**

### DIFF
--- a/lib/helpers/populate/createPopulateQueryFilter.js
+++ b/lib/helpers/populate/createPopulateQueryFilter.js
@@ -3,6 +3,7 @@
 const SkipPopulateValue = require('./SkipPopulateValue');
 const parentPaths = require('../path/parentPaths');
 const { trusted } = require('../query/trusted');
+const hasDollarKeys = require('../query/hasDollarKeys');
 
 module.exports = function createPopulateQueryFilter(ids, _match, _foreignField, model, skipInvalidIds) {
   const match = _formatMatch(_match);
@@ -13,6 +14,11 @@ module.exports = function createPopulateQueryFilter(ids, _match, _foreignField, 
     if (foreignField !== '_id' || !match['_id']) {
       ids = _filterInvalidIds(ids, foreignSchemaType, skipInvalidIds);
       match[foreignField] = trusted({ $in: ids });
+    } else if (foreignField === '_id' && match['_id']) {
+      const userSpecifiedMatch = hasDollarKeys(match[foreignField]) ?
+        match[foreignField] :
+        { $eq: match[foreignField] };
+      match[foreignField] = { ...trusted({ $in: ids }), ...userSpecifiedMatch };
     }
 
     const _parentPaths = parentPaths(foreignField);
@@ -37,6 +43,11 @@ module.exports = function createPopulateQueryFilter(ids, _match, _foreignField, 
         const foreignSchemaType = model.schema.path(foreignField);
         ids = _filterInvalidIds(ids, foreignSchemaType, skipInvalidIds);
         $or.push({ [foreignField]: { $in: ids } });
+      } else if (foreignField === '_id' && match['_id']) {
+        const userSpecifiedMatch = hasDollarKeys(match[foreignField]) ?
+          match[foreignField] :
+          { $eq: match[foreignField] };
+        match[foreignField] = { ...trusted({ $in: ids }), ...userSpecifiedMatch };
       }
     }
   }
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks! 👍 
I have no idea how I survived not being bitten by this bug all those years. 😄

vkarpov15:
@AbdelrahmanHafez do you find yourself using match with `_id`? I've never done that before, although I can see why it would be tempting.

AbdelrahmanHafez:
Now that I think about it, I don't think I ever did use `match` with `_id`, but I can see a scenario where I would want to find an intersection/subset between documents coming from one place and matching them in a populate.

yassine-cc:
Hey @vkarpov15 , i'm the one who created the issue related to this, thank you for fixing it.

I just want to point out that the test case in the PR wouldn't catch this bug as only the ids that are present get populated 
and so the issue wouldn't affect the length of the populated array  inside the document.

There is a difference between the query results returned by the db and the result that mongoose produces. 

The bug is more about the performance impact of getting a huge query result size back from the db (huge memory usage), instanciating every item in the array, and then iterating over them to match them with the ids inside the array to populate

vkarpov15:
@yassine-cc can you please clarify the issue you're seeing? I unfortunately don't understand what you mean by "only the ids that are present get populated and so the issue wouldn't affect the length of the populated array inside the document"

PR #12894, opened 2023-01-09
fix(types): updated `.lean()` to remove methods from type

### LINKED ISSUE #12880: Update `.lean` type to automatically remove methods from type
### Prerequisites

### Issue

Currently the `.lean` type does not remove methods from a given type, so something like the following is passing in typescript, but will fail at runtime:

```ts
// NodeJS: 18.10.0
// MongoDB: 5.0 (Docker)
// Typescript 4.9.4
import * as mongoose from 'mongoose'; // mongoose@6.8.0

interface IUser {
  username: string;
  test(): void;
}

const UserModel = mongoose.model<IUser>('User', new mongoose.Schema({ username: String }, { methods: { test() {} } }));

(async () => {
  await mongoose.connect(`mongodb://localhost:27017/`, {
    dbName: 'verifyMASTER',
  });

  await UserModel.create({ username: 'hello' });

  const doc = await UserModel.findOne({}).orFail().lean();
  doc.test(); // should give a type error

  await mongoose.disconnect();
})();
```

Reproduction Repository / Branch: https://github.com/typegoose/typegoose-testing/tree/mongooseLeanMethods

Note: the auto-inferred types are not affected, because they properly split all methods defined via the `methods` option into the proper generic

lpizzinidev:
fix #12880

### DIFF
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -234,3 +234,26 @@ async function _11767() {
   // expectError(examFound2Obj.questions[0].populated);
   expectType<string[]>(examFound2Obj.questions[0].answers);
 }
+
+async function gh12880() {
+  interface IUser {
+    username: string;
+    test(): void;
+  }
+
+  const UserSchema = new Schema({ username: String }, { methods: { test() {} } });
+
+  const UserModel = model<IUser>(
+    'User',
+    UserSchema
+  );
+
+  await UserModel.create({ username: 'hello' });
+
+  const doc = await UserModel
+    .findOne({})
+    .lean();
+  if (!doc) return;
+
+  expectError(doc.test());
+}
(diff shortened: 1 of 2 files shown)


vkarpov15 (commented):
We're trying hard to move away from supporting putting methods directly in the document interface, mostly because TypeScript doesn't support stripping out function keys from an object in general. See https://github.com/microsoft/TypeScript/issues/43281 : stripping function keys _does_ work in some cases, but once you add generic interfaces then things break.

hasezoey:
> mostly because TypeScript doesn't support stripping out function keys from an object in general

this PR does not actually strip the functions out, it just changes those to `never`

lean_pr

vkarpov15:
I'm going to close this one for now. We can reconsider it later, but right now TypeScript doesn't support this behavior in general.

PR #12898, opened 2023-01-10
fix(cast): remove empty conditions after strict applied

0x0a0d:
**Summary**

with `strictQuery true`, a query likes:
```js
{$or: [{notInSchema: 'any'}, {age: {$gt: 18}}]}
```
will become
```js
{$or: [{}, {age: {$gt: 18}}]}
```
and `{}` will match every documents

this pull will 
+ remove empty object `{}` from conditions

then 
+ if conditions is an empty array, remove it


vkarpov15 (approved):
Thanks :+1: will merge this into 6.9 branch for release in 6.9.0

PR #12900, opened 2023-01-10
Allow deselecting fields where select is set to true in the schema definition

IslandRhythms:


IslandRhythms:
So I checked and adding the `true` flag to `parseProjection()` is causing the remaining tests to fail. Not sure how you want to handle that.

IslandRhythms:
I misread the tickets and started refactoring some of the other tests but realized my mistake and stopped.

PR #12906, opened 2023-01-13
types(query):  `returnDocument` type safety

AbdelrahmanHafez:
This PR sets type safety `returnDocument` type as `'before' | 'after'` instead of accepting any `string`.

### DIFF
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -304,9 +304,18 @@ async function gh11602(): Promise<void> {
     returnDocument: 'after',
     rawResult: true
   });
+
   expectError(updateResult.lastErrorObject?.modifiedCount);
   expectType<boolean | undefined>(updateResult.lastErrorObject?.updatedExisting);
   expectType<ObjectId | undefined>(updateResult.lastErrorObject?.upserted);
+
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM

though i have just noticed (not related to this PR) that the JSDoc documentation does not list `returnDocument` as "deprecated" or in relation to `new`, but the TSDoc says it is deprecated (reference [Model.findOneAndUpdate](https://mongoosejs.com/docs/api/model.html#model_Model-findOneAndUpdate) (same for `findOneAndReplace` and `findByIdAndUpdate`))

AbdelrahmanHafez:
@hasezoey 
I think you meant `returnOriginal`* is deprecated, `returnDocument` is the modern way of returning the document before/after the update AFAIK.
Am I understanding this correctly?

hasezoey:
>I think you meant returnOriginal* is deprecated, returnDocument is the modern way of returning the document before/after the update AFAIK.
> Am I understanding this correctly?

yes, i misread, thanks for clarifying

vkarpov15:
Just to clarify. `returnOriginal` is deprecated, as well as `new`. `returnDocument` is the new recommended approach for specifying whether `findOneAndUpdate()` returns the document as it was before changes were applied, or after changes were applied.

PR #12909, opened 2023-01-14
Fix context undefined in $elemMatch query

### LINKED ISSUE #12902: TypeError: Cannot read property 'mongooseCollection' of undefined
### Prerequisites

### Mongoose version

6.8.3

### Node.js version

14.7.3

### MongoDB server version

6.0.3

n/a

### Description

When querying for non-existent keys on a nested schema, the following error is thrown `TypeError: Cannot read property 'mongooseCollection' of undefined`.

Seems to be have introduced in https://github.com/Automattic/mongoose/pull/12717 https://github.com/Automattic/mongoose/issues/12703 as this error does not occur in version `6.7.3`

### Steps to Reproduce

```js
'use strict';

mongoose.connect('mongodb://localhost:27017/elemMatchBug');
const conn = mongoose.connection;

const bioSchema = new mongoose.Schema({
	name: { type: String },
}, { _id: false });

const Book = mongoose.model('book', new Schema({
    name: String,
    authors: [{
        bio: bioSchema,
    }]
}));

async function run() {
    assert.strictEqual(mongoose.version, '6.8.3');
    await conn.dropDatabase();


    const book = await new Book({
        name: "Mongoose Fundamentals",
        authors: [{
            bio: {
                name: "Foo Bar",
            }
        }]
    }).save();

    const books = await Book.find({
        name: "Mongoose Fundamentals",
        authors: {
            $elemMatch: {
                "bio.name": { $in: ["Foo Bar"] },
                "bio.location": "Mandurah" // THIS KEY DOES NOT EXIST ON THE SCHEMA
            }
        }
    }).lean();

    assert.strictEqual(books.length, 1);

    return conn.close();
}

run();
```

### Expected Behavior

For the `bio.location` query term to be sent to the DB server and not match anything, as `strictQuery` is set to false.

MohOraby:
fix #12902 caused by `context` being undefined causing TypeError.


AbdelrahmanHafez:
Welcome @MohOraby 👋 
Can you please also add the original test from the issue to verify that the whole flow works?

AbdelrahmanHafez (changes requested):
Welcome @MohOraby 👋
Can you please also add the original test from the issue to verify that the whole flow works?

test/model.test.js
```diff
@@ -8900,6 +8900,40 @@ describe('Model', function() {
       assert.equal(TestModel.staticFn(), 'Returned from staticFn');
     });
   });
+
+  it('returns document with nested schema', async function() {
```
AbdelrahmanHafez, line 8904:
```suggestion
  it('returns document with nested schema (gh-12902)', async function() {
```

test/cast.test.js
```diff
@@ -193,4 +193,24 @@ describe('cast: ', function() {
       name: 'foo'
     });
   });
+
+  it('returns cast value with context undefined for nested schema', function() {
+    const nested = new Schema({ _id: Number }, {});
```
AbdelrahmanHafez, line 198:
```suggestion
  it('returns cast value with context undefined for nested schema (gh-12902)', function() {
    const nested = new Schema({ _id: Number });
```

MohOraby:
Hello @vkarpov15 👋🏻
When trying to run the tests, deno tests fail with unexpected results
unless specifying `{ strictQuery: true }` the casted query is returning properties that aren't part of the schema
I think `strictQuery` is defaulting to false in that case

vkarpov15:
Closing this one in favor of #12915.

PR #12912, opened 2023-01-16
docs(typescript): add notes about virtual context to Mongoose 6 migration and TypeScript virtuals docs

### LINKED ISSUE #12806: Issues when migrating from 5.12.3 to 6.8.0
### Prerequisites

### Mongoose version

6.8.0

### Node.js version

16.10.0

### MongoDB version

5.0.14

### Operating system

None

_No response_

### Issue

I've completed the migration of my code from using Mongoose 5.12.3 to using 6.8.0.

I've followed the guidelines in [this guide](https://mongoosejs.com/docs/migrating_to_6.html), and everything has worked as expected except from a couple of things:
### Usage of `this` in virtual functions when also using discriminators.
I have a main schema that contains some basic values and two schemas that inherit from it. Let's say, for the sake of this issue, that I have a `Payments` collection, and two types of Payments that inherit from the main `Payment` schema, `BankPayment` and `CashPayment`.

Up until this point, I would structure this as follows:
```ts
export const PaymentCategoryArray = [
    "CashPayment",
    "BankPayment",
] as const;
export type PaymentCategoryType = typeof PaymentCategoryArray[number];

interface PaymentInterface {
    owner: mongoose.Types.ObjectId;
    category: PaymentCategoryType;
}

interface BankPaymentInterface extends PaymentInterface {
    bankReference: string;
}

interface CashPaymentInterface extends PaymentInterface {
    cashAmount: number
}

export interface PaymentDocument extends PaymentInterface, Document {}

export interface BankPaymentDocument extends BankPaymentInterface, PaymentDocument {}

export interface CashPaymentDocument extends CashPaymentInterface, PaymentDocument {}

/**
 * SCHEMAS
 */
const paymentSchema = new mongoose.Schema(
    {
        owner: {
            ref: "User",
            required: true
        }
    },
    { discriminatorKey: "category" }
);
const bankPaymentSchema = new mongoose.Schema(
    {
        bankReference: { type: String, required: true },
    }
);
const cashPaymentSchema = new mongoose.Schema(
    {
        cashAmount: { type: Number }
    }
);

export const Payment = mongoose.model<PaymentDocument>("Payment", paymentSchema);
export const BankPayment = Payment.discriminator<BankPaymentDocument>("BankPayment", bankPaymentSchema);
export const CashPayment = Payment.discriminator<CashPaymentDocument>(
    "CashPayment",
    cashPaymentSchema
);
```

And up until this point, I could implement virtual methods such as this:
```ts
bankPaymentSchema.virtual("displayOwner").get(function (): string {
    return this.owner.toString();
});
```

However, when we migrated to 6.8.0, this started complaining with the following error:
```
TS2339: Property 'owner' does not exist on type 'Document  & { bankReference: string; } & { _id: ObjectId; }'.
```

And it seems like I am forced to cast the above like so:
```ts
bankPaymentSchema.virtual("displayOwner").get(function (): string {
    const document = this as PaymentInterface;
    return document.owner;
});
```

But even this fails with:
```
TS2352: Conversion of type 'Document<unknown, any, { bankReference: string; }> & { bankReference: string; } & { _id: ObjectId; }' to type 'PaymentInterface' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.   Type 'Document<unknown, any, { bankReference: string; }> & { bankReference: string; } & { _id: ObjectId; }' is missing the following properties from type 'PaymentInterface': owner, category
```

I am trying to understand what `this` is in the context of a virtual method implementation such as the above and how I can work around this.

### Population of `_id` in all subdocuments
It seems like from the moment I migrated to 6.8.0, I have to specify `_id: false` in all the `object` types in my schemas. Is there a way to default to this behaviour? I don't have any use cases I'd want a `_id` field in an object in my schema.

vkarpov15:
Fix #12806

**Summary**

Couple of quick notes in the docs re: #12806. Mostly make a note of a potentially breaking change, and also highlight the `virtual<MyTypeHere>` escape hatch for when types are not quite right.

**Examples**

### DIFF
--- a/docs/typescript/virtuals.md
+++ b/docs/typescript/virtuals.md
@@ -83,3 +83,19 @@ schema.virtual('fullName').get(function() {
   return `${this.firstName} ${this.lastName}`;
 });
 ```
+
+### Override the Type of `this` in Your Virtual
+
+In case the value of `this` in your virtual is incorrect for some reason, you can always override it using the generic parameter to `virtual()`.
+
+```ts
+interface MyCustomUserDocumentType {
+  firstName: string;
+  lastName: string;
+  myMethod(): string;
+}
+
+schema.virtual<MyCustomUserDocumentType>('fullName').get(function() {
+  return this.method(); // returns string
+});
+```
--- a/docs/migrating_to_6.md
+++ b/docs/migrating_to_6.md
@@ -511,3 +511,22 @@ The following legacy types have been removed:
 * `HookDoneFunction`
 * `SchemaTypeOpts`
 * `ConnectionOptions`
+
+Mongoose 6 infers the document's type for `this` in virtual getters and setters.
+In Mongoose 5.x, `this` would be `any` in the following code.
+
+```ts
+schema.virtual('myVirtual').get(function() {
+  this; // any in Mongoose 5.x
+});
+```
+
+In Mongoose 6, `this` will be set to the document type.
+
+```ts
+const schema = new Schema({ name: String });
+
+schema.virtual('myVirtual').get(function() {
+  this.name; // string
+});
+```


docs/typescript/virtuals.md
```diff
@@ -82,4 +82,20 @@ const schema = new Schema<UserDoc, UserModel, UserVirtuals>({ // <-- and here
 schema.virtual('fullName').get(function() {
   return `${this.firstName} ${this.lastName}`;
 });
+```
+
+### Override the Type of `this` in Your Virtual
+
+In case the value of `this` in your virtual is incorrect for some reason, you can always override it using the generic parameter to `virtual()`.
```
hasezoey, line 89:
```suggestion
In case the value of `this` in your virtual is incorrect for some reason, you can always override it using the generic parameter in the `virtual()` function.
```

change because i think `to virtual()` does not sound correct

hasezoey (approved):
LGTM, with a minor spelling suggestion

PR #12918, opened 2023-01-17
Change deno tests to also use MMS

hasezoey:
**Summary**

Follow-up to #12890

This PR changes the deno tests and CI to also use MMS for easier testing

~~WIP because:~~
- ~~MMS uses `get-port` which uses `net.createServer().unref()` but `TCP.unref()` is currently still a unstable function, see https://github.com/denoland/deno/issues/17307~~

Deno permissions changes:
- added `--allow-write` because MMS needs to save the downloaded binary and create the folders
- ~~added `--unstable` because `TCP.unref` is currently still a unstable function (and maybe some other unstable function, but i have not passed further than a `TCP.unref` error)~~


hasezoey:
note that the deno test may suddenly fail randomly, see https://github.com/Automattic/mongoose/actions/runs/3940909452/jobs/6742619925
i cannot locally reproduce this (particular) connection refused error and so makes it hard to locate where the connection does not get properly closed

also this error will always happen in deno, because in there the tests and the mocha instance running share the same context from what i can tell, so the connections that are not closed will still be existing before the instance is shut-down and so throw errors (and are hard to locate where the connection was created). in normal mocha (non-deno tests) it works because the tests themself and the fixtures run in different context (from what i can tell) so the tests are closed (with connections) before the instance stop is run and so do not cause errors

hasezoey:
Update: `TCP.unref` had the stabilization PR merged and will (probably) be part of deno `1.30.0`
https://github.com/denoland/deno/pull/17477

hasezoey:
deno 1.30.0 is released, which stabilizes `TCP.unref()`, and mentioned updates have been done in d9a6de728a45de99e0e4733d4175a7a85c14c523

vkarpov15 (approved):
Thanks :+1:

PR #12919, opened 2023-01-17
test: some cleanup re: #12890

vkarpov15:
**Summary**

Applied some suggestions from #12890. Mostly getting rid of unnecessary `connectionsToClose` - left it in one case where it makes sense. Also refactored out some `Model.watch()` cases, and fixed a potential bug due to the MongoDB node driver continuing to poll after change stream is closed.

**Examples**

### DIFF
--- a/test/collection.capped.test.js
+++ b/test/collection.capped.test.js
@@ -19,15 +19,12 @@ const Schema = mongoose.Schema;
 describe('collections: capped:', function() {
   let db;
 
-  const connectionsToClose = [];
-
-  before(function() {
-    db = start();
-  });
-
-  after(async function() {
+  afterEach(async function() {
+    if (db == null) {
+      return;
+    }
     await db.close();
-    await Promise.all(connectionsToClose.map((v) => v.close()));
+    db = null;
   });
 
   it('schemas should have option size', function() {
@@ -40,6 +37,8 @@ describe('collections: capped:', function() {
 
   it('creation', async function() {
     this.timeout(15000);
+
+    db = start();
 
     await db.dropCollection('Test').catch(() => {});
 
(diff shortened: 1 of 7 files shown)


test/collection.capped.test.js
```diff
@@ -55,7 +54,6 @@ describe('collections: capped:', function() {
 
   it('skips when setting autoCreate to false (gh-8566)', async function() {
     const db = start();
```
hasezoey, line 56:
```suggestion
    db = start();
```

test/collection.test.js
```diff
@@ -8,15 +8,18 @@ const assert = require('assert');
 const mongoose = start.mongoose;
 
 describe('collections:', function() {
-  const connectionsToClose = [];
+  let db = null;
 
-  after(async function() {
-    await Promise.all(connectionsToClose.map((v) => v.close()));
+  this.afterEach(async function() {
```
hasezoey, line 13:
```suggestion
  afterEach(async function() {
```

i dont think `this.` is required here, removing it to make it consistent with other test files

test/model.test.js
```diff
@@ -4336,7 +4326,6 @@ describe('Model', function() {
   it('save max bson size error with buffering (gh-3906)', async function() {
     this.timeout(10000);
     const db = start({ noErrorListener: true });
```
hasezoey, line 4328: from what i can tell, in this file / tests the connections do not get properly closed if the test is not successful (which `connectionsToClose` would solve)

hasezoey (changes requested):
mostly looks good to me, though there are some places that have not been changed yet, or were changed back to not properly close in case of test failure

IslandRhythms (approved):
LGTM

PR #12928, opened 2023-01-19
Island rhythms/sync schema options for discriminators

IslandRhythms:
So moving [this line](https://github.com/Automattic/mongoose/blob/7.0/lib/helpers/model/discriminator.js#L181) above the for loop causes [this test](https://github.com/Automattic/mongoose/blob/7.0/test/model.discriminator.test.js#L287) to fail. It also doesn't solve the issue. The if statement that was added cannot be appended to the check in the for loop for deep equality or it would also cause that same test to fail.


lib/helpers/model/discriminator.js
```diff
@@ -153,31 +153,36 @@ module.exports = function discriminator(model, name, schema, tiedValue, applyPlu
...
-
     const toJSON = schema.options.toJSON;
     const toObject = schema.options.toObject;
     const _id = schema.options._id;
     const id = schema.options.id;
 
     const keys = Object.keys(schema.options);
     schema.options.discriminatorKey = baseSchema.options.discriminatorKey;
-
+    const hasUserProvidedTypeKey = schema._userProvidedOptions.typeKey != null && baseSchema._userProvidedOptions.typeKey != null;
```
vkarpov15, line 163: This is a step in the right direction, but I'd like for you to generalize this check for `typeKey` to all schema options. So if user didn't specify other options in discriminator schema (capped, etc.) then use the base schema's options by default.

vkarpov15 (changes requested):

lib/helpers/model/discriminator.js
IslandRhythms: Making the changes you requested causes the test mentioned earlier in the thread from failing.

vkarpov15 (approved):

PR #12932, opened 2023-01-22
BREAKING CHANGE: remove castForQueryWrapper and _castForQuery on schematype, make castForQuery not rely on $conditional being optional

### LINKED ISSUE #12914: Remove `castForQueryWrapper`, make `castForQuery` take context as parameter
### Prerequisites

### Issue

When we first added support for executing setters with the Query as the context in https://github.com/Automattic/mongoose/commit/69e286a831d885533380e69e6fff3e93318c6cb9, we added a cumbersome `castForQueryWrapper()` method to avoid breaking `castForQuery()` 's API signature. In this next release, we should just remove `castForQueryWrapper()`, and instead make `castForQuery()` take context as a param to avoid the messy `$$context` hack.

vkarpov15:
Fix #12914

**Summary**

We created the `castForQueryWrapper` abstraction in https://github.com/Automattic/mongoose/commit/69e286a831d885533380e69e6fff3e93318c6cb9 to avoid potentially breaking custom schematype implementations by adding a 3rd param, because `castForQuery()` historically has relied on checking argument length to check if `$conditional` is set or not. But the castForQueryWrapper abstraction is a bit messy and makes it hard to write custom schematypes.

With this PR, `castForQuery()` now always takes 3 arguments, including the context. And, if you don't want to specify `$conditional`, you should pass null.

**Examples**

### DIFF
--- a/docs/migrating_to_7.md
+++ b/docs/migrating_to_7.md
@@ -12,6 +12,7 @@ you should be aware of when migrating from Mongoose 6.x to Mongoose 7.x.
 If you're still on Mongoose 5.x, please read the [Mongoose 5.x to 6.x migration guide](migrating_to_6.html) and upgrade to Mongoose 6.x first.
 
 * [`strictQuery`](#strictquery)
+* [Removed `castForQueryWrapper()`, updated `castForQuery()` signature](#removed-castforquerywrapper-updated-castforquery-signature)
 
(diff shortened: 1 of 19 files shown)


hasezoey (approved):
LGTM

this change should also be noted in the migrate_7

PR #12933, opened 2023-01-23
types(model): fix type of options of Model.aggregate

ghost91-:
**Summary**

This change adjusts the type of the `options` parameter of `Model.aggregate` to be the `AggregateOptions` type from mongoose, instead of the one from mongodb. The aggregate method just passes this parameter on to `Aggregate.option`, which is already typed to accept the `AggregateOptions` type from mongoose, so this seems like the correct choice.

The 2 types mostly seem to be compatible anyways, but there is a small inconsistency that makes them incompatible: `mongodb.AggregateOptions` specifies that the `session` property may be a `ClientSession`, or `undefined` (but not `null`), while the `AggregateOptions` type from mongoose also allows `null`. So maybe that also should simply be aligned.

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -11,7 +11,8 @@ import {
   HydratedDocument,
   HydratedDocumentFromSchema,
   Query,
-  UpdateWriteOpResult
+  UpdateWriteOpResult,
+  AggregateOptions
 } from 'mongoose';
 import { expectAssignable, expectError, expectType } from 'tsd';
 import { AutoTypedSchemaType, autoTypedSchema } from './schema.test';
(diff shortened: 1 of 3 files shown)


vkarpov15 (changes requested):
This should work fine, but to reduce the risk of any backwards breaking changes, can we change `AggregateOptions` to something like the following?

```ts
interface AggregateOptions extends Omit<mongodb.AggregateOptions, 'session'>, SessionOption {
    [key: string]: any;
  }
```

That seems to work fine, and makes it so that there's no chance we missed an option from `mongodb.AggregateOptions`.

ghost91-:
I have incorporated the suggestion. Before doing so, I have manually verified that there are no differences between the 2 `AggregateOptions` types, aside from allowing `null` as the `session` property.

vkarpov15 (approved):
Thanks :+1:

vkarpov15 (approved):
Thanks :+1:

PR #12935, opened 2023-01-23
BREAKING CHANGE: make exported model just an interface, rename exported Model const -> ModelClass

### LINKED ISSUE #12873: Naming conflict in const Model and interface Model
### Prerequisites

### Mongoose version

6.8.2

### Node.js version

16.8

### MongoDB server version

5.1

4.9.4

### Description

from models.d.ts we are exporting `const Model` and also `interface Model` and my IDE (WebStrom) is confused about what I want to import, even when I use the `import type` syntax:

### Steps to Reproduce

Screenshot 2023-01-04 at 21 07 52

as you can see MM refers to both (const and interface) and because of this, I have other issues with suggestions etc. in my IDE...

for example: 
Screenshot 2023-01-04 at 21 27 09
IDE is not sure about `MM`.

note: this 
```ts
export default (models.LicenseModel as MM<License>) ||
  model<License>('LicenseModel', Schema, 'licenses')
```
is there because of the hot module issue

PS: the `/** @type {import('mongoose').Model} */` don't help either.

### Expected Behavior

Can we rename `const Model` or `interface Model` to something else?

vkarpov15:
Fix #12873

**Summary**

#12873 points out that we have some odd legacy behavior with `Model`: `Model` is both an interface and a `const`. Besides being a bit confusing, this is a bit dangerous because it allows people to do `import { Model } from 'mongoose'; Model.findOne();`, which is ideally something that Mongoose should block. Seems like a "why would anyone do that?", but we have a couple of test cases (the ones this PR removes) that accidentally pass because they make use of that exact pattern :) 

Only downside is that classes that `extends Model` are no longer supported. Which didn't work well anyway, because `extends Model` implies `Model<any>`, and also creates a `Model` class that isn't connected to any connection, which would fail at runtime.

@hasezoey what do you think?

**Examples**


test/types/queries.test.ts
```diff
@@ -298,26 +298,6 @@ async function gh11306(): Promise<void> {
   expectType<string[]>(await MyModel.distinct<string>('name'));
 }
 
-async function gh11602(): Promise<void> {
```
hasezoey, line 301: maybe consider instead of removing just changing it to a actual model

types/models.d.ts
```diff
@@ -123,8 +123,8 @@ declare module 'mongoose' {
 
   interface RemoveOptions extends SessionOption, Omit<mongodb.DeleteOptions, 'session'> {}
 
-  const Model: Model<any>;
-  interface Model<T, TQueryHelpers = {}, TMethodsAndOverrides = {}, TVirtuals = {}, TSchema = any> extends
+  export const ModelClass: Model<any>;
+  export interface Model<T, TQueryHelpers = {}, TMethodsAndOverrides = {}, TVirtuals = {}, TSchema = any> extends
```
hasezoey, line 127: maybe consider adding tsdoc explaining what the difference between `ModelClass` and `Model` is

test/types/generics.test.ts
```diff
@@ -1,15 +0,0 @@
-import { Model, Document } from 'mongoose';
```
hasezoey, line 1: maybe consider instead of removing just changing it to a actual model

hasezoey (changes requested):
The change looks good, though with some questions to resolve

hasezoey:
Would be good if this change would also be documented in the migration guide

hasezoey:
just noticed that this change does not allow `model instanceof mongoose.Model` anymore (which works at runtime), because no type named `Model` exists (only a interface) and `ModelClass` does not exist at runtime from what i know

vkarpov15:
@hasezoey that's a good point. Do you think it is worthwhile to revert this change? I made this change assuming that there wouldn't be any significant impact on end users, but `model instanceof mongoose.Model` not working in TypeScript anymore sounds significant.

hasezoey:
> Do you think it is worthwhile to revert this change?

you can do 1 of 2 options:
- revert the change to how it was before, in just types
- add at runtime a alias for `Model` as `ModelClass`

PR #12938, opened 2023-01-24
Allow this typing in validator functions

Freezystem:
**Summary**

Improve ability to type validator functions by allowing to define the `this` keyword. This avoid type errors when using `this` in the function body.

**Examples**

```ts
export const refExists: ValidateOpts<Types.ObjectId> = {
	async validator(_id: Types.ObjectId, props: ValidatorProps & Record<string, unknown>): Promise<boolean> {
		const self = this.constructor as Model<unknown>;
		const {ref} = self.schema.paths[props.path].options;
		const model = models[ref as string];

		return await model.exists({_id}).exec() !== null;
	},
	message: 'REF_ID_DOES_NOT_EXISTS',
	propsParameter: true,
};
```
return the following error on this.constructor;

> any
> Unsafe member access .constructor on an `any` value. `this` is typed as `any`.
> You can try to fix this by turning on the `noImplicitThis` compiler option, or adding a `this` parameter to the function.eslint[@typescript-eslint/no-unsafe-member-access](https://typescript-eslint.io/rules/no-unsafe-member-access)

### DIFF
--- a/types/validation.d.ts
+++ b/types/validation.d.ts
@@ -2,32 +2,32 @@ declare module 'mongoose' {
 
   type SchemaValidator<T> = RegExp | [RegExp, string] | Function | [Function, string] | ValidateOpts<T> | ValidateOpts<T>[];
 
-  interface ValidatorProps {
+  interface ValidatorProps<T> {
     path: string;
-    value: any;
+    value: T;
   }
 
-  interface ValidatorMessageFn {
-    (props: ValidatorProps): string;
+  interface ValidatorMessageFn<T> {
+    (props: ValidatorProps<T>): string;
   }
 
-  interface ValidateFn<T> {
-    (value: T, props?: ValidatorProps & Record<string, any>): boolean;
+  interface ValidateFn<T, M = Model<unknown>> {
+    (this: M, value: T, props?: ValidatorProps<T> & Record<string, any>): boolean;
   }
 
-  interface LegacyAsyncValidateFn<T> {
-    (value: T, done: (result: boolean) => void): void;
+  interface LegacyAsyncValidateFn<T, M = Model<unknown>> {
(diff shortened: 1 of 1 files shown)


types/validation.d.ts
```diff
@@ -2,32 +2,32 @@ declare module 'mongoose' {
...
-  interface ValidatorMessageFn {
-    (props: ValidatorProps): string;
+  interface ValidatorMessageFn<T> {
+    (props: ValidatorProps<T>): string;
   }
 
-  interface ValidateFn<T> {
-    (value: T, props?: ValidatorProps & Record<string, any>): boolean;
+  interface ValidateFn<T, M = Model<unknown>> {
+    (this: M, value: T, props?: ValidatorProps<T> & Record<string, any>): boolean;
```
vkarpov15, line 15: `this` should be a HydratedDocument, not `Model`. Remember `Model` is a type, so `this` would be a model class, not an instance of a model class. I'd change the generic def from `M = Model<unknown>` to `THydratedDocumentType = HydratedDocument<unknown>`

types/validation.d.ts
```diff
@@ -2,32 +2,32 @@ declare module 'mongoose' {
 
   type SchemaValidator<T> = RegExp | [RegExp, string] | Function | [Function, string] | ValidateOpts<T> | ValidateOpts<T>[];
```
vkarpov15, line 3: Without changing `SchemaValidator`, and `SchemaTypeOptions`, there's no way for user to set `M`?

vkarpov15 (changes requested):

vkarpov15:
@Freezystem do you want to continue work on this PR?

Freezystem:
Hey, sorry I'm actually very busy and I will not have the time to work on it for the moment.  
We can close this one until someone else fixes it or I get time to do it myself.

Issue #12944, opened 2023-01-25
Chaining `.find` with `$and` condition overwrites prior conditions


Rouby:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.7.0

### Node.js version

18.x

### MongoDB server version

4.x

### Description

Chaining multiple `.find` (or other methods that manipulate the conditions) with `$and` (or other array-based conditions) results in the latest method silently overwriting any previous conditions.

This issue is relevant specifically for the use-case with the package @casl/mongoose that has utility methods to add constraints to queries based on the ability of an user.

E.g.

```
    MyModel.accessibleBy(userAbility)
      .countDocuments(additionalConditions)
      .exec(),
```

The query-conditions generated by `accessibleBy` might be silently overwritten if they happen to utilize `$and` and the `additionalConditions` do so too.

### Steps to Reproduce

```
MyModel.find({ $and: [{ type: { $in: ['X'] } }] })
        .find({ $and: [{ type: { $in: ['Y'] } }] })
        .countDocuments({ $and: [{ type: { $in: ['Z'] } }] })
        .getQuery(),

//? {"$and":[{"type":{"$in":["Z"]}}]}
```

### Expected Behavior

Since other conditions are merged, I would expect array-conditions to also be merged.

```
MyModel.find({ $and: [{ type: { $in: ['X'] } }] })
        .find({ $and: [{ type: { $in: ['Y'] } }] })
        .countDocuments({ $and: [{ type: { $in: ['Z'] } }] })
        .getQuery(),

//? {"$and":[{"type":{"$in":["X"]}},{"type":{"$in":["Y"]}},{"type":{"$in":["Z"]}}]}
```

PR #12945, opened 2023-01-25
fix(merge): allow array fields to be merged

Rouby:
**Summary**

This PR addresses the issue mentioned in #12944


vkarpov15 (commented):
I don't think we can merge this particular change, changing `utils.merge()` for this one case is a bit too risky. For example, we also use `merge()` to merge query options and query updates, and we don't necessarily want to merge array options or arrays in updates.

I'm not saying that #12944 isn't something we can't fix, but this particular implementation is unfortunately not something we can merge. Sorry.

PR #12947, opened 2023-01-25
Fix discriminator type field update for custom discriminator model name

wassil:
**Summary**

Mongoose requires unique model names, so to avoid conflicts when using discriminators, you use 3 parameter syntax: `model.discriminator(modelName, schema, discriminatorKeyValue)`.
When updating a document using e.g. `updateOne`, you're allowed to change discriminator key value when you provide `overwriteDiscriminatorKey`.
But the code responsible for updating the discriminator type was relying on `modelName` being equal to `discriminatorKeyValue`, which may not be the case if you use custom model name for your discriminator.

`Schema.discriminators` are indexed by model name and the code was using update value for `discriminatorKey`.

**Examples**

I modified an existing test to make it fail, then fixed it.

I'm not sure if I should keep the original test as is and add a new one, but the difference is just 2 lines - the model names for discriminators.

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -5100,7 +5100,9 @@ Query.prototype._castUpdate = function _castUpdate(obj, overwrite) {
   if (this._mongooseOptions.overwriteDiscriminatorKey &&
       obj[discriminatorKey] != null &&
       baseSchema.discriminators) {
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Great, thanks :+1:

PR #12948, opened 2023-01-25
fix: export mongodb types more robustly

simon-abbott:
**Summary**

Mongoose exposes its underlying `mongodb` import via the `mongo` export. However, if you use `import * as mongoose from 'mongoose'` syntax then `mongoose.mongo.X` becomes untyped. By using an alias export instead, the types are preserved, even when importing like that.

**Examples**

See the test I added.

### DIFF
--- a/test/types/mongo.test.ts
+++ b/test/types/mongo.test.ts
@@ -1,0 +1,4 @@
+import * as mongoose from 'mongoose';
+import { expectType } from 'tsd';
+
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
This looks like a good alternative to #11954 

AbdelrahmanHafez (approved):

hasezoey (approved):
LGTM

vkarpov15 (approved):
Nice, thanks :+1:

PR #12951, opened 2023-01-25
Revert "typescript: Allow null for optional document fields"

vkarpov15:
**Summary**

Probably best to defer the change from #12781 until our 7.0 release. The change seemed easy enough, but then we got bit by test failures in master that we didn't notice until after we merged our PR for 6.9. This change may require users to update method return signatures and do other additional work, so it is likely a bit too big of a change for a minor release.

**Examples**


IslandRhythms (approved):
LGTM

PR #12954, opened 2023-01-26
Fix "near" aggregation operator input type

Jokero:
**Summary**

The current input parameter type for `near` aggregation operator does not reflect what MongoDB actually supports since version `4.0` (https://www.mongodb.com/docs/v4.0/reference/operator/aggregation/geoNear/). `4.0` is minimum supported version by Mongoose `^6.0.0` (https://mongoosejs.com/docs/compatibility.html).

I've switched to the correct (already existing) interface, removed `uniqueDocs` property (which has no effect since MongoDB `2.6`) and added `num` property supported in MongoDB `4.0.x` but removed in `4.2`.

**Examples**

Without these changes, I cannot, for example, use GeoJSON point:
image

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -360,13 +360,12 @@ Aggregate.prototype.project = function(arg) {
  * #### Example:
  *
  *     aggregate.near({
- *       near: [40.724, -73.997],
+ *       near: { type: 'Point', coordinates: [40.724, -73.997] },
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks :+1:

Peace-Song:
@vkarpov15 
Hi. Sorry if this request is abrupt, but could you please apply this commit to 5.x.x as well?

PR #12955, opened 2023-01-26
BREAKING CHANGE: use mongodb@5, bson@5

vkarpov15:
**Summary**

Couple of notes: `mongoose.Types.ObjectId()` no longer works, need to do `new mongoose.Types.ObjectId()`. ObjectId class's `_bsontype` is now 'ObjectId', not 'ObjectID'.

`MongoClient.connect()` no longer supports callbacks. We will likely need to drop callback support as well in this version.

**Examples**


hasezoey (commented):
from what i can tell, this PR is not final yet, is there a ETA for mongodb's 5.0 release?

things to note:
- debug console.log's present
- documentation not yet noting that `new Types.ObjectId()` is required now
- something seems like it does not work (see tests)?
- usage of alpha dependencies (mongodb, bson)

vkarpov15:
@hasezoey you can ignore this PR for now, it is a very rough draft and not close to ready for merge. We're not shipping this with alpha dependencies.

No formal ETA for mongodb@5 release, but I would expect to see that released within the next few weeks.

lib/connection.js
```diff
@@ -530,9 +532,20 @@ Connection.prototype.transaction = function transaction(fn, options) {
  * @api public
  */
 
-Connection.prototype.dropCollection = _wrapConnHelper(function dropCollection(collection, cb) {
-  this.db.dropCollection(collection, cb);
-});
+Connection.prototype.dropCollection = function dropCollection(collection, cb) {
+  const args = Array.prototype.slice(arguments);
```
baileympearson, line 536:
```suggestion
  const args = Array.prototype.slice.call(arguments);
```
The failing async iterator tests are a result of `this` being improperly bound when `slice` is called.

This results in args being `[]`, line 542 producing `args: [cb]` instead of `args: ['Movie', cb]`, and then the driver constructs the following command:

```
{ drop: <callback function>, lsid: < binary session id > }
```

`serializeFunctions` (bson option) defaults to false, so the `drop` command is filtered out by bson and the server throws an "Unknown command "lsid"" error.

sapter:
mongodb@5 is out!

package.json
```diff
@@ -19,9 +19,9 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "bson": "^4.7.0",
+    "bson": "^5.0.0",
     "kareem": "2.5.1",
-    "mongodb": "4.13.0",
+    "mongodb": "5.0.0",
```
aksdevac, line 24: Here fixed version is used but 5.0 has some critical issues like this:
https://jira.mongodb.org/projects/NODE/issues/NODE-5052?filter=allopenissues

So we should wait for next driver release.

vkarpov15:
@hasezoey @AbdelrahmanHafez can you please take a look at this?

hasezoey (changes requested):
What a bunch of changes, the tests look way better now than the old nesting.

- it seems like the benchmark for `update` was deleted instead of updated, should there be a new one?
- typescript types seem to not have been updated yet
- there are conflicts with the base branch

i have mainly some questions & style changes.

lib/cast/decimal128.js
```diff
@@ -20,7 +20,10 @@ module.exports = function castDecimal128(value) {
     return Decimal128Type.fromString(value);
   }
 
-  if (Buffer.isBuffer(value)) {
+  if (typeof Buffer === 'function' && Buffer.isBuffer(value)) {
```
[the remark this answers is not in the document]
vkarpov15: Deno and browser. Buffer is a purely Node construct. `Uint8Array` is how the bson lib represents Decimal128's internally in environments where `Buffer` doesn't exist.

lib/helpers/discriminator/areDiscriminatorValuesEqual.js
```diff
@@ -9,7 +9,7 @@ module.exports = function areDiscriminatorValuesEqual(a, b) {
   if (typeof a === 'number' && typeof b === 'number') {
     return a === b;
   }
-  if (isBsonType(a, 'ObjectID') && isBsonType(b, 'ObjectID')) {
+  if (isBsonType(a, 'ObjectId') && isBsonType(b, 'ObjectId')) {
```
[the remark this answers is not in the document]
vkarpov15: Nope, no longer valid. That's a change in the bson lib. For end users, it shouldn't matter too much.

lib/drivers/node-mongodb-native/connection.js
```diff
@@ -144,17 +141,16 @@ NativeConnection.prototype.doClose = function(force, fn) {
...
-    // If there's queued operations, you may still get some background work
-    // after the callback is called.
-    setTimeout(() => fn(err, res), 1);
-  });
+  await this.client.close(force);
+  // Defer because the driver will wait at least 1ms before finishing closing
+  // the pool, see https://github.com/mongodb-js/mongodb-core/blob/a8f8e4ce41936babc3b9112bf42d609779f03b39/lib/connection/pool.js#L1026-L1030.
+  // If there's queued operations, you may still get some background work
+  // after the callback is called.
+  await new Promise(resolve => setTimeout(resolve, 1));
```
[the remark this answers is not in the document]
vkarpov15: I'm not certain, but I figure there isn't much harm to leaving this here. This code doesn't add to maintenance burden, and waiting an extra millisecond on `close()` isn't a huge impact on perf relative to the cost of opening/closing a connection.

test/model.populate.test.js
```diff
@@ -1946,165 +1578,64 @@ describe('model: populate:', function() {
...
-          assert.equal(note.body, 'Buy Milk');
-          assert.ok(note.author);
-          assert.equal(note.author.name, 'Alice');
-          done();
-        });
-      });
-    });
-  });
-
-  it('required works on ref fields (gh-577)', function(done) {
```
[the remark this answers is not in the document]
vkarpov15: You're right, not sure why this one was removed. Likely because we assumed it was redundant and hard to refactor to async/await. But turns out it isn't redundant.

test/model.test.js
```diff
@@ -1075,105 +1031,11 @@ describe('Model', function() {
         const doc = await post.save();
         assert.ok(doc);
       });
-
-      it('async', async function() {
```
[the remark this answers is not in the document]
vkarpov15: No, it was removed because we don't support callback-based async validators anymore as of Mongoose 6. It looks like this particular syntax still works, but we don't need tests for it anymore.

IslandRhythms (approved):
LGTM

docs/migrating_to_7.md
```diff
@@ -29,6 +33,140 @@ const docs = await MyModel.find({ notInSchema: 1 });
...
+- `Model.syncIndexes`
+- `Model.updateMany`
+- `Model.updateOne`
+- `Query.prototype.find`
+- `Query.prototype.findOne`
+- `Query.prototype.findOneAndDelete`
+- `Query.prototype.findOneAndUpdate`
+- `Query.prototype.findOneAndRemove`
+- `Query.prototype.findOneAndReplace`
+- `Query.prototype.validate`
```
hasezoey, line 127:
```suggestion
- `Query.prototype.validate`
- `Query.prototype.deleteOne`
- `Query.prototype.deleteMany`
- `Query.prototype.exec`
```

hasezoey (approved):
LGTM

though noticed some missing `Query.prototype` functions in the `callbacks` removed section

lib/model.js
```diff
@@ -482,18 +479,21 @@ function generateVersionError(doc, modifiedPaths) {
  * @param {Boolean} [options.validateBeforeSave] set to false to save without validating.
  * @param {Boolean} [options.validateModifiedOnly=false] if `true`, Mongoose will only validate modified paths, as opposed to modified paths and `required` paths.
  * @param {Number|String} [options.w] set the [write concern](https://www.mongodb.com/docs/manual/reference/write-concern/#w-option). Overrides the [schema-level `writeConcern` option](/docs/guide.html#writeConcern)
- * @param {Boolean} [options.j] set to true for MongoDB to wait until this `save()` has been [journaled before resolving the returned promise](https://www.mongodb.com/docs/manual/reference/write-concern/#j-option). Overrides the [schema-level `writeConcern` option](/docs/guide.html#writeConcern)
+ * @param {Boolean} [options.j] set to true for MongoDB to wait until this `save()` has been [journaled before resolving the returned promise](https://docs.mongodb.com/manual/reference/write-concern/#j-option). Overrides the [schema-level `writeConcern` option](/docs/guide.html#writeConcern)
```
hasezoey, line 482: why was the link changed back to its old version?

test/query.test.js
```diff
@@ -1341,53 +1253,55 @@ describe('Query', function() {
...
+    });
 
-        done();
+    describe('snapshot', function() {
+      it('works', function() {
+        const query = new Query({});
+        query.snapshot(true);
+        assert.equal(query.options.snapshot, true);
       });
     });
```
hasezoey, line 1292: i think this was accidentally added back in the merge of 7.0 into this PR

weirdly the tests seem to pass though

lib/model.js
vkarpov15: Mistake in resolving merge conflict. Fixed.

test/query.test.js
vkarpov15: mquery still has a snapshot method, that's why this test still succeeds. I guess we should also remove these from mquery.

hasezoey:
should this PR also update mquery, or should this be done after this PR?

see https://github.com/mongoosejs/mquery/pull/137

vkarpov15:
@hasezoey we'll merge this PR into 7.0. Let's open a separate PR to bump mquery.

PR #12994, opened 2023-02-03
fix(document): isModified should not be triggered when setting a nested boolean to the same value as previosly

### LINKED ISSUE #12992: IsModified is triggered when setting a nested boolean to the same value as previously
### Prerequisites

### Mongoose version

6.9.0

### Node.js version

18.x

### MongoDB server version

6.x

4.6

### Description

When setting a boolean in a nested path to the same value as previously, the document gets marked as modified when it shouldn't.

### Steps to Reproduce

```
//define schema
const assessmentSchema = new Schema({
    _id: { type: String, default: uuidv4 },
    result: new Schema({
        score: Number,
        passed: Boolean
    }, { _id: false })
})
export const Assessment = model('assessment', assessmentSchema)

//connect to db here

//first save an assessment with result.passed set to false
let assessment = new Assessment({
        result: {
            score: 40,
            passed: false
        }
    })
    await assessment.save();

    //retrieve assessment and set result to the same value as previously
    assessment = await Assessment.findById(assessment.id).orFail();
    assessment.result = {
        score: 40,
        passed: false
    }

    console.log("assessment modified", assessment.isModified(), assessment.modifiedPaths())
   // prints :  assessment true (2) ['result', 'result.passed']

```

### Expected Behavior

I would expect mongoose to return isModified() false as nothing has changed on the result object.

I have tracked down the offending code in mongoose in the Document.js file : 

```
Document.prototype.$__shouldModify = function(pathToMark, path, options, constructing, parts, schema, val, priorVal) {
   
   [...]

  //here seems to be the issue
  //Given that priorVal is false, utils.getValue get invoked and returns undefined and so the values are not considered equal
  //If we used ?? (Nullish coalescing operator ) instead of || it would work as expected

  if (!deepEqual(val, priorVal || utils.getValue(path, this))) {
    return true;
  }


}
```

lpizzinidev:
fix #12992

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1621,7 +1621,7 @@ Document.prototype.$__shouldModify = function(pathToMark, path, options, constru
     return false;
   }
 
-  if (!deepEqual(val, priorVal != null ? priorVal : utils.getValue(path, this))) {
+  if (!deepEqual(val, priorVal !== undefined ? priorVal : utils.getValue(path, this))) {
     return true;
   }
 
--- a/lib/document.js
+++ b/lib/document.js
@@ -1621,7 +1621,7 @@ Document.prototype.$__shouldModify = function(pathToMark, path, options, constru
     return false;
   }
 
-  if (!deepEqual(val, priorVal || utils.getValue(path, this))) {
+  if (!deepEqual(val, priorVal != null ? priorVal : utils.getValue(path, this))) {
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
I changed `null` to `undefined` to avoid having the same bug with null values. But otherwise LGTM :+1: Thanks for finding this issue.

PR #13004, opened 2023-02-06
fix(document): make array `$shift()` use `$pop` instead of overwriting array

vkarpov15:
**Summary**

Looks like in Mongoose 6.0.0, our switch to using proxies for arrays broke some internal behavior for `$shift()`. The TLDR; is that `$shift()` mistakenly always becomes a `$set`, as opposed to a `$pop`, since 6.0.0. Which may have performance implications for large arrays.

This issue was happening because 1) setting `_shifted` on the array proxy triggers a `$set`, and 2) calling `shift()` on an array proxy also triggers a `$set`.

**Examples**

### DIFF
--- a/lib/types/array/methods/index.js
+++ b/lib/types/array/methods/index.js
@@ -137,12 +137,13 @@ const methods = {
     this._markModified();
 
     // only allow shifting once
-    if (this._shifted) {
+    const __array = this.__array;
+    if (__array._shifted) {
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

hasezoey (approved):
LGTM

PR #13007, opened 2023-02-08
fix(connection): disable createConnection buffering if autoCreate option is changed

### LINKED ISSUE #12940: Connection autoCreate option behaves differently migrating 5 to 6
### Prerequisites

### Mongoose version

6.8.4

### Node.js version

18.13.0

### MongoDB server version

4.12.1 (driver)

_No response_

### Description

When upgrading from Mongoose 5.13.15 to 6.8.4, we switched to explicitly setting the `autoCreate` option to `false` on the connection. We previously never set the autoCreate option on our Schema objects, so they should default to null (which means use the connection's autoCreate option). Our Schemas instantiate before the connection is created.

We found that in Mongoose 6, the autoCreate commands buffered prior to the connection being created were executed, issuing createCollection commands to the database. This didn't happen previously in Mongoose 5.

Workaround was to explicitly set autoCreate to false on the Schema objects.

### Steps to Reproduce

Added to `test/connection.test.js`

```
it('with autoCreate = false after schema create', async function() {
    const schema = new Schema({ name: String }, {
      collation: { locale: 'en_US', strength: 1 },
      collection: 'gh8814_Conn'
    });

    const Model = mongoose.model('gh8814_Conn', schema);

    const conn = mongoose.connect(start.uri, {
      autoCreate: false
    });

    await conn;
    await Model.init();

    const res = await mongoose.connection.db.listCollections().toArray();
    assert.ok(!res.map(c => c.name).includes('gh8814_Conn'));
  });
```

### Expected Behavior

Schema with `autoCreate: null` option set follows the connection's `autoCreate` option, even if schema is initialized first and the createCollection command is buffered.

lpizzinidev:
fix #12940

**Summary**
Avoid usage of cached `createCollection` callback if `autoCreate` option value has changed.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1341,7 +1341,10 @@ Model.init = function init(callback) {
 
   this.schema.emit('init', this);
 
-  if (this.$init != null) {
+  const autoCreate = utils.getOption('autoCreate',
+    this.schema.options, this.db.config, this.db.base.options);
+
+  if (this.$init != null && this.$autoCreate === autoCreate) {
     if (callback) {
       this.$init.then(() => callback(), err => callback(err));
       return null;
@@ -1349,10 +1352,10 @@ Model.init = function init(callback) {
     return this.$init;
   }
 
+  this.$autoCreate = autoCreate;
+
   const Promise = PromiseProvider.get();
   const autoIndex = utils.getOption('autoIndex',
-    this.schema.options, this.db.config, this.db.base.options);
-  const autoCreate = utils.getOption('autoCreate',
     this.schema.options, this.db.config, this.db.base.options);
 
   const _ensureIndexes = autoIndex ?
--- a/lib/model.js
+++ b/lib/model.js
@@ -60,9 +60,11 @@ const modifiedPaths = require('./helpers/update/modifiedPaths');
 const parallelLimit = require('./helpers/parallelLimit');
 const parentPaths = require('./helpers/path/parentPaths');
 const prepareDiscriminatorPipeline = require('./helpers/aggregate/prepareDiscriminatorPipeline');
+const promiseOrCallback = require('./helpers/promiseOrCallback');
 const pushNestedArrayPaths = require('./helpers/model/pushNestedArrayPaths');
 const removeDeselectedForeignField = require('./helpers/populate/removeDeselectedForeignField');
 const setDottedPath = require('./helpers/path/setDottedPath');
+const STATES = require('./connectionstate');
 const util = require('util');
 const utils = require('./utils');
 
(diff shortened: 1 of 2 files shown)


lpizzinidev:
@vkarpov15 
Do you think that this is the correct approach?
If so I'll setup a MongoDB 4.0.x env to check why tests are failing on that specific version only.
Otherwise I'll search for another way to solve the issue.

lib/model.js
```diff
@@ -1341,19 +1341,22 @@ Model.init = function init(callback) {
 
   this.schema.emit('init', this);
 
-  if (this.$init != null) {
+  const autoCreate = utils.getOption('autoCreate',
+    this.schema.options, this.db.config, this.db.base.options);
+
+  if (this.$init != null && this.$autoCreate === autoCreate) {
```
vkarpov15, line 1347: The problem with this approach is that you may have multiple `$init`'s going at the same time. `Model.init(); Model.schema.options.autoCreate = false; Model.init();`. The purpose of `$init` is to avoid that.

I think a better approach would be to move the `utils.getOption()` call for `autoCreate` into the `_createCollection()` function itself. And put `_createCollection()` behind buffering like how `_wrapConnHelper()` works.

vkarpov15 (changes requested):

lib/model.js
lpizzinidev: Thanks for the feedback 💪 I'll work on this

lib/model.js
```diff
@@ -1354,15 +1356,37 @@ Model.init = function init(callback) {
...
+              fn.apply(conn, [cb]);
+            } catch (err) {
+              return cb(err);
+            }
+          }
+        });
+      });
+    };
+  };
+  const _createCollection = _wrapCollHelper(function _createCollection(cb) {
```
vkarpov15, line 1384: For consistency, let's also delay checking `autoIndex` until after buffering is done. Apply similar changes to `_ensureIndexes()`. No need to do `_wrapCollHelper` on `_ensureIndexes()` though, since `_createCollection()` will already wait for buffering.

lib/model.js
```diff
@@ -1354,15 +1356,37 @@ Model.init = function init(callback) {
...
+    const autoCreate = utils.getOption('autoCreate',
```
vkarpov15, line 1385: Also, let's not add a separate `_wrapCollHelper()` function, since that function won't be reused anywhere. Just add the `if ((conn.readyState === STATES.connecting || conn.readyState === STATES.disconnected) && conn._shouldBufferCommands())` check directly into `_createCollection()`.

vkarpov15 (changes requested):

vkarpov15 (approved):
Thanks :+1:

PR #13013, opened 2023-02-09
chore: add strict populate as exported error

emilsivervik:
**Summary**

Add and export `StrictPopulateError` class so it can be used by ts/esm modules

**Examples**
```javascript
import { MongooseError } from 'mongoose';

const err = new MongooseError.StrictPopulateError('test')

if(err instanceof MongooseError.StrictPopulateError){
  // do something
}
```


vkarpov15 (approved):
Thanks :+1:

PR #13001, opened 2023-02-05
Update broken & outdated links

hasezoey:
**Summary**

This PR does:
- update broken & outdated (308) links
- add `broken-link-checker` as a script command to easily run it
- update `api.js` (and `splitApiDocs.js`) to generate all names in one place
- update `api.js` to better handle `@inherits` links for classes that are not external

Notes:
- i wanted to integrate BLC into a custom script that is tailored to this project, but couldnt because of https://github.com/stevenvachon/broken-link-checker/issues/253 (which could be fixed if https://github.com/stevenvachon/broken-link-checker/issues/244 gets resolved)
- there were some links that i did not know how to update

Links Not updated and are broken:
- any links using `"query-js" & "model-js" & "schema-string-js" & "schema-date-js"`
- any links using `mongoosearray` (not documentarray)
- any links using `utils`

there were also broken links because of removed (previously deprecated) options:
- query option `snapshot` (in `query.js`)
- query option `maxScan` (in `query.js`)
- `autoIndexId` (in `guide.md`)
- `connection-limits-and-cluster-tier` (in `connections.md`)

Filters i used and ignored from the output (regex / file path)
- `^.*OK─[^\n]+\n`
- `^.*/docs/\d\.x[^\n]+\n`
- `^Getting links from: [^\n]+\nFinished[^\n]+\n\n`
- `/docs/api.html` (because of #12223)

re #12691

<details>
<summary>Full BLC log of still broken URLs</summary>

```txt
Getting links from: http://127.0.0.1:8089/docs/guide.html
├─BROKEN─ http://127.0.0.1:8089/docs/api/schema-string-js.html#schema-string-js (HTTP_404)
├─BROKEN─ http://www.mongodb.org/display/DOCS/Capped+Collections#CappedCollections-autoIndexId (HTTP_308)
├─BROKEN─ http://127.0.0.1:8089/docs/api/utils.html#utils_exports-toCollectionName (HTTP_404)
├─BROKEN─ http://127.0.0.1:8089/docs/api/schema-date-js.html#schema-date-js (HTTP_404)
Finished! 274 links found. 198 excluded. 6 broken.

Getting links from: http://127.0.0.1:8089/docs/schematypes.html
├─BROKEN─ http://127.0.0.1:8089/docs/api/query-js.html#query-js (HTTP_404)
Finished! 171 links found. 144 excluded. 3 broken.

Getting links from: http://127.0.0.1:8089/docs/connections.html
[...]

PR #13019, opened 2023-02-11
Update & remove old links

hasezoey:
**Summary**

This PR is a follow-up to #13001, which changes almost all the remaining things
(in a way as suggested at https://github.com/Automattic/mongoose/pull/13001#issuecomment-1426163524)

Using same ignore filters as in the mentioned PR


hasezoey:
as a note, please verify if the following example is still correct to include the option `autoIndexId`:
https://github.com/Automattic/mongoose/blob/542a20ebafd59e5255c9206cdc0a1e2cb8f902ac/docs/guide.md#capped

vkarpov15 (approved):
Thanks :+1:

PR #13021, opened 2023-02-11
feat(index): added createInitialConnection option to Mongoose constructor

### LINKED ISSUE #12965: Feature: Option to disable creation of initial connection
### Prerequisites

### 🚀 Feature Proposal

If you manage multiple custom connections, a typical pattern is to simply create new connections with `createConnection(uri, options)`.  Doing so, it adds by default to the mongoose singleton of `mongoose.connections`, so you can iterate over `mongoose.connections` to see properties such as `readyState`.

The default connection however is always present.  The only way to delete it (regardless of custom Mongoose instances) is to manually call `mongoose.connection.destroy();`.  This isn't such a good pattern, as a better option might be `new Mongoose({ createInitialConnection: false });` or something.  Adding this feature would require some refactoring though.

lpizzinidev:
closes #12965

**Summary**
If explicitly set to `false` the `createInitialConnection` option will disable the creation of the default initial connection.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -71,8 +71,11 @@ function Mongoose(options) {
     autoIndex: true,
     autoCreate: true
   }, options);
-  const conn = this.createConnection(); // default connection
-  conn.models = this.models;
+  const createInitialConnection = utils.getOption('createInitialConnection', this.options) ?? true;
+  if (createInitialConnection) {
+    const conn = this.createConnection(); // default connection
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
LGTM. Going to merge this into 7.0 branch since we're not doing another minor release before 7.0

vkarpov15:
I cherry-picked this into 6.10.

PR #13022, opened 2023-02-11
feat(query): remove all references to option "maxScan"

hasezoey:
**Summary**

This PR removes all references to query option `maxScan`, which [has been removed from the driver in 4.0](https://github.com/mongodb/node-mongodb-native/commit/0e6375a22fc3d1f4ce10846b98dafaa65044ef6d) and has been removed from [mongodb's (db) side since 4.2](https://www.mongodb.com/docs/v4.0/reference/method/cursor.maxScan/#cursor.maxScan) (and deprecated since 4.0)

PS: i am not quite sure if this *should* be removed already, because version 4.0 and 4.2 are still supported by the driver (but the options have been removed from the driver) and from what i can tell mongoose matches the drivers support of mongodb (db) versions

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -52,9 +52,7 @@ const queryOptionMethods = new Set([
   'j',
   'lean',
   'limit',
-  'maxScan',
   'maxTimeMS',
-  'maxscan',
   'populate',
   'projection',
   'read',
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #13023, opened 2023-02-11
feat(query): remove all references to option "snapshot"

hasezoey:
**Summary**

This PR removes the query option `snapshot`, which i could not find any documentation about it anymore (even for mongodb (db) 3.6), i could only find things for `readconcern: "snapshot"`

Note: this PR does not touch the read-concern option


vkarpov15 (approved):
Thanks :+1:

Issue #12997, opened 2023-02-03
Error: remove() on sub document array with map


manoj19058:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.9.0

### Node.js version

18.13.0

### MongoDB server version

6.0.2

### Typescript version (if applicable)

4.9.5

### Description

Here created a **User Schema** with two properties: 
1. _name_ : _String_
2. _subscriptions_: _Map<String, Subscription []>_

And **Subscription Schema** also contains two properties: 
1. _name_ : _String_
2. _duration_ : _Number_

**Subscription []** in _User Schema_ is an **array of sub documents**. Removing elements from this array using [remove()](https://mongoosejs.com/docs/subdocs.html#removing-subdocs) for  array of sub documents yields the error.

### Steps to Reproduce

Standalone code to reproduce
```Typescript
import mongoose from "mongoose";
const subscriptionSchema = new mongoose.Schema({
    name: String,
    duration: Number,
});

const userSchema = new mongoose.Schema({
    name: String,
    subscriptions: {
        type: Map,
        of: [subscriptionSchema],
    },
});

const User = mongoose.model("Users", userSchema);

(async function () {
    mongoose.set("strictQuery", true);
    mongoose.set("debug", true);
    await mongoose.connect("mongodb://127.0.0.1:27017/mydb");

    const user = new User({
        name: "Joe Doe",
        subscriptions: {
            api: [
                { name: "basic", duration: 3 },
                { name: "pro", duration: 1 },
            ],
        },
    });
    const doc = await user.save();
    try {
        const subscription = doc.subscriptions?.get("api");
        if (subscription == null) return;
        const id = subscription[0].id;
        console.log(`\nID of Object to be removed: ${id}\n`);
        await user.subscriptions?.get("api")?.id(id)?.remove();
    } catch (error) {
        console.log(error);
    }
})();
```

Output : 
```sh
Mongoose: users.insertOne({ name: 'Joe Doe', subscriptions: Map(1) { 'api' => [ { name: 'basic', duration: 3, _id: new ObjectId("63dd577ef901021530e0e0e6") }, { name: 'pro', duration: 1, _id: new ObjectId("63dd577ef901021530e0e0e7") } ] }, _id: new ObjectId("63dd577ef901021530e0e0e5"), __v: 0}, { session: null })

ID of Object to be removed: 63dd577ef901021530e0e0e6

[...]

PR #13027, opened 2023-02-12
Error: remove() on sub document array with map #12997

Negarnbkhti42:
Finding the small this ommit in array methods

**Summary**

**Examples**


vkarpov15 (changes requested):
Can you please undo all the code formatting changes? It is very difficult to review this PR, and we'll have to undo most of your changes anyway because they won't pass our linter.

PR #13029, opened 2023-02-13
Prevent casting null or undefined UUID field value

Freezystem:
**Summary**

In `binaryToString` function (line 67), return the _falsy_ (e.g: `null` or `undefined`) value instead of trying to cast it to string.

Optional modifications:
- in `hex2buffer` function (line 24), check for _falsy_ `hex` before trying to cast it to a `Buffer`.
- in `binary2hex` function (line 37), check for _falsy_ `buf` before trying to generate hexadecimal string from it.

**Examples**

related to https://github.com/Automattic/mongoose/discussions/13025


Freezystem:
Sorry I don't know if written tests are sufficient enough as I can't run them locally because of a Mongo Memory error.

vkarpov15 (approved):

Issue #13032, opened 2023-02-14
What is the proper way to replace `_id` type from `ObjectId` to `UUID` in mongoose 6


vkarpov15:
### Discussed in https://github.com/Automattic/mongoose/discussions/13025

<div type='discussions-op-text'>

<sup>Originally posted by **Freezystem** February 11, 2023</sup>
I've read a lot of post in SO or GH where people are trying to implement UUIDs properly instead of ObjectIds. But these posts are a bit outdated for most of them or incomplete. I also wanted to avoid third party librairies like _mongoose-uuid2_.
The official documentation is not even mentioning the official `UUID` schema type.
I also have read that storing UUIDs as `Buffer` instead of `String` improve overall storage and index performance but it could be false.
Also it is not clear if I should use `uuidv1` or `uuidv4` generator as mongoose `UUID` type mention both in its source.

So for the moment I'm doing something like this:
```ts
import type { Schema } from "mongoose";
import { randomUUID } from "crypto";  // NodeJS implementation of uuid v4.

interface IUser {
    _id: Schema.Types.UUID;
    name: string;
    organization?: Schema.Types.UUID;
}

const userSchema = new Schema<IUser>({
    _id: {
        type: Schema.Types.UUID,
        default: randomUUID,
        immutable: true,
    },
    name: {
        type: String,
        required: true,
    },
    organization: {
        type: Schema.Types.UUID,
        ref: "Organization",
        index: true,
    }
},
{ _id: false });
```

But I get some errors, especially with the organization reference when I try to destructure a newly created user and organization isn't provided, example:

```js
const {name, organization} = await User.create({name: "user"});
```

> TypeError: Cannot read properties of undefined (reading 'toString')
```
at binary2hex (node_modules/mongoose/lib/schema/uuid.js:39:19)
at model.binaryToString (node_modules/mongoose/lib/schema/uuid.js:71:11)
at SchemaUUID.Object.<anonymous>.SchemaType.applyGetters (node_modules/mongoose/lib/schematype.js:1225:20)
at model.Object.<anonymous>.Document.get (node_modules/mongoose/lib/document.js:1886:18)
at model.get [as organization] (node_modules/mongoose/lib/helpers/document/compile.js:191:32)
```
Seems like the UUID getter is missing an `undefined` or `null` test before trying to cast it or maybe I'm missing something.  
If I remove `organization` the error does not occur.</div>

PR #13034, opened 2023-02-15
schematype(UUID): added null check to prevent error on binaryToString conversion

lpizzinidev:
closes #13032


Freezystem:
seems a duplicate from https://github.com/Automattic/mongoose/pull/13029

vkarpov15 (approved):
I'll merge both this and #13029 so you both get a contrib. Sorry for delay in code review, we've been working on Mongoose 7. I think this PR has better tests, and prefer the `!= null` check, but #13029 adds a couple of extra checks that are helpful.

Freezystem:
thanks a lot 👍🏼

PR #13036, opened 2023-02-15
Mongodb v4.13 Memoryleak fix

samermurad:
**Summary**

According to https://www.mongodb.com/docs/drivers/node/current/whats-new/#what-s-new-in-4.14, `mongodb@4.13` has a memoryleak which is fixed in `mongodb@4.14`. Naturally they recommend to upgrade to 4.14

Other than that, the 4.14 version seems to only have "Deprecated methods and options that reference the legacy Logger." so the upgrade should be a no-brainer

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "bson": "^4.7.0",
     "kareem": "2.5.1",
-    "mongodb": "4.13.0",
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
LGTM, merging this into 6.10 branch.

Issue #10302, opened 2021-05-28
[Typscript] Wrong type for `new Model(doc)`


Zenthae:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

When using `new Model(doc)` to create a new model and the `save()` it. the type of `doc` is always `any` which break the auto-completion for properties added by the user model.

![image](https://user-images.githubusercontent.com/32647899/119918891-9065e400-bf69-11eb-918d-cab2e65e34b5.png)

**If the current behavior is a bug, please provide the steps to reproduce.**

[Quick Start](https://mongoosejs.com/docs/index.html) example from the website

`tsconfig.json`
```json
{
  "compilerOptions": {
    "target": "ES2020",
    "module": "commonjs",
    "outDir": "./dist",
    "rootDir": "./src",
    "strict": true,
    "moduleResolution": "node",
    "types": ["mongoose"],
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true
  },
  "exclude": ["node_modules"],
  "include": ["src/**/*.ts"]
}
```

**What is the expected behavior?**

instead of having `any` it should use the provided type.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

nodejs: v16.1.0
mongoose: 5.12.10

DVGY:
const Kitten = mongoose.model<IKitten>('Kitten', kittenSchema);

Is this line added correctly in your code ?

Zenthae:
yes it is, it's the line hidden by Vscode popup

DVGY:
> yes it is, it's the line hidden by Vscode popup
@Zenthae change the line 
from this

`const Kitten = mongoose.model('Kitten', kittenSchema);`

to this
`const Kitten = mongoose.model<Ikitten>('Kitten', kittenSchema);`

Zenthae:
same

DVGY:
> same

@Zenthae one more thing make sure your `interface IKitten extends Document` is like this

Zenthae:
still the same issue

DVGY:
> still the same issue

can you share on codepen ?

vkarpov15:
Make sure you're running Mongoose >= 5.12.6, we changed this type definition with #10074 so either you're using an older version of Mongoose or VSCode is picking up the wrong version of Mongoose type definitions.

Zenthae:
[...]

PR #13038, opened 2023-02-15
Don't allow Model.create() with keys that don't exist in the schema

JavaScriptBach:
**Summary**

Currently the Mongoose types allow to call `MyModel.create()` while specifying attributes that don't exist in the schema. This removes a lot of the benefits of static type-checking, as discussed in https://github.com/Automattic/mongoose/issues/10302.

This PR disallows nonexistent attributes without requiring the user to manually specify the type on every create call. It is not backwards compatible but I think it's a better default for people using TypeScript and is worth including in Mongoose 7. Thoughts?

**Examples**

Updated tests.

### DIFF
--- a/test/types/connection.test.ts
+++ b/test/types/connection.test.ts
@@ -120,10 +120,6 @@ export function autoTypedModelConnection() {
 
   (async() => {
   // Model-functions-test
-  // Create should works with arbitrary objects.
-    const randomObject = await AutoTypedModel.create({ unExistKey: 'unExistKey', description: 'st' });
-    expectType<AutoTypedSchemaType['schema']['userName']>(randomObject.userName);
-
     const testDoc1 = await AutoTypedModel.create({ userName: 'M0_0a' });
     expectType<AutoTypedSchemaType['schema']['userName']>(testDoc1.userName);
     expectType<AutoTypedSchemaType['schema']['description']>(testDoc1.description);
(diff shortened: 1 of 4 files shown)


hasezoey:
i think the generic way should stay for now, but then be changed to a `extends` instead of *just* a default assignment

personally i would be for either creating a new function that is always strict for create (because adding a option for `strict` would probably not work with a extends type)

vkarpov15:
Nope, not something we can reasonably support. Because of runtime type casting, you can pass all sorts of values that TypeScript thinks are invalid to `create()` and have everything work as expected.

JavaScriptBach:
> Because of runtime type casting, you can pass all sorts of values that TypeScript thinks are invalid to `create()` and have everything work as expected.

IMO this would be a feature, not a bug. It should be fine to have stricter types than what is allowed at Mongoose runtime, just like how you can write a correct JavaScript program that can't be typed using TypeScript. Developers using TypeScript inherently sacrifice some flexibility when writing code, expecting in return that the compiler will catch some mistakes. In this context, it's pretty strange to define a typed schema and then have `create()` still allow arbitrary input objects.

vkarpov15:
Not necessarily. If you're using some validation library like zod, do you expect to be able to pass in arbitrary input to the validation? [Looks like yes](https://github.com/colinhacks/zod/blob/9b7dd816e92e16ca735e488b77e280a92a84ed64/README.md#parse).

Issue #13033, opened 2023-02-14
Make sure search generation works with 7.0


vkarpov15:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Make sure search works with 7.0 branch given changes in #12223

PR #13044, opened 2023-02-17
Improve search generation script

hasezoey:
**Summary**

This PR aims to improve the search generation script, by doing:
- move the script from `/docs` to `/scripts` (because it is not a browser script)
- disconnect on error (had it hanging for me)
- add useful error when no config is found
- change to use `/docs/api/file.html` over `/docs/api.html` (re https://github.com/Automattic/mongoose/pull/12223#issuecomment-1428470684 & https://github.com/Automattic/mongoose/issues/13033)
- use `Object.entries` over creating a one-time use array
- disable `strictQuery` warning

### DIFF
--- a/docs/search.js
+++ b/docs/search.js
@@ -40,7 +40,7 @@ for (const [filename, file] of Object.entries(filemap)) {
         const content = new Content({
           title: `API: ${prop.string}`,
           body: prop.description,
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #13045, opened 2023-02-17
Update some scripts to be more consistently in place

hasezoey:
**Summary**

This PR does:
- move all generation scripts that are somewhere in `/docs` to `/scripts`
- combine some scripts to lessen file load where possible (like `home.js` or docs-subdirectory `index.js`)
- add script for `loadSponsorData` (because i couldnt find any other entry-point)

the last one is `search.js`, but that is handled by #13044


vkarpov15:
I'd prefer to keep the "data" for docs generation in docs. So I'm fine to move logic for generating the docs to `scripts`, but I think the js files that define the docs structure like `exports['docs/api.pug'] = require('./api');` should stay under `docs`. What do you think @hasezoey ?

hasezoey:
> I'd prefer to keep the "data" for docs generation in docs.

i prefer to keep the data there too, as long as they are required from the final client side, but those moved scripts are all what i classify as "generation" script, this keeps them in a more consistent (personally easier to find) place
and also does not conflict or gets confused with actual paths (or scripts that will be required for the client-side)

if i should change it back, i would move them to something like all in `/docs/source` instead of some also in `/docs` directly (makes them easier to find then)

vkarpov15:
Yeah, `docs/source` is a fine place to put them.

hasezoey:
just to confirm, only move everything that is currently (in this PR) in `/scripts/docs`, or some other scripts like `loadSponsorData` too?

vkarpov15:
Bring back `docs/source/home.js`, `docs/source/index.js`, `docs/tutorials/index.js`, `docs/typescript/index.js`. Remove the contents of those files from `scripts/docs/docsIndex.js`.

For consistency, move `docs/tutorials/index.js` -> `docs/source/tutorials.js`, `docs/typescript/index.js` -> `docs/source/typescript.js`.

hasezoey:
with 3497772b4984da2c63bc4f06d0d933ba2d641d86 i have moved the mentioned scripts to `/docs/source`, also re-created `home.js`, `typescript.js` and `tutorials.js` - but not in the state as they have been before the PR, they now use shared code (via `/docs/source/utils.js`)

scripts/loadSponsorData.js
```diff
@@ -3,14 +3,21 @@
 const axios = require('axios');
 const config = require('../.config');
 const fs = require('fs');
-const mongoose = require('../');
+const mongoose = require('../types');
```
vkarpov15: I don't think this change will work, since we use `mongoose.connect()` here.

docs/source/home.js
```diff
@@ -1,7 +1,3 @@
+const package = require('./../../package.json')
 
-var package = require('./../../package.json')
-
-module.exports = {
-    package: package
-  , title: 'ODM'
-}
+module.exports = { package: package, title: 'ODM' };
```
vkarpov15, line 3: This script doesn't look like it is imported anywhere?

docs/source/docsIndex.js
```diff
@@ -0,0 +1,95 @@
...
+docs['docs/jobs.pug'] = {
+  title: 'Mongoose MongoDB Jobs',
+  jobs
+};
+docs['docs/change-streams.md'] = { title: 'MongoDB Change Streams in NodeJS with Mongoose', markdown: true };
+docs['docs/lodash.md'] = { title: 'Using Mongoose with Lodash', markdown: true };
+docs['docs/incompatible_packages.md'] = { title: 'Known Incompatible npm Packages', markdown: true };
+
+// do sub-directories
+require('./typescript');
```
vkarpov15, line 87: I think you would need to pass `docs` in here somehow? I don't think `require('./typescript')` actually modifies the exports of this file

docs/source/docsIndex.js
```diff
@@ -0,0 +1,95 @@
+
+'use strict';
```
vkarpov15, line 2: Why not just name this file `docs/source/index.js`?

scripts/loadSponsorData.js
hasezoey: had also noticed this, i dont know why, but vscode seems to want to resolve it there, have fixed it

docs/source/home.js
hasezoey: seems like it, removed with cef2843ccd55125781b2d1c456e4add963da1211, it was just touched up by this pr because of auto-format

docs/source/docsIndex.js
hasezoey: i had renamed it originally for the move to `/scripts` and didnt rename it back, should i rename it back?

docs/source/docsIndex.js
hasezoey: true, fixed with 0681d38f0907e4d5fc8c53c4c8e100f105f50caa

hasezoey:
will re-do this PR based on the latest master with all suggestions applied, which will improve commit history and file blame and hopefully not result in conflicts like this current PR might

docs/source/home.js
hasezoey: while re-doing this PR i noticed that the entry for `home` in `source/index` was lost in translation somewhere

hasezoey:
rework is #13166

PR #13059, opened 2023-02-20
fix(query): avoid sending unnecessary empty projection to MongoDB server

### LINKED ISSUE #13050: Avoid sending empty `projection` option with queries
### Prerequisites

### Mongoose version

6.9.1

### Node.js version

16.x

### MongoDB server version

Any

_No response_

### Description

The following script shows that Mongoose sends `projection: {}` to the MongoDB driver, which is unnecessary. Would be ideal if Mongoose didn't set a `projection` option if the projection is empty.

```javascript
'use strict';

mongoose.set('debug', true);

async function run() {

  // Prints "Mongoose: tests.findOne({}, { projection: {} })"
  await Test.findOne(); // Sends `projection: {}` to MongoDB
}
```

### Steps to Reproduce

See above

### Expected Behavior

_No response_

vkarpov15:
Fix #13050

**Summary**

Right now we always send `projection` option to the MongoDB server, even if `projection` is an empty object. This is a bit clunky for numerous reasons: perf, messy logs, etc. With this change, we won't send `projection` if it is an empty object.

**Examples**

No projections in below script:

```javascript
'use strict';
  
mongoose.set('debug', true);

async function run() {

  await Test.findOne();

  await Test.find();

  await Test.findOneAndUpdate({}, { name: 'bar' });
}
```

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -1876,6 +1876,12 @@ Query.prototype.setUpdate = function(val) {
  */
 
 Query.prototype._fieldsForExec = function() {
+  if (this._fields == null) {
+    return null;
+  }
+  if (Object.keys(this._fields).length === 0) {
+    return null;
+  }
   return utils.clone(this._fields);
 };
 
@@ -1996,6 +2002,12 @@ Query.prototype._optionsForExec = function(model) {
       delete options.wtimeout;
     }
   }
+
+  const projection = this._fieldsForExec();
+  if (projection != null) {
+    options.projection = projection;
+  }
+
   return options;
 };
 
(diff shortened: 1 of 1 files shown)


hasezoey (approved):

AbdelrahmanHafez (commented):
Other than the missing tests for this, LGTM. 👍 
We could test this using `sinon.spy(...)`.

Uzlopak (approved):
makes sense, but should have a unit test.

vkarpov15:
Would be nice to have a test, I opened an issue to track.

PR #13066, opened 2023-02-21
fix(types): use MergeTypes for type overrides in HydratedDocument

### LINKED ISSUE #13040: DocumentArrays defined in TMethodsAndOverrides do not correctly override Arrays.
### Prerequisites

### Mongoose version

6.9.1

### Node.js version

12.22.10

### MongoDB server version

5

4.9.5

### Description

I was following this example from the official docs: https://mongoosejs.com/docs/typescript/subdocuments.html#subdocument-arrays.

Unfortunately, there is problem in the provided example:

```typescript
// Given
const doc = new UserModel({});

// This indeed works:
doc.names[0].ownerDocument(); // Works!

// But this does not!
doc.names.forEach(name => {
  name.ownerDocument() // Does not work! no property `ownerDocument`
});
```

Seems that Array methods, such as `forEach`, `find`, etc. are not correctly overridden by DocumentArray.

### Steps to Reproduce

 Here's a full custom example:

```typescript
// First I define a few simple types

interface Product {}

interface User {
  products: Product[];
}

// I want my `UserDocument` type to define `products` as a `DocumentArray`; I'll do this using overrides

interface UserOverrides {
  products: Types.DocumentArray<Product>;
}

interface UserModel extends Model<User, {}, UserOverrides> {}

// Here I determine the type of user documents; I could also manually define a `HydratedDocument` - makes no difference.

interface UserDocument extends InstanceType<UserModel> {}

// Assume I have an instance of `UserDocument`

let user: UserDocument;

// This is then fine:

user.products[0].ownerDocument(); // ok

// While this is not:

user.products.forEach(product => {
  product.ownerDocument(); // error - no property named `ownerDocument`
});
```

If `interface User` did not have `products: Product[]` defined, the error would go away. This is not a great solution however.

### Expected Behavior

Using Array prototype methods such as forEach, find and others on a DocumentArray override should correctly infer array entry types to be Subdocuments, not POJOs.

vkarpov15:
Fix #13040

**Summary**

Use `MergeTypes` to correctly omit doc paths that conflict with `TMethodsAndOverrides`. The problem in #13040 is that `products` has type `Product[] | Types.DocumentArray<Product>`.

The one test change is due to tsd assertion quirks. I confirmed that `Omit<X, never>` is equivalent to `X`, inasmuch as anything is equivalent in TypeScript.

image

Also bumped a couple of dependencies that we already bumped in v7.0 to avoid npm audit warnings.

**Examples**


IslandRhythms:
LGTM

IslandRhythms (approved):

PR #13067, opened 2023-02-21
Revert "Revert "typescript: Allow null for optional document fields""

vkarpov15:
**Summary**

Apply fix for #12748 to 7.0

@hasezoey can you please take a look at this and see if it causes any issues in any Typegoose sample apps you have?

**Examples**


hasezoey:
> @hasezoey can you please take a look at this and see if it causes any issues in any Typegoose sample apps you have?

i did not see any specific problems aside from what i mentioned in https://github.com/Automattic/mongoose/issues/13072

commands i used, because this PR does not base itself on 7.0 (but on master):
```sh
git checkout 7.0
git checkout -b test70revertnull
git merge upstream/revert-12951-revert-12781-infer-null
```

Though the tests are failing in this PR

vkarpov15:
I took a look and we're going to have to close this one for now. There are some unfortunate nuances that we didn't catch in prior testing: nested paths end up as `required: false` and are nullable in TypeScript, which isn't necessarily true at runtime. And this is causing the test failures in d4c2ae8. We'll defer this to a future release.

Issue #10312, opened 2021-05-31
Switch from Pug to server side rendered Vue for docs


vkarpov15:
**Do you want to request a *feature* or report a *bug*?**

Docs refactor

**What is the current behavior?**

Docs layout has been written in Pug since at least 2012, but I think it's time to make a change. The key benefit is that Vue templates are just HTML, which makes it easier to iterate on layout without tooling. See [server side rendering with Vue](https://masteringjs.io/tutorials/vue/ssr).

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

hasezoey:
small side question, why not use something like docusaurus (v2) with versioning?

vkarpov15:
Never use a framework when you can just use a for loop :)

hasezoey:
would like to actually do this for 7.0 (so it does not happen in minor / patch versions), but have still have a question to resolve:

should we really go for vue instead of something simpler like [`handlebars`](https://handlebarsjs.com/guide/#what-is-handlebars)? because we only need templating, like applying a base layout to all output files, and some basic logic with looping or conditional creation, and everything is static anyway (client side code we handle our-self) (and handlebars itself is just plain HTML too)

vkarpov15:
I'd rather not prioritize making this switch for now. We wanted to make this change for the sake of standardizing Mongoose docs with some other projects we work on, but I don't think that's important enough to justify the work. We'll close this issue for now unless you feel strongly about making this change. My only feedback would be that I'd rather keep pug than switch to handlebars, I haven't had good experiences working with handlebars.

PR #13084, opened 2023-02-24
Misc Documentation changes

hasezoey:
**Summary**

This PR does some miscellaneous documentation improvements / changes:
- remove file `prior.pug`, because it is not generated anywhere since at least mongoose 4.0
- remove file `contributing.md` (in `/docs`), because that file was not listed in any sidebar and didnt provide any helpful things
- change `api.pug` to be `api.md`, to lessen the amount of pug files (re #10312)
- change to use `fileName` in `api_split.pug`
- convert some documentation scripts from `var` to `const / let`
- fix `Query.prototype.toStringTag` not having a `ctx.constructor`
- change Symbols to now be listed as `[Symbol.function]` instead of `.function` or `.Symbol.function`
- add extra handling for symbols to be listed without a extra dot (now listed as `.prototype[Symbol.function]` instead of `.prototype.[Symbol.function]`)
- fix `schematypes`'s `isRequired` being listed as `validators` in documentation and at runtime
- change search script to use `history.pushState` instead of always reloading
- change search script to also use navbar search and disable nav-bar search when being in `/search.html` (but encourage the use of the big input)
- update `purecss` from `1.0.0` to `1.0.1`


docs/js/navbar-search.js
```diff
@@ -1,19 +1,24 @@
-var defaultVersion = '6.x';
-
 (function() {
-  var versionFromUrl = window.location.pathname.match(/^\/docs\/(\d+\.x)/);
-  var version = versionFromUrl ? versionFromUrl[1] : defaultVersion;
+  const defaultVersion = '6.x';
```
vkarpov15, line 2: Mental note to self to check why `defaultVersion = '6.x'` here

vkarpov15: Turns out it doesn't really matter what the default version is. I'll remove that code in this PR.

hasezoey: probably because of a merge of the 6.x branch, or just plainly forgotten to change with the release of 7.0 (https://github.com/Automattic/mongoose/blob/791ab3e7e7929564340ab4c56ade2c7cbb1fe5fc/docs/js/navbar-search.js#L1)

vkarpov15 (approved):

PR #13089, opened 2023-02-27
types(documentarray): typed DocumentArray constructor parameter

### LINKED ISSUE #13087: `DocumentArray` constructor accepts array of any type what can cause runtime error
### Prerequisites

### Mongoose version

6.10.0

### Node.js version

18.12.0

### MongoDB server version

N/A

4.9.5

### Description

`DocumentArray` constructor [accepts](https://github.com/Automattic/mongoose/blob/6.10.0/types/types.d.ts#L64) `values: any[]`:
```ts
class DocumentArray<T> extends Types.Array<T extends Types.Subdocument ? T : Types.Subdocument<InferId<T>> & T> {
    /** DocumentArray constructor */
    constructor(values: any[]);
```
This can result in runtime error when passed to constructor value won't have required field.

### Steps to Reproduce

This code compiles and results in runtime error:
```ts
import { Types } from 'mongoose';

interface Book {
    author: {
        name: string;
    };
}

const books: Types.DocumentArray<Book> = new Types.DocumentArray<Book>([1, 2, 3]);
console.log(books[0].author.name);
```
Error:
```
console.log(books[0].author.name);
                            ^
TypeError: Cannot read properties of undefined (reading 'name')
```

### Expected Behavior

Compile time error:
```
Type 'number' is not assignable to type 'Book'
```

lpizzinidev:
Changed `DocumentArray` constructor parameter type from `any[]` to `(T & AnyObject)[]`.

closes #13087

### DIFF
--- a/test/types/docArray.test.ts
+++ b/test/types/docArray.test.ts
@@ -1,4 +1,4 @@
-import { Schema, model, Document, Types } from 'mongoose';
+import { Schema, model, Types, InferSchemaType } from 'mongoose';
 import { expectError, expectType } from 'tsd';
 
 async function gh10293() {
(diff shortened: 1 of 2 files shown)


test/types/docArray.test.ts
```diff
@@ -24,3 +24,47 @@ async function gh10293() {
...
+  const routeSchema = new Schema({
+    points: { type: [pointSchema] }
+  });
+
+  type Route = InferSchemaType<typeof routeSchema>;
+
+  expectError(function getTestRouteData(): Route {
+    return {
+      points: new Types.DocumentArray([
+        { name: 'Test' } // "location" is missing
```
vkarpov15, line 66: The major problem with this approach is that you'll still get an error even if `location` has a default.

Even if we work around the defaults issue, still doesn't handle the case of custom getters, or properties that are set in middleware (e.g. timestamps).

types/types.d.ts
```diff
@@ -61,7 +61,7 @@ declare module 'mongoose' {
 
     class DocumentArray<T> extends Types.Array<T extends Types.Subdocument ? T : Types.Subdocument<InferId<T>> & T> {
       /** DocumentArray constructor */
-      constructor(values: any[]);
+      constructor(values: (T & AnyObject)[]);
```
vkarpov15, line 64: AnyObject is realistically the best we can do here.

vkarpov15 (changes requested):

vkarpov15 (approved):
Thanks :+1:

PR #13092, opened 2023-02-27
BREAKING CHANGE: copy schema options when merging schemas using `new Schema()` or `Schema.prototype.add()`

vkarpov15:
**Summary**

A neat change/feature that was recently requested: if using `schema1.add(schema2)` and `schema1` doesn't have a `toJSON` or `toObject` option set, use schema2's `toJSON` and `toObject` options.

**Examples**


IslandRhythms (approved):
LGTM

adamreisnz:
Yep I like this, thanks @vkarpov15 !

PR #13099, opened 2023-02-28
Fix v6 docs link

fpereira1:
**Summary**

Broken link on the website

**Examples**

Current link to v6 docs lead to a 404
image


vkarpov15 (commented):
This change is no longer necessary, https://mongoosejs.com/docs/6.x/docs/guide.html now works. Either way, `/docs/v6` was a long out of date version of the 6.x docs that we had mistakenly deployed several years ago.

PR #13113, opened 2023-03-02
Change API documentation headers to be more unique

### LINKED ISSUE #12901: [Docs] Refactor header naming to lessen conflicts
### Prerequisites

### Issue

Currently in the documentation, a naming of `#type_Type-function` is used, which can cause conflicts between *static* and *instance* methods, for example:
`#model_Model-remove` refers to both *instance* and *static* remove, which is not correct.

Also the `type_Type` is redundant, so it could be removed.
Personally i would like to not have to use band-aids for special cases and just have it correctly separated.

My proposal would be: `#Type(.prototype).function`, real examples:
- `#Model.remove` to refer to the *static* remove
- `#Model.prototype.remove` to refer to the *instance* remove

(basically like it is in JS directly)

i would target this towards 7.0 (because it would be a rather drastic change for the documentation.

as for backwards compatibility (like bookmarks), we could add hidden elements that still have the same old header id and so could still refer to the same place (and maybe some script to translate it to the new id)

cc @vkarpov15 @Uzlopak

hasezoey:
**Summary**

This PR changes the API documentation headers to be more unique and now have the style of how they would appear in js itself, also includes a script to translate the old header id's to the new ones.

Notes:
because the old id's did not differentiate between static and instance properties and also not between properties and functions, there are multiple tests to see if one exists in the new style, where the first found is used, current order is:
- properties first, functions second
- static first, instance second

closes #12901

Examples:
- `mongoose_Mongoose` -> `Mongoose()`
- `mongoose_Mongoose-model` -> `Mongoose.prototype.model()`
- `model_Model-find` -> `Model.find()`


vkarpov15 (approved):
Thanks :+1:

PR #13114, opened 2023-03-02
feat(types): add `BigInt` support

### LINKED ISSUE #13081: [7.x] Add native BigInt Support using latest driver
### Prerequisites

### 🚀 Feature Proposal

New mongodb driver 5.1 released with JS BigInt support & important bug fixes, consider updating the driver & adding support for it.
See release notes for more details:
https://github.com/mongodb/node-mongodb-native/releases/tag/v5.1.0

### Motivation

_No response_

### Example

_No response_

lpizzinidev:
**Summary**
Added support for `BigInt`.
Support for the data type has been introduced in [MongoDB driver 5.1](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.1.0).

Closes #13081


vkarpov15 (changes requested):
I don't agree with the high level direction of this PR. Specifically the decision to create a separate `mongoose.Types.BigInt` class.

The crux of BigInt support in `mongodb@5.1` is support for JavaScript's native [BigInt primitive](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt). So Mongoose's treatment of BigInts should be analogous to Mongoose's treatment of numbers, not ObjectIds.

1. `castBigInt()` should check for `typeof value === 'bigint'`, and, if so, return `value`. Otherwise it should attempt to convert using `BigInt()`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
2. We should add schema-level, collection-level, and global 'useBigInt64' options

Issue #13070, opened 2023-02-22
Fix model.populate() typescript return type to support paths


meness:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

#### Model
```typescript
/** Populates document references. */
populate(docs: Array<any>, options: PopulateOptions | Array<PopulateOptions> | string,
      callback?: Callback<(HydratedDocument<T, TMethodsAndOverrides, TVirtuals>)[]>): Promise<Array<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>>;

populate(doc: any, options: PopulateOptions | Array<PopulateOptions> | string,
      callback?: Callback<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>): Promise<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>;
```

#### Query
```typescript
/** Specifies paths which should be populated with other documents. */
populate<Paths = {}>(path: string | string[], select?: string | any, model?: string | Model<any, THelpers>, match?: any): QueryWithHelpers<UnpackedIntersection<ResultType, Paths>, DocType, THelpers, UnpackedIntersection<RawDocType, Paths>>;

populate<Paths = {}>(options: PopulateOptions | (PopulateOptions | string)[]): QueryWithHelpers<UnpackedIntersection<ResultType, Paths>, DocType, THelpers, UnpackedIntersection<RawDocType, Paths>>;
```

`query.populate()` is what I need that applies paths. I use `model.populate()` because I have to use `aggregate()`. `aggregate()` receives a type, maybe you can use the return type instead of `any`.

PR #13117, opened 2023-03-02
Add Path generic

IslandRhythms:
It wasn't yelling at me.

Also, while working on this issue I noticed the populate section is more focused on query population than calling `Model.populate()`. Maybe we should include a small section about that function since we have a section titled `Populate` in the docs.

### DIFF
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -366,7 +366,7 @@ async function gh13070() {
   const child = model<IChild>('Child', childSchema);
 
   const doc = await parent.findOne();
-  await child.populate<{child: IChild}>(doc, 'child');
+  await child.populate<{ child: IChild }>(doc, 'child');
 
 
 }
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -365,7 +365,7 @@ async function gh13070() {
   const Parent = model<IParent>('Parent', parentSchema);
   const Child = model<IChild>('Child', childSchema);
 
-  const doc = await Parent.findOne();
+  const doc = await Parent.findOne().orFail();
   const doc2 = await Child.populate<{ child: IChild }>(doc, 'child');
-  const name: string = doc.child.name;
+  const name: string = doc2.child.name;
 }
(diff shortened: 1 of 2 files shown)


test/types/populate.test.ts
```diff
@@ -339,3 +339,34 @@ function gh12136() {
...
+    });
+
+  const childSchema = new Schema(
+    {
+      name: { type: String, required: true },
+      parent: { type: Schema.Types.ObjectId, ref: 'Parent', required: true }
+    });
+
+  const parent = model<IParent>('Parent', parentSchema);
+  const child = model<IChild>('Child', childSchema);
```
vkarpov15: We always use capitalized names for models

test/types/populate.test.ts
```diff
@@ -339,3 +339,34 @@ function gh12136() {
...
+
+  const doc = await parent.findOne();
+  await child.populate<{child: IChild}>(doc, 'child');
```
vkarpov15: Didn't actually test that the result doc is correctly populated. Missed this: https://github.com/Automattic/mongoose/blob/7d2c511475fb4a36d8ffcdd30739af45cdde4d4e/test/types/populate.test.ts#L126

types/models.d.ts
```diff
@@ -303,7 +303,8 @@ declare module 'mongoose' {
       callback?: Callback<(HydratedDocument<T, TMethodsAndOverrides, TVirtuals>)[]>): Promise<Array<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>>;
     populate(doc: any, options: PopulateOptions | Array<PopulateOptions> | string,
       callback?: Callback<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>): Promise<HydratedDocument<T, TMethodsAndOverrides, TVirtuals>>;
-
+    populate<Paths={}>(docs: Array<any>, options: PopulateOptions | Array<PopulateOptions> | string): Promise<MergeType<this, Paths>>;
```
vkarpov15: If you had actually tested the populate doc result, you'd see that `this` is the wrong type to use here. In this context, `this` is the Model class, not the document. Use the other `populate()` calls on `Model` as an example for how to do this correctly.

vkarpov15 (approved):
For the future, please reference the closed issue in the commit message (#13070) in this case. And please avoid spelling mistakes in git commit messages.

PR #13120, opened 2023-03-03
fix(types/query): remove "update" function

hasezoey:
**Summary**

This PR removes the function `update` from the query types, because it has been removed for 7.0 in runtime

re #12955
re #13039


vkarpov15 (approved):
Thanks

PR #13124, opened 2023-03-03
6.x should still support node 12

lorand-horvath:
@vkarpov15 I think the bump to node 14 minimum was accidental, this reverts mongoose 6.x to still be supported on node 12.

**Examples**


vkarpov15 (approved):
Absolutely accidental. Thank you for catching this, and I'm so sorry for the trouble.

PR #13126, opened 2023-03-04
fix(aggregate): added `await` to prevent exception in aggregate exec

### LINKED ISSUE #13125: TypeError: cursor.toArray is not a function
### Prerequisites

### Mongoose version

7.0.0

### Node.js version

16.17.1

### MongoDB server version

6.0.4 (offline)

_No response_

### Description

If I run my application and keep the database offline, I can crash the application by running an aggregation due to a typo in `lib/aggregate.js`. Stack trace:
```
node_modules/mongoose/lib/aggregate.js:1054
    result = await cursor.toArray();
                          ^

TypeError: cursor.toArray is not a function
    at Aggregate.exec (node_modules/mongoose/lib/aggregate.js:1054:27)
    at async default (web/routes/index.js:4:19)
```

When I navigate to the code it's complaining about, the issue seems to be that `await cursor.toArray();` should be `(await cursor).toArray();` in the `exec` function:
```
try {
  const cursor = collection.aggregate(this._pipeline, options);
  result = await cursor.toArray();
} catch (error) {
...
```

### Steps to Reproduce

Simply run an aggregation using `.exec()` while the database connection has not been established.

### Expected Behavior

_No response_

lpizzinidev:
closes #13125

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -15,7 +15,6 @@ const stringifyFunctionOperators = require('./helpers/aggregate/stringifyFunctio
 const utils = require('./utils');
 const read = Query.prototype.read;
 const readConcern = Query.prototype.readConcern;
-const STATES = require('./connectionstate');
 
 const validRedactStringValues = new Set(['$$DESCEND', '$$PRUNE', '$$KEEP']);
 
@@ -1051,12 +1050,7 @@ Aggregate.prototype.exec = async function exec() {
   let result;
   try {
     const cursor = collection.aggregate(this._pipeline, options);
-    const conn = this._model.db;
-    if (conn._readyState !== STATES.connected) {
-      result = (await cursor).toArray();
-    } else {
-      result = await cursor.toArray();
-    }
+    result = await (await cursor).toArray();
   } catch (error) {
     await new Promise((resolve, reject) => {
       const _opts = { error: error };
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -15,6 +15,7 @@ const stringifyFunctionOperators = require('./helpers/aggregate/stringifyFunctio
 const utils = require('./utils');
 const read = Query.prototype.read;
 const readConcern = Query.prototype.readConcern;
+const STATES = require('./connectionstate');
 
 const validRedactStringValues = new Set(['$$DESCEND', '$$PRUNE', '$$KEEP']);
 
@@ -1050,7 +1051,12 @@ Aggregate.prototype.exec = async function exec() {
   let result;
   try {
     const cursor = collection.aggregate(this._pipeline, options);
-    result = await cursor.toArray();
+    const conn = this._model.db;
+    if (conn._readyState !== STATES.connected) {
+      result = (await cursor).toArray();
+    } else {
+      result = await cursor.toArray();
+    }
   } catch (error) {
     await new Promise((resolve, reject) => {
       const _opts = { error: error };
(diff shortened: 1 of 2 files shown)


wfjake:
I was able to test this fix in my project and confirmed that I no longer get a TypeError, just the expected timeout. Thanks for the good work!

lib/aggregate.js
```diff
@@ -1050,7 +1050,7 @@ Aggregate.prototype.exec = async function exec() {
   let result;
   try {
     const cursor = collection.aggregate(this._pipeline, options);
-    result = await cursor.toArray();
+    result = await (await cursor).toArray();
```
Uzlopak, line 1053: Actually very unsexy. But it is what it is.

test/aggregate.test.js
```diff
@@ -1242,4 +1242,16 @@ describe('aggregate: ', function() {
     });
   });
 
+  it('should not throw error if database connection has not been established (gh-13125)', async function() {
+    const m = new mongoose.Mongoose();
+    const mySchema = new Schema({ test: String });
+    const M = m.model('Test', mySchema);
+
+    const aggregate = M.aggregate();
+    aggregate.match({ $text: { $search: 'test' } });
```
vkarpov15, line 1251: Probably best to not use text search because that requires a special index

test/aggregate.test.js
```diff
@@ -1242,4 +1242,16 @@ describe('aggregate: ', function() {
...
+
+    assert.doesNotThrow(async function() {
+      await aggregate.exec();
```
vkarpov15, line 1254: I suspect this will cause some issues because this operation will eventually time out because of buffering. Probably best to save the promise and then await on it later, for example `const promise = aggregate.exec(); await m.connect(start.uri); const res = await promise;` similar to https://github.com/Automattic/mongoose/blob/c4fd23d64087510d3293f7168b06c284abf52a3c/test/collection.test.js#L37

lib/aggregate.js
vkarpov15: I'd prefer `await cursorPromise.then(cursor => cursor.toArray())`

vkarpov15 (changes requested):

test/aggregate.test.js
hasezoey: or like in other parts of the tests, use [`await assert.rejects`](https://nodejs.org/api/assert.html#assertrejectsasyncfn-error-message)

lib/aggregate.js
```diff
@@ -1050,7 +1050,7 @@ Aggregate.prototype.exec = async function exec() {
   let result;
   try {
     const cursor = collection.aggregate(this._pipeline, options);
```
AbdelrahmanHafez, line 1052: Why not just add an `await` here? Am I missing something?
```suggestion
    const cursor = await collection.aggregate(this._pipeline, options);
```

lib/aggregate.js
```diff
@@ -1049,7 +1049,7 @@ Aggregate.prototype.exec = async function exec() {
   const options = clone(this.options || {});
   let result;
   try {
-    const cursor = collection.aggregate(this._pipeline, options);
+    const cursor = await collection.aggregate(this._pipeline, options);
```
vkarpov15, line 1052: It looks like this still throws a `MongoNotConnectedError: Client must be connected before running operations`. I'll merge this PR into master so I can more easily debug.

vkarpov15 (approved):

PR #13135, opened 2023-03-06
fix(document): avoid setting array default if document array projected out by sibling projection

### LINKED ISSUE #13003: Populate with select on nested field returns extra sub-field of array type.
### Prerequisites

### Mongoose version

6.9.0

### Node.js version

16.13.1

### MongoDB server version

6.0.3

4.1.3

### Description

A parent schema contains a reference to a child schema. The child schema contains a nested field, with a string property and an array property. When populating a parent and selecting the nested string property, the array property is also included, with  empty contents.

### Steps to Reproduce

Example code:

```

export const ParentSchema = new mongoose.Schema({
  title: String,
  childId: { type: mongoose.Schema.Types.ObjectId, ref: "Child" }
});

export const Parent = mongoose.model("Parent", ParentSchema);

const childSchema = new mongoose.Schema({
  name: String,
  properties: { foo: String, bar: [{ baz: String, qux: Boolean }] }
});

export const Child = mongoose.model("Child", childSchema);

async function run() {
  const child = await Child.create({ name: "c1", properties: { foo: "foo1", bar: [] } });

  const parent = await Parent.create({ title: "p1", childId: child._id });

  const res = await Parent.findOne({ title: "p1" }).populate({
    path: "childId",
    model: Child,
    select: { "properties.foo": 1 }
  });
  return res;
}

mongoose
  .connect("mongodb://localhost:27017/test")
  .then((_) => run())
  .then((res) => console.log(JSON.stringify(res)));
```

This prints:

```
{
  "_id": "63e16b1747e00097cf9a5957",
  "title": "p1",
  "childId": {
    "properties": { "foo": "foo1", "bar": [] },
    "_id": "63e16b1747e00097cf9a5955"
  },
  "__v": 0
}
```

Note that the `bar` field is included, unexpectedly.  If I populate the `bar` array in the child document, such as:

```
const child = await Child.create(
    {
        name: "c1", 
        properties: { foo: "foo1", bar: [{baz: "quk", qux: true}] } 
    });
```

Then I still get an empty `bar` array in the populated parent document. 

I don't encounter the same problem when `bar` is a simple string property instead of an array of objects, e.g.:

```
properties: { foo: String, bar: String }
```

Using the original example with Mongoose 5.11.5 I get the expected result:

```
{
  "_id": "63e16b1747e00097cf9a5957",
  "title": "p1",
  "childId": {
    "properties": { "foo": "foo1" },
    "_id": "63e16b1747e00097cf9a5955"
  },
  "__v": 0
}
```

### Expected Behavior

The populated `Parent` document contains only the `properties.foo` field, similar to with mongoose 5.x.

vkarpov15:
Fix #13003

**Summary**

Without this fix, `applyDefaults()` thinks it needs to apply defaults to all sibling properties of `properties.foo` when the projection is `{ 'properties.foo': 1 }`

**Examples**

### DIFF
--- a/test/docs/debug.test.js
+++ b/test/docs/debug.test.js
@@ -93,15 +93,11 @@ describe('debug: shell', function() {
   });
 
   it('should avoid sending null session option with document ops (gh-13052)', async function() {
-    const m = new mongoose.Mongoose();
-    m.set('debug', true);
-    await m.connect(start.uri);
+    mongoose.set('debug', { shell: false });
     const schema = new Schema({ name: String });
-    const Test = m.model('gh_13052', schema);
+    const Test = db.model('gh_13052', schema);
 
     await Test.create({ name: 'foo' });
     assert.equal(false, lastLog.includes('session'));
-
-    await m.disconnect();
   });
 });
--- a/test/docs/debug.test.js
+++ b/test/docs/debug.test.js
@@ -93,11 +93,20 @@ describe('debug: shell', function() {
   });
 
   it('should avoid sending null session option with document ops (gh-13052)', async function() {
-    mongoose.set('debug', { shell: false });
+    let args = [];
+    const m = new mongoose.Mongoose();
+    m.set('debug', function() {
+      args.push([...arguments]);
+    });
+    await m.connect(start.uri);
     const schema = new Schema({ name: String });
-    const Test = db.model('gh_13052', schema);
+    const Test = m.model('gh_13052', schema);
 
     await Test.create({ name: 'foo' });
-    assert.equal(false, lastLog.includes('session'));
+    assert.equal(args.length, 1);
+    assert.equal(args[0][1], 'insertOne');
+    assert.ok(!('session' in args[0][3]));
+
+    await m.disconnect();
   });
 });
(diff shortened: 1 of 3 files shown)


test/docs/debug.test.js
```diff
@@ -93,10 +93,15 @@ describe('debug: shell', function() {
   });
 
   it('should avoid sending null session option with document ops (gh-13052)', async function() {
+    const m = new mongoose.Mongoose();
+    m.set('debug', true);
```
hasezoey, line 97: does this need to stay?

vkarpov15: Unfortunately yes. Without `m.set('debug', true)` this test was failing for me locally. Not sure why.

hasezoey: i didnt see this was placed in `debug.test.js`, so i guess it is correct to have it

though from what i can tell, this test is not about testing debug. also removing that "debug true" statement locally does not fail the test for me

vkarpov15: The test is using the debug mode setting to ensure that we're not sending `session: null` to the server. I made the test use debug a bit more explicitly.

IslandRhythms (approved):
LGTM

Issue #13136, opened 2023-03-06
syncIndexes and diffIndexes does not correctly find compound indexes involving text and non-text indexe


rdeavila94:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

5.x

### Node.js version

14.x

### MongoDB server version

4.x

### Description

Given a schema with the following:
```js
      const Test = new Schema({
        title: {
          type: String
        },
        description: {
          type: String
        },
        age: {
          type: Number
        }
      }, {
        autoIndex: false
      })
```

Calling `syncIndexes` on that schema ALWAYS attempts to delete and create it. 
Calling `diffIndexes` on that schema, even after it's been synced, always returns the compound index as both a `toDrop` entry and a `toCreate` entry

### Steps to Reproduce

Run the following tests in `test/model.indexes.test.js`
```js
    it('should not re-create a compound text index that involves non-text indexes, using syncIndexes', function(done) {
      const Test = new Schema({
        title: {
          type: String
        },
        description: {
          type: String
        },
        age: {
          type: Number
        }
      }, {
        autoIndex: false
      })
  
      Test.index({
        title:'text',
        description:'text',
        age: 1
      });

      const TestModel = db.model('Test', Test);
      TestModel.syncIndexes().then((results1) => {
        assert.strictEqual(results1, []);
        // second call to syncIndexes should return an empty array, representing 0 deleted indexes
        TestModel.syncIndexes().then((results2) => {
          assert.strictEqual(results2, []);
          done();
        })
      })
    })

    it('should not find a diff when calling diffIndexes after syncIndexes involving a text and non-text compound index', function(done) {
      const Test = new Schema({
        title: {
          type: String
        },
        description: {
          type: String
        },
        age: {
          type: Number
        }
      }, {
        autoIndex: false
      })
  
      Test.index({
        title:'text',
        description:'text',
        age: 1
      });

      const TestModel = db.model('Test', Test);
      
      TestModel.diffIndexes().then((diff) => {
        assert.deepEqual(diff, {toCreate: [{age: 1, title: 'text', description:'text'}], toDrop: []});
        TestModel.syncIndexes().then(() => {
          TestModel.diffIndexes().then((diff2) => {
            assert.deepEqual(diff2, {toCreate: [], toDrop: []});
            done();
          })
        })
      })
    })
```

### Expected Behavior

The above tests to pass.

[...]

PR #13138, opened 2023-03-06
Updated the isIndexEqual function to take into account non-text index…

rdeavila94:
…es when checking compound indexes that include both text and non-text indexes

**Summary**

This PR fixes https://github.com/Automattic/mongoose/issues/13136

### DIFF
--- a/lib/helpers/indexes/isIndexEqual.js
+++ b/lib/helpers/indexes/isIndexEqual.js
@@ -28,15 +28,19 @@ module.exports = function isIndexEqual(key, options, dbIndex) {
   //   textIndexVersion: 3
   // }
   if (dbIndex.textIndexVersion != null) {
-    const weights = dbIndex.weights;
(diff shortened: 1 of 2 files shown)


lib/helpers/indexes/isIndexEqual.js
```diff
@@ -28,15 +28,19 @@ module.exports = function isIndexEqual(key, options, dbIndex) {
   //   textIndexVersion: 3
```
rdeavila94, line 28: A compound index with text and non-text fields will look like this:
  `{ name: 'text', age:1 }` becomes:
```js
   {
     v: 2,
     key: { _fts: 'text', _ftsx: 1, age: 1 },
     name: 'name_text_age_1',
     ns: 'test.tests',
     background: true,
     weights: { name: 1 },
     default_language: 'english',
     language_override: 'language',
     textIndexVersion: 3
   }
```

vkarpov15 (approved):
Thanks :+1:

Issue #13065, opened 2023-02-21
Test coverage for empty projection fixes


vkarpov15:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Add tests for #13059, using `sinon.stub()` on `collection.findOne()`, etc.

Issue #13052, opened 2023-02-19
Avoid sending null `session` option with document ops


vkarpov15:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.9.x

### Node.js version

16.x

### MongoDB server version

Any

### Description

```javascript
'use strict';

const mongoose = require('mongoose');
mongoose.set('debug', true);

run().catch(err => console.log(err));

async function run() {
  await mongoose.connect('mongodb://localhost:27017/test');
  const schema = new mongoose.Schema({ name: String });
  const Test = mongoose.model('Test', schema);

  // Mongoose: tests.insertOne({ name: 'foo', _id: new ObjectId("63f28c9760c00f9e5b3d6001"), __v: 0 }, { session: null })
  await Test.create({ name: 'foo' });
}
```

Sending a `null` session is unnecessary and confusing. `session` shouldn't show up in debug output if you're not using sessions.

### Steps to Reproduce

See description

PR #13141, opened 2023-03-07
test(query): refactored test case for empty session property

lpizzinidev:
**Summary**
Refactored test case for #13052 to align with the approach used in the test for #13065


AbdelrahmanHafez (approved):
LGTM, thanks. 👍

vkarpov15 (approved):

Issue #13062, opened 2023-02-21
validation ignored if the path set as 'select: false'


ryan1au:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Related Issue:
#730
[https://github.com/Automattic/mongoose/issues/730](issue 730)

```
...
const User = mongoose.model('User', Schema({
  name: { type: String, required: true },
  age: { type: Number, required: true, **select:false** }
});
// set user's name and age
// user={name:'Ben',age:16}
...
// then update and save in this way
const user=await User.findOne({name}); // user={name:'Ben'}
// do some work
...
await user.save(); // does not run validation on 'age'
...
```

At most time, before doc.save(), validation will run. But if select option set to false on some path, it does not run validation on that path even there's a required validator until select set to true. This idea may come from #730. That's reasonable enough.

But it's so sad that this rationale **cannot be searched on the documentation**. I only found it in source code, around line 2610:
![image](https://user-images.githubusercontent.com/30694608/220370367-5535761e-bc3a-49bc-b034-b36cf759c002.png)

I just thought full validation will run rather than select option determines. It wasted some time.
So, for making more lives much easier, I think this rule should be added to the documentation directly and explicitly, at least a warning item of rules on validation ;)

IslandRhythms:
```javascript
const mongoose = require('mongoose');

const testSchema = new mongoose.Schema({
  name: { type: String, required: true },
  age: { type: Number, required: true, select: false }
});

const Test = mongoose.model('Test', testSchema);

async function run() {
  await mongoose.connect('mongodb://localhost:27017');
  await mongoose.connection.dropDatabase();

  await Test.create({
    name: 'Test Testerson',
    age: 0
  });

  const doc = await Test.findOne();
  doc.age = null;
  doc.name = 'Test';
  await doc.save();
  console.log('done');
}

run();
```

ryan1au:
```
const mongoose = require('mongoose');

const testSchema = new mongoose.Schema({
  name: { type: String, required: true },
  age: { type: Number, required: true, select: false },
  links: { type: String, required: true, select: false }
});

testSchema.pre('save', function() {
  **this.links = undefined;**
});

const Test = mongoose.model('Test', testSchema);

async function run() {
  await mongoose.connect('mongodb://localhost:27017');
  await mongoose.connection.dropDatabase();

  await Test.create({
    name: 'Test Testerson',
    age: 0,
[...]

Issue #5800, opened 2017-11-08
using .set on nested field in pre-save hook overwrites parent when using .select


zoellner:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Parent field gets overwritten by .set of nested field in pre-save hook when document is incomplete due to .select

**If the current behavior is a bug, please provide the steps to reproduce.**
https://gist.github.com/zoellner/f022ac62a7ed968a85784edf1a6690c8

**What is the expected behavior?**
Nested field gets updated without overwriting the parent field.

**Please mention your node.js, mongoose and MongoDB version.**
node.js 6.11.1
mongoose 4.13.0
mongoDB 3.4.7

vkarpov15:
Thanks for the detailed repro, will fix asap :+1:

dietergeerts:
I just encounter this bug in version 5.9.16. Does this mean that this got reverted, and when doing changes in a sub-document, it has to be selected in the first place?

I wrote a plugin that does some extra changes based on other changes. The plugin is sure that the sub-documents exist, and it wants to update 1 specific field. Though using `this.set` completely overwrites the whole sub-document, which is not expected. Even though it was not specifically selected, and it conforms to the schema, I would expect only that path to be updated, not the whole sub-document. I must say that this is in a sub-document array.

Issue #4651, opened 2016-10-24
(4.6.2 -> 4.6.3 regression) Calling .set on a subfield of a deselected subdoc wipes out other fields on the subdoc


n-rook:
Suppose I have a schema defining a single embedded subdocument. I don't select this subdocument when I fetch the parent document, but I set a field on it anyway using .set.

In 4.6.2, this works as expected: the field I set changes, and the rest of the subdocument remains unchanged.

In 4.6.3, this wipes out the rest of the subdocument's fields, and gives it a new ID.

This feels like a bug, though it may be intended behavior.

```javascript
const mongoose = require('mongoose');
const Promise = require('bluebird');

mongoose.Promise = Promise;
const connection = mongoose.connect('mongodb://localhost/minimal', {});
mongoose.set('debug', true);

const ballSchema = mongoose.Schema({
  kind: String,
  mass: Number
});

const sportSchema = mongoose.Schema({
  ball: ballSchema
});

const Sport = mongoose.model('Sport', sportSchema);

new Sport({ball: {kind: 'birdie', mass: 4.75 }}).save()
  .then((sport) => Sport.findById(sport._id).select('-ball'))
  .then((sport) => {
    sport.set('ball.kind', 'shuttlecock');

    // On 4.6.2, this save issues this update:
    // sports.update({ _id: ObjectId(...) },
    //   { '$set': { 'ball.kind': 'shuttlecock' } })
    // On 4.6.3, it issues this update:
    // sports.update({ _id: ObjectId("...") },
    //   { '$set': { ball: { kind: 'shuttlecock', _id: ObjectId("...") } } })
    return sport.save();
  })
  .then((sport) => Sport.findById(sport._id).select('ball'))
  .then((sport) => {
    // On 4.6.2, sport.ball.mass === 4.7.5
    // On 4.6.3, sport.ball.mass is undefined
    console.log('sport.ball.mass', sport.ball.mass);
  });
```

vkarpov15:
Thanks for the detailed repro, will investigate

vkarpov15:
Yeah this is related to the fix for #4578. I would argue this is expected behavior, because when you do `.set('ball.kind')` you're implicitly creating a new `ball`, which in turn gets tracked as setting `ball` to `{ kind: 'shuttlecock' }`. What's your use case for this behavior?

n-rook:
I think it's unexpected behavior, but I agree it's ambiguous. It's not obviously a bug.

[...]

PR #13143, opened 2023-03-07
Avoid saving changes to non-modified paths

vkarpov15:
**Summary**

#13062 pointed out an issue similar to #5800, #9457, and #4651: Mongoose saves changes to non-selected paths. This is a bit of a gotcha, and something that we'd like to consider changing. What do you think of this change @hasezoey ? There are some minor changes to existing tests, so this change is a little risky.

Also made a change in the docs to highlight the gotcha that `pre('save')` changes don't go through validation in the `validation` docs.

**Examples**

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -4,11 +4,12 @@ Before we get into the specifics of validation syntax, please keep the following
 
 - Validation is defined in the [SchemaType](schematypes.html)
 - Validation is [middleware](middleware.html). Mongoose registers validation as a `pre('save')` hook on every schema by default.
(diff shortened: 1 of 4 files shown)


hasezoey (approved):
Code looks good to me.

i think it would be better to make this a feature toggle (because when changing a value you would probably expect it to save) or at least some kind of warning (or error?) that a deselected value has been changed.

is it correct to target 6.x instead of 7.0?

vkarpov15:
I opened new issue https://github.com/Automattic/mongoose/issues/13145 to track.

PR #13146, opened 2023-03-08
Extend argument type definition of select method

rbereziuk:
**Summary**

I added a more precise type definition to the `select` method. 

[Docs](https://mongoosejs.com/docs/api/query.html#query_Query-select)

**Examples**

In the code editor, it gives more information about the argument type

### DIFF
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -532,7 +532,7 @@ declare module 'mongoose' {
     replaceOne(filter?: FilterQuery<DocType>, replacement?: DocType | AnyObject, options?: QueryOptions<DocType> | null): QueryWithHelpers<any, DocType, THelpers, RawDocType>;
 
     /** Specifies which document fields to include or exclude (also known as the query "projection") */
-    select(arg: string | any): this;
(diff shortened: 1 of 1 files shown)


types/query.d.ts
```diff
@@ -532,7 +532,7 @@ declare module 'mongoose' {
     replaceOne(filter?: FilterQuery<DocType>, replacement?: DocType | AnyObject, options?: QueryOptions<DocType> | null): QueryWithHelpers<any, DocType, THelpers, RawDocType>;
 
     /** Specifies which document fields to include or exclude (also known as the query "projection") */
-    select(arg: string | any): this;
+    select(arg: string | string[] | Record<string, number>): this;
```
vkarpov15, line 535: Record<string, number> isn't enough. At minimum you need Record<string, number | boolean | object>

rbereziuk: Thanks for the remark )

hasezoey (approved):

vkarpov15 (approved):

Issue #13152, opened 2023-03-09
missing typescript overload for cursor.eachasync


krosenk729:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.6.1

### Node.js version

18+

### MongoDB server version

6+

### Typescript version (if applicable)

4.5.5

### Description

The `fn` callback for eachAsync should allow for an index parameter; however, the typescript definitions only allow for a single DocType or DocType[] param.

The function is properly shown in the documentation to allow `i` as an index here https://mongoosejs.com/docs/api/querycursor.html#querycursor_QueryCursor-eachAsync but the typescript definitions below prevent use of the second arg

```
    eachAsync(fn: (doc: DocType[]) => any, options: EachAsyncOptions & { batchSize: number }, callback: CallbackWithoutResult): void;
    eachAsync(fn: (doc: DocType) => any, options: EachAsyncOptions, callback: CallbackWithoutResult): void;
    eachAsync(fn: (doc: DocType[]) => any, options: EachAsyncOptions & { batchSize: number }): Promise<void>;
    eachAsync(fn: (doc: DocType) => any, options?: EachAsyncOptions): Promise<void>;
```

### Steps to Reproduce

Use mongoose in a `.ts` file and try `Thing.find({}).cursor().eachAsync(async (doc, i) => { })` ... this will cause a typescript error `No overload matches this call.`

hasezoey:
closing because #13153 as been merged and will be available in 7.0.2

hasezoey:
nvm, forgot this issue was for 6.x and 7.x

PR #13153, opened 2023-03-09
fix(document): add index param to eachAsync fn

krosenk729:
**Summary**

Fix for https://github.com/Automattic/mongoose/issues/13152

**Examples**

### DIFF
--- a/test/types/querycursor.test.ts
+++ b/test/types/querycursor.test.ts
@@ -13,3 +13,10 @@ Test.find().cursor().
     expectType<string | undefined>(doc.name);
   }).
   then(() => console.log('Done!'));
+
+Test.find().cursor().
+  eachAsync(async(doc: ITest, i) => {
+    expectType<Types.ObjectId>(doc._id);
+    expectType<number>(i);
+  }).
+  then(() => console.log('Done!'));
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (changes requested):
LGTM, can you please add a test in `test/typescript` to prevent that from being accidentally deleted?

krosenk729:
yes thanks for the fast review @AbdelrahmanHafez -- added a test

vkarpov15 (approved):
Thanks :+1:

I noticed in #13152 you're using Mongoose 6.6.1. Is there anything blocking you from upgrading to Mongoose 7? Since this PR is against master branch, it will go into 7.0.2. But we can also cherry pick this into 6.x.

krosenk729:
> Thanks 👍
> 
> I noticed in #13152 you're using Mongoose 6.6.1. Is there anything blocking you from upgrading to Mongoose 7? Since this PR is against master branch, it will go into 7.0.2. But we can also cherry pick this into 6.x.

@vkarpov15  a cherry pick would be much much appreciated. our last upgrade was a little complicated, so we're still gathering resources to devote to another major version bump

vkarpov15:
@krosenk729 it looks like this fix was already applied to 6.x here: https://github.com/Automattic/mongoose/pull/13162 . Upgrade to 6.10.3 for the fix :+1:

PR #13155, opened 2023-03-10
type(query): changed return type of lean() to include null

### LINKED ISSUE #13151: typescript Model#findOne() return type is not as expected
### Prerequisites

### Mongoose version

6.10.1

### Node.js version

16

### MongoDB server version

4.2

4.6.3

### Description

When I execute `findOne` method on model I am getting type annotations like below:

image

The problem above is typescript assumes that result will be always there.
So, If `latestRowEntry` is null I am getting error accessing nested props(`latestRowEntry.time`).

Ideally, Typescript should not allow the access of nested properties without checking `latestRowEntry` is null or not.

The annotation that's getting generated is not correct.

Below is my model and it's interface

```ts
import {model, Schema, Types} from 'mongoose'

export interface DeviceRawData {
  device_id: string
  date: string
  time: string
  createdAt?: Date | string
  updatedAt?: Date | string
}

export const DeviceRawDataSchema = new Schema(
  {
    device_id: {
      type: String,
      required: true,
    },
    date: {
      type: String,
      required: true,
    },
    time: {
      type: String,
      required: true,
    },
  },
  {
    timestamps: true,
  },
)

const DeviceRawDataModel = model<DeviceRawData>('DeviceRawData', DeviceRawDataSchema)
export default DeviceRawDataModel

```

### Steps to Reproduce

Use below model code and try to query using method `model#findOne`.

```ts
import {model, Schema, Types} from 'mongoose'

export interface DeviceRawData {
  device_id: string
  date: string
  time: string
  createdAt?: Date | string
  updatedAt?: Date | string
}

export const DeviceRawDataSchema = new Schema(
  {
    device_id: {
      type: String,
      required: true,
    },
    date: {
      type: String,
      required: true,
    },
    time: {
      type: String,
      required: true,
    },
  },
  {
    timestamps: true,
  },
)

const DeviceRawDataModel = model<DeviceRawData>('DeviceRawData', DeviceRawDataSchema)
export default DeviceRawDataModel
```

### Expected Behavior

The problem above is typescript assumes that result will be always there.
So, If `latestRowEntry` is null I am getting error accessing nested props(`latestRowEntry.time`).

Ideally, Typescript should not allow the access of nested properties without checking `latestRowEntry` is null or not.
The annotation that's getting generated is not correct.

lpizzinidev:
**Summary**
The `lean()` method return type has been changed to `QueryWithHelpers<LeanResultType | null, DocType, THelpers, RawDocType>`.

Closes #13151


test/types/lean.test.ts
```diff
@@ -29,6 +29,7 @@ function gh10345() {
     expectError(leanDoc.id = 43);
 
     const doc2 = await UserModel.findOne().orFail().lean();
+    if (!doc2) return;
```
vkarpov15, line 32: This change is unfortunately a blocker. You should never need `if (!doc2)` if you have `orFail()`.

test/types/lean.test.ts
```diff
@@ -75,6 +76,7 @@ async function gh11118(): Promise<void> {
   const UserModel = model<User>('User', schema);
 
   const docs = await UserModel.find().lean().exec();
+  if (!docs) return;
```
vkarpov15, line 79: This one is also unfortunately a red flag. `find()` can never return `null`, so something is incorrect here.

types/query.d.ts
```diff
@@ -430,7 +430,7 @@ declare module 'mongoose' {
     j(val: boolean | null): this;
 
     /** Sets the lean option. */
-    lean<LeanResultType = ResultType extends any[] ? Require_id<RawDocType>[] : Require_id<RawDocType>>(val?: boolean | any): QueryWithHelpers<LeanResultType, DocType, THelpers, RawDocType>;
+    lean<LeanResultType = ResultType extends any[] ? Require_id<RawDocType>[] : Require_id<RawDocType>>(val?: boolean | any): QueryWithHelpers<LeanResultType | null, DocType, THelpers, RawDocType>;
```
vkarpov15, line 433: Probably need an extra condition here to only add `null` if `ResultType` is nullable. That should work around the `orFail()` issue.

vkarpov15 (changes requested):

types/query.d.ts
```diff
@@ -430,7 +430,7 @@ declare module 'mongoose' {
     j(val: boolean | null): this;
 
     /** Sets the lean option. */
-    lean<LeanResultType = ResultType extends any[] ? Require_id<RawDocType>[] : Require_id<RawDocType>>(val?: boolean | any): QueryWithHelpers<LeanResultType, DocType, THelpers, RawDocType>;
+    lean<LeanResultType = ResultType extends any[] ? Require_id<RawDocType>[] : Require_id<RawDocType>>(val?: boolean | any): QueryWithHelpers<null extends ResultType ? LeanResultType | null : LeanResultType, DocType, THelpers, RawDocType>;
```
hasezoey, line 433: any reason this is written as `null extends Type` instead of `Type extends null` like other parts of the code?

vkarpov15 (approved):
Thanks :+1:

Issue #12691, opened 2022-11-14
Proposal to add some tools for documentation consistency


hasezoey:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

This Issue is a proposal to add the following tools for consistent documentation style (both as raw MD and transformed HTML):
- [`eslint-plugin-markdown`](https://www.npmjs.com/package/eslint-plugin-markdown) to run ESLint against all JS code-blocks inside markdown
- [`markdownlint-cli2`](https://www.npmjs.com/package/markdownlint-cli2) to have consistent markdown files (like eslint but for markdown) using [`markdownlint`](https://www.npmjs.com/package/markdownlint)

Those tools should help to keep the documentation consistent and easy to read.

cc @Uzlopak @vkarpov15

vkarpov15:
I'm supportive of eslint-plugin-markdown. I'm less interested in markdownlint - we'd have to come up with the right set of markdown rules for us, and that seems like it would be a time-consuming endeavor without much upside.

Another tool that I think would be better to have than markdownlint would be a link checker.

hasezoey:
things added so-far:
- `eslint-plugin-markdown` (`lint`): process all code blocks that eslint supports (needs to be manually enabled because of https://github.com/eslint/eslint-plugin-markdown/issues/214)
- `markdownlint-cli2` (`lint-md`): have all markdown files be in consistent style
- `blc` (`docs:check-links`): check documentation for broken links

any other suggestions to maybe add for documentation consistency? otherwise i would close this issue as "completed"

PR #13156, opened 2023-03-10
[DOCS] Add* and run `eslint-plugin-markdown`

hasezoey:
**Summary**

This PR basically adds* and runs `eslint-plugin-markdown`, including all changes required for this to work, in more detail:
- remove `docs` from being ignored in eslint completely
- apply eslint to `test/docs`
- add overrides to use env `browser` for `docs/js/**/*.js`
- apply eslint to `docs/**/*.js` (except `*.min.js` files)
- convert all javascript blocks with acquit-require in it to language `acquit`, to not confuse eslint with invalid tokens
- convert the eslint config from `json` to `js` to allow comments
- fix all documentation problems (like invalid tokens)
- apply eslint to all `*.md` files

*Note: `eslint-plugin-markdown` has been disabled with commit 6f2dec1b2dad72188693f03cd5f7b76a9e22f2e7, because of some out-standing issues: https://github.com/eslint/eslint-plugin-markdown/issues/214

This PR (or the mentioned PRs if this is merged first) may require fixes:
- ~~#13045~~
- #13166
- #13084
- #13113

re #12691


Uzlopak (approved):
awesome

README.md
```diff
@@ -110,9 +110,9 @@ const ObjectId = Schema.ObjectId;
 
 const BlogPost = new Schema({
   author: ObjectId,
-  title: String,
-  body: String,
-  date: Date
+  title:  String,
+  body:   String,
```
vkarpov15, line 114: Can we change this particular setting? This extra indentation is inconsistent with how our docs currently define schemas, and I don't think it is worth switching.

docs/advanced_schemas.md
```diff
@@ -8,6 +8,6 @@ statics, and virtuals from an ES6 class. A class method maps to a schema
 method, a static method maps to a schema static, and getters/setters map
 to virtuals.
 
-```javascript
+```acquit
```
vkarpov15, line 11: Does syntax highlighting still work with this change?

docs/js/native.js
```diff
@@ -1,101 +1,102 @@
 'use strict';
-var _native = {
+const _native = {
```
vkarpov15, line 2: Can we please undo the changes to this file, and ignore this file in eslint? This file was provided by BSA, I don't want to risk changing it.

scripts/website.js
```diff
@@ -152,6 +152,8 @@ async function pugify(filename, options) {
 
   if (options.acquit) {
     contents = transform(contents, tests);
+
+    contents = contents.replaceAll(/^```acquit$/gmi, "```javascript");
```
vkarpov15, line 156: Neat trick

test/docs/discriminators.test.js
```diff
@@ -107,9 +107,9 @@ describe('discriminator docs', function() {
     event.__t = 'SignedUp';
     // ValidationError: ClickedLink validation failed: __t: Cast to String failed for value "SignedUp" (type string) at path "__t"
     // acquit:ignore:start
-    await assert.rejects(async () => {
+    await assert.rejects(async() => {
     // acquit:ignore:end
-    await event.save();
+      await event.save();
```
vkarpov15, line 112: Is there any way we can avoid this one indentation change? The indentation is slightly off here because the `await assert.rejects()` is stripped from the output because of `// acquit:ignore:start`, so we want the docs to appear as if `await event.save()` isn't indented at all.

vkarpov15 (changes requested):

docs/advanced_schemas.md
hasezoey: yes, in `website.js` the codeblock (with acquit) get replaced with javascript ones before markdown processing

this was necessary because eslint would complain about invalid tokens

test/docs/discriminators.test.js
hasezoey: we could disable indentation rule for that line, but would not recommend to to it globally

or the acquit (plugin?) could handle indentation changes between comments

README.md
hasezoey: i had initially added the setting, because i had found some documentation where it was indented to be on the same "column", but also thought that it would be good to keep one rule instead of disabling / enabling for specific sections

but yes, i can remove that rule again and format all documentation with only one space instead of being aligned

docs/js/native.js
hasezoey: sure, though i dont think anything practically changed

it will be ignored like `.min.js` files

hasezoey: reverted & ignored with eb349911cb25d9987ae15c82df3378ea9d9b5752

README.md
hasezoey: rule removed & applied in 88eec11d848363cc0610610dfec9c826c84eca13

hasezoey: i had it originally added, because it was in use at `schematypes.md`, see changes with the rule removed: https://github.com/Automattic/mongoose/pull/13156/files#diff-2f4335ef76e27dfe5e52253d5a42778f66bdf448427004d0b0f4ae927a2b424f

test/docs/discriminators.test.js
hasezoey: for now i could add a quick-fix of ignoring the eslint rule (but editors may still automatically apply indentation)

or we could wait for https://github.com/vkarpov15/acquit-ignore/pull/6

vkarpov15 (approved):

PR #13160, opened 2023-03-13
chore(generateSearch): move config missing error to require

hasezoey:
This PR moves the Error message that was added in 0b7bbbbc23a2b79fb05defaef27f91efc54398bb to the require, where the actual error gets thrown if the config does not exist.

also adds a message on how much content was added.


vkarpov15 (approved):
Nice, thanks :+1:

PR #13163, opened 2023-03-13
fix(model): add `results` property to unordered `insertMany()` to make it easy to identify exactly which documents were inserted

### LINKED ISSUE #12791: Mongoose bulk insert misses critical information

vkarpov15:
Fix #12791

**Summary**

#12791 pointed out that it is still very tricky to figure out exactly which documents were inserted if you're using `insertMany()` with `ordered: false`. That's because the document can either 1) fail Mongoose validation, 2) fail because of a MongoDB server error, 3) succeed. And the results for (1), (2), and (3) are all in separate properties, which means you need to do some stitching to figure out exactly what happened with a particular document.

With this PR, if there were any errors, there will be a `results` property on the error (or `res.mongoose.results` if you're using `rawResult: true`) that contains an array which has either 1) the document if the document was successfully inserted, 2) the MongoDB write error if the write failed, 3) the Mongoose validation error if the document failed validation. The index in the `results` array corresponds to the index of the original document in `arr` in the `Model.insertMany(arr)` call, so this should make it easy to determine if an individual document succeeded or not in `insertMany()`. Just check `error.results[i] instanceof mongoose.Document`.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3371,7 +3371,7 @@ Model.startSession = function() {
  * @param {Array|Object|*} doc(s)
  * @param {Object} [options] see the [mongodb driver options](https://mongodb.github.io/node-mongodb-native/4.9/classes/Collection.html#insertMany)
  * @param {Boolean} [options.ordered=true] if true, will fail fast on the first error encountered. If false, will insert all the documents it can and report errors later. An `insertMany()` with `ordered = false` is called an "unordered" `insertMany()`.
- * @param {Boolean} [options.rawResult=false] if false, the returned promise resolves to the documents that passed mongoose document validation. If `true`, will return the [raw result from the MongoDB driver](https://mongodb.github.io/node-mongodb-native/4.9/interfaces/InsertManyResult.html) with a `mongoose` property that contains `validationErrors` if this is an unordered `insertMany`.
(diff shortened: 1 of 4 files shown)


vkarpov15:
@hasezoey take a look now, I added types and docs, and fixed some test issues

hasezoey (approved):

PR #13169, opened 2023-03-14
test: try to fix "operation interrupted" errors for CI

hasezoey:
**Summary**

This PR tries to add a fix to the "operation interrupted" error in the CI

@vkarpov15 is this what this option will fix / do, or am i understanding the option wrong?

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -153,7 +153,7 @@ async function dropDBs() {
   this.timeout(60000);
 
   const db = await module.exports({ noErrorListener: true }).asPromise();
-  await db.dropDatabase();
+  await db.dropDatabase({ retryWrites: true });
   await db.close();
 }
 
--- a/test/common.js
+++ b/test/common.js
@@ -152,8 +152,26 @@ module.exports.mongodVersion = async function() {
 async function dropDBs() {
   this.timeout(60000);
 
+  // retry the "dropDBs" actions if the error is "operation was interrupted", which can often happen in replset CI tests
+  let retries = 5;
+  while (retries > 0) {
+    retries -= 1;
+    try {
+      await _dropDBs();
+    } catch (err) {
+      if (err instanceof mongoose.mongo.MongoWriteConcernError && /operation was interrupted/.test(err.message)) {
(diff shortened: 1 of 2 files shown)


vkarpov15:
@hasezoey do you have an example of where this operation interrupted error happens?

hasezoey:
example from CI:
https://github.com/Automattic/mongoose/actions/runs/4408293537/jobs/7738190013

the problem is likely that the mongodb instances report that there is a primary available (which is the event mongodb-memory-server waits for) but then shortly after steps down due to whatever reason any interrupting (stopping) all operations that were still running while it was primary.
sadly i dont know of a way to make this more reliable from the MMS side.

i also have another fix stashed, but i would rather not have to use that if this `retryWrites` option works as i think it does

vkarpov15:
Sadly, I don't think this PR will help because [dropDatabase is not a retryable write operation](https://www.mongodb.com/docs/manual/core/retryable-writes/#retryable-write-operations).

I'd say probably the best alternative would be to write a loop in the `dropDBs()` function that manually retries `dropDatabase()` up to twice if there is a `MongoWriteConcernError`. Can you please implement that in this PR @hasezoey ?

hasezoey:
> Sadly, I don't think this PR will help because [dropDatabase is not a retryable write operation](https://www.mongodb.com/docs/manual/core/retryable-writes/#retryable-write-operations).

sad, hoped a simple option would be able to fix this

> I'd say probably the best alternative would be to write a loop in the dropDBs() function that manually retries dropDatabase() up to twice if there is a MongoWriteConcernError. Can you please implement that in this PR @hasezoey ?

already had it stashed, just open a simple option would do just this

---

changed in 04a51bf307277163966dfb0b28b1d73066151d8e, how does it look?

hasezoey:
forgot to update `test/utils`, done with 5d1e95a2adfd0685f22a520dd65974911b2f47b8

maybe we should consider merging those two files / functions (because they do the same)

vkarpov15 (approved):
Thanks :+1:

PR #13178, opened 2023-03-15
fix(document): apply setters on resulting value when calling Document.prototype.$inc()

### LINKED ISSUE #13158: Mongoose schema setter not being called when using $inc
### Prerequisites

### Mongoose version

^6.9.0

### Node.js version

18.12.1

### MongoDB version

5.0.15

### Operating system

Windows

11

### Issue

I run in a problem. I have defined a setter in my model:

```js
    money: {
        balance: {
            type: Number,
            set: (v) => Number(v)?.toFixed(4),
            default: 0,
        },
    },
});
```

When I try increasing balance of a user, it does format the number when the value is positive but sometimes not when its negative, which is weird.

```js
await User.findByIdAndUpdate(userId, {
        $inc: {
            'money.balance': earnings,
        },
    },
    { runValidators: true }
);
```

Thanks for helping me

vkarpov15:
Fix #13158

**Summary**

Document.prototype.$inc() doesn't call setters right now, which is a little strange given that `findOneAndUpdate()` with `$inc` does call setters.

The biggest wrinkle here is that we're calling setters on the _resulting value_, not the value passed in to `$inc`. So `$inc('balance', 4)` will call setters with `balance = 6 + 4` if `balance` was originally 6. I think this is slightly better because setters are often used for rounding, and we want to make sure devs that are using `$inc()` instead of `+=` don't run into unexpected binary floating point precision issues.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1783,18 +1783,22 @@ Document.prototype.$inc = function $inc(path, val) {
     return this;
   }
 
+  const currentValue = this.$__getValue(path) || 0;
+  let valToSet = currentValue + val;
+  let valToInc = val;
+
   try {
     val = castNumber(val);
+    valToSet = schemaType.applySetters(currentValue + val, this);
+    valToInc = valToSet - currentValue;
   } catch (err) {
     this.invalidate(path, new MongooseError.CastError('number', val, path, err));
   }
 
-  const currentValue = this.$__getValue(path) || 0;
-
   this.$__.primitiveAtomics = this.$__.primitiveAtomics || {};
-  this.$__.primitiveAtomics[path] = { $inc: val };
+  this.$__.primitiveAtomics[path] = { $inc: valToInc };
   this.markModified(path);
-  this.$__setValue(path, currentValue + val);
+  this.$__setValue(path, valToSet);
 
   return this;
 };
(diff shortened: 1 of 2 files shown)


lib/document.js
```diff
@@ -1783,18 +1783,22 @@ Document.prototype.$inc = function $inc(path, val) {
     return this;
   }
 
+  const currentValue = this.$__getValue(path) || 0;
+  let valToSet = currentValue + val;
```
hasezoey, line 1787: is this correct to not have `castNumber(val)` too? (like in the try block)

vkarpov15: You're right. My assumption was that setting the invalid value if casting fails, we should still set the value. But that isn't correct. For example, below script still prints 42:

```javascript
'use strict';

async function run() {
  await mongoose.connect('mongodb://127.0.0.1/mongoose_test');
  const Test = mongoose.model('Test', mongoose.Schema({ num: Number }));

  const doc = new Test({ num: 42 });
  doc.num = 'abc';
  console.log(doc.num);
}
```

Thanks for calling that out. Can you re-review please?

test/document.test.js
```diff
@@ -11649,6 +11649,51 @@ describe('document', function() {
...
+        doc.$inc('counter', 1.14159);
+        assert.deepEqual(called, [2, 3.14159]);
+        assert.equal(doc.counter, 3.14);
+        await doc.save();
+
+        const res = await Test2.findById(doc);
+        assert.equal(res.counter, 3.14);
+      });
+
+      it('sets value even if setter fails (gh-13158)', async function() {
```
hasezoey, line 11672: i think this title now needs to be changed, because from my understanding it does not actually change the value anymore on fail

IslandRhythms (approved):
I think hasezoey is right, otherwise LGTM

Issue #9720, opened 2020-12-16
https://github.com/mongodb-js/mongoose-autopopulate/issues/79


vkarpov15:
https://github.com/mongodb-js/mongoose-autopopulate/issues/79

vkarpov15:
WIP in https://github.com/mongodb-js/mongoose-autopopulate/pull/80

PR #13182, opened 2023-03-16
2 failing tests because ?

IslandRhythms:
```
1) insertMany()
       insertMany() validation error with ordered false and rawResult for checking which documents failed (gh-12791):
     TypeError: res.insertedIds[0].toHexString is not a function
      at Context.<anonymous> (test\model.insertMany.test.js:382:37)

  2) insertMany()
       insertMany() populate option (gh-9720):
     TypeError: Cannot read properties of null (reading 'name')
      at Context.<anonymous> (test\model.insertMany.test.js:444:42)
      ```
     I'm genuinely confused why these two tests are failing. The first one, the id is there. The second one, the documents are created but the id is not being set to the ref property even though it is being told to set it.


vkarpov15 (approved):
Thanks :+1:

vkarpov15:
It looks like tests are passing in CI and I don't see why those tests would be failing, so I'm going to merge this one.

PR #13184, opened 2023-03-17
docs(guide+schematypes): add UUID to schematypes guide

### LINKED ISSUE #13032: What is the proper way to replace `_id` type from `ObjectId` to `UUID` in mongoose 6

vkarpov15:
Fix #13032
Re: #13025

**Summary**

Add uuid to schematypes guide so we have some documentation to point to on how to work with uuids.

**Examples**

### DIFF
--- a/docs/schematypes.md
+++ b/docs/schematypes.md
@@ -1,3 +1,4 @@
+
 <h2 id="schematypes"><a href="#schematypes">SchemaTypes</a></h2>
 
 SchemaTypes handle definition of path
@@ -53,6 +54,7 @@ Check out [Mongoose's plugins search](http://plugins.mongoosejs.io) to find plug
 - [Decimal128](api/mongoose.html#mongoose_Mongoose-Decimal128)
 - [Map](#maps)
 - [Schema](#schemas)
+- [UUID](#uuid)
 
 <h4>Example</h4>
 
@@ -506,8 +508,6 @@ const Empty4 = new Schema({ any: [{}] });
 
 <h4 id="maps">Maps</h4>
 
-_New in Mongoose 5.1.0_
-
 A `MongooseMap` is a subclass of [JavaScript's `Map` class](http://thecodebarbarian.com/the-80-20-guide-to-maps-in-javascript.html).
 In these docs, we'll use the terms 'map' and `MongooseMap` interchangeably.
 In Mongoose, maps are how you create a nested document with arbitrary keys.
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
maybe also consider adding a note about `mongoose.Types.UUID` from #13186

also it would be good to note that not just uuid-v4 is supported, but all other versions and that mongoose does not do any checking that they are the same version (but does check that they match a pattern)
https://github.com/Automattic/mongoose/blob/5b75e10815b848133bf7524eb7092bddf8ac7c48/lib/schema/uuid.js#L14

docs/schematypes.md
```diff
@@ -585,13 +585,56 @@ const userSchema = new Schema({
 const User = mongoose.model('User', userSchema);
 ```
 
-To populate every `socialMediaHandles` entry's `oauth` property, you should populate
+To populate every `socialMediaHandles` entry's `oauth` property, you should populatex
```
AbdelrahmanHafez, line 588: Accidentally added letter 😄 
```suggestion
To populate every `socialMediaHandles` entry's `oauth` property, you should populate
```

docs/schematypes.md
```diff
@@ -585,13 +585,56 @@ const userSchema = new Schema({
...
+});
+
+// 'string'
+console.log(typeof doc.docId);
+// true
+console.log(doc.toObject().docId instanceof mongoose.mongo.BSON.Binary);
+
+const rawDoc = await Test.collection.findOne({ _id: doc._id });
+// true
+console.log(rawDoc.toObject().docId instanceof mongoose.mongo.BSON.Binary);
```
AbdelrahmanHafez, line 622: Does that mean `docId` is now _the_ unique identifier? Does it have a unique index by default?
Is `_id` still an object id? Can we add `_id: Schema.Types.UUID`?

Also, I don't know how soon or late it is for this, but maybe it'd be a good idea to avoid naming a type "UUID" without the version, something like `UUIDv4` and `UUIDv1` would be easier to maintain and introduce less questions over time.

docs/schematypes.md
```diff
@@ -585,13 +585,56 @@ const userSchema = new Schema({
...
+```javascript
+const { v4: uuidv4 } = require('uuid');
+
+const schema = new mongoose.Schema({
+  docId: {
+    type: 'UUID',
+    default: () => uuidv4()
+  }
+});
+```
```
AbdelrahmanHafez, line 636: I don't think we should recommend using an external package now that we have a built-in way to achieve the same result. Starting from Node v14.17.0, we now have `const { randomUUID: uuidv4 } = require('crypto');`

docs/schematypes.md
vkarpov15: I suppose I can see that being a source of confusion. `docId` is just an arbitrary property name, no particular significance. Maybe instead I should rename to something like `bookSchema` with an `authorId` property.

I'm not familiar enough with UUID to decide whether having v1 or v4 in the name is better or not. I know the MongoDB Node driver just uses UUID, so I think we should leave the name as UUID for consistency, unless we have a strong reason not to.

hasezoey: > but maybe it'd be a good idea to avoid naming a type "UUID" without the version, something like UUIDv4 and UUIDv1 would be easier to maintain and introduce less questions over time.
> 
>  I know the MongoDB Node driver just uses UUID, so I think we should leave the name as UUID for consistency, unless we have a strong reason not to.

personally i think we should leave it without a version, because mongoose currently does not validate against a specific version and `Schema.Type.UUID` / `mongo.BSON.Binary`(subtype 4) also does not validate against version and allowing all versions, though it seems like `mongo.BSON.UUID` only supports uuidv4 [see this line](https://github.com/mongodb/js-bson/blob/5d2648ecc03e7ac062978172cbf76b97581cc6f8/src/binary.ts#L457)

AbdelrahmanHafez: I was unaware that mongo has a UUID type, so I think going with `UUID` as a name is fine.

I feel like the example can still be confusing. I'd like to propose a full example that can answer a few questions about referencing and uniqueness. Please let me know if the following is valid:

```js
const authorSchema = new Schema({
  _id: Schema.Types.UUID, // Can also do `_id: 'UUID'`
  name: String
});

const Author = mongoose.model('Author', authorSchema);

const bookSchema = new Schema({ 
  authorId: { type: Schema.Types.UUID, ref: 'Author' }
});
const Book = mongoose.model('Book', bookSchema);

const author = new Author({ name: 'Martin Fowler' });
console.log(typeof author._id); // 'string'
console.log(author.toObject()._id instanceof mongoose.mongo.BSON.Binary); // true

const book = new Book({ authorId: '09190f70-3d30-11e5-8814-0f4df9a59c41' });
```

PR #13185, opened 2023-03-17
fix(query): avoid executing transforms if query wasn't executed

### LINKED ISSUE #13165: findById returns "DocumentNotFoundError" instead of "CastError" with invalid ID
### Prerequisites

### Mongoose version

7.0.0

### Node.js version

18.13.0

### MongoDB server version

5.0.15

_No response_

### Description

When using .findById method, an error named "DocumentNotFoundError" is returned when in previous versions an error with name "CastError" would be returned.

### Steps to Reproduce

In the code below, when an invalid ID is passed in req.params.UserId a "DocumentNotFoundError" results. This causes the errorHandler function to return a return a status code of 404, when a status code of 400 would be more appropriate.

```
module.exports.getUserById = (req, res) => {
  User.findById(req.params.UserId)
    .orFail()
    .then((user) => {
      res.status(200).send(user);
    })
    .catch((err) => {
        errorHandler.errorHandler(req, res, err);
    });
};
```
### Expected Behavior

I expected an error name of "Cast Error", but instead received an error with a name "DocumentNotFoundError".

vkarpov15:
Fix #13165

**Summary**

Some of the refactoring we did to switch `exec()` to async/await created a quirk where we still execute query transforms even if the query errored out. This leads to odd consequences like throwing a `DocumentNotFoundError` if `findOne()` throws a cast error, even though Mongoose didn't execute a query.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2786,7 +2786,7 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
 
   // only validate required fields when necessary
   const pathDetails = _getPathsToValidate(this, pathsToValidate, pathsToSkip);
-  let paths = shouldValidateModifiedOnly ?
+  const paths = shouldValidateModifiedOnly ?
     pathDetails[0].filter((path) => this.$isModified(path)) :
     pathDetails[0];
   const skipSchemaValidators = pathDetails[1];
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

Issue #13170, opened 2023-03-15
updated timestamp not get updated for noop, aside of upsert usage


cherviakovtaskworld:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.0.1

### Node.js version

19.6.0

### MongoDB server version

5.1.0

### Description

We requrie to update documents the way that only updated timestamp get changed, trying to perform update operation with empty object, which is noop. Mongoose seem to catching it and does not perform any operation. If we add upsert: true option though and document exists, mongoose does perform update even if it is noop. Well we can't use upsert as it may lead to many partial documents get inserted, but in the same time seems without upsert and without any actual update to the document we can't get updated timestmap to be updated.

Checked that in pure mongodb nodejs driver when we on update adding $set it is working as expected and update timestamp field every time
```
await db.collection('users').updateOne({ name: 'test' }, { $setOnInsert: { created: new Date }, $set: { updated: new Date } })
```
therefore it is issue on mongoose side.

### Steps to Reproduce

```
import mongoose from 'mongoose'

await mongoose.connect('mongodb://127.0.0.1:27017/test')

const UsersSchema = new mongoose.Schema({
  name: String,
  // updated: { type: Date, default: Date.now() },
  // created: { type: Date, default: Date.now() }
}, {
  timestamps: {
    createdAt: 'created',
    updatedAt: 'updated'
  }
})

const UsersModel = mongoose.model('users', UsersSchema)

await UsersModel.deleteMany({})
await UsersModel.create({ name: 'test' })
const user1 = await UsersModel.findOne({ name: 'test' }).lean()
console.log('user1', user1)
await new Promise((resolve) => setTimeout(() => resolve(), 1000))
const user2 = await UsersModel.findOneAndUpdate({ name: 'test' }, {}, { new: true })
// await UsersModel.updateOne({ name: 'test' }, { name: 'test1' })
// const user2 = await UsersModel.findOne({ name: 'test1' }).lean()
console.log('user2', user2)
```

### Expected Behavior

updated timestamp get updated even on noop update operation, or there separate operations to only update timestmpas (e.g. bash touch) or there option to enforce update without creation if document not exists.

vkarpov15:
As a workaround, you can explicitly set `updated` in your update as follows:

```javascript
const user2 = await UsersModel.findOneAndUpdate({ name: 'test' }, { updated: Date.now() }, { new: true })
```

[...]

Issue #9951, opened 2021-02-18
findOneAndReplace does not set timestamps


aaronadamsCA:
**Do you want to request a *feature* or report a *bug*?**

Report a bug.

**What is the current behavior?**

I am calling `Model.findOneAndReplace()` on a model with timestamps enabled. The resulting documents have no timestamps. Adding `timestamps: true` has no effect.

Calling `Model.findOneAndUpdate()` on the same model with `overwrite: true` results in correct timestamps, leading me to believe this issue is specific to `Model.findOneAndReplace()`.

**If the current behavior is a bug, please provide the steps to reproduce.**

```js
const product = await Product.findOneAndReplace(
  { externalId },
  { externalId, ...rest },
  { upsert: true, setDefaultsOnInsert: true, new: true }
).exec();
```

I can't currently provide a standalone PR but I still wanted to get this report in, in case it's something obvious.

**What is the expected behavior?**

Insert or overwrite with timestamps, as this similar call does:

```js
const product = await Product.findOneAndUpdate(
  { externalId },
  { externalId, ...rest },
  { overwrite: true, upsert: true, setDefaultsOnInsert: true, new: true }
);
```

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Node.js 14.15.4, Mongoose 5.11.17, MongoDB 3.6.4 (MongoDB Atlas 4.4.3).

IslandRhythms:
```
const mongoose = require('mongoose');

mongoose.connect("mongodb://localhost:27017/", { useNewUrlParser: true, useUnifiedTopology: true });
const data = mongoose.connection;
data.on("open", function () {
    data.db.dropDatabase(console.log('database drop'));
});

const userSchema = new mongoose.Schema({
    email: {
      type: String,
      unique: true // `email` must be unique
    }
  },{timestamps: true});
  const User = mongoose.model('User', userSchema);
  async function test() {
    console.log(await User.create({email: 'test@google.com'}));
    console.log(await User.findOneAndReplace({email:'test@google.com'},{email: 'test2@google.com'}, {new:true, overwrite: true, upsert: true, rawResult:true}));
  }

  test();
```

PR #13196, opened 2023-03-22
fix(timestamps): set timestamps on empty `replaceOne()`

vkarpov15:
Re: #9951
Re: #13170

**Summary**

I noticed when working on #13170 that the fix for #9951 only applies to `findOneAndReplace()`, not `replaceOne()`.

**Examples**


hasezoey (approved):
LGTM

Issue #8961, opened 2020-05-09
Remove `useNestedStrict`, always use child schema's `strict` mode


vkarpov15:
Re: #8922

AbdelrahmanHafez:
Another option would be to set it as `true` by default, I've never needed to use that option, but some users might find the need to use that option.

vkarpov15:
If you want `useNestedStrict: false`, you should be able to force `strict` mode for all child schemas using the `strict` option: `Model.updateOne(filter, update, { strict: Model.schema.options.strict })`. I should double check that you can do that when working on this.

AbdelrahmanHafez:
That seems like a reasonable replacement.

AdamGerthel:
Just found this thread because I discovered the hard way that strict mode wasn't being taken into account for subdocuments when using `Model.update()`. Is `useNestedStrict` being deprecated?

AbdelrahmanHafez:
Yes, it should be in 6.0.

PR #13199, opened 2023-03-22
Added logic back to pull in strict option from schema

jerell-mendoza:
**Summary**

I wanted to add this logic back into the Query from 5.x
https://github.com/Automattic/mongoose/blob/5.x/lib/query.js#L4686

I am not sure why this was removed in 6.x and 7.x but I would like the `strict` parameter to persist from the Schema.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -4575,6 +4575,15 @@ Query.prototype._castUpdate = function _castUpdate(obj, overwrite) {
     upsert = this.options.upsert;
   }
 
+  let strict;
+  if ('strict' in this.options) {
+    strict = this.options.strict;
+  } else if (schema && schema.options) {
+    strict = schema.options.strict;
+  } else {
+    strict = true;
+  }
+
   const filter = this._conditions;
   if (schema != null &&
       utils.hasUserDefinedProperty(filter, schema.options.discriminatorKey) &&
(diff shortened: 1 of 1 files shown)


vkarpov15 (commented):
Thanks for catching this. Could you please add a test case to `Model.findOneAndUpdate.test.js` for this?

jerell-mendoza:
@vkarpov15 Yes! I'd love to add a test. Will ping you when that's ready.

lib/query.js
```diff
@@ -4575,6 +4575,15 @@ Query.prototype._castUpdate = function _castUpdate(obj, overwrite) {
     upsert = this.options.upsert;
   }
 
+  let strict;
+  if ('strict' in this._mongooseOptions) {
+    strict = this._mongooseOptions.strict;
+  } else if (schema && schema.options) {
```
vkarpov15, line 4581: It looks like this logic breaks our test for #8961. I don't think this PR as written will work, because this code only runs on the top-level schema. So this will always force nested schemas to have the same `strict` mode as the top level schema.

I think the fix for this issue is going to end up in `lib/helpers/query/castUpdate.js` similar to the fix for #8961: https://github.com/Automattic/mongoose/commit/dd122fe9b17641a89244d93330a9dc460911c34b

vkarpov15 (changes requested):

lib/query.js
jerell-mendoza: Got it, I think ok I understand now.

We were using Mongoose 5.x and we relied on the "feature/bug" that the top level schema settings would apply to the sub document / nested schemas since it only read the top-level schema.

Now that this is working properly, we just need to explicitly set `{ strict: false }` on all our nested schemas.

I will close out this PR thank you for clarifying!

PR #13202, opened 2023-03-22
fix(array): pass-through all parameters

### LINKED ISSUE #13201: Nested Discriminators in multi-dimensonal arrays use `name` instead of `value`
### Prerequisites

### Mongoose version

6.10.1

### Node.js version

19.6.0

### MongoDB server version

5.0

4.9.5

### Description

I have noticed that nested discriminator multi-dimensonal arrays use the `name` of the discriminator instead of `value`, like in non-array or one-dimensonal arrays

function parameter reference: `path.discriminator(name, schema, value)`

### Steps to Reproduce

Reproduction Repository & branch: https://github.com/typegoose/typegoose-testing/tree/mongooseNestedDiscriminatorNaming

- (have node_modules install and a mongo instance running on default port)
- execute `yarn run run` (or any package manager)
- observe output being without the discriminated properties and assert error

- invert https://github.com/typegoose/typegoose-testing/blob/39da16a7de48d26f9d37b1ead010c0d7f48ce6ec/src/test.ts#L83-L88
- execute `yarn run run` (or any package manager)
- observe discriminated properties being present and no assertion error

also includes commented-out code for a one-dimensonal array

### Expected Behavior

to use `value` like for one-dimensonal arrays and non-arrays

---

also tested mongoose 7.0.2, with same behavior

hasezoey:
**Summary**

This PR changes `SchemaArray.discriminator` to pass-through all parameters to the `arr.discriminator` function

fixes #13201

fix should also be added to 7.x

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -451,7 +451,7 @@ SchemaArray.prototype.$toObject = SchemaArray.prototype.toObject;
  * ignore
  */
 
-SchemaArray.prototype.discriminator = function(name, schema) {
+SchemaArray.prototype.discriminator = function(...args) {
   let arr = this;
   while (arr.$isMongooseArray && !arr.$isMongooseDocumentArray) {
     arr = arr.casterConstructor;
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

PR #13208, opened 2023-03-24
types(model): aligned `watch()` type for mongodb 4.6.0

### LINKED ISSUE #13206: [TypeScript] watch().on doesn't have correct type on callback, so can't access fullDocument
### Prerequisites

### Mongoose version

7.0.2 (but happening since 6.4.0, with the switch to mongodb driver 4.7.0)

### Node.js version

19.8.1

### MongoDB server version

6.0.4

5.0.2

### Description

`watch().on` is returning the generic `ChangeStreamDocument<T>` as argument for the `on` callback, which doesn't include `fullDocument` and other properties from other types of specialized `ChangeStreamDocument` types like `ChangeStreamInsertDocument` or `ChangeStreamUpdateDocument` that were introduced in mongodb driver 4.6.0.

I suppose it's because the current type for `watch()` accepts only 1 argument, but since mongodb 4.6.0, watch type requires two arguments to be able to provide the specialized return types:

https://github.com/mongodb/node-mongodb-native/releases/tag/v4.6.0

See current docs https://mongodb.github.io/node-mongodb-native/5.1/classes/Collection.html#watch

### Steps to Reproduce

```
model.watch<Schema>([], { fullDocument: "updateLookup" }).on("change", (change) => {
  console.log(change.fullDocument);
});

```

Will give a TypeScript error: `Property 'fullDocument' does not exist on type 'ChangeStreamDocument<Schema>'.`

### Expected Behavior

Being able to call `watch` with specialized type, like:
```
model
  .watch<Schema, ChangeStreamInsertDocument<Schema> | ChangeStreamUpdateDocument<Schema>>([], {
    fullDocument: "updateLookup",
  })
  .on("change", (change) => {
    console.log(change.fullDocument);
  });

```

So the `change` argument of the `on` callback has the correct type, to for example be able to access `fullDocument`.

### Workaround

As a workaround, I'm setting the type of the `on` parameter explicitly to the specialized types I need:

```
model
  .watch<Schema>([], { fullDocument: "updateLookup" })
  .on("change", (change: ChangeStreamInsertDocument<Schema> | ChangeStreamUpdateDocument<Schema>) => {
    console.log(change.fullDocument);
  });
```

But I think it makes more sense to align the current mongoose's `watch` to how it is in the current mongodb driver.

lpizzinidev:
**Summary**
You can now specify the custom type for the top-level document returned in a `change` event.
This aligns with the changes introduced in [MongoDB 4.6.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v4.6.0).

**Example**
```
TestModel.watch<ITest, ChangeStreamInsertDocument<ITest>>([]).on('change', (change) => {
    console.log(change.fullDocument); // TS won't complain
});
```

Closes #13206

### DIFF
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -382,7 +382,7 @@ declare module 'mongoose' {
     validate(optional: any, pathsToValidate: PathsToValidate): Promise<void>;
 
     /** Watches the underlying collection for changes using [MongoDB change streams](https://www.mongodb.com/docs/manual/changeStreams/). */
-    watch<ResultType extends mongodb.Document = any>(pipeline?: Array<Record<string, unknown>>, options?: mongodb.ChangeStreamOptions & { hydrate?: boolean }): mongodb.ChangeStream<ResultType>;
+    watch<ResultType extends mongodb.Document = any, ChangeType extends mongodb.ChangeStreamDocument = any>(pipeline?: Array<Record<string, unknown>>, options?: mongodb.ChangeStreamOptions & { hydrate?: boolean }): mongodb.ChangeStream<ResultType, ChangeType>;
 
     /** Adds a `$where` clause to this query */
     $where(argument: string | Function): QueryWithHelpers<Array<THydratedDocumentType>, THydratedDocumentType, TQueryHelpers, TRawDocType>;
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -17,7 +17,7 @@ import {
 } from 'mongoose';
 import { expectAssignable, expectError, expectType } from 'tsd';
 import { AutoTypedSchemaType, autoTypedSchema } from './schema.test';
-import { UpdateOneModel } from 'mongodb';
+import { UpdateOneModel, ChangeStreamInsertDocument } from 'mongodb';
 
 function rawDocSyntax(): void {
   interface ITest {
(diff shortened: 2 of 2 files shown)


types/models.d.ts
```diff
@@ -382,7 +382,7 @@ declare module 'mongoose' {
     validate(optional: any, pathsToValidate: PathsToValidate): Promise<void>;
 
     /** Watches the underlying collection for changes using [MongoDB change streams](https://www.mongodb.com/docs/manual/changeStreams/). */
-    watch<ResultType extends mongodb.Document = any>(pipeline?: Array<Record<string, unknown>>, options?: mongodb.ChangeStreamOptions & { hydrate?: boolean }): mongodb.ChangeStream<ResultType>;
+    watch<ResultType extends mongodb.Document = any, ChangeType extends mongodb.ChangeStreamDocument = any>(pipeline?: Array<Record<string, unknown>>, options?: mongodb.ChangeStreamOptions & { hydrate?: boolean }): mongodb.ChangeStream<ResultType, ChangeType>;
```
vkarpov15, line 385: Wouldn't it be better if we added a separate overload for `watch()` with `options: mongodb.ChangeStreamOptions & { fullDocument: 'updateLookup', hydrate?: boolean }` that added `fullDocument` to the result? It looks like this approach relies on the user to specify the correct type

lpizzinidev: From my understanding of the [changelog](https://github.com/mongodb/node-mongodb-native/releases/tag/v4.6.0), the added feature is that the dev can now specify `ChangeType`, the type of the change document.
It is not related to the `options` parameter (I wrote an unclear example above, now edited).

vkarpov15: I took a closer look and @lpizzinidev is correct, the expected behavior here is that the user can overwrite the valid change types. `fullDocument: 'updateLookup'` doesn't always ensure that there will be a `fullDocument` property, for example deletes. Below is an example script I used to verify:

```ts

mongoose.set('debug', true);

run().catch(err => console.error(err));

async function run() {
  await mongoose.connect('mongodb://localhost:27017,localhost:27018,localhost:27019/mongoose_test');

  const TestModel = mongoose.model('Test', new mongoose.Schema({ name: String }));
  await TestModel.createCollection();

  await TestModel.
    watch<{name:string}>([], { fullDocument: 'updateLookup' }).
    on('change', change => console.log(change));
  await TestModel.create({ name: 'foo' });
  await TestModel.updateOne({ name: 'foo' }, { name: 'bar' });
  await TestModel.deleteOne({ name: 'bar' });
  console.log('Wrote document');
}
```

vkarpov15 (approved):

PR #13213, opened 2023-03-24
test(model): make test covering `init()` errors on index build failures more robust

### LINKED ISSUE #13171: [Test] `indexes -> error should emit on the model` seems to not always emit a error
### Prerequisites

### Issue

The Test `model -> indexes -> error should emit on the model` in `model.indexes.test.js` seems to not always give a error.

test:
https://github.com/Automattic/mongoose/blob/339bf686df027f5e07bd7aaf888fce6f8d6c4dfa/test/model.indexes.test.js#L257-L272

Error:
```txt
  1) model
       indexes
         error should emit on the model:
     TypeError: Cannot read properties of null (reading 'message')
      at Context.<anonymous> (test/model.indexes.test.js:268:55)
```

from: https://github.com/hasezoey/mongoose/actions/runs/4425870265/jobs/7761457802

vkarpov15:
Fix #13171

**Summary**

I don't have a good explanation for the root cause of this test flaking. But we shouldn't rely on modifying the Test model's schema indexes after compiling, that seems brittle. With this change, we instead create a separate model that points to the same collection, and ensure that model's index builds fail.

**Examples**

### DIFF
--- a/test/model.indexes.test.js
+++ b/test/model.indexes.test.js
@@ -256,19 +256,23 @@ describe('model', function() {
 
     it('error should emit on the model', async function() {
       const schema = new Schema({ name: { type: String } });
-      const Test = db.model('Test', schema);
-
+      const Test = db.model('Test', schema, 'Test');
+      await Test.init();
       await Test.create({ name: 'hi' }, { name: 'hi' });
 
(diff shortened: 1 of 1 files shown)


hasezoey (approved):
LGTM, hopefully this fixes it

PR #13218, opened 2023-03-27
fix(model): execute valid write operations if calling `bulkWrite()` with ordered: false

### LINKED ISSUE #13176: Unordered `bulkWrite()` with validation errors doesn't execute the bulk write
### Prerequisites

### Mongoose version

6.x

### Node.js version

Doesn't matter

### MongoDB server version

Doesn't matter

_No response_

### Description

General idea is the following code throws a validation error and doesn't execute the one valid bulk write op, even though `ordered: false` is set

```javascript
      num: Number
    });
    const User = db.model('User', userSchema);


    await User.bulkWrite([{
      updateOne: {
        filter: { },
        update: { num: 'not a number' },
        upsert: true
      },
      {
        updateOne: { filter: {}, update: { num: 42 } }
      }
    }], { ordered: false }).then(() => null, err => err);
```

Reported in Slack: https://mongoosejsteam.slack.com/archives/C2CA1AWQP/p1678894562556919

"Hey I’m trying to use writeBulk with multiple updateOne unordered. When one entity throws error (For example doesn’t pass scheme validation) all the bulk get reverts, and I get a general error of cast fails. I would like to update the entities that able to be updated, and get error that contains the entities failed to update, similar to the the insertMany behavior. Does anyone have an idea?"

### Steps to Reproduce

See above

### Expected Behavior

_No response_

vkarpov15:
Fix #13176

**Summary**

Make the `bulkWrite()` API more consistent with `insertMany()`:

If `ordered: false`, validate all write operations, and strip out any invalid ones. Execute the valid operations.

1. If remaining operations succeed, add `mongoose.validationErrors` to result to indicate which operations failed validation.
2. If remaining operations failed, add `mongoose.validationErrors` to the error to indicate which operations failed validation.

I'm not thrilled with the fact that `bulkWrite()` doesn't throw an error if some operations failed validation, but that is consistent with how `insertMany()` works. And I'd prefer to maintain consistency, and consider changing the API for the next major release. What do you think @hasezoey ?

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3790,8 +3790,7 @@ Model.bulkWrite = function(ops, options, callback) {
       map(v => v.error);
 
     function completeUnorderedValidation() {
-      validOps.sort();
-      validOps = validOps.map(index => ops[index]);
+      validOps = validOps.sort().map(index => ops[index]);
 
       this.$__collection.bulkWrite(validOps, options, (error, res) => {
         if (error) {
--- a/lib/model.js
+++ b/lib/model.js
@@ -3736,33 +3736,81 @@ Model.bulkWrite = function(ops, options, callback) {
     options = null;
   }
   options = options || {};
+  const ordered = options.ordered == null ? true : options.ordered;
 
   const validations = ops.map(op => castBulkWrite(this, op, options));
 
   callback = this.$handleCallbackError(callback);
   return this.db.base._promiseOrCallback(callback, cb => {
     cb = this.$wrapCallback(cb);
-    each(validations, (fn, cb) => fn(cb), error => {
-      if (error) {
-        return cb(error);
-      }
+    if (ordered) {
+      each(validations, (fn, cb) => fn(cb), error => {
+        if (error) {
+          return cb(error);
+        }
 
-      if (ops.length === 0) {
-        return cb(null, getDefaultBulkwriteResult());
-      }
+        if (ops.length === 0) {
+          return cb(null, getDefaultBulkwriteResult());
+        }
 
-      try {
-        this.$__collection.bulkWrite(ops, options, (error, res) => {
-          if (error) {
-            return cb(error);
+        try {
+          this.$__collection.bulkWrite(ops, options, (error, res) => {
+            if (error) {
+              return cb(error);
+            }
+
+            cb(null, res);
+          });
+        } catch (err) {
+          return cb(err);
+        }
+      });
+
+      return;
+    }
+
+    let remaining = validations.length;
+    let validOps = [];
+    let validationErrors = [];
+    for (let i = 0; i < validations.length; ++i) {
+      validations[i]((err) => {
+        if (err == null) {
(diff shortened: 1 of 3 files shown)


lib/model.js
```diff
@@ -3736,33 +3736,81 @@ Model.bulkWrite = function(ops, options, callback) {
...
+    for (let i = 0; i < validations.length; ++i) {
+      validations[i]((err) => {
+        if (err == null) {
+          validOps.push(i);
+        } else {
+          validationErrors.push({ index: i, error: err });
+        }
+        if (--remaining <= 0) {
+          completeUnorderedValidation.call(this);
+        }
```
hasezoey, line 3784: why have this inside the loop instead of after the loop and checking if there are operations to be done?

lib/model.js
```diff
@@ -3736,33 +3736,81 @@ Model.bulkWrite = function(ops, options, callback) {
...
+      });
+    }
+
+    validationErrors = validationErrors.
+      sort((v1, v2) => v1.index - v2.index).
+      map(v => v.error);
+
+    function completeUnorderedValidation() {
+      validOps.sort();
+      validOps = validOps.map(index => ops[index]);
```
hasezoey, line 3794: for consistency, why not combine it into one statement (like the above `validationErrors`)?

hasezoey (commented):
i am not very familiar with this part of the code, so i cannot say much about it

> I'm not thrilled with the fact that bulkWrite() doesn't throw an error if some operations failed validation, but that is consistent with how insertMany() works

i also think if there are errors it should throw by default, but maybe also allow a option like `applyValidOperations` or something (which is by default `false` in a next major release)

lib/model.js
vkarpov15: It is inside the for loop, but only because this logic needs to be inside the callback function to validations[i].

vkarpov15:
@hasezoey it is a reasonable idea to throw an error if errors occurred. However, I'd like to add that in a separate PR because I would like to add that option to `insertMany()` as well as `bulkWrite()`.

Probably call the option `throwOnValidationError` or something like that, because we would apply valid operations either way. The difference would be that, if `throwOnValidationError` is true and `ordered` is false, then `bulkWrite()` and `insertMany()` would throw an error if there were operations that failed validation, but all operations that passed validation succeeded.

IslandRhythms (approved):
LGTM

lib/model.js
```diff
@@ -3736,33 +3736,80 @@ Model.bulkWrite = function(ops, options, callback) {
...
+        if (error) {
+          return cb(error);
+        }
 
-      if (ops.length === 0) {
-        return cb(null, getDefaultBulkwriteResult());
-      }
+        if (ops.length === 0) {
+          return cb(null, getDefaultBulkwriteResult());
+        }
```
doomhz, line 3754: This check should be moved somewhere above the `if (ordered) {` block

PR #13222, opened 2023-03-29
Add `docs/source/api` to `docs/source/index` filemap AND add better versioning support

### LINKED ISSUE #13148: [Docs] Rework scripts to allow setting easier setting of current version
### Prerequisites

### Issue

Currently when a new major version is out, fixes like https://github.com/Automattic/mongoose/pull/13147 have to be used to make all paths correct on the gh-pages branch, but as a downside this makes it unable to be viewed locally without reverting the fix, this is a tracker issue (note to self) to improve the script to allow a simple setting of a variable.

Also consider slightly rework on how the "latest versions" are fetched so that it still works for selecting newer versions from branches like 6.x.

hasezoey:
**Summary**

This PR consists of 2 parts:

part 1:
- add `dos/source/api` to the `docs/source/index` filemap
- move code from `docs/source/splitApiDocs.js` to `/scripts/website.js`
- add jsdoc types for `docs/source/api` `out` (/ `module.exports.docs` / `data`)
- rename api property `name` to `title` to match `docs/source/index` scheme
- only promisify `pug.render` once
- remove `docs/source/splitApiDocs.js`
- add watchers for all of documented `lib/*` files
- add watcher for `docs/api_split.pug` (reload all api files)
- update `scripts/generateSearch` for the changes made
- update `scripts/generateSearch` to use the already generated url instead of re-doing it
- add a log of how many files will be processed when running `docs:generate`

part 2:
- add better versioning support by making the process more automatic
- refactor how versions are parsed and from where (previously it was from CHANGELOG, but now its from `git tag`)
- add ability to deploy to sub-directy (like in gh-pages) directly like `docs/6.x`
- add dev-dependency `fs-extra` to copy the full directories (not using `ncp` because it is old and nodejs does not have a native way, would replace `ncp` later)

fixes #13148

---

Notes on the new versioning:
the new version in made so that without environment variable `DOCS_DEPLOY=true` it will behave as if no versioning is present (not outputting any new sub-directories or adding things to urls), but once the environment variable is set, it will:
- if the latest version is detected, or as fallback; it will deploy to `docs/` like always
- if a older version is detected, it will deploy to something like `docs/6.x`, including all files that are required (like js, css, and even `<root>/index.pug`)

this change makes it so that versioned docs can still be locally run without screwing up static paths (like js and css) but also easy to deploy to gh-pages (hopefully)

should i add a package.json script for deploying with `DOCS_DEPLOY=true`?


vkarpov15 (approved):
This looks great :+1: I was just having some trouble testing docs generation on 6.x branch, looks like this might help with that.

hasezoey:
@vkarpov15 would you want a backport of this PR to the 6.x branch?

vkarpov15:
Yeah backport would be nice. Merge conflicts will be a bit messy, but hopefully not too bad.

PR #13226, opened 2023-04-01
[Docs] Fix `async function` anchors

hasezoey:
**Summary**

This PR fixes `dox` not being able to recognize `async function` as proper function, resulting in anchors of `Model.ensureIndexes` instead of `Model.ensureIndexes()`

also includes a fix to automatically convert the old improper anchors to the new ones (if existing)

See https://github.com/tj/dox/issues/198


vkarpov15 (approved):
LGTM. Maybe time to re-open #8925 ?

hasezoey:
> LGTM. Maybe time to re-open #8925 ?

maybe, but i dont know with what to replace it with aside from `typedoc`, for now `dox` works and can be worked-around

PR #13247, opened 2023-04-04
fix(model): avoid unnecessary empty options to `createCollection()` with `autoCreate`

vkarpov15:
**Summary**

Tiny fix but still nice to have so we're not sending extra data unnecessarily.

**Examples**


AbdelrahmanHafez (approved):
LGTM, thanks! 👍

hasezoey (approved):

PR #13249, opened 2023-04-05
clearer error message

IslandRhythms:
**Summary**

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -4922,15 +4922,14 @@ function _orFailError(err, query) {
 
 Query.prototype.exec = function exec(op, callback) {
   const _this = this;
-  // Ensure that `exec()` is the first thing that shows up in
-  // the stack when cast errors happen.
-  const execOptions = this._optionsForExec(this.model);
-  if (execOptions.sort) {
-    const keys = Object.keys(execOptions.sort);
+  if (this.options && this.options.sort) {
+    const keys = Object.keys(this.options.sort);
     if (keys.includes('')) {
-      throw new Error('Invalid arguent passed to sort()');
+      throw new Error('Invalid field "" passed to sort()');
     }
   }
+  // Ensure that `exec()` is the first thing that shows up in
+  // the stack when cast errors happen.
   const castError = new CastError();
 
   if (typeof op === 'function') {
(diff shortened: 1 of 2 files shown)


lib/query.js
```diff
@@ -4924,6 +4924,13 @@ Query.prototype.exec = function exec(op, callback) {
   const _this = this;
   // Ensure that `exec()` is the first thing that shows up in
   // the stack when cast errors happen.
+  const execOptions = this._optionsForExec(this.model);
+  if (execOptions.sort) {
+    const keys = Object.keys(execOptions.sort);
+    if (keys.includes('')) {
+      throw new Error('Invalid arguent passed to sort()');
```
vkarpov15, line 4931: "arguent"?

lib/query.js
```diff
@@ -4924,6 +4924,13 @@ Query.prototype.exec = function exec(op, callback) {
   const _this = this;
   // Ensure that `exec()` is the first thing that shows up in
   // the stack when cast errors happen.
+  const execOptions = this._optionsForExec(this.model);
```
vkarpov15, line 4927: Why put this underneath the above comment? Above comment is not relevant to this code block.

lib/query.js
```diff
@@ -4924,6 +4924,13 @@ Query.prototype.exec = function exec(op, callback) {
...
+  if (execOptions.sort) {
```
vkarpov15, line 4928: Let's just use `this.options.sort`. `_optionsForExec()` clones the object, so this would incur a performance penalty.

test/query.test.js
```diff
@@ -4373,4 +4373,14 @@ describe('Query', function() {
...
   });
+  it('should provide a clearer error message when sorting with empty string', async function() {
+    const testSchema = new Schema({
+      name: { type: String }
+    });
+
+    const Error = db.model('error', testSchema);
+    await assert.rejects(async() => {
+      await Error.find().sort('-');
+    }, { message: 'Invalid arguent passed to sort()' });
```
vkarpov15, line 4384: On  second thought, let's make this error message `Invalid field "" passed to sort()`

vkarpov15 (changes requested):

vkarpov15 (approved):

PR #13254, opened 2023-04-06
Speed up get on top-level properties by 10x

vkarpov15:
**Summary**

`doc.myProp` does a lot of unnecessary string manipulation and comparisons. This PR trims off a bunch of the unnecessary work.

Before:

```
$ node ./benchmarks/get.js 
{
  "Model loop ms": 2011,
  "POJO loop ms": 7
}
```

After:

```
$ node ./benchmarks/get.js 
{
  "Model loop ms": 206,
  "POJO loop ms": 8
}
```

Not really sure how much faster we can make this, Mongoose will always be slower than a plain old property access because we need to support getters, etc. 

**Examples**

### DIFF
--- a/benchmarks/get.js
+++ b/benchmarks/get.js
@@ -1,40 +1,72 @@
 'use strict';
 
-const get = require('../lib/helpers/get');
+const mongoose = require('../');
 
-let obj;
+run().catch(err => {
+  console.error(err);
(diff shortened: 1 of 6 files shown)


AbdelrahmanHafez (approved):
This is great, thanks!

PR #13257, opened 2023-04-06
chore(workflows/test): update deno version to 1.32

hasezoey:
**Summary**

This PR updates the used deno version to 1.32, because 1.32.2 got a port of `node:zlib` implemented in rust (rather than a polyfill?), because it often resulted in `Error during unzip for /binary/mongodb-linux-x86_64-ubuntu2004-6.0.0.tgz: Error: incorrect header check`

hopefully this fixes this problem for deno

[Example run](https://github.com/hasezoey/mongoose/actions/runs/4631765108/jobs/8195024293)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,7 +96,7 @@ jobs:
       - name: Setup Deno
         uses: denoland/setup-deno@v1
         with:
-          deno-version: v1.30.x
+          deno-version: v1.32.x
       - run: deno --version
       - run: npm install
       - name: Run Deno tests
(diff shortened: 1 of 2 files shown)


vkarpov15:
Might just need to upgrade std?

hasezoey:
> Might just need to upgrade std?

already tried locally, but somehow it keeps failing with the same error (no base url)

hasezoey:
well turns out that i was dumb:
- i didnt notice that `test/deno.js` was explicitly defining the std version for `createRequire` (but not for flags)
- deno 1.31 and std 0.177 removed node from std and moved it to be `node:` imported
- deno somehow changed that `process` needed to be explicitly imported (for `process.stdout` modification)

AbdelrahmanHafez (approved):

vkarpov15 (approved):
Nice :+1:

PR #13262, opened 2023-04-09
Modify exemple and schema syntax

c-marc:
**Summary**

- it seems `{ name: "string", size: "string" }` is a valid syntax, but I had to run some code to confirm it is :)
- in my opinion `{ name: String, size: String }` would be more coherent with the rest of the doc (especially for new learners who might find this snippet confusing)

**Examples**


AbdelrahmanHafez (approved):
LGTM, thanks. 👍

PR #13277, opened 2023-04-14
chore: add `versionNumber` property to docs content for search

vkarpov15:
**Summary**

Re: [this comment](https://github.com/mongoosejs/backend/commit/378f7e27a2ae2379260047febddd699fe479c67a#r108335721), looks like our search backend isn't using Atlas search correctly. Unfortunately with the current `version` property being a string, we get the following error: `MongoServerError: "compound.must[0].equals.value" must be a boolean, objectId, number, or date` . Looks like there's no way for `$search` with `equals` to compare strings. So I added a `versionNumber` property.

We'll get rid of the `Content` model/s `version` property once we've merged the search changes and confirmed no issues.

**Examples**

### DIFF
--- a/scripts/generateSearch.js
+++ b/scripts/generateSearch.js
@@ -31,7 +31,8 @@ const contentSchema = new mongoose.Schema({
   title: { type: String, required: true },
   body: { type: String, required: true },
   url: { type: String, required: true },
-  version: { type: String, required: true, default: version }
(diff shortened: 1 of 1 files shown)


hasezoey (commented):
code looks good to me, but i cannot say about what search to use, because i have never used the search (or atlas) before

IslandRhythms (approved):
LGTM

Issue #13271, opened 2023-04-13
$__reset method is slow on document with large number of subdocuments


alexd6631:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the performance issue has not already been reported

### Last performant version

6.9.2

### Slowed down in version

6.9.2

### Node.js version

18

### 🦥 Performance issue

Context : 
I had an issue at work when saving a document with a large number of subdocuments.
The mongoose model (written with Typegoose) use a Map to store unique pair of keys and sub-documents, in a single mongo document.

When adding a set of keys, to a document having already thousand of keys, I noticed the .save() was quite long, at first I suspected that the change detection or write to the db itself was the most expensive. But a profiling session showed that the bottleneck was on __reset method calling isModified.

<img width="848" alt="image" src="https://user-images.githubusercontent.com/1730613/231773375-4588c483-bfc9-495a-9871-fbce2fd2e0ef.png">

I also noticed the problem was aggregated by the size of the batch of new keys added to the inner Map.

After stepping into the save/reset/isModified methods using a debugger I noticed the following : 
- $__reset iterates through all the subdocuments
https://github.com/Automattic/mongoose/blob/7971a4dbd55888f0b005e65b06024109af8352f7/lib/document.js#L3277
- For each sub-document, we iterate on the modified paths : https://github.com/Automattic/mongoose/blob/7971a4dbd55888f0b005e65b06024109af8352f7/lib/document.js#L2198

So the reset method seems to operate in O(n_subdocs * n_modified_subdocs), which seems the biggest culprit here.

As a workaround, since I am not modifying the object after save() (and I no longer use the object), replacing `.save()` by `model.updateOne(model.getChanges())` yields a large performance improvement (my bench goes from 45s to 1s)

### Steps to Reproduce

Unfortunately I can't share the benchmarking code as is. 
Having a document with a large set of subdocument (for instance in the map), and trying to modify/add new subdocuments by batch should be enough to reproduce the problem.

### Expected Behavior

$__reset should probably not a bottleneck when saving a document.

alexd6631:
After second look, it seems to be similar of : https://github.com/Automattic/mongoose/issues/13191 
I am keeping the issue open, since the profile flame-graph and pre-analysis done here, could help fix the issue.

Issue #13191, opened 2023-03-21
Mongoose takes a very long time to save large Maps with subdocuments to the database


TheYuriG:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.0.2

### Node.js version

16.9.0

### MongoDB server version

latest

### Description

Mongoose takes an absurdly long time to save a large Map to the database. This problem doesn't happen if you replace the Maps for Arrays.

This was initially reported at #5925 and requested to be made into a new issue when the reproduction script was created, so here it is.

### Steps to Reproduce

Clone my reproduction repo [here](https://github.com/TheYuriG/mongoose-map-saving-issue), then either edit your mongoKey into the `index.js` file or add a `mongoKey.js` file to `/private` so it can be imported by the `index.js` file.

Once that's done, you can either use `npm start` to start the script in colored mode or `npm run colorless` to use colorless mode.

All major steps are color-coded, if you choose to retain the colors, so you can more easily parse the information.

The script will connect to a database and create 20K documents called Minis that only have a small string and a number, then use those to sort and create a Map and an Array version of those documents, sorted by timestamp value.

I've reused the script a few times and I consistently get ~6-minute timers for saving the Maps using mongoose, while saving the Arrays usually takes around ~5-seconds.

Try out the script and let me know if this is an actual bug with the library double checking If every value actually isn't a duplicate of the other value or if this was a [PEBCAK](https://en.wiktionary.org/wiki/PEBCAK).

### Expected Behavior

Expected behavior was to not take over 6 minutes to save the Map to the database.

TheYuriG:
At last, closure! Thanks Valeri!

vkarpov15:
Thanks for your patience, took a lot of debugging but the bottleneck should be gone now :+1:

PR #13280, opened 2023-04-16
perf: speed up creating maps of subdocuments

vkarpov15:
**Summary**

Re: #13191, #13271.

Benchmark from #13191 with 2000 documents, before this change:

```
Starting the process of mapping all minis and timing it
Iterating through all minis: 5.661ms
Sorting all minis: 1.368ms
Creating and populating Map of Minis: 0.4ms
Saving Map of Minis to MongoDB using Mongoose: 1.684s
Map Process: 1.692s
Saved Minis Map successfully!
```

After:

```
Starting the process of mapping all minis and timing it
Iterating through all minis: 6.554ms
Sorting all minis: 1.872ms
Creating and populating Map of Minis: 0.555ms
Saving Map of Minis to MongoDB using Mongoose: 315.028ms
Map Process: 324.444ms
Saved Minis Map successfully!
```

I would love to make this faster and will work on it some more, as well as adding a benchmark for this, but figured I'd add a PR first.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2221,9 +2221,8 @@ Document.prototype.isModified = function(paths, modifiedPaths) {
       paths = paths.indexOf(' ') === -1 ? [paths] : paths.split(' ');
     }
 
-    const directModifiedPathsObj = this.$__.activePaths.getStatePaths('modify');
-    const directModifiedPaths = Object.keys(directModifiedPathsObj);
-    if (directModifiedPaths.length === 0) {
+    const directModifiedPathsObj = this.$__.activePaths.states.modify;
+    if (directModifiedPathsObj == null) {
       return false;
     }
     for (const path of paths) {
@@ -2237,6 +2236,7 @@ Document.prototype.isModified = function(paths, modifiedPaths) {
       return !!~modified.indexOf(path);
     });
 
+    const directModifiedPaths = Object.keys(directModifiedPathsObj);
     return isModifiedChild || paths.some(function(path) {
       return directModifiedPaths.some(function(mod) {
         return mod === path || path.startsWith(mod + '.');
@@ -3324,10 +3324,7 @@ Document.prototype.$__reset = function reset() {
   let _this = this;
 
   // Skip for subdocuments
-  if (this.$isDocumentArrayElement || this.$isSingleNested) {
-    return this;
-  }
-  const subdocs = this.$getAllSubdocs();
+  const subdocs = this.$parent() === this ? this.$getAllSubdocs() : [];
   const resetArrays = new Set();
   for (const subdoc of subdocs) {
     const fullPathWithIndexes = subdoc.$__fullPathWithIndexes();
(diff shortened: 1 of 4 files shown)


lib/document.js
```diff
@@ -2217,19 +2217,26 @@ Document.prototype[documentModifiedPaths] = Document.prototype.modifiedPaths;
...
     if (!Array.isArray(paths)) {
       paths = paths.indexOf(' ') === -1 ? [paths] : paths.split(' ');
     }
+
+    const directModifiedPathsObj = this.$__.activePaths.states.modify;
+    if (directModifiedPathsObj == null) {
+      return false;
+    }
+    for (const path of paths) {
+      if (directModifiedPathsObj.hasOwnProperty(path)) {
```
Uzlopak, line 2229: Should use Object.prototype.hasOwnProperty. but still dangerous if the object is not using NullObject.

lib/document.js
```diff
@@ -2217,19 +2217,26 @@ Document.prototype[documentModifiedPaths] = Document.prototype.modifiedPaths;
...
+    }
+
     const modified = modifiedPaths || this[documentModifiedPaths]();
     const isModifiedChild = paths.some(function(path) {
       return !!~modified.indexOf(path);
     });
 
+    const directModifiedPaths = Object.keys(directModifiedPathsObj);
     return isModifiedChild || paths.some(function(path) {
       return directModifiedPaths.some(function(mod) {
```
Uzlopak, line 2241: Can we avoid this some() call or atleast extract the function?

lib/document.js
```diff
@@ -2217,19 +2217,26 @@ Document.prototype[documentModifiedPaths] = Document.prototype.modifiedPaths;
...
         return mod === path || path.startsWith(mod + '.');
```
Uzlopak, line 2242: Maybe just calling startsWith is more ergonomic than making the direct comparison. 
Or maybe store path length in a variable and then do someting like

return mod.length === pathLength && mod === path || mod.length > pathLength && mod[pathLength] === "." && path.startsWith(mod)

Uzlopak: If we do this we would need to put it into a for loop. We could theoretically do directly a for of loop of Object.keys as Object.keys returns an iterator and it should not make a perf difference.

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

lib/document.js
vkarpov15: What alternative would you prefer? For loop?

lib/document.js
vkarpov15: Not a bad idea, but I'd like to do some additional benchmarking to see if this makes a meaningful difference. We'll keep #13191 open for future work related to this.

lib/document.js
Uzlopak: I think its interconnected with the next comment.

PR #13286, opened 2023-04-17
add clusteredIndex to schema options

jakesjews:
**Summary**

This allows specifying that a schema's collection should be created as a [clustered collection](https://www.mongodb.com/docs/manual/core/clustered-collections/).

**Examples**
```js
const schema = Schema(
  { 
    name: String 
  },
  { 
    clusteredIndex: { 
      key: { _id: 1 }
    }
  }
);
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1400,8 +1400,7 @@ Model.createCollection = async function createCollection(options) {
     this.schema.options &&
     this.schema.options.clusteredIndex;
   if (clusteredIndex != null) {
-    options = Object.assign({ clusteredIndex }, options);
-    options.clusteredIndex = { ...clusteredIndex, unique: true };
+    options = Object.assign({ clusteredIndex: { ...clusteredIndex, unique: true } }, options);
   }
 
   try {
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -1395,6 +1395,15 @@ Model.createCollection = async function createCollection(options) {
     }
   }
 
+  const clusteredIndex = this &&
+    this.schema &&
+    this.schema.options &&
+    this.schema.options.clusteredIndex;
+  if (clusteredIndex != null) {
+    options = Object.assign({ clusteredIndex }, options);
+    options.clusteredIndex.unique = true;
```
vkarpov15, line 1404: Best to clone the `clusteredIndex` value before modifying it, `options.clusteredIndex = { ...clusteredIndex, unique: true }`

vkarpov15 (changes requested):
One minor comment, looks good otherwise

lib/model.js
jakesjews: no problem!

vkarpov15 (approved):

PR #13292, opened 2023-04-19
fix(types): change "bulkWrite" to not be strict by default and allow overwriting

hasezoey:
**Summary**

This PR modifies `Model.bulkWrite` to be overwrite-able and not be strict by default (i am not quite sure if it is wanted)

this changes `bulkWrite` to be more in-line with functions like `create` & `insertMany`


vkarpov15 (approved):
LGTM :+1:

PR #13298, opened 2023-04-19
Use `collection.findOne()` for saving docs with no changes

### LINKED ISSUE #13250: Saving a doc without modification triggers post findOne with a pojo
### Prerequisites

### Mongoose version

7.0.2

### Node.js version

16.15.0

### MongoDB server version

5.0.0

_No response_

### Description

Calling save methods on a document without modification triggers post `findOne` hooks, and the doc passed to the function is a pojo with just an _id instead of a mongoose document.
The problem occurs because of the `$__handleSave` method in `/lib/model.js`, which when there is no `delta` calls `this.constructor.exists`, which in turn performs a `findOne` with a `select` on `_id`, and a call to the `lean` method, ultimately returning a pojo

### Steps to Reproduce

```js

describe('Test', () => {
 let model, doc

 before(async () => {
  const conString = 'mongodb://localhost:27017/test'
  const con = await mongoose.createConnection(conString).asPromise()
  const schema = new mongoose.Schema({ data: Boolean })

  schema.post('findOne', async doc => {
   console.log(doc, doc.constructor)
  })

  model = con.model('test', schema)
 })

 it('create data and save with modification', async () => {
  doc = await model.create({ data: false })
  doc.set({ data: true })
  await doc.save()
 })

 it('create data and save without modification', async () => {
  doc = await model.create({ data: false })
  await doc.save()
 })
})

```

### Expected Behavior

findOne hooks should not be triggered by a save

IslandRhythms:
closes #13250

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -324,7 +324,6 @@ Model.prototype.$__handleSave = function(options, callback) {
   // Make sure we don't treat it as a new object on error,
   // since it already exists
   this.$__.inserting = false;
-
   const delta = this.$__delta();
   if (delta) {
     if (delta instanceof MongooseError) {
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #13301, opened 2023-04-20
added correct link

IslandRhythms:


lib/schematype.js
```diff
@@ -825,7 +825,7 @@ SchemaType.prototype.get = function(fn) {
  * From the examples above, you may have noticed that error messages support
  * basic templating. There are a few other template keywords besides `{PATH}`
  * and `{VALUE}` too. To find out more, details are available
- * [here](#error_messages_MongooseError-messages).
+ * [here](/docs/api/error.html#Error.prototype.name).
```
vkarpov15, line 828: We make links relative in the docs now, not sure exactly which PR that was but @hasezoey worked on it. So I think this should be `[here](./error.html#Error.prototype.name)` . But please test and confirm.

vkarpov15 (changes requested):

lib/schematype.js
IslandRhythms: works

hasezoey: actually we mostly currently use absolute paths in the jsdoc, for example:

https://github.com/Automattic/mongoose/blob/master/lib/index.js#L931-L938

i wanted to change this at some point so that the urls are valid for both in-editor use and on the website (via using full paths with `https://mongoosejs.com/` and changing the paths in the website processing)

hasezoey (approved):

vkarpov15 (approved):

Andrey-Yur:
@vkarpov15 @IslandRhythms  Sorry, link does not work yet. But, I am looking for information (the list) with *_***other template keywords besides {PATH} and {VALUE}****_  So, the question regarding the basic templates in mongoose
image

hasezoey:
> Sorry, link does not work yet.

the link you have in your screenshot is the link that this PR replaces, not adds and the PR's changes are not live yet (website only updates when a version releases)

> But, I am looking for information (the list) with _other template keywords besides {PATH} and {VALUE}_

there does not seem to be any documentation about that yet, but for reference:
- `CastError` templates can be seen at https://github.com/Automattic/mongoose/blob/bc69fc8f281dcb4d2ab43a79e6cd67b23ef99e1a/lib/error/cast.js#L126-L134
- `ValidationError` supports templates based on the properties, see https://github.com/Automattic/mongoose/blob/bc69fc8f281dcb4d2ab43a79e6cd67b23ef99e1a/lib/error/validator.js#L89 and https://github.com/Automattic/mongoose/blob/bc69fc8f281dcb4d2ab43a79e6cd67b23ef99e1a/test/errors.validation.test.js#L204-L210

Note: i dont know if what i said is totally correct, i just looked at the format functions themself (not the full call chain)

Andrey-Yur:
@hasezoey @vkarpov15 Thanks for your answer! I would be grateful if you would give a link where I can find ALL the available templates keywords. From provided info, for example, it is clear for me since now, and I guess that in additional to `CastError/ValidateError` templates were just shared by you,  are present something like {SCHEMA}, {DOCUMENT} and etc.,does it? Could you share that too, pls?

hasezoey:
> Could you share whole set of the avaible mongoose templates too, pls?

as i said in the earlier comment, there is currently no documentation for all templates available (see #13311) and the listed code example i gave were just quick findings from me (i dont know if they are everything or if they are even correct)

PR #13304, opened 2023-04-20
Update document.d.ts set function params

jeffersonlipsky:
```js
/**
 * Alias for `set()`, used internally to avoid conflicts *
 * @param {String|Object} path path or object of key/vals to set
 * @param {Any} val the value to set
 * @param {Schema|String|Number|Buffer|*} [type] optionally specify a type for "on-the-fly" attributes
 * @param {Object} [options] optionally specify options that modify the behavior of the set
 * @param {Boolean} [options.merge=false] if true, setting a [nested path](/docs/subdocs.html#subdocuments-versus-nested-paths) will merge existing values rather than overwrite the whole object. So `doc.set('nested', { a: 1, b: 2 })` becomes `doc.set('nested.a', 1); doc.set('nested.b', 2);`
 * @return {Document} this
 * @method $set
 * @memberOf Document
 * @instance
 * @api public */
```

**Summary**

**Examples**

### DIFF
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -90,9 +90,9 @@ declare module 'mongoose' {
     $session(session?: ClientSession | null): ClientSession | null;
 
     /** Alias for `set()`, used internally to avoid conflicts */
-    $set(path: string, val: any, type: any, options?: DocumentSetOptions): this;
-    $set(path: string, val: any, options?: DocumentSetOptions): this;
-    $set(value: any): this;
+    $set(path: string | Record<string, any>, val: any, type: any, options?: DocumentSetOptions): this;
+    $set(path: string | Record<string, any>, val: any, options?: DocumentSetOptions): this;
+    $set(value: string | Record<string, any>): this;
 
     /** Set this property to add additional query filters when Mongoose saves this document and `isNew` is false. */
     $where: Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


hasezoey (approved):
Looks good to me, maybe consider adding some type tests so that it does not get forgotten in the future

types/document.d.ts
```diff
@@ -90,9 +90,9 @@ declare module 'mongoose' {
     $session(session?: ClientSession | null): ClientSession | null;
 
     /** Alias for `set()`, used internally to avoid conflicts */
-    $set(path: string, val: any, type: any, options?: DocumentSetOptions): this;
-    $set(path: string, val: any, options?: DocumentSetOptions): this;
-    $set(value: any): this;
+    $set(path: string | Record<string, any>, val: any, type: any, options?: DocumentSetOptions): this;
```
vkarpov15, line 93: This isn't quite correct, Mongoose ignores `val` if `path` is an object. So I think the more correct approach would be to add an additional override to `$set()` that looks like this:

```
$set(Record<string, any>, val?: null | undefined, type?: any, options?: DocumentSetOptions): this;
```

What do you think?

hasezoey: > This isn't quite correct, Mongoose ignores val if path is an object. 

if i understand it correctly, a additional overload would be needed like

```ts
$set(path: Record<string, any>, type: any, options?: DocumentSetOptions): this;
```

or is `type` not just assigned from parameter 2 and still needs `val: never`?

vkarpov15 (approved):
LGTM. I took a closer look and, in order to apply the changes I suggested, we'd also need to change the function signature for `set()`, and I'd rather not do that for now. Given that this makes our types more consistent, I'm going to merge this as is.

PR #13313, opened 2023-04-22
[Docs] reload api js files on change

hasezoey:
**Summary**

This PR changes the documentation to be able to re-load the api js file to reflect any changes, previously api would be reload-able, but the reload would not reflect any changes from the original js file, only if the pug files were changed

in more detail:
- changes the `parse` function in `docs/source/api` to be in 3 different functions:
  - process the properties `processFile`
  - load a file with dox and process file `parseFile`
  - load and process all defined files `parseAllFiles`
- removes the global storage of the raw js parsed js files in `docs/source/api` (because it only needs to be loaded once and on reload would need to reload the anyway)
- update `scripts/website` to be able to use the reload function
- fix a accidental wrong type import in `docs/source/utils.js` from a previous PR

PS: the current approach is a little "hacky" because the `docs/source/api` file will always load it once at file-load time and the file-map `docs/source/index` will always store this one value and not update on reload, but the `filemap.apiDocs` value will update, this means that unless specifically loaded, the file-maps value will never update beyond the initial load - which should not pose a problem with the way the documentation is currently done (and for any scripts which will not make use of reload)

### DIFF
--- a/docs/source/api.js
+++ b/docs/source/api.js
@@ -39,8 +39,8 @@ const files = [
   'lib/types/map.js'
 ];
 
-/** @type {DocsObj[]} */
-const out = module.exports.docs = [];
+/** @type {Map.<string, DocsObj>} */
+const out = module.exports.docs = new Map();
 
 // add custom matchers to dox, to recognize things it does not know about
 // see https://github.com/tj/dox/issues/198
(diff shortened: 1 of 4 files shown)


vkarpov15 (approved):
Thanks :+1:

Issue #13314, opened 2023-04-23
Enhance sidebar menu to understand navigation the menu.


elias-soykat:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Hi.

When i read mongoose documentation it's really hard to find out which menu/options i read or navigating. Suppose when i read the `Schemas` section using left sidebar then scroll little bit I don't understand which section I navigating currently. 

I share a screenshot to understand properly.

<img width="1169" alt="Screenshot 2023-04-23 at 1 26 28 PM" src="https://user-images.githubusercontent.com/76895393/233826334-2b89c5b4-9f24-401a-b8d1-e732276ce30f.png">

Thank you

hasezoey:
actually it gets bolded (via the `selected` css class), but somehow this does just not apply to the `Schemas` entry

but i also think that this bold-ing that is done is also not quite visible enough

hasezoey:
with #13320 and #13321 merged and also deployed, is the sidebar easier to understand now?

@elias-soykat 

(Note that only the 7.x documentation have those changes deployed currently)

PR #13320, opened 2023-04-24
docs(layout): fix "Schemas" not being highlighted when currently on page

hasezoey:
**Summary**

This PR fixes `Schemas` entry in the documentation sidebar not being highlighted while being on the page

re #13314


vkarpov15:
Looks like this conflicts with changes from #13321?

hasezoey:
maybe, will fix

PR #13326, opened 2023-04-25
types(query): make `lean()` flatten out inferred maps into Record<string, V>

### LINKED ISSUE #13010: `lean()` conflicts with `Map` type returned by `InferSchemaType`
### Prerequisites

### Issue

`InferSchemaType` returns `Map` type for a field of type `Map`, which is perfectly fine for full queries like `Model.findOne()`, but does not work for lean queries `Model.findOne().lean()`: 

```typescript
import mongoose, { InferSchemaType, model, Schema } from 'mongoose';

const schema = new Schema({
    name: { required: true, type: Map, of: String },
});

// infers { name: Map<string, string> }
export type CountryRaw = InferSchemaType<typeof schema>;

export const CountryModel = model('Country', schema);

(async () => {
    await mongoose.connect('mongodb://localhost/test');

    await CountryModel.create({
        name: {
            en: 'Croatia',
            ru: 'Хорватия',
        },
    });

    const country: CountryRaw | null = await CountryModel.findOne().lean().exec();
    if (country) {
        // produces "TypeError: country.name.get is not a function",
        // because country.name is a plain object, not a Map
        console.log('country name', country.name.get('en'));
    }
})();
```
Possible solution or quick fix is to allow to pass some flag to `lean()` function that will tell to convert plain object to a `Map` object:
```
// country.name will be a Map
const country: CountryRaw | null = await CountryModel.findOne().lean({ convertToMap: true }).exec();
```
Or maybe even better to use a special plugin for it.

vkarpov15:
Fix #13010

**Summary**

`lean()` queries never return maps, because Mongoose maps are stored as POJOs in MongoDB. This PR makes so that `lean()` uses `FlattenMaps` by default to replace maps in documents with their `Record<>` equivalent.

**Examples**

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,9 +19,9 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "bson": "^5.0.1",
+    "bson": "^5.2.0",
     "kareem": "2.5.1",
-    "mongodb": "5.1.0",
+    "mongodb": "5.3.0",
     "mpath": "0.9.0",
     "mquery": "5.0.0",
     "ms": "2.1.3",
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -125,3 +125,21 @@ async function _11767() {
   // expectError(examFound2Obj.questions[0].populated);
   expectType<string[]>(examFound2Obj.questions[0].answers);
 }
+
+async function gh13010() {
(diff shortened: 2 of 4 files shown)


hasezoey (approved):
LGTM, aside from one question

PR #13328, opened 2023-04-25
Update jsdoc to use full URLs instead of non-prefix absolute urls (also fix some urls)

hasezoey:
**Summary**

This PR updates the jsdoc to use full paths (`https://mongoosejs.com/path/to/somewhere.html`) instead of `/path/to/somewhere.html`, this way the URLs also work in editors and are arguably easier to add / maintain, also updates urls from `#type_Type-method` to `#Type(.prototype).method`.

- update some wrong urls which had `//`
- fix blc broken links
- include `MongooseArray` in documentation output
- update `eslintrc` to ignore versioned documentation too
- update `scripts/website.js` to process `https://mongoosejs.com/` links to local links and versioned links (if versioning is activated)
- update `lib/` file's jsdoc to use the full qualified url
- update `lib/` file's jsdoc to use the new header-ids

PS: some links in `schematype.html` are currently not valid, but #13310 already fixes that


vkarpov15 (approved):
Thanks :+1:

vkarpov15:
Just need to resolve merge conflicts

vkarpov15 (approved):

Issue #13327, opened 2023-04-25
Document.set doesn't work on nested Maps or plain non-strict objects.


sgpinkus:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.0.5

### Node.js version

v16.18.1

### MongoDB server version

mongo:5.0.9

### Description

Deep `Document.set` using a key with 2 more dots in it like `x.set('a.b.c', 1)` doesn't seem to work for:

 - Map of Maps
 - Map of Mixed 
 - Plain schema object with `strict: false`.
 
It should. It does work for just a top level Mixed type.

### Steps to Reproduce

```
const mongoose = require('mongoose');
const { model, Schema } = mongoose;

var TestSchema =  new Schema({
  a: Schema.Types.Mixed,
  b: {
    type: Map,
    of: Schema.Types.Mixed,
    default: {},
  },
  c: {
    type: Map,
    of: {
      type: Map,
      of: Schema.Types.Mixed,
      default: {},
    },
    default: {},
  },
  d: new Schema({
    }, {
      strict: false,
      _id: false,
    },
  )
});

var Test = model('Test', TestSchema);

async function main() {
  await mongoose.connect('mongodb://localhost:27017/test');
  await mongoose.connection.dropDatabase();
  await setThings(); // Only sets a.x.y
}

async function setThings() {
  var x = new Test({});
  x.set('a.x.y', 1);
  x.set('b.x.y', 1);
  x.set('c.x.y', 1);
  x.set('d.x.y', 1);
  await x.save();
  console.log(x.toObject());
}

main()
  .then(() => process.exit())
  .catch((e) => { console.error(e); process.exit(); })
```

**Gives:**

```
{
  b: Map(1) { 'x' => {} },
  c: Map(0) {},
  _id: new ObjectId("644799e0f03d7fd6496855fb"),
  a: { x: { y: 1 } },
  __v: 0
}
```

### Expected Behavior

All the Document.set calls should work.

Issue #7196, opened 2018-10-29
Single nested document set function is being calling twice


LucGranato:
**Do you want to request a *feature* or report a *bug*?**
*Bug*

**What is the current behavior?**
When setting a value to a single nested document field
```javascript
rootDoc.set('anything.subDoc.someField', 12);
```
The set function is called twice
```javascript
{
  ...
  someField: {
    set: function(value) {
      return value;
    }
  }
  ...
}
```
The first time with `this` as `rootDoc` and the second as `subDoc`.

**If the current behavior is a bug, please provide the steps to reproduce.**
[RunKit Example](https://runkit.com/lucgranato/mongoose-issue-set)

**What is the expected behavior?**
Calling only one time.

**Please mention your node.js, mongoose and MongoDB version.**
node: 8.11.3
mongoose: 5.3.7

PR #13339, opened 2023-04-28
fix(document): handle calling `set()` underneath a strict: false subdocument path from the top-level document

vkarpov15:
Re: #13327

**Summary**

#13327 points out some interesting corner cases that pop up when calling `set()` on a top-level document. This PR addresses one of them: `x.set('d.x.y', 1);` where `d` is a single nested subdocument with no properties but `strict: false`.

To fix this, we needed to iron out a couple of issues:

1. We don't use subdocument strict when calling `set()` on top-level document
2. `set('strict', false)` doesn't bubble down to child schemas
3. Consistent single nested subdoc setter context without needing to rely on `$__set()` to get the correct context re: #7196 

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -26,6 +26,7 @@ const flattenObjectWithDottedPaths = require('./helpers/path/flattenObjectWithDo
 const get = require('./helpers/get');
 const getEmbeddedDiscriminatorPath = require('./helpers/document/getEmbeddedDiscriminatorPath');
 const getKeysInSchemaOrder = require('./helpers/schema/getKeysInSchemaOrder');
+const getSubdocumentStrictValue = require('./helpers/document/getSubdocumentStrictValue');
 const handleSpreadDoc = require('./helpers/document/handleSpreadDoc');
 const immediate = require('./helpers/immediate');
 const isDefiningProjection = require('./helpers/projection/isDefiningProjection');
@@ -52,6 +53,7 @@ const populateModelSymbol = require('./helpers/symbols').populateModelSymbol;
 const scopeSymbol = require('./helpers/symbols').scopeSymbol;
 const schemaMixedSymbol = require('./schema/symbols').schemaMixedSymbol;
 const parentPaths = require('./helpers/path/parentPaths');
+const getDeepestSubdocumentForPath = require('./helpers/document/getDeepestSubdocumentForPath');
 
 let DocumentArray;
 let MongooseArray;
@@ -1032,7 +1034,8 @@ Document.prototype.$set = function $set(path, val, type, options) {
   let key;
   let prefix;
 
-  const strict = options && 'strict' in options
+  const userSpecifiedStrict = options && 'strict' in options;
+  let strict = userSpecifiedStrict
     ? options.strict
     : this.$__.strictMode;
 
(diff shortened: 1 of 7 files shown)


lib/helpers/document/getDeepestSubdocumentForPath.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+module.exports = function getDeepestSubdocumentForPath(doc, parts, schema) {
```
hasezoey, line 3: it would probably help to add jsdoc for this, so it can be understood more easily at a glance

lib/helpers/document/getEmbeddedDiscriminatorPath.js
```diff
@@ -15,7 +15,9 @@ const getSchemaDiscriminatorByValue = require('../discriminator/getSchemaDiscrim
 module.exports = function getEmbeddedDiscriminatorPath(doc, path, options) {
   options = options || {};
   const typeOnly = options.typeOnly;
-  const parts = path.indexOf('.') === -1 ? [path] : path.split('.');
+  const parts = Array.isArray(path) ?
```
hasezoey, line 18: this change should also be reflected in the jsdoc of the function

lib/helpers/document/getSubdocumentStrictValue.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+module.exports = function getSubdocumentStrictValue(parts, schema) {
```
hasezoey, line 3: it would probably help to add jsdoc for this, so it can be understood more easily at a glance

hasezoey (changes requested):
Code looks good to me, though some with some notes

lib/helpers/schema/getSubdocumentStrictValue.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Find the `strict` mode setting for the deepest subdocument along a given path
+ * to ensure we have the correct default value for `strict`. When setting values
+ * underneath a subdocument, we should use the subdocument's `strict` setting by
+ * default, not the top-level document's.
+ *
+ * @param {Schema} schema
+ * @param {String[]} parts
+ * @returns boolean | 'throw' | undefined
```
hasezoey, line 11:
```suggestion
 * @returns {boolean | 'throw' | undefined}
```

hasezoey (approved):
LGTM, one small jsdoc thing

IslandRhythms (approved):
LGTM

lib/document.js
```diff
@@ -1679,7 +1693,7 @@ Document.prototype.$__set = function(pathToMark, path, options, constructing, pa
       } else if (obj[parts[i]] && obj[parts[i]] instanceof Embedded) {
         obj = obj[parts[i]];
       } else if (obj[parts[i]] && !Array.isArray(obj[parts[i]]) && obj[parts[i]].$isSingleNested) {
-        obj = obj[parts[i]];
+        obj = obj[parts[i]]._doc;
```
vkarpov15, line 1696: Undoing this change fixes #13876

PR #13344, opened 2023-04-29
Miscellaneous documentation changes

hasezoey:
**Summary**

This PR contains some miscellaneous changes, mostly being style changes: 
- move carbonads styles to a common file
- move code styles to `github.css`
- remove `default.css` (was only used by the homepage, but thanks to moving all code styles to `github.css`, only that file is still needed)
- remove style `.tertiary-item` and only use `.sub-item`
- consistently set `p`'s `line-height`
- re-use definition for search bars (both for sidebar nav search and search page bar)
- remove polyfill for `unfetch`, because "all" modern browsers have `fetch`


vkarpov15 (approved):
Thanks :+1:

Issue #13345, opened 2023-04-30
`FlattenMaps` doesn't handle `DocumentArray` properly


Jokero:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.1.0

### Node.js version

16.13.1

### MongoDB server version

-

### Typescript version (if applicable)

4.9.5

### Description

Since 7.1.0, `lean()` replaces `Map` in documents with their `Record<string, V>` equivalent by using `FlattenMaps`. `FlattenMaps` returns a type incompatible with `InferSchemaType` type when there is a property of `DocumentArray` type.

### Steps to Reproduce

```ts
import { model, Schema, InferSchemaType } from 'mongoose';
import { expectAssignable } from 'tsd';

async function test() {
    const imageSchema = new Schema({
        url: { required: true, type: String }
    });

    const placeSchema = new Schema({
        images: { required: true, type: [imageSchema] },
    });

    type Place = InferSchemaType<typeof placeSchema>;

    const PlaceModel = model('Place', placeSchema);

    const place = await PlaceModel.findOne().lean().orFail().exec();
    expectAssignable<Place>(place);
}
```
Compilation error:
```
error TS2345: Argument of type 'FlattenMaps<{ images: DocumentArray<{ url: string; }>; }> & { _id: ObjectId; }' is not assignable to parameter of type '{ images: DocumentArray<{ url: string; }>; }'.
  Types of property 'images' are incompatible.
    Type 'FlattenMaps<DocumentArray<{ url: string; }>>' is not assignable to type 'DocumentArray<{ url: string; }>'.
      'number' index signatures are incompatible.
        ...
```

### Expected Behavior

It should compile and do not return an error.

PR #13346, opened 2023-04-30
Handle `DocumentArray` in `FlattenMaps`

Jokero:
Closes https://github.com/Automattic/mongoose/issues/13345

**Summary**

Since 7.1.0, `lean()` replaces `Map` in documents with their `Record<string, V>` equivalent by using `FlattenMaps`. Currently, `FlattenMaps` returns a type incompatible with `InferSchemaType` type when there is a property of `DocumentArray` type. This PR fixes it.

**Examples**

This won't produce compilation error:
```ts
import { model, Schema, InferSchemaType } from 'mongoose';
import { expectType } from 'tsd';

async function test() {
    const imageSchema = new Schema({
        url: { required: true, type: String }
    });

    const placeSchema = new Schema({
        images: { required: true, type: [imageSchema] },
    });

    type Place = InferSchemaType<typeof placeSchema>;

    const PlaceModel = model('Place', placeSchema);

    const place = await PlaceModel.findOne().lean().orFail().exec();
    expectAssignable<Place>(place);
}
```

### DIFF
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -1,5 +1,5 @@
-import { Schema, model, Document, Types } from 'mongoose';
-import { expectError, expectType } from 'tsd';
+import { Schema, model, Document, Types, InferSchemaType, FlattenMaps } from 'mongoose';
+import { expectAssignable, expectError, expectType } from 'tsd';
 
 function gh10345() {
   (function() {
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #13349, opened 2023-05-01
Temporarily remove mongodb memory server from deno tests on 6.x

vkarpov15:
**Summary**

@hasezoey FYI, I'm temporarily removing mongodb memory server from deno tests on 6.x branch because we've been experiencing some weird test failures, see:

1. Error 1: https://github.com/Automattic/mongoose/actions/runs/4853799889/jobs/8650382091
2. Error 2: https://github.com/Automattic/mongoose/actions/runs/4853740990/jobs/8650256294

I'd like to bring mongodb memory server back as soon as we can, but right now these test failures are blocking releasing 6.11.

**Examples**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,21 +78,23 @@ jobs:
   test-deno:
     runs-on: ubuntu-20.04
     name: Deno tests
-    env:
-      MONGOMS_VERSION: 6.0.0
-      MONGOMS_PREFER_GLOBAL_PATH: 1
     steps:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
       - name: Setup node
         uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
         with:
           node-version: 16
-      - name: Load MongoDB binary cache
-        id: cache-mongodb-binaries
-        uses: actions/cache@v3
-        with:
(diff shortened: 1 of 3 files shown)


IslandRhythms (approved):
LGTM

hasezoey:
> error: Uncaught Error: connect ECONNREFUSED 127.0.0.1:46425 - Local (undefined:undefined)

sadly those are the errors i had mentioned in #12918, they are quite hard to debug in this repo and these are (probably) because deno does not support process hooks (like `beforeExit` or `uncaughtException`)

they basically boil down to not closing the connections after a test / test-suite and the mongodb instance getting stopped before they are disconnected, especially because mocha does not isolate tests (and is also running in the same context as the `test/deno.js` itself), so the issues will only show up at the end of all test-suites instead of the test-suite that causes this

hasezoey:
maybe as somewhat of a fix, we could try running mocha in a different thread (like worker?) and stop that worker before stopping mms, though that would not actually fix the issue and only work-around them

vkarpov15:
The workaround you implemented with forking a separate process seems to work well. Hard to think of another alternative because I haven't looked too closely at the cause of these errors.

Why does the lack of process hooks cause an issue?

hasezoey:
> Why does the lack of process hooks cause an issue?

that was my guess, but i now think it is wrong - i dont know how mongodb handles "dangling" connections, and i think this issue didnt previous exist because mocha may run the fixtures in a different process than the test suite (like we do now manually in the deno.js files)

PR #13365, opened 2023-05-02
chore(deno): change to start mocha fixtures before mocha

hasezoey:
**Summary**

This PR changes the deno test to start the fixtures before mocha and to start mocha in a child-process, to workaround the issue about connections still being open at the end of the test suite

only downsides are that deno permissions will have to be set via the `args` array in the code now, because i dont know of a way to start a child deno process with the same permissions

re https://github.com/Automattic/mongoose/pull/13349#issuecomment-1531194531

also reverts #13349

PS: this is basically just a workaround, not fixing the actual issue

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,23 +78,21 @@ jobs:
   test-deno:
     runs-on: ubuntu-20.04
     name: Deno tests
+    env:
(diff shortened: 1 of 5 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #13367, opened 2023-05-02
types(document): correct return type for `Model.prototype.deleteOne()`: promise, not query

### LINKED ISSUE #13223: Document.deleteOne(...).exec is not a function
### Prerequisites

### Mongoose version

7.0.3

### Node.js version

18.15.0

### MongoDB server version

6.0.5

5.0.2

### Description

I'm trying to use .deleteOne method on the Document previously found with editor code suggestion.
Interesting thing is that it's not documented in https://mongoosejs.com/docs/api/document.html although it has the signature in document.d.ts.
```
/** Removes this document from the db. */
deleteOne(options?: QueryOptions): QueryWithHelpers<any, this, TQueryHelpers>;
```

I've tried to await this Query calling exec() on it as with Document.updateOne() or Model.deleteOne() but the error occured while method itself worked well deleting the document.

```
const team = await Team.findById(someId).exec();
await team.updateOne({title: 'My new title'}).exec(); // This works
await team.deleteOne().exec(); // team.deleteOne(...).exec is not a function
// await Team.deleteOne({_id: someId}).exec(); // This works
```

Of course for now i'll use Model method, there's no urgent problem, but the situation itself seemed suspicious to me, so I've reported it.

### Steps to Reproduce

```
const teamSchema = new Schema({
    title: {
        type: String,
        required: true
    }
});
const Team = model('Team', teamSchema);
const team = await Team.create({title: 'Test'});
await team.deleteOne().exec();
```

### Expected Behavior

_Deleted document and awaitable Promise received without errors_

vkarpov15:
Fix #13223

**Summary**

As implemented, `doc.deleteOne()` returns a promise, not a query. We could either update the `deleteOne()` implementation to match `updateOne()`, which _does_ return a query; or we update the TypeScript types. I think changing the TS types is the way to go, because it is less backwards breaking. But we should consider changing `doc.deleteOne()` to return a query in the next major release.

**Examples**

### DIFF
--- a/test/types/document.test.ts
+++ b/test/types/document.test.ts
@@ -39,7 +39,7 @@ const Test = model<ITest>('Test', schema);
 void async function main() {
   const doc = await Test.findOne().orFail();
 
-  expectType<Query<any, TestDocument>>(doc.deleteOne());
+  expectType<Promise<TestDocument>>(doc.deleteOne());
 }();
 
 
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -107,7 +107,7 @@ declare module 'mongoose' {
     db: Connection;
 
     /** Removes this document from the db. */
-    deleteOne(options?: QueryOptions): QueryWithHelpers<any, this, TQueryHelpers>;
+    deleteOne(options?: QueryOptions): Promise<this>;
 
     /**
      * Takes a populated field and returns it to its unpopulated state. If called with


hasezoey:
> But we should consider changing doc.deleteOne() to return a query in the next major release.

yes, is there a issue for tracking / reminder already?

hasezoey (approved):

vkarpov15:
@hasezoey I just opened one, #13369.

PR #13376, opened 2023-05-03
fix: don't modify passed options object

### LINKED ISSUE #13335: options object get mutated
### Prerequisites

### Mongoose version

7.0.3

### Node.js version

18.15.0

### MongoDB server version

5.0

_No response_

### Description

when creating mongo connection we pass a connection string and options object
in this case the options object gets mutated 
My code is taking options from [node-config]( https://github.com/node-config/node-config) which properties make it immutable. This results in crash when mongoose is trying to mutate the passed object.

```bash
mpodolsk$ node bug.mjs
/node_modules/mongoose/lib/helpers/processConnectionOptions.js:49
  opts.config = { autoIndex: false };
              ^

TypeError: Cannot add property config, object is not extensible
```

### Steps to Reproduce

```js
const opts = Object.preventExtensions({});
await mongoose.connect('mongodb+srv://x:pass@somehost.com/db?retryWrites=true&w=majority&readPreference=secondaryPreferred', opts);
```

### Expected Behavior

it copies the properties needed and uses them without modding the passed object

IslandRhythms:
closes #13335

### DIFF
--- a/test/connection.test.js
+++ b/test/connection.test.js
@@ -1548,8 +1548,12 @@ describe('connections:', function() {
     });
   });
   describe('processConnectionOptions', function() {
+    let m = null;
+    after(async() => {
+      await m.disconnect();
+    })
     it('should not throw an error when attempting to mutate unmutable options object gh-13335', async function() {
-      const m = new mongoose.Mongoose();
+      m = new mongoose.Mongoose();
       const opts = Object.preventExtensions({});
       const conn = await m.connect(start.uri + '?retryWrites=true&w=majority&readPreference=secondaryPreferred', opts);
       assert.ok(conn);
--- a/test/connection.test.js
+++ b/test/connection.test.js
@@ -1547,4 +1547,12 @@ describe('connections:', function() {
       // currently cannot write test for continueOnError or errors in general.
     });
   });
+  describe('processConnectionOptions', function() {
+    it('should not throw an error when attempting to mutate unmutable options object gh-13335', async function() {
+      const m = new mongoose.Mongoose();
+      const opts = Object.preventExtensions({});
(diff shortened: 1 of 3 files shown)


test/connection.test.js
```diff
@@ -1547,4 +1547,12 @@ describe('connections:', function() {
       // currently cannot write test for continueOnError or errors in general.
     });
   });
+  describe('processConnectionOptions', function() {
+    it('should not throw an error when attempting to mutate unmutable options object gh-13335', async function() {
+      const m = new mongoose.Mongoose();
+      const opts = Object.preventExtensions({});
+      const conn = await m.connect('mongodb://localhost:27017/db?retryWrites=true&w=majority&readPreference=secondaryPreferred', opts);
```
Uzlopak, line 1554: Maybe on github runner the /mongo database server cant use 27017 as a port and claims a different port, thus breaking the test.

Maybe use getUri from /test/common.js and retrieve the url to the database server.

Maybe create Mongo Client Adapter and pass it to mongoose and check if opts is not the same object as passed to the adapters connect function. It should then even work without starting a mongo database server.

test/connection.test.js
```diff
@@ -1547,4 +1547,12 @@ describe('connections:', function() {
       // currently cannot write test for continueOnError or errors in general.
     });
   });
+  describe('processConnectionOptions', function() {
```
vkarpov15, line 1550: Add an `after()` hook to this `describe()` that calls `await m.disconnect()`. Right now this test leaves a dangling connection.

vkarpov15 (changes requested):
One quick change but otherwise LGTM.

Also, can you please remove the logic in `lib/connection.js` that calls `clone(options)` right after calling `processConnectionOptions()`?

PR #13381, opened 2023-05-06
fix(query+timestamps): avoid sending `overwrite` and `timestamps` options to MongoDB

vkarpov15:
**Summary**

`overwrite` and `timestamps` aren't supported options on the MongoDB server, they're internal to Mongoose, so they should be stored under `_mongooseOptions`. Removing these options makes debug output cleaner and improves perf slightly.

**Examples**


hasezoey (approved):
LGTM

will this affect plugins or other middleware that use these options?

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

vkarpov15:
@hasezoey potentially. No way to know for certain. But `options.timestamps` and `options.overwrite` aren't part of Mongoose's publicly documented API, so I'm ok making this change unless there's some plugin that you know of that will be significantly affected by this change

PR #13389, opened 2023-05-08
types: make `lean()` not clobber result type for `updateOne()`, etc.

### LINKED ISSUE #13382: ts def of updateOne&findOneAndUpdate returns FlattenMap instead of UpdateWriteOpResult, when lean() is used
### Prerequisites

### Mongoose version

7.1.0

### Node.js version

16.20

### MongoDB server version

5.x

4.9.5

### Description

I tried upgrading to mongoose 7.x, but now I'm stuck with some typescript issues.
image

somehow updateOne returns a FlattenMap and not the ResultType defined in updateOne:
image

I would expect a UpdateWriteOpResult here.

### Steps to Reproduce

my UserModel is defined as 

```
const userSchema = new Schema({...});

export interface IDBUser { .. };

const DBUserModel: Model<IDBUser> = model<IDBUser>('User', userSchema);


```

just run updateOne on it and check the return type.

### Expected Behavior

typescript should correctly return the UpdateWriteOpResult

vkarpov15:
Fix #13382

**Summary**

Doing `Model.updateOne().lean()` returns an incorrect type: has `_id`, etc. With this change, queries now have a `QueryOp` generic that they can use to check the current query op. With this change, `Model.deleteOne().lean()`, etc. all reflect their runtime behavior: `lean()` is a no-op.

**Examples**

### DIFF
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -196,3 +196,13 @@ async function gh13345_3() {
   const place = await PlaceModel.findOne().lean().orFail().exec();
   expectAssignable<Place>(place);
 }
+
+async function gh13382() {
+  const schema = new Schema({
+    name: String
+  });
+  const Test = model('Test', schema);
+
+  const res = await Test.updateOne({}, { name: 'bar' }).lean();
+  expectAssignable<{ matchedCount: number, modifiedCount: number }>(res);
+}
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -193,17 +193,24 @@ declare module 'mongoose' {
     collection: Collection;
 
     /** Creates a `count` query: counts the number of documents that match `filter`. */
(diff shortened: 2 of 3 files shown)


hasezoey (approved):

AbdelrahmanHafez (approved):
LGTM, thanks. 👍 

I think it's worth taking a note to deprecate using `lean` on non-find operations like `deleteOne`/`countDocuments` in both TS types and JS runtime, it doesn't make sense to call them with `lean`.

AbdelrahmanHafez (approved):
LGTM, thanks. 👍 

I think it's worth taking a note to deprecate using `lean` on non-find operations like `deleteOne`/`countDocuments` in both TS types and JS runtime, it doesn't make sense to call them with `lean`.

vkarpov15:
Potentially. As a general rule of thumb, Mongoose doesn't throw errors on unrecognized or unused options, so throwing an error when using `lean()` on an operation that doesn't handle documents would be a bit inconsistent. I opened an issue to consider that for the future.

PR #13397, opened 2023-05-13
feat(query): add `translateAliases` option to automatically call translate aliases on query fields

### LINKED ISSUE #7511: Feature: Automatically call translateAlias for queries.
FEATURE

I plan on implementing this at some point, but would like to propose my method for doing this for some feedback.

I propose an option when defining the schema to indicate if the schema has aliased keys. If so, translateAliases can be called automatically when doing queries on the model.

If not, then the translateAliases call can be skipped entirely, so it avoids traversing the schema if the user is not using aliases.

### LINKED ISSUE #8678: Automatically `translateAliases()` for queries and updates
```
import {
  addModelToTypegoose,
  arrayProp,
  buildSchema,
  getModelForClass,
  index,
  mongoose,
  plugin,
  pre,
  prop,
  Ref,
} from '@typegoose/typegoose';



export class Product extends BaseModel<Product> {
  static readonly collectionName = 'Products';
  @prop({
    required: true,
    index: true,
    sparse: true,
    alias: 'name2',
  })
  name: string;

}


```

```
import { SchemaOptions } from 'mongoose';

export const schemaOptions: SchemaOptions = {
  strict: true,
  timestamps: true,
  selectPopulatedPaths: false,
  minimize: true,
  toJSON: {
    virtuals: true,
    getters: true,
    transform: (doc, ret, options) => {
      delete ret.deleted;
      delete ret.__v;
      delete ret._id;
    },
  },
  toObject: {
    virtuals: true,
    getters: true,
    transform: (doc, ret, options) => {
      delete ret.deleted;
      delete ret.__v;
      delete ret._id;
    },
  },
};

```

I am not able to run an update query

Debug log
`Mongoose: Products.updateOne({ _id: ObjectId("5e6e2c2ea1e74d2362c593e3") }, { '$setOnInsert': { createdAt: new Date("Tue, 17 Mar 2020 15:13:57 GMT") }, '$set': { updatedAt: new Date("Tue, 17 Mar 2020 15:13:57 GMT") }}, {})`

vkarpov15:
Fix #7511
Fix #8678

**Summary**

Add `translateAliases` option on query, schema, and global level to apply `Model.translateAliases()` on all query fields automatically: `filter`, `projection`, `update`, and `distinct`.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -2662,7 +2662,7 @@ Query.prototype._applyTranslateAliases = function _applyTranslateAliases(options
     return;
   }
 
-  if (this.model && this.model.schema && Object.keys(this.model.schema.aliases).length > 0) {
+  if (this.model?.schema?.aliases && Object.keys(this.model.schema.aliases).length > 0) {
     this.model.translateAliases(this._conditions, true);
     this.model.translateAliases(options.projection, true);
     this.model.translateAliases(this._update, true);
--- a/lib/query.js
+++ b/lib/query.js
@@ -154,6 +154,18 @@ Query.prototype = new mquery();
 Query.prototype.constructor = Query;
 Query.base = mquery.prototype;
 
+/*!
+ * Overwrite mquery's `_distinct`, because Mongoose uses that name
+ * to store the field to apply distinct on.
+ */
+
+Object.defineProperty(Query.prototype, '_distinct', {
+  configurable: true,
+  writable: true,
+  enumerable: true,
+  value: undefined
+});
+
 /**
  * Flag to opt out of using `$geoWithin`.
  *
@@ -1618,6 +1630,11 @@ Query.prototype.setOptions = function(options, overwrite) {
     this._mongooseOptions.defaults = options.defaults;
     // deleting options.defaults will cause 7287 to fail
   }
+  if ('translateAliases' in options) {
+    this._mongooseOptions.translateAliases = options.translateAliases;
+    delete options.translateAliases;
+  }
+
   if (options.lean == null && this.schema && 'lean' in this.schema.options) {
     this._mongooseOptions.lean = this.schema.options.lean;
   }
@@ -2237,6 +2254,9 @@ Query.prototype._find = async function _find() {
   });
 
   const options = this._optionsForExec();
+
+  this._applyTranslateAliases(options);
+
   const filter = this._conditions;
   const fields = options.projection;
 
@@ -2486,6 +2506,8 @@ Query.prototype._findOne = async function _findOne() {
 
   const options = this._optionsForExec();
 
+  this._applyTranslateAliases(options);
+
   // don't pass in the conditions because we already merged them in
   const doc = await this._collection.collection.findOne(this._conditions, options);
   return new Promise((resolve, reject) => {
@@ -2582,8 +2604,11 @@ Query.prototype._count = async function _count() {
   applyGlobalMaxTimeMS(this.options, this.model);
   applyGlobalDiskUse(this.options, this.model);
 
-  const conds = this._conditions;
   const options = this._optionsForExec();
+
+  this._applyTranslateAliases(options);
+
+  const conds = this._conditions;
 
   return this._collection.collection.count(conds, options);
 };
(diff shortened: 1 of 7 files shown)


lib/query.js
```diff
@@ -2609,12 +2634,43 @@ Query.prototype._countDocuments = async function _countDocuments() {
...
+ * on the following query properties: filter, projection, update, distinct.
+ */
+
+Query.prototype._applyTranslateAliases = function _applyTranslateAliases(options) {
+  let applyTranslateAliases = false;
+  if ('translateAliases' in this._mongooseOptions) {
+    applyTranslateAliases = this._mongooseOptions.translateAliases;
+  } else if (this?.model?.schema?._userProvidedOptions?.translateAliases != null) {
+    applyTranslateAliases = this.model.schema._userProvidedOptions.translateAliases;
+  } else if (this?.model?.base?.options?.translateAliases != null) {
```
hasezoey, line 2657:
```suggestion
  } else if (this.model?.schema?._userProvidedOptions?.translateAliases != null) {
    applyTranslateAliases = this.model.schema._userProvidedOptions.translateAliases;
  } else if (this.model?.base?.options?.translateAliases != null) {
```

`this` shouldnt be undefined here, so it should not need optional chaining, otherwise it would need to be added in the first `if` too

hasezoey (changes requested):
looks good to me, though there are some thing that should still be done:
- test if the alias takes priority or errors if both are defined at the same time in the query
- test if the alias has any effect if the alias is used in a query but the actual property in a update (or vise-versa)
- document the above behavior somewhere on the documentation

also i dont know how feasible it is, but maybe consider calling `_applyTranslateAliases` in `_optionsForExec`?

lib/query.js
```diff
@@ -1618,6 +1630,11 @@ Query.prototype.setOptions = function(options, overwrite) {
     this._mongooseOptions.defaults = options.defaults;
     // deleting options.defaults will cause 7287 to fail
   }
+  if ('translateAliases' in options) {
+    this._mongooseOptions.translateAliases = options.translateAliases;
+    delete options.translateAliases;
```
AbdelrahmanHafez, line 1635: Does this mutate the provided options object?

lib/query.js
```diff
@@ -2609,12 +2634,43 @@ Query.prototype._countDocuments = async function _countDocuments() {
...
+    applyTranslateAliases = this.model.base.options.translateAliases;
+  }
+  if (!applyTranslateAliases) {
+    return;
+  }
+
+  if (this.model && this.model.schema && Object.keys(this.model.schema.aliases).length > 0) {
+    this.model.translateAliases(this._conditions);
+    this.model.translateAliases(options.projection);
+    this.model.translateAliases(this._update);
```
AbdelrahmanHafez, line 2667: Would this handle update operators such as `$inc`, `$unset`?
If not, I think it's okay to ship this as is and maybe support this in a future version, but I think it's worth having a test that verifies that at least `translateAliases` doesn't break the updates that include those operators.

AbdelrahmanHafez (commented):
Got a couple of questions, but overall LGTM. Thanks! 👍

lib/query.js
```diff
@@ -2609,12 +2634,43 @@ Query.prototype._countDocuments = async function _countDocuments() {
...
+  } else if (this.model?.schema?._userProvidedOptions?.translateAliases != null) {
+    applyTranslateAliases = this.model.schema._userProvidedOptions.translateAliases;
+  } else if (this.model?.base?.options?.translateAliases != null) {
+    applyTranslateAliases = this.model.base.options.translateAliases;
+  }
+  if (!applyTranslateAliases) {
+    return;
+  }
+
+  if (this.model && this.model.schema && Object.keys(this.model.schema.aliases).length > 0) {
```
AbdelrahmanHafez, line 2664: Nitpick:
```suggestion
  if (this.model?.schema?.aliases && Object.keys(this.model.schema.aliases).length > 0) {
```

lib/query.js
vkarpov15: No, there's a shallow clone on L1590

lib/query.js
vkarpov15: Good point, we need to handle atomic update operators.

vkarpov15: Fixed in 198aec9

hasezoey (approved):

PR #13403, opened 2023-05-15
feat(ts): Export MongooseError class

### LINKED ISSUE #13387: Make sure MongooseError is exported in TypeScript
### Prerequisites

### 🚀 Feature Proposal

A tool to detect if error is MongooseError.
The most simple approach is to check `if (error instanceof MongooseError)`. But unfortunately this class is not exported, so one need to check if error is instance of all the exported descendants of it.
The idea is to export a very simple member:

```js

const ifMongooseError = (error: Error) : boolean => error instanceof MongooseError;

``` 

This will allow to find out if it's Mongoose's error and do not lead to leaking of private MongooseError class.

### Motivation

Sometimes one need to transform thrown by Mongoose and caught in `.catch()`/`catch` or passed to errors handler(s) by `.catch(next)` (if using Express). It would be very nice to have a tool to detect if it's a Mongoose error, since several sources of error might exist in the same controller, like celebrate does (celebrate.isCelebrateError() method).
At present, I use a helper that check against all 16 types of errors. It would be way more refined and elegant to have a built-in Mongoose method.

### Example

```js

// ...somewhere in errors handler

elseif (mongoose.isMongooseError(error)) {
 /* Process, transform and/or send to client error */
}

```

ramos-ph:
**Summary**

Exports the class `MongooseError` so it can be tested using `err instanceof MongooseError`

**Examples**

```ts

const Model = mongoose.model('model', new mongoose.Schema({ answer: Number }));

async function main() {
  const doc = new Model({ answer: 'not a number' });
  const err = doc.validateSync();

  console.log(err instanceof mongoose.Error); // true
  console.log(err instanceof mongoose.MongooseError); // true
  console.log(err instanceof mongoose.Error.ValidationError); // true
}

main();
```

Closes #13387

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,6 +46,7 @@ module.exports.Mixed = mongoose.Mixed;
 module.exports.Date = mongoose.Date;
 module.exports.Number = mongoose.Number;
 module.exports.Error = mongoose.Error;
+module.exports.MongooseError = mongoose.MongooseError;
 module.exports.now = mongoose.now;
 module.exports.CastError = mongoose.CastError;
 module.exports.SchemaTypeOptions = mongoose.SchemaTypeOptions;
--- a/lib/index.js
+++ b/lib/index.js
@@ -1140,6 +1140,7 @@ Mongoose.prototype.Number = SchemaTypes.Number;
  */
 
 Mongoose.prototype.Error = require('./error/index');
+Mongoose.prototype.MongooseError = require('./error/mongooseError');
 
 /**
  * Mongoose uses this function to get the current time when setting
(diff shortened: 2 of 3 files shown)


vkarpov15 (commented):
The code for this change looks fine, but I don't think this is the right approach for fixing #13387. Like OP said, the ideal approach would be just to check if `err instanceof MongooseError`. So I think the better approach would be to make sure MongooseError is exported for use in TypeScript.

vkarpov15 (approved):
LGTM :+1: I'll merge this into 7.2 branch.

PR #13404, opened 2023-05-16
fix: fixed types to allow string in interfaces for ObjectId in schemas

### LINKED ISSUE #13288: Allow string as type in interfaces for ObjectId
### Prerequisites

### 🚀 Feature Proposal

Because we need ObjectIds as strings in our application, I want to use the approach of #6996. This is great, but has one downside. We define interfaces for our models and also use them in the frontend. to achieve this to work, I have to define an id as `string | mongoose.Types.ObjectId`, otherwise the model has the following error:
```ts
Type '{ type: typeof mongoose.Schema.Types.ObjectId; ref: string; }' is not assignable to type 'SchemaDefinitionProperty<string>'.
  Types of property 'type' are incompatible.
    Type 'typeof ObjectId' is not assignable to type 'typeof Mixed | StringSchemaDefinition'.
      Type 'typeof ObjectId' is not assignable to type 'typeof Mixed'.
        Types of property 'schemaName' are incompatible.
          Type '"ObjectId"' is not assignable to type '"Mixed"'.ts(2322)
```
It would be nice if I could define the id as `string` in the interface and in the model define it as `type: Schema.Types.ObjectId`.

### Motivation

cleaner interfaces

### Example

Interface:
```ts
export interface ITestInterface {
  _id: string;
  test: string;
}
```

Schema:
```ts
export const testInterfaceSchema = new Schema<
  ITestInterface
>(
  {
    test: {
      type: Schema.Types.ObjectId,
      ref: 'TestInterface',
    },
  }
);
```

rkreienbuehl:
**Summary**

Allowing string in interfaces for ObjectId in schemas.
Fixes #13288

### DIFF
--- a/types/schematypes.d.ts
+++ b/types/schematypes.d.ts
@@ -41,7 +41,7 @@ declare module 'mongoose' {
 
   class SchemaTypeOptions<T> {
     type?:
-    T extends string ? StringSchemaDefinition :
+    T extends string ? StringSchemaDefinition | ObjectIdSchemaDefinition :
       T extends number ? NumberSchemaDefinition :
         T extends boolean ? BooleanSchemaDefinition :
           T extends NativeDate ? DateSchemaDefinition :


vkarpov15:
I don't think this is the approach we want to take to support #13288. Because this change as written would allow defining documents where the document interface says `string` but the schema says `ObjectId` even if you don't use `mongoose.ObjectId.get(v => v != null ? v.toString() : v);`

So while this change does unblock your use case, I also think this introduces behavior that many would consider a bug. I'll look for alternatives.

vkarpov15:
Closing as per comments in #13288

PR #13258, opened 2023-04-07
feat(model): added throwOnValidationError option for bulkWrite and insertMany

lpizzinidev:
**Summary**
Added `throwOnValidationError` option for `Model.bulkWrite` and `Model.insertMany` (default to `false`).
If set to `true` and `ordered: true` the methods will throw an error if there were operations that failed validation, but all operations that passed validation succeeded.

**Example**
```
const createdUser = await User.create({ name: 'Test' });

// Will throw an error
await User.bulkWrite(
  [
    {
      updateOne: {
        filter: { _id: createdUser._id },
        update: { $set: { age: 'NaN' } },
        upsert: true,
      },
    },
    {
      updateOne: {
        filter: { _id: createdUser._id },
        update: { $set: { age: 13 } },
        upsert: true,
      },
    },
  ],
  { ordered: false, throwOnValidationError: true }
);
```

**Notes**
Let me know what could be a valid error message.
I commented out a test case that was causing the test script to fail.

Closes #13256

PR #13410, opened 2023-05-17
feat(model): add throwOnValidationError option for opting into getting MongooseBulkWriteError if all valid operations succeed in `bulkWrite()` and `insertMany()`

### LINKED ISSUE #13256: throwOnValidationError for `bulkWrite()` and `insertMany()`
### Prerequisites

### 🚀 Feature Proposal

See https://github.com/Automattic/mongoose/pull/13218#issuecomment-1498189965

```javascript
await Model.bulkWrite(ops, { throwOnValidationError: true });

await Model.insertMany(ops, { throwOnValidationError: true });
```

Will opt in to throwing an error if one of the ops failed validation, but all valid ops completed successfully

### Motivation

Currently, `insertMany()` and `bulkWrite()` don't throw an error if there were validation errors, but all valid ops succeeded. This is a bit messy and inconsistent

### Example

See proposal

vkarpov15:
Fix #13256
Expands on #13258 .

**Summary**

#13256 is a bit of a strange edge case: currently, if you use `bulkWrite()` or `insertMany()` with `ordered: false`, and there are validation errors but all valid ops succeed, you don't get an error. That's because we return the raw MongoDB server result, decorated with some extra Mongoose properties. This is very inconsistent, because you would expect to get an error if one of the operations failed validation.

This PR adds a new `MongooseBulkWriteError` class that consolidates some of the patterns we like for reporting bulk write errors: namely, list of `validationErrors` and a `results` array that helps you tell what happened with each individual operation or document you passed in.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3410,7 +3410,7 @@ Model.bulkWrite = async function bulkWrite(ops, options) {
           results[validOpIndexes[i]] = null;
         }
         if (validationErrors.length > 0) {
-          if ('throwOnValidationError' in options && options.throwOnValidationError) {
+          if (options.throwOnValidationError) {
             return reject(new MongooseBulkWriteError(
               validationErrors,
               results,
--- a/lib/model.js
+++ b/lib/model.js
@@ -64,6 +64,7 @@ const setDottedPath = require('./helpers/path/setDottedPath');
 const STATES = require('./connectionstate');
 const util = require('util');
 const utils = require('./utils');
+const MongooseBulkWriteError = require('./error/bulkWriteError');
 
 const VERSION_WHERE = 1;
 const VERSION_INC = 2;
@@ -2962,6 +2963,7 @@ Model.startSession = function() {
  * @param {Boolean} [options.lean=false] if `true`, skips hydrating and validating the documents. This option is useful if you need the extra performance, but Mongoose won't validate the documents before inserting.
  * @param {Number} [options.limit=null] this limits the number of documents being processed (validation/casting) by mongoose in parallel, this does **NOT** send the documents in batches to MongoDB. Use this option if you're processing a large number of documents and your app is running out of memory.
  * @param {String|Object|Array} [options.populate=null] populates the result documents. This option is a no-op if `rawResult` is set.
+ * @param {Boolean} [options.throwOnValidationError=false] If true and `ordered: false`, throw an error if one of the operations failed validation, but all valid operations completed successfully.
  * @return {Promise} resolving to the raw result from the MongoDB driver if `options.rawResult` was `true`, or the documents that passed validation, otherwise
  * @api public
  */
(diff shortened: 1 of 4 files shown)


lib/model.js
```diff
@@ -2962,6 +2963,7 @@ Model.startSession = function() {
  * @param {Boolean} [options.lean=false] if `true`, skips hydrating and validating the documents. This option is useful if you need the extra performance, but Mongoose won't validate the documents before inserting.
  * @param {Number} [options.limit=null] this limits the number of documents being processed (validation/casting) by mongoose in parallel, this does **NOT** send the documents in batches to MongoDB. Use this option if you're processing a large number of documents and your app is running out of memory.
  * @param {String|Object|Array} [options.populate=null] populates the result documents. This option is a no-op if `rawResult` is set.
+ * @param {Boolean} [options.throwOnValidationError=false] If true and `ordered: false`, throw an error if one of the operations failed validation, but all valid operations completed successfully.
```
AbdelrahmanHafez, line 2966: I'd argue that this should default to `true`, since this is the intuitive behavior anyway, and people should opt-out if they want backward-compatible behavior for some reason, then on v8.0 we can remove the option altogether.
What do you think?

test/model.test.js
```diff
@@ -6177,6 +6216,29 @@ describe('Model', function() {
...
+
+      const err = await User.insertMany([
+        new User({ age: 12 }),
+        new User({ age: 12 }),
+        new User({ age: 'NaN' })
+      ], { ordered: false, throwOnValidationError: true })
+        .then(() => null)
+        .catch(err => err);
+
+      assert.ok(err);
```
AbdelrahmanHafez, line 6235: Maybe worth asserting that the valid operations run on the server, and the documents are successfully inserted.

lib/model.js
vkarpov15: The problem is that we have tests that assert on the counterintuitive behavior, and for stability's sake I like to avoid changing existing tests outside of major releases.

lib/model.js
```diff
@@ -3385,9 +3406,22 @@ Model.bulkWrite = async function bulkWrite(ops, options) {
           return reject(error);
         }
 
+        for (let i = 0; i < validOpIndexes.length; ++i) {
+          results[validOpIndexes[i]] = null;
+        }
         if (validationErrors.length > 0) {
-          res.mongoose = res.mongoose || {};
-          res.mongoose.validationErrors = validationErrors;
+          if ('throwOnValidationError' in options && options.throwOnValidationError) {
```
hasezoey, line 3413: wouldnt just the second part be enough here?

lib/model.js
AbdelrahmanHafez: In that case I think it's good to merge as is, and remove the option in v8.0.

What do you think? Or would you rather to keep it as an option in v8, and maybe just change the default?

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

PR #13422, opened 2023-05-20
fix(array): track correct changes when setting nested array of primitives

### LINKED ISSUE #13372: Error when trying to update primitives at the bottom of multidimensional arrays
### Prerequisites

### Mongoose version

7.1.0

### Node.js version

17.9.1

### MongoDB server version

4.4.21

5.0.4

### Description

When attempting to modify a multidimensional array of primitives  (e.g. `[[Number]]`), changing the values and attempting a save causes an error:

```
The positional operator did not find the match needed from the query.
MongoServerError: The positional operator did not find the match needed from the query
```

### Steps to Reproduce

```ts
const RecentDataPointSchema = new Schema({
    date: Date,
    value: Number,
})
const  TestingMultiDimArraysModel =  mongoose.model("TestingMultiDimArrays ", new Schema( {
    intArray: [[Number]],
    strArray: [[String]],
    objArray:[[RecentDataPointSchema]]
}));
let testObj = new TestingMultiDimArraysModel({
    intArray: [[1,2], [3,4]],
    strArray: [["a", "b"], ["c", "d"]],
    objArray: [
        [
            {date: new Date(Date.now()), value: 1},
            {date: new Date(Date.now()), value: 2},
        ],
        [
            {date: new Date(Date.now()), value: 3},
            {date: new Date(Date.now()), value: 4},
        ]
    ],
})
await testObj.save();
testObj = await TestingMultiDimArraysModel.findById(testObj._id);

//works
//testObj.getChanges() returns: {"$set":{"objArray.0.0.value":3}}
testObj.objArray[0][0].value = 3;
await testObj.save();
testObj = await TestingMultiDimArraysModel.findById(testObj._id);
expect(testObj.objArray[0][0].value).toBe(3);

//works
//testObj.getChanges() returns {"$set":{"objArray.0.0":{"date":"2023-05-03T00:25:28.100Z","value":3,"_id":"6451a9f83e792f2678d175e4"}}}
//@ts-ignore
testObj.objArray[0][0] = {date: new Date(Date.now()), value: 3};
await testObj.save();
testObj = await TestingMultiDimArraysModel.findById(testObj._id);
expect(testObj.objArray[0][0].value).toBe(3);

//works
//testObj.getChanges() returns: {"$set":{"strArray.0":["j","d"]},"$inc":{"__v":1}}
//@ts-ignore
testObj.strArray[0] = ["j", "d"];
await testObj.save();
testObj = await TestingMultiDimArraysModel.findById(testObj._id);
expect(testObj.strArray[0][0]).toBe("j");

//works
//testObj.getChanges() returns: {"$set":{"intArray.0":[9,9]},"$inc":{"__v":1}}
//@ts-ignore
testObj.intArray[0] = [9, 9];
await testObj.save();
testObj = await TestingMultiDimArraysModel.findById(testObj._id);
expect(testObj.intArray[0][0]).toBe(9);

//fails
//testObj.getChanges() returns: {"$set":{"intArray.$.0":[[null,null],[null,null]]},"$inc":{"__v":1}}
//Upon trying to save, the error message below is produced:
//    The positional operator did not find the match needed from the query.
//    MongoServerError: The positional operator did not find the match needed from the query
testObj.intArray[0][0] = 9;
await testObj.save();
testObj = await TestingMultiDimArraysModel.findById(testObj._id);
expect(testObj.intArray[0][0]).toBe(9);


//fails
//testObj.getChanges() returns: {"$set":{"strArray.$.0":[[null,null],[null,null]]},"$inc":{"__v":1}}
//Upon trying to save, the error message below is produced:
//    The positional operator did not find the match needed from the query.
//    MongoServerError: The positional operator did not find the match needed from the query
testObj.strArray[0][0] = "z";
await testObj.save();
testObj = await TestingMultiDimArraysModel.findById(testObj._id);
expect(testObj.strArray[0][0]).toBe("z");


```

I believe the problem is something to do with how the update command is generated, it seems to loose all sense of what the array is. Looking at the `getChanges()` (which I have documented in the code sample) It seems like it's not even sure where in the array it's trying to set the values and it's also trying to put an entire empty shell of the original object shape into a single slot in the array.

One other note: I was having the same problem in version 5 of mongoose. I upgraded versions hoping that would fix it. I don't have as thorough documentation of what was going wrong in version 5, so I can't be confident that nothing has changed about the bug, but based on this I feel confident saying this wasn't caused by a recent change.

### Expected Behavior

I expect the document to be updated properly and not error out.

### potential fix

I am not an expert on this repo, but my own experimenting showed that changing the arrayPath on the integer array helped substantially. The following code works properly:

```
 testObj.intArray[1][arrayPathSymbol] = "intArray.1"
//@ts-ignore
testObj.intArray[0][arrayPathSymbol] = "intArray.0"
testObj.intArray[0][0] = 9;

let changes = testObj.getChanges();
await testObj.save();
testObj = await TestingMultiDimArraysModel.findById(testObj._id);
expect(testObj.intArray[0][0]).toBe(9);

```
For some reason, the array path is being set to `intArray.$`. On a normal single array the arrayPath defaults to `intArray`. On the 2d object array, the arrayPath defaults to `objArray.0`. But for some reason on a 2d integer array, it's `intArray.$`. I think that it may be set that way for use by other commands like `populated()`, but `_markModified()` is just presuming that the path for querying whether there are any items in the list will be the same as the command to set a single item in the list?

vkarpov15:
Fix #13372

**Summary**

The problem is that, for nested arrays of primitives, the array's internal path is `intArr.$` rather than `intArr.0`. This causes problems when trying to set the path directly, because then Mongoose marks `intArr.$.0` as modified.

**Examples**

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -359,7 +359,11 @@ SchemaArray.prototype.cast = function(value, doc, init, prev, options) {
     options = options || emptyOpts;
 
     let rawValue = utils.isMongooseArray(value) ? value.__array : value;
-    value = MongooseArray(rawValue, options.path || this._arrayPath || this.path, doc, this);
+    let path = options.path || this.path;
+    if (options.arrayPathIndex != null) {
+      path += '.' + options.arrayPathIndex;
+    }
+    value = MongooseArray(rawValue, path, doc, this);
     rawValue = value.__array;
 
     if (init && doc != null && doc.$__ != null && doc.$populated(this.path)) {
--- a/test/types.array.test.js
+++ b/test/types.array.test.js
@@ -1900,7 +1900,7 @@ describe('types array', function() {
     m.Schema.Types.Array.options.castNonArrays = true;
   });
 
-  it('supports setting nested arrays directly (gh-13372)', async function() {
+  it('supports setting nested arrays directly (gh-13372)', function() {
     const Test = db.model('Test', new Schema({ intArr: [[Number]] }));
 
     const intArr = [[1, 2], [3, 4]];
--- a/test/types.array.test.js
+++ b/test/types.array.test.js
@@ -1899,4 +1899,18 @@ describe('types array', function() {
 
     m.Schema.Types.Array.options.castNonArrays = true;
   });
+
+  it('supports setting nested arrays directly (gh-13372)', async function() {
+    const Test = db.model('Test', new Schema({ intArr: [[Number]] }));
+
+    const intArr = [[1, 2], [3, 4]];
+    const doc = Test.hydrate({ intArr });
+
+    doc.intArr[1][1] = 5;
+    assert.deepStrictEqual(doc.getChanges(), {
+      $set: {
+        'intArr.1.1': 5
+      }
+    });
+  });
 });
--- a/test/types.array.test.js
+++ b/test/types.array.test.js
@@ -1906,9 +1906,11 @@ describe('types array', function() {
     const intArr = [[1, 2], [3, 4]];
     const doc = Test.hydrate({ intArr });
 
+    doc.intArr[0][0] = 2;
     doc.intArr[1][1] = 5;
     assert.deepStrictEqual(doc.getChanges(), {
       $set: {
+        'intArr.0.0': 2,
         'intArr.1.1': 5
       }
     });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+7.2.0 / 2023-05-19
+==================
+ * feat: upgrade mongodb -> 5.5.0
+ * feat(document): add flattenObjectIds option to toObject() and toJSON() #13383 #13341
+ * feat(query): add translateAliases option to automatically call translate aliases on query fields #13397 #8678 #7511
+ * feat(schema): propagate toObject and toJSON options to implicitly created schemas #13325
+ * feat(model): add throwOnValidationError option for opting into getting MongooseBulkWriteError if all valid operations succeed in bulkWrite() and insertMany() #13410 #13256
+ * feat(types+mongoose): export MongooseError #13403 #13387 [ramos-ph](https://github.com/ramos-ph)
+
 7.1.2 / 2023-05-18
 ==================
  * fix: set timestamps on single nested subdoc in insertMany() #13416 #13343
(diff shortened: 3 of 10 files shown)


hasezoey (approved):

test/types.array.test.js
```diff
@@ -1899,4 +1899,18 @@ describe('types array', function() {
 
     m.Schema.Types.Array.options.castNonArrays = true;
   });
+
+  it('supports setting nested arrays directly (gh-13372)', async function() {
```
AbdelrahmanHafez, line 1903: Nitpick:
```suggestion
  it('supports setting nested arrays directly (gh-13372)', function() {
```

test/types.array.test.js
```diff
@@ -1899,4 +1899,18 @@ describe('types array', function() {
...
+    const Test = db.model('Test', new Schema({ intArr: [[Number]] }));
+
+    const intArr = [[1, 2], [3, 4]];
+    const doc = Test.hydrate({ intArr });
```
AbdelrahmanHafez, line 1907: I love how this whole test is synchronous and is not calling any external resources. Great work!

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

lib/schema/array.js
```diff
@@ -359,7 +359,11 @@ SchemaArray.prototype.cast = function(value, doc, init, prev, options) {
     options = options || emptyOpts;
 
     let rawValue = utils.isMongooseArray(value) ? value.__array : value;
-    value = MongooseArray(rawValue, options.path || this._arrayPath || this.path, doc, this);
+    let path = options.path || this.path;
+    if (options.arrayPathIndex != null) {
```
Uzlopak, line 363: is arrayPathIndex a string or a numerical value? 

If it is a number, we should actually do Number.isInteger.

test/types.array.test.js
Uzlopak, line 1903: Maybe have "duplicate" test where we check that 0 is properly handled, as expected. Because this is why you do the != null check in array.js

lib/schema/array.js
vkarpov15: Number. I don't think it is necessary to do an extra `isInteger()` check, because `arrayPathIndex` isn't something an end user can set. Mongoose sets that in a `for` loop.

PR #13436, opened 2023-05-24
fix(cursor): handle calling `skipMiddlewareFunction()` in pre('find') middleware with cursors

### LINKED ISSUE #13411: @vkarpov15 `skipMiddlewareFunction` only works for `Query` not `QueryCursor`, here is an example
@vkarpov15 `skipMiddlewareFunction` only works for `Query` not `QueryCursor`, here is an example
``` js
// setup a pre hook for find
schema.pre('find', function (next) {
  next(skipMiddlewareFunction([]));
});
```
``` js
// err is { "args": [[]] }, an instance of skipWrappedFunction
mongooseModel.find().cursor().on('error', function (err) { console.log(err); });
```
The reason is in `lib/cursor/QueryCursor.js`, it called like this
``` js
// here may want to call query._find()
// or wrap this with kareem.createWrapper()
model.hooks.execPre('find', query, (err) => {
  if (err != null) {
    _this._markError(err);
    _this.listeners('error').length > 0 && _this.emit('error', err);
    return;
  }
  // do other things
});
```

_Originally posted by @junmiu in https://github.com/Automattic/mongoose/issues/11927#issuecomment-1548956774_

vkarpov15:
Fix #13411

**Summary**

Currently, `skipMiddlewareFunction()` in `pre('find')` middleware causes an error if you're using cursors. With this change, it is now fine to call `skipMiddlewareFunction()` in pre('find') middleware with cursors with no parameters; or with `null`, `undefined`, or empty array `[]`.

The reason why we don't support passing an array of documents is that it would be tricky to rewrite current cursor logic to use arrays. We may consider this for the future if there's demand.

**Examples**

### DIFF
--- a/lib/cursor/QueryCursor.js
+++ b/lib/cursor/QueryCursor.js
@@ -8,6 +8,7 @@ const MongooseError = require('../error/mongooseError');
 const Readable = require('stream').Readable;
 const eachAsync = require('../helpers/cursor/eachAsync');
 const helpers = require('../queryhelpers');
+const kareem = require('kareem');
 const immediate = require('../helpers/immediate');
 const util = require('util');
 
@@ -39,6 +40,7 @@ function QueryCursor(query, options) {
   Readable.call(this, { autoDestroy: true, objectMode: true });
 
   this.cursor = null;
+  this.skipped = false;
   this.query = query;
   const _this = this;
   const model = query.model;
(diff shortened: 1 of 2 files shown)


IslandRhythms (approved):
Probably gonna have merge conflicts with my PR regarding cursors

PR #13437, opened 2023-05-24
fix(query): handle plus path in projection with findOneAndUpdate()

### LINKED ISSUE #13413: Projection with '+' using findByIdAndUpdate does not return the property
### Prerequisites

### Mongoose version

7.1.1

### Node.js version

18.7.0

### MongoDB server version

5.5.0

5.0.4

### Description

When using `select: false` for a property `x` in a schema, `x` is not returned when using `findByIdAndUpdate` and `projection: '+x'`.

Only started occurring in 7.1.1, I don't see this issue <= 7.1.0.

### Steps to Reproduce

In a schema:

```
x: {
  type: String,
  select: false,
},
```

Then try to do:

```
const response = await XModel.findByIdAndUpdate(
  id,
  { $set: { y: y } },
  { projection: '+x' },
);
```

`response` does not contain `x`.

### Expected Behavior

The property is returned.

vkarpov15:
Fix #13413

**Summary**

Looks like we double applied paths in `findOneAndUpdate()`. Likely due to merge conflict issues with 6.x. Cleaned that up and added a test, so shouldn't happen again.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -3289,19 +3289,15 @@ Query.prototype._findOneAndUpdate = async function _findOneAndUpdate() {
     throw this.error();
   }
 
-  this._applyPaths();
-  this._fields = this._castFields(this._fields);
   applyGlobalMaxTimeMS(this.options, this.model);
   applyGlobalDiskUse(this.options, this.model);
 
-  const opts = this._optionsForExec(this.model);
-  this._applyTranslateAliases(opts);
-
-  if ('strict' in opts) {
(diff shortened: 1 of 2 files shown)


IslandRhythms (approved):
LGTM

AbdelrahmanHafez:
Does `+` represent anything special here? Or are we trying to be typo-tolerant?

vkarpov15:
Yep "+" means "this field is excluded by default, don't exclude it for this operation".

Issue #7156, opened 2018-10-20
Plugin: Luxon type


vkarpov15:
Re: #5110

Uzlopak:
Moment.js is deprecated.
https://momentjs.com/docs/#/-project-status/

So we should not implement a moment type anymore.

adamreisnz:
Agreed! A Luxon DateTime type might be a good alternative though.

rkreienbuehl:
Is it anyhow possible to set a default getter for Date type like ist is possible for ObjectId?
`mongoose.Schema.Types.ObjectId.get((v) => (v ? v.toString() : undefined));`
With Date this doesn't work. If this would be possible, there would be no need for a Luxon DateTime plugin.

vkarpov15:
@rkreienbuehl  Yeah there is. `mongoose.Schema.Types.Date.get((v) => DateTime.fromJSDate(v));` should work.

rkreienbuehl:
@vkarpov15 Thanks for your answer. I tried it like this before, but when I do this, I get the following error:
```ts
TypeError: mongoose_1.default.Schema.Types.Date.get is not a function
```

rkreienbuehl:
@vkarpov15 Tested it right now. I had `v7.0.3` and now I updated to `v7.1.1`, but still have this error.

vkarpov15:
@rkreienbuehl oops, it looks like some of our schematypes don't have a `.get()` function. We'll add that for a future minor release.

In the meantime, you can do the following:

```javascript
'use strict';

const mongoose = require('mongoose');
const {DateTime} = require('luxon');

mongoose.SchemaType.get.call(mongoose.Date, (v) => DateTime.fromJSDate(v));

const Test = mongoose.model('Test', mongoose.Schema({ date: Date }));
const doc = new Test({ date: new Date() });
console.log(doc.date instanceof DateTime); // true
```

rkreienbuehl:
> 

@vkarpov15 Thanks for your workaround. Had to use `mongoose.SchemaType.get.call(mongoose.Schema.Types.Date, (v) => DateTime.fromJSDate(v));` instead of `mongoose.SchemaType.get.call(mongoose.Date, (v) => DateTime.fromJSDate(v));` for TypeScript, but then it works perfectly.

PR #13438, opened 2023-05-25
feat(schema): consistently add `.get()` function to all SchemaType classes

vkarpov15:
Re: #7156

**Summary**

Looks like some schematypes are missing the `get()` function to register a getter for all instances of the schematype. Only numbers, strings, and ObjectIds currently have that function. This came up in #7156 in the context of dates.

Not urgent because there is an easy workaround, but something we should ship for 7.3.

**Examples**

### DIFF
--- a/lib/schema/bigint.js
+++ b/lib/schema/bigint.js
@@ -63,6 +63,23 @@ SchemaBigInt._cast = castBigInt;
 SchemaBigInt.set = SchemaType.set;
 
 /**
+ * Attaches a getter for all BigInt instances
+ *
+ * #### Example:
+ *
+ *     // Convert bigints to numbers
+ *     mongoose.Schema.BigInt.get(v => v == null ? number : Number(v));
+ *
+ * @param {Function} getter
+ * @return {this}
+ * @function get
+ * @static
+ * @api public
+ */
+
+SchemaBigInt.get = SchemaType.get;
+
+/**
  * Get/set the function used to cast arbitrary values to booleans.
  *
  * #### Example:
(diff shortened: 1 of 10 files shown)


AbdelrahmanHafez (commented):
Maybe add a test
```js
[...schemaTypes].forEach(schemaType=> {
  it(`${schemaType.name} has a \`get\` method`, ()=> {
    assert.strictEqual(typeof schemaType.get, 'function');
  });
})
```

lib/schema/date.js
```diff
@@ -70,6 +70,26 @@ SchemaDate._cast = castDate;
 
 SchemaDate.set = SchemaType.set;
 
+/**
+ * Attaches a getter for all ObjectId instances
```
hasezoey, line 74:
```suggestion
 * Attaches a getter for all Date instances
```

lib/schema/date.js
```diff
@@ -70,6 +70,26 @@ SchemaDate._cast = castDate;
...
+ *
+ * #### Example:
+ *
+ *     // Always convert to string when getting an ObjectId
```
hasezoey, line 78:
```suggestion
 *     // Always convert Dates to string
```

lib/schema/bigint.js
```diff
@@ -62,6 +62,23 @@ SchemaBigInt._cast = castBigInt;
 
 SchemaBigInt.set = SchemaType.set;
 
+/**
+ * Attaches a getter for all BigInt instances
+ *
+ * #### Example:
+ *
+ *     // Convert bigints to numbers
+ *     mongoose.Schema.BigInt.get(v => v == null ? number : Number(v));
```
hasezoey, line 71:
```suggestion
 *     mongoose.Schema.BigInt.get(v => v == null ? v : Number(v));
```

hasezoey (changes requested):
looks good, though some examples need adjustment

hasezoey (approved):

PR #13449, opened 2023-05-28
docs(connections): add section on multi tenant

### LINKED ISSUE #11187: Improved `useDb()` docs
Right now all we have is API docs: https://mongoosejs.com/docs/api/connection.html#connection_Connection-useDb . Should have something more substantial for using `useDb()`, especially for multi tenant

vkarpov15:
Fix #11187

**Summary**

We get lots of questions about multi tenant, so would be nice to have a section of the docs to point people to.

**Examples**

### DIFF
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -31,6 +31,7 @@ See the [mongodb connection string spec](http://www.mongodb.com/docs/manual/refe
   <li><a href="#mongos_connections">Multi-mongos support</a></li>
   <li><a href="#multiple_connections">Multiple connections</a></li>
   <li><a href="#connection_pools">Connection Pools</a></li>
+  <li><a href="#multi-tenant-connections">Multi Tenant Connections</a></li>
 </ul>
 
 <h2 id="buffering"><a href="#buffering">Operation Buffering</a></h2>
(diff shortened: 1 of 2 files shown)


docs/connections.md
```diff
@@ -449,6 +450,91 @@ const uri = 'mongodb://127.0.0.1:27017/test?maxPoolSize=10';
...
+Pattern (2) is more flexible and better for use cases with > 10k tenants and > 1 requests/second.
+Because each tenant has a separate connection pool, one tenants' slow operations will have minimal impact on other tenants.
+However, this pattern is harder to implement and manage in production.
+In particular, [MongoDB does have a limit on the number of open connections](https://www.mongodb.com/blog/post/tuning-mongodb--linux-to-allow-for-tens-of-thousands-connections), and [MongoDB Atlas has separate limits on the number of open connections](https://www.mongodb.com/docs/atlas/reference/atlas-limits), so you need to make sure the total number of sockets in your connection pools doesn't go over MongoDB's limits.
+
+```javascript
+const express = require('express');
+const mongoose = require('mongoose');
+
+mongoose.connect('mongodb://127.0.0.1:27017/main');
```
hasezoey, line 514: is this still required here?

lib/connection.js
```diff
@@ -1582,7 +1582,7 @@ Connection.prototype.syncIndexes = async function syncIndexes(options = {}) {
 };
 
 /**
- * Switches to a different database using the same connection pool.
+ * Switches to a different database using the same [connection pool](../connections.html#connection_pools).
```
hasezoey, line 1585:
```suggestion
 * Switches to a different database using the same [connection pool](https://mongoosejs.com/docs/api/connectionshtml#connection_pools).
```

since #13328 in code we use absolute paths so that the URLs are valid both in-editor and in the website

hasezoey (approved):
Looks good to me, though one question and one suggestion

docs/connections.md
vkarpov15: Yes, to highlight that you do still need to connect before using `useDb()`.

hasezoey: from what i can tell, this code example does not use `useDb`, and rather `createConnection` and never the default connection itself

vkarpov15: That's a fair point, this `mongoose.connect()` call could be confusing. I put it here because in a multi-tenant scenario you may use the default connection to look up tenants. But not necessarily, so we can remove that.

PR #13450, opened 2023-05-28
fix(update): allow setting paths with dots under non-strict paths

### LINKED ISSUE #13434: Model.findOneAndUpdate() is updating nested fields with 'strict:false'
### Discussed in https://github.com/Automattic/mongoose/discussions/13432

<sup>Originally posted by **amitpriyankar** May 24, 2023</sup>
I have observed an unusual behavior when using `Model.findOneAndUpdate` with a collection that has a flexible schema (strict: false). When I attempt to insert a document that includes a key containing a dot (.), the behavior differs based on whether I define the type of the key in the schema. If I specify the type, the key is treated as a regular string. However, if I don't specify the type, nested fields are created in the document.

Please refer the following code:
```javascript

const url = '<connection_url>;

main().catch((err) => console.log(err));

async function main() {
    await mongoose.connect(url);

    const personSchema = new mongoose.Schema(
        {
            name: String,
            // info: Object,
        },
        {strict: false}
    );

    personSchema.index({name: 1}, {unique: true});

    const savedDoc = await Person.findOneAndUpdate(
        {},
        {name: 'Tommy1', info: {'second.name': 'Shelby'}},
        {new: true, upsert: true}
    )
        .lean()
        .exec();

    console.log('Saved document: ', savedDoc);
    mongoose.disconnect();
}
```
```
Saved document:  {
  _id: new ObjectId("646dccd30b5d3da01e058fd0"),
  __v: 0,
  info: { second: { name: 'Shelby' } },
  name: 'Tommy1'
}
```

I check the doc but didn't find anything which explains this behavior. Could someone please help?

mongoose: v7.1.1
MongoDB Server: 4.0</div>

vkarpov15:
Fix #13434
Re: #10200

**Summary**

One of @AbdelrahmanHafez 's early issues #10200 pops up :) we shouldn't apply `setDottedPaths()` if schema is not strict and we don't have a schematype, because MongoDB supports storing paths with dots now.

**Examples**

### DIFF
--- a/lib/helpers/query/castUpdate.js
+++ b/lib/helpers/query/castUpdate.js
@@ -360,7 +360,7 @@ function walkUpdatePath(schema, obj, op, options, context, filter, pref) {
         try {
           if (prefix.length === 0 || key.indexOf('.') === -1) {
             obj[key] = castUpdateVal(schematype, val, op, key, context, prefix + key);
-          } else {
+          } else if (isStrict !== false || schematype != null) {
             // Setting a nested dotted path that's in the schema. We don't allow paths with '.' in
             // a schema, so replace the dotted path with a nested object to avoid ending up with
             // dotted properties in the updated object. See (gh-10200)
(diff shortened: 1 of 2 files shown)


hasezoey (approved):

AbdelrahmanHafez (approved):
LGTM, thanks!
One question though. How's this going to affect `Mixed` paths? Do we still split keys or this change would make it so that paths are now dot-separated?

PR #13471, opened 2023-06-02
docs: add basic field level encryption docs

### LINKED ISSUE #13412: Mongoose Middleware and MongoDB CSFLE Explicit Encryption / Automatic Decryption
### Prerequisites

### Mongoose version

6.5.2

### Node.js version

18

### MongoDB version

5.0.17

### Operating system

macOS

_No response_

### Issue

Using the MongoDB Community Edition, and its CSFLE Explicit Encryption and  Automatic Decryption,  I am trying to figure out the best way to insert the ClientEncryption.encrypt() in the Mongoose “save” process.

Let’s say we want to encrypt the `{ lastName: String }` field on save.

On one hand, the encrypted data must be saved on MongoDB as BinData (to benefit from automatic decryption for MongoDB), on the other: we would like to keep lastName defined as String in the Mongoose as schema.

Defining a field as String in the Mongoose schema forces the casting to String when saving the lastName (which would prevent the automatic decryption).
The `pre(‘save’)` doesn’t seem to allow to recast the String to BinData.

To get it to work, we had to change the lastName field’s SchemaType from String to Mixed.

Here is the question: is there a way to keep the SchemaType as String, and invoke the ClientEncryption.encrypt() and recast the String to BinData just before Mongoose makes the call to the MongoDB?
Maybe using a specific hook in the Mongoose middleware?

Thank you for your suggestion.

vkarpov15:
Fix #13412

**Summary**

Got another question about CSFLE, figured we should write up a quick doc so we don't have to dig through GitHub issues every time. This doc encourages using `schemaMap` rather than collection-level validators.

@alexbevi can you take a look at this too please?

**Examples**

### DIFF
--- a/docs/field-level-encryption.md
+++ b/docs/field-level-encryption.md
@@ -1,0 +1,112 @@
+# Integrating with MongoDB [Client Side Field Level Encryption](https://www.mongodb.com/docs/manual/core/csfle/)
+
+Client Side Field Level Encryption, or CSFLE for short, is a tool for storing your data in an encrypted format in MongoDB.
+For example, instead of storing the `name` property as a plain-text string, CSFLE means MongoDB will store your document with `name` as an encrypted buffer.
+The resulting document will look similar to the following to a client that doesn't have access to decrypt the data.
+
+```
+{
+  "_id" : ObjectId("647a3207661e3a3a1bc3e614"),
+  "name" : BinData(6,"ASrIv7XfokKwiCUJEjckOdgCG+u6IqavcOWX8hINz29MLvcKDZ4nnjCnPFZG+0ftVxMdWgzu6Vdh7ys1uIK1WiaPN0SqpmmtL2rPoqT9gfhADpGDmI60+vm0bJepXNY1Gv0="),
+  "__v" : 0
+}
+```
+
+You can read more about CSFLE on the [MongoDB CSFLE documentation](https://www.mongodb.com/docs/manual/core/csfle/) and [this blog post about CSFLE in Node.js](https://www.mongodb.com/developer/languages/javascript/client-side-field-level-encryption-csfle-mongodb-node/).
+
+Note that Mongoose does **not** currently have any Mongoose-specific APIs for CSFLE.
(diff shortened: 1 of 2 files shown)


IslandRhythms (approved):
LGTM

hasezoey (approved):

huynhdev24 (approved):

alexbevi:
@vkarpov15 I haven't had a chance to revisit this yet but since multiple approvals have been submitted we can merge this as-is and just improve it later to add auto-encryption

docs/field-level-encryption.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+You also need to make sure you've installed [mongocryptd](https://www.mongodb.com/docs/manual/core/queryable-encryption/reference/mongocryptd/).
+mongocryptd is a separate process from the MongoDB server that you need to run to work with field level encryption.
+You can either run mongocryptd yourself, or make sure it is on the system PATH and the MongoDB Node.js driver will run it for you.
+[You can read more about mongocryptd here](https://www.mongodb.com/docs/v5.0/reference/security-client-side-encryption-appendix/#mongocryptd).
+
+Once you've set up and run mongocryptd, first you need to create a new encryption key as follows.
+Keep in mind that the following example is a simple example to help you get started.
+The encryption key in the following example is insecure; MongoDB recommends using a [KMS](https://www.mongodb.com/docs/v5.0/core/security-client-side-encryption-key-management/).
```
AbdelrahmanHafez, line 37: Do we need to do that if we're using Atlas? I anticipate Atlas will have a simpler solution.
Can we have a comment/section explaining what people using Atlas need to do?

AbdelrahmanHafez (approved):
One comment/question, otherwise LGTM. Thanks! 👍

docs/field-level-encryption.md
vkarpov15: That's a good point, I'll have to ask the MongoDB team on that front. I have never run CSFLE in production, so I don't know the answer to this right now.

AbdelrahmanHafez: In that case we can merge this as is and add a change if needed when we get a response. This is already an improvement to the docs.

PR #13472, opened 2023-06-02
Model.create() ordered option

### LINKED ISSUE #4038: Possibility to run create() sequentially
as @vkarpov15 commented here https://github.com/Automattic/mongoose/issues/723#issuecomment-204817852 the `create()` method runs a bunch of `save()` calls in parallel.

It would be useful to implement some kind of callback or sequential promises to allow keeping the order for example in a number sequence like in the following example:

``` js
OrderSchema.pre('save', function (next) {
  if (!this.isNew) {
    return next();
  }

  CounterModel.findOneAndUpdate({_id: 'orders'}, {$inc: {n: 1}}, {'new': true, upsert: true}, (err, doc) => {
    if (err) {
      next(err);
    } else {
      this.reference = doc.n;
      next();
    }
  });
});
```

**Output**:

``` js
[ { reference: '13' },
  { reference: '21' },
  { reference: '16' },
  { reference: '5' },
  { reference: '9' },
  { reference: '19' },
  { reference: '2' },
  { reference: '10' },
  { reference: '11' },
  { reference: '20' },
  { reference: '8' },
  { reference: '18' },
  { reference: '3' },
  { reference: '15' },
  { reference: '7' },
  { reference: '14' },
  { reference: '4' },
  { reference: '17' },
  { reference: '22' },
  { reference: '6' },
  { reference: '1' },
  { reference: '12' } ]
```

My current use case, is when a user imports a file into my app, for creating a bunch of documents, the reference number goes crazy, and I would expect them to keep the order.

I know this could be easily implemented by iterating myself and running `save()` however It could be an option in the `create()` method and probably more performant :+1:

IslandRhythms:
closes #4038

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -479,6 +479,7 @@ function generateVersionError(doc, modifiedPaths) {
  *     newProduct === product; // true
  *
  * @param {Object} [options] options optional options
+ * @param {Boolean} [options.ordered] saves the docs in series rather than parallel.
  * @param {Session} [options.session=null] the [session](https://www.mongodb.com/docs/manual/reference/server-sessions/) associated with this save operation. If not specified, defaults to the [document's associated session](https://mongoosejs.com/docs/api/document.html#Document.prototype.session()).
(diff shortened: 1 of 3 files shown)


IslandRhythms:
closes #4038

vkarpov15 (approved):

PR #13474, opened 2023-06-03
docs(populate): add a little more info on why we recommend using ObjectId for `_id`

### LINKED ISSUE #13400: Ref type docs missing
### Prerequisites

### Mongoose version

7.1.1

### Node.js version

16.17.0

### MongoDB version

6.0.5

### Operating system

Linux

20.04

### Issue

What are the implications of types if set on a ref, is missing?
Location: https://github.com/Automattic/mongoose/blob/ecb0249545670678c3d215f31bc0f2f85280559f/docs/populate.md?plain=1#L36

vkarpov15:
Fix #13400

**Summary**

Re: #13400, populate docs don't really explain why we recommend using ObjectId for `_id` and `ref`. Added a note about that, and moved the note to the "Saving Refs" section.

**Examples**

### DIFF
--- a/docs/populate.md
+++ b/docs/populate.md
@@ -31,10 +31,6 @@ what tells Mongoose which model to use during population, in our case
 the `Story` model. All `_id`s we store here must be document `_id`s from
 the `Story` model.
 
-**Note**: `ObjectId`, `Number`, `String`, and `Buffer` are valid for use
-as refs. However, you should use `ObjectId` unless you are an advanced
-user and have a good reason for doing so.
-
 <ul class="toc">
   <li><a href="#saving-refs">Saving Refs</a></li>
   <li><a href="#population">Population</a></li>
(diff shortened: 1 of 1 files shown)


docs/populate.md
```diff
@@ -72,21 +68,22 @@ const author = new Person({
...
 });
+
+await author.save();
+// that's it!
 ```
 
+You can set the `ref` option on `ObjectId`, `Number`, `String`, and `Buffer` paths.
+`populate()` works with ObjectIds, numbers, strings, and buffers.
+However, we recommend using ObjectIds as `_id` properties (and thus ObjectIds for `ref` properties) unless you have a good reason not to.
+That is because MongoDB will set `_id` to an ObjectId if you create a new document without an `_id` property, so if you make your `_id` property a Number, you need to be extra careful not to insert a document without a numeric `_id`.
```
hasezoey, line 85: maybe mention that mongoose can help with that via validators, like using `required: true` and for non-default types to have to use `default: () => newType` to generate a new id (via a function instead of statically)

hasezoey (approved):

AbdelrahmanHafez (approved):

vkarpov15:
@BioSurienDG I'm not sure what your question has to do with this PR? I'd generally recommend using `mongoose.Types.ObjectId` in TypeScript; but it doesn't really matter unless you're using the `_id` getter.

BioSurienDG:
> @BioSurienDG I'm not sure what your question has to do with this PR? I'd generally recommend using `mongoose.Types.ObjectId` in TypeScript; but it doesn't really matter unless you're using the `_id` getter.

Ya no worries. I had  thought it was documentation issue that could be resolved in this but was actually bug - pls disregard my comment on the PR 🙂

PR #13475, opened 2023-06-04
docs(connection): clarify how `Connection.prototype.destroy()` is different from `close()`

vkarpov15:
**Summary**

Current docs for `destroy()` don't do much to explain why `destroy()` is different from `close()`.

**Examples**


hasezoey (approved):

AbdelrahmanHafez (approved):
LGTM, thanks. 👍

PR #13477, opened 2023-06-04
feat(populate): pass virtual to `match` function to allow merging match options

### LINKED ISSUE #12443: Add option to merge rather than overwrite schema-level `populate()` options
### Prerequisites

### Issue

hi, i have two schemas with below descriptions

**User Schema**
````typescript
class User {
    @Prop() _id: number;
    @Prop() events: object; // it has two keys : create and delete
}
````

**Custom Schema**
````typescript
class Custom {
    @Prop() _id: number;
    @Prop() user_id: number
}

CustomSchema.virtual('user' , {
  ref: 'user',
  localField: 'user_id',
  foreignField: '_id',
  options: {
      match: {
          'events.delete': null
      }
  }
})
````

custom schema has relation with user which is defined via a virtual.

# How to produce problem
run this query :
````
custom.findOne({ _id:1 }).populate({ path: 'user, match: {'events.create': {$ne : null } } }).exec();
````

# Expected behavior
populate user if ```events.delete``` is null and ```events.create``` is not null

# Result
populate user if ``` events.create ``` is not null

# What is the problem?
match defined in populate overwrites match defined in virtual.
how to append match defined in populate with match defined in virtual?

thanks.

vkarpov15:
Fix #12443

**Summary**

#12443 asks for a way to merge `match` options when populated rather than overwriting. Instead of adding yet another new option, I added a new parameter to `match()` function: the virtual definition. That makes it easy to append the virtual `match` to any custom match you want to add.

**Examples**

### DIFF
--- a/docs/populate.md
+++ b/docs/populate.md
@@ -727,6 +727,29 @@ AuthorSchema.virtual('posts', {
 });
 ```
 
+You can overwrite the `match` option when calling `populate()` as follows.
+
+```javascript
+// Overwrite the `match` option specified in `AuthorSchema.virtual()` for this
+// single `populate()` call.
+await Author.findOne().populate({ path: posts, match: {} });
+```
+
+You can also set the `match` option to a function in your `populate()` call.
+If you want to merge your `populate()` match option, rather than overwriting, use the following.
+
+```javascript
+await Author.findOne().populate({
+  path: posts,
+  // Add `isDeleted: false` to the virtual's default `match`, so the `match`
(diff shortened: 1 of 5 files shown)


hasezoey (approved):
LGTM, maybe consider changing the documentation to reflect that it is now possible (or what changed?) to [Populate Virtuals: The Match Option](https://mongoosejs.com/docs/populate.html#match)

also [Schema.prototype.virtual](https://mongoosejs.com/docs/api/schema.html#Schema.prototype.virtual()) does not list `match` as a possible option

Issue #13487, opened 2023-06-07
Model.create fails with `Cannot read property 'length' of undefined` if there is a second argument with value null or undefined.


robincsamuel:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.2.2

### Node.js version

14.20.1

### MongoDB server version

5.0.7

### Description

The function throws the following error when executing `Model.create` with a second argument having value `null` or `undefined`,

```
TypeError: Cannot read property 'length' of undefined
    at Function.create (/Users/robin/Projects/sample/node_modules/mongoose/lib/model.js:2825:14)
```

**It works fine on version `6.x`.**

### Steps to Reproduce

```
const User = model(
  'User',
  new Schema({
    name: String,
  }),
);

const createUser = (payload, options=undefined) => {
   return User.create(payload, options)
}

await createUser({name: "Robin"})

```

### Expected Behavior

The entry should get created. The options parameter should be ignored since it's `undefined`.be ignored

vkarpov15:
We'll fix the unexpected error in #13493, but please be aware that your syntax for `create()` is wrong. Calling `create()` with 2 object arguments creates 2 documents, Mongoose does **not** treat the 2nd arg as `options` **unless** the first arg is an array.

Please use the following instead:

```javascript
const createUser = (payload, options=undefined) => {
   return User.create([payload], options) // <-- array around `payload` to make Mongoose use 2nd arg as options
}
```

robincsamuel:
Great! Ah, thanks for pointing that out. I am actually passing an array but messed up when writing the example.On 09-Jun-2023, at 11:11 PM, Valeri Karpov ***@***.***> wrote:﻿
We'll fix the unexpected error in #13493, but please be aware that your syntax for create() is wrong. Calling create() with 2 object arguments creates 2 documents, Mongoose does not treat the 2nd arg as options unless the first arg is an array.
Please use the following instead:
const createUser = (payload, options=undefined) => {
   return User.create([payload], options) // <-- array around `payload` to make Mongoose use 2nd arg as options
}

PR #13491, opened 2023-06-09
fix Model.create failing when the last argument is falsy

MohOraby:
**Summary**

Related to  #13487 

You shouldn't be able to pass falsy values to `Model.create()` after doc because it will create a document with empty values (fails schema validation later on)

Added better error handling

The code only checks if `options(second argument)` and third arguments are callbacks  and doesn't take `last` into consideration 

**Examples**

 cases like 

`
const user = await User.create(
  { name: 'user 1' },
  { name: 'user 2' },
  { name: 'user 3' },
  () => console.log('test'),
  { name: 'not created' }
)
`
would pass this validation (fails later on)

and a case like `const user = await User.create(() => console.log('test'))` throws the same `cannot read length of undefined` error instead of throwing no callbacks error

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2796,9 +2796,15 @@ Model.findByIdAndRemove = function(id, options) {
  */
 
 Model.create = async function create(doc, options) {
-  if (typeof options === 'function' ||
-      typeof arguments[2] === 'function') {
+  const argumentsArray = Array.from(arguments);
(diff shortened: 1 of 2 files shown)


vkarpov15:
Sorry we didn't see this PR, we put in an alternative fix #13493 that doesn't throw an error when you do `Model.create(doc, undefined)`. Will close this PR for now.

PR #13499, opened 2023-06-12
7.3

vkarpov15:
**Summary**

**Examples**


types/query.d.ts
```diff
@@ -622,7 +622,15 @@ declare module 'mongoose' {
...
-    select(arg: string | string[] | Record<string, number | boolean | object>): this;
+    select<Paths extends { [P in keyof ResultType]?: any } = {}>(
+      arg: string | string[] | Record<string, number | boolean | object>
+    ): QueryWithHelpers<
+      UnpackedIntersection<{}, Paths>,
+      DocType,
+      THelpers,
+      UnpackedIntersection<{}, Paths>,
+      QueryOp
+    >;
```
hasezoey, line 633: this does not seem to give the proper type when testing against typegoose 11.2.0 (current latest for mongoose 7.2.3)

error:
```txt
test/tests/indexes.test.ts:25:108 - error TS2339: Property 'toObject' does not exist on type '{}'.

25     const fSExtra = (await SelectModel.findOne({}).select(['+test1', '+test3', '-test2']).orFail().exec()).toObject();
                                                                                                              ~~~~~~~~
```

code:

```ts
const fSExtra = (await SelectModel.findOne({}).select(['+test1', '+test3', '-test2']).orFail().exec()).toObject();
```
hovering over `.exec()` says it returns a `Promise<{}>`, meaning not a document

likely because generic `ResultType` (first generic) in `QueryWithHelpers` is because it operates on `{}`

changing to the following return fixes the error, though i dont know if they are the proper intended types (similar to what `.populate` returns):
```ts
QueryWithHelpers<
      UnpackedIntersection<ResultType, Paths>,
      DocType,
      THelpers,
      UnpackedIntersection<RawDocType, Paths>,
      QueryOp
    >;
```

hasezoey (changes requested):
Aside from the one types issue, this release seems to be work just fine with current latest typegoose (both tsc build and runtime & types tests)

types/query.d.ts
vkarpov15: @hasezoey I think I fixed this in https://github.com/Automattic/mongoose/pull/13499/commits/30e7833c12a1193d41d768113aad00d69d2982c2, can you please try against the latest commit on this branch?

hasezoey (approved):
current version of e8987d0d046c0b013fb7e6a57b399e244aff8564 runs file (both in types and runtime) in typegoose

PR #13513, opened 2023-06-15
feat(schema): add collectionOptions option to schemas

vkarpov15:
**Summary**

MongoDB has added [many options to `createCollection()`](https://www.mongodb.com/docs/manual/reference/method/db.createCollection/) and schemas don't quite support all of them. In the interest of making sure developers can use schemas to configure their collections, this PR adds a `collectionOptions` option to schemas; this means you can specify options like `storageEngine` and `validator` in your schema.

**Examples**

### DIFF
--- a/types/schemaoptions.d.ts
+++ b/types/schemaoptions.d.ts
@@ -50,7 +50,7 @@ declare module 'mongoose' {
     collation?: mongodb.CollationOptions;
 
     /** Arbitrary options passed to `createCollection()` */
-    collectionOptions?: Record<string, any>;
+    collectionOptions?: mongodb.CreateCollectionOptions;
 
     /** The timeseries option to use when creating the model's collection. */
     timeseries?: mongodb.TimeSeriesCollectionOptions;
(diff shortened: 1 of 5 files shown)


types/schemaoptions.d.ts
```diff
@@ -49,6 +49,9 @@ declare module 'mongoose' {
     /** Sets a default collation for every query and aggregation. */
     collation?: mongodb.CollationOptions;
 
+    /** Arbitrary options passed to `createCollection()` */
+    collectionOptions?: Record<string, any>;
```
AbdelrahmanHafez, line 53: Wouldn't it be nice if we could use MongoDB types here instead of `Record<string, any>`?

hasezoey, line 53:
```suggestion
    collectionOptions?: mongodb.CreateCollectionOptions;
```

hasezoey (approved):
Code looks good to me, though i agree with hafez to change the type to a mongodb type

AbdelrahmanHafez (approved):

vkarpov15:
Yeah wrong branch. Sorry about that.

PR #13515, opened 2023-06-16
types: augment bson.ObjectId instead of adding on own type

### LINKED ISSUE #12537: BSON ObjectId is not compatible with Types.ObjectId
### Prerequisites

### Mongoose version

6.6.5

### Node.js version

14.19.3

### MongoDB server version

4.4

### Description

I'm trying to upgrade to v6 of mongoose but keep receiving multiple errors related to ObjectId

```ts
          Types of property 'evidence' are incompatible.
            Type 'LeanDocument<ObjectId>[] | undefined' is not assignable to type 'ObjectId[] | undefined'.
              Type 'LeanDocument<ObjectId>[]' is not assignable to type 'ObjectId[]'.
                Type 'LeanDocument<ObjectId>' is missing the following properties from type 'ObjectId': toJSON, equals
```

That was new a error showing up after upgrade in a lot of places, LeanDocument<ObjectId> :thinking: ?
I was expecting it to be original ObjectId, but I'm not sure if there is any way to change that other than updating interface?

Then that lead me to reading up recommend logic for ObjectId at this page: [ObjectIds and Other Mongoose Types](https://mongoosejs.com/docs/typescript.html#objectids-and-other-mongoose-types)

### Suggested ObjectId types
To define a property of type `ObjectId`, you should use `Types.ObjectId` in the TypeScript document interface. You should use `ObjectId` or `Schema.Types.ObjectId` in your schema definition.

### Current codebase setup
Current approach is to use bson ObjectId as interface type and  `Schema.Types.ObjectId` as schema type, therefore this will require to update all interfaces to use `Types.ObjectId` instead of bson ObjectId.

Now this is where the real problem is, seems like Types.ObjectId is not actually compatible with bson ObjectId due to required `_id` property. That would require huge amount of changes to replace `new ObjectId()` with `new Types.ObjectId()`. Is this an overlook? What is the reason Types.ObjectId is not compatible the actual underlying ObjectId class?

```
// Property '_id' is missing in type ('bson").ObjectID' but required in type 'import("mongoose").Types.ObjectId'. 
`node_modules/bson/bson").ObjectID' is not assignable to parameter of type 'import("mongoose").Types.ObjectId'`
```

### Next steps
Determine if this an issue with queries now returning  `LeanDocument<ObjectId>` instead of `ObjectId`? or incompatibility between `Types.ObjectId` and bson `ObjectId`?

I'm surprised this have not yet come up more :thinking: 

Let me know what is the best way to processed with upgrade!

### Steps to Reproduce

```js
import { Schema, Types } from 'mongoose';
import { ObjectId } from 'bson';

// 1. Create an interface representing a document in MongoDB.
interface IUser {
  name: string;
  email: string;
  organizationId: ObjectId;
}

// 2. Create a Schema corresponding to the document interface.
const userSchema = new Schema<IUser>({
  // And `Schema.Types.ObjectId` in the schema definition.
  organizationId: { type: Schema.Types.ObjectId, ref: 'Organization' }
});

const User = model<IUser>('User', userSchema);

const user = new User()
// Property '_id' is missing in type ('bson").ObjectID' but required in type 'import("mongoose").Types.ObjectId'. 
user.organizationId = new ObjectId()
```

### Expected Behavior

`bson` ObjectId should be accepted as Types.ObjectId interface or query should **not** return `LeanDocument<ObjectId>`

hasezoey:
fixes #12537

**Summary**

This PR changes the `Types.ObjectId` to not have its own `_id` declaration anymore but augment `bson.ObjectId` directly

### DIFF
--- a/test/types/mongo.test.ts
+++ b/test/types/mongo.test.ts
@@ -1,4 +1,19 @@
 import * as mongoose from 'mongoose';
 import { expectType } from 'tsd';
+import * as bson from 'bson';
 
 import GridFSBucket = mongoose.mongo.GridFSBucket;
+
+function gh12537() {
+  const schema = new mongoose.Schema({ test: String });
+  const model = mongoose.model('Test', schema);
+
+  const doc = new model({});
+
+  const v = new bson.ObjectId('somehex');
+  expectType<string>(v._id.toHexString());
+
+  doc._id = new bson.ObjectId('somehex');
+}
+
+gh12537();
--- a/types/augmentations.d.ts
+++ b/types/augmentations.d.ts
@@ -1,0 +1,9 @@
+// this import is required so that types get merged instead of completely overwritten
+import 'bson';
+
+declare module 'bson' {
+  interface ObjectId {
+    /** Mongoose automatically adds a conveniency "_id" getter on the base ObjectId class */
+    _id: this;
+  }
+}
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -22,6 +22,7 @@
 /// <reference path="./validation.d.ts" />
 /// <reference path="./inferschematype.d.ts" />
 /// <reference path="./virtuals.d.ts" />
+/// <reference path="./augmentations.d.ts" />
 
 declare class NativeDate extends global.Date { }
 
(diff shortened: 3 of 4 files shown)


vkarpov15 (approved):
LGTM :+1: In the interest of caution, I'll merge this into the 7.4 branch.

PR #13519, opened 2023-06-16
add test case for connection storage scenario

### LINKED ISSUE #11821: add db.removeDb()
We have a multi tenant architecture where each tenant has it's own DB. At some point we had to manage the connections to the DBs to avoid memory leaks. 

This is the clean up function we have:

```js
const handleCleanUp = removedDb => {
    delete db.relatedDbs[removedDb.name]
    db.otherDbs = db.otherDbs.filter(otherDb => removedDb != otherDb)
}
```

It's not much of a problem but it would be interesting to be able to have an officialy supported way of doing this:

```js
const mOtherDb = db.useDb('my-other-db')
db.removeDb('my-other-db')
```

IslandRhythms:
closes #11821

### DIFF
--- a/test/connection.test.js
+++ b/test/connection.test.js
@@ -380,6 +380,18 @@ describe('connections:', function() {
     }
   });
 
+  it('should destroy the connection gh-11821', async function() {
+    const opts = {};
+    const conn1 = await mongoose.createConnection(start.uri, opts).asPromise();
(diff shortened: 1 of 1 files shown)


vkarpov15 (commented):
Thanks for confirming, turns out we do need to implement #11821 after all. This test confirms that `destroy()` on a child connection closes the parent connection, which isn't what we're looking for.

PR #13524, opened 2023-06-19
docs(connections+faq): add info on localhost vs 127.0.0.1

### LINKED ISSUE #13483: Add localhost vs 127.0.0.1 info in FAQ and connections docs
### Prerequisites

### Mongoose version

7.2.2

### Node.js version

18.12.1

### MongoDB version

6.0.4

### Operating system

Windows

11

### Issue

I can connect to mongoDB with `127.0.0.1:27017` and `localhost:27017` both, with the help of mongoDB Compass, when I am trying to connect to the same with the help of mongoose, for `localhost:27017` it is showing error message like, `ECONNREFUSED ::1:27017`, but in case if I try to connect with `127.0.0.1:27017`, it connects.

N.B: I have checked my host mapping for my system, everything is mapped correctly.

vkarpov15:
Fix #13483

**Summary**

Finish up our switch from recommending localhost to recommending 127.0.0.1. Especially important now that Node v16 is nearing EOL, so no more versions of Node that aren't affected by this issue.

**Examples**

### DIFF
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -8,7 +8,7 @@ mongoose.connect('mongodb://127.0.0.1:27017/myapp');
 
 This is the minimum needed to connect the `myapp` database running locally on the default port (27017).
 For local MongoDB databases, we recommend using `127.0.0.1` instead of `localhost`.
-That is because Node.js 18 and up prefer IPv6 addresses, which means, on many machines, Node.js will resolve `localhost` to the IPv6 address `::1` and Mongoose will be unable to connect.
+That is because Node.js 18 and up prefer IPv6 addresses, which means, on many machines, Node.js will resolve `localhost` to the IPv6 address `::1` and Mongoose will be unable to connect, unless the mongodb instance is running with ipv6 enabled.
 
 You can also specify several more parameters in the `uri`:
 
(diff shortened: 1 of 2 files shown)


docs/connections.md
```diff
@@ -6,9 +6,9 @@ You can connect to MongoDB with the `mongoose.connect()` method.
 mongoose.connect('mongodb://127.0.0.1:27017/myapp');
 ```
 
-This is the minimum needed to connect the `myapp` database running locally
-on the default port (27017). If connecting fails on your machine, try using
-`127.0.0.1` instead of `localhost`.
+This is the minimum needed to connect the `myapp` database running locally on the default port (27017).
+For local MongoDB databases, we recommend using `127.0.0.1` instead of `localhost`.
+That is because Node.js 18 and up prefer IPv6 addresses, which means, on many machines, Node.js will resolve `localhost` to the IPv6 address `::1` and Mongoose will be unable to connect.
```
hasezoey, line 11:
```suggestion
That is because Node.js 18 and up prefer IPv6 addresses, which means, on many machines, Node.js will resolve `localhost` to the IPv6 address `::1` and Mongoose will be unable to connect, unless the mongodb instance is running with ipv6 enabled.
```
maybe note that this is only the case if the instance is not running with ipv6 enabled

hasezoey (approved):

Issue #13510, opened 2023-06-14
SchemaType-level global setters to go along with global getters


vkarpov15:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

In #11462, I ran into an issue where you can set a global getter on all date instances:

```javascript
mongoose.Schema.Types.Date.get(v => {
  return Temporal.Instant.fromEpochSeconds(v.valueOf() / 1000);
});
```

But there's no way to set a corresponding global setter.

### Motivation

Getters and setters are complementary, and while it is definitely possible to get a lot out of global getters without global setters, having both is necessary.

The major roadblock is that we use `SchemaType.get()` to define global getters... but `SchemaType.set()` is already used to set global schematype options. So we may instead need to do something like `mongoose.Schema.Types.Date.setters.push(...)` instead of `mongoose.Schema.Types.Date.set()`.

PR #13538, opened 2023-06-22
Island rhythms/gh 13510

IslandRhythms:


IslandRhythms:
The test passes on my machine so I'm not sure why its failing here

test/schematype.test.js
```diff
@@ -211,6 +211,21 @@ describe('schematype', function() {
       it('SchemaType.set, is a function', () => {
         assert.equal(typeof mongoose.SchemaType.set, 'function');
       });
+      it('should allow setting values to a given property gh-13510', async function() {
+        const m = new mongoose.Mongoose();
+        await m.connect('mongodb://127.0.0.1:27017/gh13510');
```
vkarpov15, line 216: Why did you hard code the connection string here? There's nowhere else in the tests, other than a very old sharding test, where we hard code the connection string. Use `mongoose.connect(start.uri)` instead.

vkarpov15 (changes requested):

vkarpov15:
Link #13510 to this issue

Issue #4292, opened 2016-07-02
Browser Model


DylanPiercey:
Although 90% of "model.js" doesn't make sense in the browser I think implementing it with some basic features (like creating a schema from an object and storing the model name) would be handy to make the following code easier to be isomorphic.

``` js
var Test = mongoose.model('test', {
  name: { type: String },
  test: { type: String },
});

// Browser should have
Test.schema
Test.modelName
```

Currently I have to separate out that schema if I want to use it in the browser, also I have to explicitly pass around the modelName since schema's don't have names.

Thoughts?

mkastner:
I've run into the same problem. Maybe this abstracted shared schema module could work for you (with CommonJS):
```
// schema-def.js
// abstracts the schema definition from the schema object creation

var schemaDef = {
    test: {
        name: { type: String },
        test: { type: String }
    }
}

module.exports = schemasDef;
```
Now you can require the abstracted schema definition from client and server side like this:

```
// client-models.js
// sharing same schema defintion with server

var schemaDef = require('schema-def.js');

var Test = mongoose.model('test', schemaDef.test);

// of course no plugins and other fancy server stuff

```

```
// server-models.js
// sharing same schema defintion with client

var schemaDef = require('schema-def.js');

var Test = mongoose.model('test', schemaDef.test);

// Since the schema definition is abstracted,
// on the server side you can add your plugins, indexes etc. 

```
My thoughts.

vkarpov15:
I like this idea. Hasn't really been a high priority since getting documents and schemas to work in the browser was already quite an odyssey

mkastner:
Maybe a plain json file for the schema definition would even be a better solution if one is into typing quotes. In fact loading a plain old json would even be be agnostic with respect to what module type it is called from – commonjs, es6 modules and whatnot.

vkarpov15:
This will be fixed in 9.0 with #15385

PR #13542, opened 2023-06-25
Move all MongoDB-specific connection logic into driver layer, add `createClient()` method to handle creating MongoClient

vkarpov15:
**Summary**

Pulling in some work we started on with #4292 work (see [vkarpov15/gh-4292 branch](https://github.com/Automattic/mongoose/tree/vkarpov15/gh-4292)). One of the annoying pain points in building the in-memory driver was that we couldn't import Mongoose's `Connection` class without importing all of MongoDB. The driver layer is supposed to abstract that out, but connections have a bunch of MongoDB references that should really be in the driver layer.

**Examples**

### DIFF
--- a/test/timestamps.test.js
+++ b/test/timestamps.test.js
@@ -608,12 +608,13 @@ describe('timestamps', function() {
     });
 
     it('should change updatedAt when findOneAndUpdate', async function() {
+      await Cat.deleteMany({});
       await Cat.create({ name: 'test123' });
       let doc = await Cat.findOne({ name: 'test123' });
       const old = doc.updatedAt;
+      await new Promise(resolve => setTimeout(resolve, 10));
       doc = await Cat.findOneAndUpdate({ name: 'test123' }, { $set: { hobby: 'fish' } }, { new: true });
-      assert.ok(doc.updatedAt.getTime() > old.getTime());
-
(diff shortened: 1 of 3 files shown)


AbdelrahmanHafez (approved):

test/timestamps.test.js
```diff
@@ -608,12 +608,13 @@ describe('timestamps', function() {
     });
 
     it('should change updatedAt when findOneAndUpdate', async function() {
+      await Cat.deleteMany({});
       await Cat.create({ name: 'test123' });
       let doc = await Cat.findOne({ name: 'test123' });
       const old = doc.updatedAt;
+      await new Promise(resolve => setTimeout(resolve, 10));
```
hasezoey, line 615: what requires this sleep?

hasezoey (approved):
Looks good to me, though i would recommend to merge this into the next minor version instead of a patch

test/timestamps.test.js
vkarpov15: This test flaked once while I was testing, so I figured I'd make this test more robust while it was fresh in my mind. This test doesn't always fail, but you can see that without the sleep this test can fail.

vkarpov15:
Switched base branch to 7.4 :+1:

PR #13544, opened 2023-06-26
feat(model): add create option "immediateError"

### LINKED ISSUE #1731: Model.create(doc, fn) should save all valid documents.
I'm trying to save a set of documents using `Model.create(documents, function(error){...})` I found when a document is invalid, the previous documents were saved but the following will never be saved.

I think Model.create should save all valid documents. I would like an API like this.

``` javascript
//Passing one doc
Person.create({name: "John"}, function(error, john){...})

//Passing an array
var persons = [{name: "John"}, {name: "Doe"}];
Person.create(persons, function(errors, docs){...});

//Passing individual docs
Person.create({name: "John"}, {name: "Doe"}, function(errors, docs){...});
```

I would like to do a pull request, but first I want to know what do you think.

hasezoey:
**Summary**

This PR adds option `immediateError` to `create`, to not throw on the first error encountered and instead add the error to the result list (only applicable if the input is a array)

Default: `true` for backwards compatibility (current master behavior)

works with both `ordered: false` and `ordered: true`

Result of `immediateError: false`:

```ts
const result = await Model.create([{ test: "yes" }, {}, { test: "no" }], { immediateError: false });
// index 0 is the "test: yes" document
// index 1 is a error instance
// index 2 is the "test: no" document
```

is this implementation OK or should it maybe be split into returning a object of values and errors or some other way?

also moves option `ordered` from "SaveOptions" (on .save) to "CreateOptions" (to .create) because from what i can tell, the option `ordered` has no effect on `.save` itself

fixes #1731

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2802,7 +2802,7 @@ Model.findByIdAndRemove = function(id, options) {
  * @param {Array|Object} docs Documents to insert, as a spread or array
  * @param {Object} [options] Options passed down to `save()`. To specify `options`, `docs` **must** be an array, not a spread. See [Model.save](https://mongoosejs.com/docs/api/model.html#Model.prototype.save()) for available options.
  * @param {Boolean} [options.ordered] saves the docs in series rather than parallel.
- * @param {Boolean} [options.immediateError] If a Error occurs, throw the error immediately, instead of collecting in a result, default: true (backwards compatability)
+ * @param {Boolean} [options.aggregateErrors] Aggregate Errors instead of throwing the first one that occurs. Default: false
  * @return {Promise}
  * @api public
  */
@@ -2859,9 +2859,9 @@ Model.create = async function create(doc, options) {
     return Array.isArray(doc) ? [] : null;
   }
   let res = [];
-  const immediateError = typeof options.immediateError === 'boolean' ? options.immediateError : true;
+  const immediateError = typeof options.aggregateErrors === 'boolean' ? !options.aggregateErrors : true;
 
-  delete options.immediateError; // dont pass on the option to "$save"
+  delete options.aggregateErrors; // dont pass on the option to "$save"
 
   if (options.ordered) {
     for (let i = 0; i < args.length; i++) {
--- a/lib/model.js
+++ b/lib/model.js
@@ -2802,6 +2802,7 @@ Model.findByIdAndRemove = function(id, options) {
  *
  * @param {Array|Object} docs Documents to insert, as a spread or array
  * @param {Object} [options] Options passed down to `save()`. To specify `options`, `docs` **must** be an array, not a spread. See [Model.save](https://mongoosejs.com/docs/api/model.html#Model.prototype.save()) for available options.
+ * @param {Boolean} [options.immediateError] If a Error occurs, throw the error immediately, instead of collecting in a result, default: true (backwards compatability)
  * @return {Promise}
  * @api public
  */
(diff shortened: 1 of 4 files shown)


lib/model.js
```diff
@@ -2858,27 +2859,41 @@ Model.create = async function create(doc, options) {
...
+          throw err;
+        }
+      }
     }
     return res;
   } else {
-    res = await Promise.all(args.map(async doc => {
+    // ".bind(Promise)" is required, otherwise results in "TypeError: Promise.allSettled called on non-object"
+    const promiseType = !immediateError ? Promise.allSettled.bind(Promise) : Promise.all.bind(Promise);
+    let p = promiseType(args.map(async doc => {
```
vkarpov15, line 2896: Why can't this be just `Promise.allSettled(args.map(...))` ? `bind()` is relatively slow, so I would prefer to avoid using that.

lib/model.js
```diff
@@ -2802,6 +2801,8 @@ Model.findByIdAndRemove = function(id, options) {
  *
  * @param {Array|Object} docs Documents to insert, as a spread or array
  * @param {Object} [options] Options passed down to `save()`. To specify `options`, `docs` **must** be an array, not a spread. See [Model.save](https://mongoosejs.com/docs/api/model.html#Model.prototype.save()) for available options.
+ * @param {Boolean} [options.ordered] saves the docs in series rather than parallel.
+ * @param {Boolean} [options.immediateError] If a Error occurs, throw the error immediately, instead of collecting in a result, default: true (backwards compatability)
```
vkarpov15, line 2805: I'm not sure I like this option name. I was thinking `continueOnError` to be consistent with `eachAsync()` and `createCollections()`, but `continueOnError` aggregates an error object to throw when an error occurs. Do you have any other naming ideas? Maybe `useAllSettled` or `aggregateErrors`?

hasezoey: i had named it `immediateError` because of the current (master) behavior where once a error is encountered the first one is directly thrown / rejected, but all operations *should* still run and any subsequent error is just silent

lib/model.js
hasezoey: i had avoided that because using `allSettled` in all cases would mean to both wait until all operations have finished, only to then throw the first error and second the `allSettled` result has to be mapped because of what it returns

types/models.d.ts
```diff
@@ -217,7 +221,7 @@ declare module 'mongoose' {
     >;
 
     /** Creates a new document or documents */
-    create<DocContents = AnyKeys<TRawDocType>>(docs: Array<TRawDocType | DocContents>, options?: SaveOptions): Promise<THydratedDocumentType[]>;
+    create<DocContents = AnyKeys<TRawDocType>>(docs: Array<TRawDocType | DocContents>, options?: CreateOptions): Promise<THydratedDocumentType[]>;
```
vkarpov15, line 224: Need to add TypeScript types to indicate that, if `immediateError: false`, then the return type is `Promise<Array<THydratedDocumentType | Error>>`

vkarpov15 (changes requested):
Needs some TS changes. Also, please update this PR to target the 7.4 branch.

lib/model.js
vkarpov15: Maybe we name the option `aggregateErrors` instead? I think that options that are treated as `false` by default are cleaner to work with.

vkarpov15 (approved):
Will merge this into 7.4.

Issue #13547, opened 2023-06-27
[Docs] remove deprecation notices from 7.x docs for `remove`, `update` and `count`


hasezoey:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Currently the 7.x [`deprecations.md`](https://github.com/Automattic/mongoose/blob/05c6b11411922abeaa3215ad27542818c324e76b/docs/deprecations.md) ([live](https://mongoosejs.com/docs/deprecations.html)) lists the functions `remove`, `update` and `count` as deprecated, but 7.x completely removed them. so my question is if we should update this documentation for the 7.x branch to not have them listed anymore (but still for 6.x)

cc @vkarpov15

PR #13567, opened 2023-07-02
docs: remove deprecations guide, no longer relevant in Mongoose 7

vkarpov15:
**Summary**

This guide is no longer relevant in Mongoose 7 as was mentioned in #13547 

**Examples**


hasezoey:
if it is just currently not relevant anymore, how about repurposing it as a general deprecation notice instead of just mongodb driver deprecations? (a replacement to [the wiki, which seems to not be actively used anymore](https://github.com/Automattic/mongoose/wiki/6.0-Deprecation-Warnings))

and while there is nothing to list just say that there is nothing to list (yet)?

vkarpov15:
Given that there's nothing to list right now, I think we should err on the side of removing this page. We can always add it back once we have something to list.

vkarpov15:
Given that there's nothing to list right now, I think we should err on the side of removing this page. We can always add it back once we have something to list.

vkarpov15:
Given that there's nothing to list right now, I think we should err on the side of removing this page. We can always add it back once we have something to list.

hasezoey:
> Given that there's nothing to list right now, I think we should err on the side of removing this page. We can always add it back once we have something to list.

also sounds good to me

PS: i have hidden the duplicate messages that seem to be there by accident

AbdelrahmanHafez (approved):

IslandRhythms (approved):
LGTM

Uzlopak (approved):
LGTM

hasezoey:
i actually think now that #13572 has been merged, there is a reason to still have the deprecation file, mentioning that `findOneAndUpdate` option `overwrite` is deprecated and `findOneAndReplace` should be used instead

vkarpov15:
Good point @hasezoey , I'll close this and create a new PR to remove the contents of the current deprecation guide and instead add the `overwrite` deprecation there once we have that.

PR #13571, opened 2023-07-03
fix(query): throw readable error when executing a Query instance without an associated model

vkarpov15:
Fix #13570

**Summary**

Quick fix re: discussion from #13570. We don't generally support creating a Query instance without a model currently, but we should throw a more readable error when that occurs

**Examples**


lib/query.js
```diff
@@ -4354,7 +4354,10 @@ Query.prototype.exec = async function exec(op) {
   }
 
   if (this.op == null) {
-    throw new Error('Query must have `op` before executing');
+    throw new MongooseError('Query must have `op` before executing');
```
hasezoey, line 4357: maybe also add a test for a missing operation?

hasezoey (commented):
Looks good to me, just one suggestion

test/query.test.js
```diff
@@ -4104,6 +4105,33 @@ describe('Query', function() {
...
+    await M.deleteMany({});
+    await M.create({ name: 'gh13570' });
+
+    const Q = new M.Query();
+    await assert.rejects(
+      () => Q.lean(),
+      /Query must have `op` before executing/
+    );
+  });
+
```
hasezoey, line 4134:
```suggestion

```

hasezoey (approved):

PR #13572, opened 2023-07-03
fix(query): convert findOneAndUpdate to findOneAndReplace when overwrite set for backwards compat with Mongoose 6

### LINKED ISSUE #13550: .toObject() has different behavior from 6.x to 7.x
### Prerequisites

### Mongoose version

7.x

### Node.js version

18.15

### MongoDB server version

6.x

5.1.3

### Description

The `.toObject()` method called after the creation of a document has a different behavior when you compare mongoose version 6.x and 7.x

It looks like the deeply nested properties , when set to undefined, and passed to a `findByIdAndUpdate` in:

- version 6.x are removed
- version 7.x are set to `null`

removing the `toObject()` method call make the 2 mongoose versions behave the same way, as displayed in the steps to Reproduce

### Steps to Reproduce

```js

const mySchema = new mongoose.Schema({
    extended: {
        type:    Object || String,
        default: {keep: true}
    }
});

const myModel = mongoose.model('my', mySchema);

async function main() {
    await mongoose.connect('mongodb://localhost:27017/test-mongoose');

    const myObj = {
        extended: {
            something: {else: true},
            level1:    {level2: {prop: 'something'}}
        }
    };

    let added = (await myModel.create(myObj)).toObject(); // Without the toObject both versions behave the same
    console.log('>>> added:', added.extended); // { something: { else: true }, level1: { level2: { prop: something } } }

    added.extended.level1.level2.prop = undefined;
    console.log('>>> updated', added.extended); // { something: { else: true }, level1: { level2: { prop: undefined } } }

    await myModel.findByIdAndUpdate(added._id, added, {lean: true, new: true, overwrite: true});

    const found = await myModel.findById(added._id);
    console.log('>>> found:', found.extended);
    // 7.x.x { something: { else: true }, level1: { level2: { prop: null } } }
    // 6.x.x { something: { else: true } }
}

main()
    .then(() => process.exit())
    .catch(e => {
        console.error(e);
        process.exit(1);
    });

```

These are the logs for version 6.x with the `toObject()` method call:  
6x with toObject

These are the logs for version 7.x with the `toObject()` method call:  
7x with toObject

These are the logs for version 7.x without the `toObject()` method call:  
7x without toObject

### Expected Behavior

_No response_

vkarpov15:
Fix #13550

**Summary**

In Mongoose 6, `findOneAndUpdate()` with `overwrite: true` was converted into `findOneAndReplace()`. Looks like that behavior got lost in some of our callback refactoring work for Mongoose 7. This PR adds that behavior back.

What do you think about deprecating the `overwrite` option? It is no longer necessary with `findOneAndReplace()`, and is a potential source of confusion.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -3281,6 +3281,12 @@ Query.prototype.findOneAndUpdate = function(filter, doc, options) {
  */
 
 Query.prototype._findOneAndUpdate = async function _findOneAndUpdate() {
+  // For backwards compability with Mongoose 6 re: #13550
+  if (this._mongooseOptions.overwrite) {
+    this.op = 'findOneAndReplace';
+    return this._findOneAndReplace();
+  }
+
   this._castConditions();
 
   _castArrayFilters(this);
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -4104,4 +4104,14 @@ describe('Query', function() {
       await Error.find().sort('-');
     }, { message: 'Invalid field "" passed to sort()' });
   });
+  it('converts findOneAndUpdate to findOneAndReplace if overwrite set (gh-13550)', async function() {
+    const testSchema = new Schema({
+      name: { type: String }
+    });
+
+    const Test = db.model('Test', testSchema);
+    const q = Test.findOneAndUpdate({}, { name: 'bar' }, { overwrite: true });
+    await q.exec();
+    assert.equal(q.op, 'findOneAndReplace');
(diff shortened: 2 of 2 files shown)


hasezoey (approved):
> What do you think about deprecating the overwrite option? It is no longer necessary with findOneAndReplace(), and is a potential source of confusion.

sounds good to me

Issue #13529, opened 2023-06-19
[REG 6->7] Plain fields cannot be assigned in special case (typescript error)


orgads:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.3.0

### Node.js version

18.16.0

### MongoDB server version

5.x

### Typescript version (if applicable)

5.1.3

### Description

Directly assigning to fields of a document fails with v7 when the document is generic (special case, see example).

### Steps to Reproduce

```typescript
import { Model } from 'mongoose';

interface ResourceDoc {
  foo: string;
}

type ResourceIdT<ResourceIdField extends string> = {
  [key in ResourceIdField]: string;
}
type ResourceDocWithId<ResourceIdField extends string> = ResourceDoc & ResourceIdT<ResourceIdField>;

class TestClass<
  ResourceType extends string,
  DocT extends ResourceDocWithId<`${ResourceType}Id`>>
{
  constructor(dbModel: Model<DocT>) {
    const resourceDoc = new dbModel();
    resourceDoc.foo = 'bar'; // Type 'string' is not assignable to type 'Require_id<DocT>["foo"]'
  }
}
```

If I replace <code>\`${ResourceType}Id\`</code> with `'FooId'` then it works.

### Expected Behavior

It should work.

vkarpov15:
As far as I can tell, the script as written should throw an error. `ResourceType` can be any string, so there's no guarantee that there's a `foo` property on `resourceDoc`.

orgads:
@vkarpov15 Notice that `ResourceDocWithId` is an intersection of `ResourceDoc` (which has `foo`) and `ResourceIdT<ResourceIdField>` (which adds the id field).

This is caused by `Omit`. Not sure what has changed between 6 and 7, but #13573 fixes it.

Issue #13523, opened 2023-06-18
Calls to lean returns FlattenMaps<T> instead of type T specified on the Model.


thaoula:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.3.0

### Node.js version

16

### MongoDB server version

v18.16.0

### Typescript version (if applicable)

5.1.3

### Description

We have recently upgrade Mongoose from v7.0.3 -> v7.3.0 and noticed that calls to lean are now returning the Model type wrapped with FlattenMaps.

For example we we previously had a model typed in the following way -

```
// Model
const document: Model<Quote>;

// Example Repository Usage
public async getById(id: string): Promise<Quote> {
     const quote = await this.document .findById(id).lean().exec();

     // This used to be of type Quote but now FlattenMaps<Quote> and causes a Typescript error
     return quote;
}

```

Since v7.3.0 we have over 100 typescript errors complaining -
```
// Quote her is just an example it can be replaced with T
Type '(FlattenMaps<Quote> & Required<{ _id: string; }>)[]' is not assignable to type 'Quote[]' 
```

To enable us to build we to do the following -
```
// Model
const document: Model<Quote>;

// Example Repository Usage
public async getById(id: string): Promise<Quote> {
     const quote = await this.document .findById(id).lean<Quote>().exec();

     // This would be of type Quote now
     return quote;
}

```

### Steps to Reproduce

Please see description above

### Expected Behavior

The expected behaviour is that calls to to lean with be type to the T used by model unless overridden like I have done in my example above.

vkarpov15:
What does your Quote type look like - do you use maps? This change was intentional because `lean()` means Mongoose won't convert maps. Your map types will still be POJOs.

thaoula:
Hi @vkarpov15 ,

As always, really appreciate your prompt replies.

What do you mean by maps in this context?

Our Quote is a complex object a gist of the Typescript interface definition is below -
```

export interface Quote extends Entity, QuotePricingOptions, AuditDetails {

    status?: QuoteSimpleStatus;
    quoteNumber: string;

    sections?: ItemSection[];
    theme?: Entity;
    terms?: string;
    links?: EntityLink[];
    date: Date;
    expiryDate: Date;

    note?: string;
    description: string;

    account?: Contact;
    company?: Entity;
    owner?: Entity;
    address?: Address;
    contact?: ContactSummary;
[...]

PR #13577, opened 2023-07-04
types: avoid unnecessary MergeType<> if TOverrides not set, clean up statics and insertMany() type issues

vkarpov15:
**Summary**

#13529 points out some issues with `MergeType`, this PR makes it so that we avoid using `MergeType` in `HydratedDocument<>` where possible.

However, for a future release, we should remove HydratedDocument, and instead rely on inferring the hydrated document type from the schema. Inferring HydratedDocument type from the raw document type isn't possible in general, and relies on `Omit<>`, which is super brittle in TypeScript - breaks on private/protected fields, unions with generics, and other cases.

**Examples**

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -1,4 +1,3 @@
-import { ObjectId } from 'bson';
 import {
   Schema,
   Document,
@@ -77,7 +76,7 @@ async function insertManyTest() {
   });
 
   const res = await Test.insertMany([{ foo: 'bar' }], { rawResult: true });
-  expectType<ObjectId>(res.insertedIds[0]);
+  expectType<Types.ObjectId>(res.insertedIds[0]);
 
   const res2 = await Test.insertMany([{ foo: 'bar' }], { ordered: false, rawResult: true });
   expectAssignable<Error | Object | ReturnType<(typeof Test)['hydrate']>>(res2.mongoose.results[0]);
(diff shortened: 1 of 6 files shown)


hasezoey:
> However, for a future release, we should remove HydratedDocument, and instead rely on inferring the hydrated document type from the schema.

to my knowledge `HydratedDocument` is useful when you want to assign values explicitly and dont have something like typegoose would provide called `DocumentType`
(ie have a type that is guaranteed to be the correct and full Document, with `Require_id`, overrides applied and everything that might be necessary in the future)

hasezoey (commented):
Looks good to me, also mostly compiles fine with current typegoose

currently typegoose defines `HydratedDocument<Something, any>`, which makes it incompatible with `HydratedDocument<Something, {}>`, changing to the following helps, but i dont know if this is the best

```ts
TOverrides extends Record<string, never> ?
  Document<unknown, TQueryHelpers, DocType> & Require_id<DocType> :
  IfAny<TOverrides, Document<unknown, TQueryHelpers, DocType> & Require_id<DocType>, Document<unknown, TQueryHelpers, DocType> & MergeType<
    Require_id<DocType>,
    IfAny<TOverrides, {}>
  >>
```

another option (on typegoose's side) would be to remove the `, any` from the `HydratedDocument` call

vkarpov15:
The idea is to make `type HydratedDocument` just pull a generic from a schema or model, instead of attempting to reconstruct the hydrated document from the raw document. In other words, right now we infer the raw document from the schema, then infer the hydrated document type from the raw type. We should instead infer both the raw document type and the hydrated document type from the schema. That will avoid having to `Omit<>`, would also clean up our problem with FlattenMaps in #13523

vkarpov15:
@hasezoey I applied your suggested updates in 8b8f37b, can you please re-review?

hasezoey (approved):
Looks good to me and compiles and runs without problems in the current latest typegoose

orgads:
Thank you!

PR #13589, opened 2023-07-08
fix(document): clean up all array subdocument modified paths on save()

### LINKED ISSUE #13582: Modified paths does not reset properly on subdocuments after document save
### Prerequisites

### Mongoose version

6.9.0, 7.3.1

### Node.js version

18.16.0

### MongoDB server version

4.4

5.1

### Description

Currently when saving a document with multiple subdocuments `document.modifiedPaths({ includeChildren: true })` produces `[]` but every subdocument except for the first one still retain their modifiedPaths() after `save()` has been already executed. I assume that judging by the fact that first subdocument does gets cleaned up this is a non-expected behavior.

### Steps to Reproduce

```

const ElementSchema = new mongoose.Schema({
	elementName: String,
});

const MyDocSchema = new mongoose.Schema({
	docName: String,
	elements: [ElementSchema],
});

const MyDoc = mongoose.model('MyDoc', MyDocSchema);

describe('mongoose after doc save', () => {
	before(async () => {
		await mongoose.connect('mongodb://localhost:27017/autotest');
	});

	describe('modified paths', () => {

		it('should all be cleared after save with multiple elements (non-cloned)', async () => {
			let doc = new MyDoc({ docName: 'MyDocName' });
			doc.elements.push({ elementName: 'ElementName1' });
			doc.elements.push({ elementName: 'ElementName2' });
			doc = await doc.save();
			console.log(doc.elements[1].modifiedPaths())
			expect(allPaths(doc)).to.deep.equal([]);
		});
	});
});

const allPaths = (doc: mongoose.Document) => {
	return [doc, ...doc.$getAllSubdocs()].reduce((ary, d: any) => {
		console.log(d, d.modifiedPaths());
		return ary.concat(d.modifiedPaths().map((p: string) => p));
	}, [] as string[]);
};
```

### Expected Behavior

Both document and subdocuments on any depth should produce `[]` as a result of `modifiedPaths()` invocation after document has been saved.

vkarpov15:
Fix #13582

**Summary**

#13582 points out that we didn't correctly clear `modifiedPaths()` on array subdocuments, other than the first subdocument in the array. This PR fixes that, but it looks like the fix for #6818 succeeded because of the buggy behavior. So we added some new logic to explicitly validate the parent array if making changes underneath a document array.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2682,15 +2682,9 @@ function _getPathsToValidate(doc, pathsToValidate, pathsToSkip) {
       }
     }
 
-    // If underneath a document array, may need to re-validate the parent
-    // array re: gh-6818
-    if (_pathType.$parentSchemaDocArray && typeof _pathType.$parentSchemaDocArray.path === 'string') {
-      paths.add(_pathType.$parentSchemaDocArray.path);
-    }
-
     // Optimization: if primitive path with no validators, or array of primitives
     // with no validators, skip validating this path entirely.
-    if (!_pathType.caster && _pathType.validators.length === 0) {
+    if (!_pathType.caster && _pathType.validators.length === 0 && !_pathType.$parentSchemaDocArray) {
       paths.delete(path);
     } else if (_pathType.$isMongooseArray &&
       !_pathType.$isMongooseDocumentArray && // Skip document arrays...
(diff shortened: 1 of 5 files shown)


hasezoey (commented):
LGTM

Uzlopak (commented):
LGTM

AbdelrahmanHafez (approved):

PR #13608, opened 2023-07-12
feat: support generating custom cast error message with a function

### LINKED ISSUE #3162: I want to hook cast error with custom error message.
Why CastError is throw before validate. So I cannot custom error message with CastError.

like:

```
birthday:{
    type:Date
}
```

```
Schema.path('birthday').validate(function(date){
   return _.isDate(date);
},'invalid birthday');
```

And when I set birthday to a invalid date,then call save method.
But,'invalid birthday' hook is after mongoose build-in CastError , so I cannot get the error with message 'invalid birthday'.

Has any method to do it？Thank you!!

vkarpov15:
Fix #3162

**Summary**

We do already have support for custom error messages, but that is using strings with template variables, like `cast: '{VALUE} is not a number'`.

 With this PR, we'll also support "cast: [null, value => `${value} is not a number`]" as a supported syntax to create a custom error message. 

**Examples**

### DIFF
--- a/test/docs/validation.test.js
+++ b/test/docs/validation.test.js
@@ -384,6 +384,53 @@ describe('validation docs', function() {
     // acquit:ignore:end
   });
 
+  it('Cast Error Message Overwrite', function() {
+    const vehicleSchema = new mongoose.Schema({
+      numWheels: {
+        type: Number,
+        cast: '{VALUE} is not a number'
+      }
+    });
+    const Vehicle = db.model('Vehicle', vehicleSchema);
+
+    const doc = new Vehicle({ numWheels: 'pie' });
+    const err = doc.validateSync();
+
+    err.errors['numWheels'].name; // 'CastError'
+    // "pie" is not a number
+    err.errors['numWheels'].message;
+    // acquit:ignore:start
+    assert.equal(err.errors['numWheels'].name, 'CastError');
+    assert.equal(err.errors['numWheels'].message,
+      '"pie" is not a number');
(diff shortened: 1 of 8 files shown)


hasezoey (approved):
Looks good to me, but documentation about this being possible is missing

AbdelrahmanHafez (approved):
LGTM, thanks!

test/docs/validation.test.js
```diff
@@ -384,6 +384,53 @@ describe('validation docs', function() {
...
+    err.errors['numWheels'].message;
+    // acquit:ignore:start
+    assert.equal(err.errors['numWheels'].name, 'CastError');
+    assert.equal(err.errors['numWheels'].message,
+      '"pie" is not a number');
+    db.deleteModel(/Vehicle/);
+    // acquit:ignore:end
+  });
+
+  /* eslint-disable no-unused-vars */
```
hasezoey, line 412: after the next version of `acquit-ignore` would be released, it would also be possible to just disable for a specific line
```ts
// acquit:ignore-next-line
// eslint-disable-next-line no-unused-vars
const doc = new Vehicle({ numWheels: 'pie' });
```

[`ignore-next-line`](https://github.com/vkarpov15/acquit-ignore#also-supports-a-ignore-next-line)

hasezoey (approved):

test/docs/validation.test.js
vkarpov15: I'll patch that up once we release the next version of acquit-ignore

PR #13614, opened 2023-07-15
perf: speed up mapOfSubdocs benchmark by 4x by avoiding unnecessary O(n^2) loop in getPathsToValidate()

vkarpov15:
**Summary**

Flame graph helped identify that most of the time was spent in the following loop. With a faster implementation that doesn't loop through all modified paths for every subdocument, Mongoose is now 3-4x faster on the mapOfSubdocs benchmark.

```
for (const subdoc of subdocs) {
  for (const p of paths) {
    if (p == null || p.startsWith(fullPathToSubdoc + '.')) {
      paths.delete(p);
    }
  }
}
```

Before this change:

```
{
  "Average save time ms": 439.4
}
```

After this change:

```
{
  "Average save time ms": 145.8
}
```

Flamegraph now looks a bunch more reasonable too:

image

I think we can now mark #13191 and #13271 as fixed, because `$__reset()` is no longer a bottleneck and between this PR and #13280 Mongoose is now more than 10x faster on this particular benchmark.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1130,8 +1130,8 @@ Document.prototype.$set = function $set(path, val, type, options) {
       }
 
       if (utils.isNonBuiltinObject(valForKey) && pathtype === 'nested') {
-        this.$set(prefix + key, path[key], constructing, Object.assign({}, options, { _skipMarkModified: true }));
-        $applyDefaultsToNested(this.$get(prefix + key), prefix + key, this);
+        this.$set(pathName, path[key], constructing, Object.assign({}, options, { _skipMarkModified: true }));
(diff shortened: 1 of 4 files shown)


hasezoey (approved):
LGTM, is this only for 6.x or will this be added to 7.x too?

vkarpov15:
We still merge everything from 6.x back into master, so this will end up in the next 7.x release too.

Uzlopak (commented):
LGTM
Awesome

Uzlopak (approved):

AbdelrahmanHafez (approved):
Nice! LGTM, thanks! 👍

PR #13621, opened 2023-07-18
throw error after all saves are done

### LINKED ISSUE #4628: Make `create()` wait for all documents to finish inserting or error out before throwing an error if `ordered = false`
Hello,
I'm facing a very weird problem which unfortunately happens randomly, so I can put the code that actually reproduces it.

Ideally is something like this:
The model

``` javascript
var aSchema = new Schema({
  _id: {
    type: String
  }
});
mongoose.model("A", aSchema);
```

The operations:

``` javascript
var mongoose = require("mongoose");
var A = mongoose.model("A");
  A.create([{_id: "a"}, {_id: "b"}, {_id: "a"}], function(err, result){
    A.remove({_id: {$in: ["a", "b"]}}, function(err, result) {
      console.log(result);
    });
  });
```

So the problem is that sometimes, the elements do not get removed, without any error being thrown.
It is interesting to note that if i put a setTimeout (2 seconds is enough) between the create and the remove, the elements get removed.

Also, it looks like that if the create doesn't throw an error, the problem never occurs. This is the reason why I put a duplicate _id in the elements.

Any idea? I'm using mongoose 4.5.5

Thanks

IslandRhythms:
closes #4628

### DIFF
--- a/test/model.create.test.js
+++ b/test/model.create.test.js
@@ -197,6 +197,27 @@ describe('model', function() {
         const docs = await Test.find();
         assert.equal(docs.length, 5);
       });
+      it('should throw an error only after all the documents have finished saving gh-4628', async function() {
+        const countSchema = new Schema({ n: Number });
+        const testSchema = new Schema({ name: { type: String, unique: true }, reference: Number });
+
+        const Count = db.model('gh4628', countSchema);
+
+        testSchema.pre('save', async function(next) {
+          const doc = await Count.findOneAndUpdate({}, { $inc: { n: 1 } }, { new: true, upsert: true });
+          this.reference = doc.n;
+          next();
+        });
+
+        const Test = db.model('gh4628Test', testSchema);
+        const data = [];
+        for (let i = 0; i < 11; i++) {
+          data.push({ name: 'Test' + Math.abs(i - 4) });
+        }
+        const res = await Test.create(data, { ordered: false }).catch(err => err);
+        const docs = await Test.find();
+        assert.equal(docs.length, 7); // docs 1,2,3,4 should not go through 11-4 == 7
+      })
     });
   });
 });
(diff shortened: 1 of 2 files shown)


test/model.create.test.js
```diff
@@ -197,6 +197,27 @@ describe('model', function() {
         const docs = await Test.find();
         assert.equal(docs.length, 5);
       });
+      it('should throw an error only after all the documents have finished saving gh-4628', async function() {
+        const countSchema = new Schema({ n: Number });
+        const testSchema = new Schema({ name: { type: String, unique: true }, reference: Number });
+
+        const Count = db.model('gh4628', countSchema);
```
vkarpov15, line 204: Missing `await Count.init()` after this, see https://mongoosejs.com/docs/validation.html#the-unique-option-is-not-a-validator.

test/model.create.test.js
```diff
@@ -197,6 +197,27 @@ describe('model', function() {
...
+
+        testSchema.pre('save', async function(next) {
```
vkarpov15, line 206: Why is this pre hook here? Doesn't seem to be necessary. Are you just looking for a way to leave some saves in progress to test?

vkarpov15 (changes requested):
Couple of minor issues but LGTM overall

hasezoey:
dosnt #13544 already (somewhat) address this? because enabling `aggregateErrors` all errors are returned in the result and all operations are waited until finished?

test/model.create.test.js
IslandRhythms: I just copied the above test and modified it slightly. I see now that for this test that part is not necessary.

vkarpov15:
@hasezoey #13544 addresses a similar issue, but this change applies to cases where `aggregateErrors` is false. Specifically, this change makes it so that you don't need to add an extra option or change your function signature to get the benefits of #4628. `create()` still works as usual, it just waits for all documents to finish inserting before throwing.

vkarpov15 (approved):

PR #13631, opened 2023-07-19
types:  refine definition of types `UpdateQuery`

### LINKED ISSUE #13630: Definition of type `UpdateQuery` is not very accurate cause code hint not work
### Prerequisites

### Mongoose version

7.x

### Node.js version

18.x

### MongoDB server version

4.x

4.x

### Description

The definition of type `UpdateQuery` is not very accurate cause code hint not work. 

### Steps to Reproduce

The definition of types `UpdateQuery` for now:
```typescript
export type UpdateQuery<T> = _UpdateQuery<T> & AnyObject;
```

Typescript can not do code hint  with using it, when you add some field not  belong to this schema. Like this: 
```typescript
// schema
interface SchemaDemo {
  field_1: string;
  field_2: string;
}

function update(doc: UpdateQuery<SchemaDemo>) {}

update({ test_field: 'test' }); // Will not get types error
```
`update({ test_field: 'test' })` It should get a types error, because `test_field` is not a field of interface `SchemaDemo`. Types `UpdateQuery` is an intersection type which include `AnyObject`, so `test_field` matched `AnyObject`.  In a sense, ` UpdateQuery` make no senses for now.

### Expected Behavior

Don't use `AnyObject`, it make type system not work. If user want to do something about custom type, just use generic type to do that.  I think `UpdateQuery` should by defined like that: 
```typescript
export UpdateQuery<TSchema = AnyObject>  = _UpdateQuery<TSchema>  & Partial<TSchema>
```

In this design, more flexibility for users. Usage case:
```typescript
type UpdateQuery<TSchema = AnyObject>  = _UpdateQuery<TSchema>  & Partial<TSchema>

// schema
interface SchemaDemo {
  field_1: string;
  field_2: string;
}

function update(doc: UpdateQuery<SchemaDemo>) {}

update({ field_test: 1 }); //  ❌ no fields
update({ field_1: 2}); // ❌ get erro type is not matched
update({ field_1: '' }); //  ✅
update({ $set: { field_1: '' }}); ✅

function update2(doc: UpdateQuery) {}

update2({  aaa: 1 }); // ✅  used as any 
```
```[tasklist]
### Tasks
```

taozi0818:
Fix #13630

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -605,7 +605,7 @@ declare module 'mongoose' {
    * { age: 30 }
    * ```
    */
-  export type UpdateQuery<T> = _UpdateQuery<T> & AnyObject;
+  export type UpdateQuery<T = AnyObject> = _UpdateQuery<T> & Partial<T>;
 
   export type FlattenMaps<T> = {
     [K in keyof T]: FlattenProperty<T[K]>;


vkarpov15:
There's a wide variety of issues with this PR. Will close and write up details in #13630

PR #13634, opened 2023-07-19
fix(schema): avoid propagating `toObject.transform` and `toJSON.transform` option to implicitly created schemas

### LINKED ISSUE #13599: toJSON options gets passed to implicitly created schemas, including transform
### Prerequisites

### Mongoose version

7.2.0+

### Node.js version

20.x

### MongoDB server version

6.x

_No response_

### Description

A new feature was added (https://github.com/Automattic/mongoose/issues/13325) where `toJSON` and `toObject` options get passed to implicitly created schemas in a schema. I believe this was so that if you set `virtuals: true` and `getters: true` that those options would get passed. Another option gets passed that gives us unexpected results: `transform`. We use this `transform()` function to modify the document before it goes out. Before it was only being run for the top level schema, but now it gets run for every sub document.

### Steps to Reproduce

```js

  {
    items: [{ key: String }]
  },
  {
    toJSON: {
      transform (doc, obj) {
        console.log(obj)
        return obj
      }
    }
  }
)
const MyModel = mongoose.model('MyModel', schema)
await mongoose.connect('mongodb://localhost:27017/testing')
const doc = await MyModel.create({
  items: [{ key: 'foobar' }]
})
doc.toJSON()
await mongoose.disconnect()
```

This will produce logs that look something like this:

```
{ key: 'foobar', _id: new ObjectId("[objectid]") }
{
  items: [ { key: 'foobar', _id: new ObjectId("[objectid]") } ],
  _id: new ObjectId("[objectid]"),
  __v: 0
}
```

### Expected Behavior

Expected behavior is that the defined transform function would only apply to the schema on which it is defined.

The workaround is to wrap all of the implicit schemas with `new mongoose.Schema()`.

vkarpov15:
Fix #13599

**Summary**

#13599 pointed out that we introduced a potentially breaking change in #13396: `transform()` option propagates to implicitly created schemas, which means existing `transform` functions need to account that they'll run on subdocuments. This PR undoes that backwards breaking change by explicitly omitting `transform` from the `toJSON` and `toObject` options propagated down to implicitly created schemas.

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -793,14 +793,21 @@ describe('document', function() {
       assert.strictEqual(myModel.toObject().foo, void 0);
     });
 
-    it('should propogate toObject to implicitly created schemas gh-13325', async function() {
+    it('should propogate toObject to implicitly created schemas (gh-13599) (gh-13325)', async function() {
+      let transformCalls = [];
       const userSchema = Schema({
         firstName: String,
         company: {
           type: { companyId: { type: Schema.Types.ObjectId }, companyName: String }
         }
       }, {
-        toObject: { virtuals: true }
+        toObject: {
+          virtuals: true,
+          transform(doc, ret) {
+            transformCalls.push(doc);
+            return ret;
+          }
+        }
       });
 
       userSchema.virtual('company.details').get(() => 42);
@@ -809,6 +816,8 @@ describe('document', function() {
       const user = new User({ firstName: 'test', company: { companyName: 'foo' } });
       const obj = user.toObject();
       assert.strictEqual(obj.company.details, 42);
+      assert.equal(transformCalls.length, 1);
+      assert.strictEqual(transformCalls[0], user);
     });
   });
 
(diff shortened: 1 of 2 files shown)


test/document.test.js
```diff
@@ -793,14 +793,21 @@ describe('document', function() {
       assert.strictEqual(myModel.toObject().foo, void 0);
     });
 
-    it('should propogate toObject to implicitly created schemas gh-13325', async function() {
+    it('should propogate toObject to implicitly created schemas (gh-13599) (gh-13325)', async function() {
+      let transformCalls = [];
```
hasezoey, line 797:
```suggestion
      const transformCalls = [];
```

test/document.test.js
```diff
@@ -996,15 +1005,22 @@ describe('document', function() {
       assert.equal(foundAlicJson.friends, undefined);
       assert.equal(foundAlicJson.name, 'Alic');
     });
-    it('should propogate toJSON to implicitly created schemas gh-13325', async function() {
+    it('should propogate toJSON to implicitly created schemas (gh-13599) (gh-13325)', async function() {
+      let transformCalls = [];
```
hasezoey, line 1009:
```suggestion
      const transformCalls = [];
```

hasezoey (approved):
should this be reverted again for 8.0 as a breaking change?

PR #13636, opened 2023-07-20
Types: Add inline comment to to tell the default value of the runValidator flag in the queryOptions types

omran95:
**Summary**

I ran into a small issue when using a function like `findOneAndUpdate` or `updateMany`, the `runValidator` flag in the `queryOptions` interface is optional but has no information about the default value.


vkarpov15 (approved):
Thanks :+1:

PR #13641, opened 2023-07-21
feat: change all symbols to use a consistent name

hasezoey:
**Summary**

This PR changes all Symbol-strings to be consistently named
- replacing `.` and `:` with `#`
- adding `mongoose#` where it was not previously

i have made this against the 8.0 branch, because this may not affect others, but still some change


vkarpov15 (approved):
Thanks :+1:

PR #13651, opened 2023-07-24
Island rhythms/deprecate overwrite

### LINKED ISSUE #13578: Deprecate `overwrite` option to `findOneAndUpdate()`
### Prerequisites

### Issue

Re: comments on #13572, we should remove this option for next backwards breaking release. It is confusing and inconsistent with the rest of the API.

IslandRhythms:
closes #13578

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -3268,10 +3268,6 @@ Query.prototype.findOneAndUpdate = function(filter, doc, options) {
     delete options.fields;
   }
 
-  if (options.overwrite) {
-    printOverwriteDeprecationWarning();
-  }
-
   const returnOriginal = this &&
     this.model &&
     this.model.base &&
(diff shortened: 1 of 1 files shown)


lib/query.js
```diff
@@ -3259,6 +3268,10 @@ Query.prototype.findOneAndUpdate = function(filter, doc, options) {
     delete options.fields;
   }
 
+  if (options.overwrite) {
```
vkarpov15, line 3271: Check if `options.overwrite != null`, we should also warn if `overwrite: false` is set.

Also, you should move this to `_findOneAndUpdate()`, because your current implementation won't catch `findOneAndUpdate(filter, update).setOptions({ overwrite: true })`.

vkarpov15 (changes requested):

vkarpov15 (approved):

PR #13660, opened 2023-07-26
BREAKING CHANGE: make model.prototype.deleteOne() return query, not promise

### LINKED ISSUE #13369: Make `Model.prototype.deleteOne()` return a query, not a promise, to be consistent with `Model.prototype.updateOne()`
### Prerequisites

### Issue

Re: #13367, #13223.

vkarpov15:
Fix #13369 

**Summary**

Requires a little extra work beyond what `Document.prototype.updateOne()` does, because:

1. `deleteOne()` should not actually send deleteOne to the database if `this.$__.isDeleted` is set.
2. `deleteOne()` fires deleteOne hooks on subdocs as well.

**Examples**


IslandRhythms (approved):
LGTM

lib/model.js
```diff
@@ -987,18 +988,18 @@ Model.prototype.$__where = function _where(where) {
...
 
 /**
- * Removes this document from the db. Equivalent to `.remove()`.
+ * Delete this document from the db.
  *
  * #### Example:
  *
- *     product = await product.deleteOne();
+ *     await product.deleteOne();
  *     await Product.findById(product._id); // null
```
hasezoey, line 996: why was this changed but the next line not modified?

hasezoey (commented):
LGTM

huynhdev24 (approved):

lib/model.js
vkarpov15: Why would the `findById()` line be modified? The assumption is that here's a `product` document floating around that has an `_id`, so there's no need to do `product = await product.deleteOne()`. Just `await product.deleteOne()` should be sufficient.

PR #13663, opened 2023-07-27
feature: `pathsToSkip` option for Model.validate

IslandRhythms:


lib/model.js
```diff
@@ -4100,9 +4100,12 @@ Model.validate = async function validate(obj, pathsToValidate, context) {
   }
   let paths = Object.keys(schema.paths);
 
-  if (pathsToValidate != null) {
-    const _pathsToValidate = typeof pathsToValidate === 'string' ? new Set(pathsToValidate.split(' ')) : new Set(pathsToValidate);
+  if (pathsOrOptions != null) {
+    const _pathsToValidate = typeof pathsOrOptions === 'string' ? new Set(pathsOrOptions.split(' ')) : Array.isArray(pathsOrOptions) ? new Set(pathsOrOptions) : new Set(paths);
```
vkarpov15, line 4104: In the case where `pathsOrOptions = { pathsToSkip: ['name'] }`, then `paths` will equal `['pathsToSkip']`, which is incorrect.

test/model.validate.test.js
```diff
@@ -144,4 +144,26 @@ describe('model: validate: ', function() {
...
+      age: Number,
+      weight: Number
+    });
+
+    const Test = mongoose.model('Test', testSchema);
+
+    const test = { name: 1, docs: 'a doc' };
+    let pathsOrOptions = { pathsToSkip: ['name'] };
+
+    const err = await Test.validate(test, pathsOrOptions);
```
vkarpov15, line 160: This test case is severely lacking because none of the properties have any validators, so of course `validate()` will always pass. Instead, please add a validator on each path that always fails, like `validate: () => false`, and test that `validate()` returns the correct errors with each form of `Test.validate()`.

vkarpov15 (changes requested):

vkarpov15 (approved):
Merging into 7.5

AbdelrahmanHafez:
Nice!

PR #13670, opened 2023-07-28
Island rhythms/backport pull 12954

IslandRhythms:
Some tests are failing on 5.x related to geoSearch. Is that something I should concern myself with?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1356,7 +1356,7 @@ declare module 'mongoose' {
     /** Lists all paths and their type in the schema. */
     paths: {
       [key: string]: SchemaType;
-    }
+    };
 
     /** Returns the pathType of `path` for this schema. */
     pathType(path: string): string;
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -1213,6 +1213,26 @@ declare module 'mongoose' {
...
+      key?: string;
+      maxDistance?: number;
+      minDistance?: number;
+      query?: AnyObject;
+      spherical?: boolean;
+      /**
+       * Deprecated. Use only with MondoDB below 4.2 (removed in 4.2)
+       * @deprecated
+       */
+      num?: number;
```
vkarpov15, line 1232: Also please add back `uniqueDocs?: boolean`, because Mongoose 5 supports versions of MongoDB that still support that version

vkarpov15 (changes requested):

vkarpov15 (approved):

Issue #13626, opened 2023-07-19
Nested discriminator validation kills node process in 6.11.4


skrtheboss:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.11.4

### Node.js version

14.16.0

### MongoDB server version

4.2

### Typescript version (if applicable)

4.9

### Description

After upgrading mongoose to version 6.11.4, the validation stopped working in some conditions.
If a discriminator schema is created with leaf nodes that contain custom validation logic then the node process exists.

### Steps to Reproduce

```javascript
const mongoose = require("mongoose")

function createParameterizedAttributeAssignmentSchema() {
    return mongoose.Schema({
        value: {
            type: mongoose.Schema.Types.Mixed,
            validate: {
                validator: function () {
                    throw new Error(`Invalid Value!`);
                }
            }
        }
    })
}

const TestModel = mongoose.model('TestModel', new mongoose.Schema({})).discriminator(
    'type1',
    new mongoose.Schema(
        {
            config: {
                type: new mongoose.Schema({
                    attributes: {
                        type: [createParameterizedAttributeAssignmentSchema()],
                        required: false,
                    },
                })
            },
        }
    )
);

const modelInstance = new TestModel({ __t: 'type1', config: { attributes: [{ value: 'test-value' }] } });

const timeout = setTimeout(() => process.exit(0), 100);

await modelInstance.validate().catch(err => {
    console.log(err instanceof mongoose.Error.ValidationError);
    console.log(err.message);
}).finally(() => clearTimeout(timeout));
```

Mongoose 6.11.3: https://runkit.com/embed/3g09wn1jwl4q ( Works as expected ) 
Mongoose 6.11.4: https://runkit.com/embed/cfyj2e1ibkry ( Process exists with exit code 1 )

IslandRhythms:
```javascript
const mongoose = require("mongoose")

function createParameterizedAttributeAssignmentSchema() {
    return mongoose.Schema({
        value: {
            type: mongoose.Schema.Types.Mixed,
            validate: {
                validator: function () {
                    throw new Error(`Invalid Value!`);
                }
            }
        }
    })
}

const TestModel = mongoose.model('TestModel', new mongoose.Schema({})).discriminator(
    'type1',
    new mongoose.Schema(
        {
            config: {
                type: new mongoose.Schema({
                    attributes: {
[...]

PR #13671, opened 2023-07-28
fix(schema): make `Schema.prototype.clone()` avoid creating different copies of subdocuments and single nested paths underneath single nested paths

vkarpov15:
Re: #13626

**Summary**

#13626 is one of those issues where doing the "5 whys" exercise takes you down a very deep rabbit hole. There are a few issues that came together to form the issue reported in #13626. The one addressed in this PR is that `Schema.prototype.clone()` unintentionally creates duplicate constructors for subdocuments and document arrays that are nested underneath a subdocument path.

```javascript
new Schema({
  docArr1: [{ name: String }] // unaffected
  subdoc1: new Schema({
    docArr2: [{ name: String }], // affected
    subdoc2: new Schema({ name: String }) // affected
  });
}).clone();
```

The issue is that `Schema.prototype.clone()` first uses `utils.clone()` to clone the schema's paths, and then clone the single nested paths:

```javascript
s.paths = utils.clone(this.paths);
s.singleNestedPaths = utils.clone(this.singleNestedPaths);
```

`clone()` calls `SchemaType.prototype.clone()`, on each schema type it finds. This means that, after cloning `singleNestedPaths`, `paths['subdoc1'].schema.paths['subdoc2']` will not be the same as `singleNestedPaths['subdoc1.subdoc2']`. They will be distinct clones of the original schema's `singleNestedPaths['subdoc1.subdoc2']`.

Fix is to re-assemble `singleNestedPaths` from the cloned schemas' `paths`.

**Examples**

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -409,7 +409,25 @@ Schema.prototype._clone = function _clone(Constructor) {
   s.paths = utils.clone(this.paths);
   s.nested = utils.clone(this.nested);
   s.subpaths = utils.clone(this.subpaths);
-  s.singleNestedPaths = utils.clone(this.singleNestedPaths);
+  for (const schemaType of Object.values(s.paths)) {
+    if (schemaType.$isSingleNested) {
+      const path = schemaType.path;
+      for (const key of Object.keys(schemaType.schema.paths)) {
+        s.singleNestedPaths[path + '.' + key] = schemaType.schema.paths[key];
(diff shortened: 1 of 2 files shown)


hasezoey (approved):

huynhdev24 (approved):

Uzlopak (approved):
LGTM

Issue #13667, opened 2023-07-27
doc.isModified returns false whenever multiple fields are specified


ThoseGrapefruits:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.11.4, 7.4.1

### Node.js version

20.5.0

### MongoDB server version

6.0.2

### Description

Hello!

It seems like `isModified` had a regression somewhat recently and is now returning `false` whenever multiple paths are passed, even if one of those paths was modified. Not sure if I'm doing something wrong, but I was able to reproduce it on a fresh install of latest mongoose in an isolated repro script. I tested on the latest versions of 6.x and 7.x and the results are the same. I am upgrading from `6.9.0`, where this repro script works as expected.

As far as I can tell in the [documentation](https://mongoosejs.com/docs/api/document.html#Document.prototype.isModified()), this is still a supported feature.

Thanks for your time,
Logan

### Steps to Reproduce

```js
const mongoose = require('mongoose');

const Cat = mongoose.model('Cat', {
  name: String, 
  type: String
});

async function run() {
  await mongoose.connect('mongodb://localhost:27017/cats', {
    useNewUrlParser: true, useUnifiedTopology: true});

  await mongoose.connection.dropDatabase();

  const myCat = new Cat({ name: 'Lord Fluffles' })
  await myCat.save();

  myCat.name = 'Captain Fluffles';

  console.log('isModified multiple', myCat.isModified('name type'));
  console.log('isModified single', myCat.isModified('name'));
  console.log('isDirectModified', myCat.isDirectModified('name type'));
}

run().then(() => process.exit(0));
```

### Expected Behavior

The repro script should print `true` on every line, but the `isModified multiple` line prints `false`.

vkarpov15:
Fixed by #13674

k-chop:
@vkarpov15  I also encountered this issue when I updated to 6.12.0.
Would it be possible for this fix to be backported to 6.x?

I've already fixed our app by passing array instead of string, but it would be nice if this is fixed in 6.x.
I opened backport PR: https://github.com/Automattic/mongoose/pull/13940

PR #13674, opened 2023-07-30
fix: Document.prototype.isModified support for a string of keys as first parameter

gastoncasini:
**Summary**

This pull request should fix the  issue [#13667](https://github.com/Automattic/mongoose/issues/13667). 
The error was caused in a recent change of the Document.isModified method, to resolve it I returned the paths variable declaration as it was before that change.  

**Examples**

These pictures show the repro script provided in the original issue running with and without the changes

after

before

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2216,7 +2216,7 @@ Document.prototype.isModified = function(paths, modifiedPaths) {
     }
 
     if (typeof paths === 'string') {
-      paths = [paths];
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1: looks like unintentional breaking change from https://github.com/Automattic/mongoose/commit/b8ebe80e7b9c41ed45552a479e84324693976447 .

PR #13678, opened 2023-07-31
fix(query): remove unnecessary check for atomic operators in findOneAndReplace()

vkarpov15:
**Summary**

This error check is both brittle and unnecessary. Unnecessary because the MongoDB driver already throws `MongoInvalidArgumentError: Replacement document must not contain atomic operators`. Brittle because it doesn't take into account chaining, like `.findOneAndReplace().set({ $push: 'foo' })`. I also don't like how this check uses vanilla `Error` rather than `MongooseError`.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -3587,9 +3587,6 @@ Query.prototype.findOneAndReplace = function(filter, replacement, options) {
   }
 
   if (replacement != null) {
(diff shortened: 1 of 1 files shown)


vkarpov15:
Merging into 7.5 :+1:

Issue #13514, opened 2023-06-16
required property in schema when passed to InferSchemaTypes is optional


naquiuddin:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.11.2

### Node.js version

18.16.0

### MongoDB server version

6.0

### Typescript version (if applicable)

4.9.5

### Description

The InferSchemaTypes is wrongly creating type for a field that has custom message validation.

### Steps to Reproduce

For example when running InferSchemaType on a schema that has a field with custom error message validation like this:
```ts
    email: {
         type: string,
         required: [true, "email is required"]
    }
```
It returns type as 
`email: string | undefined`

### Expected Behavior

Ideally this has to be.
`email: string`
in type definition

vkarpov15:
I'm unable to repro, the following script compiles fine on TypeScript 4.9.5:

```ts
import { Schema, model } from 'mongoose';

const schema = new Schema({
  email: {
    type: String,
    required: [true, "email is required"]
  } 
});
const Test = model('Test', schema);

const doc = new Test({ email: 'bar' });
const str: string = doc.email;
```

Output:

```
$ ./node_modules/.bin/tsc --strict gh-13514.ts 

$
```

Can you please confirm whether the above script fails to compile for you, and, if so, what's your tsconfig.json?

davidbayo10:
Same problem when hovering over the schema field. It works correctly when application is built with TSC.

![image](https://github.com/Automattic/mongoose/assets/6841248/8ba65915-5201-4981-9f44-a9f030276491)

vkarpov15:
@davidbayo10 what editor are you using and can you please provide a more complete code sample?

YazanDuelLinks:
@vkarpov15 This is a sample running on mongoose v7.3.2, TS 5.1.6 under VS Code with a fresh TS project that provides a default config file.
Using an array results into optional property (copying your own code)
```ts
import { Schema, model } from 'mongoose';

const schema = new Schema({
  email: {
    type: String,
    required: [true, "email is required"]
  } 
});
const Test = model('Test', schema);

const doc = new Test({ email: 'bar' });
const str: string = doc.email;
```
Output
```
index.ts:12:7 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

12 const str: string = doc.email;
         ~~~
```
[...]

PR #13680, opened 2023-08-01
types(schema): support `required: { isRequired: true }` syntax in schema definition

vkarpov15:
Re: #13514

**Summary**

One potential workaround for #13514 is using object syntax for defining `required`, because the following works fine:

```javascript
new Schema({
  email: {
    type: String,
    required: { isRequired: true, message: 'Email is required' }
  }
});
```

But we don't currently support that in TypeScript. This PR adds support for the `required: { isRequired: true }` syntax. Unfortunately doesn't help very much with #13514.

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -1139,7 +1139,10 @@ function gh13514() {
   const schema = new Schema({
     email: {
       type: String,
-      required: [true, 'email is required']
+      required: {
+        isRequired: true,
+        message: 'Email is required'
+      } as const
     }
   });
   const Test = model('Test', schema);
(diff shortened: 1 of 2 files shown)


test/types/schema.test.ts
```diff
@@ -1134,3 +1134,16 @@ function maps() {
   expectType<Map<string, number>>(doc.myMap);
   expectType<number | undefined>(doc.myMap!.get('answer'));
 }
+
+function gh13514() {
+  const schema = new Schema({
+    email: {
+      type: String,
+      required: [true, 'email is required']
```
hasezoey, line 1142: if i understand the addition correctly, this test does not actually cover that

hasezoey (commented):
Change looks good to me, but the test not really

test/types/schema.test.ts
vkarpov15: Yeah this test is attempting to repro #13514 using tsd, no luck.

IslandRhythms (approved):
LGTM

PR #13682, opened 2023-08-01
fix(document): correctly set index when casting subdocs for validation re: #13626

vkarpov15:
**Summary**

Another related issue to #13626. The reason why #13626 was leading to an uncaught error was that the `if (value && !(value instanceof Constructor)) {}` check in `schema/DocumentArrayElement.js` was not getting the correct scope or index set, so the subdocument threw an uncaught validation error because the subdocument didn't know what array it was part of. `scope` was the top level document, not the document array. With this PR, we'll always pass `options.index` and also set `scope` to the document array.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2642,7 +2642,7 @@ function _evaluateRequiredFunctions(doc) {
  */
 
 function _getPathsToValidate(doc) {
-  const skipSchemaValidators = {};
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
LGTM

IslandRhythms (approved):
LGTM

PR #13694, opened 2023-08-02
upgrade eslint packages

AbdelrahmanHafez:
This PR upgrades both eslint related packages.
The new ESLint version seems to require `parserOptions`, and does not work with node v14, so I used v18 for linting, but kept v14 for tsd for backwards compatibility.


.github/workflows/tsd.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@e33196f7422957bea03ed53f6fbb155025ffc7b8 # v3.7.0
         with:
-          node-version: 14
+          node-version: 18
```
AbdelrahmanHafez, line 48:
```suggestion
          node-version: 14
```

AbdelrahmanHafez:
I can see arguments on both sides here.
On one hand, the tooling we're using for development isn't exposed externally, and we have the test suites to guarantee that our ts types/js code work on the minimal node versions supported by mongoose so this shouldn't affect the end user in any way, on the other hand it's probably safer to just use node 14 until it's not supported by the native MongoDB driver, then we can drop supporting it.

On a side note: node 14 has reached EOL according to [this website](https://endoflife.date/nodejs), and node 16 will [follow suit](https://nodejs.org/en/blog/announcements/nodejs16-eol) in 6 weeks.
I honestly don't have strong feelings about this, happy to go either way.
What do you think? @hasezoey @vkarpov15

Uzlopak (approved):
LGTM

It is just linting. and we are not forcing to pass lint by git hooks. So if somebody works with node 14, then so be it.

hasezoey:
i am fine with either way, wanted to point it out and that is what i would go with in my projects (like typegoose or MMS), for the "small" case of having to use a lower nodejs version to debug something and just having all the tooling working correctly too without having to switch back and just the package manager (in my cases `yarn`) to not complain

vkarpov15 (approved):
LGTM, but I'll merge this into 8.0 branch. We should upgrade these eslint plugins eventually, but it isn't worth the risk of any headache or friction from merging these into master branch when we still support Node 14.

PR #13699, opened 2023-08-03
types: add UpdateQueryKnownOnly type for stricter UpdateQuery type checking

### LINKED ISSUE #13630: Definition of type `UpdateQuery` is not very accurate cause code hint not work

vkarpov15:
Fix #13630

**Summary**

#13630 points out that there's no good way right now to enforce a more strict update syntax if you prefer to only allow known top-level keys in updates. This PR adds a separate `UpdateQueryKnownOnly` utility type for that. I'm not entirely convinced that this is useful or necessary, but it could be helpful in some cases. What do you think @hasezoey ?

**Examples**


types/index.d.ts
```diff
@@ -618,6 +618,16 @@ declare module 'mongoose' {
...
 
+  /**
+   * A more strict form of UpdateQuery that enforces updating only
+   * known top-level properties.
+   * @example
+   * ```js
+   * { age: 30 }
+   * ```
+   */
+  export type UpdateQueryKnownOnly<T> = _UpdateQuery<T, {}>;
```
hasezoey, line 629: it would probably good to have a example of how this type is meant to be used.

is it meant to be used like `.find({} as UpdateQueryKnownOnly<IUser>)` or `.find<UpdateQueryKnownOnly<IUser>({})` or a completely different way?

hasezoey (approved):
looks good to me for those that want stricter queries in types, though a example of how to apply the new type would be good

AbdelrahmanHafez (approved):
Nice! Would like to see more of stricter TS for filters/queries.
Can be challenging to apply for dot-separated paths though, but that's a good start!

types/index.d.ts
vkarpov15: More along the lines of this:

```ts
function updateUser(update: UpdateQueryKnownOnly<IUser>) {
  return User.updateOne({}, update);
}
```

Use `UpdateQueryKnownOnly` for extra strictness in params and definitions when you want it. Does that make sense?

vkarpov15:
@AbdelrahmanHafez dot-separated paths is something we're unlikely to support in the foreseeable future. The MongoDB Node driver recently removed support for that because it caused a lot of problems; performance issues, and no way to stop infinite recursion in case of self-referential types. Unfortunately a limitation that we're gonna have to live with for now.

types/index.d.ts
hasezoey: so if no conditions are given via parameters and are "inline", the following should be used, right?
`User.updateOne({}, { something: 1 } as UpdateQueryKnownOnly<IUser>)`

PR #13704, opened 2023-08-07
fix: avoid applying map property getters when saving

### LINKED ISSUE #13657: Map schema type converts UUID into plain string
### Prerequisites

### Mongoose version

7.4.1

### Node.js version

18

### MongoDB server version

4.4

_No response_

### Description

Seems Map schema type makes unnecessary conversion via `__schemaType.applyGetters` for updates query. I noted it when developing my custom types that based on MongooseBuffer, i just tried  Map + UUID combination and got the same issue.

### Steps to Reproduce

```js
import { randomUUID } from 'node:crypto';

const Schema = new mongoose.Schema(
  {
    doc_map: {
      type: mongoose.Schema.Types.Map,
      of: mongoose.Schema.Types.UUID,
    },
  },
  { collection: 'users_test' },
);

const User = mongoose.model('TestUser', Schema);

async function test() {
  const user = new User({
    doc_map: new Map([
      ['role_1', randomUUID()],
      ['role_2', randomUUID()],
    ]),
  });

  await user.save();

  // all good here
  console.log(user.doc_map);

  // trying to update
  user.doc_map.set('role_1', randomUUID());

  await user.save();

  // got string in doc_map -> role_1
}

test().catch(console.error);

```

### Expected Behavior

_No response_

vkarpov15:
Fix #13657

**Summary**

Right now, `save()` will actually apply getters on map paths when updating an existing doc. That's why the test case from #13657 shows maps of uuids getting saved as strings. This PR fixes that by making it so that `getValue()`, Mongoose's internal way to "get dotted path without Mongoose internals", always passes `getters: false` to `map.get()`.

**Examples**

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -640,7 +640,7 @@ exports.getValue = function(path, obj, map) {
 const mapGetterOptions = Object.freeze({ getters: false });
 
 function getValueLookup(obj, part) {
-  const _from = obj && obj._doc ? obj._doc : obj;
+  const _from = obj?._doc || obj;
   return _from instanceof Map ?
     _from.get(part, mapGetterOptions) :
     _from[part];
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -630,8 +630,22 @@ function _populateObj(obj) {
  */
 
 exports.getValue = function(path, obj, map) {
-  return mpath.get(path, obj, '_doc', map);
+  return mpath.get(path, obj, getValueLookup, map);
 };
+
+/*!
+ * ignore
+ */
+
+const mapGetterOptions = Object.freeze({ getters: false });
+
+function getValueLookup(obj, part) {
+  const _from = obj && obj._doc ? obj._doc : obj;
+  obj = _from instanceof Map ?
+    _from.get(part, mapGetterOptions) :
+    _from[part];
+  return obj;
+}
 
 /**
  * Sets the value of `obj` at the given `path`.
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -630,9 +630,23 @@ function _populateObj(obj) {
...
 };
 
+/*!
+ * ignore
+ */
+
+const mapGetterOptions = Object.freeze({ getters: false });
+
+function getValueLookup(obj, part) {
+  const _from = obj && obj._doc ? obj._doc : obj;
```
AbdelrahmanHafez, line 643: Nitpick:
```suggestion
  const _from = obj?._doc || obj;
```

AbdelrahmanHafez (approved):
One minor suggestion, otherwise LGTM. Thanks 👍

hasezoey (approved):

lib/utils.js
```diff
@@ -630,9 +630,23 @@ function _populateObj(obj) {
...
+  obj = _from instanceof Map ?
+    _from.get(part, mapGetterOptions) :
+    _from[part];
+  return obj;
```
Uzlopak, line 647: Is there a reason for reassigning it?

```suggestion
  return _from instanceof Map ?
    _from.get(part, mapGetterOptions) :
    _from[part];
```

Uzlopak (approved):
LGTM

lib/utils.js
vkarpov15: Works for me, we do have one place where we use `?.` in runtime code and no complaints so far. So this should work nicely.

PR #13708, opened 2023-08-08
types: allow accessing `options` from pre middleware

### LINKED ISSUE #13633: Some middleware does not have access to operation options
### Prerequisites

### Mongoose version

7.4.0

### Node.js version

16.19.1

### MongoDB server version

5

4.8.3

### Description

It is not possible to access the operation options in the following middleware functions:

- document pre/post updateOne / deleteOne
- document post save
- pre/post insertMany

I know that the pre middlewares of the above operations do in fact get a reference to the options object supplied by the client. Unfortunately, these objects are not included in the type definitions of the middleware functions, so it is not possible to access them without overriding the types.

### Steps to Reproduce

```typescript
schema.pre("updateOne", { document: true, query: false }, function(next) {
    // no way to access operation options here
})

schema.pre("updateOne", { document: true, query: false }, function(next, options) {
    // does not compile
})

schema.post("save", function(res, next) {
    // no way to access operation options here
})

schema.post("save", function(res, next, options) {
    // does not compile
})

schema.pre("insertMany", function(next, docs) {
    // no way to access operation options here
})

schema.pre("insertMany", function(next, docs, options) {
    // does not compile
})
```

Same with post updateOne, pre/post deleteOne, post insertMany

### Expected Behavior

Ideally I'd like to be able to always access operation options like this can be done with query / aggregation middleware:

```typescript
schema.pre("find", function(next) {
    this.getOptions()
})

schema.pre("aggregate", function(next) {
    this.options
})
```

I understand that this might not be possible with Document middleware, in which case it would be nice to receive the options as the last argument to the middleware. Matter of fact, this is already the case, only the types do not reflect that.

vkarpov15:
Fix #13633

**Summary**

TypeScript types don't support `options` param for document and model middleware currently. This PR adds that.

**Examples**

### DIFF
--- a/types/middlewares.d.ts
+++ b/types/middlewares.d.ts
@@ -31,8 +31,16 @@ declare module 'mongoose' {
   type SchemaPreOptions = MiddlewareOptions;
   type SchemaPostOptions = MiddlewareOptions;
 
-  type PreMiddlewareFunction<ThisType = any> = (this: ThisType, next: CallbackWithoutResultAndOptionalError) => void | Promise<void>;
-  type PreSaveMiddlewareFunction<ThisType = any> = (this: ThisType, next: CallbackWithoutResultAndOptionalError, opts: SaveOptions) => void | Promise<void>;
+  type PreMiddlewareFunction<ThisType = any> = (
+    this: ThisType,
+    next: CallbackWithoutResultAndOptionalError,
+    opts?: Record<string, any>
+  ) => void | Promise<void>;
+  type PreSaveMiddlewareFunction<ThisType = any> = (
+    this: ThisType,
+    next: CallbackWithoutResultAndOptionalError,
+    opts: SaveOptions
+  ) => void | Promise<void>;
   type PostMiddlewareFunction<ThisType = any, ResType = any> = (this: ThisType, res: ResType, next: CallbackWithoutResultAndOptionalError) => void | Promise<void>;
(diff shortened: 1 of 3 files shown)


types/middlewares.d.ts
```diff
@@ -31,8 +31,16 @@ declare module 'mongoose' {
...
-  type PreSaveMiddlewareFunction<ThisType = any> = (this: ThisType, next: CallbackWithoutResultAndOptionalError, opts: SaveOptions) => void | Promise<void>;
+  type PreMiddlewareFunction<ThisType = any> = (
+    this: ThisType,
+    next: CallbackWithoutResultAndOptionalError,
+    opts?: Record<string, any>
+  ) => void | Promise<void>;
+  type PreSaveMiddlewareFunction<ThisType = any> = (
+    this: ThisType,
+    next: CallbackWithoutResultAndOptionalError,
+    opts: SaveOptions
```
AbdelrahmanHafez, line 42: Any particular reason `opts` is not optional here, and is optional in `PreMiddlewareFunction`?
Does it make any difference when defining a last parameter for a function?

AbdelrahmanHafez (approved):
LGTM, thanks. 👍

types/middlewares.d.ts
vkarpov15: For `pre('save')`, Mongoose defaults to passing empty options, so for `schema.pre('save', function(next, opts) { opts; })` opts will always be defined. Even if you call `save()` with no args.

Not true for all middleware functions.

PR #8878, opened 2020-04-27
check discriminator existence before accessing schema in getModelsMapForPopulate

AbdelrahmanHafez:
fixes #8837

Issue #8837, opened 2020-04-22
Cannot read property 'schema' of undefined for discriminated nested array when using `refPath`


qqilihq:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

We have a schema with (1) nested documents, which (2) use discriminators, and where (3) (only) some of the discriminated types uses a `refPath`. With 5.9.9 and 5.9.10 we get an error `Cannot read property 'schema' of undefined` when we try to populate this reference.

**If the current behavior is a bug, please provide the steps to reproduce.**

```typescript
import mongoose from 'mongoose';

beforeAll(async () => mongoose.connect(process.env.MONGO_URL as string));

afterAll(async () => mongoose.disconnect());

it('Mongoose getModelsMapForPopulate issue', async () => {
  const nested = new mongoose.Schema(
    {
      // nothing here
    },
    {
      discriminatorKey: 'type'
    }
  );

  const main = new mongoose.Schema({
    items: [nested]
  });

  const nestedDiscriminated = new mongoose.Schema({
    refPath: { type: String, required: true },
    ref: { type: mongoose.Types.ObjectId, refPath: 'items.refPath' }
    // just for illustration; this works:
    // ref: { type: mongoose.Types.ObjectId, ref: 'getModelsMapForPopulateIssue_other' }
  });

  const itemsType = main.path('items') as mongoose.Schema.Types.DocumentArray;
  itemsType.discriminator('discriminated', nestedDiscriminated);

  const MainModel = mongoose.model<any>('getModelsMapForPopulateIssue_main', main);

  const OtherModel = mongoose.model<any>(
    'getModelsMapForPopulateIssue_other',
    new mongoose.Schema({
      name: { type: String }
    })
  );

  const otherDoc = new OtherModel({ name: 'hello world' });
  await otherDoc.save();

  await new MainModel({
    items: [
      {
        type: 'not_discriminated'
      },
      {
        type: 'discriminated',
        refPath: 'getModelsMapForPopulateIssue_other',
        ref: otherDoc._id
      }
    ]
  }).save();

  const result = await MainModel.find({}).populate('items.ref').exec();
  expect(result[0].items[1].ref.name).toEqual('hello world');
});
```

This might be a regression related to this?

https://github.com/Automattic/mongoose/issues/8731

**What is the expected behavior?**

The test should pass, and the document should be populated. Instead we’re now receiving the following error:

```
    TypeError: Cannot read property 'schema' of undefined

[...]

PR #13716, opened 2023-08-10
fix: accessing discriminator schema with null

Hajjiri:
**Summary**

while trying to use findByIdAndUpdate on a document where its schema uses discriminator. I faced the same error here, https://github.com/Automattic/mongoose/issues/8837

looking at this PR, https://github.com/Automattic/mongoose/pull/8878. Trying to simulate the same fix for this case.

The error (Cannot read properties of null (reading 'schema') stack that I received was, 
```
'TypeError: Cannot read properties of null (reading 'schema')  
 at getEmbeddedDiscriminatorPath (/node_modules/mongoose/lib/helpers/query/getEmbeddedDiscriminatorPath.js:78:110)   
 at walkUpdatePath (/node_modules/mongoose/lib/helpers/query/castUpdate.js:328:22)   
 at castUpdate (/node_modules/mongoose/lib/query.js:4419:28)   
 at processTicksAndRejections (node:internal/process/task_queues:95:5)   
```

This error happens when the update includes a change in the discriminator type, and that (new) type is not defined as a discriminator. What's expected is to return a validation error instead of an exception.

Version used, "mongoose": "^7.4.2",

### DIFF
--- a/test/model.findByIdAndUpdate.test.js
+++ b/test/model.findByIdAndUpdate.test.js
@@ -69,8 +69,8 @@ describe('model: findByIdAndUpdate:', function() {
 
     schema.path('shape').discriminator('gh8378_Circle',
       Schema({ radius: String, color: String }));
-    // schema.path('shape').discriminator('gh8378_Square',
-    //   Schema({ side: Number, color: String }));
+    schema.path('shape').discriminator('gh8378_Square',
+      Schema({ side: Number, color: String }));
 
     const MyModel = db.model('Test', schema);
 
@@ -89,11 +89,11 @@ describe('model: findByIdAndUpdate:', function() {
       'shape.name': 'after',
       'shape.side': 4,
       'shape.color': 'white'
-    }, { new: true });
+    }, { new: true, overwriteDiscriminatorKey: true });
 
     assert.equal(doc.shape.kind, 'gh8378_Square');
     assert.equal(doc.shape.name, 'after');
-    assert.equal(doc.shape.side, 4); // this assertion is failing, this property is not updated
+    assert.equal(doc.shape.side, 4);
     assert.equal(doc.shape.color, 'white');
   });
 });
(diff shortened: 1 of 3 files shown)


test/model.findByIdAndUpdate.test.js
```diff
@@ -0,0 +1,99 @@
...
+    let doc = await MyModel.create({
+      shape: {
+        kind: 'gh8378_Circle',
+        name: 'before',
+        radius: 5,
+        color: 'red'
+      }
+    });
+
+    doc = await MyModel.findByIdAndUpdate(doc._id, {
```
Hajjiri, line 87: should it instead of passing to throw a validation error?

vkarpov15: Right now Mongoose doesn't do any validation on discriminator keys by default. You would need to do the following:

```javascript
const shapeSchema = Schema({ kind: { type: 'String', enum: ['gh8378_Circle'] }, name: String }, { discriminatorKey: 'kind' });
```

And also make sure to set `runValidators` option:

```javascript
doc = await MyModel.findByIdAndUpdate(doc._id, {
      'shape.kind': 'gh8378_Square',
      'shape.name': 'after',
      'shape.side': 4,
      'shape.color': 'white'
    }, { new: true, runValidators: true });
```

vkarpov15 (commented):
@Hajjiri sorry for taking so long to review. I made some quick changes to your tests to make your tests pass, and I've confirmed that this PR fixes a bug. Do the changes I made make sense?

Hajjiri:
> @Hajjiri sorry for taking so long to review. I made some quick changes to your tests to make your tests pass, and I've confirmed that this PR fixes a bug. Do the changes I made make sense?

yup, thanks for continuing the work. it looks good to me.

vkarpov15 (approved):

PR #13718, opened 2023-08-10
types(schema): correct return type for `Schema.prototype.indexes()`

### LINKED ISSUE #13702: TypeScript: Schema.issues() return type is incorrect
### Prerequisites

### Mongoose version

5.11.0 through today

### Node.js version

not pertinent

### MongoDB server version

not pertinent

not pertinent

### Description

Hi. The [docs](https://mongoosejs.com/docs/api/schema.html#Schema.prototype.indexes()) and code (see below) for `schema.prototype.issues` say that the result should be an array of [Index Definition, Index Options] pairs.

But the type definition shows that it returns an array of `[IndexDefinition]`.

[`indexes(): Array<IndexDefinition>;`](https://github.com/Automattic/mongoose/blob/570be58ff8a14dbdf01bd2de387920b3801522f4/types/index.d.ts#L281)

I think it should be `indexes(): Array<[IndexDefinition, IndexOptions]>`.

Code references:

- [ `return getIndexes(this);`](https://github.com/Automattic/mongoose/blob/570be58ff8a14dbdf01bd2de387920b3801522f4/lib/schema.js#L2203)
- [`indexes.push([field, options]);`](https://github.com/Automattic/mongoose/blob/570be58ff8a14dbdf01bd2de387920b3801522f4/lib/helpers/schema/getIndexes.js#L99)

### Steps to Reproduce

Try writing [the example plugin](https://mongoosejs.com/docs/api/schema.html#Schema.prototype.indexes()) in Typescript.

```typescript
function myPlugin(schema:Schema) {
  for (const index of schema.indexes()) {
    if (index[1].unique === undefined) {
      index[1].unique = true;
    }
  }
}
```

It will complain about numerous things. For example, "Property 'unique' does not exist on type 'IndexDirection'."
and "Property 'unique' does not exist on type '1'"

This makes Typescript happy.

```typescript
  function myPlugin(schema:Schema) {
    const indexes = schema.indexes() as unknown as Array<[IndexDefinition, IndexOptions]>;
    for (const index of indexes) {
      if (index[1].unique === undefined) {
        index[1].unique = true;
      }
    }
  }
```

It would be nice if that were unnecessary.

Thank you!

### Expected Behavior

The following does not generate complaints in TypeScript.

```typescript
  function myPlugin(schema:Schema) {
    const indexes = schema.indexes();
    for (const index of indexes) {
      if (index[1].unique === undefined) {
        index[1].unique = true;
      }
    }
  }
```

vkarpov15:
Fix #13702

**Summary**

Just a quick types fix. Current return value doesn't accurately reflect runtime behavior.

**Examples**

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -278,7 +278,7 @@ declare module 'mongoose' {
      * Returns a list of indexes that this schema declares, via `schema.index()`
      * or by `index: true` in a path's options.
      */
-    indexes(): Array<IndexDefinition>;
+    indexes(): Array<[IndexDefinition, IndexOptions]>;
 
     /** Gets a schema option. */
     get<K extends keyof SchemaOptions>(key: K): SchemaOptions[K];
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -1,19 +1,21 @@
 import {
   Schema,
   Document,
+  HydratedDocument,
+  IndexDefinition,
+  IndexOptions,
+  InferSchemaType,
+  InsertManyOptions,
+  ObtainDocumentType,
+  ObtainSchemaGeneric,
+  ResolveSchemaOptions,
   SchemaDefinition,
   SchemaTypeOptions,
   Model,
-  Types,
-  InferSchemaType,
   SchemaType,
+  Types,
   Query,
-  model,
-  HydratedDocument,
-  ResolveSchemaOptions,
-  ObtainDocumentType,
-  ObtainSchemaGeneric,
-  InsertManyOptions
+  model
 } from 'mongoose';
 import { expectType, expectError, expectAssignable } from 'tsd';
 import { ObtainDocumentPathType, ResolvePathType } from '../../types/inferschematype';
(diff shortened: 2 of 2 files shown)


types/index.d.ts
```diff
@@ -278,7 +278,7 @@ declare module 'mongoose' {
      * Returns a list of indexes that this schema declares, via `schema.index()`
      * or by `index: true` in a path's options.
      */
-    indexes(): Array<IndexDefinition>;
+    indexes(): Array<[IndexDefinition, IndexOptions]>;
```
hasezoey, line 281: any specific reason this is `Array<[v, v]>` instead of `[v,v][]`?

hasezoey (approved):

types/index.d.ts
Fonger: maybe Array<[v, v]> is more readable?

vkarpov15: Yeah I think `Array<[v1, v2]>` is slightly more readable, but it's a minor point. We can change this definition if you feel strongly about one syntax or the other. But I'll merge for now because we can always change the definition later.

hasezoey: basically asked because the other definition was used for the tests [reference](https://github.com/Automattic/mongoose/pull/13718/files#diff-8603b830eee3be0bc269312df23ca0e252cc91e4cdda393d2ae7a332cbea30d6R1178)

vkarpov15: Oh yeah I used the other definition in tests as a sanity check to make sure those definitions were equivalent. I would imagine so, but with TS you never know :)

PR #13726, opened 2023-08-12
fix(connection): reset document state in between transaction retries

### LINKED ISSUE #13698: Connection#transaction: Should reset `$isNew` during the retry
### Prerequisites

### Mongoose version

7.4.1

### Node.js version

16.18.1

### MongoDB server version

4.2

4.7.3

### Description

I'm encountering a similar issue as described in https://github.com/Automattic/mongoose/issues/8852, while using the `mongoose` built-in transaction wrapper `Connection#transaction`.

To simplify the problem, I've created the reproducible code below.

The problem arises when I attempt to save a new document within a transaction. If a `TransientTransactionError` occurs during the first try and triggers a [retry](https://github.com/mongodb/node-mongodb-native/blob/0c5c0b4162f7b0263d6bd318a4121b40637b1983/src/sessions.ts#L616), the doc's $isNew flag is not reset before the second attempt. Consequently, the subsequent save operation throws a `DocumentNotFound` error because it attempts to send an update operation behind the scenes instead of an insert.

Upon digging into the code, I discovered that the state reset, which should occur [here](https://github.com/Automattic/mongoose/blob/865887862721689aa1fa6657af2d9b17e001a82e/lib/connection.js#L525), only takes place after two failed tries (in the case of a retry), and not before each retry attempt.

### Steps to Reproduce

```typescript
const doc = new Model({ a: "b" });

await connection.transaction(
  async (session) => {
    await doc.save({ session });
    throw new MongoServerError({
      errorLabels: ["TransientTransactionError"],
    });
  },
  { readPreference: "primary" }
);
```

### Expected Behavior

I expected that the state should be reset before every retry attempt, which should prevent the `DocumentNotFound` error from occurring.

At the moment, my workaround is moving `const doc = new Model({ a: "b" });` into the transaction block.

vkarpov15:
Fix #13698

**Summary**

#13698 pointed out that, although we do reset `$isNew` if the transaction fails, we do not reset `$isNew` in between transaction retries. So if you `save()` a document within a transaction that was created outside of a transaction, retrying a transient error will always fail with `DocumentNotFoundError`.

~Some followup work: we should be running transaction tests in CI. I'll look into that.~ I got transaction tests running in CI

**Examples**

### DIFF
--- a/test/docs/transactions.test.js
+++ b/test/docs/transactions.test.js
@@ -327,6 +327,7 @@ describe('transactions', function() {
     // Session isn't committed
     assert.equal(await Character.countDocuments({ title: /hand/i }), 0);
 
+    await tyrion.doesNotExist();
     await tyrion.deleteOne();
 
     // Undo both update and delete since doc should pull from `$session()`
--- a/test/docs/transactions.test.js
+++ b/test/docs/transactions.test.js
@@ -327,7 +327,7 @@ describe('transactions', function() {
     // Session isn't committed
     assert.equal(await Character.countDocuments({ title: /hand/i }), 0);
 
-    await tyrion.remove();
+    await tyrion.deleteOne();
 
     // Undo both update and delete since doc should pull from `$session()`
     await session.abortTransaction();
@@ -339,6 +339,7 @@ describe('transactions', function() {
   });
 
   it('save() with no changes (gh-8571)', async function() {
+    db.deleteModel(/Test/);
     const Test = db.model('Test', Schema({ name: String }));
 
     await Test.createCollection();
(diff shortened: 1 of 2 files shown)


hasezoey (approved):

test/docs/transactions.test.js
```diff
@@ -349,4 +350,50 @@ describe('transactions', function() {
...
+    let retryCount = 0;
+    await db.transaction(async(session) => {
+      assert.ok(doc.$isNew);
+      await doc.save({ session });
+      if (++retryCount < 3) {
+        throw new mongoose.mongo.MongoServerError({
+          errorLabels: ['TransientTransactionError']
+        });
+      }
+    });
```
AbdelrahmanHafez, line 393: I'm curious, which part of this code triggers a retry?
Does `db.transaction` retry by default on mongo errors?

AbdelrahmanHafez (approved):
One question about the test, otherwise LGTM. Thanks! 👍

test/docs/transactions.test.js
hasezoey: from what i can tell, the mongodb driver itself, snippit:

> Runs a provided callback within a transaction, retrying either the commitTransaction operation or entire transaction as needed (and when the error permits) to better ensure that the transaction can complete successfully.

[source node-mongodb-native 5.7](https://mongodb.github.io/node-mongodb-native/5.7/classes/ClientSession.html#withTransaction)

AbdelrahmanHafez: Today I learned. Thanks!

Issue #12671, opened 2022-11-09
TypeError: Invalid URL while migrating from mongoose v5.10.0 to v6.7.2


pawan-saxena:
### Prerequisites

- [X] I have written a descriptive issue title

### Mongoose version

6.7.2

### Node.js version

16.16.0

### MongoDB version

4.4.17

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

macos Monterey v12.5.1

### Issue

Hi all,
I want to migrate the mongoose from v5.10.0 to v6.7.2
I am following these two migration guides :
https://mongoosejs.com/docs/migrating_to_6.html 
https://github.com/mongodb/node-mongodb-native/blob/HEAD/etc/notes/CHANGES_4.0.0.md

My clusters uses replicaset with 3 nodes.
I am using below config :
```
let options = {
        serverSelectionTimeoutMS: 5000,
        retryWrites : true,
        appName: 'testAppName'
        replicaSet: 'test-replicaset',
        maxIdleTimeMS: 150000, 
        wtimeoutMS: 0, 
        w: "majority", 
        autoIndex: false,
        maxPoolSize: 10,
        keepAlive: true, 
        connectTimeoutMS: 60000
    };
```

`mongoose.connect(mongo_server_uri + '/' + database, options);`

my mongo_server_uri looks like this : 
`mongodb+srv://username:password@development.xyz.mongodb.net`
and database takes the value : 'abc'

so above would become:
`mongoose.connect('mongodb+srv://username:password@development.xyz.mongodb.net/abc',options)`

I am using mongoose : 6.7.2
mongodb server version : 4.4.17
node version : 16.16.0
os : macos Monterey v12.5.1

I get below error : 

```
TypeError: Invalid URL: mongodb+srv://username:password@development.xyz.mongodb.net/abc
      at new URLImpl (/root_path/node_modules/whatwg-url/lib/URL-impl.js:21:13)
      at Object.exports.setup (/root_path/node_modules/whatwg-url/lib/URL.js:54:12)
      at new URL (/root_path/node_modules/whatwg-url/lib/URL.js:115:22)
      at new URLWithoutHost (/root_path/node_modules/mongodb-connection-string-url/lib/index.js:74:1)
      at new ConnectionString (/root_path/node_modules/mongodb-connection-string-url/lib/index.js:126:13)
      at parseOptions (/root_path/node_modules/mongodb/lib/connection_string.js:200:17)
[...]

PR #13733, opened 2023-08-14
docs(migrating_to_6): add note about incompatible packages

vkarpov15:
Re: #12671

**Summary**

Re: https://github.com/Automattic/mongoose/issues/12671#issuecomment-1670998489.

**Examples**


hasezoey (approved):
Looks good to me, are there any example packages that are known incompatible because of `whatwg-url`? because from what i can tell the mentioned link only contains 2 packages, which are not directly the problem of `whatwg-url`

vkarpov15:
Both packages listed in incompatible_packages cause issues when used with whatwg-url. Doesn't really matter whether it is whatwg-url's problem or the packages' problem (IMO both).

PR #13735, opened 2023-08-14
`skipNestedIndexes` option

### LINKED ISSUE #12344: [Feature/help] schema option to set indexes on top level only
### Prerequisites

### 🚀 Feature Proposal

In my apps, I always have a global plugin that sets `timestamps` option to `true` on schemas, and creates a descending index on `createdAt`, because most of the time I wanna show documents in the GUI in descending chronological order.
```js
mongoose.plugin((schema) => {
  if (!schema.get('timestamps')) {
    schema.set('timestamps', true);
  }
  schema.index({ createdAt: -1 });
});
```

The problem with that is that this also creates indexes on all nested schemas/embedded documents.

Is there a way to have the plugin with what we currently have in mongoose to make it so that I only create `createdAt` indexes on the top-level schema only?

### Motivation

I don't want to have unnecessary indexes on my database.

### Example

```js
describe('plugins can set indexes on top-level only', function() {
    it('works with a plugin that sets indexes on top-level only', function() {
      // Arrange
      const m = new mongoose.Mongoose();
      m.plugin(schema => {
        schema.set('timestamps', true);
        schema.index({ createdAt: -1 }); // perhaps an option here `setOnlyIfTopLevel`
      });

      // Act
      const userSchema = new m.Schema({ name: String, friends: [{ name: String }] });
      m.model('User', userSchema);
      console.log(userSchema.indexes());

      // Assert
      assert.equal(userSchema.indexes().length, 1);
    });
  });
```

IslandRhythms:
closes #12344

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -527,6 +527,7 @@ Valid options:
 * [timestamps](#timestamps)
 * [storeSubdocValidationError](#storeSubdocValidationError)
 * [collectionOptions](#collectionOptions)
+* [skipNestedIndexes](#skipNestedIndexes)
 * [methods](#methods)
 * [query](#query-helpers)
 
@@ -1423,6 +1424,34 @@ const Test = mongoose.model('Test', schema);
 // Equivalent to `createCollection({ capped: true, max: 1000 })`
 await Test.createCollection();
 ```
+
+<h2 id="skipNestedIndexes">
+  <a href="#skipNestedIndexes">
+    option: skipNestedIndexes
+  </a>
+</h2>
+
+By default, Mongoose's automatic index building logic builds indexes from nested schemas, which can lead to unnecessary indexes if you are reusing schemas.
+You can set `skipNestedIndexes` to tell Mongoose to skip nested schema indexes for all nested schemas as follows.
+
+```javascript
+// Will build an index on `user.email`
+const userSchema = new Schema(
+  { email: { type: String, index: true } },
+  { autoIndex: true }
+);
+const UserModel = mongoose.model('User', userSchema);
+
+// Would normally build 2 indexes: one on `title` and one on `user.email`.
+// But will only build `title` index because of `skipNestedIndexes`.
+const eventSchema = new Schema(
+  { title: { type: String, index: true }, user: userSchema },
+  { autoIndex: true, skipNestedIndexes: true }
+);
+const UserModel = mongoose.model('User', userSchema);
+```
+
+If you want to `skipNestedIndexes`, but just for one schema, you can use [`excludeIndexes` instead](#excludeIndexes)
 
 <h2 id="es6-classes"><a href="#es6-classes">With ES6 Classes</a></h2>
 
--- a/lib/helpers/schema/getIndexes.js
+++ b/lib/helpers/schema/getIndexes.js
@@ -39,7 +39,7 @@ module.exports = function getIndexes(schema) {
         continue;
       }
 
-      if (path.$isMongooseDocumentArray || path.$isSingleNested) {
+      if ((path.$isMongooseDocumentArray || path.$isSingleNested) && !schema.options.skipNestedIndexes) {
         if (get(path, 'options.excludeIndexes') !== true &&
             get(path, 'schemaOptions.excludeIndexes') !== true &&
             get(path, 'schema.options.excludeIndexes') !== true) {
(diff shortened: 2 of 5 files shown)


AbdelrahmanHafez:
Could you please hold on merging this until I review that it actually addresses the goal I'm trying to achieve? Should have some time tomorrow evening.

vkarpov15:
@AbdelrahmanHafez sure

vkarpov15 (approved):

AbdelrahmanHafez:
That doesn't exactly achieve what I'm trying to do.
Feel free to merge the PR, or close it along with the issue as wontfix because it's admittedly a very-nice use-case.

What I'm trying to achieve is have a way within my plugin to tell whether a schema is top-level or nested, without having to go and manually add the option to a hundred+ nested schemas in my app.

So something like mongoose automatically adding an `$isNested` and/or `$isRootSchema` property to schemas on compiling models. This might break an architectural design constraint, so if you feel like this would unnecessarily complicate things, no need to spend any more effort on this.
```js
'use strict';
const mongoose = require('./');
const assert = require('node:assert');
const { Schema } = mongoose;


run();

function run() {
  mongoose.plugin((schema) => {
    schema.set('timestamps', true);
    // each schema has `$isNested` which is true 
    // for nested schemas, but not for root schemas
    if (!schema.$isNested) {
      schema.index({ createdAt: -1 });
    }
  });


    friends: [
      new Schema({
        name: String,
        customId: { type: String, index: true },
        age: Number
      }) // I do **not** need to define any options here
    ]
  });

  userSchema.index({ 'friends.name': 1 });

  mongoose.model('User', userSchema);

  assert.deepStrictEqual(
    userSchema.indexes().map(([indexKeys]) => indexKeys),
    [
      { 'friends.customId': 1 },
      { 'friends.name': 1 },
      { createdAt: -1 }
    ]
  );
}
``` 


A workaround that works with the current mongoose version without this PR and gives the same UX (other than maybe TS complaints) would be to define a custom option on all nested schemas, and check against that custom option in my plugins:

```js
mongoose.plugin((schema) => {
  schema.set("timestamps", true);

  if (!schema.options.$isNested) {
    schema.index({ createdAt: -1 });
  }
});

  friends: [
    new Schema(
      {
        name: String,
        customId: { type: String, index: true },
        age: Number,
      },
      // adding custom option `$isNested` here to check against in the plugin
      { $isNested: true }
    ),
  ],
});

userSchema.index({ "friends.name": 1 });

mongoose.model("User", userSchema);

assert.deepStrictEqual(
  userSchema.indexes().map(([indexKeys]) => indexKeys),
  [{ "friends.customId": 1 }, { "friends.name": 1 }, { createdAt: -1 }]
);

```

AbdelrahmanHafez:
Looking at this in retrospect, I may have formulated my problem poorly.

I presented the high-level problem and was hoping for any solution.
If I were to rewrite the issue today, it would be:
Is there a way to tell whether a schema is nested or root from within the plugin without having to specify a custom option on each schema?

vkarpov15:
@AbdelrahmanHafez in general no, because it is perfectly fine to use a schema as both a top-level schema and a nested schema.

```javascript
const userSchema = new Schema({ name: String });
mongoose.model('User', userSchema);

const bookSchema = new Schema({ title: String, author: userSchema });
```

I'll consider an alternative approach for your use case. But do you think this PR is still worth merging?

AbdelrahmanHafez:
@vkarpov15 
Honestly, I don't think this would help my case because I can already achieve the same results with a custom option. It might help someone in a different scenario though.

Your call to merge or close the PR, I feel neutral about it.
Thanks a lot for the effort @IslandRhythms either way this goes!

vkarpov15:
I'm going to close this PR then. On further inspection, this feature isn't as useful as I'd thought. It sounds like it wouldn't solve @AbdelrahmanHafez 's issue, and you can already disable nested indexes by default for all nested schemas using the following:

```javascript
'use strict';

const { Schema } = mongoose;
mongoose.set('debug', true);

mongoose.Schema.Types.DocumentArray.set('excludeIndexes', true);
mongoose.Schema.Types.Subdocument.set('excludeIndexes', true);

const schema = new Schema({ name: String });
schema.index({ name: 1 });

const topLevelSchema = new Schema({ subdoc: schema, subdocs: [schema], name: String });
topLevelSchema.index({ name: 1 });

const Test = mongoose.model('Test', topLevelSchema);

void async function main() {
  // Only builds top-level `name` index
  await Test.init();
  console.log('Done');
}();
```

PR #13737, opened 2023-08-14
types: export InferSchemaType

vkarpov15:
**Summary**

TypeScript doesn't seem to care whether we do `export InferSchemaType` or not, but apparently some IDEs do. Do you know if there's any potential harm in doing this @hasezoey ?

**Examples**


hasezoey (approved):
at least for me in vscode there are no problems, and as long as there is no duplicate identifier i dont see a problem explicitly exporting this

Issue #13738, opened 2023-08-15
enhancing document's methods


maybesmurf:
### Prerequisites

- [x] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

It would be nice if we could do achieve something like

```
interface IPerson {
    age: number;
    dob: Date;
    settings: {
      theme: string;
      alerts: {
        sms: boolean;
      }
    }
  }

  const schema = new Schema<IPerson>({
    age: Number,
    dob: Date,
    settings: {
      theme: String,
      alerts: {
        sms: Boolean
      }
    }
  });

  const Person = model<IPerson>('Person', schema);

  const person = new Person({ name: 'person', dob: new Date(), settings: { alerts: { sms: true }, theme: 'light' } });

  expectType<number>(person.get('age'));
  expectType<Date>(person.get('dob'));
  expectType<string>(person.get('settings.theme'));
  expectType<boolean>(person.get('settings.alerts.sms'));
  expectType<{ sms: boolean }>(person.get('settings.alerts'));
```

maybesmurf:
I've made enhancements to various methods in the Document, and I've included utility types `FlatPath` and `ExtractFromPath` sourced from the [bridge-mongo](https://github.com/bridge-codes/bridge-mongo). The changes are ready for review and will be submitted as a pull request very soon. You can get a general idea of how it works by referring to the code snippet above

maybesmurf:
I've made enhancements to various methods in the Document, and I've included utility types `FlatPath` and `ExtractFromPath` sourced from the [bridge-mongo](https://github.com/bridge-codes/bridge-mongo). The changes are ready for review and will be submitted as a pull request very soon. You can get a general idea by referring to the code snippets above and below.

```
type PersonPaths = FlatPath<Person> // "age" | "dob" | "settings" | "settings.theme" | "settings.alerts" | "settings.alerts.sms";
type SmsValue = ExtractFromPath<Person, "settings.alerts.sms">; // boolean
```

PR #13739, opened 2023-08-15
types: enhanced Document's methods

maybesmurf:
ref: #13738 

**Summary**
Improved various Document's methods

### DIFF
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -135,7 +135,7 @@ declare module 'mongoose' {
     errors?: Error.ValidationError;
 
     /** Returns the value of a path. */
-    get(path: string, type?: any, options?: any): any;
+    get<T extends FlatPath<DocType>>(path: T, type?: any, options?: any): ExtractFromPath<DocType, T>;
 
     /**
      * Returns the changes that happened to the document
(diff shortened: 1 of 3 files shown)


types/utility.d.ts
```diff
@@ -40,4 +40,34 @@ type IfEquals<T, U, Y = true, N = false> =
     (<G>() => G extends T ? 1 : 0) extends
     (<G>() => G extends U ? 1 : 0) ? Y : N;
 
+type IsAny<T> = unknown extends T ? ([keyof T] extends [never] ? false : true) : false;
```
vkarpov15, line 43: We already have `IfAny<>`, so this is unnecessary

types/document.d.ts
```diff
@@ -228,17 +230,17 @@ declare module 'mongoose' {
...
 
     /** Executes registered validation rules for this document. */
-    validate(pathsToValidate?: pathsToValidate, options?: AnyObject): Promise<void>;
+    validate<T extends FlatPath<DocType>>(pathsToValidate?: T | T[], options?: AnyObject): Promise<void>;
     validate(options: { pathsToSkip?: pathsToSkip }): Promise<void>;
 
     /** Executes registered validation rules (skipping asynchronous validators) for this document. */
     validateSync(options: { pathsToSkip?: pathsToSkip, [k: string]: any }): Error.ValidationError | null;
-    validateSync(pathsToValidate?: pathsToValidate, options?: AnyObject): Error.ValidationError | null;
+    validateSync<T extends FlatPath<DocType>>(pathsToValidate?: T | T[], options?: AnyObject): Error.ValidationError | null;
```
vkarpov15, line 244: I'm a bit wary of using `FlatPath<>` by default because of potential infinite recursion issues. Any self-referencing objects will lead to `FlatPath<>` blowing up, which is one of the reasons why the MongoDB Node driver [no longer uses their `Join` helper by default](https://github.com/mongodb/node-mongodb-native/blob/2fbb715d7d67039a15b563ea881808a3e29602a7/src/mongo_types.ts#L425).

types/document.d.ts
```diff
@@ -207,7 +209,7 @@ declare module 'mongoose' {
     populated(path: string): any;
 
     /** Sends a replaceOne command with this document `_id` as the query selector. */
-    replaceOne(replacement?: AnyObject, options?: QueryOptions | null): Query<any, this>;
+    replaceOne(replacement?: Partial<Omit<DocType, '_id'>>, options?: QueryOptions | null): Query<any, this>;
```
vkarpov15, line 212: This is incorrect. 1) `AnyObject` is more correct, because Mongoose will run casting to handle any types that don't match, 2) `Partial<>` is incorrect because required fields are still required, 3) omitting `_id` is incorrect: you can specify `_id` as long as the replacement has the same `_id` as the document, but there's no way to know whether the `_id` is the same at compile time.

types/document.d.ts
```diff
@@ -191,7 +193,7 @@ declare module 'mongoose' {
      * for immutable properties. Behaves similarly to `set()`, except for it
      * unsets all properties that aren't in `obj`.
      */
-    overwrite(obj: AnyObject): this;
+    overwrite(obj: Partial<Omit<DocType, '_id'>>): this;
```
vkarpov15, line 196: Similar here. `AnyObject` is more correct, `Partial<>` is wrong, and omitting `_id` is wrong.

types/document.d.ts
```diff
@@ -157,31 +157,33 @@ declare module 'mongoose' {
...
-    invalidate(path: string, errorMsg: string | NativeError, value?: any, kind?: string): NativeError | null;
+    invalidate<T extends FlatPath<DocType>>(path: T, errorMsg: string | NativeError, value?: any, kind?: string): NativeError | null;
 
     /** Returns true if `path` was directly set and modified, else false. */
     isDirectModified(path: string | Array<string>): boolean;
+    isDirectModified<T extends FlatPath<DocType>>(path: T): boolean;
 
     /** Checks if `path` was explicitly selected. If no projection, always returns true. */
-    isDirectSelected(path: string): boolean;
+    isDirectSelected<T extends FlatPath<DocType>>(path: T): boolean;
```
vkarpov15, line 167: This syntax is nice, but would be good to keep the `isDirectSelected(path: string): boolean;` syntax as a fallback because it is perfectly valid to call `isDirectSelected()` on a field that's not in the schema.

vkarpov15 (changes requested):
This PR has some neat ideas, but also has several issues that prevent us from merging. Overall suggestions:

1. Get rid of `FlatPath<>` entirely. Make the `T extends FlatPath<DocType>` generics into `T extends keyof DocType`. That will mean we at least have some of the syntactic benefits for top-level paths.
2. Bring back fallback syntaxes for passing in arbitrary strings.

maybesmurf:
You have some good points. Changes made as requested. Please review. Thanks

vkarpov15 (approved):
Nice, thanks :+1:

PR #13759, opened 2023-08-21
fix(ChangeStream): add _bindEvents to addListener function

yury-ivaniutsenka:
**Summary**
ChangeStream doesn't work with fromEvent function from rxjs because this function uses addListener function which doesn't have _bindEvents
https://github.com/ReactiveX/rxjs/blob/master/src/internal/observable/fromEvent.ts#L291

**Examples**
```js
const productsObservable = fromEvent(
    this.productsModel.watch(),
    'change',
).pipe(map((change) => change.fullDocument));
```
In current implementation nothing happens

### DIFF
--- a/lib/cursor/ChangeStream.js
+++ b/lib/cursor/ChangeStream.js
@@ -128,6 +128,11 @@ class ChangeStream extends EventEmitter {
     return this.driverChangeStream.next(cb);
   }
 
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #13764, opened 2023-08-22
upgrade to mongodb-memory-server 9.0.0

hasezoey:
**Summary**

This PR upgrades `mongodb-memory-server` to `9.0.0`(`-beta.2`), but also does:
- upgrade all used mongodb version's patch version
- add mongodb `7.0.0` as a test (because mms more probably supports it with 9.0.0)
- upgrade deno to 1.36 (because with the new mms version, 1.34 just gets stuck before even starting the server)

Currently known problems:
deno really does not like the new MMS (mongodb?) version, constant `Server selection timed out` and locally many test fail too.
also in a debug run in CI it seems like the server somehow has a lot of failed hearbeats, and i dont understand why

even when testing locally against a mongodb docker container (via `MONGOOSE_TEST_URI`), those same tests fail and the final error logged also is `Server selection timed out after 30000 ms`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -108,7 +108,7 @@ jobs:
       - name: Setup Deno
         uses: denoland/setup-deno@v1
         with:
-          deno-version: v1.34.x
+          deno-version: v1.37.x
       - run: deno --version
       - run: npm install
       - name: Run Deno tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,17 +41,17 @@ jobs:
       matrix:
         node: [14, 16, 18, 20]
         os: [ubuntu-20.04, ubuntu-22.04]
-        mongodb: [4.4.22, 5.0.19, 6.0.9, 7.0.0]
+        mongodb: [4.4.28, 5.0.25, 6.0.14, 7.0.7]
         include:
           - os: ubuntu-20.04 # customize on which matrix the coverage will be collected on
-            mongodb: 5.0.14
+            mongodb: 5.0.25
             node: 16
             coverage: true
         exclude:
           - os: ubuntu-22.04 # exclude because there are no 4.x mongodb builds for 2204
-            mongodb: 4.4.22
+            mongodb: 4.4.28
           - os: ubuntu-22.04 # exclude because there are no 5.x mongodb builds for 2204
-            mongodb: 5.0.19
+            mongodb: 5.0.25
     name: Node ${{ matrix.node }} MongoDB ${{ matrix.mongodb }} OS ${{ matrix.os }}
     env:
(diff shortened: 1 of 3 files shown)


hasezoey:
i still dont know what the problem is, but my theories are:
- mongodb somehow does not refresh the connection in deno
- mongodb does somehow not clear the selection timeout in deno
- there is a connection that is not closed?
- something takes longer than all the timeouts in deno, and so cannot complete it in time?
- those errors exist in nodejs, but deno exits at the first one (and node just ignores it?)

hasezoey:
i just found https://github.com/denoland/deno/issues/19831, which is the exact issue this PR is having, greatly it is already fixed (by https://github.com/denoland/deno/pull/20314), downside is that no deno version has been released yet and only `git:master` works with this code

PR Blocked until deno releases a version with this fix

hasezoey:
deno 1.37.0 released, which included the fix, have pushed the update and the tests seem to be passing now again without problems, will clean-up and update PR tomorrow (or something)

but somehow there are no tests on this PR itself, only tests on the commit (on the fork)

hasezoey:
updated PR to be based on latest master and cleaned it up

*also accidentally rebased once on 8.0 instead of master

this PR now also includes a fix for test file `schema.select.test.js` because somehow in deno some other test leaks a `_id: immutable` index and so errors the test, but i dont know which other test does that or how to properly fix this, so there is only a band-aid fix

hasezoey:
be415381ee5ff210cc9e154d56d106b30e5fd85b is the final test before releasing mongodb-memory-server 9, to ensure it also works outside of its own test-suite in its current state

will clean this PR up once it is released

hasezoey:
before being merged i would like to know if d8e1f4f2064c9c838c4f43adda70bdf4f29c462b is the best way or if someone maybe knows where the problem there is

hasezoey:
just for reference because it does not show up here, @vkarpov15 wrote on d8e1f4f2064c9c838c4f43adda70bdf4f29c462b:

> Likely better to just drop the 'tests' collection. clearTestData() assumes that it runs after a test is done and clears all data related to models that test created.

hasezoey:
Updated to use `db.dropDatabase` directly, to my understanding `clearTestData` currently is just a more robust way of calling `dropDatabase` (retrying if failed), though i can understand if it has a different meaning and so could change in the future.

just as a note, i could not just do the `db.model` first and then call `Test.collection.drop()` because somehow that resulted in a error `MongoServerError: ns not found`.

hasezoey:
changed base branch to 7.x (because of 8.0 merge), if this branch should not get merged over to 8.x, i will make another PR for that.

vkarpov15 (approved):
Sorry for taking so long to review and merge this. LGTM :+1:

Issue #13748, opened 2023-08-17
Getter firing every time on array property


dantenol:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.4.3

### Node.js version

18.17

### MongoDB server version

6.0.6

### Typescript version (if applicable)

5.1.6

### Description

While using getters on a schema with type array, the getter function is called every time the value is accessed. The expected behavior should be run the function only once.

Given this getter function on user.hobbies[]
```ts
const removeSufix = (name: string) => {
  return name.slice(0, -6);
};
```

this is the current result of two console.logs in sequence
```ts
  console.log(user.hobbies); // ["swimming", "football"] => correct value
  console.log(user.hobbies); // ["sw", "fo"] => runs getter again
```

### Steps to Reproduce

https://github.com/dantenol/mongoose-array-getters-demo

### Expected Behavior

Run the getter function only once.

Uzlopak:
Well that it gets called everytime makes sense. That it slices everytime not, because strings as all primitives are immutable.

dantenol:
type is always `string`

therefore, changing to the suggested code to force a string had the same results

PR #13774, opened 2023-08-23
fix(document): make array getters avoid unintentionally modifying array, defer getters until index access instead

vkarpov15:
**Summary**

#13748 points out that accessing `user.hobbies` actually modifies the `hobbies` array if there's a getter on elements of the `hobbies` array. That is a problem. To fix that, I moved the logic for applying getters from `SchemaArray.prototype.applyGetters()`, which is also called when you access `user.hobbies`, and into the top-level `applyGetters()` function, which is only called by `toObject()` and `toJSON()`.

As a side effect, in order to avoid breaking Schema UUID tests, we instead need to apply getters on index access. So `user.hobbies[0]` will fire getters on `hobbies.0`, but not modify `hobbies`. That's what the changes to `types/array/index.js` are for.

Still to investigate: it looks like `oneUser.get('hobbies.0')` double calls getters with this PR. Worth checking before merging.

Also need to think about whether to merge this in 7.5, or wait for 8.0. The `types/array/index.js` changes seem a bit too heavy for a patch release.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -4091,6 +4091,14 @@ function applyGetters(self, json, options) {
       if (ii === last) {
         const val = self.$get(path);
         branch[part] = clone(val, options);
+        if (Array.isArray(branch[part]) && schema.paths[path].$embeddedSchemaType) {
+          for (let i = 0; i < branch[part].length; ++i) {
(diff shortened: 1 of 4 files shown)


AbdelrahmanHafez (approved):
LGTM. 👍

Issue #13617, opened 2023-07-18
Refactor: avoid using mquery for `updateX()`


vkarpov15:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Right now all query functions bypass mquery, except for `updateX()` because of `_update` in `query.js`. That's an annoyingly inconsistent API because mquery relies on its own collection wrapper, which has a tendency to fall out of date with Mongoose's. 

Once we change this, we should also see if we can remove mquery from Mongoose entirely.

PR #13777, opened 2023-08-24
BREAKING CHANGE: stop using mquery for `updateX()`, `deleteX()`, `sort()`, always convert sort args to objects

vkarpov15:
Re: #13617

**Summary**

Unfortunately there's a lot of functionality duplicated between mongoose and mquery, so with #13617 we're moving towards moving this logic into mongoose for easier maintenance. The big change in this PR is `sort()`: we've moved sort handling into Mongoose, and we're making `options.sort` always an object. We're now converting `.sort([['a', 1]])` into `.sort({ a: 1 })`, rather than leaving it as `[['a', 1]]` under the hood. Keeping entries-style syntax internally is not necessary post ES2015, key order is still maintained, and that removes the need for us to handle edge cases like `.sort([['a', 1]]).sort({ b: 1 })`.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -37,6 +37,7 @@ const sanitizeFilter = require('./helpers/query/sanitizeFilter');
 const sanitizeProjection = require('./helpers/query/sanitizeProjection');
 const selectPopulatedFields = require('./helpers/query/selectPopulatedFields');
 const setDefaultsOnInsert = require('./helpers/setDefaultsOnInsert');
+const specialProperties = require('./helpers/specialProperties');
 const updateValidators = require('./helpers/updateValidators');
 const util = require('util');
 const utils = require('./utils');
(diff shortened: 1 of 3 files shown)


AbdelrahmanHafez (approved):
LGTM.
I would've expected some `require('mquery')` to be removed somewhere, along with its removal from `package.json`.

hasezoey:
>  I would've expected some require('mquery') to be removed somewhere, along with its removal from package.json, though.

to my understanding this PR is only for `update`, `delete` and `sort`, but `mquery` is used in some more places, mainly the `mquery.canMerge` function and also `Query.prototype` is `new mquery()` (and `Query.base`)

vkarpov15:
Yeah there are still a few places where we use mquery, most notably for chaining for geo operators. But this PR is mostly just to remove mquery from `updateX()`, `deleteX()` and `sort()`. Those are the most heavily used pieces of code where we still rely on mquery.

PR #13778, opened 2023-08-24
fix(debug): avoid putting virtuals and getters in debug output

vkarpov15:
**Summary**

I was losing my mind for a bit figuring out why I was seeing a virtual in my debug output, even though I was passing a [POJO](https://masteringjs.io/tutorials/fundamentals/pojo) into `updateOne()`. Turns out that debug output will run virtuals and getters by default on document arrays, which doesn't accurately reflect the data passed into the mongodb driver.

**Examples**


hasezoey (approved):

Uzlopak (approved):
LGTM

PR #13786, opened 2023-08-26
feat: allow top level dollar keys with findOneAndUpdate(), update()

vkarpov15:
**Summary**

Right now, MongoDB allows storing top-level keys that start with $, but our `castUpdate()` function doesn't allow top-level dollar keys.

```javascript
'use strict';
  
mongoose.set('debug', true);

void async function main() {
  //await mongoose.connection.dropDatabase();

  await mongoose.connection.collection('test').insertOne({ $set: [1,2, 3, 4] });
  console.log(await mongoose.connection.collection('test').find().toArray());
  console.log('Done');
}();
```

**Examples**

### DIFF
--- a/lib/helpers/query/castUpdate.js
+++ b/lib/helpers/query/castUpdate.js
@@ -14,6 +14,24 @@ const schemaMixedSymbol = require('../../schema/symbols').schemaMixedSymbol;
 const setDottedPath = require('../path/setDottedPath');
 const utils = require('../../utils');
 
+const mongodbUpdateOperators = new Set([
+  '$currentDate',
+  '$inc',
+  '$min',
+  '$max',
+  '$mul',
+  '$rename',
+  '$set',
+  '$setOnInsert',
+  '$unset',
+  '$addToSet',
+  '$pop',
+  '$pull',
+  '$push',
+  '$pullAll',
+  '$bit'
+]);
+
 /**
  * Casts an update op based on the given schema
  *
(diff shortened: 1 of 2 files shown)


hasezoey (dismissed):

hasezoey:
since when does mongodb allow that? because mongodb 4.4 tests fail with "not allowed"

AbdelrahmanHafez (changes requested):
I second hasezoey's concern, I feel like we should either:
A) err on the side of caution for now and keep blocking adding keys that start with `$`.
B) Make it conditional, by fetching the server information and making sure it supports those keys.

A more performant alternative approach to B is to keep things as is in the PR, but add error-handling that throws an error an error along these lines: `mongo does not allow storing keys that start with a $ sign, please upgrade your mongodb server version to version x.y or later`.

vkarpov15:
It looks like MongoDB 5 was the first version to support top-level keys that start with $.

I disagree that we should be checking the MongoDB server version for 2 reasons:

1. Historically, Mongoose has never relied on MongoDB server version. As new features are introduced, we support them, and rely on the MongoDB server to throw an error if you use a feature that a particular version of the MongoDB server doesn't support. For example, we didn't implement server version checks for change streams, transactions, etc.
2. It is entirely possible to have multiple versions of MongoDB in one replica set, or even have one version of a mongos proxy in front of a different version of mongod replica set. So "what is the current MongoDB version?" is a more nuanced version than one might think when you're talking about a library that connects to MongoDB and enables/disables functionality based on version. And opens the door to us having to test what happens during MongoDB server version upgrades.

AbdelrahmanHafez:
Fair points, we can just leave it up to the server to throw the proper errors when needed while supporting the feature for those who have more recent versions.

AbdelrahmanHafez (approved):

hasezoey (approved):
looks good to me, but this should definitely be mentioned in the changelog and in the documentation that only mongodb 5 and up support top-level `$` keys

PR #13795, opened 2023-08-29
set default value for _update when no update object is provided and versionKey is set to false

### LINKED ISSUE #13783: Issue when using _findOneAndUpdate with empty update
### Prerequisites

### Mongoose version

7.4.3

### Node.js version

18.17.0

### MongoDB server version

6.0

5.1.6

### Description

When running a `findOneAndUpdate` query with an empty update, I get a type error that happens in query.js on `delete doc._id`. Here is the error:

'TypeError: Cannot convert undefined or null to object\n' +
    '    at model.Query._findOneAndUpdate (C:\\Users\\username\\projects\\path-replacement\\node_modules\\mongoose\\lib\\query.js:3330:7)\n' +
    '    at model.Query.exec (C:\\Users\\username\\projects\\path-replacement\\node_modules\\mongoose\\lib\\query.js:4425:28)\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n' 

The reason it happens at 3330:7 is because I added console logs to check `this._update` and `doc` after cloning. They are both undefined. Here is what I am logging:

```
 if (!this._update || Object.keys(this._update).length === 0) {
    if (options.upsert) {
      // still need to do the upsert to empty doc
      console.log("update in query where we get err")
      console.log(this._update);
      const doc = clone(this._update);
      console.log("doc in query where we get err");
      console.log(doc);
      delete doc._id;
      this._update = { $set: doc };
    } else {
      this._executionStack = null;
      const res = await this._findOne();
      return res;
    }
}
```

This is what we notice logged in console:

> update in query where we get err
> undefined
> doc in query where we get err
> undefined

### Steps to Reproduce

Run a `findOneAndUpdate` query with an empty update, as follows:
```
await ExampleModel.findOneAndUpdate({ id: 1}, {}, { upsert: true, returnDocument: 'after', returnOriginal: false });
```

This should then produce the type error I am experiencing.

### Expected Behavior

I would expect to get back the response as normal, with no updates, if an update is not provided (i.e. if update is {}).

MohOraby:
**Summary**

Fixes #13783

When calling `findOneAndUpdate` with no update query the `_decorateUpdateWithVersionKey` function adds `'$setOnInsert': { __v: 0 }` to the update doc so it passes the `_mergeUpdate` first check which then sets the `_update` to `{}` but with `versionKey` set to `false` the update doc length is 0 so the `_mergeUpdate` function returns early

this PR set a value of empty object to the `_update` property even when `Object.keys(doc).length === 0` so it's defined

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -3809,14 +3809,14 @@ function _completeManyLean(schema, docs, path, opts) {
  */
 
 Query.prototype._mergeUpdate = function(doc) {
-  if (doc == null || (typeof doc === 'object' && Object.keys(doc).length === 0)) {
-    this._update = {};
-    return;
-  }
-
   if (!this._update) {
     this._update = Array.isArray(doc) ? [] : {};
   }
+
+  if (doc == null || (typeof doc === 'object' && Object.keys(doc).length === 0)) {
+    return;
+  }
+
   if (doc instanceof Query) {
     if (Array.isArray(this._update)) {
       throw new Error('Cannot mix array and object updates');
--- a/lib/query.js
+++ b/lib/query.js
@@ -3810,6 +3810,7 @@ function _completeManyLean(schema, docs, path, opts) {
 
 Query.prototype._mergeUpdate = function(doc) {
   if (doc == null || (typeof doc === 'object' && Object.keys(doc).length === 0)) {
+    this._update = {};
     return;
   }
 
--- a/test/model.findOneAndUpdate.test.js
+++ b/test/model.findOneAndUpdate.test.js
@@ -2184,4 +2184,20 @@ describe('model: findOneAndUpdate:', function() {
       /Cannot set `rawResult` option when `includeResultMetadata` is false/
     );
   });
+
+  it('successfully runs findOneAndUpdate with no update and versionKey set to false (gh-13783)', async function() {
+    const exampleSchema = new mongoose.Schema({
+      name: String
+    }, { versionKey: false });
+
+    const ExampleModel = db.model('Example', exampleSchema);
+
+    const document = await ExampleModel.findOneAndUpdate(
+      { name: 'test' },
+      {},
+      { upsert: true, returnDocument: 'after', returnOriginal: false }
+    );
+    assert.ok(document);
+    assert.ok(document.name, 'test');
+  });
 });


test/model.findOneAndUpdate.test.js
```diff
@@ -2184,4 +2184,20 @@ describe('model: findOneAndUpdate:', function() {
...
+
+    const ExampleModel = db.model('Example', exampleSchema);
+
+    const document = await ExampleModel.findOneAndUpdate(
+      { name: 'test' },
+      {},
+      { upsert: true, returnDocument: 'after', returnOriginal: false }
+    );
+    assert.ok(document);
+    assert.ok(document.name, 'test');
```
valeriuflorescu16, line 2201:
```suggestion
    assert.equal(document.name, 'test');
```

lib/query.js
```diff
@@ -3810,6 +3810,7 @@ function _completeManyLean(schema, docs, path, opts) {
 
 Query.prototype._mergeUpdate = function(doc) {
   if (doc == null || (typeof doc === 'object' && Object.keys(doc).length === 0)) {
+    this._update = {};
```
valeriuflorescu16, line 3813: I don't think we need this. Another solution would be to take
```
 if (!this._update) {
    this._update = Array.isArray(doc) ? [] : {};
  }
```
which is just below this `if` check and move it to the top of the `mergeUpdate` function. Tried this locally and it seems to fix the problem, we also wouldn't need to add a new line of code to this. 

This solution is safe since `Array.isArray(arg)` has null checks, so `doc == null` wouldn't be an issue.

MohOraby: Actually thought about this, but for some reason I saw this snippet was inside the other if statement (bad attention span and was at the end of the day)

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

valeriuflorescu16 (approved):

lib/query.js
valeriuflorescu16: I don't see a resolve button so I think you'll have to resolve this comment, all looks good now 👍🏼

vkarpov15 (approved):
LGTM :+1:

PR #13798, opened 2023-08-30
update docs to reflect change from commit 760fa73

RobertHunter-Pluto:
Found this commit [0] while I was looking for a way to omit specific virtuals using toObject/toJSON, then noticed it wasn't in the api docs. Added an extra sentence to the doc to make it clear that this is an option.

First time contributing, let me know if there's anything else I can add.

[0] [760fa73](https://github.com/Automattic/mongoose/commit/760fa73b1f6fdb6ed5c7d1c8641a56803ffd530a)


vkarpov15 (approved):
Thanks :+1:

PR #13801, opened 2023-09-01
chore(workflows/test): force color

hasezoey:
**Summary**

This PR forces color for CI test output, which should make it easier to parse if there are problems (especially for something like debugging #13764)

Color is not enabled by default in github CI because (from what i can tell) it is detected as "not a tty"


hasezoey:
>  (Linting, type-checking, etc.)

i am unsure, do they support it? (do they use `supports-color` package?)

Uzlopak:
image

Uzlopak (approved):

vkarpov15 (approved):
Thanks :+1:

PR #13818, opened 2023-09-02
fix: remove duplicated 'exists' method in Model interface in models.d.ts

ohzeno:
**Summary**

This PR removes a duplicated method signature for `exists` in the `Model` interface within the `types/models.d.ts` file. The duplication does not add any functional value and could potentially lead to confusion. This change aims to clean up the code for better readability and maintenance.


Uzlopak (approved):
LGTM

AbdelrahmanHafez (approved):

Issue #13800, opened 2023-08-31
Typescript types on "this" in instance methods


ozzyonfire:
### Prerequisites

- [X] I have written a descriptive issue title

### Mongoose version

7.5.0

### Node.js version

18.x

### MongoDB version

5.x

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

13.5

### Issue

I am trying to get Typescript typings to work on instance methods defined on my schema. I've followed along in the docs, but I can only get types working on "this" that reflect methods, and not any properties.

Example schema (I pulled from a working example I found on here):
```ts
import { Model, Schema, model } from 'mongoose';

interface IUser {
	firstName: string;
	lastName: string;
}

// Put all user instance methods in this interface:
interface IUserMethods {
	fullName(): string;
}

// Create a new Model type that knows about IUserMethods...
type UserModel = Model<IUser, {}, IUserMethods>;

// And a schema that knows about IUserMethods
const schema = new Schema<IUser, UserModel, IUserMethods>({
	firstName: { type: String, required: true },
	lastName: { type: String, required: true },
});

schema.method('fullName', function fullName() {
	return this.firstName + ' ' + this.lastName;
});

const User = model<IUser, UserModel>('User', schema);

const user = new User({ firstName: 'Jean-Luc', lastName: 'Picard' });
user.firstName; // the typings here work great
const fullName: string = user.fullName(); // 'Jean-Luc Picard'
```

But my typescript doesn't give me the proper type of "this", "firstName" or "lastName"...
![image](https://github.com/Automattic/mongoose/assets/779491/618f6141-4911-4b54-bd74-fc38011a37a4)

I am upgrading my project from an older version of mongoose, typescript, etc. and I swear I used to have proper property types on this. 

Typescript version: 5.2.2
Node version 18 and my tsconfig is set up with "strict": true
``` json
{
	"extends": "@tsconfig/node18/tsconfig.json",
	"compilerOptions": {
		"strict": true,
		"strictNullChecks": true,
		"allowSyntheticDefaultImports": true,
		"esModuleInterop": true,
	}
}
```

Thanks for any insight!

pshaddel:
I made this PR to fix this: https://github.com/Automattic/mongoose/pull/13822

valeriuflorescu16:
I think you could update your schema as follows:
```
const schema = new Schema<IUser, UserModel, IUserMethods>({
	firstName: { type: String, required: true },
[...]

PR #13822, opened 2023-09-04
types(schematypes): use DocType for instance method this

pshaddel:
Fixes this issue: https://github.com/Automattic/mongoose/issues/13800

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -1201,8 +1201,7 @@ function gh13800() {
     lastName: { type: String, required: true }
   });
   schema.method('fullName', function fullName() {
-     type This = typeof this;
-     expectType<IUser>(this);
+    expectType<IUser>(this);
   });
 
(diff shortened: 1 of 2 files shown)


pshaddel:
Right now instance method names does not have a typing(also the function we are passing) when user is passing InstanceMethod type. Let me know if I need to implement this too!

ozzyonfire:
This looks great. Thanks for implementing it so quickly. I'm wondering if "this" should also expose the methods of a HydratedDocument. Like for example, "populate". 

There are a couple times I call `this.populate('item')` or something similar inside my instance method. And right now Typescript doesn't like that.

pshaddel:
> 

Thanks for pointing that out. Just changed it to HydratedDocument.

pshaddel:
I have pushed a new commit to fix the typescript failing test.

vkarpov15 (approved):
Thanks!

PR #13837, opened 2023-09-07
BREAKING CHANGE: rename SchemaType classes to consistently start with "Schema", make file names upper case

### LINKED ISSUE #13309: change schema-type class names to be consistent
### Prerequisites

### Issue

Currently schema types may or may not have the prefix `Schema`, this issue tracks to consistenize all types to be named in the same style.

types that do have the prefix:
- `lib/schema/buffer.js` named `SchemaBuffer`
- `lib/schema/string.js` named `SchemaString`
- `lib/schema/date.js` named `SchemaDate`
- `lib/schema/number.js` named `SchemaNumber`
- `lib/schema/uuidjs` named `SchemaUUID`

types that currently dont have the prefix:
- `lib/schema/decimal128.js` named `Decimal128`
- `lib/schema/documentarray.js` named `DocumentArrayPath`
- `lib/schema/map.js` named `Map`
- `lib/schema/mixed.js` named `Mixed`
- `lib/schema/objectid.js` named `ObjectId`

personally i think those that do not have the prefix currently, should get it instead of removing it from those that have it

vkarpov15:
Fix #13309

**Summary**

Re: #13309, this PR makes all schema type class names consistently start with `Schema` to match what we already have for `SchemaNumber`, etc. I think I prefer the `SchemaNumber` naming convention to the `NumberPath` naming convention. Also changed the file names to be uppercase to match the export name, so `Schema.Types.UUID` is in `schema/UUID.js`, etc. 

**Examples**


hasezoey (approved):
definitely great to have consistent naming, though personally i would prefer if all file names would be camelCase (instead of the new PascalCase), to be consistent with other files in the project

vkarpov15:
I figure I'd prefer file name to line up with class name as much as possible. So if file exports function `doStuff`, then `doStuff.js` is fine. But if file exports class `SchemaNumber`, then I prefer `SchemaNumber.js` or `Number.js`. What would you think about renaming to `schema/SchemaNumber.js`?

hasezoey:
> What would you think about renaming to schema/SchemaNumber.js?

i think the current way is OK, personally i just prefer (in my projects) to have all files named consistently (which for js projects is camelCase, or in rarer cases snake_case)

hasezoey:
what about the filename `DocumentArrayElement.js`?

PR #13842, opened 2023-09-08
Edits to Handle null objects in arrays during populate operations 

vkarpov15:
**Summary**

#13839 , plus lint fixes and test case

**Examples**


IslandRhythms (approved):
LGTM

PR #13843, opened 2023-09-08
BREAKING CHANGE: apply minimize by default when updating document

### LINKED ISSUE #13782: Minimize does sometimes not working. Empty object is saved.
### Prerequisites

### Mongoose version

7.4.4

### Node.js version

16.20.2

### MongoDB server version

7.0.0

_No response_

### Description

Empty objects are sometimes deleted, sometimes not.

### Steps to Reproduce

```js
  mongoose.connect("mongodb://127.0.0.1:27017/test");
  const schema = new Schema({
    metadata: {
      type: {},
      default: {},
      required: true,
      _id: false
    }
  }, { minimize: true });
  const Model = model("Schema", schema);
  const m = new Model({ metadata: {} });
  await m.save();
  console.log("Document has no metadata property")

  const x = await Model.findById(m._id).exec();
  x.metadata = {};
  await x.save();
  console.log(x.$isEmpty('metadata')) // true
  console.log("Document now has empty metadata property");
```

### Expected Behavior

The saved document has the form
```json
{
  "_id": {
    "$oid": "64e87a78972e351e7096f778"
  },
  "__v": 0
}
```
but it is
```json
{
  "_id": {
    "$oid": "64e87a78972e351e7096f773"
  },
  "__v": 0,
  "metadata": {}
}
```

vkarpov15:
Fix #13782

**Summary**

This change is small code-wise, but significant and worth thinking about carefully. The [docs on minimize](https://mongoosejs.com/docs/guide.html#minimize) say that `Mongoose will, by default, "minimize" schemas by removing empty objects.` but don't specify exactly when minimize is applied.

Right now, Mongoose only applies `minimize()` when `$toObject()` is called. So that means when you `save()` a new doc, Mongoose applies `minimize()` unless you set `minimize: false` option in your schema. However, if you try to update an existing doc's nested or subdocument property, Mongoose will currently not apply minimize: Mongoose will store an empty object in MongoDB. However, if you update a document array on an existing document, Mongoose will minimize the document array.

This behavior is inconsistent. This PR makes it so that Mongoose applies `minimize()` when updating an existing document with `save()`. However, I'm a bit worried about changing this behavior; I'm worried of unforeseen consequences. Any thoughts @hasezoey @AbdelrahmanHafez ?

**Examples**


lib/helpers/minimize.js
```diff
@@ -0,0 +1,41 @@
...
+
+const { isPOJO } = require('../utils');
+
+module.exports = minimize;
+
+/**
+ * Minimizes an object, removing undefined values and empty objects
+ *
+ * @param {Object} object to minimize
+ * @return {Object}
```
hasezoey, line 11:
```suggestion
 * @return {Object|undefined}
```

hasezoey (approved):
Looks and sounds good to me, as long as all the tests still pass and this change is mentioned in the changelog, hopefully this wont turn into another `strictQuery` thing

vkarpov15:
Yeah, the `strictQuery` issue is exactly what I'm worried about with this PR. I'll think a little more about it, but I think in the interest of learning our lessons from `strictQuery`, we shouldn't hesitate to remove this behavior if it causes `strictQuery`-like problems.

PR #13846, opened 2023-09-09
fix(discriminator): make base schema paths come before discriminator schema paths when running setters, validators, etc.

### LINKED ISSUE #13794: Wrong field processing order with discriminators
### Prerequisites

### Mongoose version

7.4.5

### Node.js version

16

### MongoDB server version

-

-

### Description

When using discriminator sub schema fields are processed before the base schema fields but according to mongoose documentation the fields should be processed in order of their definition in schema.

In my use case, I need to set and validate values of the sub schema fields according to a field of base schema. But since sub schema fields are processed first the base schema fields are coming to be undefined in sub schema field's setters and validators.

### Steps to Reproduce

```
const { Schema } = mongoose;

mongoose.connect('mongodb://localhost:27017/testdb');

const E1 = new Schema({
  f1: {
    type: Number,
    set() {
      console.log('Set 1');
      return 1;
    },
    validate: [
      {
        validator: (v) => {
          console.log('Validator 1');
          return true;
        }
      }
    ]
  },

});
const E1Model = mongoose.model('E1', E1);

const E2 = new Schema({
  f2: {
    type: Number,
    set() {
      console.log('Set 2');
      return 2;
    },
    validate: [
      {
        validator: (v) => {
          console.log('Validator 2');
          return true;
        }
      }
    ]
  },
});
const E2Model = E1Model.discriminator('E2', E2, { discriminatorKey: 'kind' })

async function main() {
  let item1 = new E2Model({ f1: 1, f2: 2 });
  item1 = await item1.save()
}

main();
```

### Expected Behavior

The setters & validators of base class fields should be processed before sub class fields

vkarpov15:
Fix #13794

**Summary**

Right now, discriminators add base schema paths to the child schema, which means that setters and validators on child schema paths run _before_ setters and validators on base schema paths. This PR switches that order.

While this PR doesn't break any of our tests, I'm considering merging it into 8.0 branch just in case. Any thoughts @hasezoey @AbdelrahmanHafez ?

**Examples**


lib/helpers/model/discriminator.js
```diff
@@ -116,6 +116,7 @@ module.exports = function discriminator(model, name, schema, tiedValue, applyPlu
     }
 
     mergeDiscriminatorSchema(schema, baseSchema);
+    schema.tree = Object.assign({}, baseSchema.tree, schema.tree);
```
hasezoey, line 119: why is this call outside of `mergeDiscriminatorSchema` instead of inside as a separate case?

hasezoey (commented):
Looks good to me, though one question

lib/helpers/model/discriminator.js
vkarpov15: You're right, that's better design. Will fix.

hasezoey (approved):
LGTM, aside from the base-branch merge conflicts

Issue #13772, opened 2023-08-23
InferSchemaType produces DocumentArray instead of a simple Array


juona:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.4.4

### Node.js version

16.19.1

### MongoDB server version

5

### Typescript version (if applicable)

4.8.3

### Description

The `InferSchemaType` utility converts subdocument arrays into `DocumentArray`s. My understanding is that this utility should produce a plain type that has nothing to do with Mongoose, so it should infer simple arrays instead of `DocumentArray`s. Is my expectation incorrect?

Possibly related: https://github.com/Automattic/mongoose/issues/12030 .

### Steps to Reproduce

```typescript
const mySchema = new Schema({
    nodes: {
        type: [
            new Schema({
                foo: String
            })
        ],
        required: true
    }
});

type IMySchema = InferSchemaType<typeof mySchema>;
```

Type `IMySchema` is:

```typescript
type IMySchema = {
    nodes: Types.DocumentArray<{
        foo?: string | undefined;
    }>;
}
```

### Expected Behavior

I'd expect type `IMySchema` to be:

```typescript
type IMySchema = {
    nodes: Array<{
        foo?: string | undefined;
    }>;
}
```

vkarpov15:
"My understanding is that this utility should produce a plain type that has nothing to do with Mongoose": Ideally yes, the `Types.DocumentArray<>` workaround is one exception to that rule. We're trying to change that for 8.0.

vkarpov15:
As expected, going to be a bit hard to change this without #13856. Because `InferSchemaType<>` can't quite decide whether it wants to return the raw document type or the hydrated document type. And we don't want to do type transformations in `HydratedDocument<>` because of issues like #13523: mapping types in TypeScript get screwed up by generics, private fields, etc.

benmeiri:
any workaround for this one?

vkarpov15:
Not currently, we're working on a workaround.

vkarpov15:
Going to move this out of 8.0 to a future release with the work we're doing on #13900. We haven't been able to make progress on this issue, and I don't want this to block 8.0 any further. Plus the work from #13900 is something we can release in a minor release.

FantinRaimbault:
My solution :

```js
const likeSchema = new Schema(
    {
        userId: { type: String, required: true },
        date: { type: Date, required: true },
    },
    {
[...]

PR #13856, opened 2023-09-12
Create separate generic for schema definition

vkarpov15:
**Summary**

Re: #13523, #13772.

The current implementation of automatic type inference makes it tricky for us to automatically infer the correct hydrated document type. Because automatic type inference infers the `DocType`, which is a mix of the hydrated document type and the raw doc type: hydrated for arrays and maps, raw for everything else, including nested paths. And then we infer the schema definition type from the `DocType`, which means we lose whether object paths are [nested paths or subdocument paths](https://mongoosejs.com/docs/subdocs.html#subdocuments-versus-nested-paths).

This PR presents one approach that can help us move towards automatically inferring the raw doc type and the hydrated doc type correctly: getting the schema definition and inferring the raw doc type from that. In theory, if we have the schema definition type, we should be able to infer the hydrated document type from that.

The downside is this approach seems to require adding `as const` to schema definitions if you want to use `type: 'String'` or `type: 'Number'` as opposed to `type: String` or `type: Number`. Also has some issues with context for default functions, but those issues are present to some extent even without this PR and have easy workarounds.

cc @mohammad0-0ahmad 

**Examples**


test/types/schema.test.ts
```diff
@@ -1007,7 +1014,7 @@ function gh12869() {
 
   const dbExample = new Schema(
     {
-      active: { type: String, enum: ['foo', 'bar'], required: true }
+      active: { type: String, enum: ['foo', 'bar'] as ['foo', 'bar'], required: true }
```
hasezoey, line 1017: is this casting really required instead of just `as const`?

test/types/virtuals.test.ts
```diff
@@ -95,14 +95,15 @@ async function autoTypedVirtuals() {
   const testSchema = new Schema({
     email: {
       type: String,
-      required: [true, 'email is required']
+      required: [true, 'email is required'] as [true, string]
```
hasezoey, line 98: are those array castings really required instead of `as const`?

test/types/virtuals.test.ts
```diff
@@ -95,14 +95,15 @@ async function autoTypedVirtuals() {
...
     }
   }, {
     virtuals: {
       domain: {
         get() {
           expectType<Document<any, any, { email: string }> & AutoTypedSchemaType>(this);
-          return this.email.slice(this.email.indexOf('@') + 1);
+          const email = this.email;
+          return email.slice(email.indexOf('@') + 1);
```
hasezoey, line 106: why did this change? to inspect the type?

hasezoey (commented):
Looks OK, it is really not ideal to require explicit casting or redundant `as const` in some places, as long as it is limited to the string representation (of primitives) it should be OK

i have not yet tested this change (or 8.0 in general) against typegoose for type issues

test/types/virtuals.test.ts
vkarpov15: Unfortunately yes. I think that's more of a TypeScript 5 issue though, similar to #13514.

test/types/schema.test.ts
vkarpov15: This particular one isn't required, I'm changing.

Issue #13308, opened 2023-04-22
[meta] change file naming to be consistent


hasezoey:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

currently this project uses inconsistent file naming for the files in `lib/`, in use are:

- fully-lowercased (majority of the files)
  - `lib/connectionstate.js`
  - `lib/validoptions.js`
-  snake_cased
   - `lib/document_provider.js`
- camelCased
  - `lib/browserDocument.js`
- PascalCased
  - `lib/schema/SubdocumentPath.js`

this issue is to track renaming of those files in a consistent manner (but likely for some major version)

mjfwebb:
Hi @hasezoey, I noticed you used "consistenize" in the title, but that's actually a misspelling of the word "consistentize", which isn't actually a real word in English, in this case, you probably want to use "make consistent". I hope I'm not being too presumptuous writing this!

hasezoey:
@vkarpov15 as there is now a PR for this change already, should we change filenames to be consistently camelCased (which the majority of files already are and the PR changes some files to)?

PR #13857, opened 2023-09-12
BREAKING CHANGE: rename top-level files to camel case

vkarpov15:
**Summary**

Based on feedback from #13837, rename top-level files in `lib/` to use camel case. Camel case filenames is the path of least resistance since most of the project already uses camel case.

**Examples**

---

re #13308


hasezoey (approved):
Looks good, added a reference to #13308 in main issue body

PR #13868, opened 2023-09-16
docs: fix `pre` element overflow in home page

ghoshRitesh12:
**Summary**
When browsing through the docs of mongoose, I opened it's home page and noticed a horizontal overflow, caused by the `pre` element in the first code snippet, hence I thought of fixing it.

**Examples**
Previously (having overflow issue): 

image

Now (fixed overflow issue):

image


vkarpov15 (approved):
Thanks :+1:

PR #13888, opened 2023-09-22
Casting will now handle nested objects

### LINKED ISSUE #13880: Querying array of arrays with $elemMatch and $not throws a CastError
### Prerequisites

### Mongoose version

7.5.2

### Node.js version

18.2

### MongoDB server version

6.0.5

_No response_

### Description

When querying an array of arrays of strings field using a combination of `$elemMatch` and `$not`, mongoose's query casters throws an error.

The query is:
`{ arr: { $elemMatch: {$not: {$elemMatch: {$eq: '1'}}} } }`

Our intention is to obtain all documents that have one nested array that does not contain certain element (in this case, the string '1').

```
CastError: Cast to string failed for value "{ '$elemMatch': { '$eq': '1' } }" (type Object) at path "arr" for model "nested_arrays"
  at SchemaString.cast (/home/davids/Documents/MongooseType/node_modules/.pnpm/mongoose@7.5.2/node_modules/mongoose/lib/schema/string.js:606:11)
```

### Steps to Reproduce

This short code shows that the query itself works using mongo itself (using the `collection` property of the model) and that mongoose throws the cast error:

```

  arr: [[String]],
});

const model = mongoose.model('nested_arrays', schema);

mongoose.connect('mongodb://127.0.0.1:27017', { useNewUrlParser: true, useUnifiedTopology: true })
  .then(async () => {
    const obj = new model({ arr: [[1,2,3], [2,3,4]] })
    await obj.save();

    const query = { arr: { $elemMatch: {$not: {$elemMatch: {$eq: '1'}}} } };
    const mongo_res = await model.collection.find(query).toArray();
    console.log(mongo_res);

    const mongoose_res = await model.find(query);
    console.log(mongoose_res);
  });
```

The full output is really long, but the first lines are:
```
[
  {
    _id: new ObjectId("6509bfce3c3c67f9f8e099d6"),
    arr: [ [Array], [Array] ],
    __v: 0
  }
]
/home/davids/Documents/MongooseType/node_modules/.pnpm/mongoose@7.5.2/node_modules/mongoose/lib/schema/string.js:606
    throw new CastError('string', value, this.path, null, this);
          ^

CastError: Cast to string failed for value "{ '$elemMatch': { '$eq': '1' } }" (type Object) at path "arr" for model "nested_arrays"
```

### Expected Behavior

Using the `mongoose` `find` should return the same results than the `model.collection` when using the same query.

IslandRhythms:
closes #13880

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -412,8 +412,10 @@ function getStrictQuery(queryOptions, schemaUserProvidedOptions, schemaOptions, 
 
 // This function simply checks if the next value of the given object is also an object.
 // fix: what happens when obj is { key: { k: val } } gh3719
+// need to check if an object that the val is also not an object
 function checkNested(obj) {
   const vals = Object.values(obj);
+  console.log('what is vals', vals);
   for (let i = 0; i < vals.length; i++) {
     if (typeof vals[i] === 'object' && !Array.isArray(vals[i]) && vals[i] != null) {
       return true;
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -293,7 +293,6 @@ module.exports = function cast(schema, obj, options, context) {
         continue;
       } else if (utils.isPOJO(val)) {
         any$conditionals = Object.keys(val).some(isOperator);
-
         if (!any$conditionals) {
           obj[path] = schematype.castForQuery(
             null,
@@ -309,7 +308,9 @@ module.exports = function cast(schema, obj, options, context) {
           while (k--) {
             $cond = ks[k];
             nested = val[$cond];
-
+           if (Object.keys(nested).some(isOperator)) {
+            cast(schema, nested, options, context);
+           } else
             if ($cond === '$not') {
               if (nested && schematype) {
                 _keys = Object.keys(nested);
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -307,7 +307,6 @@ module.exports = function cast(schema, obj, options, context) {
           while (k--) {
             $cond = ks[k];
             nested = val[$cond];
-            console.log('what is nested', nested);
             if (Object.keys(nested || {}).some(isOperator) && checkNested(nested)) {
               cast(schema, nested, options, context);
             } else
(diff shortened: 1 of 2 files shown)


lib/cast.js
```diff
@@ -305,38 +304,39 @@ module.exports = function cast(schema, obj, options, context) {
...
-            if ($cond === '$not') {
-              if (nested && schematype) {
-                _keys = Object.keys(nested);
-                if (_keys.length && isOperator(_keys[0])) {
-                  for (const key in nested) {
-                    nested[key] = schematype.castForQuery(
-                      key,
-                      nested[key],
+            if (Object.keys(nested || {}).some(isOperator) && checkNested(nested)) {
+              cast(schema, nested, options, context);
```
vkarpov15, line 311: Maybe instead of casting the whole `nested` property, you instead loop through all the keys of `nested`, check if the key matches `isOperator`, and then cast that? That would save you from needing the `checkNested` function, which seems a bit unnecessary.

vkarpov15 (changes requested):
One minor suggestion to try, but overall looks good, good job

lib/cast.js
IslandRhythms: @vkarpov15  Trying you're recommended approach seems to move the PR back to square 1 as it cause the same twenty test failures that the `checkNested()` function attempted to resolve. Unless I messed up your recommendation.

My implementation of the recommendation:
```javascript
if (Object.keys(nested || {}).some(isOperator)) {
              const iterate = (obj) => {
                Object.keys(obj).forEach(key => {
                  if (isOperator(key)) {
                    cast(schema, obj[key], options, context)
                  }
                })
              };
              iterate(nested);
            } 
```

IslandRhythms: Upon further iteration the flow I have works because it catches the case where `nested = {$not: {$elemMatch: 'val'} }` and the like kicks it to the if statement on line 313

vkarpov15:
Superseded by #13893

PR #13893, opened 2023-09-25
fix: handle casting `$elemMatch` underneath `$not` underneath another `$elemMatch`

### LINKED ISSUE #13880: Querying array of arrays with $elemMatch and $not throws a CastError

vkarpov15:
Fix #13880

**Summary**

A simpler solution than #13888, went for a different branch name and different PR because branch names that end in `.js` can make our current eslint config screw up your git state.

**Examples**

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,7 +13,8 @@ module.exports = {
     '*.min.js',
     '**/docs/js/native.js',
     '!.*',
-    'node_modules'
+    'node_modules',
+    '.git'
   ],
   overrides: [
     {
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -309,8 +309,21 @@ module.exports = function cast(schema, obj, options, context) {
           while (k--) {
             $cond = ks[k];
             nested = val[$cond];
-
-            if ($cond === '$not') {
+            if ($cond === '$elemMatch') {
+              if (nested && schematype != null && schematype.schema != null) {
+                cast(schematype.schema, nested, options, context);
+              } else if (nested && schematype != null && schematype.$isMongooseArray) {
+                if (utils.isPOJO(nested) && nested.$not != null) {
+                  cast(schema, nested, options, context);
+                } else {
+                  val[$cond] = schematype.castForQuery(
+                    $cond,
+                    nested,
+                    context
+                  );
+                }
+              }
+            } else if ($cond === '$not') {
               if (nested && schematype) {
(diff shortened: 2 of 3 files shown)


IslandRhythms (approved):
LGTM

hasezoey (approved):

PR #13900, opened 2023-09-26
Add inferRawDocType helper

vkarpov15:
**Summary**

As a potential workaround for #13772, this PR adds an `inferRawDocType` helper analogous to `inferSchemaType` (and mostly a copy of `inferschematype.d.ts`) to make it easier to automatically infer the raw doc type from the schema definition. This would also help with #13523. With this change, you can overwrite `lean<>` result types, etc.

Not ideal, but at least provides a workaround to more easily infer the correct raw doc type from the schema definition.

**Examples**


vkarpov15:
Related issues: #12030

hasezoey (commented):
do i understand correctly that this type is for getting a POJO type from a schema definition type?

i dont really like to have to repeat similar code, but i guess this is necessary.

though is it really required that we re-define exact helper functions that (at least look to me) to be the same as in `inferschematype`?

vkarpov15:
A lot of the helper types are identical, I will see if there's some way to fix that.

The reason why we need a similar helper function is because of how `InferSchemaType` is structured, there's no way to pass a `UseHydratedDocument` switch to `InferSchemaType`. The problem right now is that `InferSchemaType` doesn't quite return the fully hydrated document type or the raw document type, but something in-between. `inferRawDocType` is an attempt to return the raw doc type, which I think is a good first step in the right direction. We've tried some alternative approaches, like #13856, but it's proving tricky to work around the inherent restrictions with `InferSchemaType`.

vkarpov15:
@hasezoey I found a way to reuse the helpers from `inferschematypes` :+1:

JavaScriptBach:
Isn't this just `ObtainDocumentType`? Or am I missing something?

vkarpov15:
@JavaScriptBach almost. The major difference is that `InferSchemaType` and `ObtainDocumentType` return a type that has Mongoose wrappers `Types.Array`, `Types.Subdocument`, and `Types.DocumentArray`. So not quite the lean document, but not quite the fully hydrated document either - still missing methods, virtuals, etc.

PR #13901, opened 2023-09-26
BREAKING CHANGE: allow null for optional fields in TypeScript

vkarpov15:
**Summary**

Apply #12748, #12781 to Mongoose 8. At runtime, Mongoose lets you store `null` for optional fields. TypeScript types should reflect this.

**Examples**

### DIFF
--- a/docs/migrating_to_8.md
+++ b/docs/migrating_to_8.md
@@ -16,6 +16,7 @@ If you're still on Mongoose 6.x or earlier, please read the [Mongoose 6.x to 7.x
 * [MongoDB Node Driver 6.0](#mongodb-node-driver-6)
 * [Removed `findOneAndRemove()`](#removed-findoneandremove)
 * [Removed id Setter](#removed-id-setter)
+* [Allow `null` For Optional Fields in TypeScript](#allow-null-for-optional-fields-in-typescript)
 
 <h2 id="removed-rawresult-option-for-findoneandupdate"><a href="#removed-rawresult-option-for-findoneandupdate">Removed <code>rawResult</code> option for <code>findOneAndUpdate()</code></a></h2>
 
@@ -60,3 +61,19 @@ Use `findOneAndDelete()` instead.
 
 In Mongoose 7.4, Mongoose introduced an `id` setter that made `doc.id = '0'.repeat(24)` equivalent to `doc._id = '0'.repeat(24)`.
 In Mongoose 8, that setter is now removed.
+
+<h2 id="allow-null-for-optional-fields-in-typescript"><a href="#allow-null-for-optional-fields-in-typescript">Allow <code>null</code> For Optional Fields in TypeScript</a></h2>
+
+In Mongoose 8, automatically inferred schema types in TypeScript allow `null` for optional fields.
+In Mongoose 7, optional fields only allowed `undefined`, not `null`.
+
+```typescript
+const schema = new Schema({ name: String });
+const TestModel = model('Test', schema);
+
+const doc = new TestModel();
+
+// In Mongoose 8, this type is `string | null | undefined`.
+// In Mongoose 7, this type is `string | undefined`
+doc.name;
+```
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -477,7 +477,7 @@ function gh12100() {
   const TestModel = model('test', schema_with_string_id);
   const obj = new TestModel();
 
-  expectType<string>(obj._id);
+  expectType<string | null>(obj._id);
 })();
 
 (async function gh12094() {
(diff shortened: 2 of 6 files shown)


test/types/models.test.ts
```diff
@@ -477,7 +477,7 @@ function gh12100() {
   const TestModel = model('test', schema_with_string_id);
   const obj = new TestModel();
 
-  expectType<string>(obj._id);
+  expectType<string | null>(obj._id);
```
hasezoey, line 480: why is `null` here but not `undefined`?

test/types/queries.test.ts
```diff
@@ -393,7 +393,8 @@ async function gh12342_manual() {
 
 async function gh12342_auto() {
   interface Project {
-    name?: string, stars?: number
+    name?: string | null,
+    stars?: number | null
```
hasezoey, line 397: is adding `| null` now required or is this just to test this case? (because the `?` should already add `| undefined`)

hasezoey (commented):
i have some questions, mainly is adding `| null` now required for `?: ` properties which should already have a `| undefined` added by `?`?

docs/migrating_to_8.md
```diff
@@ -59,4 +60,20 @@ Use `findOneAndDelete()` instead.
...
+```typescript
+const schema = new Schema({ name: String });
+const TestModel = model('Test', schema);
+
+const doc = new TestModel();
+
+// In Mongoose 8, this type is `string | null | undefined`.
+// In Mongoose 7, this type is `string | undefined`
+doc.name;
+```
```
hasezoey, line 79:
````suggestion
```

````

try fix md lint "should end in new-line" (i am not sure, because of githubs online editor showing ending new-lines

**Comment by vkarpov15** [comment]
@hasezoey  "is adding | null now required for ?:  properties which should already have a | undefined added by ??" <-- yes. `?:` just adds `| undefined`, `| null` is separate.

**Comment by hasezoey** [comment]
> ?: just adds | undefined, | null is separate.

i know that they are separate and i know that the "more correct" way is `|null|undefined`, but those that do not want to deal with `null` and only `undefined` (at least in terms of types) is it still required to be added? like is there some kind of type incompatibility to *not* add `|null`?

File: test/types/queries.test.ts (hunk 2 above)
Comment (vkarpov15) [reply test/types/queries.test.ts:397]: For this particular case, just for testing. If there are cases where you specifically define a type, like `const v: number | undefined = doc.stars`, or `function foo(): number | undefined { return doc.stars; }` then you will need to add `null`.

**Comment by vkarpov15** [comment]
@hasezoey I don't understand "those that do not want to deal with null and only undefined (at least in terms of types) is it still required to be added? like is there some kind of type incompatibility to not add |null?". Can you please clarify with some examples?

**Comment by hasezoey** [comment]
> I don't understand "those that do not want to deal with null and only undefined (at least in terms of types) is it still required to be added? like is there some kind of type incompatibility to not add |null?". Can you please clarify with some examples?

i mean is the following still possible (/ supported) without type errors?

```ts
interface SomeSchema {
  prop1?: string;
  prop2?: number;
}
// instead of now
interface SomeSchema {
  prop1?: string | null;
  prop2?: number | null;
}
```
which would trickle down to something like typegoose
```ts
class SomeSchema {
  @prop()
  public prop1?: string;

  @prop()
  public prop2?: number;
}
// instead of
class SomeSchema {
  @prop()
  public prop1?: string | null;

  @prop()
  public prop2?: number | null;
}
```

personally even if the more correct way would be `string | undefined | null`, i would go with `string | undefined` (unless `null` is explicitly needed) because that makes type debugging (like hover) a lot more cleaner, especially because i have common functions like `isNullOrUndefined` (which is deprecated in utils nodejs) which would take care of both at runtime, but would only need to deal with `undefined` for the "undefined" case, where essentially `undefined` and `null` are treated the same

TL;DR: `prop?: string` is a lot more cleaner overall than `prop?: string | null` (or the "expanded type" `prop: string | undefined | null`) and i am asking if the "cleaner" way is still supported or will it give type-errors?

vkarpov15:
@hasezoey you shouldn't need to change `prop?: string` to `prop?: string | null` in raw doc definitions. The following script compiles fine even with these changes.

```ts

interface IUser {
  prop?: string;
}

const schema = new mongoose.Schema<IUser>({ prop: String });
const TestModel = mongoose.model<IUser>('Test', schema);

const doc = new TestModel();
const foo: string | undefined | null = doc.prop;
```

Can you please try this PR out with Typegoose and see if anything breaks?

hasezoey:
> Can you please try this PR out with Typegoose and see if anything breaks?

seems to run fine (both runtime and types), though i will need to test it a little more in depth (against the 8.0 branch in general)

PR #13902, opened 2023-09-26
Allow importing Mongoose without any MongoDB specifics

vkarpov15:
**Summary**

We've been using this branch to make Mongoose playground work: https://github.com/mongoosejs/mongoose-playground/blob/b6032ccbbf11908ae45cfc8eb28077a5559bfbdf/package.json#L13 because otherwise there's no way to import fully fledged Mongoose in the browser.

This change is more a refactor than anything, but it is sufficiently significant that I wanted to hold off until Mongoose 8.

**Examples**


hasezoey (changes requested):
if i understand correctly, basically `lib/index.js` was moved to `lib/mongoose.js` and replaced with a shim for mongodb as the default driver

maybe consider adding some documentation on how to use it now without the mongo driver.

also consider cleaning up the git commits so that `index.js` -> `mongoose.js` is actually picked-up as a rename, this would mean not losing git history (blame) for the lines and also would make reviewing this pr a lot easier.

also with this change i think ESM compat will need to be tested again (wasnt this added at some point in time?)

vkarpov15:
You're right, the general idea is that `lib/index.js` is renamed `lib/mongoose.js`, with removal of `Mongoose.prototype.mongo`.

Re: ESM compat, that logic is in `index.js`, not `lib/index.js`, so shouldn't be affected.

PR #13903, opened 2023-09-27
BREAKING CHANGE: avoid saving changes to de-selected paths

### LINKED ISSUE #13145: Avoid saving changes to non-selected paths by default
### Prerequisites

### Issue

Re: #13062, #5800, #9457, #4651. Mongoose currently saves changes to non-selected paths. We should consider preventing that, like in this commit: https://github.com/Automattic/mongoose/pull/13143/commits/103f2bcae60fbc77dc0d7c05efb0d972d9542af4 .

You can always save changes to non-selected paths using `markModified()`. For example, `doc.links = undefined; doc.markModified('links');`. Just this change is a little too risky to make in a patch or even minor release, because we have tests that assert on the old behavior. And I'd like to consider potential tradeoffs more.

vkarpov15:
Fix #13145
Re: #13143
Re: #13062

**Summary**

Right now Mongoose saves changes to de-selected paths, and skips `required` validation on those paths. That can be problematic in the following case:

```javascript
      age: { type: Number, required: true, select: false },
      links: { type: String, required: true, select: false }
    });

    const Test = db.model('Test', testSchema);

    const { _id } = await Test.create({
      name: 'Test Testerson',
      age: 0,
      links: 'some init links'
    });

    let doc = await Test.findById(_id);
    doc.links = undefined;
    // `links` is now `undefined`, despite `required`
    await doc.save();
```

We've been considering making this change since [when we were working on Mongoose 5](https://github.com/Automattic/mongoose/issues/4651#issuecomment-258186695), and this issue continues to pop up every year or two; I think 8.0 is a good time to make this particular change.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1579,6 +1579,11 @@ Document.prototype.$__shouldModify = function(pathToMark, path, options, constru
   if (this.$isNew) {
     return true;
   }
+
+  if (!this.$__isSelected(path) && (!options || !options._forceModify)) {
+    return false;
+  }
+
   // Is path already modified? If so, always modify. We may unmark modified later.
   if (path in this.$__.activePaths.getStatePaths('modify')) {
     return true;
--- a/lib/helpers/timestamps/setDocumentTimestamps.js
+++ b/lib/helpers/timestamps/setDocumentTimestamps.js
@@ -21,6 +21,6 @@ module.exports = function setDocumentTimestamps(doc, timestampOption, currentTim
     if (doc.isNew && createdAt != null) {
       ts = doc.$__getValue(createdAt);
     }
-    doc.$set(updatedAt, ts);
+    doc.$set(updatedAt, ts, null, { _forceModify: true });
   }
 };
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -1298,16 +1298,6 @@ describe('document', function() {
     await user1.save();
     const user2 = await User.findById(user1).select('name').exec();
 
-    user2.req = undefined;
-    let err = await user2.save().then(() => null, err => err);
-    err = String(err);
-
-    const invalid = /Path `req` is required./.test(err);
-    assert.ok(invalid);
-
-    user2.req = 'it works again';
-    await user2.save();
-
     const user3 = await User.findById(user2).select('_id').exec();
     await user3.save();
   });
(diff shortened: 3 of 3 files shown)


AbdelrahmanHafez (commented):
In all my applications, all the queries I only `select` the fields that I need _to read_, and it's a common solution that I check some of the selected fields, and according to their values, I would write to an unselected field.

This change would break a lot of the behaviors in my apps and it would probably take a lot of effort from me to upgrade to mongoose v8.0. I imagine at least a few other people would have the same issue.

Looking at https://github.com/Automattic/mongoose/issues/13062 for example, I see that their complaint is missing validation on modified unselected field. For that I think the solution would be to still try to persist the modified value even if it's unselected, but run the validations for it if it's modified.

Looking at https://github.com/Automattic/mongoose/issues/4651, it seems like either way could be right. I would expect that `doc.set('ball.kind', 'shuttlestock')` to be translated into `{ $set: { 'ball.kind': 'shuttlestock' } }`, but I can imagine that there are some scenarios where we would _have_ to translate it to a full-fledged object.


If it were for #13062 only, I would try to only fix the validation when saving on an unselected field, because we validate against the newly-set value, not the value coming from the DB.

#4651 complicates things a little bit more, but then again it's been the behavior for 7 years and I am not sure it's problematic for so many people.

All-in-all, my _preference_ would be to not go in this direction because it'll make it a very costly upgrade mongoose in my apps, if we could have a global option that persists the old behavior, that would be nice.

hasezoey:
i have not really made use of "modifying un-selected paths", but i agree that "just skipping them" is not a great solution, it should still save, but actually run the validators, maybe make it configurable with options `runValidators`, `throw` and `ignore` as a schema option and/or global option?

vkarpov15:
@hasezoey @AbdelrahmanHafez I took a closer look, and actually it looks like the issue I mentioned in [first post](https://github.com/Automattic/mongoose/pull/13903#issue-1916142575) is not an issue. I pushed a test confirming that validation _does_ run if you're saving changes to de-selected paths. The issue in [this comment](https://github.com/Automattic/mongoose/issues/13062#issuecomment-1457041670) is primarily due to setting `links = undefined` in a pre('save') hook.

PR #13905, opened 2023-09-28
Allow importing Mongoose without any MongoDB specifics, take 2

vkarpov15:
**Summary**

#13902, but with cleaner git history.

We've been using this branch to make Mongoose playground work: https://github.com/mongoosejs/mongoose-playground/blob/b6032ccbbf11908ae45cfc8eb28077a5559bfbdf/package.json#L13 because otherwise there's no way to import fully fledged Mongoose in the browser.

This change is more a refactor than anything, but it is sufficiently significant that I wanted to hold off until Mongoose 8.

**Examples**


hasezoey (approved):
LGTM, a lot cleaner git history, though it seems like comparing HEAD to 8.0 still does not recognize the rename and making it more annoying to compare (but going commit by commit works)

just to confirm, the changes in `lib/query.js` are to make things consistent or was there a specific problem for this change? (the `_collection.collection` -> `mongooseCollection` change)

vkarpov15:
"just to confirm, the changes in lib/query.js are to make things consistent or was there a specific problem for this change? (the _collection.collection -> mongooseCollection change)" I don't remember the exact reasoning. The difference is that `this._collection` is an instance of `mquery.Collection`, whereas `this.mongooseCollection` is an instance of Mongoose's collection class. So I suspect there was some quirk with mquery collections that I didn't want to work around. Either way, I think using `mongooseCollection` is a maintainability improvement as we try to move away from mquery.

PR #13928, opened 2023-10-03
fix(query): handle `$round` in `$expr` as array

### LINKED ISSUE #13881: Cannot use $round in find with $expr
### Prerequisites

### Mongoose version

7.5.2

### Node.js version

18.17.1

### MongoDB server version

6.0.6

_No response_

### Description

$round in an aggregation or within $expr takes the form
`{ $round : [ <number>, <place> ] }` [MongoDB round doc](https://www.mongodb.com/docs/manual/reference/operator/aggregation/round)

However,
`Test.find({ $expr: { $eq: [{ $round: ['$value', 0] }, 2] } })`
fails with _CastError: Cast to Number failed for value "[ '$value', 0 ]" (type Array) at path "undefined" for model "Test"_

In the mongosh, the command works fine:
`db.testcol.find({ $expr: { $eq: [{ $round: ['$value', 0] }, 2] } })`

We can only use $round to round a value to an integer but cannot specify the number of digits to round to.
($round is treated as a `arithmeticOperatorNumber` in `cast$expr.js`)

### Steps to Reproduce

Repro script:

``` JS
/* eslint-disable no-console */

const { Schema } = mongoose;

async function run() {
    // const dburl = 'mongodb://localhost:19999/test';
    const dburl = 'mongodb://127.0.0.1:21121/test?directConnection=true';
    await mongoose.connect(dburl);

    const testSchema = new Schema({
        value: Number,
    });
    const Test = mongoose.model('Test', testSchema);

    await Test.create({ value: 2 });
    await Test.create({ value: 2.2 });

    // normal find with $expr works OK
    const testExprWithFindOK = await Test.find(
        { $expr: { $eq: ['$value', 2] } },
    );
    assert.equal(testExprWithFindOK.length, 1);

    // aggregate with $expr and $round works OK
    const testRoundWithAggregateOK = await Test.aggregate([
        { $match: { $expr: { $eq: [{ $round: ['$value', 0] }, 2] } } },
    ]);
    assert.equal(testRoundWithAggregateOK.length, 2);

    // find with $expr and $round (only) works with a number OK
    const testRoundNumberWithFindOK = await Test.aggregate([
        { $match: { $expr: { $eq: [{ $round: '$value' }, 2] } } },
    ]);
    assert.equal(testRoundNumberWithFindOK.length, 2);

    // find with $expr and $round FAILS
    // $round expects a number, not ['$value', 0] as described in
    // https://www.mongodb.com/docs/manual/reference/operator/aggregation/round
    try {
        const testRoundWithFindFAIL = await Test.find(
            { $expr: { $eq: [{ $round: ['$value', 0] }, 2] } },
        );
        assert.equal(testRoundWithFindFAIL.length, 2);
    } catch (err) {
        assert.fail(err.toString().substr(0, 200));
    }
}

run().catch(console.error);

```

### Expected Behavior

$round within an $expr should work the same in a find query as in an aggregate query.

vkarpov15:
Fix #13881

**Summary**

[`$round` syntax takes an array, not a number](https://www.mongodb.com/docs/manual/reference/operator/aggregation/round/). This PR fixes that.

**Examples**

### DIFF
--- a/lib/helpers/query/cast$expr.js
+++ b/lib/helpers/query/cast$expr.js
@@ -119,13 +119,10 @@ function _castExpression(val, schema, strictQuery) {
   }
   if (val.$round) {
     const $round = val.$round;
-    if (!Array.isArray($round) || $round.length !== 2) {
+    if (!Array.isArray($round) || $round.length < 1 || $round.length > 2) {
       throw new CastError('Array', $round, '$round');
     }
-    val.$round = [
-      castNumberOperator($round[0], schema, strictQuery),
-      castNumberOperator($round[1], schema, strictQuery)
-    ];
+    val.$round = $round.map(v => castNumberOperator(v, schema, strictQuery));
   }
 
   _omitUndefined(val);
--- a/lib/helpers/query/cast$expr.js
+++ b/lib/helpers/query/cast$expr.js
@@ -29,7 +29,6 @@ const arithmeticOperatorNumber = new Set([
   '$floor',
   '$ln',
   '$log10',
-  '$round',
   '$sqrt',
   '$sin',
   '$cos',
@@ -117,6 +116,16 @@ function _castExpression(val, schema, strictQuery) {
   }
   if (val.$size) {
     val.$size = castNumberOperator(val.$size, schema, strictQuery);
+  }
+  if (val.$round) {
+    const $round = val.$round;
+    if (!Array.isArray($round) || $round.length !== 2) {
+      throw new CastError('Array', $round, '$round');
+    }
+    val.$round = [
+      castNumberOperator($round[0], schema, strictQuery),
+      castNumberOperator($round[1], schema, strictQuery)
+    ];
   }
 
   _omitUndefined(val);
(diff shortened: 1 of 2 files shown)


lib/helpers/query/cast$expr.js
```diff
@@ -118,6 +117,16 @@ function _castExpression(val, schema, strictQuery) {
   if (val.$size) {
     val.$size = castNumberOperator(val.$size, schema, strictQuery);
   }
+  if (val.$round) {
+    const $round = val.$round;
+    if (!Array.isArray($round) || $round.length !== 2) {
```
hasezoey, line 122: isnt `<place>` (the second element in the array) optional as per the docs, or am i misunderstanding this condition (and later setting)?

> <place> | integer | Optional Can be any valid [expression](https://www.mongodb.com/docs/manual/meta/aggregation-quick-reference/#std-label-aggregation-expressions) that resolves to an integer between -20 and 100, exclusive. e.g. -20 < place < 100. Defaults to 0 if unspecified.

[source](https://www.mongodb.com/docs/manual/reference/operator/aggregation/round/)

hasezoey (changes requested):
looks good to me, though from what i read `<place>`(second array element) is optional, where this change makes it seem to be required

IslandRhythms (approved):
LGTM

Issue #13930, opened 2023-10-04
Typescript: InsertMany missing array overload with options


t1bb4r:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.12.0

### Node.js version

16.20.1

### MongoDB server version

4.4.x

### Typescript version (if applicable)

4.9.5

### Description

Using any options (`{ordered: false}`) with Model.InsertMany causes the return type to change. In the example I've forced the type parameter to cause a typescript error, but this also causes a problem without it. As we are inserting an array but there is no overload for this, it defaults to the 'single' insertion response type which is incorrect.

The issue is also present in mongoose 7.

https://github.com/Automattic/mongoose/issues/13904 seems closely related but not exactly, so I've decided to open a new issue.

### Steps to Reproduce

```
export interface IUser {
  name: string;
}

export const usersSchema = new mongoose.Schema({
  name: {
    type: String,
  },
});

const Users = mongoose.model<IUser>('Users', usersSchema);

// insertOrderedSingle
Users.insertMany<{ name: string; }>({ name: 'Zildjian' });
// insertArray
Users.insertMany<{ name: string; }>([{ name: 'Zildjian' }]);
// insertArrayWithOptions
Users.insertMany<{ name: string; }>([{ name: 'Zildjian' }], {});
```

### Expected Behavior

Normal options should not change the return type of InsertMany

IslandRhythms:
```typescript
import * as mongoose from 'mongoose';

interface IUser {
  name: string;
}

const usersSchema = new mongoose.Schema({
  name: {
    type: String,
  },
});

const Users = mongoose.model<IUser>('Users', usersSchema);

// insertOrderedSingle
Users.insertMany<{ name: string; }>({ name: 'Zildjian' });
// insertArray
Users.insertMany<{ name: string; }>([{ name: 'Zildjian' }]);
// insertArrayWithOptions
Users.insertMany<{ name: string; }>([{ name: 'Zildjian' }], {});
```

t1bb4r:
Fixed by https://github.com/Automattic/mongoose/pull/13931

PR #13931, opened 2023-10-04
fix(types): add insertMany array overload with options

t1bb4r:
**Summary**
This solves a typescript problem we are having with using options and insertMany, issue https://github.com/Automattic/mongoose/issues/13930. In this PR I have added an additional overload when inserting an array with options.

**Examples**
See the test case and bug report for examples


vkarpov15 (approved):
Thanks :+1:

PR #13932, opened 2023-10-04
fix(QueryCursor): avoid double-applying schema paths so you can include `select: false` fields with `+` projection using cursors

### LINKED ISSUE #13773: Projection with '+' not including field (since v7.1.1)
### Prerequisites

### Mongoose version

7.1.1

### Node.js version

16.x, 18.x

### MongoDB server version

4.4

_No response_

### Description

This issue is very similar to this: https://github.com/Automattic/mongoose/issues/13340

7.1.1 is the first version that is bad. I have tested that 7.1.0 works, and that the latest version (7.4.4) also fails.

I have a Schema where one sub-document is not selected by default (`select: false` in schema definition). I have queries that needs this sub-document and therefore select it explicitly with the "+" syntax like so: `.select("+myField")`.
The problem is that this field is not returned although being explicitly selected.

Note that one difference from #13340 is that now the other fields are included, just not the one that is selected. In #13340 only the _id was returned – no other fields.

### Steps to Reproduce

This can be reproduced by running the following code with mongoose >= 7.1.1. See that age is not returned. 

If you use mongoose 7.1.0, age is included in the response as expected.

```js

async function main() {

  const kittySchema = new mongoose.Schema({
      name: String,
      age: {
        select: false,
        type: Number
      }
    });

  const Kitten = mongoose.model('Kitten', kittySchema);
  const silence = new Kitten({ name: 'Silence', age: 2 });
  await silence.save()

  const updated = await Kitten.findByIdAndUpdate(silence._id, { age: 3 })
    .select("+age")

  console.log("UPDATED", updated) // <-- Returns _id, name and __v, but not age!
}

main()
```

Result from the console.log using mongoose 7.1.1:
```
UPDATED {
  _id: new ObjectId("64e674ea250a3fbd6b986c1c"),
  name: 'Silence',
  __v: 0
}
```

Result from the console.log using mongoose 7.1.0:
```
UPDATED {
  _id: new ObjectId("64e676bfd40527d1db7d3166"),
  name: 'Silence',
  age: 2,
  __v: 0
}
```

### Expected Behavior

When using `.select('+someField')`, `someField` should be included in the response.

vkarpov15:
Fix #13773

**Summary**

Bug fix, double applying `this._optionsForExec()` means projections lose `+age` projections that override `age: { select: false }` in schema. This double-applying only happens with `.cursor()` right now.

**Examples**

### DIFF
--- a/lib/cursor/QueryCursor.js
+++ b/lib/cursor/QueryCursor.js
@@ -34,7 +34,7 @@ const util = require('util');
  * @api public
  */
 
-function QueryCursor(query, options) {
+function QueryCursor(query) {
   // set autoDestroy=true because on node 12 it's by default false
   // gh-10902 need autoDestroy to destroy correctly and emit 'close' event
   Readable.call(this, { autoDestroy: true, objectMode: true });
@@ -46,7 +46,7 @@ function QueryCursor(query, options) {
   this._mongooseOptions = {};
   this._transforms = [];
   this.model = model;
-  this.options = options || {};
+  this.options = {};
   model.hooks.execPre('find', query, (err) => {
     if (err != null) {
       if (err instanceof kareem.skipWrappedFunction) {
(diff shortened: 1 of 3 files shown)


IslandRhythms (approved):
LGTM

PR #13937, opened 2023-10-05
8.0

vkarpov15:
**Summary**

8.0 is ready to go :+1: mostly relatively minor housekeeping changes. I think the biggest one will be allowing `null` for optional fields in TypeScript in #13901. 

**Examples**


hasezoey:
i now have tested mongoose 8.0 with typegoose, where everything passes (both runtime and types), the steps i needed to take:
- apply #13950
- update the check again `Mixed` to use `schemaName` instead of `name` (or alternatively check against `SchemaMixed`)

liviudobrea:
Let's goo boys, I need bson 6.x in my life 🙈 @vkarpov15

IslandRhythms (approved):
LGTM

PR #13946, opened 2023-10-07
chore(npmignore): ignore newer files

hasezoey:
**Summary**

This PR updates the `.npmignore` to include some files which were added since the last touch to npmignore, mainly:
- transition from `eslintrc.json` to `.eslintrc.js`
- addition of markdownlint-cli
- addition of lgtm
- move of scripts to `scripts`
- addition of `.mocharc.yml`

also excluded `tools/`, because to my knowledge these are not meant to be shipped, another exclude is `tsconfig.json`

see https://unpkg.com/browse/mongoose@7.6.0/ for context

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -47,4 +47,12 @@ webpack.base.config.js
 notes.md
 list.out
 
-eslintrc.json
+# config files
+lgtm.yml
+.mocharc.yml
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

Issue #13874, opened 2023-09-18
Using MongooseMap causes  memory leak on model.schema.subpaths 


mistyByme:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.5.0

### Node.js version

18.16.0

### MongoDB server version

5.0

### Typescript version (if applicable)

4.9.4

### Description

model fields for model.schema.subpaths will collect  the paht of  map childSchema ,it will make the memory bigger and bigger if path changed.

### Steps to Reproduce

```Typescript
import mongoose from 'mongoose';

mongoose.connect('mongodb://localhost:27017', { dbName: 'test' });
const tradeSchema = new mongoose.Schema({ tradeId: Number, content: String });

const testSchema = new mongoose.Schema(
    {
        userId: Number,
        tradeMap: {
            type: Map,
            of: tradeSchema,
        },
    },
);

const TestModel = mongoose.model('Test', testSchema);

main();

async function test() {
    const userId = 100;
    const user = await TestModel.create({ userId, tradeMap: new Map() });

    //add subDoc
    for (let id = 1; id <= 10; id++) {
        const trade = { tradeId: id, content: 'test' };
        user.tradeMap.set(trade.tradeId.toString(), trade);
    }
    await user.save();
    await TestModel.deleteOne({ userId });
}

async function main() {
    await test();
    //print TestModel.schema.subpaths
    console.log((TestModel.schema as any).subpaths);
}
    /* output
    {
  'tradeMap.1._id': undefined,
  'tradeMap.2._id': undefined,
  'tradeMap.3._id': undefined,
  'tradeMap.4._id': undefined,
  'tradeMap.5._id': undefined,
  'tradeMap.6._id': undefined,
  'tradeMap.7._id': undefined,
  'tradeMap.8._id': undefined,
  'tradeMap.9._id': undefined,
  'tradeMap.10._id': undefined,
  'tradeMap.1.tradeId': undefined,
  'tradeMap.1.content': undefined,
  'tradeMap.2.tradeId': undefined,
  'tradeMap.2.content': undefined,
  'tradeMap.3.tradeId': undefined,
  'tradeMap.3.content': undefined,
  'tradeMap.6.content': undefined,
  'tradeMap.7.tradeId': undefined,
  'tradeMap.7.content': undefined,
  'tradeMap.8.tradeId': undefined,
  'tradeMap.8.content': undefined,
[...]

Issue #3265, opened 2015-08-13
$set doesn't update subdocument


jeffwhelpley:
When I do this in mongoose 4.1.2:

```
user.findOneAndUpdate({ _id: someid }, { $set: { field1: 'val1', field2: { subfield1: 'sub1', subfield2: 'sub2' }}});
```

field1 is updated, but field2 is not. When I downgrade to mongoose 4.0.6 this works fine.

vkarpov15:
What does your schema look like?

jeffwhelpley:
The code example I gave is fake. Here is a real example:

```
var postSchema = {
  title: String,
  answers: [{
    details: String,
    stats: {
      votes: Number,
      count: Number
    }
  }]
};

// later on in code
post.findOneAndUpdate(
  { _id: somePostId, answers._id: someAnswerId }, 
  { $set: { 'answers.$': { details: 'blah', stats: { votes: 1, count: 3 }}}}
);
```

So, in this example in 4.1.2 details is correctly updated to blah, but the stats field is completely removed from the answer. Even if there was a value in there before, its like stats was $unset or something. In 4.0.6, the details and stats are set correctly.

PR #13953, opened 2023-10-08
Avoid storing a separate entry in schema `subpaths` for every element in an array

vkarpov15:
**Summary**

In fix for #3265, we made it so that we store an entry in `subpaths` for every element in an array, presumably to help make it easier to look up array paths. With a little bit of changes, it turns out this is unnecessary; and, as #13874 pointed out, adding an entry to `subpaths` for every element in an array can be considered a memory leak.

**Examples**

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1017,7 +1017,7 @@ Schema.prototype.path = function(path, obj) {
 
     // subpaths?
     return /\.\d+\.?.*$/.test(path)
-      ? getPositionalPath(this, path)
+      ? getPositionalPath(this, path, cleanPath)
       : undefined;
   }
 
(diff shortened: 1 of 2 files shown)


lib/schema.js
```diff
@@ -2638,6 +2638,9 @@ Schema.prototype._getSchema = function(path) {
       // Re: gh-5628, because `schema.path()` doesn't take $ into account.
       parts[i] = '0';
     }
+    if (/^\d+$/.test(parts[i])) {
```
Uzlopak, line 2641: extracting a regex into a variable outside the calling function should be more performant.

hasezoey (commented):
LGTM (aside from the other threads)

IslandRhythms (approved):
LGTM. Be sure to fix the lint.

PR #13954, opened 2023-10-09
fix: raw result deprecation message

simllll:
according to https://mongoosejs.com/docs/deprecations.html#rawresult it mus tbe true, not false. it would also make more sense tbh


vkarpov15 (approved):
Thanks :+1:

PR #13960, opened 2023-10-09
types(schematypes): allow defining map path using `type: 'Map'` in addition to `type: Map`

### LINKED ISSUE #13755: Support string initialization for map in schema
### Prerequisites

### 🚀 Feature Proposal

My project uses String for setting the type in schema of each keys like this:
```js
username: { type: 'string', required: true }
```

In case of map, I just tried this style of code:
```js
instance: {
  type: Schema.Types.Map,
  of: Schema.Types.Mixed,
  default: new Map<string, any>(),
}
```
This is working code. But if I use 'Map' instead of `Schema.Types.Map`, compiler aborts to compile.
```js
instance: {
  type: 'Map',
  of: 'Mixed',
  default: new Map<string, any>(),
}
```

So, is it able to allow with setting the map type with String?

### Motivation

_No response_

### Example

_No response_

vkarpov15:
Fix #13755

**Summary**

`type: 'Map'` is valid in JavaScript, but TypeScript types don't currently support that. This PR makes TypeScript recognize `type: 'Map'` in schemas.

**Examples**

### DIFF
--- a/test/types/maps.test.ts
+++ b/test/types/maps.test.ts
@@ -71,3 +71,19 @@ function gh10872(): void {
 
   doc.toJSON().map1.foo;
 }
+
+function gh13755() {
+  class Test {
+    instance: Map<string, string>;
+    constructor() {
+      this.instance = new Map<string, string>();
+    }
+  }
+
+  const testSchema = new Schema<Test>({
+    instance: {
+      type: 'Map',
+      of: 'Mixed'
+    }
+  });
+}
--- a/test/types/maps.test.ts
+++ b/test/types/maps.test.ts
@@ -1,4 +1,5 @@
 import { Schema, model, Document, Model, Types } from 'mongoose';
+import { expectType } from 'tsd';
 
 interface ITest {
   map1: Map<string, number>,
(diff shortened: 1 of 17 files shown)


AbdelrahmanHafez (approved):

test/types/maps.test.ts
```diff
@@ -71,3 +71,19 @@ function gh10872(): void {
...
+
+function gh13755() {
+  class Test {
+    instance: Map<string, string>;
+    constructor() {
+      this.instance = new Map<string, string>();
+    }
+  }
+
+  const testSchema = new Schema<Test>({
```
hasezoey, line 83: i think this test is not quite obvious what it is testing and does not actually test what this fixes (add a expectType?), a different example:

(mongoose 7.6.1)

```ts
const sch = new mongoose.Schema({
  prop: {
    type: 'Map',
    of: 'String',
  },
  prop2: {
    type: 'String',
  },
});

const model = mongoose.model('Test', sch);

const doc = new model();
doc.prop; // Type: unknown
doc.prop2; // Type: string | undefined
```

removing the second property, the type of `prop` becomes:
```txt
prop: {
    type: "Map";
    of: "String";
}
```

and adding the class to this, would only result in a error that the `prop` types are incompatible:
```txt
Type '{ type: string; of: string; }' is not assignable to type 'SchemaDefinitionProperty<Map<string, string>, Test> | undefined'.
  Types of property 'type' are incompatible.
    Type 'string' is not assignable to type 'typeof Mixed | { readonly prototype?: SchemaDefinitionProperty<Map<any, any>, any> | undefined; readonly [Symbol.species]?: SchemaDefinitionProperty<MapConstructor, any> | undefined; } | undefined'.ts(2322)
```

vkarpov15: Good call, fixed in https://github.com/Automattic/mongoose/pull/13960/commits/c4416a60ce596be64829619e48f8fabadfdab0f6

PR #13963, opened 2023-10-10
fix(document): allow calling `$model()` with no args for TypeScript

### LINKED ISSUE #13878: How to get the model name/collection name from a document middleware?
### Prerequisites

### Mongoose version

6.12.0

### Node.js version

18

### MongoDB version

6

### Operating system

None

_No response_

### Issue

I'm trying to write a query/document middleware that logs the model name + collection name in certain circumstances.

I looked at https://mongoosejs.com/docs/6.x/docs/api/document.html, and the first problem I ran into is that model/collection name are not documented in the types.

Then I tried writing:
```typescript
    let modelName = "";
    let collectionName = "";
    if (this instanceof Query) {
      modelName = this.model.modelName;
      collectionName = this.model.collection.name;
    } else if (this instanceof Document) {
      let doc = this;
      // Subdocuments seem to not have model/collection names so keep going up to the parent
      while (isSome(this.$parent()) && doc !== this.$parent()) {
        doc = this.$parent() as Document;
      }
      modelName = (doc.constructor as any).modelName;
      collectionName = doc.collection.name;
    }
```

But this does not work either in the document case, sometimes I get `TypeError: Cannot read properties of undefined (reading 'name')` because `doc.collection` is sometimes undefined. It seems like subdocuments also behave slightly differently than documents in this case.

Surely documents should know which model they came from, so there must be some way to do this. Help would be appreciated; thanks!

vkarpov15:
Fix #13878

**Summary**

TypeScript doesn't allow specifying a type for the `constructor` property, so right now there's no ergonomic way to get the model associated with a document in TypeScript. Need to use `as`. This PR makes it so that you can do `doc.$model()`. Requires a small runtime change, Mongoose doesn't currently support calling `doc.$model()` without args.

**Examples**

### DIFF
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -75,6 +75,7 @@ declare module 'mongoose' {
 
     /** Returns the model with the given name on this document's associated connection. */
     $model<ModelType = Model<unknown>>(name: string): ModelType;
+    $model<ModelType = Model<DocType>>(): ModelType;
 
     /**
      * A string containing the current operation that Mongoose is executing
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -192,6 +192,10 @@ declare module 'mongoose' {
     markModified<T extends keyof DocType>(path: T, scope?: any): void;
     markModified(path: string, scope?: any): void;
 
+    /** Returns the model with the given name on this document's associated connection. */
+    model<ModelType = Model<unknown>>(name: string): ModelType;
+    model<ModelType = Model<DocType>>(): ModelType;
+
     /** Returns the list of paths that have been modified. */
     modifiedPaths(options?: { includeChildren?: boolean }): Array<string>;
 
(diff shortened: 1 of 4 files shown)


types/document.d.ts
```diff
@@ -75,6 +75,7 @@ declare module 'mongoose' {
 
     /** Returns the model with the given name on this document's associated connection. */
     $model<ModelType = Model<unknown>>(name: string): ModelType;
+    $model<ModelType = Model<DocType>>(): ModelType;
```
hasezoey, line 78: i dont see function `model` being available in the types, is this something that is forgotten or is there a reason it is not there?

hasezoey (approved):
LGTM, aside from a somewhat unrelated question

types/document.d.ts
vkarpov15: I don't know of any reason. My best guess would be to avoid issues back when we recommended using `& Document`, but don't see any reason why not. Looks like the tests pass with `model()` signature that matches `$model()`, so I will add that into this PR.

hasezoey (approved):

PR #13964, opened 2023-10-10
types(models): add cleaner type definitions for `insertMany()` with no generics to prevent errors when using `insertMany()` in generic classes

### LINKED ISSUE #13957: insertMany return type changed from version 7.3.2 to 7.3.3
### Prerequisites

### Mongoose version

>=7.3.3

### Node.js version

18.x

### MongoDB version

6.x

### Operating system

Windows

_No response_

### Issue

It seems that the return type of the `insertMany` function changed with version 7.3.3. 
The following code runs without any error on v7.3.2 and lower
```

export class RepositoryBase<T> {
    protected model: mongoose.Model<T & mongoose.Document>;

    constructor(schemaModel: mongoose.Model<T & mongoose.Document>) {
        this.model = schemaModel;
    }

    async insertMany(elems: T[]): Promise<T[]> {
        elems = await this.model.insertMany(elems);
        return elems;
    }
}
```
while on versions >=7.3.3 the assignment inside the function generates this error: 
image

The typescript version is 5.2.2.

vkarpov15:
Fix #13957

**Summary**

Something in the way `insertMany()` generics are structured causes the errors in #13957. Adding `insertMany()` overrides that have no generics fixes that, which seems like a reasonable approach.

**Examples**

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -709,3 +709,27 @@ async function gh13746() {
   expectType<ObjectId | undefined>(findOneAndUpdateRes.lastErrorObject?.upserted);
   expectType<OkType>(findOneAndUpdateRes.ok);
 }
+
+function gh13957() {
+  class RepositoryBase<T> {
+    protected model: mongoose.Model<T>;
+
+    constructor(schemaModel: mongoose.Model<T>) {
+      this.model = schemaModel;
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
Magic

PR #13984, opened 2023-10-16
fix(populate): handle multiple spaces when specifying paths to populate using space-delimited paths

### LINKED ISSUE #13951: Populate with spaces get evaluated Cannot populate path ``  because it is not in your schema. Set the strictPopulate option to false to override.
### Discussed in https://github.com/Automattic/mongoose/discussions/13939

<sup>Originally posted by **masterbater** October  5, 2023</sup>
Cannot populate path `` because it is not in your schema. Set the `strictPopulate` option to false to override.
If you put two spaces between populate I got the error above
```typescript
 const readings = await Gwapo.find(conditions)
    .populate(
      "user  testgwapo",
      "fname lname isActive height abbrev "
    )
```
`   "mongoose": "^6.12.0",`</div>

vkarpov15:
Fix #13951

**Summary**

Right now, `populate('user  fans')` will fail with a StrictPopulateError because of the extra space. I think it is sufficiently reasonable to split by multiple spaces, so `populate('user  fans')` is equivalent to `populate('user fans')`. This may cause issues in cases where path names have whitespace characters, but given that it isn't currently possible to populate a space-delimited path anyway, I don't think that is concerning.

**Examples**

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -31,6 +31,9 @@ exports.isMongooseDocumentArray = isMongooseDocumentArray.isMongooseDocumentArra
 exports.registerMongooseArray = isMongooseArray.registerMongooseArray;
 exports.registerMongooseDocumentArray = isMongooseDocumentArray.registerMongooseDocumentArray;
 
+const oneSpaceRE = /\s/;
+const manySpaceRE = /\s+/;
+
 /**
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM, seems like it also fixes a bug that checked for tabs but did not actually split by them

PR #13989, opened 2023-10-17
Remove `overwrite` option for `updateOne()`, `findOneAndUpdate()`, etc.

vkarpov15:
**Summary**

Re: #13578, remove the `overwrite` option for 8.0. In 7.x, `Model.findOneAndUpdate(filter, update, { overwrite: true })` becomes a `findOneAndReplace()`, which is a vestigial quirk of back when `update()` with no atomic operators would overwrite the document. Given that's no longer the case, we don't really need an `overwrite` option anymore.

Removed a few tests that relied on `overwrite: true`.

**Examples**


IslandRhythms (approved):
LGTM minus the failing typescript test.

hasezoey (changes requested):
LGTM, but there are some things missing:
- ~~remove `overwrite` option from typescript types~~ (seems like option `overwrite` never even existed in the types)
- add changelog entry about removal

the current typescript types ci failing is not related to this PR, it has been introduced into the 8.0 branch by 81cba560f542618b20386d3b8d58b0284fa48409

vkarpov15:
@hasezoey can you re-review please?

hasezoey (approved):

PR #13990, opened 2023-10-17
types(models): make all properties to Model constructor optional

### LINKED ISSUE #13897: Where to put createdAt and updatedAt members when using timestamps?
### Prerequisites

### Mongoose version

6.5.2

### Node.js version

18.12.1

### MongoDB version

6.0.5

### Operating system

Windows

11

### Issue

I know this has been asked previously but reviewing the various other issues I could not find a suitable solution.

In the https://mongoosejs.com/docs/typescript.html the steps to create interfaces and schemas is pretty straightforward. But things seem to break down when `timestamps: true` is being used.

A simplified example below:
```
import {Schema, model, HydratedDocument} from 'mongoose';

interface IDocument {
  name: string
}

const documentSchema = new Schema<IDocument >({
  name: { type: String, required: true }
},
{
  timestamps: true
});

const Document = model<IDocument>('Document', documentSchema);

type DocumentDoc = HydratedDocument<IDocument>;

// Create & save
const newDoc = new Document<IDocument>({
  name: 'Foo'
});

newDoc.save();

// Read it back
const hydratedDocument = (await Document.find())[0];

// The following line does not compile since the interface definition does not have 'createdAt'
// although the instance itself does
console.log(`This document was created at '${hydratedDocument.createdAt}'`);
```

Attempting to fix this at the interface level does not work:
```
interface IDocument {
  name: string;
  createdAt: Date;
  updatedAt: Date;
}

// Now this will not compile as 'createdAt' and 'updatedAt' have not been defined in the data-to-be-stored
const newDoc = new Document<IDocument>({
  name: 'Foo'
});
```
No matter what I've tried (methods & overrides, InferSchemaType etc.) they all seem to fall flat on their face in either one of these problems:
- `new Document<IDocument>(...)` will not compile as the 'createdAt' and 'updatedAt' have not been given
- `console.log(hydratedDocument.createdAt)` does not compile because the type does not have the property

The following circumventions I know:
- Omit the `IDocument` part from the `new Document(...)`; this kind of defeats the purpose of type-safety
- Extend the HydratedDocument type with `SchemaTimestampConfigs`; this requires explicitly typing whatever is returned from `Document.find()` into `DocumentDoc | null`. It works but it is a bit unwieldy.

Are there any other options I should or could try? What is the "best" way that results in clean, readable code?

vkarpov15:
Fix #13897
Re: #11563

**Summary**

Starting with https://github.com/Automattic/mongoose/commit/4a5d712058e109a3db0250bded38524dbecb79c2, the `Model` constructor requires all paths, which #13897 pointed out is problematic for timestamps. This PR fixes that.

**Examples**


test/types/models.test.ts
```diff
@@ -751,3 +751,22 @@ function gh13957() {
...
+
+  const documentSchema = new Schema<IDocument>({
+    name: { type: String, required: true }
+  },
+  {
+    timestamps: true
+  });
+
+  const Document = model<IDocument>('Document', documentSchema);
+  const doc = new Document({ name: 'foo' });
```
hasezoey, line 770: if i understand correctly, before this change this would complain that `createdAt` and `updatedAt` are missing?

if yes, with the way it is currently implemented, maybe add a test that tests for error with `new Document<IDocument>({ name: 'foo' });` (or the proper doc type for here)

vkarpov15: @hasezoey fixed, can you please re-review?

hasezoey (approved):

Issue #13840, opened 2023-09-08
Strict Projection Object Typing


pshaddel:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

A Projection Object with strict typescript type that:
- Accepts only allowed values of MongoDB Projection object
- Supports Typing of Nested Objects in Projection
- Gives TS Errors on Inclusion and Exclusion

#### 1- Passing MongoDB expected values for projection
Based on the [documentation](https://www.mongodb.com/docs/manual/reference/method/db.collection.find/#std-label-find-projection) only `1`, `0`, `false`, or `true` should be passed(it works also with other numbers).
Example:
```typescript
 const Cat = mongoose.model('Cat', new Schema({ name: String }));
 await Cat.find({}, { name: "true" }) // No TS Error - Returned Value is "true" as string
```
#### 2- No Typing for Nested Objects in Projection
```typescript
 const Cat = mongoose.model('Cat', new Schema({ name: String, childs: [{ name: String }] }));
 await Cat.find({}, { childs: { ... } }) // Inside the `childs` we do not have typing. 
```
#### 3- Inclusion / Exclusion Rules
In projection if it is a inclusion projection, exclusion is not allowed and vice versa. But this could happen and it results in run time error:
```typescript
 const Cat = mongoose.model('Cat', new Schema({ name: String, family: String }));
 await Cat.findOne({}, { name: true, family: false }) // Runtime Error after running the query
```

### Example

I have already implemented a type that I think covers most cases and I would be happy to make the regarding PR if this change is acceptable:
```typescript
type Projector<T, Element> = T extends Array<infer U> ? Projector<U, Element> : T extends object ? {
    [K in keyof T]?: T[K] extends object ? Projector<T[K], Element> | Element : Element;
} : Element;
type InclusionProjection<T> = { _id?: true | 1 | false | 0 } & Omit<Projector<T, true | 1>, '_id'>;
type ExclusionProjection<T> = { _id?: false | 0 } & Omit<Projector<T, false | 0>, '_id'>;

export type ProjectionType<T> = InclusionProjection<T> | ExclusionProjection<T>; // New Projection Type

type RawDocType = {
    _id: string;
    a: string;
    b: {
        c: {
            d: string;
        }
    },
    g: string[];
    f: {
        _id: string;
        starts: number
    }[]
}

type Project = ProjectionType<RawDocType>;

[...]

Issue #3230, opened 2015-08-04
Ability to de-select discriminator key


vkarpov15:
Re: #3133.

Catatomik:
Any update or lead on this ?
I'd like to remove the `__t` field, note that I'm using the `lean` option.

And this is all about memory consumption ; `__t` field for a lot of documents represents a lot of memory.

vkarpov15:
@Catatomik can you provide some code samples that show what you're trying to achieve?

Catatomik:
> @Catatomik can you provide some code samples that show what you're trying to achieve?

Of course, here's a short code sample :
```ts
const dbScheduledRoutes = await TBMScheduledRoutesModel.find<DocumentType<ScheduledRoute>>({}, dbScheduledRoutesProjection)
      .sort({ _id: 1 })
      .populate("trips.schedules", { hor_app: 1, hor_estime: 1, _id: 0 })
      .lean()
      .exec()
```

My types are made with **typegoose**.
`trips.schedules` is `Ref<dbTBM_Schedules_rt>[]`.

The important part is about `schedules`, which is a discriminator.
Short schemas : 
```ts
@modelOptions({ options: { customName: TBMEndpoints.Schedules } })
export class dbTBM_Schedules extends TimeStamps {
  @prop({ required: true })
  public hor_theo!: Date;
}

@modelOptions({ options: { customName: TBMEndpoints.Schedules_rt } })
export class dbTBM_Schedules_rt extends dbTBM_Schedules {
  @prop({ required: true })
  public hor_app!: Date;

  @prop({ required: true })
  public hor_estime!: Date;
}
```

So when populating through `.populate("trips.schedules", { hor_app: 1, hor_estime: 1, _id: 0 })`, I'd like to keep **only** `hor_app` and `hor_estime`. I got to remove `_id` (well documented, this filed is the only exception where you can make an exclusion with an inclusion projection), but nothing about the `__t` field).

A possible solution would be making an exclusion projection only, and put `__t` in. Don't know if it would work, but I strongly prefer to keep an inclusion projection.

vkarpov15:
@Catatomik you can also deselect `__t` in an inclusive projection. Mongoose will automatically strip out `__t` from the projection before sending it to MongoDB, and then strip out `__t` from the result. See the following script for a Mongoose example:

```javascript
'use strict';
  
const mongoose = require('mongoose');

run().catch(err => console.log(err));

async function run() {
[...]

PR #13993, opened 2023-10-19
fix: Strict Projection Object Typing

pshaddel:
**Summary**
Fixes this issue: https://github.com/Automattic/mongoose/issues/13840

**Examples**

### DIFF
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -19,7 +19,7 @@ declare module 'mongoose' {
 
   type MongooseQueryOptions<DocType = unknown> = Pick<QueryOptions<DocType>, 'populate' | 'lean' | 'strict' | 'sanitizeProjection' | 'sanitizeFilter'>;
 
-  type ProjectionFields<DocType> = { [Key in keyof DocType]?: any } & Record<string, any>; // Test
+  type ProjectionFields<DocType> = { [Key in keyof DocType]?: any } & Record<string, any>;
 
   type QueryWithHelpers<ResultType, DocType, THelpers = {}, RawDocType = DocType, QueryOp = 'find'> = Query<ResultType, DocType, THelpers, RawDocType, QueryOp> & THelpers;
 
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -19,7 +19,7 @@ declare module 'mongoose' {
 
   type MongooseQueryOptions<DocType = unknown> = Pick<QueryOptions<DocType>, 'populate' | 'lean' | 'strict' | 'sanitizeProjection' | 'sanitizeFilter'>;
 
-  type ProjectionFields<DocType> = { [Key in keyof DocType]?: any } & Record<string, any>;
+  type ProjectionFields<DocType> = { [Key in keyof DocType]?: any } & Record<string, any>; // Test
 
   type QueryWithHelpers<ResultType, DocType, THelpers = {}, RawDocType = DocType, QueryOp = 'find'> = Query<ResultType, DocType, THelpers, RawDocType, QueryOp> & THelpers;
 
@@ -614,7 +614,7 @@ declare module 'mongoose' {
     >;
 
     /** Get/set the current projection (AKA fields). Pass `null` to remove the current projection. */
-    projection(fields?: ProjectionFields<DocType> | string): ProjectionFields<DocType>;
+    projection(fields?: ProjectionFields<DocType> | string): ProjectionFields<DocType>; //
     projection(fields: null): null;
     projection(): ProjectionFields<DocType> | null;
 
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -614,7 +614,7 @@ declare module 'mongoose' {
     >;
 
     /** Get/set the current projection (AKA fields). Pass `null` to remove the current projection. */
-    projection(fields?: ProjectionFields<DocType> | string): ProjectionFields<DocType>; //
+    projection(fields?: ProjectionFields<DocType> | string): ProjectionFields<DocType>;
     projection(fields: null): null;
     projection(): ProjectionFields<DocType> | null;
 
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -95,7 +95,7 @@ declare module 'mongoose' {
     updatedAt?: boolean;
   }
 
-  interface QueryOptions<DocType = unknown> extends
+  interface QueryOptions<DocType = any> extends
     PopulateOption,
     SessionOption {
     arrayFilters?: { [key: string]: any }[];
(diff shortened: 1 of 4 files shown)


types/query.d.ts
```diff
@@ -19,7 +19,7 @@ declare module 'mongoose' {
 
   type MongooseQueryOptions<DocType = unknown> = Pick<QueryOptions<DocType>, 'populate' | 'lean' | 'strict' | 'sanitizeProjection' | 'sanitizeFilter'>;
 
-  type ProjectionFields<DocType> = { [Key in keyof DocType]?: any } & Record<string, any>;
+  type ProjectionFields<DocType> = { [Key in keyof DocType]?: any } & Record<string, any>; // Test
```
AbdelrahmanHafez, line 22:
```suggestion
  type ProjectionFields<DocType> = { [Key in keyof DocType]?: any } & Record<string, any>;
```

types/query.d.ts
```diff
@@ -122,7 +122,7 @@ declare module 'mongoose' {
     new?: boolean;
 
     overwriteDiscriminatorKey?: boolean;
-    projection?: ProjectionType<DocType>;
+    projection?: ProjectionType<any>;
```
AbdelrahmanHafez, line 125: Would that make the `projection` not take the fields in the document into account?
Is there a way to add that extra validation without using `any`?

pshaddel: Thanks for pointing it out! I actually wanted to ask this. Looks like it is not used(or at least in the tests or ts-tests).

The projection has been used directly in queries:
Screenshot 2023-10-19 at 5 22 40 PM

For passing projection fields to query options we should skip the projection object:
```typescript
Model.findOne({}, {...skipProjection}, { projection: {here is the type that I changed } })
```

vkarpov15 (changes requested):
A couple of use cases I'd like to see tests for that don't look like this PR supports

1. Allow adding fields that aren't in the schema to the projection
2. Projection operators: [$slice](https://www.mongodb.com/docs/manual/reference/operator/projection/slice/), [`$elemMatch`](https://www.mongodb.com/docs/manual/reference/operator/projection/elemMatch/)
3. Dotted paths, like `'child.name': 1`
4. Using `$meta` with text search: `score: { $meta: 'textScore' }`

Will have to review some more to see if there are any other projection cases that we've missed

vkarpov15:
Another thing we'll need to make sure of: `$slice` and `$meta` are not "defining projections" in the sense that they can be used with either inclusive or exclusive projections. `{ name: 1, arr: { $slice: 5 } }` is valid as is `{ name: 0, arr: { $slice: 5 } }`.

pshaddel:
> Another thing we'll need to make sure of: `$slice` and `$meta` are not "defining projections" in the sense that they can be used with either inclusive or exclusive projections. `{ name: 1, arr: { $slice: 5 } }` is valid as is `{ name: 0, arr: { $slice: 5 } }`.
@vkarpov15 
`$slice` and `$elemMatch` are implemented and whenever there is a array we can use these fields.

For using `$meta`, we do not know if there is a text index on the document or not; therefore, do you think we should allow this type(`{ score: { $meta: "textScore" } }`) in any projection?

pshaddel:
> A couple of use cases I'd like to see tests for that don't look like this PR supports
> 
> 1. Allow adding fields that aren't in the schema to the projection
> 2. Projection operators: [$slice](https://www.mongodb.com/docs/manual/reference/operator/projection/slice/), [`$elemMatch`](https://www.mongodb.com/docs/manual/reference/operator/projection/elemMatch/)
> 3. Dotted paths, like `'child.name': 1`
> 4. Using `$meta` with text search: `score: { $meta: 'textScore' }`
> 
> Will have to review some more to see if there are any other projection cases that we've missed


1. resolved, we have this test:
```typsescript
Test.find({}, { someOtherField: 1 }); // This should be allowed since it's not a field in the schema
```
2. resolved
4. Dotted Path
It was quite complicated due to existence of arrays and the way we can project arrays using `.` but finally I could implement it. I had to limit the depth of the dot path to 4 because of circular types(also it could easily slow down Typescript server when the depth of the object is too much.
I also tried to support a combination of dot path and objects. For example:
```
interface X {
  docs: { users: { name: string } }[]
}

type X = Projection<X>
const x: X = { 'docs.users': { name: 1 } } // now supports typing of  this inner object and it suggests `name`
```


The next big challenge that I faced was that we are using some types, like `Date`, `Document`, `Populated`, and `Hydrated`,... that have so many methods. At the time of building the projection object, we are trying to identify the nested members of objects we might suggest fields and methods of these types as part of the projection.

Feel free to suggest your ideas and thoughts on this issue.

fider:
@pshaddel great feature!
Will typed projection parameter also have influence on query return type?
```typescript
type EntityA = { a: number, b: ... }
const a = A.findOne({}, { a: 1 }) // typeof a == { a: number }
```

pshaddel:
> @pshaddel great feature!
> Will typed projection parameter also have influence on query return type?
> ```typescript
> type EntityA = { a: number, b: ... }
> const a = A.findOne({}, { a: 1 }) // typeof a == { a: number }
> ```

Thanks! Still working on this feature. My intention is to implement Query Return Type in another task after tackling this one.

vkarpov15:
@pshaddel looks like I merged a bit too early, literally the next issue I looked at highlighted an issue with this PR :disappointed: . https://github.com/Automattic/mongoose/issues/3230#issuecomment-1825839171 shows that Mongoose lets you deselect the discriminator key in an inclusive projection in addition to `_id`. What do you think about this - should we just get rid of checking for mixed inclusive/exclusive projection? Or find a way to pass in the discriminator key to this logic?

pshaddel:
> @pshaddel looks like I merged a bit too early, literally the next issue I looked at highlighted an issue with this PR 😞 . [#3230 (comment)](https://github.com/Automattic/mongoose/issues/3230#issuecomment-1825839171) shows that Mongoose lets you deselect the discriminator key in an inclusive projection in addition to `_id`. What do you think about this - should we just get rid of checking for mixed inclusive/exclusive projection? Or find a way to pass in the discriminator key to this logic?

This should not be a problem since this discriminator key is not in the interface, so the type of the key would be `any`. it can be anything.

vkarpov15:
That's good, but it is still valid to add `discriminatorKey` to your schema: `new Schema({ __t: String })` is fully valid.

vkarpov15:
@pshaddel I reverted this PR in 8.1 for now with 5337da9a0f416b528dd79eb2be0c8a012865a5e8, because I don't want the test failures and other issues to block releasing 8.1. Please create a new PR with the suggested fixes; we will likely take another look at this PR for a future release as well.

pshaddel:
OK

alaycock:
@vkarpov15 do you know if this was ever reintroduced? I don't see a subsequent PR from @pshaddel

pshaddel:
> @vkarpov15 do you know if this was ever reintroduced? I don't see a subsequent PR from @pshaddel
> 
> If not, maybe it's worth reopening #13840

No further PR from my side, I think a test was failing because of some other changes. I can take a look.

pshaddel:
@alaycock Should I open a new one? I just pushed a commit to the old branch.

PR #13995, opened 2023-10-20
Add fullPath to ValidatorProps

Freezystem:
When using a custom validator function in sub-schemas it's currently impossible to retrieve corresponding field in the schema and thus, access info like `ref` or `refPath`.

### Summary

I just add a new field to `ValidatorOptions` called `fullPath` retrieved from `this.$fullPath` before the call of the validator function.

### Examples

```js
const { Schema, models, model } = require("mongoose");

const User = model("User", {email:  String});

const metaSchema = new Schema({
    user: {
        type: Schema.Types.ObjectId,
        ref: "User",
        validate: {
            async validator(_id, props) {
                // here context is lost if we don't have the fullPath field
                // because retrieving current nested field options is impossible using path
                // as depth related information is no longer available
                const ref = this.schema.path(props.fullPath)?.options?.ref;
                const model = models[ref];
                return (await model.exists({ _id }).exec()) !== null;
            }
        }
    }
});

const Foo = model("Foo", {meta: {type: metaSchema}});

const user = await User.create({email: "example@email.com"});

// Works using `fullPath` but will fail if only `path` is available
await Foo.create({meta: {user: user._id}}); 

// Validation Error as expected
await Foo.create({meta: {user: "e175cac316a79afdd0ad3afb"}}); 
```

### DIFF
--- a/lib/schematype.js
+++ b/lib/schematype.js
@@ -1305,6 +1305,7 @@ SchemaType.prototype.doValidate = function(value, fn, scope, options) {
 
     const validatorProperties = isSimpleValidator(v) ? Object.assign({}, v) : clone(v);
     validatorProperties.path = options && options.path ? options.path : path;
+    validatorProperties.fullPath = this.$fullPath;
     validatorProperties.value = value;
 
     if (validator instanceof RegExp) {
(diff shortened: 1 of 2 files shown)


Freezystem:
As this change is fairly simple can it also be cherrypicked into the 6.x.x branch ?

PR #13998, opened 2023-10-22
fix(document): fix differences between sample codes and documentation

suzuki:
**Summary**

- There is no `lastMod` path in the page: https://mongoosejs.com/docs/plugins.html
- I think `loadedAt` is correct here


vkarpov15 (approved):
Thanks :+1:

PR #14007, opened 2023-10-25
fix: casting an array of objects with $elemMatch

IslandRhythms:


IslandRhythms:
So basically we're playing hide and seek with the schema.

vkarpov15:
Since this PR was against master, I pushed https://github.com/Automattic/mongoose/commit/266804b995ae715cd4746d0e5b687e7fb96441fc to make this change against 7.x instead. So LGTM, just incorrect branch for the moment. Thanks :+1:

PR #14011, opened 2023-10-27
fix(schema): handle recursive schemas in discriminator definitions

### LINKED ISSUE #13978: Recursive discriminator definition causes infinite recursion
### Prerequisites

### Mongoose version

7.6.2

### Node.js version

18.16

### MongoDB server version

6.0.5

_No response_

### Description

Defining a discriminated schema that contains a field defined as the base schema fails due to infinite recursion.

### Steps to Reproduce

```javascript 'use strict';

const { Schema } = mongoose;

const base = new Schema({
    type: { type: Number, required: true },

const recursive = new Schema({
    self: { type: base, required: true }
});

base.discriminator(1, recursive);
baseModel = mongoose.model('base', base); // RangeError: Maximum call stack size exceeded
```

### Expected Behavior

In previous versions of mongoose (before 7.6.2) this kind recursion worked as expected - You can instantiate the recursive objects and they were correctly typed according to the discriminator - Though they also suffered from the bug that needed the discriminator to be explicitly defined on each path.

vkarpov15:
Fix #13978

**Summary**

Deep cloning schema `paths` property causes issues when discriminator schema contains the main schema as a subdocument. This PR makes it so that we don't deep clone unnecessarily, just clone each `path` and avoid re-applying discriminators unnecessarily (discriminators are already copied in `SubdocumentPath.prototype.clone`)

**Examples**

### DIFF
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -3172,4 +3172,20 @@ describe('schema', function() {
     const res = await Test.findOne({ _id: { $eq: doc._id, $type: 'objectId' } });
     assert.equal(res.name, 'Test Testerson');
   });
+
+  it('handles recursive definitions in discriminators (gh-13978)', function() {
+    const base = new Schema({
+      type: { type: Number, required: true }
+    }, { discriminatorKey: 'type' });
+
+    const recursive = new Schema({
+      self: { type: base, required: true }
+    });
+
+    base.discriminator(1, recursive);
+    const TestModel = db.model('Test', base);
+
+    const doc = new TestModel({ type: 1, self: { type: 1 } });
+    assert.strictEqual(doc.self.type, 1);
+  });
 });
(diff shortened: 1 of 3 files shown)


hasezoey (commented):
LGTM

test/schema.test.js
```diff
@@ -3172,4 +3172,20 @@ describe('schema', function() {
...
+    const base = new Schema({
+      type: { type: Number, required: true }
+    }, { discriminatorKey: 'type' });
+
+    const recursive = new Schema({
+      self: { type: base, required: true }
+    });
+
+    base.discriminator(1, recursive);
+    const TestModel = db.model('Test', base);
```
hasezoey, line 3186:
```suggestion
    const TestModel = db.model('gh13978', base);
```

change model name to fit style of surrounding tests and fix tests failing

PR #14017, opened 2023-10-29
fix(connection): retain modified status for documents created outside a transaction during transaction retries

### LINKED ISSUE #13973: Modified flag does not work correctly during transaction retries
### Prerequisites

### Mongoose version

7.6.1

### Node.js version

16.19.1

### MongoDB server version

5

4.8.3

### Description

Hello there! I have some code that essentially boils down to this (pseudo-code):

```typescript
const document = await getDocument({ status: "in_progress" }); // returns a document where status === "in_progress"

document.status = "canceled";

withTransaction(async () => {
    await document.save();
});
```

Mongoose v7.4.4 fixed some issues with the modified flags and now they are properly reverted to `true` in case of a transaction failure. So if the transaction is retried, the "status" field will still be "modified" and a correct update will be sent to the database when `.save()` is called again.

However, this code would not work anymore:

```typescript
const document = await getDocument({ status: "in_progress" });

withTransaction(async () => {
    document.status = "canceled";
    await document.save();
});
```

Now if the transaction is retried, the "status" field is still modified. However, after the line `document.status = "canceled";` is executed it is no longer "modified".

My first guess was that the setter simply checks if the value changed, and since in this case it didn't, the field's modified flag is set to `false`. I then tried setting the same value to the field twice in a row, and the "modified" flag worked as expected. Kind of expected, since such an obvious thing would probably have been caught a long time ago : ]

### Steps to Reproduce

I am sorry for not providing a working reproduction script this time, my deadline is literally tomorrow and I just took a few minutes to write this down. If you'd like me to provide one, please say so and I will post it next week.

### Expected Behavior

I often use a pattern like this:

```typescript
const document = await getDocument({ status: "in_progress" });

withTransaction(async () => {
    await document.cancel();
});
```

where `.cancel()` sets the status and calls `.save()`. Due to this issue, a lot of code that we've written will occasionally misbehave.

My expectation is that the "modified" flag works consistently with how it does when there are no transaction retries.

vkarpov15:
Fix #13973

**Summary**

#13973 points out that, due to the fact that we store the document's initial `savedState` during the first transaction operation, this can lead to us calling `unmarkModified()` unintentionally. Right now `$__set()` calls `unmarkModified()` if the document's `savedState` says that the value hasn't changed from the initial value. However, we also need to check if this is a transaction retry and the transaction says the document was already modified before the transaction started.

**Examples**

### DIFF
--- a/test/docs/transactions.test.js
+++ b/test/docs/transactions.test.js
@@ -396,4 +396,29 @@ describe('transactions', function() {
     assert.equal(docs.length, 1);
     assert.equal(docs[0].name, 'test');
   });
+
+  it('transaction() retains modified status for documents created outside the transaction (gh-13973)', async function() {
+    db.deleteModel(/Test/);
+    const Test = db.model('Test', Schema({ status: String }));
+
+    await Test.createCollection();
+    await Test.deleteMany({});
+
+    const { _id } = await Test.create({ status: 'test' });
+    const doc = await Test.findById(_id);
+
+    let i = 0;
+    await db.transaction(async(session) => {
+      doc.status = 'test2';
+      assert.ok(doc.$isModified('status'));
+      await doc.save({ session });
+      if (++i < 3) {
+        throw new mongoose.mongo.MongoServerError({
+          errorLabels: ['TransientTransactionError']
+        });
+      }
+    });
+
+    assert.equal(i, 3);
+  });
 });
--- a/test/docs/transactions.test.js
+++ b/test/docs/transactions.test.js
@@ -397,7 +397,7 @@ describe('transactions', function() {
     assert.equal(docs[0].name, 'test');
   });
 
-  it('transaction() retains modified status for documents created outside the transaction (gh-13973)', async function() {
+  it('transaction() retains modified status for documents created outside of the transaction then modified inside the transaction (gh-13973)', async function() {
     db.deleteModel(/Test/);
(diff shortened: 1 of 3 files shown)


test/docs/transactions.test.js
```diff
@@ -396,4 +396,29 @@ describe('transactions', function() {
...
+    await Test.deleteMany({});
+
+    const { _id } = await Test.create({ status: 'test' });
+    const doc = await Test.findById(_id);
+
+    let i = 0;
+    await db.transaction(async(session) => {
+      doc.status = 'test2';
+      assert.ok(doc.$isModified('status'));
+      await doc.save({ session });
```
hasezoey, line 414: from what i can tell, this test tests modified status *within* a transcation, but the title makes out sound like retaining modified status across transaction if modified *outside* the transaction, like (pseudo):

```js
const doc = await Model.find();
doc.test = "something"; // not saved outside of transaction
await db.transaction(async(session) => { // run at least twice, once with error
  assert.ok(doc.$isModified('test'));
  await doc.save({ session });
  throw new Error();
});
```

hasezoey (commented):
Looks good to me, though i have a test question

test/docs/transactions.test.js
hasezoey: test title was changed, so it makes it more clear what it is testing, thanks

Issue #13103, opened 2023-03-01
UUID type in Mongoose 7


wfjake:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

With the release of MongoDB 5+, MongoDB's driver now exposes the UUID type via `require('mongodb').BSON.UUID` (see [API docs](https://mongodb.github.io/node-mongodb-native/5.0/modules/BSON.html).

This means that Mongoose can expose a UUID type for aggregation like it currently does for ObjectId. This would allow developers to aggregate match on UUIDs.

Current workarounds:
```
import {UUID} from "bson"; // Mongoose 6 and below (works with Mongoose 7 but ill-advised, see https://github.com/mongodb/node-mongodb-native/releases/tag/v5.0.0 "BSON v5" section)
import {BSON} from "mongodb"; // Mongoose 7 and above, use BSON.UUID() below

let my = await My.aggregate().match({_id: new UUID("1e984535-98ad-446b-84cc-bd8f59290eb8")}).exec();
```

Proposed feature:
```
import mongoose from "mongoose";

let my = await My.aggregate().match({_id: new mongoose.Types.UUID("1e984535-98ad-446b-84cc-bd8f59290eb8")}).exec();
```

PR #14023, opened 2023-11-01
feat: add Types.UUID typings

k725:
**Summary**

UUID are now available in https://github.com/Automattic/mongoose/issues/13103 pull requests, but TypeScript type definitions were not present, so i added this.

**Examples**


vkarpov15 (commented):
Thanks :+1:

Issue #14027, opened 2023-11-01
"this" context has "any" type in Schema methods option


ruxxzebre:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.0.0 (same case as 7.2.2)

### Node.js version

20.5.0 (same case as 18.0.0)

### MongoDB server version

7.0

### Typescript version (if applicable)

5.2.2 (recently updated, same case with 4.8.4)

### Description

As stated [here](https://github.com/Automattic/mongoose/issues/13800#issuecomment-1789367289), issue with _this_ context was fixed, if we declare methods like this:
```ts
schema.method("fullName", function fullName() {
    return this.firstName + " " + this.lastName;
});
``` 

**But**, if we're trying to specify methods in schema's _method_ option AND provide IUserMethods into a Schema generic, _this_ becomes _any_ once again.

```ts
new Schema<IUser, UserModel, IUserMethods>
```

<img width="451" alt="Screenshot 2023-11-01 at 20 05 27" src="https://github.com/Automattic/mongoose/assets/54982966/4b73c3be-cfce-4c8b-97ab-447d308fed93">

However, If we **do not** provide IUserMethods into a Schema generic, _this_ becomes a Document type, as it should.

```ts
new Schema<IUser, UserModel>
or
new Schema<IUser, UserModel, {}>
```

<img width="442" alt="Screenshot 2023-11-01 at 20 06 26" src="https://github.com/Automattic/mongoose/assets/54982966/4b12aaf6-f13d-473c-8594-c6fb41c58aff">

@pshaddel thanks!

### Steps to Reproduce

Declare types/interfaces for document, methods and model.
```ts
interface IUser {
    firstName: string;
    lastName: string;
}

// Put all user instance methods in this interface:
interface IUserMethods {
    fullName(): string;
}

// Create a new Model type that knows about IUserMethods...
type UserModel = Model<IUser, {}, IUserMethods>;
```

Initialise schema with methods, **but** without third generic type, and observe the result.
```ts
const schema = new Schema<IUser, UserModel>(
    {
        firstName: { type: String, required: true },
        lastName: { type: String, required: true },
    },
    {
        methods: {
            fullName() {
                return this.firstName + " " + this.lastName;
            },
        },
    }
);
```

Initialise schema with methods **and** third generic type (IUserMethods), and observe the result.
```ts
[...]

PR #14028, opened 2023-11-01
Fix "this" of Schema's instance methods

ruxxzebre:
**Summary**

Fixes bug described in [this issue](https://github.com/Automattic/mongoose/issues/14027).

**Examples**
Before

Screenshot 2023-11-01 at 22 57 52

After

Screenshot 2023-11-01 at 22 57 09

### DIFF
--- a/types/schemaoptions.d.ts
+++ b/types/schemaoptions.d.ts
@@ -220,7 +220,7 @@ declare module 'mongoose' {
     TInstanceMethods,
     {},
     Record<any, (this: THydratedDocumentType, ...args: any) => unknown>,
-    TInstanceMethods
+    AddThisParameter<TInstanceMethods, THydratedDocumentType>
     >
 
     /**
--- a/types/utility.d.ts
+++ b/types/utility.d.ts
@@ -40,4 +40,26 @@ type IfEquals<T, U, Y = true, N = false> =
     (<G>() => G extends T ? 1 : 0) extends
     (<G>() => G extends U ? 1 : 0) ? Y : N;
 
+/**
+ * @summary Extracts 'this' parameter from a function, if it exists. Otherwise, returns fallback.
+ * @param {T} T Function type to extract 'this' parameter from.
+ * @param {F} F Fallback type to return if 'this' parameter does not exist.
+ */
+type ThisParameter<T, F> = T extends { (this: infer This): void }
+  ? This
+  : F;
+
+/**
+ * @summary Decorates all functions in an object with 'this' parameter.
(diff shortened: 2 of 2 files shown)


pshaddel:
@ruxxzebre Nice. I was also looking at this issue. One problem is that `this` type is `IUserMethods` but the type should be `HydratedDocument`(It contains all the fields and methods of model like `save`)

pshaddel:
@ruxxzebre Looks nice! Could you implement some tests that check the type of `this` in methods, both when you are not passing `InstanceMethods` and when you are passing your desired instance?

pshaddel:
@ruxxzebre Also please look at this line:
https://github.com/Automattic/mongoose/blob/master/types/index.d.ts#L300

I guess we do not have typing in this situation because of this line:
```typescript
schema.methods.fullName = function fullName() {
    expectType<string>(this.firstName);
    return ''
  }
```
And maybe re-use your implementation so the type of `this` methods stays consistent.

pshaddel (commented):
@ruxxzebre Looks nice! Could you implement some tests that check the type of `this` in methods, both when you are not passing `InstanceMethods` and when you are passing your desired instance?

pshaddel (commented):
@ruxxzebre Also please look at this line:
https://github.com/Automattic/mongoose/blob/master/types/index.d.ts#L300

I guess we do not have typing in this situation because of this line:
```typescript
schema.methods.fullName = function fullName() {
    expectType<string>(this.firstName);
    return ''
  }
```
And maybe re-use your implementation so the type of `this` methods stays consistent.

ruxxzebre:
I seem to have found similar behaviour with _statics_, as well as _query helpers_.
For query helpers there's [example in docs](https://mongoosejs.com/docs/typescript/query-helpers.html) on how it can be mitigated.
`ProjectSchema.query.byName = function byName(
  this: QueryWithHelpers<any, HydratedDocument<Project>, ProjectQueryHelpers>,
  name: string
) {
  return this.find({ name: name });
};`
With AddThisParameter we can help with typing of those also. What you think?

pshaddel:
@ruxxzebre PR Looks nice. Yes, if you can fix those two, why not trying it! 😊 
BTW I am not a maintainer so for the merge and final review you should wait for a maintainer to look at this. But good job fixing the issue!

vkarpov15:
This is excellent, thanks  for this PR. We've been struggling to figure out how to implement something like the `AddThisParameter<>` helper :+1:

vkarpov15:
Feel free to put in a separate PR for the query helper changes you mentioned in https://github.com/Automattic/mongoose/pull/14028#issuecomment-1789895290

PR #14042, opened 2023-11-02
Index: Fix index alias object key position

meabed:
**Summary**

This pr fixes an issue index object order when it has aliases, it rename the key from the alias to the original field and maintain it's position in the object as entered by user.

**Examples**

```

const UserSchema = new Schema(
  {
    n: {
      type: String,
      alias: 'name',
    },
    age: {
      type: Number,
    }
  }
)

UserSchema.index({ name: 1, age: -1 });

// The current behavior will generate index object like this 
// original fields { name: 1, age: -1 }
// after replacing the name alias with original field  { age: -1, n: 1 } ❌ 

The fix in this PR will make it work as intended and preserve the object key order
{ n: 1, age: -1 } ✅ 

```

This issue exist in version 7 and version 8 and would great to merge this PR in both versions.

Thank you!

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -2037,11 +2037,9 @@ Schema.prototype.index = function(fields, options) {
   }
   for (const key in fields) {
     if (this.aliases[key]) {
-      fields[this.aliases[key]] = fields[key];
-      delete fields[key];
+      fields = utils.renameObjKey(fields, key, this.aliases[key]);
     }
   }
-
   for (const field of Object.keys(fields)) {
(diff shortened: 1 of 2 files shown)


meabed:
I have fixed the linting comments 👍

meabed:
Hey @vkarpov15 - Hope you well.

This is a small PR - if you could have a look and see if we could merge it in :) 

Thank you 🙏

vkarpov15 (approved):
Thanks :+1:

PR #14046, opened 2023-11-04
Issue 13999 fix

donoftime2018:
I typecasted elems as T[] before returning elems in insertMany


donoftime2018:
I typecasted elems as T[] before returning elems

PR #14047, opened 2023-11-04
types(model+query): unpack arrays in distinct return type

### LINKED ISSUE #14026: Incorrect inferred return type for `distinct()` on array fields
### Prerequisites

### Mongoose version

8.0.0

### Node.js version

20.0.0

### MongoDB server version

6.0

latest

### Description

The inferred return type for `distinct()` queries with an array fields is incorrect. It's inferred to be `string[][]` for a `string[]` field when in fact it should be `string[]`

### Steps to Reproduce

```ts
interface Foo {
  bar: string[];
}

const FooModel = model<Foo>('Foo', new Schema<Foo>({ bar: [String] }));

await FooModel.create({ bar: ['a', 'b'] })

const distinctBar = await FooModel.distinct('bar');
console.log(distinctBar); // [ 'a', 'b' ]
```

`distinctBar` is actually a `string[]` but the inferred return type is `string[][]`.

### Expected Behavior

Inferred return type does not add `Array` to a type which is already an `Array`

vkarpov15:
Fix #14026

**Summary**

`Unpack<T>` returns the array element type if `T` is an array, or `T` otherwise. We already use `Unpacked<>` in some similar cases, like async iterators for `aggregate()`. 

**Examples**

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -781,4 +781,11 @@ async function gh14026() {
 
   const distinctBar = await FooModel.distinct('bar');
   expectType<string[]>(distinctBar);
+
+  const TestModel = mongoose.model(
+    'Test',
+    new mongoose.Schema({ bar: [String] })
+  );
+
+  expectType<string[]>(await TestModel.distinct('bar'));
 }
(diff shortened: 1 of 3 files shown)


hasezoey (approved):
Looks good, to my understanding this should also change if the inferred type is used, maybe add a test for that too? (no explicit interface)

hasezoey (approved):

PR #14050, opened 2023-11-05
Issue14000

donoftime2018:
**Summary**

Issue 14000 explained an issue where using updateOne to $set an element as the first item in array, items, turned said element into an array type instead of an object type. When outputting the results of the items array, the first element had the index number in front of the object, and subsequent elements did not.

[
  '0': {
    first: 'T',
    last: 'Test',
    _id: new ObjectId("654727e3b2498b1007110fe8")
  },
  {
    first: '222',
    last: '222',
    _id: new ObjectId("654727e3b2498b1007110fec")
  }
]

All I did was replace $set with $setOnInsert and the result now looks like...

[
  {
    first: 'T',
    last: 'Test',
    _id: new ObjectId("654727e3b2498b1007110fe8")
  },
  {
    first: '222',
    last: '222',
    _id: new ObjectId("654727e3b2498b1007110fec")
  }
]


donoftime2018:
> This isn't even remotely close to resembling a bug fix. At best, you tweaked a repro script to avoid a potential bug. I'm going to close this.

What would a bug fix look like?

PR #14052, opened 2023-11-05
fix(ChangeStream): correctly handle `hydrate` option when using change stream as stream instead of iterator

### LINKED ISSUE #14049: hydrate option not working on watch
### Prerequisites

### Mongoose version

6.12.2 and 8.0.0

### Node.js version

18.17.1

### MongoDB server version

6.0.11

_No response_

### Description

Hi guys...

The documentation about ```Model.watch``` says that in the options field you can use "hydrate" with "fullDocument" to get the document hydrated, but it seems not to be working. I guess "hydrate" includes virtuals.

Besides of that issue, the example uses ```doc.remove()``` which is not longer available since mongoose 6.x (AFAIR), so you should update the example.

### Steps to Reproduce

```js
import db from "mongoose";                                                             
await db.connect("mongodb://127.0.0.1/test");                                          
const Schema = new db.Schema({                                                         
  a: String                                                                            
});                                                                                    
Schema.virtual("test", {                                                               
  option1: 1,                                                                          
  option2: 2                                                                           
}).get(function test1() {                                                              
  return this.a + " virtual";                                                          
});                                                                                    
const model = db.model("model", Schema);                                               
model                                                                                  
  .watch([], {                                                                         
    hydrate: true,                                                                     
    fullDocument: "updateLookup"                                                       
  })                                                                                   
  .on("change", change => {                                                            
    console.log("from watch", change.fullDocument?.toJSON?.({virtuals: true}));        
  });                                                                                  
                                                                                       
const doc = await model.create({a: "test"});                                                
console.log("from create", doc?.toJSON?.({virtuals: true})); 
await doc.deleteOne();                          
db.disconnect();                                                                       
```
```sh
❯ node watch_hydrate.js                            
from create {                                      
  a: 'test',                                       
  _id: new ObjectId('654671dd46f8b1dc2d71d177'),   
  __v: 0,                                          
  test: 'test virtual',                            
  id: '654671dd46f8b1dc2d71d177'                   
}                                                  
from watch undefined                               
```

### Expected Behavior

My idea is to able to "use" the entire schema definition inside a watch, but instead I've to use a ```model.findById()``` or ```model.hydrate()```, besides the need to use two operations to the DB (for model.findById()), I can't use a generic watch function.

Oth

Best regards.

PS: I still have issues with transactions, I'm doing a more exhaustives test to check if it's a problem with my code or mongoose, but that code hasn't been touched in years...

vkarpov15:
Fix #14049

**Summary**

Missed a spot when handling the `hydrate` option for change streams: works fine when you use change streams as an iterator with `next()`, but doesn't work with `on()`.

**Examples**

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -5364,6 +5364,32 @@ describe('Model', function() {
           assert.equal(changeData.fullDocument.get('name'), 'Tony Stark');
         });
 
+        it('fullDocument with immediate watcher and hydrate (gh-14049)', async function() {
+          const MyModel = db.model('Test', new Schema({ name: String }));
+
+          const doc = await MyModel.create({ name: 'Ned Stark' });
+
+          const changes = [];
+          const p = new Promise((resolve) => {
+            MyModel.watch([], {
+              fullDocument: 'updateLookup',
+              hydrate: true
+            }).on('change', change => {
+              changes.push(change);
+              resolve(change);
+            });
+          });
+
+          await MyModel.updateOne({ _id: doc._id }, { name: 'Tony Stark' });
+
+          const changeData = await p;
+          assert.equal(changeData.operationType, 'update');
+          assert.equal(changeData.fullDocument._id.toHexString(),
+            doc._id.toHexString());
+          assert.ok(changeData.fullDocument.$__);
+          assert.equal(changeData.fullDocument.get('name'), 'Tony Stark');
+        });
+
         it('respects discriminators (gh-11007)', async function() {
           const BaseModel = db.model('Test', new Schema({ name: String }));
           const ChildModel = BaseModel.discriminator('Test1', new Schema({ email: String }));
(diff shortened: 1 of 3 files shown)


test/model.test.js
```diff
@@ -5364,6 +5364,32 @@ describe('Model', function() {
           assert.equal(changeData.fullDocument.get('name'), 'Tony Stark');
         });
 
+        it('fullDocument with immediate watcher and hydrate (gh-14049)', async function() {
+          const MyModel = db.model('Test', new Schema({ name: String }));
+
+          const doc = await MyModel.create({ name: 'Ned Stark' });
+
+          const changes = [];
```
hasezoey, line 5372: this array is pushed to, but never read from

hasezoey (approved):
Code looks good to me, though one thing in the tests

PR #14068, opened 2023-11-09
Island rhythms/gh 14055

### LINKED ISSUE #14055: Recursive schema array in discriminator definition causes infinite recursion
### Prerequisites

### Mongoose version

7.6.4

### Node.js version

18.16

### MongoDB server version

6.0.5

_No response_

### Description

Follow up on #13978 - The issue still reproduces when the schema field is an array of the base type

### Steps to Reproduce

```javascript
use strict';

const { Schema } = mongoose;

const base = new Schema({
    type: { type: Number, required: true },

const recursive = new Schema({
    self: { type: [base], required: true }
});

base.discriminator(1, recursive);
const baseModel = mongoose.model('base', base);
```

### Expected Behavior

_No response_

IslandRhythms:
closes #14055 
It works but the simplicity of the solution makes me feel like I might have broken a bunch of other stuff.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -412,8 +412,9 @@ Schema.prototype._clone = function _clone(Constructor) {
   s.s.hooks = this.s.hooks.clone();
 
   s.tree = clone(this.tree);
+  // recursion is triggered here
   s.paths = Object.fromEntries(
-    Object.entries(this.paths).map(([key, value]) => ([key, value.clone()]))
+    Object.entries(this.paths).map(([key, value]) => ([key, value]))
   );
   s.nested = clone(this.nested);
   s.subpaths = clone(this.subpaths);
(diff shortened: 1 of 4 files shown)


lib/schema.js
```diff
@@ -413,7 +413,7 @@ Schema.prototype._clone = function _clone(Constructor) {
 
   s.tree = clone(this.tree);
   s.paths = Object.fromEntries(
-    Object.entries(this.paths).map(([key, value]) => ([key, value.clone()]))
+    Object.entries(this.paths).map(([key, value]) => ([key, !value.$isMongooseDocumentArray ? value.clone() : value]))
```
vkarpov15, line 416: While this does work with all tests passing, I think this undercuts the core idea of `clone()`: skipping cloning doc arrays seems incorrect.

Maybe try adding `_skipApplyDiscriminators` in the DocumentArray schema file like here? https://github.com/Automattic/mongoose/pull/14011/files#diff-6f3b287d6939d72b68c31e5ccd577a699f106a13e3068bc977ff92dc13b9d934

vkarpov15 (approved):
Nice :+1:

Issue #13987, opened 2023-10-17
Mongoose types incorrect for when rawResult: true is set within findOneAndUpdate middleware


FullStackSteve:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

~7.4.4

### Node.js version

v16.19.1

### MongoDB server version

4.0

### Typescript version (if applicable)

~11.4.1

### Description

Initially reported here https://github.com/typegoose/typegoose/issues/886 and recommended to post here re: [comment](https://github.com/typegoose/typegoose/issues/886#issuecomment-1766182237):

Incorrect types for the doc in the post findOneAndUpdate middleware when rawResult: true is set as option to the model.findOneAndUpdate(...) method

### Steps to Reproduce

In typegoose:

```ts
// MongoDB: 5.0 (Docker)
// Typescript 4.9.5
import { getModelForClass, prop, post } from '@typegoose/typegoose'; // @typegoose/typegoose@11.5.1
import * as mongoose from 'mongoose'; // mongoose@7.5.4

@post<User>('findOneAndUpdate', function (doc, next) {
  // @ts-ignore
  doc.value.id = doc.value._id;  
  // @ts-ignore
  doc.value.save();
  next();
})
class User {
  @prop()
  public id: string;

  @prop()
  public username?: string;
}

const UserModel = getModelForClass(User);

async function main() {
  await mongoose.connect(`mongodb://localhost:27017/`, {
    dbName: 'verifyMASTER',
  });

  const doc = new UserModel({ username: 'user1' });

  const upsertResult = await UserModel
    .findOneAndUpdate({}, { username: 'user1' }, {
      upsert: true,
      new: true,
      rawResult: true
    })
    .exec();

  console.log(doc);

  await mongoose.disconnect();
}

main();
```

Roughly the mongoose equivalent:
```js
const UserModel = mongoose.model('User', new mongoose.Schema({
  id: String,
  username: Number
}));

UserModel.post('findOneAndUpdate', function(doc, next) {
   // TYPES indicate doc.id but actual passed is doc.value.id
   doc.value.id = doc.value._id;  
   // TYPES indicate doc.save() but actual passed is doc.value.save()
   doc.value.save();
   next();
});


async function main() {
  await mongoose.connect(`mongodb://localhost:27017/`, {
    dbName: 'verifyMASTER',
  });

  const doc = new UserModel({ username: 'user1' });

[...]

PR #14078, opened 2023-11-13
Fix: Mongoose types incorrect for when includeResultMetadata: true is set

prathamVaidya:
**Summary**

Fixed Issue #13987 
Mongoose Version: 8.0

When `includeResultMetadata` is set to `true`. The document parameter in post middleware can also be `ModifyResult<>`.

Functions that support `includeResultMetadata` option are : `findOneAndUpdate`, `findByIdAndUpdate`, `findOneAndReplace`, `findByIdAndDelete`, and `findOneAndDelete`.

Changes Done:

- [x] Introduced a separate document type for the post middleware, specifically for these functions, ensuring they also return `ModifyResult<>`.
- [x] Some of these functions were not returning `ModifyResult<>` when `includeResultMetadata` was set to True. I've addressed this by adding appropriate return types.
- [x] In cases where `includeResultMetadata` is false and no matching document is found in any of these functions, the document parameter in the post middleware is `NULL`. So I have included `NULL` as a possible doc type in the middleware parameter.

**Examples**

```ts
import mongoose, { ModifyResult, Query } from "mongoose";

const UserSchema = new mongoose.Schema({
  id: String,
  username: String,
});

UserSchema.post<
  Query<
    mongoose.HydratedDocument<mongoose.InferSchemaType<typeof UserSchema>>,
    mongoose.InferSchemaType<typeof UserSchema>
  >
>("findOneAndDelete", function (doc) {
// doc can be NULL | QueryResultType<T> | ModifyResult<QueryResultType<T>>
  if (doc && "value" in doc) {
    console.debug("DELETE VALUE = ", doc.value);
  }
});

const UserModel = mongoose.model("User", UserSchema);

async function main() {
  await mongoose.connect(`mongodb://localhost:27017/mongoose-test`);

  const doc = new UserModel({ username: "user1" });

  await doc.save();

  const upsertResult = await UserModel.findOneAndDelete(
    {username: "user1"},
    {
      includeResultMetadata: true,
    }
  ).exec();

  if (upsertResult) {
    console.log("Result=", upsertResult.value); // it has type set to ModifyResult now
  } else {
    console.log("NULL", upsertResult);
  }
}

main();

```

I have fixed the issue for v8.0. If the solution is acceptable then I will also do it for v7.4 that was originally specified in the Issue. @hasezoey


types/models.d.ts
```diff
@@ -565,6 +569,10 @@ declare module 'mongoose' {
       TRawDocType,
       'findOneAndDelete'
     >;
+    findByIdAndRemove<ResultDoc = THydratedDocumentType>(
```
vkarpov15, line 572: `findByIdAndRemove()` is no longer supported in Mongoose 8, so please remove this

types/middlewares.d.ts
```diff
@@ -5,6 +5,7 @@ declare module 'mongoose' {
   type MongooseDistinctDocumentMiddleware = 'save' | 'init' | 'validate';
   type MongooseDocumentMiddleware = MongooseDistinctDocumentMiddleware | MongooseQueryAndDocumentMiddleware;
 
+  type MongooseRawResultQueryMiddleware = 'findOneAndUpdate' | 'findOneAndReplace' | 'findOneAndDelete' | 'findByIdAndUpdate' | 'findOneAndDelete';
```
vkarpov15, line 8: `findByIdAndUpdate` doesn't have distinct middleware, that function fires `findOneAndUpdate` middleware

vkarpov15: Also it looks like you have `findOneAndDelete` twice here

types/models.d.ts
```diff
@@ -640,11 +647,6 @@ declare module 'mongoose' {
       replacement: TRawDocType | AnyObject,
       options: QueryOptions<TRawDocType> & { includeResultMetadata: true }
     ): QueryWithHelpers<ModifyResult<ResultDoc>, ResultDoc, TQueryHelpers, TRawDocType, 'findOneAndReplace'>;
-    findOneAndReplace<ResultDoc = THydratedDocumentType>(
```
vkarpov15, line 643: Why did you remove `findOneAndReplace` here?

vkarpov15 (changes requested):
Thanks for your PR. Please fix lint, add a test case, and address the comments in this review :+1:

types/models.d.ts
prathamVaidya: There was a duplicate of findOneAndReplace with option ` { includeResultMetadata: true }` .
It was from old commit where someone replaced `rawResult` option to  `includeResultMetadata` which created duplicates.

I found 2 occurrences of this code. So removed one. Also removed duplicate for `findByIdAndUpdate`. 

```
findOneAndReplace<ResultDoc = THydratedDocumentType>(
      filter: FilterQuery<TRawDocType>,
      replacement: TRawDocType | AnyObject,
      options: QueryOptions<TRawDocType> & { includeResultMetadata: true }
```

types/middlewares.d.ts
prathamVaidya: Thanks @vkarpov15,  I'll update.

types/models.d.ts
prathamVaidya: Ok, should I delete all the `findByIdAndRemove` type definitions?

prathamVaidya: I have delete all type definations for `findByIdAndRemove`

types/middlewares.d.ts
```diff
@@ -5,6 +5,7 @@ declare module 'mongoose' {
   type MongooseDistinctDocumentMiddleware = 'save' | 'init' | 'validate';
   type MongooseDocumentMiddleware = MongooseDistinctDocumentMiddleware | MongooseQueryAndDocumentMiddleware;
 
+  type MongooseRawResultQueryMiddleware = 'findOneAndUpdate' | 'findOneAndReplace' | 'findOneAndDelete';
   type MongooseDistinctQueryMiddleware = 'count' | 'estimatedDocumentCount' | 'countDocuments' | 'deleteMany' | 'distinct' | 'find' | 'findOne' | 'findOneAndDelete' | 'findOneAndReplace' | 'findOneAndUpdate' | 'replaceOne' | 'updateMany';
```
hasezoey, line 9: shouldnt `MongooseDistinctQueryMiddleware` inherit `MongooseRawResultQueryMiddleware` instead of defining it duplicately?

types/models.d.ts
```diff
@@ -587,11 +575,6 @@ declare module 'mongoose' {
       update: UpdateQuery<TRawDocType>,
       options: QueryOptions<TRawDocType> & { includeResultMetadata: true }
     ): QueryWithHelpers<ModifyResult<ResultDoc>, ResultDoc, TQueryHelpers, TRawDocType, 'findOneAndUpdate'>;
-    findByIdAndUpdate<ResultDoc = THydratedDocumentType>(
-      id: mongodb.ObjectId | any,
-      update: UpdateQuery<TRawDocType>,
-      options: QueryOptions<TRawDocType> & { includeResultMetadata: true }
-    ): QueryWithHelpers<ModifyResult<ResultDoc>, ResultDoc, TQueryHelpers, TRawDocType, 'findOneAndUpdate'>;
```
hasezoey, line 594: was this also removed because of a duplicate?

hasezoey (approved):
Looks good to me, though some minor questions / suggestions.

also does someone know why the tests are not running for the latest commits?

types/middlewares.d.ts
prathamVaidya: While `MongooseDistinctQueryMiddleware` is a superset of `MongooseRawResultQueryMiddleware`. I don't think its a good idea to inherit it because in case where we create a new type for middlewares that support some other option than that type can also contain middlewares from `MongooseRawResultQueryMiddleware` and vice versa. According to me, It's better if we maintain a separate type for this. Let me know your final decision and I'll update accordingly.

types/models.d.ts
prathamVaidya: Yes, this was also duplicate. I removed duplicates for these two: `findByIdAndUpdate` and `findOneAndReplace`

types/middlewares.d.ts
hasezoey: i will be fine with it either way, @vkarpov15 what do you think?

prathamVaidya: @hasezoey, once all the changes are completed and it's ready to be approved, there are conflicts of the branch merging process due to the latest pushes. I just wanted to know, do I have to resolve these conflicts, or will the reviewer handle it?

vkarpov15: I think avoiding inheriting is a reasonable idea.

vkarpov15:
Can you please fix merge conflicts?

prathamVaidya:
Fixed Merge Conflicts ✅

vkarpov15 (approved):
LGTM, thanks :+1:

PR #14095, opened 2023-11-18
types: allow defining document array using `[{ prop: String }]` syntax

### LINKED ISSUE #13424: When migrating from mongoose 5 to 7, how to get LeanDocument funcitonality?
### Prerequisites

### Mongoose version

7.2.0

### Node.js version

18.16.0

### MongoDB version

6.0

### Operating system

macOS

_No response_

### Issue

I'm trying to migrate our nestjs/mongoose codebase from version 5 to 7.  

We've been using using `T.lean()` or `T.toObject()` and our return types would be `LeanDocument<T>`. Mongoose 7 no longer exports LeanDocument, and the existing migration guide suggests using the [following setup][1]:

```typescript
// Do this instead, no `extends Document`
interface ITest {
  name?: string;
}
const Test = model<ITest>('Test', schema);

// If you need to access the hydrated document type, use the following code
type TestDocument = ReturnType<(typeof Test)['hydrate']>;
```

But this gives me `HydratedDocument` that I can get by `HydratedDocument<T>`, which is not what I want since it has all the document methods on it.  
As an alternative I can use just `T` as my return type, but then any `Document<T>` is matching `T`.

I'd like to enforce that the result is a POJO, `LeanDocument` or just the documents fields and virtuals, and not a `Document`.
What approach can I take?

P.S.
I've asked the [same question on stackoverflow](https://stackoverflow.com/questions/76227334/when-migrating-from-mongoose-5-to-7-how-to-get-leandocument-funcitonality), feel free to answer there also, I'd gladly accept the answer.

  [1]: https://mongoosejs.com/docs/migrating_to_7.html#removed-leandocument-and-support-for-extends-document

vkarpov15:
Fix #13424

**Summary**

In #13424 we noticed that Mongoose doesn't infer `{ type: [{ name: String }] }` as a documentarray, whereas `{ type: [new Schema({ name: String })] }` does. This PR fixes that by explicitly looking through the inferred type of the array in the schema and checking against possible definitions like `type: [String]` and `type: [Schema.Types.ObjectId]`.

**Examples**


test/types/schema.test.ts
```diff
@@ -455,7 +455,7 @@ export function autoTypedSchema() {
     array1: [String],
     array2: Array,
     array3: [Schema.Types.Mixed],
-    array4: [{}],
+    // array4: [{}],
```
hasezoey, line 458: what happened to this? shouldnt this be removed if it is going to be unused?

test/types/schema.test.ts
```diff
@@ -408,7 +408,7 @@ export function autoTypedSchema() {
     array1: string[];
     array2: any[];
     array3: any[];
-    array4: any[];
+    // array4: any[];
```
hasezoey, line 411: what happened to this? shouldnt this be removed if it is going to be unused?

hasezoey (approved):
Looks good to me, with some questions

AbdelrahmanHafez (approved):

test/types/schema.test.ts
vkarpov15: This was mistakenly commented out for testing purposes, looks like my fix is incomplete :( thanks for pointing this out.

imranbarbhuiya:
Hi @vkarpov15 , today I was updating mongoose and found all the prop values of the schema types became `unknown` so I tried to check previous versions and found it's starting from `8.0.2`, after checking commits and reverting them, and found this commit it causing the issue. 

Here's my schema

```ts
export const UserSchema = new Schema(
    {
        reminders: {
            type: [
                {
                    type: { type: SchemaTypes.String },
                    date: { type: SchemaTypes.Date },
                    toggle: { type: SchemaTypes.Boolean },
                    notified: { type: SchemaTypes.Boolean },
                },
            ],
            default: [
                { type: 'vote', date: new Date(), toggle: false, notified: false },
                { type: 'daily', date: new Date(), toggle: false, notified: false },
                { type: 'drop', date: new Date(), toggle: false, notified: false },
                { type: 'claim', date: new Date(), toggle: false, notified: false },
                { type: 'work', date: new Date(), toggle: false, notified: false },
            ],
        },
        avatar: {
            type: SchemaTypes.String,
        },
    },
    { timestamps: true },
);

export type IUser = InferSchemaType<typeof UserSchema>;
```

Lemme know if I'm doing anything wrong with the schema. Or if it's a bug, I'll be happy to contribute, lemme me know how can I fix it.

If you want me to create a separate issue instead of this comment then also lemme know and I'll create a separate issue. Thanks

hasezoey:
@imranbarbhuiya 

> If you want me to create a separate issue instead of this comment then also lemme know and I'll create a separate issue. Thanks

seeing as there has not been a reply and this not being the best place to report this, consider opening a new issue referencing this PR, if this is the problem

PR #14102, opened 2023-11-20
Island rhythms/cleanup

vkarpov15:
**Summary**

The `docs:prepare:publish:XYZ` scripts are becoming hard to maintain and test. This cleans things up a bit so at least the logic from `docs:clean:stable` and `mv ./docs/7.x ./tmp` are moved to the `website.js` script. So the actions that don't change the branch are in one script.

**Examples**

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "docs:prepare:publish:stable": "npm run docs:checkout:gh-pages && npm run docs:merge:stable && npm run docs:clean:stable && npm run docs:generate && npm run docs:generate:search",
     "docs:prepare:publish:5x": "npm run docs:checkout:5x && npm run docs:merge:5x && npm run docs:clean:stable && npm run docs:generate && npm run docs:copy:tmp && npm run docs:checkout:gh-pages && npm run docs:copy:tmp:5x",
     "docs:prepare:publish:6x": "npm run docs:checkout:6x && npm run docs:merge:6x && npm run docs:clean:stable && env DOCS_DEPLOY=true npm run docs:generate && npm run docs:move:6x:tmp && npm run docs:checkout:gh-pages && npm run docs:copy:tmp:6x",
-    "docs:prepare:publish:7x": "env DOCS_DEPLOY=true npm run docs:generate && npm run docs:checkout:gh-pages && rimraf ./docs/7.x && ncp ./tmp ./docs/7.x",
+    "docs:prepare:publish:7x": "env DOCS_DEPLOY=true npm run docs:generate && npm run docs:checkout:gh-pages && rimraf ./docs/7.x && mv ./tmp ./docs/7.x",
     "docs:check-links": "blc http://127.0.0.1:8089 -ro",
     "lint": "eslint .",
     "lint-js": "eslint . --ext .js --ext .cjs",
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "docs:prepare:publish:stable": "npm run docs:checkout:gh-pages && npm run docs:merge:stable && npm run docs:clean:stable && npm run docs:generate && npm run docs:generate:search",
     "docs:prepare:publish:5x": "npm run docs:checkout:5x && npm run docs:merge:5x && npm run docs:clean:stable && npm run docs:generate && npm run docs:copy:tmp && npm run docs:checkout:gh-pages && npm run docs:copy:tmp:5x",
     "docs:prepare:publish:6x": "npm run docs:checkout:6x && npm run docs:merge:6x && npm run docs:clean:stable && env DOCS_DEPLOY=true npm run docs:generate && npm run docs:move:6x:tmp && npm run docs:checkout:gh-pages && npm run docs:copy:tmp:6x",
-    "docs:prepare:publish:7x": "git checkout 7.x && npm run docs:clean:stable && env DOCS_DEPLOY=true npm run docs:generate && mv ./docs/7.x ./tmp && npm run docs:checkout:gh-pages && rimraf ./docs/7.x && ncp ./tmp ./docs/7.x",
+    "docs:prepare:publish:7x": "env DOCS_DEPLOY=true npm run docs:generate && npm run docs:checkout:gh-pages && rimraf ./docs/7.x && ncp ./tmp ./docs/7.x",
     "docs:check-links": "blc http://127.0.0.1:8089 -ro",
     "lint": "eslint .",
     "lint-js": "eslint . --ext .js --ext .cjs",
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "docs:merge:6x": "git merge 6.x",
     "docs:test": "npm run docs:generate && npm run docs:generate:search",
     "docs:view": "node ./scripts/static.js",
-    "docs:prepare:publish:stable": "npm run docs:checkout:gh-pages && npm run docs:merge:stable && npm run docs:clean:stable && npm run docs:generate && npm run docs:generate:search",
+    "docs:prepare:publish:stable": "npm run docs:checkout:gh-pages && npm run docs:merge:stable && npm run docs:generate && npm run docs:generate:search",
     "docs:prepare:publish:5x": "npm run docs:checkout:5x && npm run docs:merge:5x && npm run docs:clean:stable && npm run docs:generate && npm run docs:copy:tmp && npm run docs:checkout:gh-pages && npm run docs:copy:tmp:5x",
     "docs:prepare:publish:6x": "npm run docs:checkout:6x && npm run docs:merge:6x && npm run docs:clean:stable && env DOCS_DEPLOY=true npm run docs:generate && npm run docs:move:6x:tmp && npm run docs:checkout:gh-pages && npm run docs:copy:tmp:6x",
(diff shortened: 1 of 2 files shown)


scripts/website.js
```diff
@@ -76,6 +78,48 @@ const tests = [
...
+  for (const folder of foldersToClean) {
+    let files = [];
+    try {
+      files = fs.readdirSync(folder);
+    } catch (err) {
+      if (err.code === 'ENOENT') {
+        continue;
+      }
+    }
+    for (let index = 0; index < files.length; index++) {
```
hasezoey, line 106: wouldnt it be handy to not just also use `const file of files`? (like in the folders above?)

scripts/website.js
```diff
@@ -76,6 +78,48 @@ const tests = [
...
+      }
+    }
+  }
+}
+
+function moveDocsToTemp() {
+  rimraf.sync('./tmp');
+  const folder = `./docs/7.x`;
+  const directory = fs.readdirSync(folder);
+  for (let i = 0; i < directory.length; i++) {
```
hasezoey, line 118: same here: wouldnt it be handy to not just also use `const file of files`? (like in the folders above?)

scripts/website.js
```diff
@@ -76,6 +78,48 @@ const tests = [
...
+      if (files[index].endsWith('.html')) {
+        fs.unlinkSync(`${folder}/${files[index]}`);
+      }
+    }
+  }
+}
+
+function moveDocsToTemp() {
+  rimraf.sync('./tmp');
+  const folder = `./docs/7.x`;
```
hasezoey, line 116: this likely should rely on `versionObj` like the website generation, or are there problems with using that?

scripts/website.js
```diff
@@ -4,9 +4,11 @@ Error.stackTraceLimit = Infinity;
 
 const acquit = require('acquit');
 const fs = require('fs');
+const fsextra = require('fs-extra');
 const path = require('path');
 const pug = require('pug');
 const pkg = require('../package.json');
+const rimraf = require('rimraf');
```
hasezoey, line 11: unless i missed something, i dont think we actually need `rimraf` in the script itself (it is still handy as a cli tool) because nodejs natively supports rm and rmdir (recursive) now

hasezoey (commented):
i think this could be cleaned-up even more by relying on the `versionObj` like the generation of the website

also if all the logic is now in `website.js` for `docs:clean:stable`, shouldnt the script also be changed to use that instead (so we dont have duplicate actions)

if you want you can merge it as-is and i will try to update it with `versionObj` later

vkarpov15:
@hasezoey I addressed your comments, can you please re-review?

scripts/website.js
```diff
@@ -76,6 +77,48 @@ const tests = [
...
+    for (const file of files) {
+      if (file.endsWith('.html')) {
+        fs.unlinkSync(`${folder}/${file}`);
+      }
+    }
+  }
+}
+
+function moveDocsToTemp() {
+  fs.rmSync('./tmp', { recursive: true });
```
hasezoey, line 114: this will likely need the `ENOENT` treatment

scripts/website.js
```diff
@@ -76,6 +77,48 @@ const tests = [
...
+  const folder = versionObj.versionedPath.replace(/^\//, '');
+  const directory = fs.readdirSync(folder);
+  for (const file of directory) {
+    fsextra.moveSync(`${folder}/${file}`, `./tmp/${file}`);
+  }
```
hasezoey, line 119: this likely will need a check for a non-empty `versionObj.versionedPath` or a note in the JSDOC, because if `versionedDeploy` is `false` OR `currentVersion === latestVersion` the `versionedPath` is ""(empty, project root), not `/docs/X.X/` and so would move *all* the project into tmp (and i dont think this is intended, guessing by [`docs:prepare:publish:stable`](https://github.com/Automattic/mongoose/blob/4b065f2aed1391dc9b0b5f5f914d7f01847a3723/package.json#L96C6-L96C33) and the old [`docs:prepare:publish:7x`](https://github.com/Automattic/mongoose/blob/ea855148c7ecdcfc60f46f286165617380083bc7/package.json#L99))
[see this code](https://github.com/Automattic/mongoose/blob/4b065f2aed1391dc9b0b5f5f914d7f01847a3723/scripts/website.js#L261C102-L263)

scripts/website.js
```diff
@@ -76,6 +77,48 @@ const tests = [
...
+    if (err.code !== 'ENOENT') {
+      throw err;
+    }
+  }
+  const foldersToClean = [
+    './docs',
+    './docs/tutorials',
+    './docs/typescript',
+    './docs/api',
+    './docs/source/_docs',
```
hasezoey, line 93: those would likely need to be prefixed with `versionObj.versionedPath` (guessing by [this workflow](https://github.com/Automattic/mongoose/blob/ea855148c7ecdcfc60f46f286165617380083bc7/package.json#L99) and how this function is always called)

scripts/website.js
vkarpov15: This shouldn't be strictly necessary because `moveDocsToTemp()` is only called if `versionedDeploy` is set. But I'll add a redundant check just in case to prevent errors.

scripts/website.js
vkarpov15: I don't think these paths need to be prefixed with `versionedPath`. This `deleteAllHtmlFiles()` function is meant to replace `npm run docs:clean:stable`; which cleans out `index.html`, `docs/*.html`, etc. regardless of whether this is a versioned deploy or not.

In other words, the workflow for a versioned deploy is 1) clear all docs, 2) generate docs, 3) copy docs to `/tmp`, 4) clear `versionedPath`, 5) copy `/tmp` to `versionedPath`.  `deleteAllHtmlFiles()` takes care of step 1, not step 4.

package.json
```diff
@@ -96,7 +96,7 @@
     "docs:prepare:publish:stable": "npm run docs:checkout:gh-pages && npm run docs:merge:stable && npm run docs:clean:stable && npm run docs:generate && npm run docs:generate:search",
     "docs:prepare:publish:5x": "npm run docs:checkout:5x && npm run docs:merge:5x && npm run docs:clean:stable && npm run docs:generate && npm run docs:copy:tmp && npm run docs:checkout:gh-pages && npm run docs:copy:tmp:5x",
     "docs:prepare:publish:6x": "npm run docs:checkout:6x && npm run docs:merge:6x && npm run docs:clean:stable && env DOCS_DEPLOY=true npm run docs:generate && npm run docs:move:6x:tmp && npm run docs:checkout:gh-pages && npm run docs:copy:tmp:6x",
-    "docs:prepare:publish:7x": "git checkout 7.x && npm run docs:clean:stable && env DOCS_DEPLOY=true npm run docs:generate && mv ./docs/7.x ./tmp && npm run docs:checkout:gh-pages && rimraf ./docs/7.x && ncp ./tmp ./docs/7.x",
+    "docs:prepare:publish:7x": "env DOCS_DEPLOY=true npm run docs:generate && npm run docs:checkout:gh-pages && rimraf ./docs/7.x && ncp ./tmp ./docs/7.x",
```
hasezoey, line 99:
```suggestion
    "docs:prepare:publish:7x": "env DOCS_DEPLOY=true npm run docs:generate && npm run docs:checkout:gh-pages && rimraf ./docs/7.x && mv ./tmp ./docs/7.x",
```

to my understanding it does not actually need to copy `tmp` to `/docs/X.X`

also dosnt this still need the `git checkout 7.x`? (because this is not the latest branch and will likely be merged to master)

package.json
```diff
@@ -96,7 +96,7 @@
     "docs:prepare:publish:stable": "npm run docs:checkout:gh-pages && npm run docs:merge:stable && npm run docs:clean:stable && npm run docs:generate && npm run docs:generate:search",
```
hasezoey, line 96:
```suggestion
    "docs:prepare:publish:stable": "npm run docs:checkout:gh-pages && npm run docs:merge:stable && npm run docs:generate && npm run docs:generate:search",
```

shouldnt be necessary anymore, right?

scripts/website.js
hasezoey: > just to be on a consistent page, this is the current workflow (to my understanding) Not entirely correct. 7.x and 6.x workflows don't check out gh-pages, so no steps 1/2. But otherwise correct.

as it said `if latest` - which applies to the master branch (currently 8.x) and `if versioned` it means the `X.x` branches (like `7.x` and `6.x`) - versioned uses gh-pages checkout at step 9

scripts/website.js
```diff
@@ -79,22 +79,24 @@ const tests = [
...
-    './docs',
-    './docs/tutorials',
-    './docs/typescript',
-    './docs/api',
-    './docs/source/_docs',
+    path.join('.', versionObj.versionedPath, 'docs'),
+    path.join('.', versionObj.versionedPath, 'docs', 'tutorials'),
+    path.join('.', versionObj.versionedPath, 'docs', 'typescript'),
+    path.join('.', versionObj.versionedPath, 'docs', 'api'),
+    path.join('.', versionObj.versionedPath, 'docs', 'source', '_docs'),
```
hasezoey, line 94: as a suggestion, you should probably not use `'.'` and use [variable `cwd` instead](https://github.com/Automattic/mongoose/blob/140a118dda67cbf9d9ce62ba0eac1ad0efd13154/scripts/website.js#L16)

PR #14116, opened 2023-11-23
Update `scripts/website.js` to fetch all `test/docs/` files automatically

hasezoey:
**Summary**

This PR changes `scripts/website.js` to automatically fetch and use all `test/docs/` files (unless ignored) plus additional files outside of that directory.
And some additional style changes i noticed.

### DIFF
--- a/scripts/website.js
+++ b/scripts/website.js
@@ -53,7 +53,7 @@ markdown.setOptions({
   }
 });
 
-const testPath = path.resolve(cwd, 'test')
+const testPath = path.resolve(cwd, 'test');
 
 /** additional test files to scan, relative to `test/` */
 const additionalTestFiles = [
@@ -503,7 +503,7 @@ const pathsToCopy = [
   'docs/js',
   'docs/css',
   'docs/images'
-]
+];
 
 /** Copy all static files when versionedDeploy is used */
 async function copyAllRequiredFiles() {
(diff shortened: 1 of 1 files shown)


scripts/website.js
```diff
@@ -52,29 +52,53 @@ markdown.setOptions({
...
+];
+
+/**
+ * Load all test file contents with acquit
+ * @returns {Object[]} acquit ast array
+ */
+async function getTests() {
+  const promiseArray = [];
+
+  for (const file of additionalTestFiles) {
```
vkarpov15, line 76: Can we please combine these two into one loop, `for (const file of [...fs.readdirSync(testDocs), ...additionalTestFiles]) {}` and use `readFileSync()`? With no `flatMap()` at the end. Parallelization is nice, but I would prefer to optimize for concise readability in this code rather than perf; shaving off a few ms on a docs build isn't worth it IMO.

vkarpov15 (commented):
One minor style comment

scripts/website.js
hasezoey: is c0c408c8ac011cd15f483337ae2ec0b3add477fa better? i couldnt avoid the last `.flat()` because otherwise it would constantly create a new array in each loop

vkarpov15 (approved):

Issue #13546, opened 2023-06-27
Option to not create the collection when compiling model


NotAstrid:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Placing `null` in the collectionName argument should make it not create a collection for that model

### Example

`mongoose.model(name, schema, null)`

vkarpov15:
This is an interesting idea. What is the use case for this behavior? Is this mostly a way to create subdocuments?

robertsspaceindustries:
> This is an interesting idea. What is the use case for this behavior? Is this mostly a way to create subdocuments?

Lets you have a model that represents a log, you have multiple websites you want to have logs off. So to improve performance, you want to have the websites' logs in a different collection from each other.

Got it?

vkarpov15:
That's an interesting idea. How would you change the log model's collection?

robertsspaceindustries:
> That's an interesting idea. How would you change the log model's collection?

Current way of doing it (tmk) is getting the mongo connection and inserting the document (`const document = new Model()`) into a collection
Could add an option in `document.save` to set the collection.

Issue #11380, opened 2022-02-11
Mongoose execution time increased 2 times after upgrading 4 to 6


ekorolevanyrun:
**Do you want to request a *feature* or report a *bug*?**

Probably bug

**What is the current behavior?**

Execution time increased after upgrading mongoose (4.13.21 -> 6.2.1) in 2 times

**If the current behavior is a bug, please provide the steps to reproduce.**

We have 2 mongodb replicas (primary and two secondary ones). You can reproduce the problem with a simple script:
```js
const mongoose = require('mongoose');
const RECORDS_AMOUNT = 10000;
const DATA_AMOUNT = { min: 40, max: 400, };

const options = {
    socketTimeoutMS: 15000,
    keepAlive: true,
    ...(mongoose.version === '6.2.1'
    ? { minPoolSize: 32, maxPoolSize: 64 }
    : { poolSize: 64 }),
    ignoreUndefined: true,
  };

const connectionOpts = {
    set: [ { ip: 'mongo_1', port: 27017 }, { ip: 'mongo_2', port: 27017 } ],
    db: 'meteor',
    username: 'docker',
    pass: 'dockerpass',
    rs: 'rs0',
    ssl: false,
    sslCA: '',
    sslCert: ''
}
const connectionString = 'mongodb://docker:dockerpass@mongo_1:27017,mongo_2:27017/meteor?authSource=admin&replicaSet=rs0'
const mongoOptions = {
    ...options,
    ...(connectionOpts.ssl && {
        ssl: true,
        sslValidate: true,
        sslCA: connectionOpts.sslCA,
        sslCert: connectionOpts.sslCert,
        sslKey: connectionOpts.sslCert,
    }),
};

const slowsaveSchema = new mongoose.Schema({
    randomData: [Number,],
});
const slowsaveModel = mongoose.model('slowsave_experiment', slowsaveSchema);

let last100_count=0, last100_sum=0, total_count=0, total_sum=0;
async function addData() {
    const record = new slowsaveModel();
    for (let i = 0; i < Math.random()*(DATA_AMOUNT.max-DATA_AMOUNT.min) + DATA_AMOUNT.min; ++i) {
        record.randomData.push(Math.random());
    };

    const startTime = Date.now();
    await record.save();
    const time= Date.now()-startTime;;
    
    last100_count++; total_count++;
    last100_sum+=time; total_sum+=time;
    if (last100_count>=100) {
[...]

PR #14120, opened 2023-11-24
Perf improvements when creating arrays of primitives

vkarpov15:
**Summary**

Re: #11380, #13546. This one quick improvement of using `push()` instead of `concat()` already helps shave off maybe 10% off of #11380.

Before:

```
l100.avg=1140.53, total.avg=3175.7342
```

After:

```
l100.avg=1249.86, total.avg=2870.4085
```

Gonna work to see what other performance wins we can turn up.

**Examples**

### DIFF
--- a/lib/types/array/methods/index.js
+++ b/lib/types/array/methods/index.js
@@ -375,7 +375,7 @@ const methods = {
         atomics.$push = val;
       } else {
         if (val.length === 1) {
-          atomics.$push.$each.push(val);
+          atomics.$push.$each.push(val[0]);
         } else if (val.length < 10000) {
           atomics.$push.$each.push(...val);
         } else {
--- a/lib/types/array/methods/index.js
+++ b/lib/types/array/methods/index.js
@@ -374,7 +374,13 @@ const methods = {
       if (val != null && utils.hasUserDefinedProperty(val, '$each')) {
         atomics.$push = val;
       } else {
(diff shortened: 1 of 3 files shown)


vkarpov15:
With 7d26ad2 we've managed to shave off a total of 40%:

```
l100.avg=1117, total.avg=1898.9985
```

vs latest 7.x:

```
l100.avg=1116.11, total.avg=3078.2941
```

Mongoose 7.x is now significantly faster (about 35%) than latest Mongoose 4 at pushing elements onto an array:

```
l100.avg=492.34, total.avg=2761.561
```

vkarpov15:
New flamegraph:

image

lib/types/array/methods/index.js
```diff
@@ -374,7 +374,15 @@ const methods = {
       if (val != null && utils.hasUserDefinedProperty(val, '$each')) {
         atomics.$push = val;
       } else {
-        atomics.$push.$each = atomics.$push.$each.concat(val);
+        if (val.length === 1) {
+          atomics.$push.$each.push(val[0]);
+        } else if (val.length < 10000) {
```
hasezoey, line 379: maybe note why this magic number was chosen? (i guess it is because of a stack overflow error)

hasezoey (commented):
Looks good to me, but i am not quite sure what actually changed or performance characteristics (it seems like setters are called less?)

PR #14123, opened 2023-11-25
types: make property names show up in intellisense for UpdateQuery

### LINKED ISSUE #14090: Typescript IntelliSense not working for model.updateOne second parameter
### Prerequisites

### Mongoose version

8.0.1

### Node.js version

18.18.2

### MongoDB server version

7.0.3 Community

5.2.2

### Description

Typescripts interface works for `model.findOne({ /* Autocompletion Working */ })` but not for the second parameter of `model.updateOne({ /* Autocompletion Working */}, { /* Autocompletion missing */ })`

### Steps to Reproduce

```ts
import {Schema, model} from 'mongoose'

export interface IRoles {
    name: string;
    permissions: [string];
    inherit_from: IRoles;
}
export const RolesSchema = new Schema<IRoles>({
    permissions: [{type: String}],
    inherit_from: {type: Schema.Types.ObjectId, ref: 'Roles'}
}, {timestamps: true})


export const RolesModel = model<IRoles>('Roles', RolesSchema)


RolesModel.updateOne({name: '123'}, { /* Autocompletion not working here */ })

RolesModel.findOne({name: '123'})
```

I'm only get the default IntelliSense
image

Vscode seems to get the Inferred type, the second parameter type is `UpdateWithAggregationPipeline | UpdateQuery<IRoles>`
image

### Expected Behavior

I should get the same autocompletion as the first `updateOne` parameter.

vkarpov15:
Fix #14090

**Summary**

I took a look, and with this changes it looks like intellisense picks up property names as well as update operators 
image

**Examples**

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -634,7 +634,7 @@ declare module 'mongoose' {
    * { age: 30 }
    * ```
    */
-  export type UpdateQuery<T> = _UpdateQuery<T> & AnyObject;
+  export type UpdateQuery<T> = AnyKeys<T> & _UpdateQuery<T> & AnyObject;
 
   /**
    * A more strict form of UpdateQuery that enforces updating only


vkarpov15:
This doesn't change the cast type at all, you still end up with `AnyObject` at the end; just helps with intellisense. I'm open to alternatives, and `Partial<TSchema> | AnyObject` may be better in some cases, but the approach in this PR is likely the minimum viable change and I don't see any reason why `Partial<TSchema> | AnyObject` would be better.

pshaddel:
I thought, this issue is complaining about the types: 
https://github.com/Automattic/mongoose/issues/14090#issuecomment-1821260735

What you said is true, the only way is to replace AnyObject(which i guess breaks some stuff).

hasezoey:
should this fix maybe be targeted at 7.x too?

vkarpov15:
@hasezoey yeah I'll cherry pick

PR #14124, opened 2023-11-27
feat(schematype): merge rather than overwrite default schematype validators

### LINKED ISSUE #14070: merge validators from schemaType.prototype.validate definition and schemaType.set
### Prerequisites

### 🚀 Feature Proposal

Hi, As ```validate``` attribute in a schemaType could be ```Function|Array|Object```, and you can set it via ```schemaType.set``` I'll be nice to merge both definitions.

### Motivation

I want to add a validation set for the schemaType Array (I need this for several types with different validations), to check min and max elements. but if I add other validator in the schemaType definition, it get overwritten.

For now my only choice is to create a plugin who searches in all paths and check if there is a min or max option in an "Array" instance and add a pre or post validate hook to the schema or merge the default validators with the ones declared in Schema, either way Schema.prototype.validate ain't accepts arrays

### Example

```js
import db from "mongoose";
await db.connect("mongodb://127.0.0.1/test");
//SchemaType.prototype.validate() can't get an array to set several validators but Schema.Types.set("validate", ...) can
//Comment this plugin to check the problem.
db.plugin(schema => {
  schema.eachPath((path, schemaType) => {
    if (schemaType.validators !== schemaType.constructor.defaultOptions.validate) {
      const validators = [
        ...new Set([].concat(schemaType.constructor.defaultOptions.validate ?? [], schemaType.validators ?? []))
      ];
      schemaType.validators = validators;
    }
  });
});
//custom validators for type Array
db.Schema.Types.Array.set("validate", [
  {
    validator(val, rest) {
      const {options} = this.constructor.schema.path(rest.path);
      if (options.min) {
        if (val.length < options.min) throw `El campo ${rest.path} debe tener al menos ${options.min} elementos`;
      }
      return true;
    },
    propsParameter: true
  },
  {
    validator(val, rest) {
      const {options} = this.constructor.schema.path(rest.path);
      if (options.max) {
        if (val.length > options.max) throw `El campo ${rest.path} debe tener a lo más ${options.max} elementos`;
      }
      return true;
    },
    propsParameter: true
  }
]);
const Schema = new db.Schema({
  a: {
    type: [String],
    min: 2,
    max: 3,
    validate: {
      validator(v) {
        return v.includes("test3");
      },
      message: "el campo {PATH} no incluye 'test3'"
    }
  }
});
const model = db.model("model", Schema);
//console.log("validators", Schema.path("a").validators);                                                                              
let doc;                                                                                                                               
const a = [];                                                                                                                          
for (let i = 1; i <= 5; i++) {                                                                                                         
  a.push("test" + i);                                                                                                                  
  doc = new model({a});                                                                                                                
  console.log(doc);                                                                                                                    
  const error = doc.validateSync();                                                                                                    
  console.log(error?.errors?.a?.reason ?? error?.errors?.a?.message ?? "Sin errores");                                                 
}                                                                                                                                      
db.disconnect();                                                                                                                       

```

vkarpov15:
Fix #14070

**Summary**

#14070 makes a fair point that `validate` overwriting default schematype validators is a point of friction, and merging validators would make the default validator API much more useful. I figure it is reasonable to ship this with 8.1, what do you think @AbdelrahmanHafez @hasezoey ?

**Examples**

### DIFF
--- a/test/schematype.test.js
+++ b/test/schematype.test.js
@@ -210,11 +210,15 @@ describe('schematype', function() {
 
   it('merges default validators (gh-14070)', function() {
     class TestSchemaType extends mongoose.SchemaType {}
-    TestSchemaType.set('validate', v => typeof v === 'string');
+    TestSchemaType.set('validate', checkIfString);
 
     const schemaType = new TestSchemaType('test-path', {
-      validate: v => v.length === 2
+      validate: checkIfLength2
     });
+
+    assert.equal(schemaType.validators.length, 2);
+    assert.equal(schemaType.validators[0].validator, checkIfString);
+    assert.equal(schemaType.validators[1].validator, checkIfLength2);
 
     let err = schemaType.doValidateSync([1, 2]);
     assert.ok(err);
@@ -226,6 +230,13 @@ describe('schematype', function() {
 
     err = schemaType.doValidateSync('ab');
     assert.ifError(err);
+
+    function checkIfString(v) {
+      return typeof v === 'string';
+    }
+    function checkIfLength2(v) {
+      return v.length === 2;
+    }
   });
 
   describe('set()', function() {
--- a/test/schematype.test.js
+++ b/test/schematype.test.js
@@ -208,6 +208,25 @@ describe('schematype', function() {
     });
   });
 
+  it('merges default validators (gh-14070)', function() {
+    class TestSchemaType extends mongoose.SchemaType {}
+    TestSchemaType.set('validate', v => typeof v === 'string');
+
+    const schemaType = new TestSchemaType();
+    schemaType.validate(v => v.length === 2);
+
+    let err = schemaType.doValidateSync([1, 2]);
+    assert.ok(err);
+    assert.equal(err.name, 'ValidatorError');
+
+    err = schemaType.doValidateSync('foo');
+    assert.ok(err);
+    assert.equal(err.name, 'ValidatorError');
+
+    err = schemaType.doValidateSync('ab');
+    assert.ifError(err);
+  });
+
   describe('set()', function() {
     describe('SchemaType.set()', function() {
       it('SchemaType.set, is a function', () => {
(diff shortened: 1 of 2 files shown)


test/schematype.test.js
```diff
@@ -208,6 +208,25 @@ describe('schematype', function() {
     });
   });
 
+  it('merges default validators (gh-14070)', function() {
+    class TestSchemaType extends mongoose.SchemaType {}
+    TestSchemaType.set('validate', v => typeof v === 'string');
```
hasezoey, line 213: maybe consider making it a stand-alone function and later check against it being still contained

test/schematype.test.js
```diff
@@ -208,6 +208,25 @@ describe('schematype', function() {
     });
   });
 
+  it('merges default validators (gh-14070)', function() {
```
hasezoey, line 211: i dont understand what this test tests, because it passes even latest master (without the changes)

if i understand the code in `schematype.js` correctly, shouldnt there be options passed to `new TestSchemaType();`?

hasezoey (changes requested):
The test added currently passes on master (84db3a0351b6093cc14daf80abe811e590ab9b94), without requiring the changes from this PR

test/schematype.test.js
vkarpov15: You're right, fixed in 4d3243f. This is additional motivation for this change: the following retains schematype-level default validators, even without the change in this PR

```javascript
const schema = new Schema({ arr: [Number] });
schema.path('arr').validate(v => v.length === 2);
```

The following does not without this change

```javascript
const schema = new Schema({ arr: { type: [Number], validate: v => v.length === 2 } });
```

A bit of a gotcha.

hasezoey: > The following does not without this change
> const schema = new Schema({ arr: { type: [Number], validate: v => v.length === 2 } });
> A bit of a gotcha.

with this PR the defaults are also applied there then?

hasezoey (approved):
now the test looks good to me

PR #14135, opened 2023-12-01
Call transform object with single id instead of array when populating a justOne path under an array

### LINKED ISSUE #14073: Question: I'm getting an array of ids on the populate `transform` callback
### Prerequisites

### Mongoose version

6.11.1

### Node.js version

18.17.1

### MongoDB version

5.0

### Operating system

Linux

Ubuntu 22.04

### Issue

Hi, I'm not sure if this is a bug, or just expected behavior is some case, so I'm posting this as a Help topic.

I'm using the `transform` callback to leave the `id` when the object being populated isn't found, but I found that the id argument (the 2nd argument) sometimes is an id (presumably, when populating a query, instead of a single document).

This doesn't seems to be documented.

Can you please clarify about this? Thanks in advance!

vkarpov15:
**Summary**

Fix #14073 

A bit of a weird fix, but it works. The idea is that, if you call populate on `items.brand`, where `items` is a doc array and `brand` is an ObjectId ref, the `transform` option will get an array of ids as 2nd arg, not a single id. To work around this, we need to track what position in the array is being assigned, and get the original id by that index.

**Examples**

### DIFF
--- a/lib/helpers/populate/assignVals.js
+++ b/lib/helpers/populate/assignVals.js
@@ -80,6 +80,8 @@ module.exports = function assignVals(o) {
       return valueFilter(val[0], options, populateOptions, _allIds);
     } else if (o.justOne === false && !Array.isArray(val)) {
       return valueFilter([val], options, populateOptions, _allIds);
+    } else if (o.justOne === true && !Array.isArray(val) && Array.isArray(_allIds)) {
+      return valueFilter(val, options, populateOptions, val == null ? val : _allIds[o.rawOrder[val._id]]);
     }
     return valueFilter(val, options, populateOptions, _allIds);
   }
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
i dont quite know what `justOne` has to do with this here, unless it is implicitly set somewhere (i dont see it mentioned in the test aside from the name).

aside from that i think it looks good (though i am not fully sure about the call structure and the index resetting)

vkarpov15:
Mongoose infers `justOne` from the schema, so if you have `docArr: [{ ref: { type: 'ObjectId', ref: 'MyModel' } }]` Mongoose will infer that `docArr.ref` has `justOne = true` based on the schema definition. We allow overriding `justOne` as a populate option as a workaround in case of bugs.

PR #14156, opened 2023-12-05
refactor: remove `utils.options` / `utils.object.shallowCopy`

hasezoey:
**Summary**

This PR removes function `utils.options` (`lib/utils.js`) which shallow-copies objects, which is now solved with JS-native spread-operator.

i got this idea because of #14155.

my only concern would be the removal of `clone`'s that was done previously (with 6820605d5082f94bba5a5b7ccae2dfe8c5d0a0c8), no tests fail and the comments mentioned it was because `utils.options` directly assigned to the second parameter.

as for performance, i have not measured, but i would assume it is about the same, if not better thanks to the removal of `clone`s


vkarpov15 (approved):
Thanks :+1:

PR #14159, opened 2023-12-06
add missing Type for applyDefaults

jaypea:
**Summary**
The Typescript types definition file is missing a definition for `applyDefaults()` in `Model` Interface.

**Examples**

Error was
```
TS2339: Property 'applyDefaults' does not exist on type 'Model<Item, {}, {}, {}, any>'.
```

### DIFF
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -185,7 +185,7 @@ declare module 'mongoose' {
 
     /* Apply defaults to the given document or POJO. */
     applyDefaults(obj: AnyObject|TRawDocType): TRawDocType;
-    
+
     /**
      * Sends multiple `insertOne`, `updateOne`, `updateMany`, `replaceOne`,
      * `deleteOne`, and/or `deleteMany` operations to the MongoDB server in one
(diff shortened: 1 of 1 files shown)


types/models.d.ts
```diff
@@ -183,6 +183,9 @@ declare module 'mongoose' {
     /* Cast the given POJO to the model's schema */
     castObject(obj: AnyObject, options?: { ignoreCastErrors?: boolean }): TRawDocType;
 
+    /* Apply defaults to the given document or POJO. */
+    applyDefaults(obj: AnyObject|TRawDocType): TRawDocType;
```
vkarpov15, line 187: Minor detail but I think having 2 separate overrides `applyDefaults(obj: TRawDocType): TRawDocType;` and `applyDefaults(obj: AnyObject): AnyObject` would be more correct. You can technically pass any object into `applyDefaults()`, even one that doesn't match the schema, and it will still apply defaults.

vkarpov15 (changes requested):

types/models.d.ts
jaypea: Thanks for the suggestions. i've changed it accordingly

vkarpov15 (approved):
Thanks :+1:

PR #14161, opened 2023-12-07
fix(update): set CastError path to full path if casting update fails

### LINKED ISSUE #14114: CastError doesn't return the entire nested path
### Prerequisites

### Mongoose version

8.0.0

### Node.js version

20.9.0

### MongoDB version

6.2.0

### Operating system

None

_No response_

### Issue

Hello,
I'm testing Mongoose error feedback. When testing a ValidationError, the response returns a nested path, which allows me to trace the field in error in my object that I passed in my query. 
Example: `'accessories.0.additionals.0.k'`.
It's very useful for browsing and finding fields in error.

However, when testing a CastError, the return sends me the error by browsing the object. This means I can't go back down into my object to find my error correctly. 
Example: for arrays, the index isn't specified, so it's impossible to retrieve it.

Schema : 
```ruby 
{
    id: Schema.Types.ObjectId,
    name: String,
    accessories: [
        {
            isEnabled: Boolean,
            additionals: [
                {
                    k: String,
                    v: Number,
                },
            ],
        }
    ],
    actions: [
        {
            type: String,
            isEnabled: Boolean,
            subActions: [
                {
                    type: String,
                    isEnabled: Boolean,
                }
            ],
        },
    ],
}
```

Example of a body sent :
```ruby 
{
    "id": "6499a3ff77eb47522397bbc1",
    "name": "Soldier",
    "accessories": [
        {
            "isEnabled": true,
            "additionals": [
                {
                    "k": ["helmet"], 	        // mock an error type here
                    "v": 1,
                },
                {
                    "k": "shoes",
                    "v": 2,
                },
            ],
        }
    ],
    "actions": [
        {
            "type": "move",
            "isEnabled": true,
            "subActions": [
                {
                    "type": "move right",
                    "isEnabled": true,
                }
            ],
        },
    ],
}
```

And the server returns this CastError : 
```
CastError: Cast to embedded failed for value "{
   isEnabled: true,
   additionals: [
     { k: [Array], v: 1 },
     { k: 'shoes', v: 50 },
   ],
   _id: '654d005f44b6b34ae2e19c43'
 }" (type Object) at path "accessories" because of "CastError" {
     stringValue: '"{\n' +
       "  isEnabled: true,\n" +
       '  additionals: [\n' +
       "    { k: [Array], v: 1 },\n" +
       "    { k: 'shoes', v: 2 },\n" +
       '  ],\n' +
       "  _id: '654d005f44b6b34ae2e19c43'\n" +
       '}"',
   messageFormat: undefined,
   kind: 'embedded',
   value: {
       isEnabled: true,
       additionals: [ [Object], [Object] ],
       _id: '654d005f44b6b34ae2e19c43'
   },
   path: 'accessories', 
   reason: CastError: Cast to embedded failed for value "{
         k: [ 'helmet' ],
         v: 1,
         _id: '654d005f44b6b34ae2e19c44'
     }" (type Object) at path "additionals" because of "CastError" {
         stringValue: '"{\n' +
           "  k: [ 'helmet' ],\n" +
           "  v: 1,\n" +
           "  _id: '654d005f44b6b34ae2e19c44'\n" +
           '}"',
       messageFormat: undefined,
       kind: 'embedded',
       value: { k: [Array], v: [Object], _id: '654d005f44b6b34ae2e19c44' },
       path: 'additionals',
       reason: CastError: Cast to string failed for value "[ 'helmet' ]" (type Array) at path "k" {
            stringValue: `"[ 'helmet' ]"`,
            messageFormat: undefined,
            kind: 'string',
            value: [Array],
            path: 'k', 
            reason: null,
            valueType: 'Array'
          },
       valueType: 'Object'
     },
   valueType: 'Object'
 }
```

Example of a ValidationError : 
```
ValidationError : 
Error: Validation failed: accessories.0.additionals.0.k: "helme" doesn't belong to the defined values. {
   errors: {
     'accessories.0.additionals.0.k': ValidatorError: "helme" doesn't belong to the defined values.) {
       properties: [Object],
       kind: 'enum',
       path: 'accessories.0.additionals.0.k',                         // expected path for CastError
       value: 'helme',
       reason: undefined,
       [Symbol(mongoose:validatorError)]: true
     }
   },
   _message: 'Validation failed'
 }
```

vkarpov15:
Fix #14114

**Summary**

Currently, CastError only shows the path relative to parent, which is less than ideal in the case where casting update throws an error; in that case, Mongoose just throws the CastError, rather than wrapping the CastError in a `ValidationError` like how `save()` does. This PR makes CastError use `$fullPath` instead when casting updates

**Examples**

### DIFF
--- a/lib/schema/bigint.js
+++ b/lib/schema/bigint.js
@@ -212,7 +212,14 @@ SchemaBigInt.prototype.castForQuery = function($conditional, val, context) {
     return this.applySetters(null, val, context);
   }
 
-  return this.applySetters(val, context);
+  try {
+    return this.applySetters(val, context);
+  } catch (err) {
+    if (err instanceof CastError && err.path === this.path && this.$fullPath != null) {
+      err.path = this.$fullPath;
+    }
+    throw err;
+  }
 };
 
 /**
--- a/lib/schema/boolean.js
+++ b/lib/schema/boolean.js
@@ -256,7 +256,14 @@ SchemaBoolean.prototype.castForQuery = function($conditional, val, context) {
     return this.applySetters(null, val, context);
   }
 
-  return this.applySetters(val, context);
+  try {
+    return this.applySetters(val, context);
+  } catch (err) {
+    if (err instanceof CastError && err.path === this.path && this.$fullPath != null) {
+      err.path = this.$fullPath;
+    }
+    throw err;
+  }
 };
 
 /**
--- a/lib/schema/buffer.js
+++ b/lib/schema/buffer.js
@@ -279,7 +279,16 @@ SchemaBuffer.prototype.castForQuery = function($conditional, val, context) {
     }
     return handler.call(this, val);
   }
-  const casted = this.applySetters(val, context);
+
+  let casted;
+  try {
+    casted = this.applySetters(val, context);
+  } catch (err) {
+    if (err instanceof CastError && err.path === this.path && this.$fullPath != null) {
+      err.path = this.$fullPath;
+    }
+    throw err;
+  }
   return casted ? casted.toObject({ transform: false, virtuals: false }) : casted;
 };
 
--- a/lib/schema/date.js
+++ b/lib/schema/date.js
@@ -407,7 +407,14 @@ SchemaDate.prototype.$conditionalHandlers = {
 
 SchemaDate.prototype.castForQuery = function($conditional, val, context) {
   if ($conditional == null) {
-    return this.applySetters(val, context);
+    try {
+      return this.applySetters(val, context);
+    } catch (err) {
+      if (err instanceof CastError && err.path === this.path && this.$fullPath != null) {
+        err.path = this.$fullPath;
+      }
+      throw err;
+    }
   }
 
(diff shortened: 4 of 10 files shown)


vkarpov15:
@hasezoey this change is just for `castForQuery()`. `Decimal128` uses SchemaType base class's `castForQuery`, so that's covered. It looks like uuid isn't using `applySetters()` though, I'll check.

hasezoey (commented):
Code looks good to me, though it seems deno does not like the UUID change

vkarpov15:
@hasezoey the deno issue looks unrelated, it's an unfortunate quirk with how our tests are written: `Test.findOneAndUpdate({ _id: doc._id })` is equivalent to `Test.findOneAndUpdate({}, { _id: doc._id })` because `findOneAndUpdate()` with 1 arg means Mongoose treats the 1 arg as an `update` parameter, not a `filter`. Which means that, if there's any dangling data in the underlying collection, the test will fail. I'm not sure why there's dangling data in deno tests here, but either way we should improve the tests to make them more durable, which I did in 6f7b659

PR #14171, opened 2023-12-11
fix: cast error when there is an elemMatch in the and clause

tosaka-n:
**Summary**
- after this PR is merged
	- https://github.com/Automattic/mongoose/pull/13938
- Error in cast because options.discriminatorKey is obtained before checking if the schema is null
	- here:
		- https://github.com/Automattic/mongoose/pull/13938/files#diff-5078cdf8af89c694947cdcb5e60fcd3e29dda608e62642ecf5fc3eeb97e43ce1R71
- Changed to refer to context.schema as well as this PR
	- https://github.com/Automattic/mongoose/commit/266804b995ae715cd4746d0e5b687e7fb96441fc

**Examples**
query:
```
{
  users: {
    $elemMatch: {
      $and: [
        { name: "test" },
        $or: [
          { is_deleted: true },
          { is_deleted: null } 
        ]
      ]
    }
  }
}
```

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -637,7 +637,7 @@ handle.$or = createLogicalQueryOperatorHandler('$or');
 handle.$and = createLogicalQueryOperatorHandler('$and');
 handle.$nor = createLogicalQueryOperatorHandler('$nor');
 
-function createLogicalQueryOperatorHandler(op, context) {
+function createLogicalQueryOperatorHandler(op) {
   return function logicalQueryOperatorHandler(val, context) {
     if (!Array.isArray(val)) {
       throw new TypeError('conditional ' + op + ' requires an array');
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -637,15 +637,15 @@ handle.$or = createLogicalQueryOperatorHandler('$or');
 handle.$and = createLogicalQueryOperatorHandler('$and');
 handle.$nor = createLogicalQueryOperatorHandler('$nor');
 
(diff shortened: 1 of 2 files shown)


lib/schema/array.js
```diff
@@ -637,15 +637,15 @@ handle.$or = createLogicalQueryOperatorHandler('$or');
 handle.$and = createLogicalQueryOperatorHandler('$and');
 handle.$nor = createLogicalQueryOperatorHandler('$nor');
 
-function createLogicalQueryOperatorHandler(op) {
-  return function logicalQueryOperatorHandler(val) {
+function createLogicalQueryOperatorHandler(op, context) {
```
satoshun00, line 640: It doesn't seem necessary to take `context` as an argument in this scope.

vkarpov15 (changes requested):
Code is LGTM, but can you please add a test case?

test/model.query.casting.test.js
```diff
@@ -791,6 +791,38 @@ describe('model query casting', function() {
     assert.ok(res);
     assert.deepStrictEqual(res.map(doc => doc.arr[1].id), ['two', 'three']);
   });
+
+  it('should not throw a cast error when dealing with an array of objects in combination with $elemMatch and nested $and', async function() {
```
tosaka-n, line 795: nits: I couldn't think of an appropriate test name, so I'm open to suggestions.

vkarpov15 (approved):
Thanks :+1:

Issue #13859, opened 2023-09-14
Assigning a list that contains null to itself causes error


Spade-and-Archer:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.5.0

### Node.js version

18.17.1

### MongoDB server version

6.0.9

### Typescript version (if applicable)

5.2.2

### Description

When you have a list that contains null in a mongoose document and then assign it to itself, you get an immediate error:

`TypeError: Cannot set properties of null (setting 'isNew')`

This occurs on assignment, not on save. But it isn't just because the list contains a null value, because a "bare" list with a null value is fine. It's only when it's a "hydrated" mongoose list with a null value that it's a problem. This was not a problem in mongoose 5.13.5, but only because I problem when we upgraded to the latest version.

Also, just for the record, the reason I was doing this was to assign values based on a post request using the patter `doc.property = req.body.property ?? doc.property`. My interim solution was to just do `if(req.body.property){doc.property = req.body.property}`

Code to reproduce this that explains exactly what works, what doesn't, and what is expected.  

### Steps to Reproduce

```
const RecentDataPointSchema = new Schema({
    date: Date,
    value: Number,
})
const  TestingAssigningPropsToThemselvesModel =  mongoose.model("TestingMultiDimArrays ", new Schema( {
    intArray: [Number],
    strArray: [String],
    objArray:[RecentDataPointSchema]
}));
let testObject = new TestingAssigningPropsToThemselvesModel({
    intArray: [1, 2, 3],
    strArray: ["a", "b", "c"],
    objArray: [
        {date: new Date(1000), value: 1},
        {date: new Date(2000), value: 2},
        {date: new Date(3000), value: 3},
    ]
})
//I expect this to save and I expect the self assignment to work
await testObject.save();
testObject.intArray = testObject.intArray;
testObject.strArray = testObject.strArray;
testObject.objArray = testObject.objArray;

//this works no problem. So assigning an array with null values is not intrinsically problematic
//(this also works if you do it when the list originally had nulls in it)
testObject.intArray = [1, 2, 3, null];
testObject.strArray = ["a", null, "c"];
testObject.objArray = [
    {date: new Date(1000), value: 1},
    null,
    {date: new Date(3000), value: 3},
];

//the object does save and the self assignment does work. No errors, all as expected
await testObject.save();

//this object has nulls mixed into it's lists. I personally leverage this ability to have nulls in
[...]

PR #14172, opened 2023-12-11
null check

rohanricky:
**Summary**

null check is missing in $__set method which is causing `TypeError: Cannot set properties of null (setting 'isNew')`

**Repro script**

```
var mongoose = require('mongoose')
console.log(mongoose.version)
let Schema = mongoose.Schema
var eventSchema = new Schema({
  name: { type: String },
  field3: { type: Schema.Types.Mixed },
  __stateBeforeSuspension: {
    field1: { type: String },
    field2: { type: String },
    jsonField: {
        name: { type: String },
        ex: { type: String }
    }
  }
})

eventSchema.add({
  test: { 
    val1: { type: String },
    val2: { type: String }
   }
})

eventSchema.add({
  subSchema4: [{ field1: { type: String } }, { field2: { type: String } }]
})

eventObj.save(function (err, eventObj) {
  var newObject = eventObj.toObject()
  newObject.subSchema4[0] = null
  eventObj.set(newObject)

  eventObj.save(function (err, eventObj1) {

    var newObject2 = eventObj1.toObject()
    newObject2.__stateBeforeSuspension.jsonField.name = 'test4'
    eventObj1.set(newObject2)
  })
})

```

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1689,7 +1689,7 @@ Document.prototype.$__set = function(pathToMark, path, options, constructing, pa
     val[arrayAtomicsSymbol] = priorVal[arrayAtomicsSymbol];
     val[arrayAtomicsBackupSymbol] = priorVal[arrayAtomicsBackupSymbol];
     if (utils.isMongooseDocumentArray(val)) {
-      val.forEach(doc => { doc.isNew = false; });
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
LGTM. Looks like this is a duplicate of #13859, which was fixed by #13883. I'll add a test case from #13883 after merging

PR #14173, opened 2023-12-11
fix(document): avoid treating nested projection as inclusive when applying defaults

### LINKED ISSUE #14115: Unexpected behavior using different kinds of select
### Prerequisites

### Mongoose version

6.x.x

### Node.js version

20.9.0

### MongoDB version

5.10.19

Hey guys 

I have an issue with using different kinds of `.select()` query

First case. The code below returns the expected behavior 
```
const selectFields = {
  _id: 1,
  livemode: 1,
  'payment.clientPaymentMethod': 1,
  'donation.frequency': 1,
};
User.findOne().select(selectFields) 
/* result   
  livemode: true,
  donation: { frequency: 'once' },
  payment: { clientPaymentMethod: 'paypal' },
  */

```

Second case. If I use another kind of selection we could see different behaviour

```
const selectFields = {
  _id: 1,
  livemode: 1,
  'payment': {
     clientPaymentMethod: 1
   },
  'donation': {
     frequency: 1
   },
};
User.findOne().select(selectFields) 
/* result   
  livemode: true,
  donation: { frequency: 'once', amount: 0, isMobile: false, },  <- here extra fields are selected and set by default
  payment: { clientPaymentMethod: 'paypal', id: '' },
  */

```
I am not sure that the second case is valid for use, but it works. There are no rules for selecting nested fields in [docs](https://mongoosejs.com/docs/api/query.html#Query.prototype.select())(maybe I missed)

I expected that the second case would work like the first case. Is it valid behavior?

vkarpov15:
Fix #14115

**Summary**

Right now nested projections, like `{ prop: { subprop: 1 } }`, throw off the logic that checks for projections when applying defaults. So `{ prop: { subprop: 1 } }` will still apply defaults on all properties underneath `prop` that have defaults _other than_ `subprop`.

With this PR, we won't treat `{ prop: { subprop: 1 } }` as `{ prop: 1}` for applying defaults, which is what our `applyDefaults()` function does now.

**Examples**

### DIFF
--- a/lib/helpers/document/applyDefaults.js
+++ b/lib/helpers/document/applyDefaults.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const isNestedProjection = require('../projection/isNestedProjection');
+
 module.exports = function applyDefaults(doc, fields, exclude, hasIncludedChildren, isBeforeSetters, pathsToSkip) {
   const paths = Object.keys(doc.$__schema.paths);
   const plen = paths.length;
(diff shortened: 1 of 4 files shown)


hasezoey (commented):
the code looks good, but any tests for this?

vkarpov15:
@hasezoey thanks for reminder, I pushed tests in dadac4e

hasezoey (approved):

PR #14176, opened 2023-12-13
Allow defining index on base model that applies to all discriminators

peplin:
**Summary**

We would like to define an index on a base model, with a partial filter so that it only applies to discriminators. It's not a discriminator index (as described in #11424), because it's not specific to any one child. 

We were surprised that these indexes were being silently ignored during syncing, and wondered if you are open to changing the behavior. Alternatively, please let me know if this key is just not necessary in the filter. I'm new to both Mongoose and MongoDB. Thanks!

**Examples**

I've added a test, where you can see that getRelatedIndexes no longer skips indexes on base models with the discriminator key, if it's just checking existence:

```
__t: { $exists: true }
```

### DIFF
--- a/lib/helpers/indexes/getRelatedIndexes.js
+++ b/lib/helpers/indexes/getRelatedIndexes.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const hasDollarKeys = require('../query/hasDollarKeys');
+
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
I made a quick change to avoid all partial filter expressions that aren't an explicit `$eq`. But otherwise LGTM, thanks :+1:

Issue #14164, opened 2023-12-07
Mongoose `bulkWrite` Inconsistency: Alters Model Instance and Fails on Schemas with Timestamps


NavpreetDevpuri:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.0.2

### Node.js version

21.0.0

### MongoDB server version

6.0.4

### Description

The primary issue lies with the `bulkWrite` operation in Mongoose. When using a Model instance in the `$set` clause, the behavior varies depending on the model's schema configuration:

- **With Timestamps**: The `bulkWrite` operation unexpectedly modifies the original object passed in the `$set` clause.
- **Without Timestamps**: The operation fails, throwing an error "Update document requires atomic operators".

In contrast, the `updateOne` method works correctly in both scenarios, maintaining the integrity of the original object and completing the update without errors. This difference emphasizes the specific challenge with `bulkWrite` in handling Model instances.

### Steps to Reproduce

Following code reproduce the issue 
```node
const mongoose = require('mongoose');

mongoose.connect('mongodb://localhost:27017/test', {
  useNewUrlParser: true,
  useUnifiedTopology: true
});

const productSchema = new mongoose.Schema({
  name: String,
  properties: { type: mongoose.Schema.Types.Mixed, default: {} },
}, {
  timestamps: true
});

const productSchemaWithoutTimestamps = new mongoose.Schema({
  name: String,
  properties: { type: mongoose.Schema.Types.Mixed, default: {} },
});

const Product = mongoose.model('Product', productSchema);
const ProductWithoutTimestamps = mongoose.model('ProductWithoutTimestamps', productSchemaWithoutTimestamps);

function printLogsForProduct(Model, product) {
  console.log(`   JSON.stringify(product) -> ${JSON.stringify(product)}`);
  console.log(`   typeof product -> ${typeof product}`);
  console.log(`   product instanceof Model -> ${product instanceof Model}`);
  console.log(`   product._id -> ${product._id}`);
}

async function tryBulkWriteUpdateOne(Model, product) {
  product.properties.color = "Red";
  console.log(`\n${Model.modelName}: Before bulkWrite: `)
  printLogsForProduct(Model, product);
  await Model.bulkWrite([{
    updateOne: {
      filter: { _id: product._id },
      update: { $set: product },
    }
  }]);
  console.log(`\n${Model.modelName}: After bulkWrite: `)
[...]

PR #14197, opened 2023-12-27
fix(model): deep clone bulkWrite() updateOne arguments to avoid mutating documents in update

vkarpov15:
**Summary**

#14164 points out that we don't clone arguments to `bulkWrite()`, so Mongoose can actually end up updating documents in `bulkWrite()` ops.

**Examples**

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -41,7 +41,7 @@ integrating Mongoose with external tools and frameworks.
 * [Transactions](transactions.html)
 * [MongoDB Driver Deprecation Warnings](deprecations.html)
 * [Testing with Jest](jest.html)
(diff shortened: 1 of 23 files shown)


hasezoey (commented):
Code looks good to me, though i am not fully sure what the code in question does

aside from that, there seem to be a bunch of unrelated empty-line removals, are those intended?

vkarpov15:
The code in question casts the `updateOne` ops in a `bulkWrite()` call, like the following:

```javascript
await Time.bulkWrite([{
  updateOne: {
    filter: { _id: timeDoc._id },
    update: { $set: timeDoc } // <-- cast this
  }
}]);
```

I'll undo the whitespace changes, those are unnecessary

PR #14206, opened 2023-12-29
fix(ChangeStream): avoid suppressing errors in closed change stream

### LINKED ISSUE #14177: Change stream: Mongoose suppresses error event after initial close
### Prerequisites

### Mongoose version

7.5.3

### Node.js version

20.5.1

### MongoDB server version

6.0

_No response_

### Description

In Mongoose, a change in the error event handling for MongoDB change streams appears to have inadvertently affected the ability to catch errors after the connection has been closed. Specifically, the commit [386c366](https://github.com/Automattic/mongoose/commit/386c366817d3931aef7494d5a3d2b208072458b4) introduced a change where the error event is suppressed after the change stream has been initially closed. This suppression prevents the propagation of errors that the MongoDB driver throws during reconnection attempts, as outlined in the MongoDB documentation for [Change Streams](https://www.mongodb.com/docs/v7.0/changeStreams/#connect).

The workaround is to subscribe directly on the driverChangeStream:
```
this.changeStream.driverChangeStream.on('error', (error) => {
      console.log('Driver change stream error');
      //this.reconnect();
});
```

### Steps to Reproduce

1. Set up a MongoDB change stream using Mongoose.
2. Intentionally disconnect from the database to close the change stream.
3. Wait for the MongoDB to try to reconnect the change stream
4. Error thrown by MongoDB driver is not propagated to the change stream "error" event subscription.

### Expected Behavior

The expected behavior is that Mongoose should emit error events from the MongoDB change stream even after the stream has been closed. This would enable applications to catch and handle these errors, especially to implement custom reconnection logic in scenarios where maintaining an active change stream is critical.

vkarpov15:
Fix #14177

**Summary**

It looks like the potential bug we mentioned in #12890 was actually correct behavior; MongoDB change streams can emit 'error' events after 'close'. The original reason for this change was to avoid a flakey test, so I think we should just remove the `if (ev === 'error' && this.closed)` check.

**Examples**

### DIFF
--- a/lib/cursor/ChangeStream.js
+++ b/lib/cursor/ChangeStream.js
@@ -60,12 +60,6 @@ class ChangeStream extends EventEmitter {
 
         driverChangeStreamEvents.forEach(ev => {
           this.driverChangeStream.on(ev, data => {
-            // Sometimes Node driver still polls after close, so
-            // avoid any uncaught exceptions due to closed change streams
-            // See tests for gh-7022
-            if (ev === 'error' && this.closed) {
-              return;
-            }
             if (data != null && data.fullDocument != null && this.options && this.options.hydrate) {
               data.fullDocument = this.options.model.hydrate(data.fullDocument);
             }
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
Code looks good to me

> It looks like the potential bug we mentioned in https://github.com/Automattic/mongoose/pull/12890 was actually correct behavior

i dont quite understand what #12890 has to do with this, from my quick re-skim, it does not mention a potential bug

vkarpov15:
@hasezoey sorry, I referenced the incorrect PR. Should have referenced #12919:

image

PR #14212, opened 2024-01-01
Issue14204

donoftime2018:
**Summary**

**Examples**

### DIFF
--- a/insertMany.ts
+++ b/insertMany.ts
@@ -1,0 +1,14 @@
+import * as mongoose from "mongoose";
+
+export class RepositoryBase<T> {
+    protected model: mongoose.Model<T & mongoose.Document>;
+
+    constructor(schemaModel: mongoose.Model<T & mongoose.Document>) {
+        this.model = schemaModel;
+    }
+
+    async insertMany(elems: T[]): Promise<T[]> {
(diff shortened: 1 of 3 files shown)


insertMany.ts
```diff
@@ -0,0 +1,14 @@
+import * as mongoose from "mongoose";
```
vkarpov15, line 1: These changes are irrelevant and unnecessary

types/query.d.ts
```diff
@@ -99,82 +99,10 @@ declare module 'mongoose' {
     PopulateOption,
     SessionOption {
     arrayFilters?: { [key: string]: any }[];
-    batchSize?: number;
```
vkarpov15, line 102: What makes you think that deleting all the query options that operations other than `updateOne()` use is a valid solution?

updateItem.js
```diff
@@ -0,0 +1,46 @@
+const mongoose = require('mongoose');
```
vkarpov15, line 1: This isn't relevant to the issue at hand, please delete

vkarpov15 (commented):
This PR is way off base, and #14228 fixes this issue anyway, so I will close this

Issue #14024, opened 2023-11-01
modifiedpaths are not returning as expected when we use .set() with nested paths as an object in v5.12.5 vs v6.x.x


sakethyerra:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.12.1

### Node.js version

14.18.1

### MongoDB server version

4.4

### Description

Hi, the following code behaves differently in both versions v5.12.5 and v6.x.x.
Note: modified paths are not returning as expected when we use .set() with nested paths as an object.

```
var mongoose = require('mongoose')
let Schema = mongoose.Schema
var eventSchema = new Schema({
  name: { type: String },
  __stateBeforeSuspension: {
    field1: { type: String },
    field2: { type: String },
    jsonField: {
        name: { type: String }
    }
  }
})
var Event = mongoose.model('Event', eventSchema)
async function run() {
  await mongoose.connect('mongodb://localhost:27017');
  let eventObj = new Event({ name: 'event object', __stateBeforeSuspension: { field1: 'test', jsonField: { name: 'test3' }} })
  eventObj.save(function (err, eventObj) {
    var newObject = eventObj.toObject()
    newObject.__stateBeforeSuspension = { field1: '12', jsonField: { name: 'test4' }}
    eventObj.set(newObject)
    var modifiedPaths = eventObj.modifiedPaths()
    console.log(modifiedPaths)
   
  })
}
run();
```
Output with v5.12.5 :
```
[
  '__stateBeforeSuspension',
  '__stateBeforeSuspension.field1',
  '__stateBeforeSuspension.jsonField',
  '__stateBeforeSuspension.jsonField.name'
]
```

Output with v6.12.1 :
```
[ '__stateBeforeSuspension', '__stateBeforeSuspension.jsonField' ]
```

The modified paths are returning differently in both versions.
Note: We have raised a different issue similar to .set() one. for more info please refer to https://github.com/Automattic/mongoose/issues/14022

### Steps to Reproduce

Added the full-fledged details on how to reproduce the issue in the description with a code snippet.

sakethyerra:
We observed another thing non-modified paths are showing as modified paths for more info please refer to the below code.

```
var mongoose = require('mongoose')
let Schema = mongoose.Schema
var eventSchema = new Schema({
  name: { type: String },
  __stateBeforeSuspension: {
    test: { type: String },
    test2: { type: String },
    test3: {
[...]

PR #14213, opened 2024-01-01
fix: add ignoreAtomics option to isModified() for better backwards compatibility with Mongoose 5

vkarpov15:
Re: #14024

**Summary**

Some of the work we did to make `isModified()` more consistent with what paths Mongoose validates and sends to MongoDB caused some trouble for users, so this PR adds a `ignoreAtomics` option to `isModified()` that can help smooth out the issues. Basically, with this `ignoreAtomics` option, `isModified()` will return `false` if passed an array path where the top-level array has a non-`$set` atomic operator. So if `post.comments` has a `$push` and no `$set`, then `post.isModified('comments.0.title', { ignoreAtomics: true })` will return `false`.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2223,8 +2223,9 @@ Document.prototype[documentModifiedPaths] = Document.prototype.modifiedPaths;
  * @api public
  */
 
-Document.prototype.isModified = function(paths, modifiedPaths) {
+Document.prototype.isModified = function(paths, options, modifiedPaths) {
   if (paths) {
+    const ignoreAtomics = options && options.ignoreAtomics;
     const directModifiedPathsObj = this.$__.activePaths.states.modify;
     if (directModifiedPathsObj == null) {
       return false;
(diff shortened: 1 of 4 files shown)


lib/document.js
```diff
@@ -2223,8 +2223,9 @@ Document.prototype[documentModifiedPaths] = Document.prototype.modifiedPaths;
  * @api public
  */
 
-Document.prototype.isModified = function(paths, modifiedPaths) {
+Document.prototype.isModified = function(paths, options, modifiedPaths) {
```
hasezoey, line 2226: the jsdoc needs to be updated to reflect the new parameter, also is `options` optional or required?

heyfirst:
Screenshot 2024-01-03 at 17 18 56

hey @vkarpov15, thanks for promptly raise the PR for providing backward-compatibility option! We tested in our use-case and it works! Looking forward for this to be in release,

Also, another way if you want to release this one as canary like `v6.12.4.14024` it's also works as well, we will able to install and try with our codebase 🙂

lib/document.js
vkarpov15: Optional. I updated jsdoc in 0960fae

vkarpov15:
@heyfirst we shipped 6.12.5 yesterday with the fix

PR #14226, opened 2024-01-02
fix(document): allow setting nested path to `null`

### LINKED ISSUE #14205: `Document.prototype.set` is not working properly when setting the Nested property to `null` with the `path` of object type
### Prerequisites

### Mongoose version

5.13.20

### Node.js version

18.8.0

### MongoDB server version

4.2

4.9.5

### Description

When I try to set the nested property to null with the object `path` argument,
the `nested` property is not updated to the value `null`.

But, setting the property to `null` with the form of `model.set(path, value)` or assigning directly `null` value to the property worked as I expected.

I think it is a bug, so I searched this repository for similar issues, but I couldn't find it as far as I searched.
So, I am reporting this issue like this.

### Steps to Reproduce

When the Model seems like:
```js
const schema = new Schema({
    nested: {
        key1: String,
        key2: String,
    },
});

const Model = mongoose.model('Model', schema);
```

Behavior different from what I expected:
```js
const func = async (id) => {
    const doc = await Model.findById(id);
    /**
     * doc => {
     *   nested: {
     *     key1: 'value1',
     *     key2: 'value2',
     *   }
     * }
     */
    doc.set({
        nested: null,
    });
    /**
     * doc => {
     *   nested: {
     *     key1: 'value1',
     *     key2: 'value2',
     *   }
     * }
     */
    await doc.save();
};

```

Behavior same as I expected:
```js
const func = async (id) => {
    const doc = await Model.findById(id);
    /**
     * doc => {
     *   nested: {
     *     key1: 'value1',
     *     key2: 'value2',
     *   }
     * }
     */
    doc.set('nested', null);
    // or doc.nested = null
    /**
     * doc => {
     *   nested: null
     * }
     */
    await doc.save();
};
```

### Expected Behavior

```js
const func = async (id) => {
    const doc = await Model.findById(id);
    /**
     * doc => {
     *   nested: {
     *     key1: 'value1',
     *     key2: 'value2',
     *   }
     * }
     */
    doc.set({
        nested: null,
    });
    /**
     * doc => {
     *   nested: null
     * }
     */
    await doc.save();
};
```

vkarpov15:
Fix #14205

**Summary**

Still debugging the need for the `valForKey === null` check, but otherwise this works. Without the `=== null` check, the `with conflicted data in db` test in `model.test.js` fails; it looks like because we don't store validation errors on nested paths. Will take another look tomorrow.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1153,6 +1153,8 @@ Document.prototype.$set = function $set(path, val, type, options) {
           } else {
             throw new StrictModeError(key);
           }
+        } else if (pathtype === 'nested' && valForKey === null) {
+          this.$set(pathName, valForKey, constructing, options);
         }
       } else if (valForKey !== void 0) {
         this.$set(pathName, valForKey, constructing, options);
--- a/lib/document.js
+++ b/lib/document.js
@@ -1153,7 +1153,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
           } else {
             throw new StrictModeError(key);
           }
-        } else if (pathtype === 'nested' && valForKey === null) {
+        } else if (pathtype === 'nested' && valForKey == null) {
           this.$set(pathName, valForKey, constructing, options);
         }
       } else if (valForKey !== void 0) {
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

vkarpov15:
I took a closer look and the `== null` check is necessary because of [this code](https://github.com/Automattic/mongoose/blob/037a82325ef82413830875ba9149c7c4921055d1/lib/document.js#L1176) later on in `$set` which handles setting a nested path to non-nullish.

PR #14228, opened 2024-01-02
types(model+query): use stricter typings for updateX(), replaceOne(),deleteX() Model functions

### LINKED ISSUE #14204: Type definition for `Query.updateOne` is wrong - options such as `new` are not supported by Mongo
### Prerequisites

### Mongoose version

8.1

### Node.js version

20.x

### MongoDB server version

7.0

5.x

### Description

The type definition of `Query.updateOne` is incorrect - it contains extra fields that are not used by Mongo, which can cause confusion for users. Specifically, `updateOne` in Mongo only supports (https://www.mongodb.com/docs/manual/reference/method/db.collection.updateOne/#syntax):
- `upsert`
- `writeConcern`
- `collation`
- `arrayFilters`
- `hint`
whereas the current definition in Mongoose uses `QueryOptions` (https://github.com/Automattic/mongoose/blob/8.1/types/query.d.ts#L730)
```
updateOne(
      filter?: FilterQuery<RawDocType>,
      update?: UpdateQuery<RawDocType> | UpdateWithAggregationPipeline,
      options?: QueryOptions<DocType> | null
    ): QueryWithHelpers<UpdateWriteOpResult, DocType, THelpers, RawDocType, 'updateOne'>;
```
where `QueryOptions` has many extra fields, including (https://github.com/Automattic/mongoose/blob/8.1/types/query.d.ts#L98):
- `new`
- `projection`
- `batchSize`
- `returnOriginal`
- `returnDocument`,
etc.

These extra fields cause confusion because they are not used by Mongo and therefore should be removed to provide accurate typing

### Steps to Reproduce

Type definitions do not need reproduction steps

### Expected Behavior

_No response_

vkarpov15:
Fix #14204

**Summary**

While Mongoose's current options typings are technically correct because of chaining, using stricter types in the Model functions that return a query lines up better with how developers use it. Plus with `[other: string]: any`, devs can still use other options.

I only made these changes for `deleteX()`, `replaceOne()`, and `updateX()` because Mongoose's `QueryOptions` has a `projection` property that doesn't line up exactly with the MongoDB Node driver's: the MongoDB Node driver's `projection` type is just `Record<string, any>`, doesn't take into account schema definitions.

**Examples**

### DIFF
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -222,7 +222,7 @@ declare module 'mongoose' {
     /** Creates a `countDocuments` query: counts the number of documents that match `filter`. */
     countDocuments(
       filter?: FilterQuery<TRawDocType>,
-      options?: (mongodb.CountOptions & MongooseSpecificQueryOptions) | null
+      options?: (mongodb.CountOptions & Omit<MongooseSpecificQueryOptions, 'lean' | 'timestamps'>) | null
     ): QueryWithHelpers<
       number,
       THydratedDocumentType,
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -222,7 +222,7 @@ declare module 'mongoose' {
     /** Creates a `countDocuments` query: counts the number of documents that match `filter`. */
     countDocuments(
       filter?: FilterQuery<TRawDocType>,
-      options?: QueryOptions<TRawDocType>
+      options?: (mongodb.CountOptions & MongooseSpecificQueryOptions) | null
     ): QueryWithHelpers<
       number,
       THydratedDocumentType,
@@ -254,7 +254,7 @@ declare module 'mongoose' {
      */
     deleteMany(
       filter?: FilterQuery<TRawDocType>,
-      options?: QueryOptions<TRawDocType>
+      options?: (mongodb.DeleteOptions & Omit<MongooseSpecificQueryOptions, 'lean' | 'timestamps'>) | null
     ): QueryWithHelpers<
       mongodb.DeleteResult,
       THydratedDocumentType,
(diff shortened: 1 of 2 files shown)


types/query.d.ts
```diff
@@ -95,33 +95,70 @@ declare module 'mongoose' {
...
+  interface MongooseSpecificQueryOptions {
+    /**
+     * If truthy, mongoose will return the document as a plain JavaScript object rather than a mongoose document.
+     */
+    lean?: boolean | Record<string, any>;
+
+    multipleCastError?: boolean;
+
+    overwriteDiscriminatorKey?: boolean;
+    runValidators?: boolean;
```
hasezoey, line 107:
```suggestion
    /**
     * Set to true to enable `update validators`
     * (https://mongoosejs.com/docs/validation.html#update-validators). Defaults to false.
     */
    runValidators?: boolean;
```

missing tsdoc that was previously there?

types/query.d.ts
```diff
@@ -136,40 +173,11 @@ declare module 'mongoose' {
      * Another alias for the `new` option. `returnOriginal` is deprecated so this should be used.
      */
     returnDocument?: 'before' | 'after';
-    /**
-     * Set to true to enable `update validators`
-     * (https://mongoosejs.com/docs/validation.html#update-validators). Defaults to false.
-     */
-    runValidators?: boolean;
-    /* Set to `true` to automatically sanitize potentially unsafe user-generated query projections */
-    sanitizeProjection?: boolean;
```
hasezoey, line 145: this options was removed without being added to `MongooseSpecificQueryOptions`, is this intended?

types/models.d.ts
```diff
@@ -222,7 +222,7 @@ declare module 'mongoose' {
     /** Creates a `countDocuments` query: counts the number of documents that match `filter`. */
     countDocuments(
       filter?: FilterQuery<TRawDocType>,
-      options?: QueryOptions<TRawDocType>
+      options?: (mongodb.CountOptions & MongooseSpecificQueryOptions) | null
```
hasezoey, line 225: if i understand this function correctly, then `lean` (and likely `timestamps`) are not options for this right? (maybe also other options)

hasezoey (approved):

PR #14235, opened 2024-01-05
types(schema): add missing `omit()` method

amitbeck:
**Summary**

Adding the missing `omit()` method to the `Schema` class type definition

**Examples**


vkarpov15 (approved):
Thanks :+1:

PR #14247, opened 2024-01-10
Adding missing pluralizations, fixing pluralization: virus -> viruses

ItsBradyDavis:
**Summary**
Adds pluralization for woman, cactus, focus, fungus, nucleus, and fixes the pluralization of virus from `viri` to [`viruses`](https://simple.wiktionary.org/wiki/viruses)

**Examples**
woman -> women
virus -> viruses
cactus -> cacti
focus -> foci
fungus -> fungi
nucleus -> nuclei

### DIFF
--- a/lib/helpers/pluralize.js
+++ b/lib/helpers/pluralize.js
@@ -8,13 +8,13 @@ module.exports = pluralize;
 
 exports.pluralization = [
   [/human$/gi, 'humans'],
-  [/(m)an$/gi, '$1en'],
+  [/(m|wom)an$/gi, '$1en'],
   [/(pe)rson$/gi, '$1ople'],
   [/(child)$/gi, '$1ren'],
   [/^(ox)$/gi, '$1en'],
   [/(ax|test)is$/gi, '$1es'],
-  [/(octop|vir)us$/gi, '$1i'],
-  [/(alias|status)$/gi, '$1es'],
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez (approved):
LGTM, only concern would be that this is a breaking change that we'd have to ship in v9.

vkarpov15 (approved):
LGTM, but I'm going to have to merge this into 9.0 branch because this is backwards breaking.

markstos:
I disagree with most of these, because they are acceptable alternatives listed in official dictionaries as such. Given that, in the spirit of backwards compatible, I recommend keeping "cactuses", "focuses", "funguses" and "nucleuses"

Ref:
  * https://www.merriam-webster.com/dictionary/nucleus
  * https://www.merriam-webster.com/dictionary/fungus
  * https://www.merriam-webster.com/dictionary/cactus
  * https://www.merriam-webster.com/dictionary/focus

I may be a little salty because I'm many, many hours into a Mongoose 5➡️6 upgrade that's so far about 60 commits and 2,000 lines deep. It's a big, old code base that has triggered a number of the documented breaking changes as well as least a couple undocumented breaking changes.

I think Mongoose has made a lot of good calls with the breaking changes it has made, but unnecessary changes can become paper cuts, or worse, new bugs in production.

So while, yes, the recommended plurals are the most correct, the ones in place now are recognized as valid English and may save some cactus farmers some upgrade headache.

PR #14257, opened 2024-01-13
feat(connection): add listCollections() helper to connections

vkarpov15:
**Summary**

We have `dropCollection()` and `dropDatabase()` helpers on connections, but not `listCollections()`, so there's no way to list all connections in MongoDB without bypassing Mongoose. This PR adds a new `Connection.prototype.listCollections()` helper that wraps MongoDB's `listCollections()`, including handling the fact that `listCollections()` returns a cursor.

**Examples**

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -615,10 +615,6 @@ Connection.prototype.dropCollection = async function dropCollection(collection) 
  */
 
 Connection.prototype.listCollections = async function listCollections() {
-  if (arguments.length >= 2 && typeof arguments[1] === 'function') {
-    throw new MongooseError('Connection.prototype.dropCollection() no longer accepts a callback');
-  }
-
   if ((this.readyState === STATES.connecting || this.readyState === STATES.disconnected) && this._shouldBufferCommands()) {
     await new Promise(resolve => {
       this._queue.push({ fn: resolve });
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -606,6 +606,30 @@ Connection.prototype.dropCollection = async function dropCollection(collection) 
 };
 
 /**
+ * Helper for MongoDB Node driver's `listCollections()`.
+ * Returns an array of collection objects.
+ *
+ * @method listCollections
+ * @return {Promise<Collection[]>}
+ * @api public
+ */
+
+Connection.prototype.listCollections = async function listCollections() {
+  if (arguments.length >= 2 && typeof arguments[1] === 'function') {
+    throw new MongooseError('Connection.prototype.dropCollection() no longer accepts a callback');
+  }
+
+  if ((this.readyState === STATES.connecting || this.readyState === STATES.disconnected) && this._shouldBufferCommands()) {
+    await new Promise(resolve => {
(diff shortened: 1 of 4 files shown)


lib/connection.js
```diff
@@ -605,6 +605,30 @@ Connection.prototype.dropCollection = async function dropCollection(collection)
...
+ * Helper for MongoDB Node driver's `listCollections()`.
+ * Returns an array of collection objects.
+ *
+ * @method listCollections
+ * @return {Promise<Collection[]>}
+ * @api public
+ */
+
+Connection.prototype.listCollections = async function listCollections() {
+  if (arguments.length >= 2 && typeof arguments[1] === 'function') {
```
AbdelrahmanHafez, line 618: Nitpick: since we're doing `>=`, I think it's more accurate to check for the last argument instead of argument at position `1`.
```suggestion
  if (arguments.length >= 2 && typeof arguments[arguments.length - 1] === 'function') {
```

types/connection.d.ts
```diff
@@ -121,6 +121,12 @@ declare module 'mongoose' {
      */
     readonly id: number;
 
+    /**
+     * Helper for MongoDB Node driver's `listCollections()`.
+     * Returns an array of collection names.
+     */
+    listCollections(collection: string): Promise<Pick<mongodb.CollectionInfo, 'name' | 'type'>[]>;
```
AbdelrahmanHafez, line 128: I'm a bit confused, why does `listCollections` take a `collection` parameter? I can see that the implementation in `lib/connection.js` is not defining or passing any parameters to the driver's `listCollections` method.

AbdelrahmanHafez: Also, maybe a TS test for this would be nice.

AbdelrahmanHafez (commented):
A couple of minor suggestions and a question, otherwise LGTM. 👍

test/connection.test.js
```diff
@@ -195,6 +195,14 @@ describe('connections:', function() {
...
       assert.ok(newCollectionsNames.indexOf('gh5712') !== -1);
     });
+
+    it('listCollections()', async function() {
+      await conn.dropDatabase();
+      await conn.createCollection('test1176');
+
+      const collections = await conn.listCollections();
+      assert.deepStrictEqual(collections.map(coll => coll.name), ['test1176']);
+    });
```
hasezoey, line 205: maybe consider adding another collection, to verify it does more than just one (because of cursor)

types/connection.d.ts
hasezoey:
```suggestion
    listCollections(): Promise<Pick<mongodb.CollectionInfo, 'name' | 'type'>[]>;
```

the jsdoc does not mention any parameter and the js also does not have any parameter, so i also think this is not correct
(likely also a copy-paste from `dropCollection`?)

vkarpov15:
Fixed all comments. You're right, I copy/pasted `dropCollection()` and missed a few spots.

hasezoey (approved):

AbdelrahmanHafez (approved):

PR #14260, opened 2024-01-15
8.1

vkarpov15:
**Summary**

**Examples**


hasezoey (approved):
Change look good to me, i have also tested with typegoose where in runtime it is fine, but somehow for types it broke the else path of `isDocument`, likely related to the typehint pr (but fixed when using typescript 5.3).

example (pseudo)
```ts
class AutoObjectId {
  public _id!: typegoose.mongoose.Types.ObjectId;
}

class AutoRef {
  @typegoose.prop()
  public objectid: typegoose.Ref<AutoObjectId>;
}

const AutoRefModel = typegoose.getModelForClass(AutoRef);
const doc = new AutoRefModel({});

if (isDocument(doc.objectid)) {
  // type is as said in both versions
  expectType<typegoose.DocumentType<AutoObjectId>>(doc.objectid);
} else {
  // in mongoose 8.0 this type is correct
  // but in mongoose 8.1 the type is "typegoose.types.Ref<AutoObjectId>", but i do not understand what changed for this to happen, maybe too-complex type?
  expectType<typegoose.mongoose.Types.ObjectId>(doc.objectid);
}
```

typescript 5.2, but works again with typescript 5.3

PR #14265, opened 2024-01-18
 null check

rohanricky:
@vkarpov15 added this null check due to the error I'm facing ( refer below screenshot ). hoping this can be merged, if this is a straight forward use-case. If not,  I will work on a reproducible when I find time.

image


vkarpov15:
I'll merge, but if you can come up with a erepro script that would be very helpful. I wouldn't be surprised if you got a different error once this PR is merged.

rohanricky:
@vkarpov15 when is 6.12.6 release planned? Is there any chance 6.12.6 can be released as soon as possible with currently closed issues and move the rest of the pending items to 6.12.7? We are running short of time for our release cycle due to this issue.

vkarpov15:
@rohanricky we will ship 6.12.6 early next week

PR #14287, opened 2024-01-24
fix(document): handle setting nested path to spread doc with extra properties

### LINKED ISSUE #14269: Cannot patch a sub-document with destructuring
### Prerequisites

### Mongoose version

8.1.0

### Node.js version

20.11.0

### MongoDB server version

7.x

5.3.3

### Description

When patching a sub-document with destructuring, the changes are not saved correctly.

```
const person = await personModel.findOne({...});

person.address = {
      ...person.address,
      zip: 54321,
    };

    await person.save();
```

However, after saving (and even before the save is called) the person.address.zip is still the old value.

### Steps to Reproduce

Cone this repo:  https://github.com/a88zach/mongoose-bug

run `$ yarn test`

Notice the 1 test that is successful and the 1 test that fails.  The one that fails is using destructuring to patch the sub document

### Expected Behavior

The sub-document should be updated with the existing information plus any overrides based on the patch

vkarpov15:
Fix #14269

**Summary**

Quick improvement for handling spread docs in `$set()`. `true` means we retain existing properties, so `{ ...doc, foo: true }` will still have `foo: true`.

Worth noting that we don't currently have a good way of handling key order - I'm looking into that. But `{ foo: true, ...doc }` currently will still have `foo: true`, even if `doc` has `foo: false`.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1162,7 +1162,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
 
   // Assume this is a Mongoose document that was copied into a POJO using
   // `Object.assign()` or `{...doc}`
-  val = handleSpreadDoc(val);
+  val = handleSpreadDoc(val, true);
 
   // if this doc is being constructed we should not trigger getters
   const priorVal = (() => {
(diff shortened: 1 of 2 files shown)


IslandRhythms (approved):
LGTM

PR #14299, opened 2024-01-26
fix(QueryCursor): eachAsync parameter of AggregationCursor and QueryCursor

DevooKim:
**Summary**

Currently, [documentation](https://mongoosejs.com/docs/api/querycursor.html#QueryCursor.prototype.eachAsync()) for QueryCursor.prototype.eachAsync write that it takes a callback as a parameter.

However, with the update to version 7, eachAsync no longer supports callbacks. - [Show commits](https://github.com/Automattic/mongoose/commit/b911bef9e41ed622ec3961b03d130edc5c2e4c09#diff-b9b6c3043d3ed32b9229dc3abc6d06af6f84d23106dc129cdefe46db4f89de54)
So, I fix parameters of eachAsync and the migrating guide to 7.

it has two more changes.
1. Add `AggregationCursor.prototype.next`, which was missing the guide.
2. fix conditional statements conditional in `QueryCursor.prototype.next`

### DIFF
--- a/docs/migrating_to_7.md
+++ b/docs/migrating_to_7.md
@@ -87,6 +87,7 @@ They always return promises.
 * `Aggregate.prototype.exec`
 * `Aggregate.prototype.explain`
 * `AggregationCursor.prototype.close`
+* `AggregationCursor.prototype.eachAsync`
 * `Connection.prototype.startSession`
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #14314, opened 2024-02-01
fix(populate): call setter on virtual populated path with populated doc instead of undefined

### LINKED ISSUE #14285: virtual setter called with value undefined on populate
### Prerequisites

### Mongoose version

8.1.0

### Node.js version

18.18.0

### MongoDB server version

7.0.5

_No response_

### Description

Mongoose populates the virtual internally (`set()` is never called with the populated value), but additionally calls its setter function with the value `undefined`. This can lead to bad, unexpected side effects.

I'm not sure whether that is expected behavior, but if so, please advise on how we should implement the below use case.

### Steps to Reproduce

Example script which outputs a person with `userId: null` and `userType: null`.

```js
const uri = "mongodb://mongo1:27017/test";

async function run() {
    try {
        await mongoose.connect(uri);
        console.log("connected");

        const ObjectId = Schema.ObjectId;

            email: String,
            name: "String",
        });

        const user = await User.updateOne({
            email: "admin@example.com",
        }, {
            email: "admin@example.com",
            name: "Admin",
        }, {
            upsert: true,
        });

        const personSchema = new Schema({
            userId: ObjectId,
            userType: String,
        });

        personSchema.virtual('user', {
            ref() {
                return this.userType;
            },
            localField: "userId",
            foreignField: "_id",
            justOne: true,
        })
        .set(function (user) {
            console.log("set user:", user);
            if (user) {
                this.userId = user._id;
                this.userType = toModelName(user);
            } else {
                this.userId = null;
                this.userType = null;
            }
        });

        personSchema.pre("findOne", function () {
            this.populate("user");
        });

        const fetchedUser = await User.findOne({ email: "admin@example.com" })

        const person = new Person({
            userId: fetchedUser._id,
            userType: "User",
        });

        await person.save();

        const fetchedPerson = await Person.findOne({ _id: person._id });

        console.log("person fetched:", fetchedPerson.toJSON({ virtuals: true }));
    } catch (e) {
        console.error(e);
    } finally {
        await mongoose.disconnect();
    }
}

run().catch(console.dir).then(() => console.log("Done"));
```

### Expected Behavior

The populated field (i.e. `person`) to be populated and the setter not to be called or called with the value that was populated.

vkarpov15:
Fix #14285

**Summary**

Looks like the virtual populate setter doesn't correctly return the populated doc, this PR fixes that. Also did a quick bit of refactor to make it so that we have less populate logic in `lib/schema.js`.

**Examples**

### DIFF
--- a/lib/helpers/populate/setPopulatedVirtualValue.js
+++ b/lib/helpers/populate/setPopulatedVirtualValue.js
@@ -1,0 +1,23 @@
+'use strict';
+
+module.exports = function setPopulatedVirtualValue(populatedVirtuals, name, v, options) {
+  if (options.justOne || options.count) {
+    populatedVirtuals[name] = Array.isArray(v) ?
+      v[0] :
+      v;
+
+    if (typeof populatedVirtuals[name] !== 'object') {
+      populatedVirtuals[name] = options.count ? v : null;
+    }
+  } else {
+    populatedVirtuals[name] = Array.isArray(v) ?
+      v :
+      v == null ? [] : [v];
+
+    populatedVirtuals[name] = populatedVirtuals[name].filter(function(doc) {
+      return doc && typeof doc === 'object';
+    });
+  }
+
+  return populatedVirtuals[name];
+};
--- a/lib/helpers/populate/setPopulatedVirtualValue.js
+++ b/lib/helpers/populate/setPopulatedVirtualValue.js
@@ -1,5 +1,15 @@
 'use strict';
 
+/**
+ * Set a populated virtual value on a document's `$$populatedVirtuals` value
+ *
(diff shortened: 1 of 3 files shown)


lib/helpers/populate/setPopulatedVirtualValue.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+module.exports = function setPopulatedVirtualValue(populatedVirtuals, name, v, options) {
```
hasezoey, line 3: maybe consider adding some jsdoc (at least types and descriptions of the parameters)

hasezoey (approved):
Looks good to me, aside from some jsdoc

hasezoey (approved):

PR #14336, opened 2024-02-06
types(model): correct return type for findOneAndUpdate with `includeResultMetadata` and `lean` set

### LINKED ISSUE #14303: includeResultMetadata types not changing return type?
### Prerequisites

### Mongoose version

8.1.1

### Node.js version

20.x

### MongoDB server version

6.x

5.1.6

### Description

I tried upgrading to mongoose 8.x and changed rawResult: true to includeResultMetadata: true, but now typescript complains:
```
TS2339: Property value does not exist on type
FlattenMaps<IDBJobField> & Required<{   _id: ObjectId; }>

```

also it seems that the retutrn type is not changed depending on the value of includeResultMetadata
image
it's the same as in the screenshot, regardless of this flag.

Furthermore; i found it kinda confusing with the mongodb docs, as they say that includeResultMetadata is true by default. Does mongoose set it to false then by default? Or is it actually true, but mongoose just returns the document and strips the rest away? Which would sound to me of a waste of some extra bytes?

error TS2339: Property 'value' does not exist on type 'FlattenMaps<IDBJobField> & Required<{ _id: ObjectId; }>'.
error TS2339: Property 'lastErrorObject' does not exist on type 'FlattenMaps<IDBJobField> & Required<{ _id: ObjectId; }>'.
error TS2339: Property 'value' does not exist on type 'FlattenMaps<IDBJobField> & Required<{ _id: ObjectId; }>'.

### Steps to Reproduce

use an example query for findOneAndupdate and play around with the includeResultMetadata flag

### Expected Behavior

should return the "ModifyResult" instead, like it did with rawResult: true

vkarpov15:
Fix #14303

**Summary**

Quick fix for correct return type when calling `findOneAndUpdate()` with `includeResultMetadata: true` and `lean: true`: return a `ModifyResult` with the lean doc type.

**Examples**

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -17,7 +17,7 @@ import mongoose, {
 } from 'mongoose';
 import { expectAssignable, expectError, expectType } from 'tsd';
 import { AutoTypedSchemaType, autoTypedSchema } from './schema.test';
-import { UpdateOneModel, ChangeStreamInsertDocument, ObjectId } from 'mongodb';
+import { ModifyResult, UpdateOneModel, ChangeStreamInsertDocument, ObjectId } from 'mongodb';
 
 function rawDocSyntax(): void {
   interface ITest {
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
Looks good to me

PR #14350, opened 2024-02-13
refactor: make query helpers more consistently fit our loose definition of helpers

vkarpov15:
**Summary**

This PR is more of an architecture discussion. In my mind, a "helper" ideally should be a function that takes as argument only non-Mongoose objects, with the potential exception of schemas and SchemaTypes. So no models or queries or constructors or anything like that, _unless_ they are only used as context for some callback function like in `castFilterPath`.

Why is this important? Makes helpers much easier to test, reason about, and reuse. Once you have to instantiate a model and a connection, that makes testing slower.

In that vein, in this PR I'm moving a couple of functions in `helpers/query` out into `query.js`, refactored a couple of helpers to use just POJO options objects rather than a full model, and moved some constants into a `constants.js` file (helpers should always be functions).

WDYT @AbdelrahmanHafez @hasezoey @IslandRhythms ?

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -2482,7 +2482,6 @@ Query.prototype._completeOne = function(doc, res, callback) {
  * @param {Object} [opts]
  * @param {Array} [opts.populated]
  * @param {ClientSession} [opts.session]
- * @param {Function} callback
  * @api private
  */
 


lib/query.js
```diff
@@ -2473,6 +2472,38 @@ Query.prototype._completeOne = function(doc, res, callback) {
   });
 };
 
+/**
+ * Given a model and an array of docs, hydrates all the docs to be instances
+ * of the model. Used to initialize docs returned from the db from `find()`
+ *
+ * @param {Model} model
```
hasezoey, line 2479:
```suggestion
```

there is now no such parameter

lib/query.js
```diff
@@ -2473,6 +2472,38 @@ Query.prototype._completeOne = function(doc, res, callback) {
...
+ * @param {Array} docs
+ * @param {Object} fields the projection used, including `select` from schemas
+ * @param {Object} userProvidedFields the user-specified projection
+ * @param {Object} [opts]
+ * @param {Array} [opts.populated]
+ * @param {ClientSession} [opts.session]
+ * @param {Function} callback
```
hasezoey, line 2486:
```suggestion
```

no such parameter

hasezoey (approved):
Looks good, some small documentation issues though

AbdelrahmanHafez (approved):
LGTM, definitely feels lighter and helps with re-usability.

hasezoey (approved):

PR #14352, opened 2024-02-13
Island rhythms/deeply hydrate doc

### LINKED ISSUE #4727: Hydrate: restore populated data
Currently when you use either `new Model(data)` or `Model.hydrate(data)` and data contains pre-populated stuff, the model instance won't have these populated fields.

Would it be possible to restore a model from a plain object while restoring populated data as well?

IslandRhythms:
closes #4727

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -767,15 +767,14 @@ function init(self, obj, doc, opts, prefix) {
       }
     } else {
       // Retain order when overwriting defaults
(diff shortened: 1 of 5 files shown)


vkarpov15 (commented):
Code LGTM, but please fix up the TypeScript types as well.

vkarpov15 (approved):

PR #14358, opened 2024-02-19
feat: add middleware for bulkWrite() and createCollection()

### LINKED ISSUE #7893: Hook to collection creation
Feature
I want to execute some custom commands on collection creation.
Is there a way to hook to or to override the automatic collection creation by mongoose ?

### LINKED ISSUE #14263: Hook into bulkWrite
### Prerequisites

### 🚀 Feature Proposal

Mongoose already [supports middlewares](https://mongoosejs.com/docs/middleware.html) for the majority of operations. But by reviewing the documentation and code I couldn't find a way to hook into `bulkWrite`.

Which I found surprising, as `castBulkWrite` is called on each `bulkWrite` operation giving at least a perfect entry point for 
`pre` modifying `bulkWrite` operations before they are executed:

https://github.com/Automattic/mongoose/blob/ebf7d07692b5b826267a46b7a6f886ac7c196c2a/lib/model.js#L3424

This allows mongoose internally for example to add the discriminator:

https://github.com/Automattic/mongoose/blob/master/lib/helpers/model/castBulkWrite.js#L70

Or `updatedAt` and `createdAt` timestamps:

https://github.com/Automattic/mongoose/blob/master/lib/helpers/model/castBulkWrite.js#L75

### Motivation

I am already using [mongoose discriminators](https://mongoosejs.com/docs/discriminators.html) to [clearly seperate tenants within a single database](https://www.nmajor.com/posts/2018-09-28-multi-tenancy-with-expressmongoose).

Mongoose only allows setting a single `discriminatorKey` per model, so if I would like to "discriminate" additionally on another key I can't use discriminators for it.

Now I want to shard data based on geolocation for which each query needs to include both the `discriminatorKey` and the `location`. This is fairly easy to accomplish via hooks:

```js
// ... higher order method that calls each model with the location

// Modify mongoose queries to include location
Model.schema.pre( /find|update/, function () {

    this.setQuery( { ...this.getQuery(), location } );

} );

// Add a $match location stage to the beginning of each aggregation pipeline
Model.schema.pre( 'aggregate', function() {

    this.pipeline().unshift( { $match: { location } } );

} );
```

But this doesn't touch `bulkWrite` operations. Here I would have to remember to include the `location` within each `filter` key. I would prefer if I could just hook into the `castBulkWrite` method instead similar to how mongoose is doing it internally with discriminators.

vkarpov15:
Fix #7893
Fix #14263

**Summary**

Add support for `schema.pre('bulkWrite', function(next, ops) { ... })` and `schema.pre('createCollection', function() { ... })`. Support for 2 new model middleware functions :+1:

**Examples**

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -67,13 +67,17 @@ Model middleware is supported for the following model functions.
 Don't confuse model middleware and document middleware: model middleware hooks into *static* functions on a `Model` class, document middleware hooks into *methods* on a `Model` class.
 In model middleware functions, `this` refers to the model.
 
+* [bulkWrite](api/model.html#model_Model-bulkWrite)
+* [createCollection](api/model.html#model_Model-createCollection)
 * [insertMany](api/model.html#model_Model-insertMany)
 
 Here are the possible strings that can be passed to `pre()`
 
 * aggregate
+* bulkWrite
 * count
 * countDocuments
+* createCollection
 * deleteOne
 * deleteMany
 * estimatedDocumentCount
--- a/lib/model.js
+++ b/lib/model.js
@@ -1415,6 +1415,18 @@ Model.createCollection = async function createCollection(options) {
     throw new MongooseError('Model.createCollection() no longer accepts a callback');
   }
 
+  const shouldSkip = await new Promise((resolve, reject) => {
+    this.hooks.execPre('createCollection', this, [options], (err) => {
+      if (err != null) {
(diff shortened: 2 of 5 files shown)


hasezoey (commented):
Looks good to me

though maybe consider doing this for a minor version instead of patch

vkarpov15:
Yeah this is going into 8.2.

PR #14363, opened 2024-02-20
fix: strict check "ok" on validate function

encloinc:
**Summary**

The problem is that when val is a string and its length is 0 - 0 will be a "falsey" value and the function will error. We should strictly check if the validate function explictly made "ok" false to prevent this behavior, since empty string should be allowed even if the required property is true. 

**Examples**

SCHEMA DEF && CODE (was initially failing on contextResponse being an empty string - when I expect it to fail when its undefined):

```
const messageSchema = new Schema({
	chat_id: {
        type: Schema.Types.ObjectId, 
        ref: 'Chat',
        required: true
	},

	role: {
		type: String,
		required: true
	},

	content: {
		type: String,
		required: true
	},

       embeds :{
                type: [Object],
                required: true
 
       },

    contextResponse: {
        type: String,
        required: true
    }
    

}, {timestamps: true});

  await models.Message.create({
        chat_id: '65bcedc1a78c24f230b4be7d',
        role: 'user',
        content: 'message',
        embeds: [],
        contextResponse: ''
    })
```

### DIFF
--- a/lib/schemaType.js
+++ b/lib/schemaType.js
@@ -1356,7 +1356,7 @@ SchemaType.prototype.doValidate = function(value, fn, scope, options) {
     if (err) {
       return;
     }
-    if (ok === undefined || (ok && ok !== 0)) {
+    if (ok === undefined || ok || ok === 0) {
       if (--count <= 0) {
         immediate(function() {
           fn(null);
--- a/lib/schemaType.js
+++ b/lib/schemaType.js
@@ -1356,7 +1356,7 @@ SchemaType.prototype.doValidate = function(value, fn, scope, options) {
     if (err) {
       return;
     }
-    if (ok === undefined || ok !== false) {
+    if (ok === undefined || (ok && ok !== '')) {
       if (--count <= 0) {
         immediate(function() {
           fn(null);
--- a/lib/schemaType.js
+++ b/lib/schemaType.js
@@ -1356,7 +1356,7 @@ SchemaType.prototype.doValidate = function(value, fn, scope, options) {
     if (err) {
       return;
     }
-    if (ok === undefined || (ok && ok !== '')) {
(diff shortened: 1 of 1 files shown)


lib/schemaType.js
```diff
@@ -1356,7 +1356,7 @@ SchemaType.prototype.doValidate = function(value, fn, scope, options) {
     if (err) {
       return;
     }
-    if (ok === undefined || ok) {
+    if (ok === undefined || ok !== false) {
```
vkarpov15, line 1359: Maybe instead we add an explicit check for `|| (ok && ok !== '')` ? As written, this change is pretty risky and causes a lot of test failures, partially due to breaking several of our built-in validators. For example, [string `match` validator relies on returning `null` meaning "ok"](https://github.com/Automattic/mongoose/blob/f1ed13f5b4bacbc739ae4c52a19dfcd1e355555a/lib/schema/string.js#L542-L544).

In general, I agree that treating `return ''` as an error but not `return 'test'` is not ideal. But I think this change as written is a bit too risky.

vkarpov15 (changes requested):
Made some comments, I'll try making those changes myself to see how they affect test output.

encloinc (commented):
This will fail, as ok is an integer that is the length of the string. 

The falsey value is not "", its 0.  Since ok will never be "" strictly, this implementation is the same as the one that there is currently, just with more code haha.

vkarpov15:
Oh I see, the issue you're trying to fix is that empty string `''` doesn't pass `required` check. That is expected behavior. But there is a [one-line workaround if you want Mongoose to allow empty strings to pass `required` check](https://mongoosejs.com/docs/api/schemastring.html#SchemaString.checkRequired()):

```javascript
// Allow empty strings to pass `required` check
mongoose.Schema.Types.String.checkRequired(v => v != null);
```

PR #14375, opened 2024-02-24
feat(query): add `options` parameter to `Query.prototype.sort()`

### LINKED ISSUE #14365: query.sort(null) should reset the sort options
### 🚀 Feature Proposal

Calling `.sort(null)` should allow to unset whatever was set as a sort previously (in `query.options.sort`)

### Motivation

Sometimes it's useful to override default behaviors from plugins, later in the query chain

Similar to https://github.com/Automattic/mongoose/issues/14180

### Example

```
q=I.find().sort({ startDate: -1, _id: -1 });

q.options.sort // { startDate: -1, _id: -1 }

q.sort(null); // reset sort (doesn't work currently)
q.sort({ _id: -1 }); // this doesn't overwrite a previous non-null sort

q.options.sort // now is { _id: -1 }, but in current version it still returns { startDate: -1, _id: -1 }
```

I'm proposing `.sort(null)`, but I've tried using `.setOptions(null)` or `.setOptions({})` or `.setOptions({sort: {_id: -1}})` without success. Any way would be fine

vkarpov15:
Fix #14365

**Summary**

Right now, `sort(null)` is a no-op. There's no good way to unset the current sort option other than `delete query.getOptions().sort`, which works but looks a bit janky. This PR adds `q.sort(null, { override: true })`, the `override` option lets you explicitly override the current sort.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -2861,17 +2861,25 @@ Query.prototype.distinct = function(field, conditions) {
  * Cannot be used with `distinct()`
  *
  * @param {Object|String|Array<Array<(string | number)>>} arg
+ * @param {Object} [options]
+ * @param {Boolean} [options.override=false] If true, replace existing sort options with `arg`
  * @return {Query} this
  * @see cursor.sort https://www.mongodb.com/docs/manual/reference/method/cursor.sort/
  * @api public
  */
 
-Query.prototype.sort = function(arg) {
-  if (arguments.length > 1) {
-    throw new Error('sort() only takes 1 Argument');
+Query.prototype.sort = function(arg, options) {
+  if (arguments.length > 2) {
+    throw new Error('sort() takes at most 2 arguments');
+  }
+  if (options != null && typeof options !== 'object') {
+    throw new Error('sort() options argument must be an object or nullish');
   }
 
   if (this.options.sort == null) {
+    this.options.sort = {};
+  }
+  if (options && options.override) {
     this.options.sort = {};
   }
   const sort = this.options.sort;
--- a/test/query.test.js
+++ b/test/query.test.js
@@ -795,7 +795,7 @@ describe('Query', function() {
         e = err;
       }
       assert.ok(e, 'uh oh. no error was thrown');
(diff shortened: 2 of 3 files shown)


FaizBShah:
This looks good to me. Can we also add the documentation for the 2nd parameter `override` in `setOptions()` function in this PR? (Or is that out of scope of this PR?)

lib/query.js
```diff
@@ -2861,19 +2861,27 @@ Query.prototype.distinct = function(field, conditions) {
...
+    throw new Error('sort() takes at most 2 arguments');
+  }
+  if (options != null && typeof options !== 'object') {
+    throw new Error('sort() options argument must be an object or nullish');
   }
 
   if (this.options.sort == null) {
     this.options.sort = {};
   }
+  if (options && options.override) {
```
FaizBShah, line 2882: In JS, somebody can pass `override: 1` or `override: '  '`, and even then this condition will pass. I think we should change the condition to `options && options.override === true`

hasezoey (approved):
Looks good to me. 
As for the "condition boolean checks instead of truthy" issue, i am impartial, as i think both methods are using in mongoose and not anything specific is set to be used.

lib/query.js
caub: agreed

Also options.override is clearer at least than a boolean arg like in [`setOptions`](https://github.com/Automattic/mongoose/blob/master/lib/query.js#L1570)

Should we consider also adding this options.override for `.select()`? (I know there is `.projection()`, but it could simplify things to have just one method in the future, and maybe keep the other one as an alias, so same behavior)

vkarpov15: We prefer to treat `override: 1` or `override: '  '` as equivalent to `override: true`. There's 2 alternatives, either treat `override: 1` as `override: false` or throw an error; I don't think either is better in this case.

Re: adding `options.override` to `select()`, I think that's a reasonable idea to consider, but outside the scope of this issue. If this API works well, we'll consider adding it to `select()` and other query methods where it makes sense.

PR #8507, opened 2020-01-14
Support `lean` option for `Model.insertMany()` #8234

ntsekouras:
**Summary**

This PR fixes issue #8234.

You can now pass a `lean` option to `Model.insertMany` to bypass validation.

**Examples**

```javascript

MyModel.insertMany(arr, { lean: true }, (err, docs) => {});

```

PR #14376, opened 2024-02-25
perf(model): make `insertMany()` `lean` option skip hydrating Mongoose docs

### LINKED ISSUE #14372: Model.insertMany with `lean: true` is very slow compared to native driver
### Prerequisites

### Mongoose version

6.6.1

### Node.js version

16.x

### MongoDB version

6.0.1

### Operating system

macOS

_No response_

### Issue

Hey 👋 

I am currently working on a piece of functionality that requires bulk inserting of documents using Mongoose. In the process, I've noticed a considerable performance difference between `Model.collection.insertMany(docs)` method and the `Model.insertMany(docs, {lean: true})` method with the `{lean: true}` flag. Based on the documentation, I understand that the `{lean: true}` option helps to skip the process of hydrating and validating the documents before insertion. I had thought that this would effectively make these two methods roughly equivalent in terms of performance.

However, upon performing benchmark tests, I found that `Model.collection.insertMany` is significantly faster than `Model.insertMany({lean: true})`. I am hoping to better understand why this is the case. I am also attaching a script to replicate the issue.

```nodejs

const MyModelSchema = new Schema({
    field1: String,
    field2: Number,
    field3: Boolean,
    field4: Date,
    field5: Array
});

const MyModel = mongoose.model('MyModel', MyModelSchema);

async function clearCollection() {
    await MyModel.deleteMany({});
}
async function mongooseInsert(docs) {
    console.time("Mongoose Insert");
    await MyModel.insertMany(docs, { ordered: false, lean: true });
    console.timeEnd("Mongoose Insert");
}

async function nativeInsert(docs) {
    console.time("Native Insert");
    await MyModel.collection.insertMany(docs, { ordered: false });
    console.timeEnd("Native Insert");
}

async function chunkedNativeInsert(docs) {
    console.time("Chunked Native Insert");
    for (let i = 0; i < docs.length; i += 1000) {
        await MyModel.collection.insertMany(docs.slice(i, i + 1000), { ordered: false });
    }
    console.timeEnd("Chunked Native Insert");

}
async function main() {
    await mongoose.connect("mongodb://localhost:27017/test");
    let documents = [];

    for (let i = 0; i < 50000; i++) {
        documents.push({
            field1: `String${i}`,
            field2: i,
            field3: i % 2 === 0,
            field4: new Date(),
            field5: [i, i+1, i+2]
        });
    }
    await mongooseInsert(documents);
    await clearCollection();
    await chunkedNativeInsert(documents);
    await clearCollection();
    await nativeInsert(documents);
    await clearCollection();
}


main();
```

Thanks!

vkarpov15:
Fix #14372

**Summary**

Initial implementation of `lean` for `insertMany` in #8507 made `lean` just skip validation, not hydrating docs, which makes `insertMany` with `lean` still significantly slower. I'm inclined to change it so that `insertMany` with `lean` also skips hydrating Mongoose docs, because the name `lean` implies that we skip all hydration and other Mongoose-specific stuff similar to [query lean](https://mongoosejs.com/docs/tutorials/lean.html).

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3430,6 +3430,13 @@ Model.$__insertMany = function(arr, options, callback) {
   const results = ordered ? null : new Array(arr.length);
   const toExecute = arr.map((doc, index) =>
     callback => {
+      // If option `lean` is set to true bypass validation and hydration
+      if (lean) {
+        // we have to execute callback at the nextTick to be compatible
+        // with parallelLimit, as `results` variable has TDZ issue if we
+        // execute the callback synchronously
+        return immediate(() => callback(null, doc));
+      }
       if (!(doc instanceof _this)) {
         try {
           doc = new _this(doc);
@@ -3439,13 +3446,6 @@ Model.$__insertMany = function(arr, options, callback) {
       }
       if (options.session != null) {
         doc.$session(options.session);
-      }
-      // If option `lean` is set to true bypass validation
-      if (lean) {
-        // we have to execute callback at the nextTick to be compatible
-        // with parallelLimit, as `results` variable has TDZ issue if we
-        // execute the callback synchronously
-        return immediate(() => callback(null, doc));
       }
       doc.$validate({ __noPromise: true }, function(error) {
         if (error) {
@@ -3510,7 +3510,7 @@ Model.$__insertMany = function(arr, options, callback) {
       callback(null, []);
       return;
     }
-    const docObjects = docAttributes.map(function(doc) {
+    const docObjects = lean ? docAttributes : docAttributes.map(function(doc) {
       if (doc.$__schema.options.versionKey) {
         doc[doc.$__schema.options.versionKey] = 0;
       }
@@ -3572,6 +3572,9 @@ Model.$__insertMany = function(arr, options, callback) {
             return !isErrored;
           }).
           map(function setIsNewForInsertedDoc(doc) {
+            if (lean) {
+              return doc;
+            }
             doc.$__reset();
             _setIsNew(doc, false);
             return doc;
@@ -3588,9 +3591,11 @@ Model.$__insertMany = function(arr, options, callback) {
         return;
       }
 
-      for (const attribute of docAttributes) {
-        attribute.$__reset();
-        _setIsNew(attribute, false);
+      if (!lean) {
+        for (const attribute of docAttributes) {
+          attribute.$__reset();
+          _setIsNew(attribute, false);
+        }
       }
 
       if (rawResult) {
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
Looks ok, but i will hold-off on approving until the main 6.x branch does not have failing tests anymore

vkarpov15:
@hasezoey can you please re-review? This PR has fix for test issues in 6.x.

AbdelrahmanHafez:
@vkarpov15 @hasezoey 
If I understand this correctly, would this mean that documents created by `insertMany` and `lean` options would bypass the casting step? If that is true, then that feels like a foot-gun, and might be useful to have a big warning for that behavior.

If not, then LGTM :grin:

hasezoey:
> If I understand this correctly, would this mean that documents created by insertMany and lean options would bypass the casting step?

i guess that would be true, but isnt `lean`'s whole point that it basically returns what the driver returns (/ like when using mongodb directly) instead of going through mongoose's steps?

also the documentation already mentions it:
> The [lean option](https://mongoosejs.com/docs/api/query.html#query_Query-lean) tells Mongoose to skip [hydrating](https://mongoosejs.com/docs/api/model.html#model_Model-hydrate) the result documents
> ...
> The downside of enabling lean is that lean docs don't have:
> 
>    Change tracking
>    Casting and validation
>    Getters and setters
>    Virtuals
>    save()

[source](https://mongoosejs.com/docs/tutorials/lean.html)

AbdelrahmanHafez:
Yes, and that's fine in the case of reads, assuming that it's mostly used in the context of reading and then returning in a JSON response.

But in the case of writes, very often data is not cast to the proper types, and it would be stored in the database with the wrong type. I feel like people would use that option thinking that it's only a lighter way to "return" the data, unaware of what it means under the hood. 🤔

vkarpov15:
I can see @AbdelrahmanHafez 's point, but I'm not sure I fully agree. Even before this change, `lean` would bypass validation, so it would still potentially lead to storing invalid data. I can see the justification for adding a warning in the docs that `lean` with `insertMany` skips casting and validation entirely, but I don't think adding a casting step by default is beneficial. Users can just use `Model.castObject()` to cast, like `TestModel.insertMany(docs.map(doc => TestModel.castObject(doc)), { lean: true })`

AbdelrahmanHafez:
Yeah, I am not referring to _this change_ as the concern, it's generally having the `lean` option with `insertMany` that skips the casting step.

Now that I think about it, a warning would be enough, especially that we have `castObject`.

Issue #14378, opened 2024-02-26
Missing Typescript detail on Mongoose-specific `options` for updateOne, updateMany, etc. 


sderrow:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.2.0

### Node.js version

18.18.0

### MongoDB server version

7.0

### Typescript version (if applicable)

4.5.2

### Description

In #14342 , the `MongooseQueryOptions` type was edited to include `& { [other: string]: any }` in order to support arbitrary options. Arbitrary options are fine, but the specific implementation broke Typescript support for all of the official Mongoose option properties, in the sense that there's no more auto-completion or in-line JSDoc displayed when you start typing something like "runValidators".

This is caused by the use of the `Omit` Typescript utility in the type for `options` in `updateOne`, `updateMany`, and a few others. Since now `Omit` is being called on a type that has `[other: string]: any` in it, that particular weak "other" property essentially wipes out all the official properties (I think due to the underlying implementation of `Exclude`).

### Steps to Reproduce

```ts
export const queryOptionTypeTest = async (): Promise<void> => {
  type Test = {
    _id: mongoose.Types.ObjectId;

    testProp: string;

    createdAt: number;
    updatedAt: number;
  };

  const schema = new mongoose.Schema<Test>(
    {
      testProp: { type: String },
      createdAt: { type: Number },
      updatedAt: { type: Number },
    },
    {
      timestamps: { currentTime: () => new Date().valueOf() },
    }
  );

  const M = mongoose.model<Test>("Test", schema);

  // Typescript offers no help for the `timestamps` or `runValidators` options below.
  // It treats them as if they were arbitrary options instead of formal ones with known meanings
  await M.updateMany({}, { testProp: "hello" }, { timestamps: false, runValidators: true }).exec();
};
```
Here's a screenshot of my IDE with the bug and trying to auto-complete the "timestamps" options. You can see it only shows "maxTimeMS" instead. 
![updateOptionsBugWithoutFix](https://github.com/Automattic/mongoose/assets/28607517/12be08c5-2737-496f-8f93-65d14b378f7d)

### Expected Behavior

Here's what I would expect my IDE to show me. You can see it includes auto-complete and helpful documentation for "timestamps" now.
![updateOptionsBugWithFix](https://github.com/Automattic/mongoose/assets/28607517/bc4bce3a-3e62-43c3-9c9e-a14254464db5)

[...]

PR #14379, opened 2024-02-26
[fix] Add back Typescript help for true Mongoose-specific `options` for updateOne, updateMany, etc. 

sderrow:
**Summary**
Addresses #14378. `MongooseQueryOptions` type was recently edited to include `& { [other: string]: any }` in order to support arbitrary options. This implementation broke Typescript support for all of the official Mongoose option properties, in the sense that there's no more auto-completion when you start typing something like "runValidators", and there's no in-line JSDoc displayed.

This is caused by the use of the `Omit` Typescript utility in the type for `options` in `updateOne`, `updateMany`, and a few others. Resolution is to redo the implementation with a build-up approach instead of using `Omit`.

**Examples**
See example and screenshots in #14378

### DIFF
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -228,7 +228,7 @@ declare module 'mongoose' {
     /** Creates a `countDocuments` query: counts the number of documents that match `filter`. */
     countDocuments(
       filter?: FilterQuery<TRawDocType>,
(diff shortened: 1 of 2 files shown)


sderrow:
Fixes #14378

vkarpov15 (approved):
I think I prefer this approach over #14382, but I also want to merge the `ExcludeKeys` helper because `Omit<>` is awful. I'll merge this one as the fix, but also merge #14382 with some changes.

PR #14382, opened 2024-02-26
fix: missing typescript details on options params of updateMany, updateOne, etc.

### LINKED ISSUE #14378: Missing Typescript detail on Mongoose-specific `options` for updateOne, updateMany, etc.

FaizBShah:
Fixes #14378

**Summary**
Fixes the missing typescript details from the `options` parameter of functions like `updateMany()`, `updateOne()`, etc. The issue happened because the `Omit` in `Omit<MongooseQueryOptions<TRawDocType>, 'lean'>` was making the type of the keys vague because `MongooseQueryOptions` contains `[other: string]: any`. To fix this, introdduced a new TS Util function called `ExcludeKeys` which fixes this logic without weakening the type

**Examples**
The property `timestamps` and other such params are now visible
Screenshot 2024-02-26 190616

Its still supporting arbitrary options like `fallback` without any type issues
Screenshot 2024-02-26 190710

### DIFF
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -291,7 +291,7 @@ declare module 'mongoose' {
      */
     deleteOne(
       filter?: FilterQuery<TRawDocType>,
-      options?: (mongodb.DeleteOptions & ExcludeKeys<MongooseQueryOptions<TRawDocType>, 'lean' | 'timestamps'>) | null
+      options?: (mongodb.DeleteOptions & MongooseBaseQueryOptions<TRawDocType>) | null
     ): QueryWithHelpers<
       mongodb.DeleteResult,
       THydratedDocumentType,
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -228,7 +228,7 @@ declare module 'mongoose' {
     /** Creates a `countDocuments` query: counts the number of documents that match `filter`. */
     countDocuments(
       filter?: FilterQuery<TRawDocType>,
-      options?: (mongodb.CountOptions & Omit<MongooseQueryOptions<TRawDocType>, 'lean' | 'timestamps'>) | null
+      options?: (mongodb.CountOptions & ExcludeKeys<MongooseQueryOptions<TRawDocType>, 'lean' | 'timestamps'>) | null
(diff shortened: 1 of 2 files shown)


sderrow:
Can't formally review, but LGTM!

Issue #14113, opened 2023-11-22
4.7.7 -> 5.13.21 reduction performance


0Ams:
### Prerequisites

- [X] I have written a descriptive issue title

### Mongoose version

5.13.21

### Node.js version

18.15.0

### MongoDB version

4.4.25

### Operating system

None

### Issue

# Context
* 4.7.7 -> 5.13.21 reduction performance
* Schema validate and some features appear to be slowing down. Is it common for v5 to perform less than v4? 

# Test Code
```
const mongoose4 = require('mongoose4');
const mongoose5 = require('mongoose5');
const {getModel} = require('./model/test_model');
const DATA_AMOUNT = {min: 40, max: 400};

async function run() {
    mongoose4.Promise = global.Promise;
    await mongoose4.connect('mongodb://localhost:47017/m4', {useMongoClient: true});
    await mongoose5.connect('mongodb://localhost:47017/m5', {useNewUrlParser: true, useUnifiedTopology: true});

    const Model4 = mongoose4.model('Test', getModel(mongoose4));
    const Model5 = mongoose5.model('Test', getModel(mongoose5));

    const reps = 100;

    let m5total = 0;
    for (let i = 0; i < reps; ++i) {
        m5total += await profile(Model5);
    }
    let m4total = 0;
    for (let i = 0; i < reps; ++i) {
        m4total += await profile(Model4);
    }

    console.log(`mongoose v5 spendTime:`, m5total, 'average:', m5total / reps);
    console.log(`mongoose ${mongoose4.version} spendTime:`, m4total, 'average:', m4total / reps);
    console.log('mongoose 5 perf reduction in ms:', (m5total - m4total) / reps);

    process.exit(0);
}

async function profile(Model) {
    await Model.remove({});
    for (let i = 0; i < 100; ++i) {
        const record = new Model();
        for (let i = 0; i < Math.random() * (DATA_AMOUNT.max - DATA_AMOUNT.min) + DATA_AMOUNT.min; ++i) {
            const random = Math.random();
            record.randomNumberData.push(random);
            record.randomStringData.push(`${random}`);
        }
        await record.save();
    }

    let totalMS = 0;

    const cursor = Model.find().cursor();
    for (let doc = await cursor.next(); doc != null; doc = await cursor.next()) {
        const start = Date.now();
        await Model.findById(doc._id);
        totalMS += Date.now() - start;
    }

    return totalMS;
}

run().catch((error) => console.error(error.stack));
```
# Result 
[...]

Issue #13456, opened 2023-06-01
Why do Mongoose v6 and v7 cause a significant increase in garbage collections and event loop latency compared to v5?


emilisaksson:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the performance issue has not already been reported

### Last performant version

5.12.4

### Slowed down in version

6.11.1

### Node.js version

16.18.0

### 🦥 Performance issue

Background:
We have been trying to migrate from Mongoose v5 to v6 in our SaaS platform for the past year with no luck. The issues described here also occurs when running Mongoose 7.
Our software have a decent amount of web traffic(about 25 requests per second) and is using a single MongoDB Atlas cluster where each SaaS client is using its own database.
The software is hosted on Heroku.

The issue:
When we deploy a version of our software that is using Mongoose 6 or 7 it runs fine for up to an hour, then we start to see an increase in number of garbage collections, time spent in garbage collection, and event loop latency. This continues until the servers event loop is completely hogged and cannot handle any more incoming requests.
The memory usage is unaffected, so it doesn't seem to be a memory leak.

### Steps to Reproduce

The code:

Upon start, we create a database connection with a pool of 300 connections:
```
global.databaseConnection = await mongoose.createConnection(global.config.dbConnectionString, {
	maxPoolSize: 300,
	minPoolSize: 300
});
```

When a request for a specific client is received, a connection to that specific clients database is drawn from the connection pool using the useDb-function:
```
clientDatabaseConnection = global.databaseConnection.useDb(`client-${client.databaseName}`, {
	noListener: true,
	useCache: true
});
```

The client database connection is then used to do various quesries required by the request, for example:
```
let clientUserCollection = connection.model<IClientUserModel>("ClientUser", ClientUserSchema, 'users');
let user = await clientUserCollection.findById(userId).exec();
```

This is basically the structure of the application. What is it that is different in v6 and v7 compared to v5 that could cause these performance implications?

hasezoey:
> This is basically the structure of the application. What is it that is different in v6 and v7 compared to v5 that could cause these performance implications?

one of the biggest changes from v5 to v6 was that mongodb (the underlying driver) was completely re-written in typescript, which could be a likely cause

thandaanda:
We are also experiencing the same issue. We are restarting our servers every 24 hours. Memory seems fine. CPU hovers around 50%., but app becomes completely useless till we restart.

samirkatyar:
[...]

PR #14383, opened 2024-02-26
perf(document+schema): small optimizations to make `init()` faster

vkarpov15:
Re: #14113

**Summary**

Without this change:

```
$ node gh-14113.js 
6.12.6
mongoose spendTime: 24358 average: 48.716
$ node gh-14113.js 
6.12.6
mongoose spendTime: 24956 average: 49.912
$ node gh-14113.js 
6.12.6
mongoose spendTime: 25182 average: 50.364

```

With this change:

```
$ node gh-14113.js 
6.12.6
mongoose spendTime: 24003 average: 48.006
$ node gh-14113.js 
6.12.6
mongoose spendTime: 24222 average: 48.444
$ node gh-14113.js 
6.12.6
mongoose spendTime: 24247 average: 48.494
```

Seems like a small difference, but if you remove `init` entirely the average goes down to about 47.5, so this is actually a fairly significant speedup of `init()`. I'm assuming that the rest of the relatively small performance degradation in Mongoose 5 and 6 is due to bson slowdowns discussed in #13456.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -741,7 +741,7 @@ function init(self, obj, doc, opts, prefix) {
     if (i === '__proto__' || i === 'constructor') {
       return;
     }
-    path = prefix + i;
+    path = prefix ? prefix + i : i;
     schemaType = docSchema.path(path);
 
     // Should still work if not a model-level discriminator, but should not be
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
Looks ok, but i will hold-off on approving until the main 6.x branch does not have failing tests anymore

PR #14384, opened 2024-02-26
types(querycursor): correct cursor async iterator type with `populate()` support

### LINKED ISSUE #14374: Populate with TypeScript not working on Cursor
### Prerequisites

### Mongoose version

8.2.0

### Node.js version

20.x

### MongoDB server version

7.05

5.3.3

### Description

Applying the `Paths` generic parameter as described in the [mongoose documentation](https://github.com/Automattic/mongoose/blob/master/docs/typescript/populate.md) and fetching documents in a cursor does not result in the correct type.

### Steps to Reproduce

```ts
import { Schema, model, Types } from 'mongoose'

// `Parent` represents the object as it is stored in MongoDB
interface Parent {
    child?: Types.ObjectId
    name?: string
}
const ParentModel = model<Parent>(
    'Parent',
    new Schema({
        child: { type: Schema.Types.ObjectId, ref: 'Child' },
        name: String
    })
)

interface Child {
    name: string
}
const childSchema: Schema = new Schema({ name: String })
const _ChildModel = model<Child>('Child', childSchema)

async function main() {
    // This works
    const result = await ParentModel.find({}).populate<{ child: Child }>('child').orFail()
    for (const doc of result) {
        const _t: string = doc.child.name
    }

    // This does not
    const cursor = ParentModel.find({}).populate<{ child: Child }>('child').orFail()
    for await (const doc of cursor) {
        const _t: string = doc.child.name // doc.child is Types.ObjectId
    }
}
```

### Expected Behavior

The populated `child` should be of type `Child` as defined in the `Paths` generic parameter

vkarpov15:
Fix #14374

**Summary**

Right now, using `for await` with a cursor doesn't quite work in TypeScript: you get `DocType[]` rather than `ResultType`. Similarly, that means `populate<Paths>()` doesn't work right with `cursor()`.

**Examples**

### DIFF
--- a/test/types/querycursor.test.ts
+++ b/test/types/querycursor.test.ts
@@ -43,7 +43,7 @@ async function gh14374() {
   // This does not
   const cursor = ParentModel.find({}).populate<{ child: Child }>('child').cursor();
   for await (const doc of cursor) {
-    const _t: string = doc.child.name;
+    expectType<Child>(doc.child);
   }
 
 }
--- a/test/types/querycursor.test.ts
+++ b/test/types/querycursor.test.ts
@@ -20,3 +20,30 @@ Test.find().cursor().
     expectType<number>(i);
   }).
   then(() => console.log('Done!'));
+
+async function gh14374() {
+  // `Parent` represents the object as it is stored in MongoDB
+  interface Parent {
+    child?: Types.ObjectId
+    name?: string
+  }
+  const ParentModel = model<Parent>(
+    'Parent',
+    new Schema({
+        child: { type: Schema.Types.ObjectId, ref: 'Child' },
+        name: String
+    })
+  )
+
+  interface Child {
+    name: string
+  }
+  const childSchema: Schema = new Schema({ name: String });
+
+  // This does not
+  const cursor = ParentModel.find({}).populate<{ child: Child }>('child').cursor()
+  for await (const doc of cursor) {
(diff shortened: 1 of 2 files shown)


test/types/querycursor.test.ts
```diff
@@ -20,3 +20,30 @@ Test.find().cursor().
...
+      name: String
+    })
+  );
+
+  interface Child {
+    name: string
+  }
+  const childSchema: Schema = new Schema({ name: String });
+
+  // This does not
```
hasezoey, line 43:
```suggestion
```

incomplete and likely unnecessary comment?

test/types/querycursor.test.ts
```diff
@@ -20,3 +20,30 @@ Test.find().cursor().
...
+  const cursor = ParentModel.find({}).populate<{ child: Child }>('child').cursor();
+  for await (const doc of cursor) {
+    const _t: string = doc.child.name;
```
hasezoey, line 46: maybe add a `expectType` here for `doc`, or is it a very complex type here?

hasezoey (approved):
Looks good, some small suggestions / question

test/types/querycursor.test.ts
vkarpov15: `doc` is a bit complex, but I added an `expectType` on `doc.child`.

PR #14385, opened 2024-02-26
feat: `pathsToSave` option to `save()` function

### LINKED ISSUE #9583: Support passing a list of paths to `save()`
Like how [`validate()` supports a list of paths to validate](https://mongoosejs.com/docs/api/document.html#document_Document-validate), we allow passing an array `pathsToSave` that tells Mongoose to only validate and save the paths in `pathsToSave`.

See this Tweet for motivation: https://twitter.com/dreinidaho/status/1331404327412494338

IslandRhythms:
closes #9583


lib/model.js
```diff
@@ -736,8 +736,11 @@ function handleAtomics(self, where, delta, data, value) {
  */
 
 Model.prototype.$__delta = function() {
-  const dirty = this.$__dirty();
-
+  let dirty = this.$__dirty();
+  if (this.$__.saveOptions && this.$__.saveOptions.pathsToSave) {
```
vkarpov15, line 740: This works, but on further thought I'd prefer if the logic that handles `pathsToSave` was in `$__handleSave()` rather than `$__delta()`. `$__handleSave()` calls `$__delta()` so shouldn't be too hard

vkarpov15 (changes requested):

FFUV (approved):
LGTM!

test/model.test.js
```diff
@@ -2537,6 +2537,18 @@ describe('Model', function() {
       assert.ok(!doc.$__.$versionError);
       assert.ok(!doc.$__.saveOptions);
     });
+    it('should only save paths specificed in the `pathsToSave` array (gh-9583)', async function() {
+      const schema = new Schema({ name: String, age: Number, weight: Number, location: String });
+      const Test = db.model('Test', schema);
+      await Test.create({ name: 'Test Testerson', age: 1, weight: 180, location: 'Florida' });
+      const doc = await Test.findOne();
+      doc.name = 'Test';
+      doc.age = 100;
```
vkarpov15, line 2546: We should also test that `pathsToSave` bypasses validation on paths that aren't in `pathsToSave`. For example, if we added `validate: v => v == null || v >= 140` to `weight` schema path, would the below `save()` still succeed? It should.

vkarpov15 (changes requested):
Let's also add some more test cases that cover:

1. Single nested subdocs. What happens if I do `save({ pathsToSave: ['subdoc.name'] })` and `save({ pathsToSave: ['subdoc'] })` where we have `subdoc: Schema({ name: String })`? Make sure to cover both cases where `name` is modified like `doc.subdoc.name = 'foo'` as well as `doc.subdoc = { name: 'bar' }`
2. Doc arrays using the same cases as above

lib/document.js
```diff
@@ -2927,6 +2930,11 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
   let total = 0;
 
   for (const path of paths) {
+    if (_this.$__.saveOptions && _this.$__.saveOptions.pathsToSave) {
+      if (!_this.$__.saveOptions.pathsToSave.includes(path)) {
+        continue;
+      }
+    }
```
AbdelrahmanHafez, line 2937: Nitpick:
```suggestion
if (
  Array.isArray(_this.$__.saveOptions?.pathsToSave) &&
  _this.$__.saveOptions.pathsToSave.includes(path) === false
  ) {
    continue;
}
```

Issue #14353, opened 2024-02-14
BSONError thrown when calling _id.toString() after using R.clone after mongoose bump


Axosoft-Ashley:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.1.2

### Node.js version

18

### MongoDB server version

6.0

### Description

After bumping from v5.13.20 to 8.1.2, when R.clone() is called on an array of mongoose documents a BSON error is thrown, this did not happen on the older version of mongoose.
`throw new BSONError(`Cannot create Buffer from ${String(potentialBuffer)}`);`

the ramda version ^0.29.1 was used for both old and new mongoose test

### Steps to Reproduce

schema setup
```
const userSchema = new mongoose.Schema({
  _id: { type: String, default: () => uuid.v4() }, // no need to set the index on _id, mongo will do that for us
});
const UserModel = mongoose.model('User', userSchema);

const organizationUser = {
  user: { type: String, ref: 'User' }
};

const organizationSchema = new mongoose.Schema({
  members: [organizationUser],
}, { usePushEach: true });
```
crash code:
```
    foundOrgModel = await OrganizationModel.findOne({})
    const members = R.clone(foundOrgModel.members);
    console.log(members[0]._id.toString()) // crashes
```

### Expected Behavior

it should not throw but instead just log the id as a string like id did in v5.13.20 of mongoose

gitkrakel:
I am also able to reproduce with `Document.prototype.$clone` instead of Ramda's clone by modifying an existing unit-test to use a subdocument array:

```js
  it.only('modification of test "$clone() (gh-11849)"', async function() {
    const schema = new mongoose.Schema({
      subDocArray: [{
        name: {
          type: String,
          validate: {
            validator: (v) => v !== 'Invalid'
          }
        }
      }]
    });
    const Test = db.model('Test', schema);

    const item = await Test.create({ subDocArray: [{ name: 'Test' }] });

    const doc = await Test.findById(item._id);
    const clonedDoc = doc.$clone();

    // This assertion fails
    assert.deepEqual(clonedDoc.subDocArray[0], doc.subDocArray[0]);
  });
```

IslandRhythms:
```javascript
const mongoose = require('mongoose');
const assert = require('assert');

const testSchema = new mongoose.Schema({
  subDocArray: [{
    name: {
      type: String,
      validate: {
[...]

PR #14395, opened 2024-02-29
fix(document): make `$clone` avoid converting subdocs into POJOs

vkarpov15:
Re: #14353

**Summary**

Fix for [this issue from #14353](https://github.com/Automattic/mongoose/issues/14353#issuecomment-1944564724), there are a couple of unrelated issues reported in #14353. This particular issue is that document `$clone` method converts subdocs into POJOs, which is definitely incorrect.

**Examples**

### DIFF
--- a/lib/helpers/clone.js
+++ b/lib/helpers/clone.js
@@ -41,6 +41,17 @@ function clone(obj, options, isArrayChild) {
     if (options && options._skipSingleNestedGetters && obj.$isSingleNested) {
       options = Object.assign({}, options, { getters: false });
     }
+    if (options != null && options.retainDocuments != null && obj.$__ != null) {
+      const clonedDoc = obj.$clone();
+      if (obj.__index != null) {
+        clonedDoc.__index = obj.__index;
+      }
+      if (obj.__parentArray != null) {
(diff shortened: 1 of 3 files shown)


lib/helpers/clone.js
```diff
@@ -41,6 +41,17 @@ function clone(obj, options, isArrayChild) {
     if (options && options._skipSingleNestedGetters && obj.$isSingleNested) {
       options = Object.assign({}, options, { getters: false });
     }
+    if (options != null && options.retainDocuments != null && obj.$__ != null) {
```
hasezoey, line 44: this condition seems out of style with the other if's around it, any specific reason everything is tested against null?
(minor style suggestion)

hasezoey (commented):
Looks good, would recommend to merge master again to verify tests, also one minor style question / suggestion.

lib/helpers/clone.js
vkarpov15: Good call, this is a bug, `options.retainDocuments != null` is incorrect because `options.retainDocuments === false` should turn off `retainDocuments` functionality.

hasezoey (approved):
Looks good to me; i didnt initially notice that the check for a boolean would not be working correctly.

PR #14398, opened 2024-03-01
fix: fixed the typing of FilterQuery<T> type to prevent it from only getting typed to any

### LINKED ISSUE #14397: Filter query `Condition` type might as well be `any`
### Prerequisites

### Mongoose version

8.2.0

### Node.js version

18.18.0

### MongoDB server version

7.0

4.5.2

### Description

The following appears in [query.d.ts](https://github.com/Automattic/mongoose/blob/4b34967e0c2d33fe3440e0742672dd47e01c0a0c/types/query.d.ts#L4-L9):
```ts
export type ApplyBasicQueryCasting<T> = T | T[] | (T extends (infer U)[] ? U : any) | any;
type Condition<T> = ApplyBasicQueryCasting<T> | QuerySelector<ApplyBasicQueryCasting<T>>;

type _FilterQuery<T> = {
  [P in keyof T]?: Condition<T[P]>;
} & RootQuerySelector<T>;
```

The `| any` union in `ApplyBasicQueryCasting<T>`, means that `ApplyBasicQueryCasting` is `any`, which carries through to `Condition<T>` being `any` as well.

I think that defeats the purpose of the `Condition` type in the first place, since now there's no support for prompting the user with valid Mongo operators like `$regex` and `$gte` or displaying type errors if you try to pass an invalid value into such an operator (e.g., `{ $regex: 402 }` or `{ $exists: "hello" }`).

### Steps to Reproduce

```ts
type Test = {
  name: string;
  age: number;
};

// No Typescript error below even though `$regex` cannot be passed a number
// Additionally, there's no auto-complete help for the values of the query object
export const query: FilterQuery<Test> = { name: { $regex: 10 } };
```

### Expected Behavior

Additional Typescript support when it comes to writing filter queries. I understand there's a need for `any`, but is it possible to do it without sacrificing the operator help?

FaizBShah:
Fixes #14397 

**Summary**
Fixed the typing to of `FilterQuery<T>` by preventing it from getting type-casted to `any` always. Strengthened the type by removing `any` and still keeping the ability to add generic fields. Fixed the typescript issue of `$regex`, `$eq` not showing the hint messages.

**Examples**
Screenshot 2024-03-01 233516

### DIFF
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -84,6 +84,8 @@ Test.find({ parent: { $in: ['0'.repeat(24)] } });
 Test.find({ name: { $in: ['Test'] } }).exec().then((res: Array<ITest>) => console.log(res));
 Test.find({ tags: 'test' }).exec();
 Test.find({ tags: { $in: ['test'] } }).exec();
+Test.find({ tags: /test/ }).exec();
+Test.find({ tags: { $in: [/test/] } }).exec();
 
 // Implicit `$in`
 Test.find({ name: ['Test1', 'Test2'] }).exec();
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -324,8 +324,8 @@ function gh11964() {
   let idCondition: Condition<WithId<TestUser>['id']>;
   let filter: FilterQuery<WithId<TestUser>>;
 
-  expectType<typeof idCondition>(id);
-  expectType<typeof filter>({ id });
+  expectAssignable<typeof idCondition>(id);
+  expectAssignable<typeof filter>({ id });
 }
 
 function gh12091() {
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -88,6 +88,9 @@ Test.find({ tags: { $in: ['test'] } }).exec();
 // Implicit `$in`
 Test.find({ name: ['Test1', 'Test2'] }).exec();
 
+// Implicit `$in` for regex string
+Test.find({ name: [/Test1/, /Test2/] });
+
 Test.find({ name: 'test' }, (err: Error | null, docs: ITest[]) => {
   console.log(!!err, docs[0].age);
 });
@@ -311,16 +314,18 @@ function gh11964() {
 
   type WithId<T extends object> = T & { id: string };
 
-  class Repository<T extends object> {
-    /* ... */
+  type TestUser = {
+    name: string;
+    age: number;
+  };
 
-    find(id: string) {
-      const idCondition: Condition<WithId<T>>['id'] = id; // error :(
-      const filter: FilterQuery<WithId<T>> = { id }; // error :(
+  const id: string = 'Test Id';
 
-      /* ... */
-    }
-  }
+  let idCondition: Condition<WithId<TestUser>['id']>;
+  let filter: FilterQuery<WithId<TestUser>>;
+
+  expectType<typeof idCondition>(id);
+  expectType<typeof filter>({ id });
 }
 
 function gh12091() {
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -316,6 +316,19 @@ function gh11964() {
 
   type WithId<T extends object> = T & { id: string };
 
+  class Repository<T extends object> {
+    find(id: string) {
+      const idCondition: Condition<ReturnType<(arg: WithId<T>) => typeof arg.id>> = id;
+      const filter: FilterQuery<WithId<T>> = { id } as FilterQuery<WithId<T>>;
+    }
+  }
+}
+
+function gh14397() {
+  type Condition<T> = ApplyBasicQueryCasting<T> | QuerySelector<ApplyBasicQueryCasting<T>>; // redefined here because it's not exported by mongoose
+
+  type WithId<T extends object> = T & { id: string };
+
   type TestUser = {
     name: string;
     age: number;
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -16,7 +16,7 @@ declare module 'mongoose' {
           ? BufferQueryTypeCasting
           : T;
 
-  export type ApplyBasicQueryCasting<T> = T | T[] | (T extends (infer U)[] ? U : never);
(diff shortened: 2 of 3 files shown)


FaizBShah:
Hi @AbdelrahmanHafez , can you review this PR whenever you feel free?

types/query.d.ts
```diff
@@ -237,7 +252,7 @@ declare module 'mongoose' {
     allowDiskUse(value: boolean): this;
 
     /** Specifies arguments for an `$and` condition. */
-    and(array: FilterQuery<RawDocType>[]): this;
+    and(array: FilterQuery<DocType>[]): this;
```
vkarpov15, line 255: Why changing from `RawDocType` to `DocType` here? I think `RawDocType` is more correct to avoid filtering by document properties like `$isNew`.

test/types/queries.test.ts
```diff
@@ -311,16 +316,18 @@ function gh11964() {
 
   type WithId<T extends object> = T & { id: string };
 
-  class Repository<T extends object> {
-    /* ... */
+  type TestUser = {
```
vkarpov15, line 319: Please leave the `class Repository<>` here, this test case was specifically for an issue that occurred with generic classes.

types/query.d.ts
```diff
@@ -1,8 +1,23 @@
...
+    ? StringQueryTypeCasting
+    : T extends Types.ObjectId
+      ? ObjectIdQueryTypeCasting
+      : T extends Types.UUID
+        ? UUIDQueryTypeCasting
+        : T extends Buffer
+          ? BufferQueryTypeCasting
+          : T;
+
+  export type ApplyBasicQueryCasting<T> = T | T[] | (T extends (infer U)[] ? QueryTypeCasting<U> : never);
```
vkarpov15, line 19: Does `ApplyBasicQueryCasting` still support arbitrary keys? Because it should.

vkarpov15 (commented):
In general I'm supportive of better autocomplete for FilterQuery, but we need to make sure that we still support arbitrary keys in FilterQuery and make sure tests still pass.

types/query.d.ts
sderrow: Chiming in here since I raised the original issue. Thanks for tackling @FaizBShah !

@vkarpov15 does `ApplyBasicQueryCasting` specifically need to accept arbitrary keys? Or is it just that `FilterQuery` should accept arbitrary keys? It appears that `RootQuerySelector` already accepts arbitrary keys, so we're covered on that front. Is there an example use case where there's a defined property of a schema where you need to query that specific property based on arbitrary value?

vkarpov15: That makes sense, I didn't look closely enough to see that `RootQuerySelector` supports arbitary keys. You're right that the important thing is for `FilterQuery` to support arbitrary keys. In that case this is fine.

types/query.d.ts
FaizBShah: The reason for changing this was because some test cases like the following `autoTypedQuery()` were failing w.r.t. types:
```
function autoTypedQuery() {
  const AutoTypedModel = autoTypedModel();
  const query = AutoTypedModel.find();
  expectType<typeof query>(AutoTypedModel.find().byUserName('')); // This was throwing type error
}
```

Now, the reason these function were failing because in `models.d.ts`, a function like `find()` was defined like this:
```
find<ResultDoc = THydratedDocumentType>(
    ): QueryWithHelpers<Array<ResultDoc>, ResultDoc, TQueryHelpers, TRawDocType, 'find'>;
```

Notice how in the above declared `find()` type, DocType = ResultDoc and RawDocType = TRawDocType for the corresponding `Query` type which will be generated by the `QueryWithHelpers` type. This also means that the corresponding filters type for that query will be `FilterQuery<TRawDocType>`

**BUT**, in the above autoTypedTest's `byUserName()` method, the `this.query` context being used inside that method has the type `Query` with both DocType and RawDocType being set as the THydratedDocumentType due to auto typing. Thus, when you call the `byUserMethod`, it expects filters of type `FilterQuery<ResultDoc>` instead of `FilterQuery<TRawDocType>`, thus creating a type mismatch error.

To prevent the above type conflict, I chose to replace the `RawDocType` with just `DocType`

test/types/queries.test.ts
FaizBShah: I actually did it because it was failing, but I figured out a way to fix this now by writing it like this and keeping the class Repository<>:
```
class Repository<T extends object> {
    find(id: string) {
      const idCondition: Condition<WithId<T>['id']> = id as Condition<WithId<T>['id']>;
      const filter: FilterQuery<WithId<T>> = { id } as FilterQuery<WithId<T>>;
    }
  }
```

Not sure if this is the best way, but its passing the tests + you still get autocompletion for stuffs like `$eq`, `$in`, etc. in the above test too

vkarpov15: I would strongly prefer if this test passed as written before we merge this PR. I won't say it's a hard blocker, but we shouldn't merge this PR unless either this test passes or we have a very good explanation of why this test is failing. And "just write code this other way" is generally not a good enough reason, because different people write code in different ways and we should meet users where they are.

FaizBShah: That is the reason why I wrote the above test. If you approve it, I can write the above test in a separate test function and push that commit to this PR. As for the existing old test, do suggest what should be done about it?

FaizBShah: @vkarpov15 Also, this is a wrong statement which will always fail even if you give a concrete interface implementation instead of just parameter `T`:
```
const idCondition: Condition<WithId<T>>['id'] = id;
```

The reason is because `Condition<K>` will return something like `K | K[] | ...`. Now, if the return type of `Condition<K>` is `K[]`, then it will not have the property `['id']` obviously since its an array. So, the above statement was also written with incorrect typescript assumption

FaizBShah: @vkarpov15 I have added new commit separating out the tests. Can you review the changes now?

vkarpov15:
Closing in favor of #14436. Thanks @FaizBShah :+1:

PR #14415, opened 2024-03-05
docs(model): add extra note about `lean` option for `insertMany()` skipping casting

vkarpov15:
Re: #14376

**Summary**

Addressing Hafez's comments on https://github.com/Automattic/mongoose/pull/14376#issuecomment-1978590482

**Examples**


lib/model.js
```diff
@@ -3372,7 +3372,7 @@ Model.startSession = function() {
...
  * @param {Boolean} [options.rawResult=false] if false, the returned promise resolves to the documents that passed mongoose document validation. If `true`, will return the [raw result from the MongoDB driver](https://mongodb.github.io/node-mongodb-native/4.9/interfaces/InsertManyResult.html) with a `mongoose` property that contains `validationErrors` and `results` if this is an unordered `insertMany`.
- * @param {Boolean} [options.lean=false] if `true`, skips hydrating and validating the documents. This option is useful if you need the extra performance, but Mongoose won't validate the documents before inserting.
+ * @param {Boolean} [options.lean=false] if `true`, skips hydrating the documents. This means Mongoose will **not** cast or validate any of the documents passed to `insertMany()`. This option is useful if you need the extra performance, but comes with data integrity risk. Consider using with [`castObject()`](#Model.castObject()).
```
hasezoey, line 3375:
```suggestion
 * @param {Boolean} [options.lean=false] if `true`, skips hydrating the documents. This means Mongoose will **not** cast or validate any of the documents passed to `insertMany()`. This option is useful if you need the extra performance, but comes with data integrity risk. Consider using with [`castObject()`](#model_Model-castObject).
```

6.x still uses the old api naming, will have to be changed to the new one for 7.x and up

PR #14431, opened 2024-03-12
Added Null check in case schema.options['type'][0] is undefined

Atharv-Bobde:
**Summary**

In the process of creating a new document, I encountered a TypeError. Upon investigating the error stack, it revealed that the issue was related to an undefined property 'ref' during the execution of model.$set in the Mongoose library.

Upon further examination, I noticed that there was a missing null check for schema.options[typeKey][0], where typeKey is a crucial part of the schema. In my case, schema.options[typeKey][0] was undefined, leading to a subsequent error when trying to access the 'ref' property. To address this issue, I have proposed a pull request to the Mongoose repository that includes the necessary null check for schema.options[typeKey][0]. This modification ensures a smoother handling of cases where the referenced type is absent.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1388,6 +1388,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
     if (schema.options &&
         Array.isArray(schema.options[typeKey]) &&
         schema.options[typeKey].length &&
+        schema.options[typeKey][0] &&
         schema.options[typeKey][0].ref &&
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

vkarpov15:
If you're able to provide a repro script that demonstrates this issue, that would be very helpful so we can write a test for this. But this change is sufficiently low risk that I'm comfortable merging and releasing without tests.

PR #14434, opened 2024-03-13
Add function `SchemaType.prototype.validateAll`

### LINKED ISSUE #6910: Add validateMany feature method
Feature

There is no way to run multiple validators on the shame field.

```
toySchema.path('name').validate(function(v) {
  if (v !== 'Turbo Man') {
    throw new Error('Need to get a Turbo Man for Christmas');
  }
  return true;
}, 'Name `{VALUE}` is not valid');
```

We should to be able to add multiple custom validators, and get back the result all of them.

I saw that there were some issue what reported the same, and the answer was, we can use mongoose-validator-all package. But that package have no more maintained for 1,5 years and anyway it should be supported by mongoose i think.

To prevent breaking changes, instead of overwrite the current validate system, just add a new one. 
```
toySchema.path('password').validateMany([
function(v, f) {
  if (this.name == v) {    
   return {
        'passed': false,
         'message': `field [${f}] can be the same as name, value: [${v}]`
    };
  }
  return {
    'passed': true,
  };
},function(v, f) {
  if (this.confirmPw == v) {    //just for example
   return {
        'passed': false,
         'message': `field [${f}] have to be the same as password, confirmPw, password: [${this.confirmPw}, ${v}]`
    };
  }
  return {
    'passed': true,
  };
},
]);
```

Errors:
```
{
 "password": [{error Object}, {error Object}]
}
```

UI:
For be able to compare fields whit others e.g password, this should to contain the document itself as in pre hooks.

IslandRhythms:
closes #6910

### DIFF
--- a/lib/schemaType.js
+++ b/lib/schemaType.js
@@ -803,6 +803,12 @@ SchemaType.prototype.get = function(fn) {
   return this;
 };
 
+SchemaType.prototype.validateAll = function(approvalSeekers) {
+  for (let i = 0; i < approvalSeekers.length; i++) {
+    this.validate(approvalSeekers[i]);
+  }
+}
+
 /**
  * Adds validator(s) for this document path.
  *
--- a/lib/schemaType.js
+++ b/lib/schemaType.js
@@ -803,10 +803,19 @@ SchemaType.prototype.get = function(fn) {
   return this;
 };
 
-SchemaType.prototype.validateAll = function(approvalSeekers) {
-  for (let i = 0; i < approvalSeekers.length; i++) {
-    this.validate(approvalSeekers[i]);
+/**
+ * Adds multiple validators for this document path.
+ * Calls `validate()` for every element in validators.
+ *
+ * @param {Array<RegExp|Function|Object>} validators
+ * @returns this
+ */
+
+SchemaType.prototype.validateAll = function(validators) {
+  for (let i = 0; i < validators.length; i++) {
+    this.validate(validators[i]);
   }
+  return this;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


lib/schemaType.js
```diff
@@ -803,6 +803,12 @@ SchemaType.prototype.get = function(fn) {
   return this;
 };
 
+SchemaType.prototype.validateAll = function(approvalSeekers) {
```
hasezoey, line 806: consider adding JSDoc.

test/schematype.test.js
```diff
@@ -281,4 +281,29 @@ describe('schematype', function() {
       });
     });
   });
+  it('demonstrates the `validateAll()` function (gh-6910)', async function() {
+
+    const validateSchema = new Schema({ name: String, password: String });
```
hasezoey, line 286:
```suggestion
  it('demonstrates the `validateAll()` function (gh-6910)', async function() {
    const validateSchema = new Schema({ name: String, password: String });
```

test/schematype.test.js
```diff
@@ -281,4 +281,29 @@ describe('schematype', function() {
       });
     });
   });
+  it('demonstrates the `validateAll()` function (gh-6910)', async function() {
```
hasezoey, line 284:
```suggestion
  it('demonstrates the `validateAll()` function (gh-6910)', function() {
```

function does not need to be async, from what i can tell

hasezoey (commented):
Looks good, would like to see some JSDoc and Typescript types.
Also some minor style nitpicks.

PS: changed the title to better reflect what this issue is about

hasezoey (approved):

vkarpov15 (approved):

Issue #11964, opened 2022-06-21
[Types] Adjusments to ApplyBasicQueryCasting (ff2bf3e8b3abe0e09a523465e5db8743083702b9) break FilterQueries for generic types


ghost91-:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.4.0

### Node.js version

16.x

### MongoDB server version

5.x

### Description

The changes from https://github.com/Automattic/mongoose/commit/ff2bf3e8b3abe0e09a523465e5db8743083702b9 break `FilterQuery`s for generic types that have some know properties. With this change, it's not possible anymore to assign values for the known properties.

It seems like the issue is the `T[0] &` in
```ts
  export type ApplyBasicQueryCasting<T, defaultT = T | T[] | {}> = T extends any[] ? T[0] & defaultT: defaultT;
```
What exactly is the purpose of having `T[0] & defaultT`? For arrays, that would mean the property needs to be of the element type, and the array type or an array of the array type or any non nullable type, at the same time. That doesn't make a lot of sense to me.

Also, I doubt `{}` is the correct type here. It means "any non nullable value", not "empty object". In particular, you can do stuff like `const foo: {} = true`. `null` and `undefined` are the only values that can _not_ be assigned to it. I doubt that's what's meant here.

From what I can see, the second generic parameter (`defaultT`) isn't ever set explicitly, so what is the purpose of it?

And lastly, that commit also introduces some tests, but it is in no way clear how the tests are related to the changes. And it seems like they don't cover all parts of the changes. So I think the tests should be expanded a bit here.

### Steps to Reproduce

```ts
import { ApplyBasicQueryCasting, FilterQuery, QuerySelector } from 'mongoose';

type Condition<T> = ApplyBasicQueryCasting<T> | QuerySelector<ApplyBasicQueryCasting<T>>; // redefined here because it's not exported by mongoose

type WithId<T extends object> = T & { id: string };

class Repository<T extends object> {
  /* ... */

  find(id: string) {
    const idCondition: Condition<WithId<T>>['id'] = id; // error :(
    const filter: FilterQuery<WithId<T>> = { id }; // error :(

    /* ... */
  }
}
```

### Expected Behavior

Using `FilterQuery` with generic types with know properties works as it has previously.

vkarpov15:
[...]

PR #14436, opened 2024-03-14
types: improve the typing of FilterQuery<T> type to prevent it from only getting typed to any

### LINKED ISSUE #14397: Filter query `Condition` type might as well be `any`

vkarpov15:
Re: #14398
Fix #14397

@FaizBShah WDYT?

**Summary**

#14398 but with a couple of changes:

1. Avoided changing `FilterQuery<RawDocType>` back to `FilterQuery<DocType>` for everything but `getFilter()` and `getQuery()`. For some reason, tests fail without that
2. Fixes for the #11964 test. The fact that we had to change that test was concerning, but after diving down the rabbit hole I'm convinced the #11964 test is just incorrect and will never work unless `FilterQuery` types to `any`. [Here's a couple of TypeScript playground examples demonstrating the issue](https://www.typescriptlang.org/play?#code/PTAEGUAsHsHcGdQBdIEMkBpQFMBu2A7UASwDNQBPaAV1FQBN7QADAMlAG8T6AuUeJACdiBAOagAvsywpsoQdnjUANkhHikFAA5z60RQQDkSUAWwBjRfFTDlFAFAhQafHSLNivfkPXNQWwWgdQU17TR1QABVQAF5Obj4zfEFJUHYuTz4BYTFJAG57c2gCAVAADz5ouIyvQ1JoaEN8+0cwKDhEFHQWTNM8bEE-AWJlZRwypEJ6RA8vbN9C5VR4RAAlbC1oeGIkaEEKAB5o7AmpxBqsn1yJAD5Oe1BQUhF6AApe+bEASnvHx6KSiZSABGSqxeKeZqPCT2GEA0oUPjrTbbXb7A4cXpJAa3cFmWCgZFbHZ7Q6YrzYwS3V5fPJAA), which is that with `class Repository<T>`, there's no way to type `T` so that `T` excludes `{ id: never }`, and `WithId<{ id: never }>` still types to `{id: never}`. 

**Examples**

### DIFF
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -1,4 +1,5 @@
 import {
+  Condition,
   HydratedDocument,
   Schema,
   model,
@@ -84,9 +85,14 @@ Test.find({ parent: { $in: ['0'.repeat(24)] } });
 Test.find({ name: { $in: ['Test'] } }).exec().then((res: Array<ITest>) => console.log(res));
 Test.find({ tags: 'test' }).exec();
 Test.find({ tags: { $in: ['test'] } }).exec();
+Test.find({ tags: /test/ }).exec();
+Test.find({ tags: { $in: [/test/] } }).exec();
 
 // Implicit `$in`
 Test.find({ name: ['Test1', 'Test2'] }).exec();
+
+// Implicit `$in` for regex string
+Test.find({ name: [/Test1/, /Test2/] });
 
 Test.find({ name: 'test' }, (err: Error | null, docs: ITest[]) => {
   console.log(!!err, docs[0].age);
@@ -307,20 +313,34 @@ function autoTypedQuery() {
 }
 
 function gh11964() {
+  class Repository<T extends { id: string }> {
+    find(id: string) {
+      const idCondition: Condition<T['id']> = id as Condition<T['id']>;
+
+      // `as` is necessary because `T` can be `{ id: never }`,
+      // so we need to explicitly coerce
+      const filter: FilterQuery<T> = { id } as FilterQuery<T>;
+    }
+  }
+}
+
+function gh14397() {
   type Condition<T> = ApplyBasicQueryCasting<T> | QuerySelector<ApplyBasicQueryCasting<T>>; // redefined here because it's not exported by mongoose
 
   type WithId<T extends object> = T & { id: string };
 
-  class Repository<T extends object> {
-    /* ... */
+  type TestUser = {
+    name: string;
+    age: number;
+  };
 
-    find(id: string) {
-      const idCondition: Condition<WithId<T>>['id'] = id; // error :(
-      const filter: FilterQuery<WithId<T>> = { id }; // error :(
+  const id: string = 'Test Id';
 
-      /* ... */
-    }
-  }
+  let idCondition: Condition<WithId<TestUser>['id']>;
+  let filter: FilterQuery<WithId<TestUser>>;
+
+  expectAssignable<typeof idCondition>(id);
+  expectAssignable<typeof filter>({ id });
 }
 
 function gh12091() {
--- a/scripts/tsc-diagnostics-check.js
+++ b/scripts/tsc-diagnostics-check.js
@@ -3,7 +3,7 @@
 const fs = require('fs');
 
 const stdin = fs.readFileSync(0).toString('utf8');
-const maxInstantiations = isNaN(process.argv[2]) ? 120000 : parseInt(process.argv[2], 10);
+const maxInstantiations = isNaN(process.argv[2]) ? 130000 : parseInt(process.argv[2], 10);
 
 console.log(stdin);
 
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -1,8 +1,20 @@
 declare module 'mongoose' {
   import mongodb = require('mongodb');
 
-  export type ApplyBasicQueryCasting<T> = T | T[] | (T extends (infer U)[] ? U : any) | any;
-  type Condition<T> = ApplyBasicQueryCasting<T> | QuerySelector<ApplyBasicQueryCasting<T>>;
+  type StringQueryTypeCasting = string | RegExp;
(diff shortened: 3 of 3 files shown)


FaizBShah:
Yeah, this looks good to me!! Even I personally didn't liked changing `RawDocType` to to `DocType` everywhere, but did it because I thought maybe some types might fail, but this soln. looks cleaner. And yes, the test was incorrect because it was written with the assumption that `id` will never be able to be `never`. LGTM!!

test/types/queries.test.ts
```diff
@@ -307,20 +313,34 @@ function autoTypedQuery() {
 }
 
 function gh11964() {
+  class Repository<T extends { id: string }> {
```
hasezoey, line 316: maybe we can make / have a helper that is a `BasicPojoDocument` for easy / quick / shared access

test/types/queries.test.ts
```diff
@@ -307,20 +313,34 @@ function autoTypedQuery() {
...
+    find(id: string) {
+      const idCondition: Condition<T['id']> = id as Condition<T['id']>;
+
+      // `as` is necessary because `T` can be `{ id: never }`,
+      // so we need to explicitly coerce
+      const filter: FilterQuery<T> = { id } as FilterQuery<T>;
```
hasezoey, line 322: can `T` really be `{ id: never }` if it is now explicitly set as `T extends { id: string }`? is this some kind of typescript magic i am missing?

hasezoey (commented):
Looks good to me, though i am not really involved with the semantics of these types.

test/types/queries.test.ts
FaizBShah: @hasezoey @vkarpov15 I have just tested it and checked that even writing `const idCondition: Condition<T['id']> = id` without the `as` works fine.

The 2nd one fails without the `as` keyword because [in typescript, a concrete instance is not allowed to be assigned to a type parameter](https://stackoverflow.com/questions/56505560/how-to-fix-ts2322-could-be-instantiated-with-a-different-subtype-of-constraint#:~:text=In%20typescript%2C%20a%20concrete%20instance%20is%20not%20allowed%20to%20be%20assigned%20to%20a%20type%20parameter.). This means that even if you try to write `const obj: T = { id: string };`, then it will fail. Unless you add the `as` keyword and change it to `const obj: T = { id: string } as T;` Thus, we also need to add the `as` keyword for the 2nd statement, so that it follows typescript rule. Weird behaviour by typescript, but what's not weird about typescript :)

vkarpov15: @hasezoey yep try this TypeScript playground: https://www.typescriptlang.org/play?#code/PTAEGUAsHsHcGdQBdIEMkBpQFMBu2A7UASwDNQBPaAV1FQBN7QADAMlAG8T6AuUeJACdiBAOagAvsywpsoQdnjUANkhHikFAA5z60RQQDkSUAWwBjRfFTDlFAFAhQafHSLNivfkPXNQWwWgdQU17TR1QABVQAF5Obj4zfEFJUHYuTz4BYTFJAG57c2gCAVAADz5ouIyvQ1JoaEN8+0cwKDhEFHQWTNM8bEE-AWJlZRwypEJ6RA8vbN9C5VR4RAAlbC1oeGIkaEEKAB5o7AmpxBqsn1yJAD5Oe1BQUhF6AApe+bEASnvHx6KSiZSABGSqxeKeZqPCT2GEA0oUPjrTbbXb7A4cXpJAa3cFmWCgZFbHZ7Q6YrzYwS3V5fPJAA . It is very weird that `{ id: never }` does seem to satisfy `extends { id: string }` but that's TypeScript for you :)

@FaizBShah thanks for the explanation :+1:

FaizBShah: Yupp!! Glad I could help :D

hasezoey: @vkarpov15 your example seemed to have errors, but i could reproduce it with [just this](https://www.typescriptlang.org/play?#code/PTAEGUAsHsHcGdQBdIEMmgAYEsAmAuUAOwFMA3EgJ01HiWwBsHQSAPJEo3RHA2pStiIBzTACgAxg1TxEAJRIAHaPGxJolAJ4AeACot2nbqADeoPITqCRoAL4A+U2NCgAZkNwAKC-2vCAlE4uLs52YraS0ER0oJqECsqq6lraJj6kFJQOoAC8xCSwoAkqaho6aXwZVA6e-gDcQA).

PR #14437, opened 2024-03-14
fix(model): improve update minimizing to only minimize top-level properties in the update

### LINKED ISSUE #14420: Can not update sub schema to empty object
### Prerequisites

### Mongoose version

8.2.1

### Node.js version

20.11.1

### MongoDB server version

4.4

_No response_

### Description

We ran in to an issue while migrating from mongoose 6 when trying to set the path of a sub scehma to an empty object (clearing all data). The change is reflected in the local document and the `modifiedPaths()` says that the attribute has been changed. But the change is not saved in the database.

### Steps to Reproduce

```
await mongoose.connect('mongodb://localhost:27017/mongoose_test');

const SubSchema = new mongoose.Schema({
	name: {type: String}
}, { _id: false });

const MainSchema = new mongoose.Schema({
	sub: { 
			type: SubSchema,
		}
});

const MainModel = mongoose.model('Main', MainSchema);

const doc = new MainModel({ sub: { name: 'Hello World' } });
console.log('init', doc.sub); // { name: 'Hello World' }
await doc.save();

doc.set({ sub: {} });
await doc.save();
console.log('save', doc.sub); // {} empty as expected

const savedDoc = await MainModel.findById(doc.id).orFail();
console.log('findById', savedDoc.sub); // { name: 'Hello World' } still has old value
```

### Expected Behavior

The change in the local document to be updated in the database.

vkarpov15:
Fix #14420
Re: #13782

**Summary**

The issue in #14420 is that the update to `save()` is `{ $set: { sub: {} } }`, which gets minimized away by #13782. With this PR, we'll only minimize top-level properties of update operators in the `save()` update. So in the #14420 case, we'll just minimize `sub` to `null`, but leave the rest of the update. I think this is the most consistent way to handle #14420 without undoing the changes we made in #13782.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -351,12 +351,22 @@ Model.prototype.$__handleSave = function(options, callback) {
 
     const update = delta[1];
     if (this.$__schema.options.minimize) {
-      minimize(update);
-      // minimize might leave us with an empty object, which would
-      // lead to MongoDB throwing a "Update document requires atomic operators" error
-      if (Object.keys(update).length === 0) {
-        handleEmptyUpdate.call(this);
-        return;
+      for (const updateOp of Object.values(update)) {
+        if (updateOp == null) {
+          continue;
+        }
+        for (const key of Object.keys(updateOp)) {
+          if (updateOp[key] == null || typeof updateOp[key] !== 'object') {
+            continue;
+          }
+          if (!utils.isPOJO(updateOp[key])) {
+            continue;
+          }
+          minimize(updateOp[key]);
+          if (Object.keys(updateOp[key]).length === 0) {
+            updateOp[key] = null;
+          }
+        }
       }
     }
 
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -13041,6 +13041,51 @@ describe('document', function() {
     assert.ok(doc.docArr.toString().includes('child'), doc.docArr.toString());
     assert.ok(doc.docArr.toString().includes('test child'), doc.docArr.toString());
   });
+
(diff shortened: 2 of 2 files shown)


lib/model.js
```diff
@@ -351,12 +351,22 @@ Model.prototype.$__handleSave = function(options, callback) {
...
+          if (updateOp[key] == null || typeof updateOp[key] !== 'object') {
+            continue;
+          }
+          if (!utils.isPOJO(updateOp[key])) {
+            continue;
+          }
+          minimize(updateOp[key]);
+          if (Object.keys(updateOp[key]).length === 0) {
+            updateOp[key] = null;
+          }
```
hasezoey, line 368:
```suggestion
          // minimize might leave us with an empty object, which would
          // lead to MongoDB throwing a "Update document requires atomic operators" error
          if (Object.keys(updateOp[key]).length === 0) {
            updateOp[key] = null;
          }
```

unless the semantics changed, it would likely be good to keep the comment from before the PR for this case

hasezoey (commented):
Looks good to me

lib/model.js
vkarpov15: Semantics changed. Before, we could minimize `{ $set: { sub: {} } }` down to an empty object `{}`, and MongoDB server throws an error if you try to send `updateOne()` with empty object as `update`. With this change, `{ $set: { sub: {} } }` just minimizes down to `{ $set: { sub: null } }`, which is perfectly fine.

Issue #14418, opened 2024-03-06
Populated documents are flattened to ObjectIds when assigned to another document


neubee:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.12.6

### Node.js version

16.20.2

### MongoDB server version

5.0.25

### Description

I'm having a problem where Mongoose will "flatten" documents to just an ObjectId when assigned to other documents. It's been difficult to narrow down when exactly this behavior occurs, but I do know that it will happen when populating an array and then assigning one, or all, of those populated documents to an array in another document. There could be other cases were this happens as well that I haven't found.

I'm not certain if this is a bug or expected behavior, but this is causing a lot of trouble in my project. Also, while I've always had trouble with this type of Mongoose behavior it started happening in new areas of my code when I updated from Mongoose 5.13.20 to 6.12.6

If this is expected behavior, how can I stop it from happening? Right now I need to resort to keeping/passing around the populated version of the documents or I need to populate them again after assigning them to the second parent.

### Steps to Reproduce

```js
const mongoose = require("mongoose");
const { MongoMemoryServer } = require("mongodb-memory-server");

(async () => {
  const mongoServer = await MongoMemoryServer.create();
  await mongoose.connect(mongoServer.getUri(), { dbName: "test" });

  const cartSchema = new mongoose.Schema({
    products: [
      {
        product: {
          type: mongoose.Schema.Types.ObjectId,
          ref: "Product",
        },
        quantity: Number,
      },
    ],
    singleProduct: {
      type: mongoose.Schema.Types.ObjectId,
      ref: "Product",
    },
  });
  const purchaseSchema = new mongoose.Schema({
    products: [
      {
        product: {
          type: mongoose.Schema.Types.ObjectId,
          ref: "Product",
        },
        quantity: Number,
      },
    ],
    singleProduct: {
      type: mongoose.Schema.Types.ObjectId,
      ref: "Product",
    },
  });

  const productSchema = new mongoose.Schema({
    name: String,
  });

  const Cart = mongoose.model("Cart", cartSchema);
  const Purchase = mongoose.model("Purchase", purchaseSchema);
  const Product = mongoose.model("Product", productSchema);

  const dbProduct = new Product({
    name: "Bug",
  });
  await dbProduct.save();

  const dbCart = new Cart({
    products: [
      {
        product: dbProduct,
        quantity: 2,
      },
    ],
[...]

PR #14458, opened 2024-03-21
Avoid depopulating populated subdocs underneath document arrays when copying to another document

vkarpov15:
**Summary**

#14418 is a tricky issue, there's a couple of causes:

1. [This `toObject()` call](https://github.com/Automattic/mongoose/blob/d516f50f2073f69c62cabd6bf0e765e08abb0208/lib/schema/documentArray.js#L456) swallows up the fact that you're setting the doc array element to a document that potentially includes populated values
2. [This check fails](https://github.com/Automattic/mongoose/blob/d516f50f2073f69c62cabd6bf0e765e08abb0208/lib/document.js#L1381) because `$__.wasPopulated` was `true` rather than an object with `value`

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1087,7 +1087,11 @@ Document.prototype.$set = function $set(path, val, type, options) {
       if (path.$__isNested) {
         path = path.toObject();
       } else {
-        path = path._doc;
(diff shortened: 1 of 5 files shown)


hasezoey (commented):
Looks good to me

PR #14461, opened 2024-03-23
Removed unused hook from docs

bernardarhia:
The update method was removed in v7.2 along with the middleware/hook it comes with. However the current v8 docs still shows that udpate hook is supported, which is invalid.

I had an issue after upgrading from v6 - v8 with the update hook. The docs misled me and it took me about 5hours before I got to know that update is no longer supported in v8.


vkarpov15 (approved):
Thanks :+1:

PR #14463, opened 2024-03-23
fix(types/query): bring "getFilter" and "getQuery" in-line with "find" and other types

hasezoey:
**Summary**

This PR changes `Query.getFilter` and `Query.getQuery` to use `RawDocType` again, as this is the same argument as in `.find`'s `filter` parameter.
In addition also pass-through the `QueryHelpers` and `DocType` in `SchemaOptions`'s `query`'s object (otherwise there is a test failure).

This change from e3f320515297240f6bca4ec6be7b14c442a97a46 somehow made it so that `PopulatedDoc<Document, RefType>` could not be narrowed down to the used `RefType` anymore, and i dont know why, but this change fixed it again for typegoose.

### DIFF
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -493,7 +493,7 @@ declare module 'mongoose' {
     get(path: string): any;
 
     /** Returns the current query filter (also known as conditions) as a POJO. */
-    getFilter(): FilterQuery<DocType>;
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Nice find, thanks :+1:

PR #14464, opened 2024-03-25
types(schema): re-export the defintion for `SearchIndexDescription`

noseworthy:
**Summary**

I'm trying to add support for creating search indices to `typegoose` here: https://github.com/typegoose/typegoose/pull/921

Since `typegoose` doesn't depend on or import `mongodb` I've had to create a copy of the `SearchIndexDescription` type from `mongodb` that is going to become out of date if the type in `mongodb` changes. I thought it'd be better if `mongoose` could re-export the type so that `typegoose`, `mongoose`, and `mongodb` can all be in sync.

**Examples**

N/A

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -279,7 +279,7 @@ declare module 'mongoose' {
      *
      * @remarks Search indexes are only supported when used against a 7.0+ Mongo Atlas cluster.
      */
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #14468, opened 2024-03-26
fix(document): handle virtuals that are stored as objects but getter returns string with toJSON

### LINKED ISSUE #14446: Recursion check in toObject breaks mongoose-intl plugin
### Prerequisites

### Mongoose version

>=6.3.5

### Node.js version

>=18

### MongoDB server version

7

_No response_

### Description

As of Mongoose v6.3.5, the [mongoose-intl](https://github.com/alexsk/mongoose-intl) plugin does not work anymore when used in subdocument arrays. Calling `toJSON` or `toObject` with getters enabled on such objects results in an error.

In order to narrow down the source of the problem, I recreated what the plugin is doing in a standalone [gist](https://gist.github.com/yelworc/af6a199a30e4dc0cb5cbb92b53729711). The behavior and the resulting error is the same as my actual issue with mongoose-intl in a closed-source production system.

The changes in commit [8325e563](https://github.com/Automattic/mongoose/commit/8325e563) are what introduced the problem; specifically, the recursion checking `seen` map will cache and return subdocuments where the getters for localized string fields have already been applied (e.g. `{field: 'some text'}`), but then the toObject mechanism tries to clone the single-language schema paths (like `field.en`), which conflicts with the already present `field` value.

Now, I realize that mongoose-intl is doing something fishy here with its virtuals that Mongoose wouldn't normally allow (having a "real" `field.xyz` schema path and a `field` virtual) :smile:. However, it might be relatively easy to work around; changing

```js
branch[part] = clone(val, options);
```

to

```js
if (utils.isObject(branch)) branch[part] = clone(val, options);
```

in `document/applyGetters` fixes the problem (and a similar change might be necessary in `document/applyVirtuals`) – but of course that might entail unintended side effects.

If you consider this a bug (resp. invalid behavior) in mongoose-intl, I'd be grateful for any ideas for fixing it there; preferably in ways that wouldn't require a DB migration :smile:

### Steps to Reproduce

See the [gist](https://gist.github.com/yelworc/af6a199a30e4dc0cb5cbb92b53729711); if it helps, I could also provide sample code that is using mongoose-intl.

### Expected Behavior

_No response_

vkarpov15:
Fix #14446

**Summary**

#14446 points out that we continue to loop through and try to create properties when getting nested paths underneath a virtual even if the virtual returned a non-object. Since this will always throw an error, we should just avoid that and assume this is just the virtual returning a non-object by design.

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -12538,6 +12538,38 @@ describe('document', function() {
     assert.equal(purchaseFromDbCart.products[0].product.name, 'Bug');
     assert.equal(purchaseFromDbCart.singleProduct.name, 'Bug');
   });
+
+  it('handles virtuals that are stored as objects but getter returns string with toJSON (gh-14446)', async function() {
+    const childSchema = new mongoose.Schema();
+
+    childSchema.virtual('name')
+      .set(function(values) {
+        for (const [lang, val] of Object.entries(values)) {
+          this.set(`name.${lang}`, val);
+        }
+      })
+      .get(function() {
+        return this.$__getValue(`name.${this.lang}`);
+      });
+
+    childSchema.add({ name: { en: { type: String }, de: { type: String } } });
+
+    const ChildModel = db.model('Child', childSchema);
+    const ParentModel = db.model('Parent', new mongoose.Schema({
+      children: [childSchema]
+    }));
+
+    const child = await ChildModel.create({ name: { en: 'Stephen', de: 'Stefan' } });
+    child.lang = 'en';
+    assert.equal(child.name, 'Stephen');
+
+    const parent = await ParentModel.create({
+      children: [{ name: { en: 'Stephen', de: 'Stefan' } }]
+    });
+    parent.children[0].lang = 'de';
+    const obj = parent.toJSON({ getters: true });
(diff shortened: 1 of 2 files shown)


test/document.test.js
```diff
@@ -12538,6 +12538,38 @@ describe('document', function() {
...
+
+    childSchema.add({ name: { en: { type: String }, de: { type: String } } });
+
+    const ChildModel = db.model('Child', childSchema);
+    const ParentModel = db.model('Parent', new mongoose.Schema({
+      children: [childSchema]
+    }));
+
+    const child = await ChildModel.create({ name: { en: 'Stephen', de: 'Stefan' } });
+    child.lang = 'en';
```
hasezoey, line 12563: from what i can tell `lang` is not actually part of the schema, is this required for this issue or should it be added?

hasezoey (commented):
Looks good to me, aside from a slight test inconsistency

test/document.test.js
vkarpov15: I think that's by design, it's meant to be a property that isn't saved to the db. Could make it into a virtual, but I don't think that's strictly necessary for this test.

PR #14500, opened 2024-04-04
fix(model): make `Model.recompileSchema()` also re-apply discriminators

### LINKED ISSUE #14444: Recompiling schema does not reapply discriminators
### Prerequisites

### Mongoose version

8.2.2

### Node.js version

18.16

### MongoDB server version

6.3.0

_No response_

### Description

I have a use case where I define a schema that uses a discriminated schema in a base package. I want to enable dependent packages to add their own implementations of the discriminated schema. However, it seems that once the model that uses the schema is compiled, adding further discriminated schemas are not applied.

### Steps to Reproduce

```javascript
'use strict';

const { Schema } = mongoose;

// Define discriminated schema
const decoratorSchema = new Schema({

class Decorator {
    type;
    whoAmI() { return "I am BaseDeco"; }
}
decoratorSchema.loadClass(Decorator);

// Define discriminated class before model is compiled
class Deco1 extends Decorator { whoAmI() { return "I am Deco1"; }};
const deco1Schema = new Schema({}).loadClass(Deco1);
deco1Schema.loadClass(Deco1);
decoratorSchema.discriminator("Deco1", deco1Schema);

// Define model that uses discriminated schema
const shopSchema = new Schema({
    item: { type: decoratorSchema, required: true }
});

class Shop {};
shopSchema.loadClass(Shop);
const shopModel = mongoose.model("shop", shopSchema);

// Define another discriminated class after the model is compiled
class Deco2 extends Decorator { whoAmI() { return "I am Deco2"; }};
const deco2Schema = new Schema({}).loadClass(Deco2);
deco2Schema.loadClass(Deco2);
decoratorSchema.discriminator("Deco2", deco2Schema);

// Shouldn't recompiling the schema fix this?
shopModel.recompileSchema();

void async function main() 
{
    let instance = new shopModel({ item: {type: "Deco1"} }); // "I am Deco1" - Works
    console.log(instance.item.whoAmI());

    instance = new shopModel({ item: {type: "Deco2"} }); // "I am BaseDeco" - Should be Deco2
    console.log(instance.item.whoAmI());
}();
```

### Expected Behavior

After recompiling a schema, it should also apply new discriminators of its fields.

vkarpov15:
Fix #14444

**Summary**

Right now `recompileSchema()` applies most schema changes, but not additional `Schema.prototype.discriminator()` or `SchemaType.prototype.discriminator()` calls. This PR makes `recompileSchema()` go through and reapply every discriminator that was added to the schema.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -23,6 +23,7 @@ const VersionError = require('./error/version');
 const ParallelSaveError = require('./error/parallelSave');
 const applyDefaultsHelper = require('./helpers/document/applyDefaults');
 const applyDefaultsToPOJO = require('./helpers/model/applyDefaultsToPOJO');
+const applyEmbeddedDiscriminators = require('./helpers/discriminator/applyEmbeddedDiscriminators');
 const applyHooks = require('./helpers/model/applyHooks');
 const applyMethods = require('./helpers/model/applyMethods');
 const applyProjection = require('./helpers/projection/applyProjection');
@@ -5001,6 +5002,15 @@ Model.__subclass = function subclass(conn, schema, collection) {
 
 Model.recompileSchema = function recompileSchema() {
   this.prototype.$__setSchema(this.schema);
+
+  if (this.schema._applyDiscriminators != null) {
+    for (const disc of this.schema._applyDiscriminators.keys()) {
+      this.discriminator(disc, this.schema._applyDiscriminators.get(disc));
+    }
+  }
+
+  const overwriteExisting = true;
+  applyEmbeddedDiscriminators(this.schema, new WeakSet(), overwriteExisting);
 };
 
 /**
--- a/lib/model.js
+++ b/lib/model.js
@@ -5009,8 +5009,7 @@ Model.recompileSchema = function recompileSchema() {
     }
   }
 
-  const overwriteExisting = true;
-  applyEmbeddedDiscriminators(this.schema, new WeakSet(), overwriteExisting);
+  applyEmbeddedDiscriminators(this.schema, new WeakSet(), true);
 };
 
 /**
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -7419,6 +7419,47 @@ describe('Model', function() {
     assert.equal(doc.myVirtual, 'Hello from myVirtual');
   });
 
+  it('supports recompiling model with new discriminators (gh-14444) (gh-14296)', function() {
+    // Define discriminated schema
+    const decoratorSchema = new Schema({
+      type: { type: String, required: true }
+    }, { discriminatorKey: 'type' });
+
+    class Decorator {
+      whoAmI() { return 'I am BaseDeco'; }
+    }
+    decoratorSchema.loadClass(Decorator);
+
+    // Define discriminated class before model is compiled
+    class Deco1 extends Decorator { whoAmI() { return 'I am Test1'; }}
+    const deco1Schema = new Schema({}).loadClass(Deco1);
+    deco1Schema.loadClass(Deco1);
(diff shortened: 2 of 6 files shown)


lib/model.js
```diff
@@ -5001,6 +5002,15 @@ Model.__subclass = function subclass(conn, schema, collection) {
...
   this.prototype.$__setSchema(this.schema);
+
+  if (this.schema._applyDiscriminators != null) {
+    for (const disc of this.schema._applyDiscriminators.keys()) {
+      this.discriminator(disc, this.schema._applyDiscriminators.get(disc));
+    }
+  }
+
+  const overwriteExisting = true;
+  applyEmbeddedDiscriminators(this.schema, new WeakSet(), overwriteExisting);
```
hasezoey, line 5013: and specific reason on why `overwriteExisting` is not inlined?

test/model.test.js
```diff
@@ -7419,6 +7419,47 @@ describe('Model', function() {
...
+    deco1Schema.loadClass(Deco1);
+    decoratorSchema.discriminator('Test1', deco1Schema);
+
+    // Define model that uses discriminated schema
+    const shopSchema = new Schema({
+      item: { type: decoratorSchema, required: true }
+    });
+
+    class Shop {}
+    shopSchema.loadClass(Shop);
```
hasezoey, line 7445: this class seems unnecessary, is this just here to keep consistent with the other schemas?

test/model.test.js
```diff
@@ -7419,6 +7419,47 @@ describe('Model', function() {
...
+
+    class Decorator {
+      whoAmI() { return 'I am BaseDeco'; }
+    }
+    decoratorSchema.loadClass(Decorator);
+
+    // Define discriminated class before model is compiled
+    class Deco1 extends Decorator { whoAmI() { return 'I am Test1'; }}
+    const deco1Schema = new Schema({}).loadClass(Deco1);
+    deco1Schema.loadClass(Deco1);
```
hasezoey, line 7436:
```suggestion
    const deco1Schema = new Schema({});
    deco1Schema.loadClass(Deco1);
```

test/model.test.js
```diff
@@ -7419,6 +7419,47 @@ describe('Model', function() {
...
+    });
+
+    class Shop {}
+    shopSchema.loadClass(Shop);
+    const shopModel = db.model('Test', shopSchema);
+
+    // Define another discriminated class after the model is compiled
+    class Deco2 extends Decorator { whoAmI() { return 'I am Test2'; }}
+    const deco2Schema = new Schema({}).loadClass(Deco2);
+    deco2Schema.loadClass(Deco2);
```
hasezoey, line 7451:
```suggestion
    const deco2Schema = new Schema({});
    deco2Schema.loadClass(Deco2);
```

test/model.test.js
```diff
@@ -7419,6 +7419,47 @@ describe('Model', function() {
...
+    decoratorSchema.discriminator('Test2', deco2Schema);
+
+    shopModel.recompileSchema();
```
hasezoey, line 7454: shouldnt there be a check before recompiling to test that the second is not actually applied yet?

hasezoey (changes requested):
Looks good, though some test nitpicks

lib/model.js
vkarpov15: I was thinking about doing this for readability, but given that there's only 1 boolean param I don't think it helps readability much. I removed that in [1e8753f](https://github.com/Automattic/mongoose/pull/14500/commits/1e8753fa6a94fc8649ad5cb27b5247d4d4f3c459)

Issue #14462, opened 2024-03-23
Types break in semver patch change


Anonymous4078:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.2.3

### Node.js version

20.11.1

### MongoDB server version

5.x

### Typescript version (if applicable)

5.4.3

### Description

```yaml
› yarn build
Error: src/listeners/message/messageBulkDelete.ts(28,49): error TS2769: No overload matches this call.
  Overload [1](https://github.com/Anonymous4078/suggester/actions/runs/8382964808/job/22957771088?pr=327#step:6:1) of [4](https://github.com/Anonymous4078/suggester/actions/runs/8382964808/job/22957771088?pr=327#step:6:5), '(filter?: FilterQuery<GuildConfig> | undefined, projection?: ProjectionType<GuildConfig> | null | undefined, options?: QueryOptions<...> | ... 1 more ... | undefined): Query<...>', gave the following error.
    Type 'string | null' is not assignable to type 'Condition<string> | undefined'.
      Type 'null' is not assignable to type 'Condition<string> | undefined'.
  Overload 2 of 4, '(filter?: FilterQuery<GuildConfig> | undefined, projection?: ProjectionType<GuildConfig> | null | undefined): Query<...>', gave the following error.
    Type 'string | null' is not assignable to type 'Condition<string> | undefined'.
  Overload 3 of 4, '(filter?: FilterQuery<GuildConfig> | undefined): Query<(Document<unknown, {}, GuildConfig> & GuildConfig & { _id: ObjectId; }) | null, Document<...> & ... 1 more ... & { ...; }, {}, GuildConfig, "findOne">', gave the following error.
    Type 'string | null' is not assignable to type 'Condition<string> | undefined'.
Error: src/listeners/message/messageBulkDelete.ts(32,1[5](https://github.com/Anonymous4078/suggester/actions/runs/8382964808/job/22957771088?pr=327#step:6:6)): error TS2322: Type 'string | null' is not assignable to type 'Condition<string> | undefined'.
Error: src/listeners/message/messageDelete.ts(21,45): error TS2769: No overload matches this call.
  Overload 1 of 4, '(filter?: FilterQuery<GuildConfig> | undefined, projection?: ProjectionType<GuildConfig> | null | undefined, options?: QueryOptions<...> | ... 1 more ... | undefined): Query<...>', gave the following error.
[...]

Issue #14459, opened 2024-03-22
FilterQuery Typing fails for generic type


david-loe:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.2.3

### Node.js version

20

### MongoDB server version

7.0.6

### Typescript version (if applicable)

5.4.3

### Description

Repro Script throws
```
No overload matches this call.
  Overload 1 of 5, '(filter: FilterQuery<ModelType>, projection?: ProjectionType<ModelType> | null | undefined, options?: QueryOptions<ModelType> | null | undefined): Query<...>', gave the following error.
    Argument of type '{ _id: string | Types.ObjectId; }' is not assignable to parameter of type 'FilterQuery<ModelType>'.
      Type '{ _id: string | Types.ObjectId; }' is not assignable to type '{ [P in keyof ModelType]?: Condition<ModelType[P]> | undefined; }'.
  Overload 2 of 5, '(filter: FilterQuery<ModelType>, projection?: ProjectionType<ModelType> | null | undefined): Query<IfAny<ModelType, any, Document<...> & Require_id<...>>[], IfAny<...>, {}, ModelType, "find">', gave the following error.
    Argument of type '{ _id: string | Types.ObjectId; }' is not assignable to parameter of type 'FilterQuery<ModelType>'.
      Type '{ _id: string | Types.ObjectId; }' is not assignable to type '{ [P in keyof ModelType]?: Condition<ModelType[P]> | undefined; }'.
  Overload 3 of 5, '(filter: FilterQuery<ModelType>): Query<IfAny<ModelType, any, Document<unknown, {}, ModelType> & Require_id<ModelType>>[], IfAny<...>, {}, ModelType, "find">', gave the following error.
    Argument of type '{ _id: string | Types.ObjectId; }' is not assignable to parameter of type 'FilterQuery<ModelType>'.
      Type '{ _id: string | Types.ObjectId; }' is not assignable to type '{ [P in keyof ModelType]?: Condition<ModelType[P]> | undefined; }'.ts(2769)
```

### Steps to Reproduce

```typescript
import { Model, Types } from "mongoose";

function findById<ModelType extends {_id: Types.ObjectId | string}>(model: Model<ModelType>, _id: Types.ObjectId | string){
    return model.find({_id: _id})
}
```

### Expected Behavior

No Error

david-loe:
Error comes from v8.2.3.
[...]

PR #14510, opened 2024-04-08
types(query): make `FilterQuery` props resolve to `any` for generics support

### LINKED ISSUE #14473: Cannot use FilterQuery<Type> when `Type` is a class
### Prerequisites

### Mongoose version

8.2.3

### Node.js version

20

### MongoDB server version

7

^5

### Description

When I create a generic function that receive a generic type `D` and D extends from `AbstractSchema` then I could not use `FilterQuery<D>` with any type

### Steps to Reproduce

```
export class AbstractSchema {
  _id: any
  createdAt: Date
  updatedAt: Date
  deletedAt: Date 
}
export const generateExists = <D extends AbstractSchema = AbstractSchema>() => {
  const query : FilterQuery<D>= {deletedAt : {$ne : null}} // or {deletedAt : new Date()}
  //Typescript show error at query Type 'null' is not assignable to type 'ApplyBasicQueryCasting<QueryTypeCasting<D["deletedAt"]>> | undefined'
}
```

### Expected Behavior

_No response_

vkarpov15:
Fix #14473
Re: #14459

Tagging @FaizBShah @sderrow  re: #14436, #14397

**Summary**

It looks like the major reason for `FilterQuery<T>` setting all props to `any` is because TypeScript disallows setting an object with a concrete type to an abstract type. This includes generic function parameters, like calling a function with `FilterQuery<>` as a parameter where `FilterQuery<>` depends on the generic. For example:

```ts
function findById<ModelType extends {_id: Types.ObjectId | string}>(model: Model<ModelType>, _id: Types.ObjectId | string){
    // Will not compile without `as FilterQuery<ModelType>`, even if `FilterQuery<T> = T` because `FilterQuery`
    // depends on generic type `ModelType`
    return model.find({_id: _id});
}
```

From [this SO answer](https://stackoverflow.com/questions/56505560/how-to-fix-ts2322-could-be-instantiated-with-a-different-subtype-of-constraint): "Never assign a concrete type to a generic type parameter, consider it as read-only!". That includes calling a function with arguments that rely on the generic type parameter. This restriction makes sense because of `never`

```ts
type ModelType = { _id: never };
const TestModel = model<ModelType>('test', new Schema());

// Compiles fine, apparently `{ _id: never } extends { _id: Types.ObjectId | string }`
findById<ModelType>(TestModel, 'foo');
```

This whole "generic function for wrapping Mongoose model functions" pattern isn't one that I would personally use, but it seems to be sufficiently common that we've gotten 3 distinct bug reports related to this change #14459, #14473, #14462

**Examples**

### DIFF
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -12,7 +12,6 @@ import {
   FilterQuery,
   UpdateQuery,
   UpdateQueryKnownOnly,
-  ApplyBasicQueryCasting,
   QuerySelector,
   InferSchemaType,
   ProjectionFields,
@@ -325,7 +324,7 @@ function gh11964() {
 }
 
 function gh14397() {
-  type Condition<T> = ApplyBasicQueryCasting<T> | QuerySelector<ApplyBasicQueryCasting<T>>; // redefined here because it's not exported by mongoose
+  type Condition<T> = T | QuerySelector<T>; // redefined here because it's not exported by mongoose
 
   type WithId<T extends object> = T & { id: string };
 
@@ -591,4 +590,25 @@ function mongooseQueryOptions() {
     lean: true,
     populate: 'test'
   });
+}
+
+function gh14473() {
+  class AbstractSchema {
+    _id: any;
+    createdAt: Date;
+    updatedAt: Date;
+    deletedAt: Date;
+
+    constructor() {
+      this._id = 4;
+      this.createdAt = new Date();
+      this.updatedAt = new Date();
+      this.deletedAt = new Date();
+    }
+  }
+
+  const generateExists = <D extends AbstractSchema = AbstractSchema>() => {
+    const query: FilterQuery<D> = { deletedAt: { $ne: null } };
+    const query2: FilterQuery<D> = { deletedAt: { $lt: new Date() } };
+  };
 }
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -1,24 +1,7 @@
 declare module 'mongoose' {
   import mongodb = require('mongodb');
 
-  type StringQueryTypeCasting = string | RegExp;
-  type ObjectIdQueryTypeCasting = Types.ObjectId | string;
-  type UUIDQueryTypeCasting = Types.UUID | string;
-
-  type QueryTypeCasting<T> = T extends string
-    ? StringQueryTypeCasting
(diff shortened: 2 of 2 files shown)


FaizBShah:
@vkarpov15 It seems that its a very common pattern which is followed by a lot of devs, so in that case I don't think there's much to be done in this case other than making `FilterQuery<>` handle casting to `any` type. I'll try to think if there is some way we can improve the typing of this type without compromising the pattern in future, but for now this PR looks good to me. LGTM!!

hasezoey (approved):
Looks good to me, though asking because i dont fully understand this back-and-forth of the `FilterQuery` type (i have only skimmed the issues / PRs):

the `Condition` type has been simplified and now does less checking / work to support more general cases because of typescript weirdness, while still trying to avoid the outcome of `any`?

vkarpov15:
Less checking to allow defining generic functions like the following. Any type checking in `FilterQuery` for prop values causes the following to fail to compile.

```ts
function findById<ModelType extends {_id: Types.ObjectId | string}>(model: Model<ModelType>, _id: Types.ObjectId | string){
    return model.find({_id: _id});
}
```

Evertt:
Hi @vkarpov15,

I'm a bit puzzled by the Condition<T> type definition. I understand that the goal is to provide flexibility, but I'm concerned that including any in the union type might compromise type safety.

From my understanding of TypeScript, when any is included in a union type, it effectively makes the entire type equivalent to any. This means that the type checker can't provide the usual warnings and autocompletion.

I'd love to see a more precise type definition that balances flexibility with type safety. Perhaps we could explore alternative approaches that would allow TypeScript to warn developers about potential errors, like the ones in this example code:

```ts
MyModel.find({
  // This contains a typo and so will not affect the outcome of the query
  // but it would be nice if TypeScript could warn me about that.
  createdAtt: { $lt: new Date() },

  // And this will even cause an error when trying to execute the query at runtime
  // and again, I get no warning from TypeScript about this.
  createdAt: () => {
    return console.log('createdAt is a date and so this query filter will result in an error')
  },
})
```

I've spent countless hours refining my types to get exactly the behavior I need, and I appreciate the importance of having a robust type system. I'm confident that with some creative problem-solving, we can find a solution that meets our goals.

Are you open to maybe collaborate on finding a middle-ground that we'd both be happy with?

PR #14513, opened 2024-04-09
types(models): fix incorrect bulk write options

emiljanitzek:
**Summary**
The `bulkWrite` options are incorrectly typed. `timestamp` does not exists as an option on `updateOne`/`updateMany` (there was a test but it didn't work, I fixed this as well). And the type incorrectly showed a timestamp as options to `bulkWrite`, which doesn't work. 

Also extended the mongodb `BulkWriteOptions` making it easier to use the `MongooseBulkWriteOption` without the need to also always extend mongodb.

**Examples**

See fixed test

### DIFF
--- a/test/types/middleware.test.ts
+++ b/test/types/middleware.test.ts
@@ -1,6 +1,6 @@
-import { Schema, model, Model, Document, SaveOptions, Query, Aggregate, HydratedDocument, PreSaveMiddlewareFunction, ModifyResult } from 'mongoose';
+import { Schema, model, Model, Document, SaveOptions, Query, Aggregate, HydratedDocument, PreSaveMiddlewareFunction, ModifyResult, AnyBulkWriteOperation } from 'mongoose';
 import { expectError, expectType, expectNotType, expectAssignable } from 'tsd';
-import { AnyBulkWriteOperation, CreateCollectionOptions } from 'mongodb';
+import { CreateCollectionOptions } from 'mongodb';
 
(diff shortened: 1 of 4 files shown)


vkarpov15 (approved):
Thanks :+1:

emiljanitzek:
> @emiljanitzek @vkarpov15 FYI the `timestamps` option at the `bulkWrite` level is actually valid, but it's only applied for `insertOne`s, as seen [here](https://github.com/Automattic/mongoose/blob/78bbcb52e9551f56da64de12437485ccb025f1cb/lib/helpers/model/castBulkWrite.js#L32).
> 
> Should I open a PR to revert that particular piece of this PR?
> 
> Longer-term, probably makes sense to be consistent and remove `timestamps` from the outer options level and move it to the `insertOne` level, so it mirrors the structure of `updateOne` and `updateMany`. But that would be a breaking change, so maybe for 9.0?

Seems very dangerous to have a function wide option that only applies sometimes. It does make sense as an option but then it should be respected for updateOne/updateMany as well. Or rename the option to `insertTimestamps` to avoid confusion and unexpected behaviors.

sderrow:
I also wonder if the same option should be included for `replaceOne`, and implement similar behavior?

vkarpov15:
@sderrow you're right, we should support the `timestamps` option for all bulk write elements. If you're willing to put in a separate PR for supporting that, that would be great. I'll open a separate issue, and we'll get to it in the next couple of weeks if you don't get to it first.

PR #14519, opened 2024-04-11
Add `Model.listSearchIndexes()`

### LINKED ISSUE #14450: Add `Model.listSearchIndexes()`
### Prerequisites

### Mongoose version

8.2.2

### Node.js version

16.20.2

### MongoDB server version

Atlas latest

_No response_

### Description

When I use model.createSearchIndex() with an index that already exists, my node application is going down with a SIGTERM

### Steps to Reproduce

Use an existing test case to create a search index, but attempt to create it twice with the same name 
```
//just there 
const vectorModelsSupported = [
		{
			name: 'text_embedding_3_small',
			size: 1536,
			enabled: true
		},
		{ 
			name: 'text_embedding_3_small', 
			size: 1536,
			enabled: true
		}
	]

const EmbeddingSchema = new Schema({
	chunkText: { type: String, maxLength: 32764, default: "" },
	embedding_text_embedding_3_small: { type: Array, maxLength: 1536 },
	embedding_text_embedding_3_large: { type: Array, },
	embedding_text_embedding_ada_002: { type: Array, maxLength: 1536 },
	chunkLen: Number,
	chunkIndex: Number,
	upload: {
		type: ObjectId,
		ref: "upload",
	},
	chunkUrl: { type: String, maxLength: 1000, }, /*ex: #page=2 for PDF or #id for HTML*/
	created: { type: Date, default: Date.now },
	usage: { type: Number },
	status: { type: String, default: "inactive" }, // inactive, active | error, selected, deleted
})

const createSearchIndexes = async () => { 
	
	//debug("createSearchIndexes", "This is the list of indexes:", await embedding.listIndexes())
	debug("createSearchIndexes", "This is the list of indexes:", await embedding.searchIndexes)
	if (!modelsSupported) return;
	
	modelsSupported.forEach((model) =>{
		if (model.enabled){
			
			//const field =
			
			const fields = {}
			fields[`embedding_${model.name}`]=  [
				{
					"dimensions": model.size, // 1536,
					"similarity": "euclidean",
					"type": "knnVector"
				}
			]

			try{
				embedding.createSearchIndex({
					name: model.name, // "text_embedding_3_small",
					definition: {
						"mappings": {
							"fields": fields
							}
						}
					}
				)
			}
			catch (err){
				debug("createSearchIndexes", "Failed to create index for model ",model, err)
			}
		
		}

	})
}

embedding.on('index',(event) => {
	debug("Embedding", "===> This is an index event", event)
	
	createSearchIndexes()

})
```
### Expected Behavior

the createSearchIndex should throw an error I can catch, or there should be an API that returns the search index. 

The Model.listIndexes does not return the search indexes.

vkarpov15:
**Summary**

Fix #14450 . Looks like we missed a spot when implementing the search indexes API

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1655,6 +1655,15 @@ Model.dropSearchIndex = async function dropSearchIndex(name) {
   return await this.$__collection.dropSearchIndex(name);
 };
 
+
+Model.listSearchIndexes = async function listSearchIndexes() {
+  _checkContext(this, 'listSearchIndexes');
+
+  const cursor = await this.$__collection.listSearchIndexes();
+
+  return await cursor.toArray();
+}
+
 /**
  * Does a dry-run of `Model.syncIndexes()`, returning the indexes that `syncIndexes()` would drop and create if you were to run `syncIndexes()`.
  *
--- a/lib/model.js
+++ b/lib/model.js
@@ -1662,7 +1662,7 @@ Model.listSearchIndexes = async function listSearchIndexes() {
   const cursor = await this.$__collection.listSearchIndexes();
 
   return await cursor.toArray();
-}
+};
 
 /**
  * Does a dry-run of `Model.syncIndexes()`, returning the indexes that `syncIndexes()` would drop and create if you were to run `syncIndexes()`.
--- a/lib/model.js
+++ b/lib/model.js
@@ -1655,11 +1655,27 @@ Model.dropSearchIndex = async function dropSearchIndex(name) {
   return await this.$__collection.dropSearchIndex(name);
 };
 
+/**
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -1655,6 +1655,15 @@ Model.dropSearchIndex = async function dropSearchIndex(name) {
   return await this.$__collection.dropSearchIndex(name);
 };
 
+
+Model.listSearchIndexes = async function listSearchIndexes() {
```
hasezoey, line 1659: i think there should be jsdoc so that it gets added to the documentation and also typescript types.
i dont think tests can be easily added as search indexes are atlas-exclusive (to my knowledge)

hasezoey (changes requested):

Issue #14502, opened 2024-04-05
Inconsistent behaviour of  isDirectModified and ignoreAtomics 


khanguyen-unity:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.6.9

### Node.js version

20.11.0

### MongoDB server version

5.0

### Typescript version (if applicable)

4.9.5

## `{ ignoreAtomics: true }`

The `{ ignoreAtomics: true }` option was introduced in Mongoose **6.12.5** as a solution for issue #14024. The intended behavior is that, when using `{ ignoreAtomics: true }`, the `isModified` method should return `false` for elements not affected by atomic operators such as `push` and `pull`. While I've confirmed that this option functions correctly with `push`, its behavior appears inconsistent with the `pull` operator.

When I pull `comment1` or `comment2` out of the `comments` array, all other elements are modified.

```typescript
blogPost.init({
  title: 'Test',
  slug: 'test',
  comments: [comment1, comment2, comment3],
});

console.log('Is comment1.title modified: ', blogPost.comments[0].isModified('title')); // -> False (expected)
console.log('Is comment3.title modified: ', blogPost.comments[2].isModified('title')); // -> False (expected)

blogPost.comments.pull({ _id: comment2._id });
// Now there are 2 elements in blogPost.comments
console.log(
  'Is comment1.title modified after pull: ',
  blogPost.comments[0].isModified('title', { ignoreAtomics: true }),
); // -> True, expected False
console.log(
  'Is comment3.title modified after pull: ',
  blogPost.comments[1].isModified('title', { ignoreAtomics: true }),
); // -> True, expected False
```
But when I pull `comment3` out of the array, the rest of the array are left unmodified.

```typescript
blogPost.init({
  title: 'Test',
  slug: 'test',
  comments: [comment1, comment2, comment3],
});

console.log('Is comment1.title modified: ', blogPost.comments[0].isModified('title')); // -> False (expected)
console.log('Is comment3.title modified: ', blogPost.comments[2].isModified('title')); // -> False (expected)

blogPost.comments.pull({ _id: comment3._id });
// Now there are 2 elements in blogPost.comments
console.log(
  'Is comment1.title modified after pull: ',
  blogPost.comments[0].isModified('title', { ignoreAtomics: true }),
); // -> False (expected)
console.log(
  'Is comment3.title modified after pull: ',
[...]

PR #14531, opened 2024-04-22
fix(array): avoid converting to `$set` when calling `pull()` on an element in the middle of the array

vkarpov15:
Re: #14502

**Summary**

Part of #14502 is that `pull()` uses the Mongoose array, rather than the internal raw array, for making updates, which mixes up change tracking.

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -13055,6 +13055,64 @@ describe('document', function() {
     const obj = parent.toJSON({ getters: true });
     assert.equal(obj.children[0].name, 'Stefan');
   });
+
+  it('isDirectModified on paths underneath direct modified subdoc (gh-14502)', async function() {
+    const JsonFieldSchema = new Schema({
+      fieldA: String,
+      fieldB: String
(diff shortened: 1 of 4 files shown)


test/document.test.js
```diff
@@ -13055,6 +13055,64 @@ describe('document', function() {
...
+      body: String,
+      jsonField: JsonFieldSchema
+    });
+
+    const blogPostSchema = new Schema({
+      comment: CommentSchema
+    });
+
+    const Comments = db.model('Comments', CommentSchema);
+    const BlogPost = db.model('BlogPost', blogPostSchema);
```
hasezoey, line 13076:
```suggestion
    const BlogPostSchema = new Schema({
      comment: CommentSchema
    });

    const Comments = db.model('Comments', CommentSchema);
    const BlogPost = db.model('BlogPost', BlogPostSchema);
```

some consistency

hasezoey (approved):

Issue #14527, opened 2024-04-16
schema.recompileSchema exception when subDocument discriminator is defined before the model is compiled


ZachLeviPixel:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.3.1

### Node.js version

18.16

### MongoDB server version

6.3.0

### Description

The following example throws "discriminator with name 'X' already exists" when the model of the top level schema is compiled after the discriminator definitions have been created on the subdocument schema.

If you create the model first, recompile seems to work as expected.
Important to note that on either case, the schema behaves as expected without recompiling.

### Steps to Reproduce

```javascript
const mongoose = require("mongoose");

const shopItemSchema = new mongoose.Schema({}, { discriminatorKey: "type" });
const shopSchema = new mongoose.Schema({
    items: { type: [shopItemSchema], required: true },
});

const shopItemSubType = new mongoose.Schema({});
// const shopModel = mongoose.model("shop", shopSchema); // If the shop model is created before the schema is discriminated, recompile will work
shopItemSchema.discriminator(2, shopItemSubType);
const shopModel = mongoose.model("shop", shopSchema);

shopModel.recompileSchema(); // Error: Discriminator with name "2" already exists
```

### Expected Behavior

The schema should recompile without error.

vkarpov15:
Fixed by 23869fb10cd3e81c1bb2b51878a206ea30bb935e, I got my issue numbers mixed up

PR #14541, opened 2024-04-26
types(document): make document _id type default to unknown instead of any

### LINKED ISSUE #14520: Why is the default type of the generic field _id in the class Document of type "any"?
### Prerequisites

### Mongoose version

8.3.1

### Node.js version

20.12.1

### MongoDB version

6.0

### Operating system

macOS

14.2.1 (23C71)

### Issue

I just fell victim to an issue similar to: 
https://github.com/Automattic/mongoose/issues/11085

Which was caused by legacy code we've been using in NestJS:
https://github.com/nestjs/docs.nestjs.com/pull/2517

After researching this for quite some time, I undersand we are supposed to use the `HydratedDocument` these days, but I'd still be nice if somebody can shade some light into why it is what it is at the moment.

So here is my example:
Our schemas are defined like this (which is how it was found in NestJS docs, a few years ago):
```
import { Prop, Schema, SchemaFactory } from '@nestjs/mongoose';
import { Document } from 'mongoose';
import { Types } from 'mongoose';

@Schema()
export class Cat {
  _id: Types.ObjectId;

  @Prop()
  name: string;

  @Prop()
  age: number;

  @Prop()
  breed: string;
}

export type CatDocument = Cat & Document;
export const CatSchema = SchemaFactory.createForClass(Cat);

```

The interesting part is here is:
`export type CatDocument = Cat & Document;`

Which basically makes _id be of type `any` instead of `Types.ObjectId`.
As per above issue, I wonder why document is defined as `class Document<T = any, TQueryHelpers = any, DocType = any>` in https://github.com/Automattic/mongoose/blob/master/types/document.d.ts#L19 while the HydratedDocument was changed to use `unknown` to circumvent the problem with `any` inference.

Wouldn't it make sense to either use `unkown` or `class Document<T = Types.ObjectId, TQueryHelpers = any, DocType = any>` to default to ObjectId and leave the user a way to use any instead?

Thanks for any answers to this question.

vkarpov15:
Fix #14520

**Summary**

 I don't see any reason why `IdType` can't default to `unknown`, tests pass and `unknown` seems more correct. @hasezoey can you please check if this causes any problems for typegoose?

**Examples**

### DIFF
--- a/lib/schema/documentArray.js
+++ b/lib/schema/documentArray.js
@@ -195,7 +195,7 @@ SchemaDocumentArray.prototype.discriminator = function(name, schema, options) {
     schema = schema.clone();
   }
 
-  schema = discriminator(this.casterConstructor, name, schema, tiedValue);
+  schema = discriminator(this.casterConstructor, name, schema, tiedValue, null, null, options?.overwriteExisting);
 
   const EmbeddedDocument = _createConstructor(schema, null, this.casterConstructor);
   EmbeddedDocument.baseCasterConstructor = this.casterConstructor;
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -7463,6 +7463,23 @@ describe('Model', function() {
     assert.equal(instance.item.whoAmI(), 'I am Test2');
   });
 
+  it('overwrites existing discriminators when calling recompileSchema (gh-14527) (gh-14444)', async function() {
+    const shopItemSchema = new mongoose.Schema({}, { discriminatorKey: 'type' });
+    const shopSchema = new mongoose.Schema({
+      items: { type: [shopItemSchema], required: true }
+    });
+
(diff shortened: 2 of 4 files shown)


hasezoey (approved):
Looks good to me, also no problems with typegoose.
typegoose already had (more or less always) typed the first parameter of `Document` to `unknown`: [`DocumentType`](https://github.com/typegoose/typegoose/blob/2c2ab6dad3d558e92392cb676a78042ab9487443/src/types.ts#L17)

vkarpov15:
I accidentally pushed fix for #14520 into this branch, so I will merge into 8.4 and then cherry pick 23869fb

hasezoey:
> I accidentally pushed fix for https://github.com/Automattic/mongoose/issues/14520 into this branch

i was slightly confused, do you mean #14527 and #14444 instead? as #14520 is the issue this PR was made for

Issue #14286, opened 2024-01-24
Typescript schema doesn't have _id field.


abarriel:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.0.4

### Node.js version

16

### MongoDB server version

6.0.6

### Typescript version (if applicable)

5.2.2

### Description

Hello,

I think the `SchemaType` of a  schema should have an  `_id` field **by default**

For e.g, we have `createdAt`, and `updatedAt` field in the type when we set the `option { timestamps: true }`.

Which is not the case when we set `options: { _id: true }`

### Steps to Reproduce

```ts
import mongoose, { type InferSchemaType, Schema } from "mongoose";

const eventSchema = new mongoose.Schema({ url: String }, { timestamps: true });
const Event = mongoose.model("Event", eventSchema);

type EventDoc = InferSchemaType<typeof eventSchema>;

async function run() {
  const event = {} as EventDoc;

  // createdAt exists because timestamps: true, which is totally OK
  event.createdAt = new Date();

  //  WRONG: _id does not exist on type EventDoc, this should be set, since options._id is to true by default
  event._id; // ERROR
}

run();

```

IslandRhythms:
```typescript
import mongoose, { type InferSchemaType, Schema } from "mongoose";

const eventSchema = new mongoose.Schema({ url: String }, { timestamps: true });
const Event = mongoose.model("Event", eventSchema);

type EventDoc = InferSchemaType<typeof eventSchema>;

async function run() {
  const event = {} as EventDoc;

  // createdAt exists because timestamps: true, which is totally OK
  event.createdAt = new Date();

  //  WRONG: _id does not exist on type EventDoc, this should be set, since options._id is to true by default
  event._id; // ERROR
}

run();
```

vkarpov15:
I agree this is a bit confusing, however currently the issue is that `InferSchemaType` doesn't return the full hydrated doc type. You still need to pass `InferSchemaType` through `HydratedDocument<>`. The easiest way to infer the full hydrated document type currently is the following:

```ts
type EventDoc = ReturnType<(typeof Event)['hydrate']>; //InferSchemaType<typeof eventSchema>;
```

vkarpov15:
[...]

PR #14542, opened 2024-04-27
docs: de-emphasize `InferSchemaType<>` in TypeScript docs in favor of automatic inference

vkarpov15:
Re: #14286

**Summary**

Did a bunch of work on the TypeScript docs to make it more clear that using `InferSchemaType<>` is usually not necessary. Because `InferSchemaType<>` returns a mostly raw document interface but with DocumentArrays and Subdocs, it typically isn't useful for end users because you end up with a mostly-lean document. `InferSchemaType<>` is still useful internally because it helps automatic type inference internally, but it unfortunately tries to satisfy both the lean document type and the hydrated document type, which leads to confusion.

For user-facing API, we should instead encourage `inferRawDocType` from #13900, which will infer the raw doc type from the schema, and `inferHydratedDocumentType`, which will infer the fully hydrated document type but is not yet written.

**Examples**

### DIFF
--- a/docs/typescript/schemas.md
+++ b/docs/typescript/schemas.md
@@ -66,7 +66,7 @@ For example, the above code won't throw an error if `email` is optional in the d
 The Mongoose `Schema` class in TypeScript has 9 [generic parameters](https://www.typescriptlang.org/docs/handbook/2/generics.html):
 
 * `RawDocType` - An interface describing how the data is saved in MongoDB
-* `M` - The Mongoose model type. Can be omitted if there are no query helpers or instance methods to be defined.
+* `TModelType` - The Mongoose model type. Can be omitted if there are no query helpers or instance methods to be defined.
   * default: `Model<DocType, any, any>`
 * `TInstanceMethods` - An interface containing the methods for the schema.
   * default: `{}`
--- a/docs/typescript/schemas.md
+++ b/docs/typescript/schemas.md
@@ -72,6 +72,7 @@ The Mongoose `Schema` class in TypeScript has 9 [generic parameters](https://www
   * default: `{}`
 * `TQueryHelpers` - An interface containing query helpers defined on the schema. Defaults to `{}`.
 * `TVirtuals` - An interface containing virtuals defined on the schema. Defaults to `{}`
+* `TStaticMethods` - An interface containing methods on a model. Defaults to `{}`
 * `TSchemaOptions` - The type passed as the 2nd option to `Schema()` constructor. Defaults to `DefaultSchemaOptions`.
 * `DocType` - The inferred document type from the schema.
 * `THydratedDocumentType` - The hydrated document type. This is the default return type for `await Model.findOne()`, `Model.hydrate()`, etc.
--- a/docs/typescript/schemas.md
+++ b/docs/typescript/schemas.md
@@ -1,14 +1,49 @@
 # Schemas in TypeScript
 
 Mongoose [schemas](../guide.html) are how you tell Mongoose what your documents look like.
-Mongoose schemas are separate from TypeScript interfaces, so you need to either define both a *document interface* and a *schema*; or rely on Mongoose to automatically infer the type from the schema definition.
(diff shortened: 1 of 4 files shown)


docs/typescript/schemas.md
```diff
@@ -1,89 +1,86 @@
...
-3. Mongoose adds `createdAt` and `updatedAt` to your schema if you specify the `timestamps` option in your schema, *except* if you also specify `methods`, `virtuals`, or `statics`. There is a [known issue](https://github.com/Automattic/mongoose/issues/12807) with type inference with timestamps and methods/virtuals/statics options. If you use methods, virtuals, and statics, you're responsible for adding `createdAt` and `updatedAt` to your schema definition.
-
-If automatic type inference doesn't work for you, you can always fall back to document interface definitions.
+By default, Mongoose does **not** check if your raw document interface lines up with your schema.
+For example, the above code won't throw an error if `email` is optional in the document interface, but `required` in `schema`.
 
 ## Generic parameters
 
 The Mongoose `Schema` class in TypeScript has 4 [generic parameters](https://www.typescriptlang.org/docs/handbook/2/generics.html):
```
hasezoey, line 66:
```suggestion
The Mongoose `Schema` class in TypeScript has 9 [generic parameters](https://www.typescriptlang.org/docs/handbook/2/generics.html):
```

if counting all the generics in the current schema type, unless i missed something

https://github.com/Automattic/mongoose/blob/5c4051ac56b4414afe7a6a02f6b946c7360c5ea8/types/index.d.ts#L219-L235

docs/typescript/schemas.md
```diff
@@ -1,89 +1,86 @@
...
 
-* `DocType` - An interface describing how the data is saved in MongoDB
+* `RawDocType` - An interface describing how the data is saved in MongoDB
 * `M` - The Mongoose model type. Can be omitted if there are no query helpers or instance methods to be defined.
```
hasezoey, line 69:
```suggestion
* `TModelType` - The Mongoose model type. Can be omitted if there are no query helpers or instance methods to be defined.
```

current generic name, more explaining than a simple `M`

docs/typescript/schemas.md
```diff
@@ -1,89 +1,86 @@
...
   * default: `Model<DocType, any, any>`
 * `TInstanceMethods` - An interface containing the methods for the schema.
   * default: `{}`
 * `TQueryHelpers` - An interface containing query helpers defined on the schema. Defaults to `{}`.
+* `TVirtuals` - An interface containing virtuals defined on the schema. Defaults to `{}`
+* `TSchemaOptions` - The type passed as the 2nd option to `Schema()` constructor. Defaults to `DefaultSchemaOptions`.
```
hasezoey, line 75:
```suggestion
* `TVirtuals` - An interface containing virtuals defined on the schema. Defaults to `{}`
* `TStaticMethods` - An interface containing methods on a model. Defaults to `{}`
* `TSchemaOptions` - The type passed as the 2nd option to `Schema()` constructor. Defaults to `DefaultSchemaOptions`.
```

docs/typescript/schemas.md
```diff
@@ -1,89 +1,86 @@
...
+* `DocType` - The inferred document type from the schema.
+* `THydratedDocumentType` - The hydrated document type. This is the default return type for `await Model.findOne()`, `Model.hydrate()`, etc.
 
 <details>
   <summary>View TypeScript definition</summary>
 
   ```typescript
-  class Schema<DocType = any, M = Model<DocType, any, any>, TInstanceMethods = {}, TQueryHelpers = {}> extends events.EventEmitter {
+  class Schema<RawDocType = any, M = Model<DocType, any, any>, TInstanceMethods = {}, TQueryHelpers = {}> extends events.EventEmitter {
```
hasezoey, line 83: that should likely be updated with the full current schema type definition, or a plain link to the file in the github repo

hasezoey (approved):
Looks good, some suggestions if we touch that documentation

PR #14568, opened 2024-05-05
fix(error): remove model property from CastError to avoid printing all model properties to console

### LINKED ISSUE #14529: Dumps *gigantic* object into console on error
### Prerequisites

### Mongoose version

8.3.1

### Node.js version

21.6.2

### MongoDB server version

Irrelevant

_No response_

### Description

Every time I have a problem with mongoose, it decides to dump a gigantic (larger than my terminal scrollback, which is set to 1 billion lines) object into the console.
This makes it completely impossible to diagnose anything.
I have mongoose.set('debug', false),
I don't have any other loggers enabled.

### Steps to Reproduce

Happens in random places in this repo: https://github.com/acheung-94/pet-portal/tree/buildNewPetForm
notably when you send a put request to /api/pets/:id

### Expected Behavior

I expect there to be some way to get mongoose to not dump the entire mongoose object to console

vkarpov15:
Fix #14529

**Summary**

The following script currently prints > 10k lines of output:

```
'use strict';

mongoose.connect('mongodb://127.0.0.1:27017/mongoose_test2');

const TestModel = mongoose.model('Test', mongoose.Schema({ 
  num: { type: Number }
}));

TestModel.findOne({ num: 'foo' }).exec();

```

Because `CastError` contains the full model, and Node.js' `inspect()` seems to ignore custom inspect handlers for error instances. So the above script prints out `TestModel` and all its subprops, including connections and connection strings in plaintext. This is a bit of a nasty surprise as we saw in #14529.

I did a bunch of digging to try to find some way to get Node's error formatting to avoid printing `model`, the only approach that seems to work is making `model` a getter and storing the actual mapping of error -> model in a WeakMap. However that seems a bit excessive.

**Examples**

### DIFF
--- a/lib/error/cast.js
+++ b/lib/error/cast.js
@@ -75,7 +75,6 @@ class CastError extends MongooseError {
    * ignore
    */
   setModel(model) {
-    this.model = model;
     this.message = formatMessage(model, this.kind, this.value, this.path,
       this.messageFormat, this.valueType);
   }


hasezoey (approved):
Looks good to me, just consider maybe putting this in the next minor version instead of a patch

vkarpov15:
I agree @hasezoey

PR #14582, opened 2024-05-09
feat(mongoose): export omitUndefined() helper

### LINKED ISSUE #14569: Export `omitUndefined()` function similar to what `cast$expr` has
We already have an internal `_omitUndefined()` function that would be useful for deleting keys that are `===` undefined for cases like #14547.

vkarpov15:
Fix #14569

**Summary**

Export `omitUndefined()` helper for #14569 re: discussion from #14547

**Examples**

### DIFF
--- a/test/types/base.test.ts
+++ b/test/types/base.test.ts
@@ -69,3 +69,5 @@ function setAsObject() {
 
   expectError(mongoose.set({ invalid: true }));
 }
+
+const x: { name: string } = mongoose.omitUndefined({ name: 'foo' });
--- a/lib/helpers/omitUndefined.js
+++ b/lib/helpers/omitUndefined.js
@@ -1,0 +1,20 @@
+'use strict';
+
+module.exports = function omitUndefined(val) {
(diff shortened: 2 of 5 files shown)


test/types/base.test.ts
```diff
@@ -69,3 +69,5 @@ function setAsObject() {
 
   expectError(mongoose.set({ invalid: true }));
 }
+
+const x: { name: string } = mongoose.omitUndefined({ name: 'foo' });
```
hasezoey, line 73: consider also adding a test that actually removes something

hasezoey (approved):

test/types/base.test.ts
vkarpov15: A reasonable suggestion, but not really worth it for TypeScript since we don't actually run the TypeScript tests, we're just testing the type

hasezoey: right, forgot the typescript type didnt actually modify the input type

PR #14583, opened 2024-05-10
feat: add `transactionAsyncLocalStorage` option to opt in to automatically setting `session` on all transactions

### LINKED ISSUE #13889: Use AsyncLocalStorage for transaction context support
### Prerequisites

### 🚀 Feature Proposal

A few other ORMs use the NodeJS `AsyncLocalStorage` API to provide a more ergonomic way of using transactions. One example is Sequelize: https://sequelize.org/docs/v7/querying/transactions/#disabling-cls

Basically this allows someone to write a bunch of business logic that might span several different function calls etc. while having all the ORM database calls automatically use a transaction that was started earlier in the callstack, without having to pass an object like `session` around and manually provide it to each ORM call.

I think this would be a pretty powerful alternative interface for transactions

### Motivation

We are writing an API where we have a use-case that requires us to start a transaction at the beginning of every request and then use it for each DB operation that occurs within that request's business logic, before committing the transaction at the end.

Currently that requires us to find our own way to pass the session object around, eg:

```typescript
async function requestHandler() {
  await connection.withTransaction(async (session) => {
     await someBusinessLogic(session)
 
  }
}

async function someBusinessLogic(session) {
 // need to explicitly provide session
  MyModel.findOneAndUpdate({}, {}, {session})
}
```

but with this proposal you wouldn't need to provide the session:
```typescript
async function requestHandler() {
  await connection.withTransactionContext(async (session) => {
     await someBusinessLogic()
 
  }
}

async function someBusinessLogic() {
 // automatically uses transaction by finding it via AsyncLocalStorage context provided by `withTransactionContext` wrapper
   MyModel.findOneAndUpdate({}, {})
}
```

### Example

_No response_

vkarpov15:
Fix #13889

**Summary**

Having to pass `session` option to every operation in a transaction is cumbersome and error prone. Node 16+ has a stable AsyncLocalStorage class as part of the async_hooks API; we can use AsyncLocalStorage to set `session` option by default on all operations in a function as long as the user uses `Connection.prototype.transaction()` wrapper. See `docs/transactions.md` changes for an example.

**Examples**

### DIFF
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -102,7 +102,7 @@ const doc = new Test({ name: 'test' });
 await connection.transaction(async() => {
   await doc.save(); // Notice no session here
   throw new Error('Oops');
-});
+}).catch(() => {});
 
 // false, `save()` was rolled back
 await Test.exists({ _id: doc._id });
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -1,8 +1,6 @@
 # Transactions in Mongoose
 
-[Transactions](https://www.mongodb.com/transactions) are new in MongoDB
-4.0 and Mongoose 5.2.0. Transactions let you execute multiple operations
-in isolation and potentially undo all the operations if one of them fails.
+[Transactions](https://www.mongodb.com/transactions) let you execute multiple operations in isolation and potentially undo all the operations if one of them fails.
 This guide will get you started using transactions with Mongoose.
 
 <h2 id="getting-started-with-transactions"><a href="#getting-started-with-transactions">Getting Started with Transactions</a></h2>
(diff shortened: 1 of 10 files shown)


docs/transactions.md
```diff
@@ -86,6 +84,33 @@ Below is an example of executing an aggregation within a transaction.
...
+
+const Test = mongoose.model('Test', mongoose.Schema({ name: String }));
+
+const doc = new Test({ name: 'test' });
+
+// Save a new doc in a transaction that aborts
+await connection.transaction(async() => {
+  await doc.save(); // Notice no session here
+  throw new Error('Oops');
+});
```
hasezoey, line 105:
```suggestion
await connection.transaction(async() => {
  await doc.save(); // Notice no session here
  throw new Error('Oops');
}).catch(() => {});
```

i am not entirely sure, but wouldnt this `await connection.transaction` throw a error and so never get past that?

hasezoey (changes requested):
Mostly looks good to me, though one minor code example thing and failing Replica set tests because of this PR (this time not because of watch errors)

test error for reference:
```txt
   1) transactions
       transactionAsyncLocalStorage option
         transaction() sets `session` by default if transactionAsyncLocalStorage option is set:
     AssertionError [ERR_ASSERTION]: The input did not match the regular expression /Oops!/. Input:

'AssertionError [ERR_ASSERTION]: 0 == 1'

      at async Context.<anonymous> (test/docs/transactions.test.js:374:7)

```

hasezoey (approved):

PR #14605, opened 2024-05-18
fix(update): cast array of strings underneath doc array with array filters

### LINKED ISSUE #14595: When updating elements of primitive array using positional filtered operator, update value is incorrectly cast to array
### Prerequisites

### Mongoose version

7.6.2

### Node.js version

20.13.1

### MongoDB server version

7.0.8

5.2.2

### Description

When using the [filtered positional operator](https://www.mongodb.com/docs/manual/reference/operator/update/positional-filtered/) to update specific element(s) of an array of primitives (e.g. strings), mongoose incorrectly changes the input value from a primitive (e.g. string) to an array containing the input value.

This seems to be due to a misunderstanding of the intent of the operation. Mongoose identifies that the schema type of the path to be updated is an array, and therefore casts the input value to be an array. However, the intent of the update is to modify individual elements of that array, so the schema type for the input value should be identified as the type of the elements of the array, rather than an array itself.

This bug is not present in 7.6.1, but it appears in 7.6.2 and is still present in 7.6.11.

### Steps to Reproduce

Here's an example snippet. The additional nesting might not be entirely necessary to reproduce the bug, but it's how I encountered it. 

```js

const userSchema = {
  groups: [{
    document: mongoose.Schema.Types.ObjectId,
    tags: [String],
  }],
};

const groupId = new mongoose.Types.ObjectId();

await User.create({
  groups: [{
    document: groupId,
    tags: ['tag-first', 'tag-to-update', 'tag-last'],
  }],
});

await User.updateMany(
  { 
    groups: {
      $elemMatch: {
        document: groupId,
        tags: 'tag-to-update',
      },
    },
  },
  {
    $set: {
      'groups.$[group].tags.$[tag]': 'tag-modified',
    },
  },
  {
    arrayFilters: [
      { 'group.document': groupId },
      { tag: { $eq: 'tag-to-update' } },
    ],
  },
);

const userAfter = await User.findOne({});
console.log(userAfter.groups[0].tags);
// ['tag-first', ['tag-modified'], 'tag-last']
```

### Expected Behavior

Since the intent of the update is to modify individual element(s) of the array, rather than the array itself, the schema type of the elements of the array (e.g. in this case, `String`) should be used for casting.

The output of the example script should be:

```js
// ['tag-first', 'tag-modified', 'tag-last']
```

vkarpov15:
Fix #14595

**Summary**

`Schema.prototype._getSchema()` doesn't quite handle the case where you have `docArr.$.primitiveArr.$`, that last `$` gets dropped off because `primitiveArr.schema == null`.

**Examples**

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -2584,6 +2584,10 @@ Schema.prototype._getSchema = function(path) {
           // If there is no foundschema.schema we are dealing with
           // a path like array.$
           if (p !== parts.length) {
+            if (p + 1 === parts.length && foundschema.$embeddedSchemaType && (parts[p] === '$' || isArrayFilter(parts[p]))) {
+              return foundschema.$embeddedSchemaType;
+            }
+
             if (foundschema.schema) {
               let ret;
               if (parts[p] === '$' || isArrayFilter(parts[p])) {
--- a/test/helpers/update.castArrayFilters.test.js
+++ b/test/helpers/update.castArrayFilters.test.js
@@ -311,4 +311,42 @@ describe('castArrayFilters', function() {
 
     assert.strictEqual(q.options.arrayFilters[0]['element.number'], 1);
   });
+
+  it('correctly casts array of strings underneath doc array (gh-12565)', function() {
+    const userSchema = new Schema({
+      groups: [{
+        document: 'ObjectId',
+        tags: [String]
+      }]
+    });
+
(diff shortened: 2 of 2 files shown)


hasezoey (commented):
Looks good to me, though i am not very familiar with the code here.

PR #14606, opened 2024-05-20
Pass options to `clone` instead of `get` in applyVirtuals.

andrews05:
**Summary**

A bug introduced in v8.3.3 by #14543 caused `toJSON` to not be executed for virtual sub documents. The issue is caused by the `options` being incorrectly passed into the `get` function instead of the `clone` function.

**Examples**

I don't have an easy examples sorry, but I think the fix should be fairly evident. The intentions of #14543 was to pass `noDottedPath: true` to the `get` function but it mistakenly included the clone options which should have been passed to `clone` instead, as on line 4139 below it.


vkarpov15 (approved):
Very good catch, thank you!

PR #14618, opened 2024-05-29
Fix: The remove function was removed from the document, so it was modified.

swtpumpkin:
This code is missing differently from the previous version, resulting in an error.

    /** Removes this document from the db. */
    remove(options: QueryOptions, callback: Callback): void;
    remove(callback: Callback): void;
    remove(options?: QueryOptions): Promise<this>;

Therefore, I would like to request a correction.

### DIFF
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -1,5 +1,5 @@
-declare module 'mongoose' {
-  import mongodb = require('mongodb');
+declare module "mongoose" {
+  import mongodb = require("mongodb");
 
   /** A list of paths to skip. If set, Mongoose will validate every modified path that is not in this list. */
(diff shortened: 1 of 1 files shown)


vkarpov15:
Mongoose documents no longer have a `remove()` function as of Mongoose 7: https://mongoosejs.com/docs/migrating_to_7.html#removed-remove , so I don't think this PR is correct.

Is there something I'm missing here? It looks like there's a lot of unnecessary code formatting changes, which makes this PR a bit tricky to review.

swtpumpkin:
I didn't know the feature was deleted from mongoose 7 version.
I will close the request.
Thank you.

PR #14623, opened 2024-05-30
More toObject() perf improvements

vkarpov15:
**Summary**

#14606 caused some performance regression because `options` is 3rd param to `get()`, not 2nd.

Also did some work to remove `cloneOptions` in favor of just plain old `options`, and relying more on `_calledWithOptions` to handle the distinction between options that the user specified in the `toObject()` / `toJSON()` function call vs options that are set from the schema or mongoose global

**Examples**


andrews05:
> #14606 caused some performance regression because `options` is 3rd param to `get()`, not 2nd.

In my defence... "It was like that when I got here!" 😄

lib/document.js
```diff
@@ -3877,33 +3844,47 @@ Document.prototype.$toObject = function(options, json) {
...
     if (options.minimize) {
       ret = minimize(ret) || {};
     }
   }
 
-  if (options.virtuals || (options.getters && options.virtuals !== false)) {
-    applyVirtuals(this, ret, cloneOptions, options);
+  const virtuals = options._calledWithOptions.virtuals
+    ?? defaultOptions.virtuals
+    ?? undefined;
```
hasezoey, line 3871: just to confirm, there is no `options.virtuals` that needs to be checked?

hasezoey (commented):
LGTM, one question though

lib/document.js
vkarpov15: No, because the order of precedence is 1) `toObject({ virtuals: true })`, 2) `schema.set('toObject', { virtuals: true })`, 3) `mongoose.set('toObject', { virtuals: true })`. (1) is from `_calledWithOptions`; (2) and (3) are rolled into `defaultOptions`.

`options` is primarily for calling `clone()`. `$toObject()` should primarily be concerned with `_calledWithOptions`

vkarpov15:
@andrews05 no need to apologize, it's our fault, we should've caught that in code review.

Issue #14394, opened 2024-02-29
BulkSave on entities with virtual properties *and* a transaction: massive performance hit and OOM crashes


hardcodet:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.2.0

### Node.js version

20.x

### MongoDB server version

Atlas 7.x

### Typescript version (if applicable)

5.2.2

### Description

*SOLVED* - problem due to calls to `toObject` on complex hydrated documents, not the transactions.

We noticed massive performance issues in PROD on rather trivial bulk updates. I ran some local tests and narrowed it down to virtual properties in combination with transactions/sessions).

This is the `bulkSave` we perform to update our entities.

```
await this.model.bulkSave(batch, { session: transaction?.session });
```

Our entities have 4 virtual properties, e.g.:

```
CredentialSchema.virtual("parentAuthorization")
    .get(function() {
        return (this as any)._parentAuthorization;
    })
    .set(function(v) {
        (this as any)._parentAuthorization = v;
    });
```

For my tests, I ran the `bulkSave`

1. with the virtual properties declared in the schema, and populated (the problem)
2. with the virtual properties declared in the schema, but set the field values to `undefined` before running the update
3. without virtual properties in the schema, and setting field values to `undefined` before running the update.

2 and 3 resulted in the same performance, so I'll just list measures for 1 and 2 going forward.

#### Test Results

Here's some measurements with a clear outlier:

- Updating 100 entities with populated virtual properties and no transaction: < 150 ms
- **Updating 100 entities with populated virtual properties within a transaction: ~ 4 seconds**
- Updating 100 entities without populated virtual properties and no transaction: < 150 ms
- Updating 100 entities without populated virtual properties within a transaction: < 150 ms

With bigger arrays, the performance degradation is absolutely fatal, resulting in timeouts etc.
A test with 10K entities crashed my process because the JS heap ran out of memory (`V8::FatalProcessOutOfMemory+662`))

This looks like a bug to me, as the virtual properties shouldn't even be touched. My guess is that mongoose iterates through those virtual properties in order to detect changes rather than skipping them, which seems to introduce a _massive_ overhead. That wouldn't explain why it only occurs with a transaction/session in the mix. 

Right now, this really is a show stopper for us. Is there a better short-time solution than setting those properties to undefined before every update, and then reassigning them?

Thanks!

### Steps to Reproduce

[...]

PR #14626, opened 2024-05-31
remove old benchmarks, add new benchmark for #14394

vkarpov15:
**Summary**

Currently the benchmarks dir is filled with old benchmarks that haven't been maintained since 2013. Given that we're looking to focus more on performance, we should remove the stale benchmarks and instead start building out a new benchmark suite based on benchmarks we've found useful, like #14394.

Here are some priorities for new benchmarks:

1. Minimal overhead: just a plain Node.js script. No frameworks like benchmark, because benchmark adds (admittedly minimal) overhead that adds extra noise to output from tools like [0x](https://npmjs.com/package/0x), which I now rely on heavily for perf debugging.
2. Includes setup code, but also allows skipping expensive setup using `MONGOOSE_BENCHMARK_SKIP_SETUP` environment variable. Also for flamegraphs, as well as faster iteration.
3. Outputs only a JSON object that contains timing info, which we can then later store in MongoDB. We used to store benchmark results for our TypeScript benchmark, we disabled that temporarily but I would like to get that going again. It would be wonderful to have a website like `perf.mongoosejs.io` or something similar that shows Mongoose's performance on benchmarks over time.

**Examples**


hasezoey (approved):
LGTM, though why is this PR targeting `vkarpov15/gh-14394-2` (is it because it is build on it)?

vkarpov15:
Yep exactly, we will merge this PR into #14623, and then merge #14263.

Issue #12064, opened 2022-07-07
[Types] Query with dotted paths


hasezoey:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Currently Query's do not have dotted path suggestions, but would be possible with current typescript.

See [This Stackoverflow Answer](https://stackoverflow.com/a/47058976/8944059)

TL;DR:
a simple example of what the stackoverflow answer has does (note: the types are directly copied from the answer)
```ts
interface Test {
  somenested?: {
    path: string;
  };
  somethingelse?: string;
}

type PathsToStringProps<T> = T extends string
  ? []
  : {
      [K in Extract<keyof T, string>]: [K, ...PathsToStringProps<T[K]>];
    }[Extract<keyof T, string>];

type Join<T extends string[], D extends string> = T extends []
  ? never
  : T extends [infer F]
  ? F
  : T extends [infer F, ...infer R]
  ? F extends string
    ? `${F}${D}${Join<Extract<R, string[]>, D>}`
    : never
  : string;

function onlyDotted(input: Join<PathsToStringProps<Test>, '.'>) {
  console.log('input', input);
}

onlyDotted('somenested.path'); // works, with suggestions
onlyDotted('somethingelse'); // also works, with suggestions
onlyDotted('ERROR'); // error, property does not exist
```

Disclaimer: i am by no means a expert at complex typescript types, i also have no clue about the effect on types performance this would bring

hasezoey:
just noticed that #10180 is a similar Issue, should that one be closed in favor of this (or vise versa)?

vkarpov15:
I'm really wary of doing this because I imagine it'll be horrifically slow and cause a lot of impossible to debug "Type instantiation is excessively deep and possibly infinite" type errors.

I've tried adding the following to `query.d.ts`:

```ts
  type PathsToStringProps<T> = T extends string
    ? []
    : {
        [K in Extract<keyof T, string>]: [K, ...PathsToStringProps<T[K]>];
      }[Extract<keyof T, string>];

  type Join<T extends string[], D extends string> = T extends []
    ? never
    : T extends [infer F]
    ? F
    : T extends [infer F, ...infer R]
    ? F extends string
      ? `${F}${D}${Join<Extract<R, string[]>, D>}`
      : never
    : string;

  type _FilterQuery<T> = {
    [P in keyof T]?: Condition<T[P]>;
[...]

PR #14632, opened 2024-06-01
types(models+query): infer return type from schema for 1-level deep nested paths

### LINKED ISSUE #14615: Type is incorrect for Query.distinct when field is a nested field
### Prerequisites

### Mongoose version

8.1.2

### Node.js version

18.x

### MongoDB server version

4.2.8

_No response_

### Description

When using the `distinct` query for nested fields, the declared type of the return value is `unknown[]`, even though the type should be statically discoverable as `string[]`.

A minimal example 

## myrecord.schema.ts

```typescript

import { Prop, Schema, SchemaFactory } from '@nestjs/mongoose';
import { Document } from 'mongoose';


@Schema()
export class Foo {
  @Prop()
  one: string;
}

export const FooSchema = SchemaFactory.createForClass(Foo);


@Schema()
export class MyRecord extends Document {

  @Prop()
  _id: string;

  @Prop({ type: FooSchema })
  foo: Foo;

export const MyRecordSchema = SchemaFactory.createForClass(MyRecord);
```

## myrecord.service.ts

```typescript
import { Injectable } from '@nestjs/common';
import { InjectModel } from '@nestjs/mongoose';
import { Model } from 'mongoose';
import { MyRecord } from '../schemas/myrecord.schema';

@Injectable()
export class MyService {
  constructor(
    @InjectModel('myrecord') private model: Model<MyRecord>,
  ) {}

  async myQuery(): Promise<void> {
    const foos: string[] = await this.model.distinct('foo.one').exec()
  }

}
```

The type checking error shows:

```Type 'unknown[]' is not assignable to type 'string[]'```

### Steps to Reproduce

I've included some sample code above. It's written in the style of nestjs, so it may need to be modified a bit for a simpler use case without nestjs.

1. Create a document schema with a subdocument.
2. Write a `distinct` query for the document on one of the nested fields on the subdocument.
3. The specified return type will be `unknown[]`

### Expected Behavior

The resolved return type should be an array of the subdocument field type. In this case, `string[]`.

### PR DESCRIPTION
**Pull request by vkarpov15**
Fix #14615
Re: #13836
Re: #12064

**Summary**

Fixes #14615 by inferring the type of 1-level deep schema paths to help with query filters. Consider the following schema:

```
  const FooSchema = new Schema({
    one: { type: String }
  });

  const MyRecordSchema = new Schema({
    _id: { type: String },
    foo: { type: FooSchema }
  });
```

Currently, Mongoose can infer `await MyRecord.distinct('foo').exec()` return type because `foo` is a top-level path. But not `await MyRecord.distinct('foo.one').exec()` because `foo.one` is nested.

As per discussion in #12064, we don't want to infer arbitrarily deep nested paths: too risky because we don't want to run into the dreaded `Type instantiation is excessively deep and possibly infinite` error in an unforeseen case, and we don't want to cause performance issues.

However, #14615 encourages us to consider just the case of 1-level deep nested paths. If we cut off at 1 level, we don't need to worry about infinite recursion and risk of performance degradation is much lower.

This PR adds a reasonably simple `WithLevel1NestedPaths` helper which, given a type, also adds all the 1-level deep nested paths to the result type. That makes it easy to get the correct result type for `distinct()` with 1-level deep nested paths. This PR intentionally avoids arrays.

**Examples**

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -13,7 +13,9 @@ import mongoose, {
   Query,
   UpdateWriteOpResult,
   AggregateOptions,
-  StringSchemaDefinition
+  WithLevel1NestedPaths,
+  NestedPaths,
+  InferSchemaType
 } from 'mongoose';
 import { expectAssignable, expectError, expectType } from 'tsd';
 import { AutoTypedSchemaType, autoTypedSchema } from './schema.test';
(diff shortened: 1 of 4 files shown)


hasezoey (commented):
LGTM

vkarpov15:
Merging this into 8.5 branch in the interest of caution

PR #14646, opened 2024-06-03
check the value of overwriteModels in options

uditha-g:
**Summary**

When we call the discriminator function of a model we can pass options with overwriteModels Boolean field. 
But that field isn't considered in the logic. 

**Examples**

model.discriminator(discriminatorName, schema, { overwriteModels: true });


vkarpov15 (approved):
Good find, thank you :+1:

Issue #8469, opened 2020-01-02
Run `refPath` and `ref` functions on subdocument being populated rather than top level doc


vkarpov15:
Re: #8452. This makes it exceedingly difficult to use `ref` functions as a replacement for `refPath`, and makes it hard to populate conditionally based on the contents of the subdocument.

```javascript
    const ImageSchema = Schema({ imageName: String });
    const Image = db.model('gh8452_Image', ImageSchema);

    const TextSchema = Schema({ textName: String });
    const Text = db.model('gh8452_Text', TextSchema);

    const opts = { _id: false };
    const ItemSchema = Schema({ objectType: String }, opts);
    const ItemSchemaA = Schema({
      data: { 
        type: ObjectId,
        refPath: function() { return 'list.objectType'; } // `this` will be the top-level `Example` doc
      },
      objectType: String,
    }, opts);
    const ItemSchemaB = Schema({
      data: { sourceId: Number },
      objectType: String,
    }, opts);

    const ExampleSchema = Schema({ test: String, list: [ItemSchema] });
    ExampleSchema.path('list').discriminator('gh8452_ExtendA', ItemSchemaA);
    ExampleSchema.path('list').discriminator('gh8452_ExtendB', ItemSchemaB);
    const Example = db.model('gh8452_Example', ExampleSchema);
```

dennisleung:
is there any solution for this? ...

vkarpov15:
@dennisleung not currently. We're working on a fix for v6.0.0.

vkarpov15:
`ref` is done in e141409, did some work for `refPath` in https://github.com/Automattic/mongoose/commit/6179267b61bb34d0b0b95ba36e35103db5ae2b7f.

Worth noting: I think we should treat `refPath` as being relative to the top-level document rather than the subdocument. There is a benefit to calling `refPath` with the subdocument as context (say, switching `refPath` based on the type of `this`), but Mongoose should treat the returned path as relative to the top level document, so users can reference top-level properties.

PR #14652, opened 2024-06-05
BREAKING CHANGE: call virtual `ref` function with subdoc, not top-level doc

### LINKED ISSUE #12363: Dynamic virtual populate not working on subdocument with separate schema

vkarpov15:
Fix #12363
Fix #12440

**Summary**

The primary issue from #12363 and #12440 is that right now we call `ref()` functions on virtuals with the top-level document rather than the subdocument, even if the virtual is defined on the subdocument. For example, in Mongoose 8, with the following setup, `ref()` would see `this` as an instance of `ParentSchema`, not the `b` subdocument.

```javascript
        const ASchema = new Schema({
          name: String
        });

        const BSchema = new Schema({
          referencedModel: String,
          aId: ObjectId
        });

        BSchema.virtual('a', {
          ref: function() {
            return this.referencedModel;
          },
          localField: 'aId',
          foreignField: '_id',
          justOne: true
        });

        const ParentSchema = new Schema({
          b: BSchema
        });

        const A1 = db.model('Test1', ASchema);
        const A2 = db.model('Test2', ASchema);
        const Parent = db.model('Parent', ParentSchema);
```

With this PR, we will call `ref()` with the subdocument. This is more consistent with how things work in conventional populate: `ref()` would get called with the subdoc in virtual populate, see #8469 and [this test](https://github.com/Automattic/mongoose/blob/c71ba5e044ee1f84d1fa85a2c21a90fad2e90f72/test/model.populate.test.js#L9870-L9920).

**Examples**

If you're using `ref()` as a function, it is easier to get the top-level document from the subdocument (using `subdoc.ownerDocument()`) than it is to get the correct subdocument from the top-level document. Plus, calling `ref()` with the subdocument helps make schemas easier to reuse.

### DIFF
--- a/lib/helpers/populate/getModelsMapForPopulate.js
+++ b/lib/helpers/populate/getModelsMapForPopulate.js
@@ -410,7 +410,15 @@ function _virtualPopulate(model, docs, options, _virtualRes) {
       justOne = options.justOne;
     }
 
-    modelNames = virtual._getModelNamesForPopulate(doc);
+    // Use the correct target doc/sub-doc for dynamic ref on nested schema. See gh-12363
+    if (_virtualRes.nestedSchemaPath && typeof virtual.options.ref === 'function') {
+      const subdocs = utils.getValue(_virtualRes.nestedSchemaPath, doc);
+      modelNames = Array.isArray(subdocs)
+        ? subdocs.flatMap(subdoc => virtual._getModelNamesForPopulate(subdoc))
+        : virtual._getModelNamesForPopulate(subdocs);
+    } else {
+      modelNames = virtual._getModelNamesForPopulate(doc);
+    }
     if (virtual.options.refPath) {
       justOne = !!virtual.options.justOne;
       data.isRefPath = true;
--- a/test/model.populate.test.js
+++ b/test/model.populate.test.js
@@ -4280,10 +4280,11 @@ describe('model: populate:', function() {
         ]);
         await Parent.create([
           { b: { name: 'test1', referencedModel: 'Test1', aId: as[0]._id } },
-          { b: { name: 'test2', referencedModel: 'Test2', aId: as[1]._id } }
+          { b: { name: 'test2', referencedModel: 'Test2', aId: as[1]._id } },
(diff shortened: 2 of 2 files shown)


hasezoey (approved):
LGTM, should a migrate entry be added in this PR itself, or later?

vkarpov15:
Later, since we don't have a `migrating_to_9` doc yet

PR #14674, opened 2024-06-18
Fix AWS Lambda docs link

ghemingway:
Both of the links at the end of the first paragraph pointed to the same Vercel page.  I believe that the last link in the first paragraph should instead point to the appropriate Mongoose AWS page.


vkarpov15 (approved):
Thanks :+1:

PR #14679, opened 2024-06-21
Remove default connection if setting createInitialConnection to false after Mongoose instance created

vkarpov15:
**Summary**

Right now there's always a default `mongoose.connection`. We added a `createInitialConnection` option that prevents new `Mongoose` instances from creating a default connection, but that doesn't apply to existing `Mongoose` instances, including the default one. So there's no way to really "delete" the default connection in the API.

With this PR, `mongoose.set('createInitialConnection', false)` will also remove the default connection, presuming it isn't connected and there are no models already on it.

**Examples**

### DIFF
--- a/lib/mongoose.js
+++ b/lib/mongoose.js
@@ -436,6 +436,11 @@ Mongoose.prototype.connect = async function connect(uri, options) {
   }
 
   const _mongoose = this instanceof Mongoose ? this : mongoose;
+  if (_mongoose.connection == null) {
+    const conn = this.createConnection(); // default connection
+    conn[defaultConnectionSymbol] = true;
+    conn.models = this.models;
+  }
   const conn = _mongoose.connection;
 
   return conn.openUri(uri, options).then(() => _mongoose);
--- a/lib/mongoose.js
+++ b/lib/mongoose.js
@@ -33,6 +33,7 @@ const SetOptionError = require('./error/setOptionError');
 const applyEmbeddedDiscriminators = require('./helpers/discriminator/applyEmbeddedDiscriminators');
 
 const defaultMongooseSymbol = Symbol.for('mongoose:default');
+const defaultConnectionSymbol = Symbol('mongoose:defaultConnection');
 
 require('./helpers/printJestWarning');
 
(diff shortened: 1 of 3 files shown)


lib/mongoose.js
```diff
@@ -292,6 +294,16 @@ Mongoose.prototype.set = function(key, value) {
       } else if (!optionValue && _mongoose.transactionAsyncLocalStorage) {
         delete _mongoose.transactionAsyncLocalStorage;
       }
+    } else if (optionKey === 'createInitialConnection') {
+      if (optionValue && !_mongoose.connection) {
+        const conn = this.createConnection(); // default connection
+        conn[defaultConnectionSymbol] = true;
+        conn.models = this.models;
```
hasezoey, line 301: i dont know if this will be modified often, but maybe consider putting this code in a function to de-duplicate default connection creation

test/index.test.js
```diff
@@ -1180,4 +1180,14 @@ describe('mongoose module:', function() {
       }
     });
   });
+
+  it('setting createInitialConnection after creation deletes existing connection (gh-8302)', async function() {
```
hasezoey, line 1184: maybe also add a test for when a model is registered on that connection and/or when the connection is not "disconnected" before setting the option

hasezoey (approved):

test/index.test.js
vkarpov15: Done in 8eafe6e

lib/mongoose.js
vkarpov15: Fixed in c04a75d

test/index.test.js
```diff
@@ -1180,4 +1180,52 @@ describe('mongoose module:', function() {
...
+
+    beforeEach(function() {
+      m = new mongoose.Mongoose();
+    });
+
+    afterEach(async function() {
+      await m.disconnect();
+    });
+
+    it('should delete existing connection when setting createInitialConnection to false (part 1)', function() {
```
hasezoey, line 1195:
```suggestion
    it('should delete existing connection when setting createInitialConnection to false', function() {
```

as there is seemingly no `Part 2`

hasezoey (approved):

PR #14685, opened 2024-06-24
fix(document): avoid passing validateModifiedOnly to subdocs so subdocs get fully validating if they're directly modified

### LINKED ISSUE #14677: Document.prototype.validate() does not validate subdocument required fields with validateModifiedOnly option
### Prerequisites

### Mongoose version

8.4.3

### Node.js version

20.14.0

### MongoDB server version

7.0.11

_No response_

### Description

In versions 8.2.4 and prior, a call to `Document.prototype.validate(null, {validateModifiedOnly: true})` would throw an error if required fields in a subdocument were missing. In 8.3.0 and subsequent versions, `validate()` no longer checks for required fields in subdocuments if the `validateModifiedOnly` option is set to true.

If subdocuments are added, the required fields should be validated as was done in 8.2.4 and prior.

### Steps to Reproduce

The following code will throw an error in mongoose versions 8.2.4 and prior, and succeed in 8.3.0 and subsequent:

```
      const embedSchema = new mongoose.Schema({
        field1: {
          type: String,
          required: true,
        },
        field2: String,
      });
        testField: {
          type: String,
          required: true,
        },
        testArray: [embedSchema],
      });
      const TestModel = mongoose.model('Test', testSchema);
      const m = new TestModel({testArray: [{field2: 'test'}]});
      await m.validate(null, {validateModifiedOnly: true});
```

With the `validateModifiedOnly` option, `testField` is not required, however it also skips the check for `field1` in the subdocument (which is missing) starting in version 8.3.0.

### Expected Behavior

If a subdocument is added, `validate()` should check that all required fields are present when the `validateModifiedOnly` is set to true.

vkarpov15:
Fix #14677

**Summary**

Right now we propagate `validateModifiedOnly` to subdocument `validate()`, which causes problems when a subdocument is directly modified. If you overwrite the full subdocument, we should validate the whole subdocument, not just the subdocs modified paths.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -3053,7 +3053,6 @@ Document.prototype.$__validate = function(pathsToValidate, options, callback) {
       const doValidateOptions = {
         ...doValidateOptionsByPath[path],
         path: path,
-        validateModifiedOnly: shouldValidateModifiedOnly,
         validateAllPaths
       };
 
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -2556,6 +2556,88 @@ describe('document', function() {
       // Does not throw
       await Model.create({ name: 'test' });
     });
+
+    it('fully validates modified subdocs (gh-14677)', async function() {
+      const embedSchema = new mongoose.Schema({
+        field1: {
+          type: String,
+          required: true
+        },
+        field2: String
+      });
+      const testSchema = new mongoose.Schema({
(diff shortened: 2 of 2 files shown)


hasezoey (approved):
Looks good to me, aside from the lint errors.

PR #14690, opened 2024-06-25
types(connection): fix return type of withSession()

tt-public:
**Summary**

This fixes `Connection.withSession()`'s type definition, which currently doesn't seem to match the implementation [[here](https://github.com/Automattic/mongoose/blob/3ad5b4f932180609ed7e68926fd7ca4abc6228b1/lib/connection.js#L458)].

**Examples**

(a test is added instead)


vkarpov15 (approved):
Thanks :+1:

PR #14691, opened 2024-06-25
feat(query): make sanitizeProjection prevent projecting in paths deselected in the schema

vkarpov15:
**Summary**

[sanitizeProjection option](https://thecodebarbarian.com/whats-new-in-mongoose-5-13-sanitizeprojection.html) currently exists to prevent cases like `select({ name: '$password' })`, which would cause the `name` property to contain the value of the `password` property in newer versions of MongoDB.

While that is helpful, `sanitizeProjection` can do a bit more to prevent inclusion of sensitive data when the projection is potentially untrusted. With this PR, if `sanitizeProjection` is enabled, there is no way to project in a field that's deselected with `select: false` in the schema definition. If `password` has `{ type: String, select: false }` and `sanitizeProjection` is set, then `select('+password')`, `select('password')`, etc. will be ignored.

@hasezoey what do you think about this PR, is this a reasonable feature and do you think it's reasonable to ship in 8.5?

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -1143,6 +1143,59 @@ Query.prototype.select = function select() {
 };
 
 /**
+ * Sets this query's `sanitizeProjection` option. If set, `sanitizeProjection` does
+ * two things:
+ *
+ * 1. Enforces that projection values are numbers, not strings.
+ * 2. Prevents using `+` syntax to override properties that are deselected by default.
+ *
+ * With `sanitizeProjection()`, you can pass potentially untrusted user data to `.select()`.
+ *
+ * #### Example
+ *
+ *     const userSchema = new Schema({
+ *       name: String,
+ *       password: { type: String, select: false }
+ *     });
+ *     const UserModel = mongoose.model('User', userSchema);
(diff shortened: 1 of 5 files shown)


hasezoey (approved):
> @hasezoey what do you think about this PR, is this a reasonable feature and do you think it's reasonable to ship in 8.5?

i think it is reasonable to ship it with 8.5.0.

---

i personally have not used this yet, but wouldnt this (according to the test cases) completely disable `find().select("+field -field")`, or is this just for fields which have a schema `select: false`?

as a side note, i had tried to search for option `sanitizeProjection` in the documentation, and the only mention of this is in `Query.prototype.setOptions` as `The following options are only for find(), findOne(), findById(), findOneAndUpdate(), findOneAndReplace(), findOneAndDelete(), and findByIdAndUpdate():`, no documentation about what this option does.

AbdelrahmanHafez (approved):

vkarpov15:
@hasezoey re: `select('+field')`, just for `+` and just for fields that have `select: false` in the schema. Other fields are fine. I also added a `Query.prototype.sanitizeProjection` method in 0904a18 and added docs with examples to address the docs issue.

PR #14692, opened 2024-06-25
fix(query): remove `count()` and `findOneAndRemove()` from query chaining

### LINKED ISSUE #14689: Query.count / Query.findOneAndRemove are still alive
### Prerequisites

### Mongoose version

8.4.3

### Node.js version

20.14

### MongoDB server version

7.0

_No response_

### Description

`Query.prototype.count()` and  `findOneAndRemove()` are still alive.

I found that the issue was that the query was using mquery as it is. - [permlink](https://github.com/Automattic/mongoose/blob/17334785f52e001610200380f10416806c674910/lib/query.js#L154)
mquery@5 has count() and findOneAndRemove().

Therefore, mongoose@8 removed `count()` and `findOneAndRemove()`, but the methods from mquery are being called.

### Steps to Reproduce

```javascript
await User.updateOne().findOne().findOneAndRemove();
await User.findOne().count();
```
I don't think this is good code, but it still works.

### Expected Behavior

_No response_

vkarpov15:
Fix #14689

**Summary**

A bit of an unfortunate case where mquery still supports `count()` and `findOneAndRemove()`, which means you can still access those functions in Mongoose 8 with query chaining. This PR removes that.

**Examples**

### DIFF
--- a/docs/nextjs.md
+++ b/docs/nextjs.md
@@ -3,7 +3,7 @@
 Next.js is a popular framework for building full stack applications with React.
 Mongoose works out of the box with Next.js.
 If you're looking to get started, please use [Next.js' official Mongoose sample app](https://github.com/vercel/next.js/tree/canary/examples/with-mongodb-mongoose).
-Furthermore, if you are using Next.js with [Vercel Serverless Functions](https://vercel.com/docs/concepts/functions/serverless-functions), please review [Mongoose's AWS Lambda docs](https://vercel.com/docs/concepts/functions/serverless-functions).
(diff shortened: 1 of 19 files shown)


hasezoey (approved):

DevooKim (approved):

AbdelrahmanHafez (approved):
LGTM, do we need to modify the TS declarations for this as well?

hasezoey:
> do we need to modify the TS declarations for this as well?

to my knowledge that has already been removed, the only mentions about `count` in the types are about `$count` and `aggregate.count()` (and `estimatedDocumentCount`)

PR #14717, opened 2024-07-02
feat: allow setting array default value to `null`

### LINKED ISSUE #6691: Unable to set array to `null` as default

Bug in the latest mongoose version (5.2.1)

Schemas ignore `default: null` for arrays, empty array is created instead.

```js
  arrayThatShouldBeNull: {
    type: [String],
    default: null,
    required: false
  },

  arrayThatShouldContainTwoEntries: {
    type: [String],
    default: ['asd', '123'],
    required: false
  }
}, {
  minimize: false // not sure if relevant
});

const Model = mongoose.model('test', schema);
const instance = new Model();

expect(instance.arrayThatShouldBeNull).to.be.null; // fails
expect(instance.arrayThatShouldContainTwoEntries).to.deep.equal(['asd', '123']); // passes
```

`instance.arrayThatShouldBeNull` should be set to null, but instead an empty array is created

vkarpov15:
Fix #6691

**Summary**

I took another look at #6691 and I can't see any reason for us continuing to disallow `default: null` on array paths. We already allow `default: undefined` and `default: () => null`, the latter of which is practically identical. Silently ignoring `default: null` for arrays feels more like a bug than anything, and changing it doesn't break any tests. 8.5 release is a good opportunity to make this change.

**Examples**

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -314,7 +314,7 @@ const Kitten = connection.model('Kitten', kittySchema);
 
 <a class="anchor" href="#array-defaults">**Q**</a>. How can I change mongoose's default behavior of initializing an array path to an empty array so that I can require real data on document creation?
 
-**A**. You can set the default of the array to a function that returns `undefined`.
+**A**. You can set the default of the array to `undefined`.
 
 ```javascript
 const CollectionSchema = new Schema({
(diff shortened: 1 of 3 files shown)


hasezoey (changes requested):
LGTM, it does not break any tests, but it would likely be good to have new tests for this

PR #14720, opened 2024-07-02
types: correct `this` for `validate.validator` schematype option

### LINKED ISSUE #14696: Validate in schema options is not properly typed
### Prerequisites

### Mongoose version

8.3.1

### Node.js version

18.17.1

### MongoDB server version

6

4.8

### Description

The validate config is not properly typed in schema options e.g

```
 isActive: {
            type: Boolean,
            default: false,
            validate: {
                validator(v) {
                   // this & v are any
                    return !v || this.name === "super admin"
                },
            },
        },
```

### Steps to Reproduce
```
import { Schema } from "mongoose";

interface User {
  name: string;
  isActive: boolean;
}

const userSchema = new Schema<User>({
  name: {
    type: String,
    required: [true, "Name on card is required"],
  },
  isActive: {
    type: Boolean,
    default: false,
    validate: {
      validator(v) {
        return !v || this.name === "super admin";
      },
    },
  },
});
```

### Expected Behavior

I would expect that `this` is properly typed.

vkarpov15:
Fix #14696

**Summary**

Ensure `this` has the correct type in `validate.validator` below.

```ts
    isActive: {
      type: Boolean,
      default: false,
      validate: {
        validator(v: any) {
          expectAssignable<User>(this);
          return !v || this.name === 'super admin';
        }
      }
    }
```

I made 2 changes to support this:

1) passing `EnforcedDocType` down to `ValidateFn` via `ValidateOpts` and `SchemaValidator` 
2) Removing `LegacyAsyncValidateFn`. For some reason, `LegacyAsyncValidateFn` breaks these tests, and I have no idea why. It has something to do with the `done` function. Either way, `LegacyAsyncValidateFn` is no longer necessary because we removed support for validators that receive a callback in [Mongoose 6](https://mongoosejs.com/docs/migrating_to_6.html#removed-validator-isasync), so I think it is safe to remove.

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -18,7 +18,8 @@ import {
   SchemaType,
   Types,
   Query,
-  model
+  model,
+  ValidateOpts
 } from 'mongoose';
 import { expectType, expectError, expectAssignable } from 'tsd';
 import { ObtainDocumentPathType, ResolvePathType } from '../../types/inferschematype';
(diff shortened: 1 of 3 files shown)


hasezoey (approved):

pharapeti:
**For context**: I'm upgrading from Mongoose v8.2.2 to v8.6.3

When performing a document create/update, `this` is the document (`mongoose.Document`). 
When performing a `MyModel.findOneAndUpdate` query with options `{ runValidators: true }`, `this` becomes is the query (`mongoose.Query`).

However, I believe due to these type changes make in this PR, Typescript seems to believe that `this` is **always** the document type - even in the case when it is a query.

Do you have any idea on how I can type the validator function so that Typescript knows that `this` is either a document OR a query type? @vkarpov15 @hasezoey

vkarpov15:
@pharapeti you're right that the workaround is to explicitly set the type of `this` in the `validator` function. We don't currently have `| Query` in the fix in this PR, and adding it would be a breaking change. We're going to put in a PR to fix this issue and some related issues, but you'll likely have to use the workaround with explicit `this` typing for the near future.

pharapeti:
@vkarpov15 Thanks for confirming that, will implement the workaround for now :+1: 

Cheers for the speedy reply :pray:

RadotJar:
pharapeti's workaround no longer appears to work as of v8.9.7. 

Using the same example provided, I get the following error on the type of 'this':

errorMongoose

It seems that the validator function is now expecting the 'this' argument to be of the 'PassPersisted' type, which I assume is the raw unhydrated data type representing what is stored in MongoDB. This does not align with my understanding of what is written in the docs at all.

vkarpov15:
@RadotJar can you please open a new issue and follow the issue template?

PR #14724, opened 2024-07-03
Performance improvements for `insertMany()`

vkarpov15:
**Summary**

The big changes are: 

1. a `_skipParallelValidateCheck` option for `validate()`. Looks like creating a new `ParallelValidateError()` each time is relatively expensive, and unnecessary for `insertMany()` since the user won't have access to the document references anyway outside of middleware.
2. Skip `$toObject()` in favor of `$__toObjectInternal()`, which skips options processing and only does a shallow clone, when using `insertMany()` on documents which only have primitive values (no subdocs, nested paths, or arrays).

Before this PR:

```
$ node gh-14719.js 
{
  "Average insertMany time ms": 54.99
}
```

After this PR:

```
$ node gh-14719.js 
{
  "Average insertMany time ms": 20.66
}
```

Looking at about 2.5x performance improvement.

Potential alternative would be to change `this.$__.validating` to use a boolean instead of an error instance, will investigate.

**Examples**


lib/document.js
```diff
@@ -3868,17 +3868,21 @@ Document.prototype.$toObject = function(options, json) {
...
-    ret = this._doc
-      ? options.minimize ? (minimize({ ...this._doc }) || {}) : { ...this._doc }
-      : {};
-    if (!options.minimize || options.flattenObjectIds) {
-      const keys = Object.keys(ret);
-      for (const key of keys) {
-        // If `minimize` is false, still need to remove undefined keys
-        if (!options.minimize && ret[key] === undefined) {
+    if (this._doc == null) {
+      ret = {};
```
billouboq, line 3874: we are doing this below, do we still need this ?

lib/document.js
```diff
@@ -3912,6 +3927,26 @@ Document.prototype.$toObject = function(options, json) {
   return ret;
 };
 
+/*!
+ * Internal shallow clone alternative to `$toObject()`: much faster, no options processing
+ */
+
+Document.prototype.$__toObjectInternal = function $__toObjectInternal() {
```
hasezoey, line 3934: i dont think `Internal` is a good suffix for this, maybe consider something along the lines of `$__toObjectShallow`?

vkarpov15: Fixed in 8223f91, good suggestion :+1:

Issue #14722, opened 2024-07-03
Model.bulkSave() and setting the query with "save" and "bulkWrite" pre middleware causes path conflict


blowfishlol:
### Prerequisites

- [X] I have written a descriptive issue title

### Mongoose version

8.4.4

### Node.js version

18.19.0

### MongoDB version

6.0.15 Community

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

14.4.1 (23E224)

### Issue

I have the following code that have both save and bulkWrite pre-hooks installed to a schema:
```javascript
const mongoose = require('mongoose');

const getUser = () => ({
    _id: new mongoose.Types.ObjectId("66852317541ef0e22ae5214c"),
    name: "blowfishlol",
    email: "blowfish@test.com"
});

const updateInfoSchema = new mongoose.Schema({
    name: {
        type: String, required: true
    },
    email: {
        type: String,
        required: true
    }
}, {
    versionKey: false,
});
const schema = new mongoose.Schema({ name: String, updater: updateInfoSchema });
schema.pre(["updateOne", "updateMany", "findOneAndUpdate", "save"], function (next) {
    this.set("updater", getUser());
    next();
});
schema.pre("bulkWrite", function (next, ops) {
    for (const op of ops) {
        op.updateOne.update["$set"].updater = getUser();
    }
    next();
});

const main = async () => {
    await mongoose.connect('mongodb://localhost:27017/mongoose_test');

    const TestModel = mongoose.model('Test', schema, "tests");

    const doc1 = new TestModel({ name: 'foobar', updater: getUser() });
    const doc2 = new TestModel({ name: 'bazinga', updater: getUser() });
    await doc1.save();
    await doc2.save();


    const docs = await TestModel.find().exec();
    for (const doc of docs) {
        doc.name = doc.name + "edit";
    }
    await TestModel.bulkSave(docs);

    const result = await TestModel.find().exec();

    console.log('Result', result.map(doc => doc.toObject()));
    await mongoose.connection.collection("tests").drop();
};

main();
```
In the example above, i will get an error:
```
Uncaught MongoBulkWriteError MongoBulkWriteError: Updating the path 'updater' would create a conflict at 'updater'
    at handleWriteError (/Users/blowfish/Source/finway/laboratory/mongoose-test/node_modules/mongodb/lib/bulk/common.js:826:22)
[...]

Issue #4145, opened 2016-05-11
Direct object assignment causes wrong array validation


dbellavista:
Soo this is a bit tricky. If you have a Schema defining an object with a nested array of schemas and you perform a direct assignent of the nested object, the validator are called twice:

* Once with `this` set as the parent object
* Once with `this` set as the array schema (desiderata)

This is the minimum code I was able to write to reproduce the issue:

```javascript
const mongoose = require('mongoose');
// Creating the inner schema with custom validator
const InnerSchema = new mongoose.Schema({
  myfield: {
    type: String,
    validate: {
      validator: myValidator,
      message: 'Message'
    }
  },
  sibling: String
});
// Creating the main schema. The array is nested inside an object
const MySchema = new mongoose.Schema({
  nest: {
    myarray: [InnerSchema]
  },
  rootSibling: String
});

const Model = mongoose.model('MyModel', MySchema);

const inst = new Model({
    rootSibling: 'This is the root sibling'
  });
// Direct object assignment
inst.nest = {
    myarray: [{
      myfield: 'This is my field',
      sibling: 'This is the nested sibling'
    }]
};

inst.save();
// Expected output:
//        This is my field , This is the nested sibling , undefined
// Actual output:
//         This is my field , undefined , This is the root sibling
//         This is my field , This is the nested sibling , undefined


// My validator prints the value, this.sibling and this.rootSibling
function myValidator(v) {
  console.log(v, ',', this.sibling, ',', this.rootSibling);
}
```

dbellavista:
And if I do this:

```javascript
const inst = new Model({
    rootSibling: 'This is the root sibling',
    nest: {
      myarray: [{
        myfield: 'This is my field',
        sibling: 'This is the nested sibling'
      }]
    }
  });
```

it **works** as expected. Also this code works:

```javascript
const inst = new Model({
    rootSibling: 'This is the root sibling'
  });
inst.nest.myarray = [{
  myfield: 'This is my field',
  sibling: 'This is the nested sibling'
}];
```

dbellavista:
Sorry, forgot: Node `v6.1.0`, mongoose `4.4.16`. I confirm that version `4.4.15` doesn't have this bug.

PR #14728, opened 2024-07-05
Avoid leaving subdoc defaults on top-level doc when setting subdocument to same value

vkarpov15:
**Summary**

The cause of #14722 is that, when creating a new subdocument, Mongoose first applies defaults to the subdocument, and those defaults bubble up to the top-level document. Which leaves any default paths in `default` state on the top-level document, even if we're setting the subdocument to the exact same value.

This PR makes it so that we skip parent change tracking on defaults when creating a completely new subdocument - the idea is, if we're creating a new subdocument, we'll be setting the subdocument to the top-level document, so we will track the change as "set the subdocument" rather than "set the subdocument, and also set this default on the subdocument."

There's also some followup work I want to do related to #4145: #4145 succeeds because we leave document array underneath a nested path in `default` state, even though we modify the document array.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1691,6 +1691,13 @@ Document.prototype.$__set = function(pathToMark, path, options, constructing, pa
         }
       });
     }
+  } else if (schema && schema.instance === 'Embedded') {
+    const defaultPaths = Object.keys(this.$__.activePaths.states['default'] ?? {});
+    for (const defaultPath of defaultPaths) {
+      if (defaultPath.startsWith(path + '.')) {
+        this.$__.activePaths.clearPath(defaultPath);
+      }
+    }
   }
 
   let obj = this._doc;
--- a/lib/document.js
+++ b/lib/document.js
@@ -155,7 +155,9 @@ function Document(obj, fields, skipId, options) {
     // By default, defaults get applied **before** setting initial values
     // Re: gh-6155
     if (defaults) {
-      applyDefaults(this, fields, exclude, hasIncludedChildren, true, null);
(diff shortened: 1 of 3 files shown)


hasezoey (commented):
I dont fully understand what is fixed here, but the code looks good to me.

vkarpov15:
Lol. To put it another way, the issue is that, when you do the following, `user.$isDefault('updater._id')` will be `true` even though `updater` was set to the exact same value it was already.

```javascript
const getUser = () => ({
    _id: new mongoose.Types.ObjectId("66852317541ef0e22ae5214c"),
    name: "blowfishlol",
    email: "blowfish@test.com"
});

const { _id } = await TestModel.create({ name: 'foobar', updater: getUser() });

const user = await TestModel.findById(_id).orFail();
user.set('updater', getUser());

user.$isDefault('updater._id'); // true without this PR, which shouldn't happen since we set `updater` to the same value
```

lukas-becker0:
Is there any way to disable this if I want the old logic ?

vkarpov15:
@lukas-becker0 unfortunately not. What sort of issues is this change causing for you?

lukas-becker0:
Hi @vkarpov15 , 
If I understand this change correctly, It also means that getters are no longer applied recursively to nested/child documents ?

Since mongoose 8.5.1 I noticed that some of our tests are failing due to some getters on properties in deeply nested subDocuments (3 levels down) being called but the result not showing up in the resulting object when using `queryResult.toObject({ getters: true, virtuals: false})`.

PR #14730, opened 2024-07-06
Update some dev-dependencies

hasezoey:
**Summary**

This PR updates some dev-dependencies


vkarpov15 (approved):
I undid the axios upgrade, because we had some issues with Axios 1.2 and it isn't worthwhile right now to test whether Axios 1.7 works correctly for our very limited use case. I want to remove Axios dependency in the near future anyway.

PR #14732, opened 2024-07-07
fix(document): ensure post('deleteOne') hooks are called when calling `save()` after `subdoc.deleteOne()`

### LINKED ISSUE #9885: Pre-remove hooks not called for subdocuments
Bug

Pre-remove middleware is not called when removing an item from an array of subdocuments.  According to https://github.com/Automattic/mongoose/issues/6224, this seemed to work with Mongoose <5.0.0.  @vkarpov15 asked @georgehess to open a new issue on this in the same https://github.com/Automattic/mongoose/issues/6224, though it appears not to have happened, thus I'm running it up the flagpole.

This behaviour presents a problem for me because I need to know when a document is removed from an array, so that a corresponding cache entry can be deleted.  Unfortunately, modifiedPaths doesn't provide enough context to tell me when something is removed, only that something happened.  I can, of course, clear the cache entry manually from outside of my Mongoose code, but this seems like a less-than-elegant solution.

```

const SubSchema = mongoose.Schema({
  myValue: {
    type: String
  }
}, {});

SubSchema.pre('remove', async function() {
  console.log('Subdoc got removed!')
});



const Schema = mongoose.Schema({
  foo: {
    type: String,
    required: true
  },
  mySubdoc: {
    type: [SubSchema],
    required: true
  }
}, {minimize: false, collection: 'test'});

const Model = mongoose.model('TestModel', Schema);


const initMongoose = async () => {
  const mongooseUrl = `mongodb://user:pass@dbhost:27017/test`;
  try {
    await mongoose.connect(mongooseUrl, {
      useUnifiedTopology: true,
      useNewUrlParser: true,
      promiseLibrary: global.Promise,
      useFindAndModify: false
    });
  }
  catch (err) {
    console.error(err);
    process.exit(1);
  }
  console.log('Connected');
};


const test = async () => {
  await Model.deleteMany({}); // remove all existing documents
  const newModel = {
    foo: 'bar',
    mySubdoc: [{myValue: 'some value'}]
  };
  const document = await Model.create(newModel);
  console.log('Created document');
  console.log('document:', document);
  
  console.log('Removing subDocument');
  document.mySubdoc[0].remove();
  
  
  console.log('Saving document');
  await document.save().catch( (error) => {
    console.error(error);
    process.exit(1);
  });
  console.log('document:', document);

  console.log(`Notice that SubSchema.pre('remove') never ran`);
};



const main = async () => {
  await initMongoose();
  await test();
  process.exit(0);
};

main();
```

Output from the above is:
```
Connected
Created document
document: {
  _id: 601831bcddc22e01c640d6b6,
  foo: 'bar',
  mySubdoc: [ { _id: 601831bcddc22e01c640d6b7, myValue: 'some value' } ],
  __v: 0
}
Removing subDocument
Saving document
document: { _id: 601831bcddc22e01c640d6b6, foo: 'bar', mySubdoc: [], __v: 1 }
Notice that SubSchema.pre('remove') never ran
```

Pre-remove hooks should run when removing sub-documents.

Mongoose 5.11.14
Node.js 15.5.1
MongoDB 4.4.1

vkarpov15:
Fix #9885

**Summary**

It looks like when we removed `remove()`, we didn't make it so that `post('deleteOne')` hooks fire when you `save()` a deleted doc. This PR makes Mongoose consistently trigger post('deleteOne') middleware on subdocs when you call `subdoc.deleteOne(); await topLevelDoc.save();`, which is what used to happen when you did `subdoc.remove(); await topLevelDoc.save();`

**Examples**

### DIFF
--- a/lib/plugins/saveSubdocs.js
+++ b/lib/plugins/saveSubdocs.js
@@ -36,6 +36,27 @@ module.exports = function saveSubdocs(schema) {
     });
   }, null, unshift);
 
+  schema.s.hooks.post('save', function saveSubdocsPostDeleteOne(doc, next) {
+    const removedSubdocs = this.$__.removedSubdocs;
+    if (!removedSubdocs || !removedSubdocs.length) {
+      return next();
+    }
+
+    each(removedSubdocs, function(subdoc, cb) {
+      subdoc.$__schema.s.hooks.execPost('deleteOne', subdoc, [subdoc], function(err) {
+        cb(err);
+      });
+    }, (error) => {
+      this.$__.removedSubdocs = null;
+      if (error) {
+        return this.$__schema.s.hooks.execPost('deleteOne:error', this, [this], { error: error }, function(error) {
+          next(error);
+        });
+      }
+      next();
+    });
+  });
+
   schema.s.hooks.post('save', function saveSubdocsPostSave(doc, next) {
     if (this.$isSubdocument) {
       next();
--- a/lib/plugins/saveSubdocs.js
+++ b/lib/plugins/saveSubdocs.js
@@ -36,25 +36,26 @@ module.exports = function saveSubdocs(schema) {
     });
   }, null, unshift);
 
-  schema.s.hooks.post('save', function saveSubdocsPostDeleteOne(doc, next) {
+  schema.s.hooks.post('save', async function saveSubdocsPostDeleteOne() {
     const removedSubdocs = this.$__.removedSubdocs;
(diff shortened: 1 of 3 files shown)


vkarpov15:
I also refactored the subdoc post save hook to use async/await. Looks to be about 7% faster on the `mapOfSubdocs` benchmark on my local, so this refactor may actually be slightly faster.

We will do some work to add promise support for kareem, to avoid the need to explicitly instantiate `new Promise()`

PR #14735, opened 2024-07-09
types: make `_id` required on Document type

### LINKED ISSUE #14660: TS: _id not required on DocumentArray properties of documents returned from query
### Prerequisites

### Mongoose version

8.2.3

### Node.js version

20.14.0

### MongoDB version

?

### Operating system

macOS

14.1

### Issue

I have a Mongoose model which has a subdocument. I have a function that queries the model via the `.find()` method. The function's output is expected to conform to an output interface, however it is not playing nicely with the the auto-generated TypeScript types for the model. The [mongoose docs state](https://mongoosejs.com/docs/subdocs.html#finding-a-subdocument): "Each subdocument has an `_id` by default", however the types for the `DocumentArray` properties returned from a `find()` operation do not have `_id` marked as required.  

This feels like a bug on the Mongoose side but I am new to using the library so it's possible that I'm missing something.

**Questions:**

* Am I correct that it is safe to assume that any subdocuments will have an `_id` property when a document is retrieved via the `.find()` operations? (safe for any situations where I explicitly instruct Mongoose to _not_ place an `_id` on the subdocument, of course)
* Short of a change in `mongoose` code itself, how can I customize the output of the types in a succinct manner?  I'd prefer to not write my own types for the Schema and Model as mongoose does a good job inferring these values and I would like to avoid maintaining redundant code. I see that `.find()` is a generic function ([link](https://github.com/Automattic/mongoose/blob/8.4.1/types/models.d.ts#L660-L685)), so I could provide my own type that instructs `mongoose` that the `_id` is required, however I'm struggling to achieve that.

**Simplified example:**

```ts

const ChildSchema = new mongoose.Schema({
});

const ParentSchema = new mongoose.Schema({
  _id: { type: String, required: true },
  parts: { type: [ChildSchema] },
});

const ParentModel = mongoose.model('Parent', ParentSchema);


export const getParents = async (): Promise<ExpectedParent[]> => {
  const parents = await ParentModel.find();
  return parents;  // TypeScript error here!
};

interface ExpectedParent {
  _id: string;
  name: string;
  parts: Array<{ _id: string; name: string }>;
}
```

I see the following error from TS:

```
Type '(Document<unknown, {}, { name: string; _id: string; parts: DocumentArray<{ name: string; }>; }> & { name: string; _id: string; parts: DocumentArray<{ name: string; }>; } & Required<...>)[]' is not assignable to type 'ExpectedParent[]'.
  Type 'Document<unknown, {}, { name: string; _id: string; parts: DocumentArray<{ name: string; }>; }> & { name: string; _id: string; parts: DocumentArray<{ name: string; }>; } & Required<...>' is not assignable to type 'ExpectedParent'.
    The types returned by 'parts.pop()' are incompatible between these types.
      Type '(Subdocument<ObjectId> & { name: string; }) | undefined' is not assignable to type '{ _id: string; name: string; } | undefined'.
        Type 'Subdocument<ObjectId> & { name: string; }' is not assignable to type '{ _id: string; name: string; } | undefined'.
          Type 'Subdocument<ObjectId> & { name: string; }' is not assignable to type '{ _id: string; name: string; }'.
            Types of property '_id' are incompatible.
              Type 'ObjectId | undefined' is not assignable to type 'string'.
                Type 'undefined' is not assignable to type 'string'.ts(2322)
```

vkarpov15:
Fix #14660

**Summary**

`_id: T` is more correct because `_id` should always exist on a hydrated document by default. And this doesn't break any tests other than ones that assert on the behavior from #14660. @hasezoey what do you think, does this work fine with Typegoose?

**Examples**

### DIFF
--- a/test/types/docArray.test.ts
+++ b/test/types/docArray.test.ts
@@ -91,7 +91,7 @@ async function gh13424() {
   const TestModel = model('Test', testSchema);
 
   const doc = new TestModel();
-  expectType<Types.ObjectId | undefined>(doc.subDocArray[0]._id);
+  expectType<Types.ObjectId>(doc.subDocArray[0]._id);
 }
 
 async function gh14367() {
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -1386,7 +1386,7 @@ function gh13424() {
   const TestModel = model('TestModel', new Schema(testSchema));
 
   const doc = new TestModel({});
-  expectType<Types.ObjectId | undefined>(doc.subDocArray[0]._id);
+  expectType<Types.ObjectId>(doc.subDocArray[0]._id);
 }
 
 function gh14147() {
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -22,7 +22,7 @@ declare module 'mongoose' {
     constructor(doc?: any);
 
     /** This documents _id. */
-    _id?: T;
+    _id: T;
 
     /** This documents __v. */
     __v?: any;


hasezoey (approved):
LGTM, works in typegoose, likely because typegoose already always did a [`mongoose.Document & mongoose.RequireId`](https://github.com/typegoose/typegoose/blob/423109ab12f0ae4f7287d4b05d0ac9a5d5e31905/src/types.ts#L18) for typegoose's `DocumentType` (though it could not be removed because of the inference of `_id` as `ObjectId` if not explicitly defined)

PR #14740, opened 2024-07-11
Fix ChangeStream.close to return a Promise<void> like the driver

orgads:
**Summary**

Client code that calls `stream.close().then(...)` fails with TypeError, since mongoose `close()` returns undefined.


vkarpov15 (approved):
Thanks :+1:

PR #14742, opened 2024-07-11
feat: add transactionAsyncLocalStorage option to opt in to automatically setting session on all transactions

vkarpov15:
Backport #14583 to 7.x
Re: #13889

**Summary**

Add `transactionAsyncLocalStorage` option to 7.x. Should work fine because Node v14.x has "experimental" support for the same AsyncLocalStorage API that became stable in 16.x.

**Examples**

### DIFF
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -88,7 +88,7 @@ Below is an example of executing an aggregation within a transaction.
 
 One major pain point with transactions in Mongoose is that you need to remember to set the `session` option on every operation.
 If you don't, your operation will execute outside of the transaction.
-Mongoose 8.4 is able to set the `session` operation on all operations within a `Connection.prototype.transaction()` executor function using Node's [AsyncLocalStorage API](https://nodejs.org/api/async_context.html#class-asynclocalstorage).
+Mongoose 7.8 is able to set the `session` operation on all operations within a `Connection.prototype.transaction()` executor function using Node's [AsyncLocalStorage API](https://nodejs.org/api/async_context.html#class-asynclocalstorage).
 Set the `transactionAsyncLocalStorage` option using `mongoose.set('transactionAsyncLocalStorage', true)` to enable this feature.
 
 ```javascript
--- a/docs/transactions.md
+++ b/docs/transactions.md
@@ -1,8 +1,6 @@
 # Transactions in Mongoose
 
-[Transactions](https://www.mongodb.com/transactions) are new in MongoDB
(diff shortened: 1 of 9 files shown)


docs/transactions.md
```diff
@@ -86,6 +84,33 @@ Below is an example of executing an aggregation within a transaction.
 [require:transactions.*aggregate]
 ```
 
+<h2 id="asynclocalstorage"><a href="#asynclocalstorage">Using AsyncLocalStorage</a></h2>
+
+One major pain point with transactions in Mongoose is that you need to remember to set the `session` option on every operation.
+If you don't, your operation will execute outside of the transaction.
+Mongoose 8.4 is able to set the `session` operation on all operations within a `Connection.prototype.transaction()` executor function using Node's [AsyncLocalStorage API](https://nodejs.org/api/async_context.html#class-asynclocalstorage).
```
hasezoey, line 91:
```suggestion
Mongoose 7.8 is able to set the `session` operation on all operations within a `Connection.prototype.transaction()` executor function using Node's [AsyncLocalStorage API](https://nodejs.org/api/async_context.html#class-asynclocalstorage).
```

types/mongooseoptions.d.ts
```diff
@@ -203,6 +203,13 @@ declare module 'mongoose' {
      */
     toObject?: ToObjectOptions;
 
+    /**
+     * Set to true to make Mongoose use Node.js' built-in AsyncLocalStorage (Node >= 16.0.0)
```
hasezoey, line 207:
```suggestion
     * Set to true to make Mongoose use Node.js' built-in AsyncLocalStorage (Added in: v13.10.0, v12.17.0; Stable since 16.4.0)
```

hasezoey (changes requested):
Code looks good to me, though i would recommend to add a note to the documentation that it makes use of a experimental nodejs API below `16.4.0` (as added in the types by me and listed [under `AsyncLocalStorage`'s History](https://nodejs.org/api/async_context.html#class-asynclocalstorage)).

also i just noticed that option `transactionAsyncLocalStorage` (in latest docs) only returns a result for `docs/transactions.html`, maybe it should be mentioned as a possible option for `mongoose.set`.

aside from that, there are some minor doc suggestions that should be done.

hasezoey (commented):
LGTM, would wait until #14744 is merged (and included here) before approving

vkarpov15:
@hasezoey I backported #14744 in e909e0b, can you take a look?

hasezoey (approved):

PR #14743, opened 2024-07-11
fix(model): handle `transactionAsyncLocalStorage` option with `insertMany()`

### LINKED ISSUE #14738: transactionAsyncLocalStorage not honored in bulk operations like insertMany
### Prerequisites

### Mongoose version

8.4.0

### Node.js version

20.15.0

### MongoDB server version

6

_No response_

### Description

The new `transactionAsyncLocalStorage` mechanism seems to work great for individual operations like `save` and `create` on models. It doesn't seem to be honored on bulk operations such as `insertMany`.

### Steps to Reproduce

Forgive the slightly convoluted code here, we have a wrapper over mongoose models with some singleton logic for working in AWS Lambda.

This test passes as I'd expect:
```
it.only('supports transactions without explicitly passing the session object', async () => {
  mongoose.set('transactionAsyncLocalStorage', true);
  await ChildDao.instance.find({ filter: {}, context }); // just forcing a preflight to get a connection, ignore
  await ChildDao.instance.model.db
    .transaction(async () => {
      await ChildDao.instance.model.create({ name: 'test' });
      throw new Error('Rollback');
    })
    .catch(() => {});

  expect(await ChildDao.instance.model.findOne({ name: 'test' })).toBeNull();
});
```

This test fails:
```
it.only('supports transactions without explicitly passing the session object', async () => {
  mongoose.set('transactionAsyncLocalStorage', true);
  await ChildDao.instance.find({ filter: {}, context });
  await ChildDao.instance.model.db
    .transaction(async () => {
      await ChildDao.instance.model.insertMany([{ name: 'test' }]);
      throw new Error('Rollback');
    })
    .catch(() => {});

  expect(await ChildDao.instance.model.findOne({ name: 'test' })).toBeNull();
});
```

This test passes, explicitly passing the session:
```
it.only('supports transactions on bulk operations if explicitly passed a session', async () => {
  mongoose.set('transactionAsyncLocalStorage', true);
  await ChildDao.instance.find({ filter: {}, context });
  await ChildDao.instance.model.db
    .transaction(async (session) => {
      await ChildDao.instance.model.insertMany([{ name: 'test' }], { session });
      throw new Error('Rollback');
    })
    .catch(() => {});

  expect(await ChildDao.instance.model.findOne({ name: 'test' })).toBeNull();
});
```

### Expected Behavior

The `transactionAsyncLocalStorage` should allow the more ergonomic flow for bulk operations as well. This might fall under a feature request, but without the caveats explained in the docs I'd be inclined to call it a bug.

vkarpov15:
Fix #14738

**Summary**

Looks like we missed a spot when implementing `transactionAsyncLocalStorage`, we're not adding `session` option to `insertMany()`. I reviewed and don't see any other places in `Model` where we aren't setting `session`

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2838,6 +2838,11 @@ Model.$__insertMany = function(arr, options, callback) {
   const throwOnValidationError = typeof options.throwOnValidationError === 'boolean' ? options.throwOnValidationError : false;
   const lean = !!options.lean;
 
+  const asyncLocalStorage = this.db.base.transactionAsyncLocalStorage?.getStore();
+  if ((!options || !options.hasOwnProperty('session')) && asyncLocalStorage?.session != null) {
+    options = { ...options, session: asyncLocalStorage.session };
+  }
+
   if (!Array.isArray(arr)) {
     arr = [arr];
   }
--- a/test/docs/transactions.test.js
+++ b/test/docs/transactions.test.js
@@ -388,6 +388,8 @@ describe('transactions', function() {
           }();
           assert.equal(doc.name, 'test_transactionAsyncLocalStorage');
 
+          await Test.insertMany([{ name: 'bar' }]);
+
           throw new Error('Oops!');
         }),
         /Oops!/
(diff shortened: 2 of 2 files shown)


hasezoey (approved):
LGTM, should this be added to #14742 before it is merged?

vkarpov15:
Yep, will do

Issue #13079, opened 2023-02-23
Why is `any` used as the type for `Document.id`?


juona:
### Prerequisites

- [X] I have written a descriptive issue title

### Mongoose version

6.9.1

### Node.js version

16

### MongoDB version

5

### Operating system

Linux

### Issue

The virtual `id` getter very clearly always returns a String:

```javascript
function idGetter() {
  if (this._id != null) {
    return String(this._id);
  }

  return null;
}
```

However the type for the `id` field on `Document` objects is `id?: any;`.

Why not string? What am I not seeing?

vkarpov15:
See https://github.com/Automattic/mongoose/pull/10248#pullrequestreview-660464370, #9773, etc.

We should be able to change this in Mongoose 7. But we need to keep `id?: any` in Mongoose 6 to support `extends Document` with non-string ids.

vkarpov15:
On further review, I think we still need to keep `id: any` for 7.0. That's because, in TypeScript, `{ id: string } & { id: number }` resolves to `{ id: never }` :man_facepalming: 

![image](https://user-images.githubusercontent.com/1620265/220956872-5f34704e-ae31-4e28-bf63-f7f9dfbf9f11.png)

With `id?: any`, everything works as expected:

![image](https://user-images.githubusercontent.com/1620265/220957139-b37afc24-8a9d-4728-baec-7c6649847086.png)

We'll have to defer making this change until we manage to figure out how to use `MergeType` instead of `&` with `Document<>` in the `HydratedDocument` definition here: https://github.com/Automattic/mongoose/blob/05c30268277c7771cfa617f7e2ee359a237783a3/types/index.d.ts#L144-L146 . Right now, that causes a bunch of hard to debug test failures.

juona:
Gotcha! Thanks for the explanation. I knew there had to be a reason. Not a problem for me, feel free to close this as I was mostly just curious. : ]

mjfwebb:
@vkarpov15 Was there a particular reason you aren't using a type union? There is no error when using a union instead:

```
type A = { id: string }
type B = A | { id: number }

const x: B = { id: 'bar' }
const x2: B = { id: 1 }
```

![image](https://user-images.githubusercontent.com/28074382/233955357-c16b71b6-473b-4f6f-b467-a5f704c6c88a.png)

vkarpov15:
@mjfwebb the problem is that Mongoose's `HydratedDocument<>` type uses `&`, not `|`.

PR #14747, opened 2024-07-15
Updating id type to be string instead of any

vlio20:
I am chanigng the type of the document id property to be a string.


vkarpov15:
This is a duplicate of #13079, #9773, and others. Long story short, we can't merge this as written because it would break any case where a user defines a path named `id` with a different type, like `new Schema({ id: Number })`.

Issue #13086, opened 2023-02-26
Auto remove empty object in condition


adostal:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.9.2

### Node.js version

19.0.1

### MongoDB server version

6

### Typescript version (if applicable)

4.9.5

### Description

I use casl.js in my project.
The conditions themselves are generated automatically from the @casl/mongoose library
The problem is that newer versions of mongoose remove the empty object that is sometimes part of the condition.
After updating mongoose from **6.7.5** to **6.9.2**, the empty object started to be removed.

If it would be possible to set this behavior in the mongoose configuration, that would be great.

### Steps to Reproduce

```typescript
// query for all people (conditions are from casl.js - sometimes with empty object)
const people = PeopleModel.find({$or: [
  {},
  {type: "xxx"},
  .... other conditions
]})
```

### Expected Behavior

In version 6.9.2 is finally condition without empty object {}:
```
condition: $or: [{"type": "xxx", ... other conditions}]
```

In version 6.7.5 is finally condition with empty object {}:
```
condition: $or: [{}, {"type": "xxx", ... other conditions}]
```

vkarpov15:
This looks like it is caused by https://github.com/Automattic/mongoose/pull/12898 . We merged this issue because we thought that `$or: [{}]` is likely indicative of a potential bug where `strictQuery` removed some fields. However, you're right, there are some valid use cases for this behavior. We'll revert #12898 in 6.10.1.

PR #14749, opened 2024-07-15
Improve fix cast empty query

0x0a0d:
due to update on [6.10.1](https://github.com/Automattic/mongoose/blob/master/CHANGELOG.md#6101--2023-03-03) was reverted #12898 because of #13086
I believe that my commit has actually fixed a bug where `{}` was unintentionally created through the strict process, which was not the developer's intention. I am updating this pull request to address the issue I just mentioned.
Now, only allows empty condition when the developer directly provides it. Additionally, if an empty condition appears due to the casting process, it will be removed.

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -65,7 +65,7 @@ module.exports = function cast(schema, obj, options, context) {
       if (!Array.isArray(val)) {
         throw new CastError('Array', val, path);
       }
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Thanks :+1: I made a couple of minor changes

0x0a0d:
@vkarpov15 please check review

Issue #9076, opened 2020-05-30
How to programmatically create a new subdocument from outside document?


niftylettuce:
For example, the only way right now that is documented to create a subdocument is:

```js
const Model = mongoose.model('MyModel', schema);
const obj = new Model();
const subDocument = obj.someSubSchema.create(); // <--
```

I think it might be nice to document, if possible, a way to create the `subDocument` without having to create an instance of the model?

niftylettuce:
I suppose you could also do this:

```js
const subDocument = Model.create().someSubSchema.create();
``

Is this worth mentioning in the docs somewhere?

**Comment by niftylettuce** [comment]
Never knew about `hydrate`! <https://mongoosejs.com/docs/api.html#model_Model.hydrate

Can this be done for subdocs somehow? @vkarpov15

**Comment by vkarpov15** [comment]
Document arrays have a `create()` method:

```javascript
'use strict';
  
const mongoose = require('mongoose');

mongoose.set('useFindAndModify', false);

const { Schema } = mongoose;

run().catch(err => console.log(err));

async function run() {
  await mongoose.connect('mongodb://localhost:27017/test', {
    useNewUrlParser: true,
    useUnifiedTopology: true
  });

  await mongoose.connection.dropDatabase();

  const nestedSchema = new Schema({ name: String });
  nestedSchema.methods.getAnswer = () => 42;

  const schema = new Schema({
    arr: [nestedSchema]
  });
  const Model = mongoose.model('Test', schema);

  const doc = new Model({ arr: [] });
  const subdoc = doc.arr.create({ name: 'foo' });

  console.log(subdoc instanceof mongoose.Document); // true
  console.log(subdoc.getAnswer()); // 42
}
```

But I've struggled to find a case where this is useful. Can you elaborate on your use case for this?

niftylettuce:
Yes, see my comment/solution here: https://github.com/mblarsen/mongoose-hidden/issues/58#issuecomment-636508840

vkarpov15:
@niftylettuce I took a closer look at this, it looks like your issue won't be solved by `.create()` since you want to convert a POJO to a subdoc without creating a top-level document. In that case, you can use `SchemaType#cast()` to make Mongoose cast the POJO as shown below:

```javascript
'use strict';
  
const mongoose = require('mongoose');

mongoose.set('useFindAndModify', false);

const { Schema } = mongoose;

run().catch(err => console.log(err));

async function run() {
[...]

PR #14756, opened 2024-07-21
types: allow calling `SchemaType.cast()` without `parent` and `init` parameters

### LINKED ISSUE #14748: Optional Arguments in SchemaTypes.path().cast required by Typescript
### Prerequisites

### Mongoose version

7.6.7

### Node.js version

18.20.3

### MongoDB server version

7.0.12

5.2.2

### Description

It looks like the typescript signature for SchemaTypes.path().cast require some (what appear to be) optional fields.

I wanted to create a subdocument instance without creating a model. With reference to https://github.com/Automattic/mongoose/issues/9076,

I had 
image

which works exactly as intended with the //@ts-ignore, but otherwise gives me a type error. Upon further examination,
it appears that 
image
doc and init are required fields in the typescript signature, whereas they seem to be optional in a lot of the documentation I've seen.

This is my top-level schema file:
image

Please let me know if I'm mistaken somewhere, many thanks!! 

### Steps to Reproduce

Please see above. 

### Expected Behavior

Optional (?) fields should be required by typescript signature

vkarpov15:
Fix #14748
Re: #9076

**Summary**

The `doc` and `init` parameters to `cast()` should be optional in TypeScript, we have plenty of tests that call `SchemaType.prototype.cast()` with only 1 arg. Plus #9076 recommends using `Schema.Types.Subdocument.prototype.cast()` as an alternative for instantiating a subdocument without an associated parent document, which is the root cause of the issue in #14748.

**Examples**


test/types/schema.test.ts
```diff
@@ -1555,3 +1556,17 @@ function gh14696() {
...
+  const schema = new Schema({
+    arr: [nestedSchema],
+    singleNested: nestedSchema
+  });
+
+  // Cast to single nested subdoc
+  const subdoc = schema.path('singleNested')
+    .cast<HydratedArraySubdocument<{ name: string }>>({ name: 'bar' });
+  expectAssignable<{ name: string }>(subdoc);
+}
```
hasezoey, line 1572:
```suggestion
}

```

fix lint

test/types/schema.test.ts
```diff
@@ -1555,3 +1556,17 @@ function gh14696() {
...
+
+  const schema = new Schema({
+    arr: [nestedSchema],
+    singleNested: nestedSchema
+  });
+
+  // Cast to single nested subdoc
+  const subdoc = schema.path('singleNested')
+    .cast<HydratedArraySubdocument<{ name: string }>>({ name: 'bar' });
+  expectAssignable<{ name: string }>(subdoc);
```
hasezoey, line 1571: slight suggestion, also add a test line for the non-set return type overload

hasezoey (approved):
Looks good, some slight suggestions

vkarpov15:
@hasezoey please take a look at the new implementation, I added a test based on your comments, and also some more logic to allow passing `DocType` down to the Schema Subdocument class

hasezoey (approved):

PR #14760, opened 2024-07-23
fix(model+document): avoid depopulating manually populated doc as getter value

### LINKED ISSUE #14759: Manually populated documents get saved as strings instead of `ObjectId`s in combination with SchemaType getter
### Prerequisites

### Mongoose version

8.5.1

### Node.js version

20.15.1

### MongoDB server version

7.0.12

_No response_

### Description

I use a SchemaType getter on `ObjectId` because I think strings are more convenient to work with than ObjectIds.
Their use is described and mentioned here:
 * https://mongoosejs.com/docs/api/schematype.html#SchemaType.prototype.get()
 * https://thecodebarbarian.com/whats-new-in-mongoose-54-global-schematype-configuration.html#schematype-getters
 * https://github.com/Automattic/mongoose/issues/6996#issuecomment-434063799
 
Have a look at the code snippet below. I cannot query for a pet anymore, given its owner.

I think that is because `pet.save()` unexpectedly saved the owner id as a string into the mongoDB instead of an `ObjectId`.

image

### Steps to Reproduce

```js
/**
 * I want to globally work with strings instead of ObjectIds
 * because that allows me to reuse my underlying types
 * across server and client.
 *
 * Also see:
 * https://mongoosejs.com/docs/api/schematype.html#SchemaType.prototype.get()
 * https://thecodebarbarian.com/whats-new-in-mongoose-54-global-schematype-configuration.html#schematype-getters
 * https://github.com/Automattic/mongoose/issues/6996#issuecomment-434063799
 */
mongoose.Schema.ObjectId.get(
  value => {
    if (value?.constructor?.name?.toLowerCase() === 'objectid') return String(value)
    return value
  }
)

const ownerSchema = new mongoose.Schema({ name: 'String' })
const petSchema = new mongoose.Schema({
  name: 'String',
  owner: { type: 'ObjectId', ref: 'owner' }
})

const Owner = mongoose.model('owner', ownerSchema)
const Pet = mongoose.model('pet', petSchema)

const owner = new Owner({ name: 'Alice' })
const pet = new Pet({ name: 'Kitty', owner: owner })

await owner.save()
// pet.owner = pet.owner._id // Manually depopulating it like this makes the whole thing work
await pet.save()

/**
 * The property 'owner' of 'pet' has unexpectedly been saved
 * as a string into the mongoDB instead of ObjectId.
 * 
 * No matter how I query for it, mongoose is not
 * able to find it...
 */

const reloadAttempts = await Promise.all([
  Pet.findOne({ owner: owner }),
  Pet.findOne({ owner: owner._id }),
  Pet.findOne({ owner: String(owner._id) }),
  Pet.findOne({ owner: new mongoose.Types.ObjectId(owner._id) }),
])

console.log(reloadAttempts) // Logs [ null, null, null, null ] but shouldn't
```

### Expected Behavior

I expect my `reloadAttempts` to not be all `null`.

Also the ObjectId getter should not change the mongoDB data structure.

vkarpov15:
Fix #14759

**Summary**

When depopulating, it looks like Mongoose applies `_id` getters, which is problematic if you add a default getter to convert all ObjectIds to strings like here: https://thecodebarbarian.com/whats-new-in-mongoose-54-global-schematype-configuration.html#schematype-getters. Fixed by replacing `doc._id` with `doc._doc._id` to get the raw id and bypass any getters.

We should backport this to 7.x and 6.x. Same issue impacts those versions.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1394,7 +1394,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
 
     let didPopulate = false;
     if (refMatches && val instanceof Document && (!val.$__.wasPopulated || utils.deepEqual(val.$__.wasPopulated.value, val._id))) {
-      const unpopulatedValue = (schema && schema.$isSingleNested) ? schema.cast(val, this) : val._id;
+      const unpopulatedValue = (schema && schema.$isSingleNested) ? schema.cast(val, this) : val._doc._id;
       this.$populated(path, unpopulatedValue, { [populateModelSymbol]: val.constructor });
       val.$__.wasPopulated = { value: unpopulatedValue };
       didPopulate = true;
@@ -1412,7 +1412,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
       this.$populated(path, val.map(function(v) { return v._id; }), popOpts);
 
       for (const doc of val) {
-        doc.$__.wasPopulated = { value: doc._id };
+        doc.$__.wasPopulated = { value: doc._doc._id };
       }
       didPopulate = true;
     }
@@ -3840,7 +3840,7 @@ Document.prototype.$toObject = function(options, json) {
   // _isNested will only be true if this is not the top level document, we
   // should never depopulate the top-level document
   if (depopulate && options._isNested && this.$__.wasPopulated) {
-    return clone(this.$__.wasPopulated.value || this._id, options);
+    return clone(this.$__.wasPopulated.value || this._doc._id, options);
   }
 
   // merge default options with input options.
--- a/lib/document.js
+++ b/lib/document.js
@@ -1393,7 +1393,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
     })();
 
     let didPopulate = false;
-    if (refMatches && val instanceof Document && (!val.$__.wasPopulated || utils.deepEqual(val.$__.wasPopulated.value, val._id))) {
+    if (refMatches && val instanceof Document && (!val.$__.wasPopulated || utils.deepEqual(val.$__.wasPopulated.value, val._doc._id))) {
       const unpopulatedValue = (schema && schema.$isSingleNested) ? schema.cast(val, this) : val._doc._id;
       this.$populated(path, unpopulatedValue, { [populateModelSymbol]: val.constructor });
       val.$__.wasPopulated = { value: unpopulatedValue };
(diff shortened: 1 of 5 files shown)


lib/document.js
```diff
@@ -1394,7 +1394,7 @@ Document.prototype.$set = function $set(path, val, type, options) {
 
     let didPopulate = false;
     if (refMatches && val instanceof Document && (!val.$__.wasPopulated || utils.deepEqual(val.$__.wasPopulated.value, val._id))) {
```
fardolieri, line 1396: Not completely sure, but did you miss the `_doc` in the `deepEqual` check?

```diff
-    if (refMatches && val instanceof Document && (!val.$__.wasPopulated || utils.deepEqual(val.$__.wasPopulated.value, val._id))) {
+    if (refMatches && val instanceof Document && (!val.$__.wasPopulated || utils.deepEqual(val.$__.wasPopulated.value, val._doc._id))) {
```

fardolieri: Otherwise we run into this situation

```js
const owner = new Owner({ name: 'Alice' })
const pet1 = new Pet({ name: 'Kitty1', owner: owner })
const pet2 = new Pet({ name: 'Kitty2', owner: owner })

console.log(typeof pet1.owner) // 'object'
console.log(typeof pet2.owner) // 'string'
```

test/model.populate.test.js
```diff
@@ -11014,4 +11014,35 @@ describe('model: populate:', function() {
...
+    const Pet = db.model('Pet', petSchema);
+
+    const ownerId = new mongoose.Types.ObjectId();
+    const owner = new Owner({
+      _id: ownerId,
+      name: 'Alice'
+    });
+    await owner.save();
+    const pet = new Pet({ name: 'Kitty', owner: owner });
+    await pet.save();
```
hasezoey, line 11043: this could likely be reduced to `.create` calls

hasezoey (commented):
LGTM, minor style suggestion and would like to see the question above resolved

lib/document.js
vkarpov15: Fixed in 1ad7531, thanks for catching this

test/model.populate.test.js
vkarpov15: Fixed in 1ad7531

PR #14764, opened 2024-07-26
Refine query selector to highlight unexpected query keys

alex-statsig:
**Summary**

Previously, there was a catch-all case in RootQuerySelector which would accept any key with type "any". This still allows nice IDE typehints, but can lead to bugs where the incorrect key is used for a filter with no warning that the key is clearly wrong.

With this PR, only keys that contain a "." are unsafely typed as any, which should reduce the blast radius of the hacky types dramatically.

**Examples**
Previously, the following would be a valid type:
```ts
const testDoc: FilterQuery<{ testKey: string }> = {
  invalidKey: 'test',
};
```
Now, the above case gives a type error for "invalidKey".

Note that a case such as this will still be accepted (since its hard to properly type any nested queries):
```ts
const testDoc: FilterQuery<{ testKey: string }> = {
  "invalidKey.test": 'test',
};
```

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -215,9 +215,8 @@ function find() {
   Project.find({});
   Project.find({ name: 'Hello' });
 
-  // just callback
-  // @ts-expect-error: Callback to find is longer supported
-  Project.find((error: CallbackError, result: IProject[]) => console.log(error, result));
+  // just callback; this is no longer supported on .find()
+  expectError(Project.find((error: CallbackError, result: IProject[]) => console.log(error, result)));
 
   // filter + projection
   Project.find({}, undefined);
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -981,10 +981,10 @@ function testWithLevel1NestedPaths() {
 
 function gh14764TestFilterQueryRestrictions() {
   const TestModel = model<{ validKey: number }>('Test', new Schema({}));
-  // @ts-expect-error: A key not in the schema should be invalid
-  TestModel.find({ invalidKey: 0 });
-  // @ts-expect-error: A key not in the schema should be invalid for simple root operators
-  TestModel.find({ $and: [{ invalidKey: 0 }] });
+  // A key not in the schema should be invalid
+  expectError(TestModel.find({ invalidKey: 0 }));
+  // A key not in the schema should be invalid for simple root operators
+  expectError(TestModel.find({ $and: [{ invalidKey: 0 }] }));
 
   // Any "nested" keys should be valid
   TestModel.find({ 'validKey.subkey': 0 });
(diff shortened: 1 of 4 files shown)


alex-statsig:
@hasezoey @IslandRhythms please let me know if there's anything I can do to help get this merged, or if there's any blocking concerns (can't tag reviewers)

hasezoey (commented):
i am partial to this, but i think the new types only allow a single level of nesting.

in any case, to get this merged (not considering if this is even wanted or not), you should probably add some ts tests for this new addition (and fix existing ones). some tests to consider:
- that there is a error on a unknown key
- 1 level nesting should give `any`
- more than 1 level nesting?
- overwrite to allow the old behavior in addition to known properties?

alex-statsig:
@hasezoey Thanks for the feedback!

I've fixed the existing tests by adding a `RootFilterQuery` which supports `Query` and `ObjectId`, which previously didn't have explicit type support (relied on the `[key: string]: any` types afaict). I considered just adding those to `FilterQuery` to avoid changing as much, but I believe its technically not valid to use this in all places `FilterQuery` was used, such as within a `$and` operator.

I've also added some new test cases covering those you mentioned as well as some others I thought would be relevant.

Addressing some comments:
- > i think the new types only allow a single level of nesting.

It actually allows anything containing at least one "." to retain the previous behavior, so multiple levels of nesting should work fine still

- > overwrite to allow the old behavior in addition to known properties?

I'm not sure how I could go about supporting this, and that's my main concern with this change that I'm open to feedback on. Only approach I could see is a separate version of functions with the other types, which feels far too messy. I'm not if this change would need to be delayed until a breaking major version; it seems possible it would cause type errors for many users, but also likely that it would catch many bugs in doing so. I could fallback to implementing this as a patch in my repo's types, but was hoping it could help others more broadly.

test/types/models.test.ts
```diff
@@ -977,3 +978,29 @@ function testWithLevel1NestedPaths() {
...
+  TestModel.find(TestModel.find());
+  // @ts-expect-error: A Query should not be a valid type for a FilterQuery within an op like $and
+  TestModel.find({ $and: [TestModel.find()] });
+
+  const id = new Types.ObjectId();
+  // Any ObjectId should be accepted as the root argument
+  TestModel.find(id);
+  // @ts-expect-error: A ObjectId should not be a valid type for a FilterQuery within an op like $and
+  TestModel.find({ $and: [id] });
+}
```
hasezoey, line 1006: this is not how the type errors are tested, look at other tests.
instead of `@ts-expect-error` comments [`expectError`](https://github.com/Automattic/mongoose/blob/ea2a4ef7b20f73217b82dd21068a2f85bc9b197d/test/types/models.test.ts#L257) should be used (even if your IDE says there is a error)

test/types/models.test.ts
```diff
@@ -216,6 +216,7 @@ function find() {
   Project.find({ name: 'Hello' });
 
   // just callback
+  // @ts-expect-error: Callback to find is longer supported
   Project.find((error: CallbackError, result: IProject[]) => console.log(error, result));
```
hasezoey, line 220: i dont know why no `expectError` was here previously (maybe because `any` somehow allowing it?), but maybe remove the comment again

alex-statsig: Ah sorry, forgot to update this one, thanks

Yeah the type `{ [key: string]: any }` (which is what `FilterQuery` basically reduced to before) is super permissive, so a function actually passes it (as well as `ObjectId` and `Query` which were supposed to be valid, but weren't very explicitly typed to be valid)

vkarpov15 (approved):
LGTM. I think this can be a reasonable compromise between Mongoose's flexible queries and strict type checking. In the interest of caution, I'll merge this into 8.6 branch though. We may end up having to revert this if it causes unforeseen consequences.

vkarpov15:
@alex-statsig it looks like there's some test failures against 8.6 branch, the crux of the issue is likely 8.6 changing the `RawDocType` default in `class Query<ResultType, DocType, THelpers = {}, RawDocType = DocType, QueryOp = 'find', TInstanceMethods = Record<string, never>> implements SessionOperation {` to `RawDocType = unknown`.

I think the fix would be to just update queryhelpers.test.ts to correctly type `type ProjectModelQuery = Query<any, HydratedDocument<Project>, ProjectQueryHelpers>` -> `type ProjectModelQuery = Query<any, HydratedDocument<Project>, ProjectQueryHelpers, any>`

PR #14769, opened 2024-07-29
fix: allow setting document array default to `null`

### LINKED ISSUE #6691: Unable to set array to `null` as default

vkarpov15:
Fix #6691
Re: #14717

**Summary**

Looks like #14717 missed a spot, we still need a change to support `default: null` for document arrays

**Examples**

### DIFF
--- a/lib/schema/documentArray.js
+++ b/lib/schema/documentArray.js
@@ -69,7 +69,7 @@ function SchemaDocumentArray(key, schema, options, schemaOptions) {
 
   const fn = this.defaultValue;
 
-  if (!('defaultValue' in this) || fn !== void 0) {
+  if (!('defaultValue' in this) || fn != null) {
     this.default(function() {
       let arr = fn.call(this);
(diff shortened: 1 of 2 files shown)


IslandRhythms (approved):
LGTM

PR #14772, opened 2024-08-01
feat(model+query): support `options` parameter for distinct()

### LINKED ISSUE #8006: Support options arg for `distinct()`: Model.distinct() doesn't have { session }
Unlike other methods `distinct()` doesn't have { session } option.
How to workaround it?

vkarpov15:
Fix #8006

**Summary**

Make it easier to use `distinct()` with transactions, so you don't need to chain `setOptions()`. `distinct()` looks to be the only query function where you can't pass in `options`.

**Examples**

### DIFF
--- a/test/docs/transactions.test.js
+++ b/test/docs/transactions.test.js
@@ -338,6 +338,26 @@ describe('transactions', function() {
     assert.deepEqual(fromDb, { name: 'Tyrion Lannister' });
   });
 
+  it('distinct (gh-8006)', async function() {
+    const Character = db.model('gh8006_Character', new Schema({ name: String, rank: String }, { versionKey: false }));
+
+
+    const session = await db.startSession();
+
+    session.startTransaction();
+    await Character.create([{ name: 'Will Riker', rank: 'Commander' }, { name: 'Jean-Luc Picard', rank: 'Captain' }], { session });
+
+    let names = await Character.distinct('name', {}, { session });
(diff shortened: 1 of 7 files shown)


test/docs/transactions.test.js
```diff
@@ -338,6 +338,26 @@ describe('transactions', function() {
     assert.deepEqual(fromDb, { name: 'Tyrion Lannister' });
   });
 
+  it('distinct (gh-8006)', async function() {
+    const Character = db.model('gh8006_Character', new Schema({ name: String, rank: String }, { versionKey: false }));
+
+
+    const session = await db.startSession();
```
hasezoey, line 345:
```suggestion
    const Character = db.model('gh8006_Character', new Schema({ name: String, rank: String }, { versionKey: false }));

    const session = await db.startSession();
```

i dont know if this intended, but i think a single new line would be enough

hasezoey (changes requested):
LGTM, but the `options` parameter still needs to be added to the query types (currently only in the model's types) and a similar test to `using options parameter for distinct` on query should likely be in for Model too, or is the transaction test enough?

As a note, the `Model.distinct` and `Query.prototype.distinct` have differing `options` checks, is this intentional?

vkarpov15:
Thanks for the suggestions @hasezoey , I've implemented them both. Added `options` param to query `distinct()` types and removed the `typeof options === 'object'` check in Query.prototype.distinct (and countDocuments() + estimatedDocumentCount() for consistency)

hasezoey (approved):

PR #14790, opened 2024-08-02
docs: add schema type widening advice

JstnMcBrd:
### Summary

While using Mongoose with TypeScript, @ShunnShine and I discovered the dangers of defining the schema in a separate variable when we realized all of our fields with `required: false` were being treated as `required: true`.

We eventually figured out how to fix the type definitions by using `as const`.

We looked at the [Schemas in TypeScript](https://mongoosejs.com/docs/typescript/schemas.html) documentation and saw a warning about assigning a schema to a temporary variable (fair enough), but did not see any information about using `as const`, even though it is used in the example code right below - without explanation.

This PR adds a little section about type widening and `as const` to help users understand why this was done.


vkarpov15 (approved):
Thanks :+1:

Issue #14794, opened 2024-08-07
Populate + lean + multiple populated documents leads to "Cannot read properties of undefined (reading '_id')" in version 8.5.2


Systerr:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.5.2

### Node.js version

22

### MongoDB server version

7.0

### Description

Populating with 'lean' and more then one model in a populated object cause in error in v8.5.2. All works fine i v.8.5.1. Without lean() all works fine too

```javascript
Error: TypeError: Cannot read properties of undefined (reading '_id')
    at _assign (//node_modules/mongoose/lib/model.js:4471:38)
    at _done (/node_modules/mongoose/lib/model.js:4300:9)
    at _next (//node_modules/mongoose/lib/model.js:4288:7)
    at //node_modules/mongoose/lib/model.js:4396:7
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Looks like it happens because of https://github.com/Automattic/mongoose/commit/7c2cfa898f04edd566220c04b1c46bffe9864b3a#diff-305dbcc98ad8fc959dd0c0be51280893962af34c2d8b18abc871cffc2d80c5f9R4471 as part of fix https://github.com/Automattic/mongoose/issues/14759

### Steps to Reproduce

This an abstract sample how to reproduce it

```javascript
import mongoose from "mongoose";

await mongoose.connect("mongodb://localhost:27017/blogdb");

const authorSchema = new mongoose.Schema({
  group: String, // our populate key
  name: String,
});

const postSchema = new mongoose.Schema({
  authorGroup: String,
  title: String,
  content: String,
});

const Author = mongoose.model("Author", authorSchema);
const Post = mongoose.model("Post", postSchema);

async function createTestData() {
  await Author.create({ group: "AUTH1", name: "John Doe" });
  await Author.create({ group: "AUTH2", name: "Jane Smith" });
  await Author.create({ group: "AUTH2", name: "Will Jons" }); // second author with the same group (it will cause an issue). One populated record works just fine

  await Post.create({
    authorGroup: "AUTH1",
    title: "First Post",
    content: "Content 1",
  });
  await Post.create({
    authorGroup: "AUTH2",
    title: "Second Post",
    content: "Content 2",
  });
}

async function runProblemQuery() {
  try {
    const posts = await Post.find()
      .populate({
        path: "authorGroup",
[...]

PR #14799, opened 2024-08-09
fix populate called with lean (gh-14794)

MohOraby:
**Summary**

in #14794 due to being called with `lean()` there's no `_doc` object so that throws an error

another issue here is documents with duplicate foreignField so the `rawDocs[key]` already exists and its value is being overwritten.

this change shouldn't affect (gh-14759) and its test case still runs successfully.

another solution would be to check for `rawDocs[key]._id`  and `val._doc` first using the following statement instead
```js
(rawDocs[key]._doc && String(rawDocs[key]._doc._id)) !== (val._doc && String(val._doc._id)))

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -4459,7 +4459,7 @@ function _assign(model, vals, mod, assignmentOpts) {
         }
       } else {
         if (_val instanceof Document) {
-          _val = _val._doc._id;
+          _val = _val._id;
         }
         key = String(_val);
         if (rawDocs[key]) {
@@ -4468,7 +4468,7 @@ function _assign(model, vals, mod, assignmentOpts) {
             rawOrder[key].push(i);
           } else if (isVirtual ||
             rawDocs[key].constructor !== val.constructor ||
-            String(rawDocs[key]._doc._id) !== String(val._doc._id)) {
+            String(rawDocs[key]._id) !== String(val._id)) {
             // May need to store multiple docs with the same id if there's multiple models
             // if we have discriminators or a ref function. But avoid converting to an array
             // if we have multiple queries on the same model because of `perDocumentLimit` re: gh-9906
--- a/lib/model.js
+++ b/lib/model.js
@@ -4459,7 +4459,7 @@ function _assign(model, vals, mod, assignmentOpts) {
         }
       } else {
         if (_val instanceof Document) {
-          _val = _val._id;
+          _val = _val._doc._id;
         }
         key = String(_val);
         if (rawDocs[key]) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/model.js [hunk 1]
```diff
@@ -4459,7 +4459,7 @@ function _assign(model, vals, mod, assignmentOpts) {
         }
       } else {
         if (_val instanceof Document) {
-          _val = _val._doc._id;
+          _val = _val._id;
```
vkarpov15, line 4462: This change isn't necessary. L4462 will never run if lean.

lib/model.js
```diff
@@ -4468,7 +4468,7 @@ function _assign(model, vals, mod, assignmentOpts) {
             rawOrder[key].push(i);
           } else if (isVirtual ||
             rawDocs[key].constructor !== val.constructor ||
-            String(rawDocs[key]._doc._id) !== String(val._doc._id)) {
+            String(rawDocs[key]._id) !== String(val._id)) {
```
vkarpov15, line 4471: Can we expand this change to check if `rawDocs[key]` and `val` are `instanceof Document`, and, if so, use `val._doc._id`, otherwise `val._id`?

vkarpov15 (commented):
A couple of quick changes, #14759 made us very wary of cases where we may accidentally be checking for quality on `_id` after getters applied, so we want to avoid undoing the `val._doc._id` changes in cases where we know `val instanceof Document`.

Thanks for finding the issue

vkarpov15 (changes requested):
See previous comments, I accidentally submitted review with "Comment" selected rather than "Request changes"

lib/model.js
MohOraby: Ok, for this would you prefer it be a ternary operator inside the else if check like this
```js
(rawDocs[key] instanceof Document ? String(rawDocs[key]._doc._id) : String(rawDocs[key]._id)) !== (val instanceof Document ? String(val._doc._id) : String(val._id))
```
which looks kind of long and hard to read or move them to constant above the if check like this
```js
const keyDocId = rawDocs[key] instanceof Document ? rawDocs[key]._doc._id : rawDocs[key]._id;
const valDocId = val instanceof Document ? val._doc._id : val._id;
```
both are fine and would do the same thing, I just don't wanna keep pushing commits since the CI would rerun

MohOraby: I went for the the first approach currently, just to move things forward, if it needs refactoring please let me know

vkarpov15 (approved):
Thanks :+1:

PR #14802, opened 2024-08-12
feat: `useConnection(connection)` function

IslandRhythms:


test/model.test.js
```diff
@@ -7575,6 +7575,16 @@ describe('Model', function() {
 
     assert.strictEqual(doc.__v, 0);
   });
+  it('updates the model\'s db property to point to the provided connection instance and vice versa', async function() {
+    const schema = new mongoose.Schema({
+      name: String
+    });
+    const Model = db.model('Test', schema);
+    const connection = start();
```
vkarpov15, line 7583: We should expand this test to use a different uri than `db`, and assert that `Model.find()` executes queries against the new uri rather than the old uri. 

As written, I don't think that will work because on further review we'll need to also update the model's collection properties (see `Model.compile()`)

```
  const collection = connection.collection(
    collectionName,
    collectionOptions
  );

  model.prototype.collection = collection;
  model.prototype.$collection = collection;
  model.prototype[modelCollectionSymbol] = collection;
```

lib/model.js
```diff
@@ -146,6 +146,21 @@ Model.prototype.$isMongooseModelPrototype = true;
...
+
+Model.useConnection = function useConnection(connection) {
+  if (!connection) {
+    throw new Error('Please provide a connection.');
+  }
+  if (this.db) {
+    delete this.db.models[this.modelName];
+  }
+  this.db = connection;
+  connection.models[this.modelName] = this;
```
vkarpov15, line 161: We'll need to test this a bit more thoroughly, it looks like we also need to modify the model's collection as well

vkarpov15 (changes requested):
More testing required. Also please add TypeScript type definition and TypeScript test please

vkarpov15 (approved):

lib/model.js
```diff
@@ -148,6 +148,62 @@ Model.prototype.$isMongooseModelPrototype = true;
...
+    delete this.prototype.db;
+    delete this.prototype[modelDbSymbol];
+    delete this.prototype.collection;
+    delete this.prototype.$collection;
+    delete this.prototype[modelCollectionSymbol];
+  }
+
+  this.db = connection;
+  const collection = connection.collection(this.modelName, connection.options);
+  this.prototype.collection = collection;
```
meshde-flux, line 195: I believe this makes it such that the name of the collection is always set to the modelName which would be different from the name of the collection assigned initially, which could either be a user-defined collection name or one that is automatically generated by mongoose using the `utils.toCollectionName` as mentioned here: https://mongoosejs.com/docs/api/connection.html#Connection.prototype.model().

This leads to the case where a new collection is created and used once `useConnection` is invoked on the model. Is that intentional?

vkarpov15: That's a good point @meshde-flux , I will open up new issue

Issue #10592, opened 2021-08-23
Depopulate not working for populated docs inside array


LiaanM:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
A schema containing an array of documents with a nested reference inside of the document in the array does not depopulate correctly.

**If the current behavior is a bug, please provide the steps to reproduce.**
The following is a test to reproduce the error.  The `band.members` and `band.lead` is depopulated correctly, but the `band.embeddedMembers` are not. 

    it('depopulate all should depopulate nested array population', async function() {

         const Person = db.model('Person', {
              name: String
          });

          const Band = db.model('Band', {
              name: String,
              members: [{ type: Schema.Types.ObjectId, ref: 'Person' }],
              lead: { type: Schema.Types.ObjectId, ref: 'Person' },
              embeddedMembers: [{
                  active: Boolean,
                  member :{
                      type: Schema.Types.ObjectId, ref: 'Person'
                  }
              }]
          });

          const people = [{ name: 'Axl Rose' }, { name: 'Slash' }];
          try {
              let docs = await Person.create(people);
              let band = {
                  name: 'Guns N\' Roses',
                  members: [docs[0]._id, docs[1]],
                  lead: docs[0]._id,
                  embeddedMembers: [{active:true,member:docs[0]._id},{active:false, member:docs[1]._id}]
              };
              band = await Band.create(band);
              await band.populate('members lead').populate('embeddedMembers.member').execPopulate();
              assert.ok(band.populated('members'));
              assert.ok(band.populated('lead'));
              assert.ok(band.populated('embeddedMembers.member'));
              assert.equal(band.members[0].name, 'Axl Rose');
              assert.equal(band.embeddedMembers[0].member.name, 'Axl Rose');
              band.depopulate();
              assert.ok(!band.populated('members'));
              assert.ok(!band.populated('lead'));
              assert.ok(!band.populated('embeddedMembers.member'));
              assert.ok(!band.embeddedMembers[0].member.name);
              return null
          } catch (err) {
              /* handle error */
              assert.ifError(err);
              return err;
          }
      });

**What is the expected behavior?**
The array with nested population should depopulate.  
I suspect the $set is not taking into account the array indexes here : https://github.com/Automattic/mongoose/blob/a803885a49cb1f0498181263e140e8b688c3e145/lib/document.js#L4206

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
[...]

Issue #1635, opened 2013-08-13
Document.array.pull() does not seem to work for populated arrays when pulling an _id string


JedWatson:
I have the following function designed to update the ObjectIds stored in a given document `doc`, at `path`, to match those provided in a form submission `data` by using atomic `push` and `pull` functionality.

(ObjectIds are stored in an array like this: `schema.add({ myPath: [{type: ObjectId, ref: 'OtherSchema'}] });`

```
function(doc, data, path) {

    var storedValue = doc.populated(path),
        arr = doc.get(path),
        _old = storedValue.map(function(i) { return String(i) }),
        _new = _.compact(data[path].split(','));

    console.log('Stored Value:');
    console.log(storedValue)

    console.log('Data at Path:');
    console.log(arr);

    console.log('Remove:');
    console.log(_.difference(_old, _new));

    // remove ids
    _.difference(_old, _new).forEach(function(val) {
        console.log('pulling ' + val);
        arr.pull(val);
    });

    console.log('Add:');
    console.log(_.difference(_new, _old));

    // add new ids
    _.difference(_new, _old).forEach(function(val) {
        console.log('pushing ' + val);
        arr.push(val);
    });

    console.log('Updated Value:');
    console.log(arr);

}
```

If the `doc` has already had `path` populated, two unexpected behaviours occur:
1. ObjectIds aren't removed
2. ObjectIds that are added exist in a half-populated state

`Console.log` output demonstrating both in the above function:

```
Stored:
[ 51f9fe338507730000000004 ]
Item Data:
[ { name: 'Jimmy',
  _id: 51f9fe338507730000000004 } ]
Remove:
[ '51f9fe338507730000000004' ]
pulling 51f9fe338507730000000004
Add:
[ '51f9fe338507730000000007' ]
pushing 51f9fe338507730000000007
Updated:
[ { name: 'Jimmy',
  _id: 51f9fe338507730000000004 }, { _id: 51f9fe338507730000000007 } ]
```

re: 1, this seems like a bug. The documentation [here](http://mongoosejs.com/docs/api.html#types_array_MongooseArray-pull) implies that you should be able to remove ids from an array, although it doesn't mention what happens if the array has been populated.

re: 2, as you can see the newly added object _looks_ populated (is an object) but hasn't been loaded from the database yet (so has no name property). What's the best way to handle this? Repopulate the path after saving the document?

[...]

PR #14803, opened 2024-08-12
fix(document): make sure depopulate does not convert hydrated arrays to vanilla arrays

vkarpov15:
**Summary**

One issue that popped up while I was working on #1635 is that `depopulate()` converts Mongoose arrays to vanilla JavaScript arrays. So `band.depopulate('members'); band.members.addToSet(id);` will error out because `band.members` is no longer a Mongoose array.

The approach I came up with is to use `getValue()`: if `getValue()` returns a Mongoose array, then we just modify the array in-place to avoid triggering setters. If not a Mongoose array, then we have a case where mpath is returning an array of values underneath a document array, like in the #10592 test case `doc.depopulate('docArr.members')`, so we fall back to `setValue()`. 

I left `assert.ok(!band.embeddedMembers[0].$populated('member'));` commented out because that assertion fails, but I want to look at it some more when we come back to #1635.

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -10805,7 +10805,10 @@ describe('document', function() {
     assert.ok(!band.populated('members'));
     assert.ok(!band.populated('lead'));
     assert.ok(!band.populated('embeddedMembers.member'));
+    assert.ok(band.members.isMongooseArray);
+    assert.ok(band.embeddedMembers.isMongooseArray);
     assert.ok(!band.embeddedMembers[0].member.name);
+    // assert.ok(!band.embeddedMembers[0].$populated('member'));
   });
 
   it('should allow dashes in the path name (gh-10677)', async function() {
--- a/lib/document.js
+++ b/lib/document.js
@@ -4774,7 +4774,23 @@ Document.prototype.depopulate = function(path) {
         continue;
       }
       delete populated[key];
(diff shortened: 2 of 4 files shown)


test/document.test.js
```diff
@@ -10805,7 +10805,10 @@ describe('document', function() {
     assert.ok(!band.populated('members'));
     assert.ok(!band.populated('lead'));
     assert.ok(!band.populated('embeddedMembers.member'));
+    assert.ok(band.members.isMongooseArray);
+    assert.ok(band.embeddedMembers.isMongooseArray);
     assert.ok(!band.embeddedMembers[0].member.name);
+    // assert.ok(!band.embeddedMembers[0].$populated('member'));
```
hasezoey, line 10811: > I left assert.ok(!band.embeddedMembers[0].$populated('member')); commented out because that assertion fails, but I want to look at it some more when we come back to https://github.com/Automattic/mongoose/issues/1635.

was this working before the PR?

hasezoey (approved):

test/document.test.js
vkarpov15: I confirmed this assertion fails before this PR. We will need to do some extra work to make sure `depopulate()` also updates subdocs' `$populated()`.

PR #14808, opened 2024-08-15
docs: upgrade marked -> 14.x, use new {#id-here} syntax for declaring custom header ids

### LINKED ISSUE #13211: [Docs] Proposal of using custom header ids without needing html tags
### Prerequisites

### Issue

This issue is a proposal to convert current manual headers (`<h2 id="test-id"><a href="#test-id>test`) to markdown headers with custom id (`## test {#test-id}`).

This should make the documentation files themself be more readable and editors (like vscode) that can use "jump to heading" will then work with the markdown headers.

Downside is that this change would not be compatible with using the documentation files raw in github (like directly viewing [this file](https://github.com/Automattic/mongoose/blob/master/docs/migrating_to_7.md), the headers would appear as `strictQuery {#strictquery}` if written like `## strictQuery {#strictquery}`).

For example documentation frameworks like `docusaurus` support this.

[Example branch with changes to the generation and some files](https://github.com/hasezoey/mongoose/tree/customHeaderId), this branch includes:
- the changes to the header generation to support this change
- some updated files to show the difference (most notably `docs/migrate_to_7.md`)

cc @vkarpov15 @AbdelrahmanHafez  @Uzlopak

vkarpov15:
Fixes #13211

**Summary**

@hasezoey I really like your suggestion from #13211 on further research. In this PR I:

1. Upgraded marked to v14 and used their new renderer API to handle headers
2. Inlined the logic we use to convert header text to slugs by default. This will make it so that we don't have to worry as much about breaking links when upgrading dependencies
3. Moved about half of the inlined `<h2>` elements to the new syntax

I did part of the migration using the following script:

```js
const cheerio = require('cheerio');

const handle3 = (line) => { const $ = cheerio.load(line); return `## ${($('h2')).text()} {#${$('h2').attr('id')}}`; };

const str = require('fs').readFileSync('./docs/migrating_to_5.md').toString('utf8');

const lines = str.split('\n');

console.log(lines.map(line => line.startsWith('<h2') ? handle3(line) : line).join('\n'));

```

**Examples**


scripts/website.js
```diff
@@ -10,11 +10,16 @@ const pug = require('pug');
...
-require('./generateSearch');
+
 // using "__dirname" and ".." to have a consistent CWD, this script should not be runnable, even when not being in the root of the project
 // also a consistent root path so that it is easy to change later when the script should be moved
 const cwd = path.resolve(__dirname, '..');
 
+// support github-like custom header ids
+// Example:
+// # Some Header {#custom-id}
+const CustomIdRefex = /{#([a-z][\w-]*)}(?: *)$/;
```
hasezoey, line 21:
```suggestion
const CustomIdRefex = /{#([a-zA-Z0-9_-]+)}(?: *)$/;
```

add `_` as otherwise `guide.md`'s `option: _id` does not get recognized, also expand `\w` to be more explicit in what is supported and only have one range (as we are in control of both processing and writing the headings)

hasezoey (changes requested):
Looks good for the most part, some suggestions regarding typos and general suggestions i came across while looking at the changes.

also some other to-dos:
- when also touching `migrationg_to_5`, we should touch all the `migrating_to_*.md` files
- also apply the same for and `h3`
- some files seem to have been missed in the translation (like `docs/subdocs.md`, `docs/schematypes.md` and `docs/tutorials/` and more)

scripts/website.js
```diff
@@ -600,3 +617,12 @@ if (isMain) {
...
 }
+
+// Modified from github-slugger
+function createSlug(value) {
+  if (typeof value !== 'string') {
+    return '';
+  }
+  value = value.toLowerCase();
+  return value.replace(/[^a-z0-9-_\s]/, '').replace(/ /g, '-');
+}
```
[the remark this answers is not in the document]
vkarpov15: Yeah default slugger was removed in marked@9 apparently. Recommendation is to switch to github-slugger npm package. See discussion from this issue: https://github.com/markedjs/marked/issues/2983.

I copied most of the core logic from github-slugger into this function. github-slugger is more sophisticated, but I wanted to inline our slugging logic because I don't want to have to go through every single header every time we have to upgrade github-slugger.

docs/queries.md
```diff
@@ -97,11 +97,7 @@ await Person.
 
 A full list of [Query helper functions can be found in the API docs](api/query.html).
 
-<h2 id="queries-are-not-promises">
-  <a href="#queries-are-not-promises">
-    Queries are Not Promises
-  </a>
-</h2>
+## Queries are Not Promises
```
vkarpov15, line 100: I confirmed that the custom id isn't necessary here

docs/migrating_to_7.md
```diff
@@ -276,7 +276,7 @@ oid._bsontype; // 'ObjectId' in Mongoose 7, 'ObjectID' in older versions of Mong
 Please update any places where you use `_bsontype` to check if an object is an ObjectId.
 This may also affect libraries that use Mongoose.
 
-<h2 id="removed-mapreduce"><a href="#removed-mapreduce">Removed mapReduce</a></h2>
+## Removed mapReduce {#removed-mapreduce}
```
hasezoey, line 279:
```suggestion
## Removed `mapReduce` {#removed-mapreduce}
```

not in original but would look better

docs/migrating_to_6.md
```diff
@@ -130,7 +130,7 @@ await q;
 await q.clone(); // Can `clone()` the query to allow executing the query again
 ```
 
-<h2 id="model-exists-returns-a-lean-document-instead-of-boolean"><a href="#model-exists-returns-a-lean-document-instead-of-boolean">Model.exists(...) now returns a lean document instead of boolean</a></h2>
+## Model.exists(...) now returns a lean document instead of boolean {#model-exists-returns-a-lean-document-instead-of-boolean}
```
hasezoey, line 133:
```suggestion
## `Model.exists(...)` now returns a lean document instead of boolean {#model-exists-returns-a-lean-document-instead-of-boolean}
```

not in original but would look better

docs/queries.md
hasezoey: other places also define still define even if not necessary currently, maybe this should then also be like the others for now?

docs/guide.md
```diff
@@ -727,7 +729,7 @@ const p = new Page({ name: 'mongodb.org' });
 console.log(p.id); // undefined
 ```
 
-<h2 id="_id"><a href="#_id">option: _id</a></h2>
+## option: _id {#_id}
```
hasezoey, line 732: duplicate id with `## Ids {#_id}`, maybe we should change one of them?

hasezoey (changes requested):
Some more suggestions and still some files missing:
- `docs/middleware.md` (h3)
- `docs/subdocs.md` (h3)
- `docs/tutorials/`
  - `findoneandupdate.md` (h2)
  - `lean.md` (h2)

docs/guide.md
vkarpov15: Good catch, fixed in edd94fc78ea9edd506a931f39489c571050cb91f

vkarpov15:
@hasezoey re: "other places also define still define even if not necessary currently", that's a reasonable suggestion, I'll write a script to do that

Also, I left one of the h2's in `findOneAndUpdate.md` in legacy style. That is because the h2 `<h2 id="includeresultmetadata">The <code>includeResultMetadata</code> Option<h2 id="rawresult"></h2></h2>` actually has 2 h2s so as not to break links to `findoneandupdate#rawresult`. I'll keep the old syntax there for now because of the exception, but the rest should be fixed.

hasezoey (approved):
looks good now, only one h2 remaining, but with reason.

aside from that, while going through this i found some issues unrelated to this PR, see #14822.

PR #14812, opened 2024-08-17
fix(connection): avoid returning `readyState = connected` if connection state is stale

### LINKED ISSUE #14727: Improve readyState behavior to more reliably report connection state
### Prerequisites

### 🚀 Feature Proposal

Since `readyState` updates are based on topology status updates, it's less reliable in stateless environments (such as AWS Lamba) as topology update events may be missed while the execution environment is frozen.

This can result in `readyState` tests returning stale state information.

### Motivation

_No response_

### Example

_No response_

vkarpov15:
Fix #14727

cc @alexbevi 

**Summary**

The problem here is that `readyState` can return `connected` immediately after an AWS Lambda container thaws after some time. `readyState` only looks at the last heartbeat, but not how long ago the last heartbeat occurred - with this PR, if the last heartbeat is stale (more than 2 heartbeat intervals) then we assume we're disconnected.

This PR still needs significant testing, but I wanted to put it out for review to solicit early feedback.

**Examples**

### DIFF
--- a/lib/drivers/node-mongodb-native/connection.js
+++ b/lib/drivers/node-mongodb-native/connection.js
@@ -23,6 +23,11 @@ const utils = require('../../utils');
 function NativeConnection() {
   MongooseConnection.apply(this, arguments);
   this._listening = false;
+  // Tracks the last time (as unix timestamp) the connection received a
+  // serverHeartbeatSucceeded or serverHeartbeatFailed event from the underlying MongoClient.
+  // If we haven't received one in a while (like due to a frozen AWS Lambda container) then
+  // `readyState` is likely stale.
+  this._lastHeartbeatAt = null;
 }
 
 /**
@@ -106,6 +111,7 @@ NativeConnection.prototype.useDb = function(name, options) {
       _opts.noListener = options.noListener;
     }
     newConn.db = _this.client.db(name, _opts);
+    newConn._lastHeartbeatAt = _this._lastHeartbeatAt;
     newConn.onOpen();
   }
 
@@ -409,6 +415,12 @@ function _setClient(conn, client, options, dbName) {
       }
     });
   }
+  client.on('serverHeartbeatSucceeded', () => {
+    conn._lastHeartbeatAt = Date.now();
+  });
+  client.on('serverHeartbeatFailed', () => {
+    conn._lastHeartbeatAt = Date.now();
+  });
 
   if (options.monitorCommands) {
     client.on('commandStarted', (data) => conn.emit('commandStarted', data));
@@ -417,6 +429,9 @@ function _setClient(conn, client, options, dbName) {
   }
 
   conn.onOpen();
+  if (client.topology?.s?.state === 'connected') {
+    conn._lastHeartbeatAt = Date.now();
+  }
 
   for (const i in conn.collections) {
     if (utils.object.hasOwnProperty(conn.collections, i)) {
--- a/lib/drivers/node-mongodb-native/connection.js
+++ b/lib/drivers/node-mongodb-native/connection.js
@@ -418,9 +418,6 @@ function _setClient(conn, client, options, dbName) {
   client.on('serverHeartbeatSucceeded', () => {
     conn._lastHeartbeatAt = Date.now();
   });
-  client.on('serverHeartbeatFailed', () => {
-    conn._lastHeartbeatAt = Date.now();
-  });
 
   if (options.monitorCommands) {
     client.on('commandStarted', (data) => conn.emit('commandStarted', data));
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -103,6 +103,15 @@ Object.setPrototypeOf(Connection.prototype, EventEmitter.prototype);
 
 Object.defineProperty(Connection.prototype, 'readyState', {
   get: function() {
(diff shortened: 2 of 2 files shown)


lib/drivers/node-mongodb-native/connection.js
```diff
@@ -409,6 +415,12 @@ function _setClient(conn, client, options, dbName) {
       }
     });
   }
+  client.on('serverHeartbeatSucceeded', () => {
+    conn._lastHeartbeatAt = Date.now();
+  });
+  client.on('serverHeartbeatFailed', () => {
+    conn._lastHeartbeatAt = Date.now();
+  });
```
hasezoey, line 423: should this really count as a heartbeat?

hasezoey (approved):
Code looks good to me, though i agree this should need some testing

lib/drivers/node-mongodb-native/connection.js
vkarpov15: It shouldn't matter because heartbeat failed would make Mongoose readyState `disconnected`. Which is a good argument for removing this code, because it doesn't have any impact. Thanks for the suggestion :+1:

vkarpov15:
Going to move this one back because I've been having trouble reproducing this issue in AWS lambda

vkarpov15:
It took some time, but I was able to repro this in Lambda. The following function is adapted from Mongoose's Lambda docs:

```

let conn = null;

const uri = 'mongodb+srv://chris2022:PennywinklePlaydays@zevo-dev.6lxkgiz.mongodb.net/chris2022?retryWrites=true&w=majority';

exports.handler = async function(event, context) {
  // Make sure to add this so you can re-use `conn` between function calls.
  // See https://www.mongodb.com/blog/post/serverless-development-with-nodejs-aws-lambda-mongodb-atlas
  context.callbackWaitsForEmptyEventLoop = false;

  // Because `conn` is in the global scope, Lambda may retain it between
  // function calls thanks to `callbackWaitsForEmptyEventLoop`.
  // This means your Lambda function doesn't have to go through the
  // potentially expensive process of connecting to MongoDB every time.
  if (conn == null) {
    conn = mongoose.createConnection(uri, {
      // and tell the MongoDB driver to not wait more than 5 seconds
      // before erroring out if it isn't connected
      serverSelectionTimeoutMS: 5000
    });

    // `await`ing connection after assigning to the `conn` variable
    // to avoid multiple function calls creating new connections
    await conn.asPromise();
    conn.model('Test', new mongoose.Schema({ name: String }));
  }

  console.log('Start', conn.readyState, new Date(conn._lastHeartbeatAt), Date.now() - conn._lastHeartbeatAt);

  const M = conn.model('Test');

  const doc = await M.findOne();
  console.log('After find', conn.readyState, new Date(conn._lastHeartbeatAt), Date.now() - conn._lastHeartbeatAt);
  console.log(doc);

  return doc;
};

```

Output:

```

Response
{
  "_id": "641b1cb82525faca0c0be001",
  "name": "Run Hook",
  "__v": 0
}

Function Logs
START RequestId: ec6e7091-f2dd-47f8-8a4a-8cee7cccc1fa Version: $LATEST
2024-09-17T16:36:30.288Z	ec6e7091-f2dd-47f8-8a4a-8cee7cccc1fa	INFO	Start 0 2024-09-17T16:36:08.749Z 21539
2024-09-17T16:36:30.408Z	ec6e7091-f2dd-47f8-8a4a-8cee7cccc1fa	INFO	After find 1 2024-09-17T16:36:30.371Z 37
2024-09-17T16:36:30.408Z	ec6e7091-f2dd-47f8-8a4a-8cee7cccc1fa	INFO	{
  _id: new ObjectId('641b1cb82525faca0c0be001'),
  name: 'Run Hook',
  __v: 0
}
END RequestId: ec6e7091-f2dd-47f8-8a4a-8cee7cccc1fa
REPORT RequestId: ec6e7091-f2dd-47f8-8a4a-8cee7cccc1fa	Duration: 204.32 ms	Billed Duration: 205 ms	Memory Size: 128 MB	Max Memory Used: 104 MB

Request ID
ec6e7091-f2dd-47f8-8a4a-8cee7cccc1fa
```

So `readyState` is correctly 0 at the beginning, but then changes to 1 when Mongoose gets a successful heartbeat back.

halilbrhmtrn:
`    if (
      this._readyState === STATES.connected &&
      this._lastHeartbeatAt != null &&
      typeof this.client?.topology?.s?.description?.heartbeatFrequencyMS === 'number' &&
      Date.now() - this._lastHeartbeatAt >= this.client.topology.s.description.heartbeatFrequencyMS * 2) {
      return STATES.disconnected;
    }`
    
This logic seems overly aggressive. Network latencies caused by high load on the MongoDB server could prevent heartbeats from being received within the heartbeatFrequencyMS * 2 time frame. Relying on such logic may lead Mongoose to falsely report a disconnected state. If this implementation is necessary for cold-start environments, it should be configurable and invoked only when required. Additionally, a more reliable method for detecting staleness should be employed. FYI
@vkarpov15

halilbrhmtrn:
It seems there are already couple of issues opened related with this. @vkarpov15 
https://github.com/Automattic/mongoose/issues/15042
https://github.com/Automattic/mongoose/issues/15015
https://github.com/Automattic/mongoose/issues/14971

PR #14819, opened 2024-08-19
types: make toObject() and toJSON() not generic by default to avoid type widening

### LINKED ISSUE #12883: TypeScript does not complain about type issues when returning `doc.toObject()`
### Prerequisites

### Mongoose version

6.8.3

### Node.js version

n/a

### MongoDB server version

n/a

4.9.4

### Description

I am having a hard time understanding why TypeScript complains in `getUser_1` and `getUser_3` seen below, but not in `getUser_2`.

Is this an issue with mongoose?

### Steps to Reproduce

```ts

    name: String,
})

const UserModel = mongoose.model('User', userSchema)

type GetUserOutput = {
    _id: string
    firstname?: string
}

async function getUser_1(): Promise<GetUserOutput> {
    const user = await UserModel.findOne({}).exec()

    if (user === null) throw new Error(`No user found.`)

    return user // TypeScript complains:
    // Type 'Document<unknown, any, { name?: string | undefined; }> & { name?: string | undefined; } & { _id: ObjectId; }' is not assignable to type 'GetUserOutput'.
    //   Types of property '_id' are incompatible.
    //     Type 'ObjectId' is not assignable to type 'string'.ts(2322)
}

async function getUser_2(): Promise<GetUserOutput> {
    const user = await UserModel.findOne({}).exec()

    if (user === null) throw new Error(`No user found.`)

    return user.toObject() // TypeScript does not complain.
}

async function getUser_3(): Promise<GetUserOutput> {
    const user = await UserModel.findOne({}).exec()

    if (user === null) throw new Error(`No user found.`)

    const o = user.toObject()

    return o // TypeScript complains:
    // Type 'LeanDocument<{ name?: string | undefined; }> & { _id: ObjectId; }' is not assignable to type '{ firstname?: string | undefined; _id: string; }'.
    //   Types of property '_id' are incompatible.
    //     Type 'ObjectId' is not assignable to type 'string'.ts(2322)
}
```

### Expected Behavior

TypeScript should complain in `getUser_2`, too.

vkarpov15:
Fix #12883

**Summary**

Making document methods generic by default makes TypeScript do some strange type widening where it automatically infers the generic parameter if the variable being assigned has an explicit type. This PR makes `toJSON()` and `toObject()` not generic by default.

In the interest of caution, I'm targeting merging this in v8.6 because it may cause some issues for users that rely on the broken behavior.

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -22,6 +22,7 @@ import {
   model,
   ValidateOpts
 } from 'mongoose';
+import { IsPathRequired } from '../../types/inferschematype';
 import { expectType, expectError, expectAssignable } from 'tsd';
 import { ObtainDocumentPathType, ResolvePathType } from '../../types/inferschematype';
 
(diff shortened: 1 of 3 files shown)


hasezoey (approved):
Looks good to me, would have to change something in typegoose as we seemingly relied on this behavior.

simplified example of what the use-case was:
```ts
interface User {
  name: string;
}

const pojo: User & { _id: mongoose.Types.ObjectId; id: string } = doc.toObject({ virtuals: true, getters: true });
// Error:
// Property 'id' is missing in type 'User & { _id: ObjectId; }' but required in type '{ _id: ObjectId; id: string; }'.
```
(TL;DR: we needed types for the default `id` virtual / getter, but it does not currently exist in mongoose yet)

PR #14822, opened 2024-08-21
docs(middleware): update some more "remove" related notes

hasezoey:
**Summary**

re [7.0: Removed remove()](https://mongoosejs.com/docs/migrating_to_7.html#removed-remove)

22f0722a217b69ed37caa905400f15a2a1cae889 was only partially updating everything related to "remove"

found this while browsing through #14808

EDIT:
now also includes another fix i found while making sure #14808 didnt break anything

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -31,7 +31,6 @@ In document middleware functions, `this` refers to the document. To access the m
 
 * [validate](api/document.html#document_Document-validate)
 * [save](api/model.html#model_Model-save)
-* [remove](api/model.html#model_Model-remove)
 * [updateOne](api/document.html#document_Document-updateOne)
 * [deleteOne](api/model.html#model_Model-deleteOne)
 * [init](api/document.html#document_Document-init) (note: init hooks are [synchronous](#synchronous))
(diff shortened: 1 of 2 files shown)


docs/middleware.md
```diff
@@ -382,11 +379,11 @@ Mongoose has both query and document hooks for `deleteOne()`.
 ```javascript
 schema.pre('deleteOne', function() { console.log('Removing!'); });
 
-// Does **not** print "Removing!". Document middleware for `remove` is not executed by default
+// Does **not** print "Removing!". Document middleware for `deleteOne` is not executed by default
```
hasezoey, line 382: is this statement still true for `deleteOne`? (originally it was for `remove`)

vkarpov15 (approved):
Thanks :+1:

docs/middleware.md
vkarpov15: This is true for `deleteOne()`, https://github.com/Automattic/mongoose/blob/9572cbd7436f1acb2e95f89f6f1267a1201fe5c1/test/query.middleware.test.js#L287 shows that `pre('deleteOne')` is for queries by default

PR #14870, opened 2024-09-06
set merges deeply nested objects

ianHeydoc:
**Summary**
solves the issue of deeply nested documents not merging when merge option is set

**Examples**

Example is in the test


vkarpov15 (approved):
Thanks :+1:

ianHeydoc:
Thanks! 
Will this fix also work on versions 6 and 7 of mongoose? 
Or should I write PRs for those branches as well

Issue #14842, opened 2024-08-29
TypeScript returns an error if you run a query on a subclass field of a model


nikzanda:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.6.0

### Node.js version

20.10.0

### MongoDB server version

6.0.2

### Typescript version (if applicable)

5.5.4

### Description

First of all, I find the new feature introduced by the minor release 8.6.0 very useful, which checks if fields exist in the main interface of the model when running queries! It’s a great idea!

However, I encountered a problem with models and subclasses: before 8.6.0, it was possible to run a query on a subclass field from the main model, but now TypeScript reports an error [(solved old issue)](https://github.com/Automattic/mongoose/issues/14664#issue-2348333549).

interfaces:
![image](https://github.com/user-attachments/assets/eaf45b96-f042-49e3-bf86-416464d11c67)
![image](https://github.com/user-attachments/assets/c27bda6d-459a-4b5f-abe0-4c2bce39dd1c)
![image](https://github.com/user-attachments/assets/dad7474b-e2bd-4f04-971f-533fe7ceed74)

query:
<img width="235" alt="Screenshot 2024-08-29 at 11 49 20" src="https://github.com/user-attachments/assets/2fe4870e-b299-461d-9eca-5dab3fe007ba">
<img width="1472" alt="Screenshot 2024-08-29 at 11 49 33" src="https://github.com/user-attachments/assets/6fb8488c-9ba5-4df0-9b15-c00dd1b5b480">

### Steps to Reproduce

Reproduction link: https://stackblitz.com/edit/vitejs-vite-jkwelc?file=src%2Findex.ts

### Expected Behavior

Although the main interface does not include the fields of the subclasses, you should still be able to run a query from the main model, as it worked prior to version 8.6.0.

vkarpov15:
This issue will be fixed by #14874 .

This is another good point for #14874 re: #14764: Mongoose does support automatically inferring discriminator schema if discriminator key is set in the query filter. So if we implement #14764, we would need to come up with a way to support that.

vkarpov15:
Fixed by #14874

PR #14874, opened 2024-09-09
types: allow arbitrary keys in query filters again

### LINKED ISSUE #14862: Breaking TypeScript Change in updateMany from v8.5.5 to v8.6.0
### Prerequisites

### Mongoose version

8.6.0

### Node.js version

20.15.1

### MongoDB server version

7

5.5.4

### Description

Hi Mongoose team,

In version 8.6.0, the type definition for the filter parameter in updateMany was changed from FilterQuery<TRawDocType> to RootFilterQuery<TRawDocType>. This change restricts the use of certain MongoDB operators (such as $pull) that were allowed in 8.5.5.

This change causes TypeScript errors in existing codebases that rely on the old type definition, making it a breaking change. According to Semantic Versioning, this should have warranted a major version bump rather than a minor version change.

I think this change is a breaking change for existing codebases that use MongoDB operators in the filter parameter of updateMany and suggest that it may require a major version bump according to SemVer principles.

This is my prod code snippet, which works with version 8.5.5:

```
			await GuardianModel.updateMany(
				{ $pull: { wards: { $in: [userId] } } },
				{ session },
			);
```

with version 8.6.0, I need to add something to replace the filter to keep Typescript happy:

```
			await GuardianModel.updateMany(
				{},
				// or undefined
				// or { _id: { $exists: true } }
				{ $pull: { wards: { $in: [userId] } } },
				{ session },
			);
```

This workaround does feel like a hack and is not intuitive, especially when your goal is to search all documents or apply a filter based on nested values.

Suggested Actions:

- Revert the type change or consider bumping to a new major version.

- Clearly document this change in the release notes to warn users.

### Steps to Reproduce

1. Initialize a new Node.js project with TypeScript.
2. Install Mongoose version 8.5.5 first to demonstrate the working scenario, and then 8.6.0 to show the issue.
3. Create a TypeScript file (index.ts) and define a simple Mongoose schema.

```
import mongoose, { Schema, Document } from 'mongoose';

// Define a simple Guardian schema
interface Guardian extends Document {
  wards: Schema.Types.ObjectId[];
}

const GuardianSchema = new Schema<Guardian>({
  wards: { type: [Schema.Types.ObjectId] }
});

const GuardianModel = mongoose.model<Guardian>('Guardian', GuardianSchema);

// Sample function that demonstrates the issue
async function testUpdateMany(userId: string) {
  // This should work in version 8.5.5 but throws an error in 8.6.0
  await GuardianModel.updateMany(
    { $pull: { wards: { $in: [userId] } } }, // Filter with $pull operator
    {},
    {}
  );
}
```

4. Compile and Run the Example with Version 8.5.5
5. Upgrade to Mongoose Version 8.6.0
6. Run TypeScript compilation and observe the TypeScript error: Object literal may only specify known properties, and '$pull' does not exist in type 'RootFilterQuery<Guardian>'.

### Expected Behavior

The TypeScript code should compile without errors as it did in version 8.5.5, allowing MongoDB update operators like $pull in the filter parameter of updateMany.

Actual Behavior: The TypeScript compiler throws an error in version 8.6.0 due to the stricter type definition RootFilterQuery.

### LINKED ISSUE #14863: Breaking TypeScript Change in find from v8.5.5 to v8.6.0
### Prerequisites

### Mongoose version

8.6.0

### Node.js version

20.15.1

### MongoDB server version

7

5.5.4

### Description

Hi Mongoose team,

Mongoose version 8.6.0 introduced a breaking change in the type definitions for the find method. The filter parameter type was changed from FilterQuery<TRawDocType> to RootFilterQuery<TRawDocType>, which no longer allows fields or operators that are not explicitly defined in the schema.

This change significantly affects the use of generic utility functions in TypeScript. For example, functions designed to operate across multiple Mongoose models with different schemas (using a generic Model<T>) now fail to compile when using dynamically added fields or common filters that are not explicitly declared in every schema.

Previously, the more permissive FilterQuery<TRawDocType> allowed such flexibility, enabling code reuse and reducing code duplication across models. The stricter typing of RootFilterQuery<TRawDocType> now forces developers to either:

- Refrain from using dynamic fields or filters that are not explicitly defined in the schema.

- Resort to less type-safe workarounds, such as using type assertions or @ts-ignore, to bypass TypeScript errors.

### Steps to Reproduce

1. Initialize a new Node.js project with TypeScript and install Mongoose version 8.5.5
2. Define a Mongoose schema and a function that uses find with a filter containing fields that might not be strictly defined in the schema:

```
import mongoose, { Schema, Document, Model, Types } from 'mongoose';

// Define a simple schema with a limited number of fields
interface ExampleDoc extends Document {
  isActive: boolean;
  isCustom: boolean;
}

const ExampleSchema = new Schema<ExampleDoc>({
  isActive: Boolean,
  isCustom: Boolean
});

const ExampleModel: Model<ExampleDoc> = mongoose.model<ExampleDoc>('Example', ExampleSchema);

async function initializeRelatedParametersHelper<T extends Document>(
  model: Model<T>,
  createdForId: Types.ObjectId,
  createdForType: string,
  trainingZoneId: Types.ObjectId
): Promise<void> {
  // This should work in version 8.5.5 but throws an error in 8.6.0
  const defaults = await model
    .find({
      createdForType: 'system', // This field is not defined in ExampleSchema
      isActive: true,
      isCustom: false,
    })
    .lean()
    .exec();

  for (const defaultParam of defaults) {
    const newParam = new model({
      ...defaultParam,
      createdForId,
      createdForType,
      trainingZoneId,
      isCustom: false,
    });
    await newParam.save();
  }
}
```

3. Compile and Run with Mongoose Version 8.5.5. The code should compile without errors.
4. Upgrade to Mongoose Version 8.6.0
5. Recompile the TypeScript code and observe the TypeScript error: Object literal may only specify known properties, and 'createdForType' does not exist in type 'RootFilterQuery<T>'.

### Expected Behavior

TypeScript should compile without errors, allowing the flexibility to use generic utility functions across different models with dynamic fields or query parameters not strictly defined in the schema.

vkarpov15:
Revert #14764
Fix #14863
Fix #14862

**Summary**

#14764 introduced issues for quite a few devs, see #14863. We should revert for now, and consider adding #14764 in 9.0.

cc @alex-statsig 

**Examples**

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -216,7 +216,7 @@ function find() {
   Project.find({ name: 'Hello' });
 
   // just callback; this is no longer supported on .find()
-  expectError(Project.find((error: CallbackError, result: IProject[]) => console.log(error, result)));
+  Project.find((error: CallbackError, result: IProject[]) => console.log(error, result));
 
   // filter + projection
   Project.find({}, undefined);
@@ -976,30 +976,4 @@ function testWithLevel1NestedPaths() {
     foo?: { one?: string | null | undefined } | null | undefined,
     'foo.one': string | null | undefined
   }>({} as Test2);
-}
-
-function gh14764TestFilterQueryRestrictions() {
-  const TestModel = model<{ validKey: number }>('Test', new Schema({}));
-  // A key not in the schema should be invalid
-  expectError(TestModel.find({ invalidKey: 0 }));
-  // A key not in the schema should be invalid for simple root operators
-  expectError(TestModel.find({ $and: [{ invalidKey: 0 }] }));
-
-  // Any "nested" keys should be valid
-  TestModel.find({ 'validKey.subkey': 0 });
-
-  // And deeply "nested" keys should be valid
-  TestModel.find({ 'validKey.deep.nested.key': 0 });
-  TestModel.find({ validKey: { deep: { nested: { key: 0 } } } });
-
-  // Any Query should be accepted as the root argument (due to merge support)
-  TestModel.find(TestModel.find());
-  // A Query should not be a valid type for a FilterQuery within an op like $and
-  expectError(TestModel.find({ $and: [TestModel.find()] }));
-
-  const id = new Types.ObjectId();
-  // Any ObjectId should be accepted as the root argument
-  TestModel.find(id);
-  // A ObjectId should not be a valid type for a FilterQuery within an op like $and
-  expectError(TestModel.find({ $and: [id] }));
 }
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -12,7 +12,7 @@ declare module 'mongoose' {
    */
   type RootFilterQuery<T> = FilterQuery<T> | Query<any, any> | Types.ObjectId;
 
-  type FilterQuery<T> ={
+  type FilterQuery<T> = {
     [P in keyof T]?: Condition<T[P]>;
   } & RootQuerySelector<T> & { _id?: Condition<string>; };
 
@@ -117,10 +117,8 @@ declare module 'mongoose' {
     /** @see https://www.mongodb.com/docs/manual/reference/operator/query/comment/#op._S_comment */
     $comment?: string;
     $expr?: Record<string, any>;
-    // we could not find a proper TypeScript generic to support nested queries e.g. 'user.friends.name'
-    // this will mark all unrecognized properties as any (including nested queries) only if
-    // they include a "." (to avoid generically allowing any unexpected keys)
-    [nestedSelector: `${string}.${string}`]: any;
+    // this will mark all unrecognized properties as any (including nested queries)
+    [key: string]: any;
   };
 
   interface QueryTimestampsConfig {


alex-statsig:
Sorry, sounds good to me! Waiting for 9.0 to make it a breaking change seems reasonable. Guess we could even expose a "typesafe" or "unsafe" version of each function to allow people to opt out, although unsafe is the same as just doing `model.query(queryObj as any)`, so maybe thats a simpler workaround in 9.0

vkarpov15:
@alex-statsig there's also another issue that popped up in #14842: Mongoose discriminators support specifying the discriminator key in the query filter, and Mongoose will infer the schema based on the discriminator key. For example:

```javascript
const Person = new Schema({ name: String });
const Employee = Person.discriminator('Employee', new Schema({ title: String }));

// Because discriminator key `__t` is set, Mongoose is able to infer the type of the `title` property, even
// though the model is `Person` not `Employee`
await Person.findOne({ __t: 'Employee', title: 'CEO' });
```

Is there a way to support this with TypeScript? I imagine we would need some way of defining what discriminators are tied to a particular model in TS.

test/types/models.test.ts
```diff
@@ -216,7 +216,7 @@ function find() {
   Project.find({ name: 'Hello' });
 
   // just callback; this is no longer supported on .find()
-  expectError(Project.find((error: CallbackError, result: IProject[]) => console.log(error, result)));
+  Project.find((error: CallbackError, result: IProject[]) => console.log(error, result));
```
hasezoey, line 219: after the revert, does the "expectError" now fail?

hasezoey (approved):
Looks good, lets wait for 9.0 to re-consider introducing this strictness

test/types/models.test.ts
vkarpov15: Yes. Functions match "object with any keys" in TypeScript, and telling TypeScript "this parameter can be any non-function object" is tricky, I haven't been able to figure out how to do that.

image

I would love to keep the "this parameter can't be a function" behavior, but not sure how that would work

PR #14875, opened 2024-09-10
fix(transaction): avoid unnecessarily updating initial state in between transactions

### LINKED ISSUE #14848: Array `addToSet()` and `push()` methods do not work with transaction retries
### Prerequisites

### Mongoose version

8.6.0

### Node.js version

20.16.0

### MongoDB server version

5

N/A

### Description

Hello there,

Scenario:

1. A document has a subdocument array.
2. The array is modified via `push()`.
3. The document is `save()`d inside a transaction.
4. The transaction is retried (the whole transaction as opposed to just the commit).

Result: duplicate entries are saved. Actually, this is exactly the same as in #14340 - the total number of documents added to the array grows exponentially with every retry - it is 2^n, where n is the number of retries.

I chose not to reopen the old issue because I have a hunch that this time the cause might be different.

There is also a closely-related problem when using `addToSet()`. Other array methods (`pop()`, `pull()`, `remove()`, `set()`, `shift()` and even `splice()`) all seem to be working correctly.

### My analysis

Let's take this for example (you'll find the full script below):

```typescript
doc.items.push({ name: "test3" });

  console.log(`Attempt: ${attempt}`);
  await doc.save({ session });
});
```

The first attempt to `save()` generates the following `updateOne` query:

```json
{"$push":{"items":{"$each":[{"name":"test3"}]}},"$inc":{"__v":1}}
```

If the transaction is retried, the second attempt to `save()` generates this `updateOne` query:

```json
{"$set":{"__v":1},"$push":{"items":{"$each":[{"name":"test3"},{"name":"test3"}]}}}
```

Attempt 3:

```json
{"$set":{"__v":1},"$push":{"items":{"$each":[{"name":"test3"},{"name":"test3"},{"name":"test3"},{"name":"test3"}]}}}
```

Etc. - as you can see, with each attempt the number of elements in `$push.items.$each` doubles.

A side-effect of this problem is that the update query can grow significantly in size. If it grows too large, it might fail with one of these errors:

```
RangeError: The value of "offset" is out of range. It must be >= 0 && <= 17825792. Received 17825793

MongoServerError: BSONObj size: 16928643 (0x1024F83) is invalid. Size must be between 0 and 16793600(16MB) First element: update: "mymodels"
```

As a matter of fact, this was the first problem that I ran into since I noticed that saving after `addToSet()` was _sometimes_ failing. Now I know that _sometimes_ was when a transaction was retried a lot.

### Steps to Reproduce

```typescript
import { MongoServerError, MongoErrorLabel } from "mongodb";

console.log("connecting...");

// Connecting to a *ReplicaSet* specifically to make sure that transactions work correctly.
// Nevertheless, whether or not an actual ReplicaSet is used has no effect on the issue.
await mongoose.connect(
  "mongodb://mongo-node1:27017,mongo-node2:27018,mongo-node3:27019/xxx?replicaSet=eof"
);

console.log("connected!");

// Define some schemas...
const subItemSchema = new mongoose.Schema(
  {
  },
  { _id: false }
);

const itemSchema = new mongoose.Schema(
  {
    subItems: { type: [subItemSchema], required: true },
  },
  { _id: false }
);

  items: { type: [itemSchema], required: true },
});

// ...and a model
const Model = mongoose.model("MyModel", schema);

// Clear the collection...
await Model.deleteMany({});

// ...and create one document
await Model.create({
  items: [
    { name: "test1", subItems: [{ name: "x1" }] },
    { name: "test2", subItems: [{ name: "x2" }] },
  ],
});

const doc = await Model.findOne();

doc.items.addToSet({ name: "test3", subItems: [{ name: "x3" }] });

let attempt = 0;

// Controls the number of retries.
// 17 works, 18 fails
let retries = 18;

  console.log(`Attempt: ${attempt}`);

  await doc.save({ session });

  if (attempt < retries) {
    attempt += 1;
    throw new MongoServerError({
      message: "Test transient transaction failures & retries",
      errorLabels: [MongoErrorLabel.TransientTransactionError],
    });
  }
});

await mongoose.disconnect();
```

vkarpov15:
**Summary**

Updating initial state on every save seems like incorrect behavior, since the idea of "initial state" is that is the state we should restore the document to if transaction fails.

Fix #14848

**Examples**

### DIFF
--- a/lib/plugins/trackTransaction.js
+++ b/lib/plugins/trackTransaction.js
@@ -27,13 +27,6 @@ module.exports = function trackTransaction(schema) {
       initialState.atomics = _getAtomics(this);
 
       session[sessionNewDocuments].set(this, initialState);
-    } else {
-      const state = session[sessionNewDocuments].get(this);
-
-      for (const path of Object.keys(this.$__.activePaths.getStatePaths('modify'))) {
-        state.modifiedPaths.add(path);
-      }
-      state.atomics = _getAtomics(this, state.atomics);
     }
   });
 };
--- a/test/docs/transactions.test.js
+++ b/test/docs/transactions.test.js
@@ -567,28 +567,23 @@ describe('transactions', function() {
 
   it('transaction() avoids duplicating atomic operations (gh-14848)', async function() {
     db.deleteModel(/Test/);
-    // Define some schemas...
     const subItemSchema = new mongoose.Schema(
-      { 
-        name: { type: String, required: true },
+      {
+        name: { type: String, required: true }
       },
       { _id: false }
     );
-
     const itemSchema = new mongoose.Schema(
       {
         name: { type: String, required: true },
-        subItems: { type: [subItemSchema], required: true },
+        subItems: { type: [subItemSchema], required: true }
       },
       { _id: false }
     );
-
     const schema = new mongoose.Schema({
-      items: { type: [itemSchema], required: true },
+      items: { type: [itemSchema], required: true }
     });
-
-    // ...and a model
-    const Test = db.model("Test", schema);
+    const Test = db.model('Test', schema);
 
 
     await Test.createCollection();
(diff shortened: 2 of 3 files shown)


hasezoey (commented):
LGTM

PR #14881, opened 2024-09-10
types: make SchemaType static `setters` property accessible in TypeScript

### LINKED ISSUE #14879: Setters property of schema types not available in TS declarations
### Prerequisites

### Mongoose version

8.6.1

### Node.js version

20.13.1

### MongoDB server version

7.0.8

5.5.2

### Description

In #13538 , a new static `setters` property was added to several schema type classes as well as an instance `setters` property to the base `SchemaType` class. These properties are not available in the TS declaration files, so accessing the setters property is issuing a type error.

image

### Steps to Reproduce

```ts
mongoose.Schema.Types.String.setters.push((val?: unknown) => val);
```
Type error will occur on the access of the missing `setters` property.

### Expected Behavior

The `setters` property should allow pushing a setter function without type error.

vkarpov15:
Fix #14879

**Summary**

There are places in the docs where we recommend pushing default setters to schematypes like `mongoose.Schema.String.setters.push(fn)`, but TypeScript types currently don't support that.

**Examples**

### DIFF
--- a/types/schematypes.d.ts
+++ b/types/schematypes.d.ts
@@ -216,6 +216,9 @@ declare module 'mongoose' {
     /** Attaches a getter for all instances of this schema type. */
     static get(getter: (value: any) => any): void;
 
+    /** Array containing default setters for all instances of this SchemaType */
+    static setters: Function[];
+
     /** The class that Mongoose uses internally to instantiate this SchemaType's `options` property. */
     OptionsConstructor: SchemaTypeOptions<T>;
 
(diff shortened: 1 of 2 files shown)


types/schematypes.d.ts
```diff
@@ -216,6 +216,9 @@ declare module 'mongoose' {
     /** Attaches a getter for all instances of this schema type. */
     static get(getter: (value: any) => any): void;
 
+    /** Array containing default setters for all instances of this SchemaType */
+    static setters: Function[];
```
hasezoey, line 220:
```suggestion
    static setters: ((val?: unknown) => unknown)[];
```

i think it is better with known arguments, unless there are more arguments?

hasezoey (approved):

PR #14884, opened 2024-09-11
fix(model): throw MongooseBulkSaveIncompleteError if bulkSave() didn't completely succeed

### LINKED ISSUE #14763: bulkSave doesn't throw an error even if one of the submitted documents was not updated
### Prerequisites

### Mongoose version

8.x.x

### Node.js version

20.11

### MongoDB version

7.0

### Operating system

None

_No response_

### Issue

Hi all,

While troubleshooting we ran into some unexpected behavior of the `bulkSave` function. 
If one of the submitted documents is outdated (has a lower version than the copy on the DB), this operation will not throw any error, even though the document won't be updated. 
We expected it to behave similarly to the `save` function, which throws a `VersionError` in these cases. 

Repro:
```js
    public async foo() {
        const fooSchema = new mongoose.Schema({
            bar: { type: Number },
        });
        const model = mongoose.model("foo", fooSchema);
        await model.deleteMany();

        const foo = await model.create({
            bar: 0
        });

        // update 1
        foo.bar = 1;
        await foo.save();

        // parallel update
        const fooCopy = await model.findById(foo._id);
        fooCopy.bar = 99;
        await fooCopy.save();

        // update 2 - this should throw a version error
        foo.bar = 2;
        // await foo.save(); // throws a VersionError
        await model.bulkSave([foo]) // no error thrown, but update not applied

        // log result
        console.log(`bar: ${(await model.findById(foo._id)).bar}`); // logs 99
    }
```

Is this the expected behavior of `bulkSave`? Is there any possibility of making it fail if one of the documents has been already replaced by a newer version instead of returning silently?

vkarpov15:
Fix #14763

**Summary**

Make it so that `bulkSave()` throws an error if not all documents were successfully updated due to version errors or optimistic concurrency. I thought more about #14763 and this seems like the way to go: if not all documents were updated, we don't know which documents we can run post save hooks on, or which documents to update `$isNew` on.

**Examples**

### DIFF
--- a/lib/error/bulkSaveIncompleteError.js
+++ b/lib/error/bulkSaveIncompleteError.js
@@ -1,0 +1,44 @@
+/*!
+ * Module dependencies.
+ */
+
+'use strict';
+
+const MongooseError = require('./mongooseError');
+
+
+/**
+ * If the underwriting `bulkWrite()` for `bulkSave()` succeeded, but wasn't able to update or
+ * insert all documents, we throw this error.
+ *
+ * @api private
+ */
+
+class MongooseBulkSaveIncompleteError extends MongooseError {
+  constructor(modelName, documents, bulkWriteResult) {
+    const matchedCount = bulkWriteResult?.matchedCount ?? 0;
+    const insertedCount = bulkWriteResult?.insertedCount ?? 0;
+    let preview = documents.map(doc => doc._id).join(', ');
+    if (preview.length > 100) {
+      preview = preview.slice(0, 100) + '...';
+    }
+
+    const numDocumentsNotUpdated = documents.length - matchedCount - insertedCount;
+    super(`${modelName}.bulkSave() was not able to update ${numDocumentsNotUpdated} of the given documents due to incorrect version or optimistic concurrency, document ids: ${preview}`);
+
+    this.modelName = modelName;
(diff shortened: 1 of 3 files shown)


lib/error/bulkSaveIncompleteError.js
```diff
@@ -0,0 +1,44 @@
...
+
+const MongooseError = require('./mongooseError');
+
+
+/**
+ * If the underwriting `bulkWrite()` for `bulkSave()` succeeded, but wasn't able to update or
+ * insert all documents, we throw this error.
+ *
+ * @api private
+ */
```
hasezoey, line 15: i think this block should be on `constructor` instead of on `class` like the other errors

hasezoey (approved):
LGTM, minor jsdoc suggestion and the error should likely be added to the types too

lib/error/bulkSaveIncompleteError.js
vkarpov15: We're actually mixed on whether the comment is on the class or the constructor. For example, `MongooseBulkWriteError` and `CastError` have the comment on the class, but `DivergentArrayError` and `InvalidSchemaOptionError` have the comment on the constructor. We'll make this more consistent in a future PR.

Issue #12959, opened 2023-01-30
Unexpected TS types for _id and _v in inferred types, particularly for subdocument arrays


shawnmcknight:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.9.0

### Node.js version

16.18.1

### MongoDB server version

4.4.18

### Typescript version (if applicable)

4.9.4

### Description

I'm encountering a few unexpected behaviors with the types being inferred from both schema helpers (`InferSchemaType` and `HydratedDocumentFromSchema`), the resulting types when executing a model method, and the resulting types when converting that result to a POJO., particularly as it pertains to subdocument arrays.  The included sample code hopefully tells the story, but this is a summary of what I have found that _feels_ incorrect.

### Subdocument Array inference is returning types for `_id` and `__v` when using `InferSchemaType`
This is less of an error than a surprise based on what I read in #11967.  As per that issue, `InferSchemaType` does not return an `_id` type.  Although it's not indicated in that issue, it also doesn't have a type for `__v` but that seems to be consistent with not having an `_id` type.  However, I was very surprised to see that subdocument arrays DO have a type for the `_id` and the `__v` value.  It seems like neither the parent document nor the subdocuments should have these types.

### Hydrated documents have wide types for `_id` and `__v` for subdocument arrays
I ran my schema through `HydratedDocumentFromSchema`.  This correctly provided a type of `mongoose.types.ObjectId` for the parent document's `_id` property.  However, the subdocument arrays have a type of `mongoose.Types.ObjectId | undefined`.  I would have not expected this to be an optional property.

Additionally, the types for `__v` on both the parent document as well as the subdocuments is `any` with the optional flag.  I traced through some history and it seems like `any` was put there instead of number to try to avoid some other issues.  I was surprised to see the optional flag on this type, although with `any` it just widens it to `any`, so maybe it doesn't matter.  If this type ever was able to be more accurate and be `number` then I wouldn't think this should be `number | undefined`.

### Document instance exhibits the same types as the `HydratedDocumentFromSchema`
This seems to track as I would expect them to be consistent with each other.  However, all of what I described that feels off in that section also applies to the document instances.

### Document instance converted to POJO additionally lacks a type for `__v`
[...]

PR #14892, opened 2024-09-16
types: make __v a number, only set __v on top-level documents

vkarpov15:
Re: #12959

**Summary**

Version key `__v` is only set on top-level documents at runtime, so TypeScript types should support that. We'll put this change in 8.7 in the interest of caution, because this may cause issues for people who use `extends Document`

**Examples**


test/types/document.test.ts
```diff
@@ -359,3 +359,19 @@ function gh13738() {
   expectType<Date>(person.get('dob'));
   expectType<{ theme: string; alerts: { sms: boolean } }>(person.get('settings'));
 }
+
+async function gh12959() {
+  const subdocSchema = new Schema({ foo: { type: 'string', required: true } });
+
+  const schema = new Schema({
+    subdocArray: { type: [subdocSchema], required: true },
```
hasezoey, line 367:
```suggestion
    subdocArray: { type: [subdocSchema], required: true }
```

fix lint

test/types/document.test.ts
```diff
@@ -359,3 +359,19 @@ function gh13738() {
...
+  });
+
+  const Model = model('test', schema);
+
+  const doc = await Model.findById('id').orFail();
+  expectType<Types.ObjectId>(doc._id);
+  expectType<number | undefined>(doc.__v);
+
+  expectError(doc.subdocArray[0].__v);
+}
```
hasezoey, line 377:
```suggestion
}

```

fix lint

hasezoey (approved):

PR #14894, opened 2024-09-18
fix(projection): avoid setting projection to unknown exclusive/inclusive if elemMatch on a Date, ObjectId, etc.

### LINKED ISSUE #14893: Mongoose >= 7.0.0 inserts default values after querying with a single field projection and $elemMatch
### Prerequisites

### Mongoose version

7.8.0

### Node.js version

20.16.0

### MongoDB version

5.0.28

### MongoDB driver version

5.9.2

### Operating system

None

_No response_

### Issue

Before mongoose 7, If we made a query with an $elemMatch projection (on a single subdocument array - no other fields are mentioned in the projection),  the document will return with the requested subdocument array item only (as requested) and without any other document fields.

Example Schema:

```
const MySchema = new Schema({
    details: [{
        name: String,
        id: String,
        age: Number
    }],
    addresses: [{
        streetName: String,
        streetNumber: String
    }],
    permissions: {
        add: {type: Boolean, default: false},
        other: {type: Number, default: 3}
    }
});
```

Example:
```
const doc = await Model.findById(some-id, 'details').select({details: {$elemMatch: {id: some-id}}}).exec();
```
The returned doc will look as follows:
```
{
    details: [{
        name: 'Gary',
        age: 23,
        id: '12345'
    }]
}
```
If I try to call doc.permissions it will be undefined

After v7, when using an $elemMatch projection it seems as if the document returns with all the default fields so that when eventually we call a save() action on the document, it saves the document and overrides any field that has a default value

Example:
```
const doc = await Model.findById(some-id, 'details').select({details: {$elemMatch: {id: some-id}}}).exec();
```
The returned doc will look as follows:
```
{
    details: [{
        name: 'Gary',
        age: 23,
        id: '12345'
    }],
    addresses: [],
    permissions: {add: false, other: 3}
}
```
The addresses subDocument array returns with the default schema values even though it may have multiple items in the array. The same goes for the permissions object which may have an "add" value of true.

If I modify the doc (the specific subdocument array item requested in the projection) and then call save(), the document saves all fields with their default values. It overwrites the current values with the defaults event though the fields were not requested in the projection.

Is this the proper behavior? is there any way to overcome this issue and not get the default values if they are not requested in the projection?

Thanks

vkarpov15:
Fix #14893

**Summary**

https://github.com/Automattic/mongoose/commit/2b197b21dff7fb53333a623426d564442f78552e#diff-af7545780ae0b49c587fcbe08b131ce99659ccac5786983702610b6c7a3b8a24 made it so that `isExclusive()` would return `null` if the projection had a `$elemMatch` where the first property was an ObjectId, Date, Buffer, or other value object.

**Examples**

### DIFF
--- a/lib/helpers/projection/isExclusive.js
+++ b/lib/helpers/projection/isExclusive.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const isDefiningProjection = require('./isDefiningProjection');
+const isPOJO = require('../isPOJO');
 
 /*!
  * ignore
@@ -22,10 +23,12 @@ module.exports = function isExclusive(projection) {
       // Explicitly avoid `$meta` and `$slice`
       const key = keys[ki];
       if (key !== '_id' && isDefiningProjection(projection[key])) {
-        exclude = (projection[key] != null && typeof projection[key] === 'object') ?
-          isExclusive(projection[key]) :
+        exclude = isPOJO(projection[key]) ?
+          (isExclusive(projection[key]) ?? exclude) :
           !projection[key];
-        break;
+        if (exclude != null) {
+          break;
+        }
       }
     }
   }
--- a/lib/helpers/projection/isInclusive.js
+++ b/lib/helpers/projection/isInclusive.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const isDefiningProjection = require('./isDefiningProjection');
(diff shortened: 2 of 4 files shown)


hasezoey (commented):
LGTM

Issue #14903, opened 2024-09-22
hook is not of type Aggregate


dragontaek-lee:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.6.1

### Node.js version

20.10.0

### MongoDB server version

7.0.9

### Description

this context in hooks are not of type Aggregate. 

This issue arises when custom static methods overwrite built-in aggregate functions. (as seen in the implementation in `mongoose-delete`)

This issue is similar to the past one (https://github.com/Automattic/mongoose/issues/7790), where the context in hooks was not of type Query. I noticed that it has been fixed in this PR (https://github.com/Automattic/mongoose/commit/f848b54d33cc5f61be3b1f0539370f9c43d9d4e2), which filters custom static methods that overwrite existing query middleware in `applyStaticHooks.js`.

### Steps to Reproduce

I created a minimum reproduction repository. (https://github.com/dragontaek-lee/mongoose-staticHooks-overwrite-support-more-cases)

I created some test cases to reproduce this issue in the reproduction repository.
The code can be found in `index.js`, `test/aggregate.middleware.test.js`, or `test/query.middleware.test.js`. 
Tests and scripts can also be run using the npm commands mentioned in the README.

**The summarized reproduction code is as follows:**

```
schema1.statics.aggregate = function(pipeline) {
    let match = { $match: { deleted: { '$ne': false } } };

    if (pipeline.length && pipeline[0].$match) {
        pipeline[0].$match.deleted = { '$ne': false };
    } else {
        pipeline.unshift(match);
    }

    const query = Model.aggregate.apply(this, [pipeline]);
    return query;
};
```

Assume that the static method overwrites the built-in Mongoose aggregate function.

```
schema1.pre('aggregate', function(next) {
    console.log(this);
    next();
});

schema1.post('aggregate', function() {
    console.log(this);
});
```

In this situation, also assume that there are pre and post hooks for aggregate.

**Test 1**

- both the custom static method and the built-in Mongoose aggregate hooks are executed and both the pre and post hooks are called twice

**Test 2**

- this context differs between the custom static method and the built-in aggregate function.
- in the first pre-hook and the second post-hook call, `this` context **is not an instance of Aggregate**, but in the second pre-hook and the first post-hook call, **it becomes an instance of Aggregate.**

### Expected Behavior

[...]

PR #14904, opened 2024-09-22
fix(model): skip applying static hooks by default if static name conflicts with aggregate middleware

dragontaek-lee:
Fix https://github.com/Automattic/mongoose/issues/14903

**Summary**

Make it so that if the custom static overwrites an existing aggregate middleware, don't apply middleware to it by default. This should be achieved by filtering custom static-named aggregates in `applyStaticHooks.js`. 

Furthermore, aside from this PR, I believe we should extend this filtering not only to query and aggregation middleware but also to other [Mongoose middlewares](https://mongoosejs.com/docs/middleware.html), including Document and Model middleware. Does that make sense? If so, I would like to create another PR for this work.

**Examples**

```
schema.statics.aggregate = function(pipeline) {
    let match = { $match: { deleted: { '$ne': false } } };

    if (pipeline.length && pipeline[0].$match) {
        pipeline[0].$match.deleted = { '$ne': false };
    } else {
        pipeline.unshift(match);
    }

    const query = Model.aggregate.apply(this, [pipeline]);
    return query;
};
```

Assume that the static method overwrites the built-in Mongoose aggregate function.

```
schema.pre('aggregate', function(next) {
    console.log(this);
    next();
});

schema.post('aggregate', function() {
    console.log(this);
});
```

In this situation, also assume that there are pre and post hooks for aggregate.

With this PR, when a custom static method overwrites an existing aggregate middleware as described above, the middleware is not applied by default. As a result, the `this` context type in pre/post hooks is correctly set to Aggregate, as intended.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -34,3 +34,13 @@ const queryMiddlewareFunctions = queryOperations.concat([
 ]);
 
 exports.queryMiddlewareFunctions = queryMiddlewareFunctions;
+
+/*!
+ * ignore
+ */
+
+const aggregateMiddlewareFunctions = [
+  'aggregate'
+];
+
+exports.aggregateMiddlewareFunctions = aggregateMiddlewareFunctions;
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks :+1:

vkarpov15:
And yes you can create a separate PR to apply this filtering to other middleware types.

PR #14905, opened 2024-09-22
feat(model): add `Model.applyVirtuals()` to apply virtuals to a POJO

### LINKED ISSUE #14818: Model.applyVirtuals(): apply schema virtuals to a POJO, similar to Model.applyDefaults()
### Prerequisites

### 🚀 Feature Proposal

Supporting virtual fields in aggregate will be good feature.

I know aggregate is server-side feature and virtual fields is client-side feature and aggregation results can change shape of document.
However, some aggregate results just extends document data and it is possible to add virtual fields to aggregate results, just I mentioned in motivation.
I think we can add `hasVirtualFields` to aggregate options which default is false and add virtual fields to aggregate results.

### Motivation

I am using aggregate function for pagination. 
Since aggregate doesn't support virtual fields, I need to hydrate each aggregation result and add virtual fields to each aggregation result manually.
It will be good to do in mongoose itself.

### Example

```
UserModel.aggregate([
  // add pipelines
], {
  hasVirtualFields: true
});
```

So, we can add virtual fields like `fullName` to aggregate results.

vkarpov15:
Fix #14818

**Summary**

Logic largely copied from https://github.com/mongoosejs/mongoose-lean-virtuals/blob/master/index.js to apply a Model's virtuals to a POJO.

**Examples**

### DIFF
--- a/lib/helpers/document/applyVirtuals.js
+++ b/lib/helpers/document/applyVirtuals.js
@@ -1,0 +1,141 @@
+'use strict';
+
+const mpath = require('mpath');
+
+module.exports = applyVirtuals;
+
+function applyVirtuals(schema, doc, virtuals, parent) {
+  if (doc == null) {
+    return doc;
+  }
+
+  let virtualsForChildren = virtuals;
+  let toApply = null;
+
+  if (Array.isArray(virtuals)) {
+    virtualsForChildren = [];
+    toApply = [];
+    const len = virtuals.length;
+    for (let i = 0; i < len; ++i) {
+      const virtual = virtuals[i];
+      if (virtual.length === 1) {
+        toApply.push(virtual[0]);
+      } else {
+        virtualsForChildren.push(virtual);
+      }
+    }
+  }
+
+  applyVirtualsToChildren(this, schema, doc, virtualsForChildren, parent);
+  return applyVirtualsToDocs(schema, doc, toApply);
+}
+
+function applyVirtualsToDocs(schema, res, toApply) {
+  if (Array.isArray(res)) {
+    const len = res.length;
+    for (let i = 0; i < len; ++i) {
+      applyVirtualsToDoc(schema, res[i], toApply);
+    }
+    return res;
+  } else {
+    return applyVirtualsToDoc(schema, res, toApply);
+  }
+}
+
+function applyVirtualsToChildren(doc, schema, res, virtuals, parent) {
+  const len = schema.childSchemas.length;
+  let attachedVirtuals = false;
+  for (let i = 0; i < len; ++i) {
+    const _path = schema.childSchemas[i].model.path;
+    const _schema = schema.childSchemas[i].schema;
+    if (!_path) {
+      continue;
+    }
+    const _doc = mpath.get(_path, res);
+    if (_doc == null || (Array.isArray(_doc) && _doc.flat(Infinity).length === 0)) {
+      continue;
+    }
+
+    let virtualsForChild = null;
+    if (Array.isArray(virtuals)) {
+      virtualsForChild = [];
+      const len = virtuals.length;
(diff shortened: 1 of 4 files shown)


lib/model.js
```diff
@@ -3503,6 +3507,40 @@ Model.applyDefaults = function applyDefaults(doc) {
...
+ *     const userSchema = new Schema({ name: String });
+ *     userSchema.virtual('upper').get(function() { return this.name.toUpperCase(); });
+ *     const User = mongoose.model('User', userSchema);
+ *
+ *     const obj = { name: 'John' };
+ *     User.applyVirtuals(obj);
+ *     obj.name; // 'John'
+ *     obj.upper; // 'JOHN', Mongoose applied the return value of the virtual to the given object
+ *
+ * @param {Object} obj object or document to apply virtuals on
```
hasezoey, line 3524:
```suggestion
 * @param {Object} doc object or document to apply virtuals on
```

consistency with the parameter name

lib/model.js
```diff
@@ -3503,6 +3507,40 @@ Model.applyDefaults = function applyDefaults(doc) {
...
+ * @param {Array<string>} [virtualsToApply] optional whitelist of virtuals to apply
+ * @returns {Object} obj
```
hasezoey, line 3526:
```suggestion
 * @returns {Object} doc
```

test/model.test.js
```diff
@@ -7781,6 +7781,208 @@ describe('Model', function() {
...
+          first: 'Bill',
+          last: 'Gates'
+        }
+      }, ['name.lastLower']);
+      assert.strictEqual(res.name.first, 'Bill');
+      assert.strictEqual(res.name.last, 'Gates');
+      assert.strictEqual(res.fullName, undefined);
+      assert.strictEqual(res.name.firstUpper, undefined);
+      assert.strictEqual(res.name.lastLower, 'gates');
+    });
```
hasezoey, line 7949: isnt this literally a copy of the test before it?

lib/helpers/document/applyVirtuals.js
```diff
@@ -0,0 +1,141 @@
...
+      continue;
+    }
+    const virtualType = schema.virtuals[virtual];
+    const sp = Array.isArray(virtual)
+      ? virtual :
+      virtual.indexOf('.') === -1
+        ? [virtual]
+        : virtual.split('.');
+    let cur = doc;
+    for (let j = 0; j < sp.length - 1; ++j) {
```
hasezoey, line 123: just to confirm, this is meant to apply to all but the last `sp`? (to my knowledge `<` is exclusive, like the other for loops are using it)

hasezoey (commented):
mostly looks good to me, some minor style suggestions, aside from that:
- typescript types for the new `applyVirtuals` are not added yet
- there is a distinct lack of documentation in the `applyVirtuals` file

test/model.test.js
vkarpov15: Good catch, sorry about that. Fixed.

lib/helpers/document/applyVirtuals.js
vkarpov15: Yes, this loop is meant to get the last nested object in the path so we can set `cur[sp.length - 1]`.

vkarpov15:
@hasezoey thanks for the feedback. I added docs and addressed issues :+1:

lib/model.js
vkarpov15: I changed consistently to `obj` because `doc` implies the parameter should be a document, which isn't the case.

hasezoey (commented):
Looks good to me, typescript types for the new `applyVirtuals` on `Model` are still missing though

hasezoey (commented):
Looks good to me

PR #14908, opened 2024-09-24
fix(model): filter applying static hooks by default if static name conflicts with mongoose middleware

dragontaek-lee:
Additional Fix for PR: https://github.com/Automattic/mongoose/pull/14904

**Summary**

Ensure that if a custom static method overwrites an existing mongoose middleware, the middleware is not applied by default, similar to how it works with query and aggregate middleware.

#### applyStaticHooks.js
```js
const middlewareFunctions = [
  ...[
    ...queryMiddlewareFunctions,
    ...aggregateMiddlewareFunctions,
    ...modelMiddlewareFunctions,
    ...documentMiddlewareFunctions
  ].reduce((s, hook) => s.add(hook), new Set())
];
```
I used a `Set()` to eliminate duplicates in the following three function names: updateOne, deleteOne, and validate.

1) Document middleware's **updateOne** and Query middleware's **updateOne**
2) Document middleware's **deleteOne** and Query middleware's **deleteOne**
3) Document middleware's **validate** and Query middleware's **validate**

**Examples**

The following example demonstrates the newly implemented filtering applied to model and document middleware.

```js
schema.statics.save = function() {
  ...
};

schema.statics.insertMany = function() {
  ...
};
```

Suppose that the static method overwrites the built-in Mongoose `save` / `insertMany` function.

```js
schema.pre('save', function(next) {
    console.log(this);
    next();
});

schema.post('save', function() {
    console.log(this);
});

schema.pre('insertMany', function(next) {
    console.log(this);
    next();
});

schema.post('insertMany', function() {
    console.log(this);
});
```

In this case, also suppose that there are pre and post hooks for save, insertMany.

With this PR, when a custom static method overwrites an existing model or document middleware as described above, the middleware is no longer applied by default. As a result, the this context type in pre/post hooks is correctly set as intended.

### DIFF
--- a/lib/helpers/model/applyStaticHooks.js
+++ b/lib/helpers/model/applyStaticHooks.js
@@ -3,14 +3,14 @@
 const promiseOrCallback = require('../promiseOrCallback');
 const { queryMiddlewareFunctions, aggregateMiddlewareFunctions, modelMiddlewareFunctions, documentMiddlewareFunctions } = require('../../constants');
 
-const middlewareFunctions = [
-  ...[
+const middlewareFunctions = Array.from(
+  new Set([
     ...queryMiddlewareFunctions,
     ...aggregateMiddlewareFunctions,
     ...modelMiddlewareFunctions,
     ...documentMiddlewareFunctions
-  ].reduce((s, hook) => s.add(hook), new Set())
-];
+  ])  
+);
 
 module.exports = function applyStaticHooks(model, hooks, statics) {
   const kareemOptions = {
--- a/lib/helpers/model/applyStaticHooks.js
+++ b/lib/helpers/model/applyStaticHooks.js
@@ -1,11 +1,15 @@
 'use strict';
 
 const promiseOrCallback = require('../promiseOrCallback');
-const { queryMiddlewareFunctions, aggregateMiddlewareFunctions } = require('../../constants');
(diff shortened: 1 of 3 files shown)


lib/helpers/model/applyStaticHooks.js
```diff
@@ -1,11 +1,15 @@
 'use strict';
 
 const promiseOrCallback = require('../promiseOrCallback');
-const { queryMiddlewareFunctions, aggregateMiddlewareFunctions } = require('../../constants');
+const { queryMiddlewareFunctions, aggregateMiddlewareFunctions, modelMiddlewareFunctions, documentMiddlewareFunctions } = require('../../constants');
 
 const middlewareFunctions = [
-  ...queryMiddlewareFunctions,
-  ...aggregateMiddlewareFunctions
+  ...[
```
vkarpov15: You can just do `Array.from(new Set([...middleware1, ...middleware2]))` to avoid the need for `reduce()`.

vkarpov15 (approved):
I made one quick refactor but otherwise LGTM :+1:

lib/helpers/model/applyStaticHooks.js
dragontaek-lee: I also think your refactoring is a better approach! Thank you 👍

PR #14910, opened 2024-09-24
fix(document): avoid massive perf degradation when saving new doc with 10 level deep subdocs

### LINKED ISSUE #14897: Infinitely hangs inserting complex document with recursive schema
### Prerequisites

### Mongoose version

8.4.1+

### Node.js version

20.x

### MongoDB server version

5.x,6.x

5.4.5

### Description

[mongoose_test.zip](https://github.com/user-attachments/files/17068839/mongoose_test.zip)

This test case demonstrates inserting a single document with a complex schema. it is actually simplified from out actual schema in use, and I've replace several subdocuments with `Mixed`.  But I expect it is very likely due to the potential recursive nature of the schema.

Works as expected in version 8.2.2, 8.3.0 and 8.4.0. 
Hangs in 8.4.1 - 8.6.3

### Steps to Reproduce

build and run attached app.

### Expected Behavior

inserts the document.

vkarpov15:
Fix #14897

**Summary**

Saving a new doc with many very deep subdocs is extremely expensive: the test case in this PR fails after 35 seconds on my local without the change I made to `lib/document.js`. Saving a new doc 10 levels deep is still more expensive than I'd like, this test now takes maybe 300ms, I'll investigate why this is so slow and add a benchmark.

**Examples**

### DIFF
--- a/benchmarks/createDeepNestedDocArray.js
+++ b/benchmarks/createDeepNestedDocArray.js
@@ -1,0 +1,37 @@
+'use strict';
+
+const mongoose = require('../');
+
+run().catch(err => {
+  console.error(err);
+  process.exit(-1);
+});
+
+async function run() {
+  await mongoose.connect('mongodb://127.0.0.1:27017/mongoose_benchmark');
+
+  const levels = 12;
+
+  let schema = new mongoose.Schema({ test: { type: String, required: true } });
+  let doc = { test: 'gh-14897' };
+  for (let i = 0; i < levels; ++i) {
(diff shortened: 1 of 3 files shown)


hasezoey (commented):
Looks good, will this be in the next minor version for caution?

vkarpov15:
No meaningful change on `insertManySimple` benchmark. But `createDeepNestedDocArray` benchmark difference is massive: 223ms vs 34802ms on master on 10 level deep subdocs. Over 100x speedup :sunglasses:

Issue #14877, opened 2024-09-10
Mongoose findOneAndUpdate hangs under heavy load


andreialecu:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.5.1

### Node.js version

20.17.0

### MongoDB server version

7.0.12

### Description

We have a batch job that does a bunch of updates that used to work a while ago, but recently started hanging on calls to `findOneAndUpdate`. The promise stops resolving after a while.

I have been able to create a repro for this, but it seems to only viably reproduce on MongoDB Atlas (or probably other remote servers). On localhost it doesn't seem to reproduce.

Also if I bypass mongoose entirely and use the mongodb driver directly, it seems to work.

### Steps to Reproduce

```js
const mongoose = require('mongoose');
const Parallel = require('async-parallel');
const { setTimeout } = require('timers/promises');

const ObjectId = mongoose.Types.ObjectId;
const Schema = mongoose.Schema;

const testSchema = new Schema({
  fieldA: {
    type: Schema.Types.ObjectId,
    required: true,
  },
  fieldB: {
    type: String,
    required: true,
  },
  fieldC: {
    type: Number,
    default: 0,
  },
  fieldD: {
    subFieldA: [String],
    subFieldB: [String],
    subFieldC: [String],
  },
  fieldK: {
    type: Date,
    default: Date.now,
  },
});

const Test = mongoose.model('Test', testSchema);

async function createObject() {
  await setTimeout(100 + Math.random() * 100);
  return {
    fieldA: new ObjectId(),
    // random hex
    fieldB: new ObjectId().toHexString(),
    fieldC: 0,
    fieldD: {
      // make this big:
      subFieldA: Array.from({ length: 2000 }, () =>
        new ObjectId().toHexString(),
      ),
      subFieldB: Array.from({ length: 2000 }, () =>
        new ObjectId().toHexString(),
      ),
      subFieldC: Array.from({ length: 2000 }, () =>
        new ObjectId().toHexString(),
      ),
    },
  };
}

async function run() {
  await mongoose.connect(
    // 'mongodb://localhost:27017/testmongoose', // could not reproduce on localhost
    'mongodb+srv://....mongodb.net/testmongoose?ssl=true&authSource=admin', // replace with a mongodb atlas connection string
  );

  let tidx = 0;
  let total = 10000;
  await Parallel.each(
    Array.from({ length: total }),
[...]

PR #14916, opened 2024-09-27
feat: add semver range for mongodb node driver and bson versions

vkarpov15:
**Summary**

Instead of hard-coding an exact version of the mongodb node driver, add a semver range. This should help with issues like #14877, where users want to use an older version of the mongodb node driver without downgrading Mongoose.

I also modified GitHub workflows to run tests against both the lowest mongodb node driver version in the semver range, and the latest version.

This is a fairly substantial change, I'll think about this some more over the weekend. Mongoose has pinned exact versions since before I became involved with the project, because historically Mongoose was a bit too tightly coupled to the MongoDB Node driver and patch releases in the MongoDB Node driver would create breaking changes in Mongoose on occasion. However, it has been years since we've seen that happen, both due to increased stability in the MongoDB Node driver and due to Mongoose no longer reaching into internal connection state regularly.

**Examples**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,6 +47,7 @@ jobs:
           - os: ubuntu-20.04 # customize on which matrix the coverage will be collected on
             mongodb: 5.0.26
             node: 16
+            node-driver: latest
             coverage: true
         exclude:
           - os: ubuntu-22.04 # exclude because there are no 4.x mongodb builds for 2204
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
Looks good to me, though i think going this far is quite a substantial change, how about (at least until 9.0) allow all patch versions like `~6.9.0` instead of directly only `6.9.0`? (though i am not familiar how heavy the performance degradation was between versions the versions in the mentioned issue)

vkarpov15:
It looks like we have an alternative fix for #14877 with https://github.com/mongodb/node-mongodb-native/pull/4249, so we should instead just bump to `mongodb@6.10.0` when this PR lands: https://github.com/mongodb/node-mongodb-native/pull/4241.

Given that change, I think it is reasonable to make Mongoose 8.8 rely on `mongodb@~6.10.0` as a first step toward opening up the semver range as per @hasezoey's comments

PR #14919, opened 2024-09-30
Update migrating_to_6.md about mongoose.modelSchemas

markstos:
It was removed by this commit that first appeared in the 6.x series:

6865fea216220a88d2e7128d3a8eb8228f29155e

My codebase only used this property for deleting models in a few places. Please peer-review the note
in case there are other uses of this property worth mentioning.


markstos:
I may be wrong about recommending replacing `delete mongoose.modelSchema.Foo` with `mongoose.deleteModel('Foo')`.

What's working better in my code base is just to delete the lines where we have `delete mongoose.modelSchemas.Foo`;

vkarpov15 (approved):
Thanks :+1:

Issue #13648, opened 2023-07-24
flattenObjectIds toObject option side effects


stevemit007:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.4.0

### Node.js version

16.17.0

### MongoDB server version

6.0.3

### Typescript version (if applicable)

5.1.6

### Description

Creating a schema with option 
```
toObject: {
  flattenObjectIds: true
}
```
has unwanted side effects.

MongoDB Compass shows the autogenerated `_id` field has type `string`, and `Model.findById()` returns `null`.

A workaround is to use `toJSON()` instead of `toObject()`.

### Steps to Reproduce

```
import { Model, Schema, model } from "mongoose";

interface IUser {
  name: string;
}
const UserSchema = new Schema<IUser>(
  {
    name: String
  },
  {
    toObject: { flattenObjectIds: true }
  }
);
const User: Model<IUser> = model("User", UserSchema);
User.create({ name: "George" }).then((d) => {
  User.findById(d._id)
    .exec()
    .then((d) => console.log(d));
});

```

### Expected Behavior

Result is not `null`.

IslandRhythms:
```typescript
import { Model, Schema, model, connect, connection } from "mongoose";

interface IUser {
  name: string;
}
const UserSchema = new Schema<IUser>(
  {
    name: String
  },
  {
    toObject: { flattenObjectIds: true }
  }
);
const User: Model<IUser> = model("User", UserSchema);
async function run() {
  await connect('mongodb://localhost:27017');
  await connection.dropDatabase();
  User.create({ name: "George" }).then((d) => {
    User.findById(d._id)
      .exec()
      .then((d) => console.log(d));
  });
}

run();

```

PR #14938, opened 2024-10-06
fix: set flattenObjectIds to false when calling toObject() for internal purposes

### LINKED ISSUE #14935: find on ObjectIds not working after insertMany
### Prerequisites

### Mongoose version

8.7.0

### Node.js version

18.x

### MongoDB server version

7.0.12

^5.1.6

### Description

After `insertMany` calling `find` with on an ObjectId parameter, it will fail to return any rows.

The issue happens when I have both a property of `type: [Schema.Types.Mixed]` and `toObject: { flattenObjectIds: true }`
I'm not sure why, but I've done written a jest test to show.

### Steps to Reproduce

```

import db from '../helper/mockDb'
import mongoose, { Schema, Types } from 'mongoose'

const userId = new Types.ObjectId()

beforeAll(async () => {
  await db.connect()
})

afterAll(async () => {
  await db.disconnect()
})

describe('Basic insertMany test', () => {
  const TestSchema = new Schema(
    {
      professionalId: {
        type: Schema.Types.ObjectId,
      },
      firstName: {
        type: String,
      },
    },
    {
      toObject: { flattenObjectIds: true },
    },
  )

  const TestModel = mongoose.model('testSchema', TestSchema)

  const SimpleExternalClientSchema = new Schema(
    {
      professionalId: {
        type: Schema.Types.ObjectId,
      },
      firstName: {
        type: String,
        default: '',
        required: true,
        trim: true,
      },
      addresses: {
        type: [Schema.Types.Mixed],
        default: [],
      },
    },
    {
      toObject: { flattenObjectIds: true },
    },
  )

  const SimpleExternalClientModel = mongoose.model('simpleExternalClient', SimpleExternalClientSchema)

  test('Should insert multiple records', async () => {
    const data = [
      { professionalId: userId, firstName: 'Testy' },
      { professionalId: userId, firstName: 'Johny' },
    ]

    await TestModel.insertMany(data)

    const records = await TestModel.find({}).exec()
    expect(records.length).toEqual(2)

    const recordFiltered = await TestModel.find({ professionalId: userId }).exec()
    expect(recordFiltered.length).toEqual(2) // <- This works
  })

  test('Should insert client multiple records', async () => {
    const externalClientData = [{
      firstName: 'Testy',
      professionalId: userId,
    }, {
      firstName: 'Testy',
      professionalId: userId,
    }]

    await SimpleExternalClientModel.insertMany(externalClientData)

    const records = await SimpleExternalClientModel.find({}).exec()
    expect(records.length).toEqual(2)

    const recordFiltered = await SimpleExternalClientModel.find({ professionalId: userId }).exec()
    expect(recordFiltered.length).toEqual(2) // <-- This is the line that is failing
  })
})
```

### Expected Behavior

Second test should pass

vkarpov15:
Fix #14935

**Summary**

`insertMany()` currently respects schema-level flattenObjectIds option settings, which means if schema sets `flattenObjectIds: true` then Mongoose will store ObjectIds as strings. 

This does not affect `save()` because `save()` was fixed in #13648, but that issue still affects `insertMany()`.

In general, `internalToObjectOptions` should always set `flattenObjectIds` to `true` because we rely on those options when trying to get the MongoDB server representation of an object for queries, `insertMany()`, etc.

We should also backport this fix to 6.x and 7.x.

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -13926,6 +13926,32 @@ describe('document', function() {
     cur.subdocs[0] = { test: 'updated' };
     await savedDoc.save();
   });
+
+  it('avoids flattening objectids on insertMany (gh-14935)', async function() {
+    const TestSchema = new Schema(
+      {
+        professionalId: {
+          type: Schema.Types.ObjectId
+        },
+        firstName: {
+          type: String
+        },
+        nested: {
+          test: String
+        }
+      },
+      {
+        toObject: { flattenObjectIds: true }
+      }
+    );
+    const Test = db.model('Test', TestSchema);
+
+    const professionalId = new mongoose.Types.ObjectId();
+    await Test.insertMany([{ professionalId, name: 'test' }]);
+
+    const doc = await Test.findOne({ professionalId }).lean().orFail();
+    assert.ok(doc.professionalId instanceof mongoose.Types.ObjectId);
+  });
 });
 
 describe('Check if instance function that is supplied in schema option is available', function() {
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -13952,7 +13952,7 @@ describe('document', function() {
     const doc = await Test.findOne({ professionalId }).lean().orFail();
     assert.ok(doc.professionalId instanceof mongoose.Types.ObjectId);
   });
-    
+
   it('handles buffers stored as EJSON POJO (gh-14911)', async function() {
     const pdfSchema = new mongoose.Schema({
       pdfSettings: {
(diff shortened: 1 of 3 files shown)


test/document.test.js
```diff
@@ -13927,6 +13927,32 @@ describe('document', function() {
...
+      {
+        professionalId: {
+          type: Schema.Types.ObjectId
+        },
+        firstName: {
+          type: String
+        },
+        nested: {
+          test: String
+        }
```
hasezoey, line 13941:
```suggestion
        }
```
property seems unused and not relevant to this test

PR #14955, opened 2024-10-11
feat: allow defining virtuals on arrays, not just array elements

### LINKED ISSUE #2326: virtual path on array
in mongose guild doc there is a cool example. define `.full` virtual on sub documents `name: {first:String, last:String}`. then get virtual value by `name.full`.

but if i define virtual on array like this:

``` js
var schema = new Schema({
  nums: [Number]
});
schema.virtual('nums.odd').get(...).
```

i cannot get it by dot path `nums.odd`. only write like this `model.get('nums.odd')`.

is there a way to fix this?
i'm running mongoose@3.9.3

vkarpov15:
**Summary**

Neat little feature that allows defining virtuals on arrays, so you can define a `last` virtual on an array that gets you the last element in the array:

```js
schema.virtual('nums.last', { applyToArray: true }).get(function() {
  return this[this.length - 1];
});

// Later
doc.nums.last; // last element in the array
```

Fix #2326

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -13978,6 +13978,49 @@ describe('document', function() {
     assert.ok(Buffer.isBuffer(reloaded.pdfSettings.fileContent));
     assert.strictEqual(reloaded.pdfSettings.fileContent.toString('utf8'), 'hello');
   });
+
+  describe('gh-2306', function() {
+    it('allow define virtual on non-object path', function() {
+      const schema = new mongoose.Schema({ num: Number, str: String, nums: [Number] });
+      schema.path('nums').virtual('last').get(function() {
+        return this[this.length - 1];
+      });
+      schema.virtual('nums.first', { applyToArray: true }).get(function() {
+        return this[0];
+      });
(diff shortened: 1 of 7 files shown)


test/document.test.js
```diff
@@ -13978,6 +13978,49 @@ describe('document', function() {
...
+    });
+
+    it('works on document arrays', function() {
+      const schema = new mongoose.Schema({ books: [{ title: String, author: String }] });
+      schema.path('books').virtual('last').get(function() {
+        return this[this.length - 1];
+      });
+      schema.virtual('books.first', { applyToArray: true }).get(function() {
+        return this[0];
+      });
```
hasezoey, line 14016: just to make sure, these 2 ways (`.path().virtual()` & `.virtual('path', applyToArray)`) are equivalent?

hasezoey (approved):

test/document.test.js
vkarpov15: Yes they are equivalent.

PR #14957, opened 2024-10-11
docs(model): add more info on acknowledged: false, specifically that Mongoose may return that if the update was empty

### LINKED ISSUE #14951: acknowledged: false should also generate descriptive logging
### Prerequisites

### Issue

In certain cases, an result of {acknowledged: false} can be returned. The official docs say vaguely that this means "things didn't go smoothly". 

To resolve the issue of why it didn't go smoothly, it would be of great help if there were more logging about what exactly went wrong--  this would still be useful even if it was only emitted at the debug logging level. 

https://github.com/Automattic/mongoose/discussions/14936 is an example where a user is spending a lot of time how to resolve an "{acknowlwedged: false}" issue after an upgrade. 

The more precise the logging about this could be, the faster the true root cause can found and addressed. 

Ref: Here's where it's mentioned:

vkarpov15:
Fix #14951

**Summary**

Help clarify why #14951 is happening

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3771,7 +3771,7 @@ Model.hydrate = function(obj, projection, options) {
  *     const res = await Person.updateMany({ name: /Stark$/ }, { isDeleted: true });
  *     res.matchedCount; // Number of documents matched
  *     res.modifiedCount; // Number of documents modified
- *     res.acknowledged; // Boolean indicating everything went smoothly.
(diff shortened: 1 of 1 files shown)


markstos:
Looks good to me. If people end up after upgrading to Mongoose 6 because they are have started getting `acknowledged:false`, see this discussion about one possible reason that appeared for me during the Mongoose 6 upgrade:

https://github.com/Automattic/mongoose/discussions/14936

markstos (approved):

Issue #10983, opened 2021-11-17
findOne with nested refpath in large array of subdocuments causes out of memory exception


jessicagoldq4:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

A findOne() invocation, paired with a dynamic populate() call against a document with an array of subdocuments that contain a nested refpath, 'arrayName.typeField', causes an out of memory exception. This particularly happens when the subdocument array contains a large number of records (i.e. 2100). I don't fully understand the logic, so I will just list my findings

\------

During the Model.populate('arrayName.idField') -> getModelsMapForPopulate() -> _getModelNames() method invocation, the array of modelNames returned contains 2100 elements, each of which with a string value of the referred Model . Down the line, this causes the addModelNamesToMap() method to create a map with a very large memory footprint, where an individual model in the map has a total of 2.1 million records in it's allIds property, which just seems wrong:

![count undefined](https://user-images.githubusercontent.com/90337173/142239732-c4851d3d-7c12-41eb-bb49-eb6a779a22f0.png)

Each of the arrays in allIds are copies of each other, which brings into question why we need them all. In this scenario, we ultimately see an out of memory exception during an invocation of Model.populate('arrayName.idField') -> _done() -> _assign -> utils.clone(mod.allIds):

![RUNNING](https://user-images.githubusercontent.com/90337173/142239804-2d057ff3-9302-4396-a5b7-280bbbdd82b7.png)

This does not seem right. If we return the modelNames array from _getModelNames() with just 3 elements, 'Model1', 'Model2', and 'Model3', populate('arrayName.idField') seems to produce the correct results without this huge amount of memory usage. Maybe I don't really know what I'm talking about, but this seems like overkill

---

**If the current behavior is a bug, please provide the steps to reproduce.**

Pre-requisite: 1000+ elements exist in the document's 'items' array

```
const SampleSchema = new mongoose.Schema(
  {
    name: String,
    items: [{
      itemId: {
        type: mongoose.Schema.Types.ObjectId,
        required: true,
        refPath: 'items.type'
      },
      type: {
        type: String,
        required: true,
        enum: ['Model1, Model2, Model3']
      }
    }],
 },
 {
    timestamps: {
      createdAt: 'create_date',
      updatedAt: 'update_date'
    }
 }
)

const SampleModel = mongoose.model('Sample', SampleSchema)

async.waterfall([
      (cb) => {
        SampleModel
          .findOne({
[...]

Issue #6870, opened 2018-08-15
Dynamic references with different id types


joaohencke:
Hey there!

The problem: Dynamic ref (``refPath``) is not working when I have mixed content. 

Example: 

I have a model Test with an _id as Number and a model Lesson with an _id as a String.
Then, I have a Study Plan with an array of Parts and these parts have an array of Contents. 

```javascript
const mongoose = require('mongoose');
const db = mongoose.createConnection('mongodb://localhost:27017/dynamicRef');

const TestSchema = new mongoose.Schema({
    _id: Number,
    exercises: [String],
  });
const LessonSchema = new mongoose.Schema({
    _id: String,
    url: String,
  });
const StudyPlanSchema = new mongoose.Schema({
    parts: [
      {
        title: String,
        contents: [
          {
            item: {
              type: mongoose.Schema.Types.Mixed,
              refPath: 'parts.contents.kind',
            },
            kind: String,
          },
        ],
      },
    ],
  });
const StudyPlan = db.model('StudyPlan', StudyPlanSchema);
const Test = db.model('Test', TestSchema);
const Lesson = db.model('Lesson', LessonSchema);

const test = new Test({ _id: 123, exercises: ['t1', 't2'] });
const lesson = new Lesson({ _id: 'lesson', url: 'https://youtube.com' });
await Promise.all([test.save(), lesson.save()]);
const studyPlan = new StudyPlan({
    parts: [
      {
        title: 'Study Plan 01',
        contents: [
          // THIS CODE WORKS IF U COMMENT ONE OF THESE OBJECTS... IF I HAVE ONLY A TEST OR A LESSON, THE POPULATE WORKS FINE.
          {
            item: test._id,
            kind: 'Test',
          },
          {
            item: lesson._id,
            kind: 'Lesson',
          },
        ],
      },
    ],
  });

  await studyPlan.save();
```
When I have a Study Plan with a TEST and a LESSON in the array, the following populate command fails.
```javascript 
await StudyPlan.findOne({}).populate('parts.contents.item');
```
`` CastError: Cast to number failed for value "lesson" at path "_id" for model "Test" ``

Ps.: This code works if you change the _id type of Test and Lesson to ``ObjectId``

Versions:
* mongoose 5.2.8
* mongodb 4.0.1
* node.js 10.6.0

lineus:
here's a runnable repro script for this and it's output on 5.2.8:

### 6870.js
```js
#!/usr/bin/env node
'use strict';

[...]

PR #14965, opened 2024-10-15
fix(populate): handle array of ids with parent refPath

vkarpov15:
Re: #10983

Also related to #6870, that issue addresses a slightly different case where the `refPath` is within the same doc array as the id.

**Summary**

There was a bug in our fix for #10983 where we mishandle `refPath` in an array of ObjectIds where the `refPath` refers to a parent of the array.

```javascript
      const Parent = db.model(
        'Parent',
        new mongoose.Schema({
          docArray: [
            {
              type: {
                type: String,
                enum: ['Child', 'OtherModel']
              },
              ids: [
                {
                  type: mongoose.Schema.ObjectId,
                  refPath: 'docArray.type' // <-- note that `docArray.type` is not within the `ids` array
                }
              ]
            }
          ]
        })
      );
```

In `getModelsMapForPopulate()`, we end up with a case where we have a flat array of all ids and a flat array of all model names, but no way to match the two. In `flat.filter((val, i) => modelNamesForRefPath[i] === modelName);` we then end up accidentally filtering out incorrect ids.

Without flattening, we end up with the following list of model names:

```
[
  undefined, 'Child',
  'Child',   undefined,
  'Child',   undefined,
  'Child'
]
```

and the following ids:

```
[
  [],
  [
    new ObjectId('6671a008596112f0729c2045'),
    new ObjectId('667195f3596112f0728abe24'),
    new ObjectId('6671bd39596112f072cda69c'),
    new ObjectId('6672c351596112f072868565')
  ],
  [ new ObjectId('66734edd596112f0727304a2') ],
  [],
  [ new ObjectId('66726eff596112f072f8e834') ],
  [],
  [ new ObjectId('667267ff596112f072ed56b1') ]
]
```

We just need to correctly filter which ids belong to the `child` model by recursively going through the model names and ids.

**Examples**

### DIFF
--- a/lib/helpers/populate/getModelsMapForPopulate.js
+++ b/lib/helpers/populate/getModelsMapForPopulate.js
@@ -478,9 +478,10 @@ function addModelNamesToMap(model, map, available, modelNames, options, data, re
     return;
   }
 
-  let k = modelNames.length;
+  const flatModelNames = utils.array.flatten(modelNames);
+  let k = flatModelNames.length;
   while (k--) {
-    let modelName = modelNames[k];
+    let modelName = flatModelNames[k];
     if (modelName == null) {
       continue;
     }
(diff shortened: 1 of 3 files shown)


hasezoey (commented):
LGTM

PR #14968, opened 2024-10-16
types: improve goto definition for inferred schema definitions

forivall:
**Summary**

Improve Go to Definition for inferred schema definitions

**Examples**

In [schema.test.ts](https://github.com/Automattic/mongoose/blob/c187fd84d197e2f29759a157c798538661d36079/test/types/schema.test.ts#L471), add a type alias such as

```ts
  type InferredSchemaDecimal2 = InferredTestSchemaType['decimal2'];
```

Without this change, using "Go to Definition" does not work, and the typescript language service reports "No definition found". With this change, "Go to Definition" navigates directly to the definition in the schema.

[Screencast](https://github.com/user-attachments/assets/9d59082d-b592-48a7-9021-264ac46a2331)

I don't think there's a way to test for this with `tsd`, you just need to know the pattern to make it work nicely.


vkarpov15 (approved):
Thanks :+1:

PR #14972, opened 2024-10-17
docs(documents): add section on setting deeply nested properties, including warning about nullish coalescing assignment

### LINKED ISSUE #14944: Nullish coalescing assignment operator does not work as expected with mongoose.
### Prerequisites

### Mongoose version

7.2.2

### Node.js version

16.20.2

### MongoDB server version

7.0.2

5.5.4

### Description

When setting a field in an optional subdocument via [nullish coalescing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment), the incoming value is not set on the subdocument.

Usually, the nullish coalescing operator allows for patterns like this:

```ts
const obj: { nested?: { a?: number } } = {}

(obj.nested ??= {}).a = 5

console.log(obj.nested.a)
// 5
```

However, this does not work when used with a mongoose document:

```ts


(doc.nested ??= {}).a = 5

console.log(doc.nested.a)
// undefined
```

The reason is the fact that mongoose wraps the 'drive-by' assignment of `subdocument` in a `SingleNested` object. Nullish coalescing assignment returns the right hand operand if the left hand operand is `nullish`. Setting a field on that object (POJO) will not be picked up by mongoose, because `doc.nested` does not reference the POJO but rather the new object mongoose created under the hood.

### Steps to Reproduce

```ts
const { Schema } = mongoose

const schema = new Schema({
    nested: {
        a: Number
    }
})

const Model = mongoose.model('Test', schema)

const doc = new Model()

(doc.nested ??= {}).a = 5

console.log(doc.nested.a)
// undefined

```

The issue becomes even more apparent when looked at without the property assignment:

```ts
const doc = new Model()

const nested: (typeof doc)['nested'] = {}

const x = doc.nested ??= nested

console.log(x === nested)
// false <- Should be true!

```

### Expected Behavior

This is somewhat of an edge case as the nullish coalescing assignment operator might be considered quite exotic.
However, it's a big gotcha for people who don't know that mongoose routinely wraps and transforms objects under the hood.

Mongoose should not change how js operators function.
At the very least, the documentation should explicitly point out the behaviour.

vkarpov15:
Fix #14944

**Summary**

Add section in the docs about setting deeply nested paths. Clarify that `?.` and `??` are fine, we also support `.get()` and `.set()`, but `??=` comes with caveats.

**Examples**

### DIFF
--- a/docs/documents.md
+++ b/docs/documents.md
@@ -8,6 +8,7 @@ to documents as stored in MongoDB. Each document is an instance of its
   <li><a href="#documents-vs-models">Documents vs Models</a></li>
   <li><a href="#retrieving">Retrieving</a></li>
   <li><a href="#updating-using-save">Updating Using <code>save()</code></a></li>
+  <li><a href="#setting-nested-properties">Setting Nested Properties</a></li>
   <li><a href="#updating-using-queries">Updating Using Queries</a></li>
   <li><a href="#validating">Validating</a></li>
   <li><a href="#overwriting">Overwriting</a></li>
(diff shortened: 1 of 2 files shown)


docs/documents.md
```diff
@@ -81,6 +82,54 @@ doc.name = 'foo';
...
+```javascript
+// The following works fine
+const doc3 = new TestModel();
+doc3.nested.subdoc ??= {};
+doc3.nested.subdoc.name = 'John Smythe';
+
+// The following does **NOT** work.
+// Do not use the following pattern with Mongoose documents.
+const doc4 = new TestModel();
+(doc4.nested.subdoc ??= {}).name = 'Charlie Smith';
```
hasezoey, line 128: does this not work because mongoose is cloning on assignment? (i have tried in a node REPL)

```js
ret = (doc4.nested.subdoc ??= {})
ret === doc4.nested.subdoc // eval "false"
```

hasezoey (approved):

docs/documents.md
vkarpov15: Yes. The problem is the following:

```javascript
const x = {};
(doc4.nested.subdoc ??= x) === x; // true
doc4.nested.subdoc === x; // false
```

Basically `a.b ??= x` evaluates to `x`, not the value of `a.b` after the assignment. Which isn't a big deal in most cases, but with Mongoose the distinction matters.

We need to clone because `doc4.nested.subdoc` needs change tracking. This may be a case where using proxies instead of Object.defineProperty() for change tracking would help.

PR #14973, opened 2024-10-18
docs(migration-guide-v7): correct link to the section "Id Setter"

rb-ntnx:
**Summary**

Currently link to the section "Id Setter" is not working due to a missing id & the auto-generated one is not the correct one!


vkarpov15 (approved):
Thanks :+1:

PR #14982, opened 2024-10-23
fix(cursor): close underlying query cursor when calling destroy()

vkarpov15:
**Summary**

Implemented using `_destroy()` like recommended in Node docs here: https://nodejs.org/api/stream.html#writable_destroyerr-callback because that's the preferred way to handle async cleanup in destroy(), because destroy() itself needs to be sync.

**Examples**


hasezoey (approved):
LGTM, slight question though, why set `_closed` to false in the destroy case?

PR #14986, opened 2024-10-27
feat(query): add schemaLevelProjections option to query to disable schema-level select: false

### LINKED ISSUE #11474: Select entire document after using select: false

Question / feature

Currently, if we use `select: false` in a schema, we need to do `Model.findOne({ _id }).select("+field_a")` to get its value.

We could have an option to select the entire document, ignoring the `select: false`, it would not require us to do something like `Model.findOne({ _id }).select("+field_b +field_c +field_d +field_e ... +field_n")`. I feel this is necessary for some situations, but I couldn't figure a proper way to achieve this. I couldn't find anywhere an answer to this and I did not get an answer in [stack overflow](https://stackoverflow.com/questions/71300357/how-can-i-select-entire-document-in-mongoose-after-using-select-false). Any idea is welcome!

node v16.14.0
mongoose v6.2.4
mongodb v5.0.6

vkarpov15:
Fix #11474

**Summary**

Schema-level `select: false` is very useful, but there are cases where you want to disable `select: false` for all fields, like building an admin panel. This option will be very useful for mongoose studio.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -1143,6 +1143,38 @@ Query.prototype.select = function select() {
 };
 
 /**
+ * Enable or disable schema level projections for this query. Enabled by default.
+ * Set to `false` to include fields with `select: false` in the query result by default.
+ *
+ * #### Example:
+ *
+ *     const userSchema = new Schema({
+ *       email: { type: String, required: true },
+ *       passwordHash: { type: String, select: false, required: true }
+ *     });
+ *     const UserModel = mongoose.model('User', userSchema);
+ *
+ *     const doc = await UserModel.findOne().orFail().schemaLevelProjections(false);
+ *
+ *     // Contains password hash, because `schemaLevelProjections()` overrides `select: false`
+ *     doc.passwordHash;
+ *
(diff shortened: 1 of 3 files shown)


lib/query.js
```diff
@@ -1142,6 +1142,38 @@ Query.prototype.select = function select() {
...
+ *
+ *     // Contains password hash, because `schemaLevelProjections()` overrides `select: false`
+ *     doc.passwordHash;
+ *
+ * @method schemaLevelProjections
+ * @memberOf Query
+ * @instance
+ * @param {Boolean} value
+ * @return {Query} this
+ * @see SchemaType https://mongoosejs.com/docs/api/schematype.html
```
hasezoey, line 1167: why does this link to `SchemaType`?

hasezoey (approved):

lib/query.js
vkarpov15: Because SchemaType docs are where we have an explanation for `select: false` and `select: true`. I improved the link to link to a more specific header.

PR #14987, opened 2024-10-28
feat(model): add hideIndexes option to syncIndexes() and cleanIndexes()

### LINKED ISSUE #14868: syncIndex option to hide instead of drop index
### Prerequisites

### 🚀 Feature Proposal

I love the syncIndex() feature that helps to have a single source of truth for our MondoDB schema indexing.

However, dropping an index is never 100% sure, especially on large MongoDB databases with large codebase querying the collections. There can be some rarely used piece of code that need once a time an index and thus will not update the $indexStats frequently enough to identify that this index is actually needed.

One recommandation from the MongoDB team is to **hide** an index before actually dropping it. It will exclude the index from the query planner options and whenever you identify a performance issue due to this exclusion, you can instantly make it visible again. This is way faster and less resource consuming than re-creating the index from scratch. The latter can last for hours on large collections and your system suffers during this re-creation.

It would be nice to pass an option to the syncIndex(**{hide: true}**) to hide the indexes missing in the schema instead of dropping them immediately. Once the system has run for a certain period of time without problem, then we can call the syncIndex({hide:false}) to actually drop them.

### Motivation

Secure the synchronization of the indexes and prevent mis-dropping of rarely used indexes with immediate rollback if necessary.

### Example

```javascript
model.syncIndex({hide: true}); // hide missing schema indexes instead of dropping them
model.syncIndex(); // actually drop missing schema indexes. Same as {hide:false}
```

vkarpov15:
Fix #14868

**Summary**

Add an option to make `syncIndexes()` mark indexes as hidden (see [MongoDB hidden indexes docs](https://www.mongodb.com/docs/v5.1/core/index-hidden/)) as opposed to dropping them. Hidden indexes are useful because rebuilding an index can be expensive, so if you accidentally drop a index that you shouldn't it can take a while to recover. With hidden indexes, you can test whether removing this index will cause performance issues without actually dropping the index. `syncIndexes({ hideIndex: true })` will hide all non-schema indexes rather than dropping them.

Hidden index support was added in MongoDB 4.4 so we should make a quick note about version compatibility, since Mongoose 8 technically supports all the way back to MongoDB 3.6.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1220,7 +1220,7 @@ Model.createCollection = async function createCollection(options) {
  *
  * @param {Object} [options] options to pass to `ensureIndexes()`
  * @param {Boolean} [options.background=null] if specified, overrides each index's `background` property
- * @param {Boolean} [options.hideIndexes=false] set to `true` to hide indexes instead of dropping
+ * @param {Boolean} [options.hideIndexes=false] set to `true` to hide indexes instead of dropping. Requires MongoDB server 4.4 or higher
  * @return {Promise}
  * @api public
  */
@@ -1443,7 +1443,7 @@ function getIndexesToDrop(schema, schemaIndexes, dbIndexes) {
  *
  * @param {Object} [options]
  * @param {Array<String>} [options.toDrop] if specified, contains a list of index names to drop
- * @param {Boolean} [options.hideIndexes=false] set to `true` to hide indexes instead of dropping
+ * @param {Boolean} [options.hideIndexes=false] set to `true` to hide indexes instead of dropping. Requires MongoDB server 4.4 or higher
  * @return {Promise<String>} list of dropped index names
  * @api public
  */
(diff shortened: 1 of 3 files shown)


lib/model.js
```diff
@@ -1440,8 +1441,10 @@ function getIndexesToDrop(schema, schemaIndexes, dbIndexes) {
  *
  * The returned promise resolves to a list of the dropped indexes' names as an array
  *
- * @param {Function} [callback] optional callback
- * @return {Promise|undefined} Returns `undefined` if callback is specified, returns a promise if no callback.
+ * @param {Object} [options]
+ * @param {Array<String>} [options.toDrop] if specified, contains a list of index names to drop
+ * @param {Boolean} [options.hideIndexes=false] set to `true` to hide indexes instead of dropping. Requires MongoDB server 4.4 or higher
+ * @return {Promise<String>} list of dropped index names
```
hasezoey, line 1447:
```suggestion
 * @return {Promise<String>} list of dropped or hidden index names
```

does the hiding also return the index names? if yes this should likely be applied

hasezoey (approved):

lib/model.js
vkarpov15: Yep you're right, thank you for the suggestion

PR #15000, opened 2024-11-01
fix(query): add `overwriteImmutable` option to allow updating immutable properties without disabling strict mode

### LINKED ISSUE #8619: Mongoose update method cannot change createdAt field
Bug

Model update method does not change `createdAt` field!

I want to change createdAt field to newCreatedAt value. I wrote the bellow script and it wasn't useful:
```
        const result = await Donations.findOneAndUpdate(
           { _id },
           { createdAt: newCreatedAt },
         ).exec();
```
Instead of `findOneAndUpdate`, I tried with `update` method too and result is the same.
However, this script works!
```
        const [d] = await Donations.find({_id}).exec();
        d.createdAt = newCreatedAt;
        const result = await d.save();
```
Update script should be able to update the `createdAt` field.

mongoose: 5.1.3,
node: v10.9.0
Mongo: 4.2.2

vkarpov15:
Fix #8619

**Summary**

#8619 pointed out that we don't have docs in place for how to change a document's `createdAt` without bypassing Mongoose entirely, and the example in #8619 no longer works now that `createdAt` is an immutable property. So this PR adds `overwriteImmutable` property to queries to allow updating immutable properties.

**Examples**


docs/timestamps.md
```diff
@@ -236,3 +236,20 @@ await User.findOneAndUpdate({}, { $setOnInsert: { updatedAt: new Date() } }, {
   timestamps: { createdAt: true, updatedAt: false }
 });
 ```
+
+## Updating Timestamps
+
+If you need to disable Mongoose's timestamps and update a document's timestamps to a different value using `updateOne()` or `findOneAndUpdate()`, you need to do the following:
+
+1. Set the `timestamps` option to `false` to prevent Mongoose from setting `updatedAt`.
+2. Set `overwriteImmutable` to `false` to allow overwriting `createdAt`, which is an immutable property by default.
```
hasezoey, line 245:
```suggestion
2. Set `overwriteImmutable` to `true` to allow overwriting `createdAt`, which is an immutable property by default.
```

docs/timestamps.md
```diff
@@ -236,3 +236,20 @@ await User.findOneAndUpdate({}, { $setOnInsert: { updatedAt: new Date() } }, {
...
+
+```javascript
+const createdAt = new Date('2011-06-01');
+// Update a document's `createdAt` to a custom value. Normally Mongoose would prevent doing this because `createdAt`
+// is immutable.
```
hasezoey, line 250:
```suggestion
// Update a document's `createdAt` to a custom value.
// Normally Mongoose would prevent doing this because `createdAt` is immutable.
```

consider moving the entire second sentence to the second line.

lib/helpers/query/handleImmutable.js
```diff
@@ -2,7 +2,7 @@
 
 const StrictModeError = require('../../error/strict');
 
-module.exports = function handleImmutable(schematype, strict, obj, key, fullPath, ctx) {
+module.exports = function handleImmutable(schematype, strict, obj, key, fullPath, options, ctx) {
```
hasezoey, line 5: maybe consider adding some documentation to this function in the future

lib/query.js
```diff
@@ -3318,6 +3323,7 @@ function prepareDiscriminatorCriteria(query) {
  * @param {Boolean} [options.returnOriginal=null] An alias for the `new` option. `returnOriginal: false` is equivalent to `new: true`.
  * @param {Boolean} [options.translateAliases=null] If set to `true`, translates any schema-defined aliases in `filter`, `projection`, `update`, and `distinct`. Throws an error if there are any conflicts where both alias and raw property are defined on the same object.
  * @param {Boolean} [options.overwriteDiscriminatorKey=false] Mongoose removes discriminator key updates from `update` by default, set `overwriteDiscriminatorKey` to `true` to allow updating the discriminator key
+ * @param {Boolean} [options.overwriteImmutable=false] Mongoose removes updated immutable properties from `update` by default (excluding $setOnInsert), set `overwriteImmutable` to `true` to allow updating immutable properties using other update operators.
```
hasezoey, line 3326:
```suggestion
 * @param {Boolean} [options.overwriteImmutable=false] Mongoose removes updated immutable properties from `update` by default (excluding $setOnInsert). Set `overwriteImmutable` to `true` to allow updating immutable properties using other update operators.
```

i think it reads better if it is a new sentence

lib/query.js
```diff
@@ -4016,6 +4022,7 @@ Query.prototype._replaceOne = async function _replaceOne() {
  * @param {Boolean} [options.timestamps=null] If set to `false` and [schema-level timestamps](https://mongoosejs.com/docs/guide.html#timestamps) are enabled, skip timestamps for this update. Does nothing if schema-level timestamps are not set.
  * @param {Boolean} [options.translateAliases=null] If set to `true`, translates any schema-defined aliases in `filter`, `projection`, `update`, and `distinct`. Throws an error if there are any conflicts where both alias and raw property are defined on the same object.
  * @param {Boolean} [options.overwriteDiscriminatorKey=false] Mongoose removes discriminator key updates from `update` by default, set `overwriteDiscriminatorKey` to `true` to allow updating the discriminator key
+ * @param {Boolean} [options.overwriteImmutable=false] Mongoose removes updated immutable properties from `update` by default (excluding $setOnInsert), set `overwriteImmutable` to `true` to allow updating immutable properties using other update operators.
```
hasezoey, line 4025:
```suggestion
 * @param {Boolean} [options.overwriteImmutable=false] Mongoose removes updated immutable properties from `update` by default (excluding $setOnInsert). Set `overwriteImmutable` to `true` to allow updating immutable properties using other update operators.
```

lib/query.js
```diff
@@ -4086,6 +4093,7 @@ Query.prototype.updateMany = function(conditions, doc, options, callback) {
  * @param {Boolean} [options.timestamps=null] If set to `false` and [schema-level timestamps](https://mongoosejs.com/docs/guide.html#timestamps) are enabled, skip timestamps for this update. Note that this allows you to overwrite timestamps. Does nothing if schema-level timestamps are not set.
  * @param {Boolean} [options.translateAliases=null] If set to `true`, translates any schema-defined aliases in `filter`, `projection`, `update`, and `distinct`. Throws an error if there are any conflicts where both alias and raw property are defined on the same object.
  * @param {Boolean} [options.overwriteDiscriminatorKey=false] Mongoose removes discriminator key updates from `update` by default, set `overwriteDiscriminatorKey` to `true` to allow updating the discriminator key
+ * @param {Boolean} [options.overwriteImmutable=false] Mongoose removes updated immutable properties from `update` by default (excluding $setOnInsert), set `overwriteImmutable` to `true` to allow updating immutable properties using other update operators.
```
hasezoey, line 4096:
```suggestion
 * @param {Boolean} [options.overwriteImmutable=false] Mongoose removes updated immutable properties from `update` by default (excluding $setOnInsert). Set `overwriteImmutable` to `true` to allow updating immutable properties using other update operators.
```

hasezoey (approved):
One slight docs typo, otherwise suggestions.

lib/helpers/query/handleImmutable.js
vkarpov15: Good suggestion, fixed in 388c5b2

Issue #14948, opened 2024-10-10
findOne method returns first document, even when the query value passed is "undefined" - can create privacy issues


Alef5750:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the performance issue has not already been reported

### Last performant version

8.7.0

### Slowed down in version

8.7.0

### Node.js version

20.17.0

### 🦥 Performance issue

When using the findOne() method, passing it a query value of undefined, like such:
`User.findOne(undefined)`
the response is the first document in the `User` collection.
This is unique to Mongoose, and different than `db.collection.findOne(undefined)` in regular MongoDB, which returns `null` in such a case.

**The problem: privacy issues. A hacker can pass 'undefined' and retrieve first user in the database which may have sensitive information.**

### Steps to Reproduce

1. Create a User collection in a mongoDB database.
2. add a users to the database with some fields (username, password, etc.)
3. in your server's controller, run a function `getUser`:`export const getUser = async (req: Request, res: Response) => {
  try {
    const user = await User.findOne(undefined);
    **console.log(user);**
    res.status(200).json(user);
  } catch (error) {
    res.status(500).json({ message: "Error fetching user", error });
  }
};`

### Expected Behavior

Notice what the console logs as "user" to be the first user in your database, (and not `null`)

IslandRhythms:
I believe this is expected behavior. This is not the first time this issue has been raised. If I'm not confusing this with something else, what happens under the hood is that it takes out the undefined in the search query and so what gets sent to mongo is `{}` which would return one document. Changing this could have some consequences on people's environment's so its something that will need to be considered.

vkarpov15:
This is currently expected behavior, and not unique to Mongoose. The following script shows that `findOne(undefined)` also returns the first document for with the MongoDB Node.js driver

```javascript
'use strict';

const mongoose = require('mongoose');

void async function main() {
  await mongoose.connect('mongodb://127.0.0.1:27017/mongoose_test');

  const Test = mongoose.model('Test', mongoose.Schema({ name: String }));
  await Test.findOneAndUpdate({ name: 'test' }, { name: 'test' }, { upsert: true });

  // The following all print the first document in the db
  console.log(await Test.findOne(undefined));
  console.log(await Test.findOne(null));
  console.log(await Test.collection.findOne(undefined));

  // Throws "MongoServerError: Expected field filterto be of type object"
  console.log(await Test.collection.findOne(null));
}();

```

[...]

PR #15019, opened 2024-11-05
fix(model+query): make `findOne(null)`, `find(null)`, etc. throw an error instead of returning first doc

vkarpov15:
Re: #14948

**Summary**

As we found out in #14948, `findOne(null)` throws an error with the MongoDB Node driver, but Mongoose treats it as `findOne({})`. I do think that making `findOne(null)` and `findOne(undefined)` throw an error is more correct (while still allowing `findOne({})` and `findOne()` to return the first doc), and I think this change is a step in the right direction, but I'm wondering if we should put this in a minor release 8.10, or hold off until 9.0?

**Examples**

### DIFF
--- a/lib/error/objectParameter.js
+++ b/lib/error/objectParameter.js
@@ -17,10 +17,9 @@ const MongooseError = require('./mongooseError');
  */
 
 class ObjectParameterError extends MongooseError {
-
   constructor(value, paramName, fnName) {
     super('Parameter "' + paramName + '" to ' + fnName +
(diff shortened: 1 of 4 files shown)


hasezoey (approved):
LGTM, i dont know how much impact this gonna have, so maybe move it to 9.0 as a breaking change?

AbdelrahmanHafez (approved):
LGTM, this looks like a breaking change to me, would ship at v9.0.

PR #15020, opened 2024-11-05
BREAKING CHANGE: change `this` to HydratedDocument for default() and required(), HydratedDocument | Query for validate()

vkarpov15:
**Summary**

Currently, `this` is a raw document for `required()`, `default()`, and `validate()`. While this type works for many cases, it isn't fully correct. And the `validate()` also excludes the possibility that `this` is a Query in the case of query validators.

Re: #14696

This is a breaking change, and should go into 9.0 if merged.

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -1232,12 +1232,24 @@ async function gh13797() {
   interface IUser {
     name: string;
   }
-  new Schema<IUser>({ name: { type: String, required: function() {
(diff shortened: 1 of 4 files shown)


hasezoey (commented):
LGTM, this PR should likely target branch `9.0`, right?

vkarpov15:
Yep, I merged into 9.0

Issue #14906, opened 2024-09-23
Remove `_executionStack` due to performance issues


0Ams:
### Prerequisites

- [X] I have written a descriptive issue title

### Mongoose version

8.6.3

### Node.js version

v18.20.4	

### MongoDB version

4.4.25

### Operating system

Linux

### Issue

# Description:
I am currently in the process of upgrading Mongoose from version 4 to version 8 in our project, and I've encountered a significant performance issue.

# Summary:
* v4 performance: Acceptable speeds.
* v8 performance: Approximately 40% slower for findOne queries, despite find queries operating as expected (and even faster than before).

# Details:
- Indexes and schema remain identical between v4 and v8.
- No special configurations or optimizations were applied during the upgrade.
- find queries exhibit improved performance under v8, but findOne queries are noticeably slower.
- same setting & same environment & same index

# Steps to Reproduce:
- Run the same query using findOne on both Mongoose v4 and v8.
- Compare the query performance between the two versions.

# Expected Behavior:
- Performance should be on par or improved in v8, compared to v4.

# Actual Behavior:
- findOne queries are approximately 30% slower in v8 than in v4.

IslandRhythms:
Mongoose version 4.13.21 `1.938`ms
```javascript
const mongoose = require('mongoose');


const testSchema = new mongoose.Schema({
    name: String,
    val: Number
});


const Test = mongoose.model('Test', testSchema);

async function main() {
    for (let i = 0; i < 100; i++) {
        const doc = new Test();
        doc.name = 'Test',
        doc.val = i;
        doc.save(function(err) {

        });
    }
}

async function run() {
    Test.findOne({}, function (err, doc) {
        console.log('doc', doc);
    });
}

(async () => {
    await mongoose.connect('mongodb://localhost:27017/performance-test');
    // await mongoose.connection.dropDatabase();
    await main();
    console.time('performanceTest')
    await run();
    console.timeEnd('performanceTest');
    console.log(mongoose.version)
    process.exit(0);
})();
```

Mongoose version 8.7.0 `10.779`ms
```javascript
const mongoose = require('mongoose');


const testSchema = new mongoose.Schema({
    name: String,
    val: Number
});


const Test = mongoose.model('Test', testSchema);

async function main() {
    for (let i = 0; i < 100; i++) {
        await Test.create({
            name: 'Test',
            val: i
        })
    }
}

async function run() {
    const res = await Test.findOne();
    console.log(res);
}

(async () => {
[...]

PR #15022, opened 2024-11-06
perf: make a few micro-optimizations to help speed up findOne()

vkarpov15:
Re: #14906

**Summary**

Some small improvements that maybe give us an extra 1-2% on performance. Not huge, but a step in the right direction. I'm still debugging to see where we can find some extra perf.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -2248,6 +2248,9 @@ Query.prototype.error = function error(err) {
  */
 
 Query.prototype._unsetCastError = function _unsetCastError() {
+  if (this._error == null) {
+    return;
+  }
   if (this._error != null && !(this._error instanceof CastError)) {
     return;
   }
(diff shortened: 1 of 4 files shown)


hasezoey (approved):

lib/query.js
```diff
@@ -2291,9 +2293,9 @@ Query.prototype.mongooseOptions = function(v) {
 
 Query.prototype._castConditions = function() {
   let sanitizeFilterOpt = undefined;
-  if (this.model != null && utils.hasUserDefinedProperty(this.model.db.options, 'sanitizeFilter')) {
+  if (this.model != null && this.model.db.options?.sanitizeFilter != null) {
```
AbdelrahmanHafez, line 2296: Can also be shortened to:
```suggestion
  if (this.model?.db.options?.sanitizeFilter != null) {
```

lib/query.js
```diff
@@ -2291,9 +2293,9 @@ Query.prototype.mongooseOptions = function(v) {
...
     sanitizeFilterOpt = this.model.db.options.sanitizeFilter;
-  } else if (this.model != null && utils.hasUserDefinedProperty(this.model.base.options, 'sanitizeFilter')) {
+  } else if (this.model != null && this.model.base.options?.sanitizeFilter != null) {
```
AbdelrahmanHafez, line 2298: Same here
```suggestion
  } else if (this.model?.base.options?.sanitizeFilter != null) {
```

AbdelrahmanHafez (approved):
A couple of minor suggestions, otherwise LGTM.

lib/query.js
```diff
@@ -2248,6 +2247,9 @@ Query.prototype.error = function error(err) {
  */
 
 Query.prototype._unsetCastError = function _unsetCastError() {
+  if (this._error == null) {
+    return;
+  }
   if (this._error != null && !(this._error instanceof CastError)) {
```
BenaliDjamel, line 2253: @vkarpov15  Since you already checked `(this._error == null` above, you can simplify the condition as below. 
`if (!(this._error instanceof CastError)) {}`

vkarpov15: See #15104 :+1:

PR #15027, opened 2024-11-09
CI: add Node 22

stscoundrel:
**Summary**

Node 22 is the current Node.js version and LTS version. It was missing from CI test matrix.

I've already been running Mongoose on Node 22 for a while without issues, but it should be included in the official CI too. Therefore:

- Add 22 to test matrix, in addition to existing ones.
- Replace older single-version actions with 22. Some were run in Node 16, which is end-of-life. Some in 18, which will be end-of-life in six months. If an action is only run in one version, might as well be the latest.

For successful CI run, see the [PR in this fork](https://github.com/stscoundrel/mongoose/pull/1)


vkarpov15 (approved):
Thanks :+1:

PR #15039, opened 2024-11-13
fix: save execution stack in query as string

durran:
**Summary**

While debugging some unexpected memory usage in a production cluster, we noticed that mongoose's `Query` object was constructing a new `Error` instance in order to be able to preserve the original stack when throwing an error after attempting to execute the query more than once. This is problematic in that until the stack is converted to a string, the error retains references to every object in the call site info. This PR changes the execution stack object to the actual stack string to avoid this problem. Since this variable appears intended to not be part of the public API it seems reasonable to change.

**Examples**
Screenshot 2024-11-13 at 22 25 56 copy

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -4433,10 +4433,10 @@ Query.prototype.exec = async function exec(op) {
       str = str.slice(0, 60) + '...';
     }
     const err = new MongooseError('Query was already executed: ' + str);
-    err.originalStack = this._executionStack.stack;
+    err.originalStack = this._executionStack;
     throw err;
   } else {
-    this._executionStack = new Error();
+    this._executionStack = new Error().stack;
   }
 
   let skipWrappedFunction = null;


durran:
Note see related more detailed reasoning for a similar change in Node itself: https://github.com/nodejs/node/pull/34103#issuecomment-652002364

vkarpov15:
We also noticed some performance issues with the `_executionStack` pattern [here](https://github.com/Automattic/mongoose/issues/14906#issuecomment-2453595276). I think this PR confirms my suspicions that we should just remove `_executionStack` and `originalStack` in Mongoose 9.0 because those cause performance overhead on every single query, and the original rationale is somewhat obsolete. The original rationale was to help users debug duplicate queries, but the most common cause of duplicate queries was mixing `await` and callbacks, like `await Model.find({}, function(err, docs) {})`. Given that Mongoose no longer supports callbacks, duplicate queries are significantly less likely.

TLDR; thanks for the PR, LGTM. But we will probably remove the `_executionStack` logic in our next major release.

vkarpov15 (approved):

PR #15044, opened 2024-11-15
feat: add forceRepopulate option for populate() to allow avoiding repopulating already populated docs

### LINKED ISSUE #14979: Introduce a simpler way to "populate unless populated"
### 🚀 Feature Proposal

Is there a method to `populate` only if the value isn't already there? If not, maybe it's worth adding?

### Motivation

Currently, a `populate` call (both on a Model and an instance) queries the database, even if the value is already populated.

However, it's quite common that objects come from different code paths, and they may have paths populated already. So, performance-wise, we should not do it again.

For a single instance, we can write
```js
if (!user.populated('friends')) {
  await user.populate('friends');
}
```

...But it's cumbersome to repeat this often.

Also, for an array of `users` we need to check each item, which makes even longer code compared to a single call, e.g. `User.populateMissing(users, 'friends')`, which could work just as `User.populate`(single query to fetch all users), but omit already-populated paths.

### Additional thoughts

To say the truth, after many years of development with Mongoose, my experience is that the current behaviour (always querying the database) is much less needed than the lazy one (query only if populated path doesn't exist).

vkarpov15:
Fix #14979

**Summary**

It is a little strange that `doc.populate('child')` will populate 'child' even if 'child' is already populated. This PR adds an option to disable that behavior, either for an individual populate call using `doc.populate({ path: 'child', forceRepopulate: false })` or globally with `mongoose.set('forceRepopulate', false)`.

We should consider making `forceRepopulate` default to false for Mongoose 9.

**Examples**

### DIFF
--- a/test/model.populate.test.js
+++ b/test/model.populate.test.js
@@ -11279,4 +11279,89 @@ describe('model: populate:', function() {
     assert.strictEqual(doc.node.length, 1);
     assert.strictEqual(doc.node[0]._id, '65c7953e-c6e9-4c2f-8328-fe2de7df560d');
   });
+
+  it('avoids repopulating if forceRepopulate disabled (gh-14979)', async function() {
+    const ChildSchema = new Schema({ name: String });
+    const ParentSchema = new Schema({
+      children: [{ type: Schema.Types.ObjectId, ref: 'Child' }],
+      child: { type: 'ObjectId', ref: 'Child' }
+    });
+
+    const Child = db.model('Child', ChildSchema);
+    const Parent = db.model('Parent', ParentSchema);
+
+    const child = await Child.create({ name: 'Child test' });
+    let parent = await Parent.create({ child: child._id, children: [child._id] });
+
+    parent = await Parent.findOne({ _id: parent._id }).populate(['child', 'children']).orFail();
+    child.name = 'Child test updated 1';
+    await child.save();
+
+    await parent.populate({ path: 'child', forceRepopulate: false });
+    await parent.populate({ path: 'children', forceRepopulate: false });
+    assert.equal(parent.child.name, 'Child test');
+    assert.equal(parent.children[0].name, 'Child test');
+
+    await Parent.populate([parent], { path: 'child', forceRepopulate: false });
+    await Parent.populate([parent], { path: 'children', forceRepopulate: false });
+    assert.equal(parent.child.name, 'Child test');
+    assert.equal(parent.children[0].name, 'Child test');
+
+    parent.depopulate('child');
+    parent.depopulate('children');
+    await parent.populate({ path: 'child', forceRepopulate: false });
+    await parent.populate({ path: 'children', forceRepopulate: false });
+    assert.equal(parent.child.name, 'Child test updated 1');
+    assert.equal(parent.children[0].name, 'Child test updated 1');
+  });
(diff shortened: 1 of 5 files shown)


test/model.populate.test.js
```diff
@@ -11279,4 +11279,89 @@ describe('model: populate:', function() {
     assert.strictEqual(doc.node.length, 1);
     assert.strictEqual(doc.node[0]._id, '65c7953e-c6e9-4c2f-8328-fe2de7df560d');
   });
+
+  it('avoids repopulating if forceRepopulate disabled (gh-14979)', async function() {
```
hasezoey, line 11283:
```suggestion
  it('avoids repopulating if forceRepopulate is disabled (gh-14979)', async function() {
```

hasezoey (approved):

AbdelrahmanHafez (approved):
LGTM, I also agree it should default to `false` in v9.0.

One minor addition I would introduce here is a TS type test.

igoradamenko:
@vkarpov15, I believe that the introduced flag does not work with `populate` option field that is used for nested population.

The flag should probably be passed somewhere around here to the recursive `model.populate` call:

https://github.com/Automattic/mongoose/blob/dd39133c004d07770ee755d755ad7702dd7e0d63/lib/model.js#L4552-L4557

I have a pretty vague understanding of Mongoose internals, just noticed that `{ path: 'a', populate: 'b', forceRepopulate: false }` does not work for me. But when it added `forceRepopulate: populateOptions.forceRepopulate` to the `opts` in the code above, it worked.

Also, if I understand correctly, it's not intended that `Model.populate` accepts lean documents (POJO), right? I mean, I used it that way for lean docs and everything worked fine, until I needed to disable forced repopulation. It does not work with lean documents since the check is used `doc.populated` that does not exists on POJO. I see that `Model.populate` doc describes that first argument should be a Document or array of them. What is the valid way to populate lean documents, if it's even possible?

vkarpov15:
@igoradamenko are you using `Model.populate()` with `forceRepopulate`? `forceRepopulate` is only supported for `doc.populate()`, not `Model.populate()`.

igoradamenko:
@vkarpov15, well, then the doc is wrong, is not it? It says that `Model.populate` has this option:

https://mongoosejs.com/docs/api/model.html#Model.populate()

I mean, I can be wrong here, but `Model.populate` works to me 🤔 

Tbh, it looks like it should work:

1. `Model.populate` runs `_populatePath` here: https://github.com/Automattic/mongoose/blob/18f2ca9b2305d427d96457af4555eacae04b6671/lib/model.js#L4438

2. `_populatePath` runs `getModelsMapForPopulate` here: https://github.com/Automattic/mongoose/blob/18f2ca9b2305d427d96457af4555eacae04b6671/lib/model.js#L4471

5. `getModelsMapForPopulate` checks for `forceRepopulate` here: https://github.com/Automattic/mongoose/blob/18f2ca9b2305d427d96457af4555eacae04b6671/lib/helpers/populate/getModelsMapForPopulate.js#L57-L62

Again, not trying to be right in any case here. I just noticed that this feature did not work as I expected, and I decided to not to use it. But I thought that I might help to identify some wrong behavior. If it's all working as expected, than I probably got the idea wrong.

PR #15054, opened 2024-11-22
feat(NODE-6503): Add Int32 to SchemaType

aditi-khare-mongoDB:
This [PR follows the one opened on the mongodb fork](https://github.com/mongodb-js/mongoose/pull/1,), Val's initial round of feedback has been addressed.
### Summary

**Motivation**
Support a 32-bit integer type in mongoose for compatibility with CSFLE/QE. The current `Number` schema type can result in either Int32, Int64, or Doubles that are inserted into the database using type inference but CSFLE and QE require exact BSON types in schemas.

**Summary of Changes**
- Add `SchemaType` `Int32`, which has the following behavior:
  - its default `cast` method throws when provided a 
    - a non-numeric string
    - a non-integer value
    - a `NaN`
    - a number beyond the bounds of INT32_MAX or INT32_MIN
  - it is queryable as `$type: int` or `$type: number`, but NOT `$type: double`
 - Add API Docs for new query option and new schema type

### Example

```
const vehicleSchema = new Schema({ numberOfTires: mongoose.Int32 });
```


aditi-khare-mongoDB:
That works for me, thanks!

vkarpov15 (approved):
LGTM. I will merge this into 8.9 branch.

PR #15055, opened 2024-11-22
perf: cache results from getAllSubdocs() on saveOptions, only loop through known subdoc properties

### LINKED ISSUE #15029: mongose docReducer function inside __getAllSubdocs taking server hgh cpu even if my use-case doesn't involve embedded docs at all?
### Prerequisites

- [x] I have searched existing issues to ensure the performance issue has not already been reported

### Last performant version

4.9.9

### Slowed down in version

4.9.9

### Node.js version

v16.16.0

### 🦥 Performance issue

In the Mongoose lib CPU profiling, I noticed that the docReducer function present in document.js is consuming 30% of Application CPU resources. Upon further investigation, I found multiple calls to __getAllSubdocs during the document save call, which calls docReducer. This function recursively checks each key in the document to see if there are embedded documents. However, my collection does not contain embedded documents(embedded is not needed for my usecase), and I would like to avoid this overhead. Is there a flag to skip the docReducer call, or do you guys have any suggestions on how to handle this? Or what can be done here?

Shall i comment the code var subDocs = Object.keys(this._doc).reduce(docReducer.bind(this), []); and simply return []; ?

In the latest Mongo version docReducer function is optimized but its still there with big overhead.

`
        
    Document.prototype.$__getAllSubdocs = function() {
    DocumentArray || (DocumentArray = require('./types/documentarray'));
    Embedded = Embedded || require('./types/embedded');

      function docReducer(seed, path) {
        var val = this[path];

    if (val instanceof Embedded) {
      seed.push(val);
    }
    
    if (val && val.$isSingleNested) {
      seed = Object.keys(val._doc).reduce(docReducer.bind(val._doc), seed);
      seed.push(val);
    }
    
    if (val && val.isMongooseDocumentArray) {
      val.forEach(function _docReduce(doc) {
        if (!doc || !doc._doc) {
          return;
        }
        if (doc instanceof Embedded) {
          seed.push(doc);
        }
        seed = Object.keys(doc._doc).reduce(docReducer.bind(doc._doc), seed);
      });
    } else if (val instanceof Document && val.$__isNested) {
      if (val) {
        seed = Object.keys(val).reduce(docReducer.bind(val), seed);
      }
    }
    
    return seed;
      }

      var subDocs = Object.keys(this._doc).reduce(docReducer.bind(this), []); 
      return subDocs; 
    // shall i comment var subDocs = Object.keys(this._doc).reduce(docReducer.bind(this), []);  ? and return []
    };

  

`

### Steps to Reproduce

calling mongoose save docReducer of __getAllSubdocs gets called

### Expected Behavior

_No response_

vkarpov15:
Fix #15029

**Summary**

#15029 points out that we spend a lot of unnecessary CPU cycles in `save()` in `$getAllSubdocs()`, even for schemas that have no subdocs. That's because currently `save()` calls `$getAllSubdocs()` 5 times, and `$getAllSubdocs()` loops through every path in the document searching for subdocs.

With this PR, we will only call `$getAllSubdocs()` 1x if calling `save()` with no subdocs, and also `$getAllSubdocs()` will only loop through paths we know can have subdocs based on the schema. So if the schema doesn't have any subdocuments, document arrays, or maps of subdocs, then `$getAllSubdocs()` will have an empty `for` loop.

In my quick experiment with the `saveSimple` benchmark, I'm seeing a 2.7% improvement in performance, which isn't huge, but that's 2.7% on every save call.

I'll also work on adding a benchmark with subdocuments to see how much this helps with subdocs.

**Examples**

### DIFF
--- a/benchmarks/saveSimple.js
+++ b/benchmarks/saveSimple.js
@@ -1,0 +1,57 @@
+'use strict';
+
+const mongoose = require('../');
+
+run().catch(err => {
+  console.error(err);
+  process.exit(-1);
+});
+
+async function run() {
+  await mongoose.connect('mongodb://127.0.0.1:27017/mongoose_benchmark');
+  const FooSchema = new mongoose.Schema({
+    prop1: String,
+    prop2: String,
+    prop3: String,
+    prop4: String,
+    prop5: String,
+    prop6: String,
+    prop7: String,
+    prop8: String,
+    prop9: String,
+    prop10: String
+  });
+  const FooModel = mongoose.model('Foo', FooSchema);
+
+  if (!process.env.MONGOOSE_BENCHMARK_SKIP_SETUP) {
+    await FooModel.deleteMany({});
+  }
+
+  const numIterations = 500;
+  const saveStart = Date.now();
+  for (let i = 0; i < numIterations; ++i) {
+    for (let j = 0; j < 10; ++j) {
+      const doc = new FooModel({
+        prop1: `test ${i}`,
+        prop2: `test ${i}`,
+        prop3: `test ${i}`,
+        prop4: `test ${i}`,
+        prop5: `test ${i}`,
+        prop6: `test ${i}`,
+        prop7: `test ${i}`,
+        prop8: `test ${i}`,
+        prop9: `test ${i}`,
+        prop10: `test ${i}`
+      });
+      await doc.save();
+    }
+  }
+  const saveEnd = Date.now();
+
+  const results = {
+    'Average save time ms': +((saveEnd - saveStart) / numIterations).toFixed(2)
+  };
+
+  console.log(JSON.stringify(results, null, '  '));
+  process.exit(0);
+}
--- a/lib/document.js
+++ b/lib/document.js
@@ -2711,7 +2711,7 @@ function _getPathsToValidate(doc, pathsToValidate, pathsToSkip, isNestedValidate
 
   if (!isNestedValidate) {
     // If we're validating a subdocument, all this logic will run anyway on the top-level document, so skip for subdocuments
-    const subdocs = doc.$getAllSubdocs();
+    const subdocs = doc.$getAllSubdocs({ useCache: true });
     const modifiedPaths = doc.modifiedPaths();
     for (const subdoc of subdocs) {
       if (subdoc.$basePath) {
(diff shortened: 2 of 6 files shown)


vkarpov15:
Quick check with the `createDeepNestedDocArray` benchmark, which is a good example of `save()` with a lot of very deeply nested subdocs, shows a 5% speedup.

billouboq:
Does it improves performance of create or insertMany ?

benchmarks/saveSimple.js
```diff
@@ -0,0 +1,57 @@
...
+  const FooModel = mongoose.model('Foo', FooSchema);
+
+  if (!process.env.MONGOOSE_BENCHMARK_SKIP_SETUP) {
+    await FooModel.deleteMany({});
+  }
+
+  const numIterations = 500;
+  const saveStart = Date.now();
+  for (let i = 0; i < numIterations; ++i) {
+    for (let j = 0; j < 10; ++j) {
```
hasezoey, line 33: why are there 2 loops with `j` being unused (from what i can tell)?

lib/document.js
```diff
@@ -3679,57 +3679,50 @@ Document.prototype.$__getArrayPathsToValidate = function() {
  * @instance
  */
 
-Document.prototype.$getAllSubdocs = function() {
+Document.prototype.$getAllSubdocs = function(options) {
```
hasezoey, line 3682: the jsdoc should likely be updated to reflect the new parameter

hasezoey (approved):
LGTM, aside from the minor jsdoc change necessary. I would recommend to put this into the next minor version for caution.

hasezoey:
@billouboq 

> Does it improves performance of create or insertMany ?

It will affect any function making use of `$getAllSubdocs`, which by extension is `$validate` (which is then made use of in `insertMany`, `save`, etc), thought to different extends depending on previous amount of calls to `$getAllSubdocs`.
(note that `create` is [using `.save`](https://github.com/Automattic/mongoose/blob/b34aba65bd64540e330665477f542eb79c877909/lib/model.js#L2537), so any changes to `save` also affects `create` directly).

benchmarks/saveSimple.js
vkarpov15: Easier to track the perf difference in 10 sequential saves than 1 save, because 1 save is still very fast.

vkarpov15:
`insertMany()` uses `validate()` under the hood, so this should help a bit on `insertMany()` too. Definitely on `create()`, because `create()` calls `save()`.

PR #15059, opened 2024-11-25
feat: allow specifying error message override for duplicate key errors `unique: true`

vkarpov15:
**Summary**

Allow overriding MongoDB's default "duplicate key error" message in your schemas:

```javascript
new Schema({
  email: { type: String, unique: [true, 'Email must be unique'] }
});
```

This works for `save()` and query methods, but does **not** work for `bulkWrite()` and `insertMany()` because the duplicate key error in that case does not return `keyPattern`, which means we would either have to guess the duplicate key by the index name (brittle) or parse the index key from the error message (we would have to pull in Acorn or something like that, because duplicate key error message is not formatted as JSON).

**Examples**

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -2799,8 +2799,13 @@ Schema.prototype._getPathType = function(path) {
   return search(path.split('.'), _this);
 };
 
-/*!
+/**
+ * Transforms the duplicate key error by checking for duplicate key error messages by path.
+ * If no duplicate key error messages are found, returns the original error.
  *
+ * @param {Error} error The error to transform
+ * @returns {Error} The transformed error
+ * @api private
  */
 
 Schema.prototype._transformDuplicateKeyError = function _transformDuplicateKeyError(error) {
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -2800,6 +2800,34 @@ Schema.prototype._getPathType = function(path) {
 };
 
 /*!
+ *
+ */
+
+Schema.prototype._transformDuplicateKeyError = function _transformDuplicateKeyError(error) {
+  if (!this._duplicateKeyErrorMessagesByPath) {
+    return error;
+  }
+  if (error.code !== 11000 && error.code !== 11001) {
+    return error;
+  }
+
+  if (error.keyPattern != null) {
+    const keyPattern = error.keyPattern;
+    const keys = Object.keys(keyPattern);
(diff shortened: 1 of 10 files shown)


lib/schema.js
```diff
@@ -2799,6 +2799,34 @@ Schema.prototype._getPathType = function(path) {
   return search(path.split('.'), _this);
 };
 
+/*!
+ *
+ */
```
hasezoey, line 2804: this should likely have docs or a `ignore` comment like other places

lib/schemaType.js
```diff
@@ -447,21 +446,38 @@ SchemaType.prototype.index = function(options) {
...
+ *     unique: [true, 'Email must be unique']
+ *   }
+ * });
+ * ```
+ *
+ * Note that the above syntax does **not** work for `bulkWrite()` or `insertMany()`. `bulkWrite()` and `insertMany()`
+ * will still throw MongoDB's default `E11000 duplicate key error` message.
+ *
+ * @param {Boolean} value
+ * @param {String} message
```
hasezoey, line 466:
```suggestion
 * @param {String} [message]
```

if i understand correctly, `message` can be optional (theoretically `value` could be too)

types/index.d.ts
```diff
@@ -309,7 +309,7 @@ declare module 'mongoose' {
     eachPath(fn: (path: string, type: SchemaType) => void): this;
 
     /** Defines an index (most likely compound) for this schema. */
-    index(fields: IndexDefinition, options?: IndexOptions): this;
+    index(fields: IndexDefinition, options?: Omit<IndexOptions, 'unique'> & { unique?: boolean | undefined | [true, string] }): this;
```
hasezoey, line 312:
```suggestion
    index(fields: IndexDefinition, options?: Omit<IndexOptions, 'unique'> & { unique?: boolean | [true, string] }): this;
```
i think this `undefined` can be omitted because property is already marked `?`

hasezoey (approved):
LGTM, minor docs things

AbdelrahmanHafez (approved):
LGTM

types/index.d.ts
vkarpov15: Confirmed. I thought the `undefined` was necessary to support `unique: undefined` as opposed to omitting `unique`, but that doesn't look to be the case.

PR #15061, opened 2024-11-26
feat(NODE-6504): Add Double as a SchemaType

aditi-khare-mongoDB:
### Summary

**Motivation**
Support a IEEE 754-2008 floating point type in mongoose for compatibility with CSFLE/QE. The current `Number` schema type can result in either Int32, Int64, or Doubles that are inserted into the database using type inference but CSFLE and QE require exact BSON types in schemas.

The reason the Double constructor returns a BSON.Double is so even when an integer is inserted, the field is still queryable as a `$double`:

```
  await Test.create({ myDouble: '42' });
  const doc = await Test.findOne({ myDouble: { $type: 'double' } });
  assert.deepEqual(doc.myDouble, new BSON.Double(42));
```

**Summary of Changes**
- Same changes as NODE-6503 but for Double.
- Add `SchemaType` `Double`, which has the following behavior:
  - its default `cast` method throws when provided a 
    - a non-numeric string
  - it is queryable as `$type: double` or `$type: number`
 - Add API Docs for new query option and new schema type

### Example

```
const vehicleSchema = new Schema({ gasLevel: mongoose.Schema.Types.Double });
```

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -80,6 +80,7 @@ The permitted SchemaTypes are:
 * [Decimal128](api/mongoose.html#mongoose_Mongoose-Decimal128)
 * [Map](schematypes.html#maps)
 * [UUID](schematypes.html#uuid)
+* [Double](schematypes.html#double)
 
 Read more about [SchemaTypes here](schematypes.html).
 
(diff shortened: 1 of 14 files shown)


vkarpov15 (approved):
Merging into 8.9

vkarpov15:
There's some test failures and lint issues from merge conflicts, I'll fix those in the 8.9 branch

PR #15072, opened 2024-12-03
types: only apply BufferToBinary if type strictly equals Buffer | null | undefined

### LINKED ISSUE #15057: Mongoose 8.7.2 breaks Union Types
### Prerequisites

### Mongoose version

8.7.2

### Node.js version

22

### MongoDB server version

6.11

5.7.2

### Description

Before 8.7.2, only the type of the union was returned and that was fine. But now the type is returned as binary, but the union has no binaries in it.
It does not matter that the correct union value is not parsed here by mongoose, as this is done by zod.
This error seems to be related to this issue: https://github.com/Automattic/mongoose/issues/14974

### Steps to Reproduce

```ts
import mongoose, { Model } from "mongoose";

type Attachment =
  | {
      type: "foo";
      value?: undefined;
    }
  | {
      type: "string";
      value?: string;
    };

const TestSchema = new mongoose.Schema<Attachment>({
  value: { type: String },
});

export const AttachmentModel = mongoose.model<Attachment>("test", TestSchema);

const main = async () => {
  const item = await AttachmentModel.findOne().lean();

  if (!item) return;

  doSomeThing(item);
};

const doSomeThing = (item: Attachment) => {
  console.log(item);
};


```

`doSomeThing` returns error
Image

### Expected Behavior

_No response_

vkarpov15:
Fix #15057

**Summary**

The issue in #15057 is  that BufferToBinary adds `Binary` to any type that `extends undefined` or `extends null`, which is not intentional. This PR now only applies the BufferToBinary transform if the type strictly equals Buffer | null | undefined. I also made a similar change to other transforms we use for calculating the JSON serialized type.

**Examples**

### DIFF
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -206,3 +206,34 @@ async function gh13382() {
   const res = await Test.updateOne({}, { name: 'bar' }).lean();
   expectAssignable<{ matchedCount: number, modifiedCount: number }>(res);
 }
+
+async function gh15057() {
+  type Attachment =
+    | {
+        type: 'foo';
+        value?: undefined;
+      }
+    | {
+        type: 'string';
+        value?: string;
+      };
+
+  const TestSchema = new Schema<Attachment>({
+    type: { type: String, required: true },
+    value: { type: String }
+  });
+
+  const AttachmentModel = model<Attachment>('test', TestSchema);
+
+  const main = async() => {
+    const item = await AttachmentModel.findOne().lean();
+
+    if (!item) return;
+
+    doSomeThing(item);
+  };
+
+  const doSomeThing = (item: Attachment) => {
+    console.log(item);
+  };
+}
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -715,13 +715,16 @@ declare module 'mongoose' {
   export type BufferToBinary<T> = T extends TreatAsPrimitives ? T : T extends Record<string, any> ? {
     [K in keyof T]: T[K] extends Buffer
       ? mongodb.Binary
-      : T[K] extends (Buffer | null | undefined)
-        ? mongodb.Binary | null | undefined
-        : T[K] extends Types.DocumentArray<infer ItemType>
+      : IfEquals<
+        T[K],
+        Buffer | null | undefined,
+        mongodb.Binary | null | undefined,
+        T[K] extends Types.DocumentArray<infer ItemType>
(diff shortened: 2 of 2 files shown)


hasezoey (commented):
This PR causes typegoose (in its current state) fail to build (types):

```txt
src/typeguards.ts:21:12 - error TS2677: A type predicate's type must be assignable to its parameter's type.
  Type 'Array<DocumentType<NonNullable<T>>>' is not assignable to type 'Array<Ref<T, S>>'.
    Type 'DocumentType<NonNullable<T>>' is not assignable to type 'Ref<T, S>'.
      Type 'DocumentType<NonNullable<T>>' is not assignable to type 'DocumentType<T, BeAnObject>'.
        Type 'DocumentType<NonNullable<T>>' is not assignable to type 'Document<unknown, BeAnObject, T>'.

21 ): docs is mongoose.Types.Array<DocumentType<NonNullable<T>>>;
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/typeguards.ts:27:102 - error TS2677: A type predicate's type must be assignable to its parameter's type.
  Type 'DocumentType<NonNullable<T>>[]' is not assignable to type 'Ref<T, S>[]'.
    Type 'DocumentType<NonNullable<T>>' is not assignable to type 'Ref<T, S>'.
      Type 'DocumentType<NonNullable<T>>' is not assignable to type 'DocumentType<T, BeAnObject>'.
        Type 'DocumentType<NonNullable<T>>' is not assignable to type 'Document<unknown, BeAnObject, T>'.

27 export function isDocumentArray<T, S extends RefType>(docs: Ref<T, S>[] | null | undefined): docs is DocumentType<NonNullable<T>>[];
                                                                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/typeguards.ts:21
```

removing the `NonNull<>` from [those 2 `is` locations](https://github.com/typegoose/typegoose/blob/cc755df79c95ab5d9e2261cc44534c730a9ef11f/src/typeguards.ts#L21-L27) seemingly fixes it, though it is quite weird to not be able to assert that `T` is *not* `null` or `undefined` anymore. (though it would be in-line with the `isDocument` definition from typegoose)

vkarpov15:
Closing this in favor of #15103

PR #15080, opened 2024-12-05
fix: mark documents that are populated using hydratedPopulatedDocs option as populated in top-level doc

### LINKED ISSUE #15048: Hydrate populated fields not mark them as populated in the document
### Prerequisites

### Mongoose version

8.8.2

### Node.js version

20.7.0

### MongoDB server version

5.0.2

5.6.3

### Description

On hydrating a document with populated fields (`hydratedPopulatedDocs: true`), they doesn't pass `document.populated(...)` check.

### Steps to Reproduce

```js
  name: String
});
const companySchema = new Schema({
  name: String,
  users: [{ ref: 'User', type: Schema.Types.ObjectId }]
});

db.model('UserTestHydrate', userSchema);
const Company = db.model('CompanyTestHydrate', companySchema);

const users = [{ _id: new mongoose.Types.ObjectId(), name: 'Val' }];
const company = { _id: new mongoose.Types.ObjectId(), name: 'Booster', users: [users[0]] };

const C = Company.hydrate(company, null, { hydratedPopulatedDocs: true });
assert(C.populated('users')); // fails
```

AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
assert(C.populated('users'))

### Expected Behavior

Hydrating populated fields should write population info in the document to make `populated` method works.

vkarpov15:
Fix #15048

**Summary**

#15048 points out that `hydratedPopulatedDocs` option for `hydrate()` doesn't mark the paths as `populated()`. Also turns out that `hydrate()` doesn't actually hydrate the nested populated docs.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -624,16 +624,17 @@ Document.prototype.toBSON = function() {
 };
 
 /**
- * Initializes the document without setters or marking anything modified.
+ * Hydrates this document with the data in `doc`. Does not run setters or mark any paths modified.
  *
- * Called internally after a document is returned from mongodb. Normally,
+ * Called internally after a document is returned from MongoDB. Normally,
  * you do **not** need to call this function on your own.
  *
  * This function triggers `init` [middleware](https://mongoosejs.com/docs/middleware.html).
  * Note that `init` hooks are [synchronous](https://mongoosejs.com/docs/middleware.html#synchronous).
  *
- * @param {Object} doc document returned by mongo
+ * @param {Object} doc raw document returned by mongo
  * @param {Object} [opts]
+ * @param {Boolean} [opts.hydratedPopulatedDocs=false] If true, hydrate and mark as populated any paths that are populated in the raw document
  * @param {Function} [fn]
  * @api public
  * @memberOf Document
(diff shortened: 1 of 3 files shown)


test/model.hydrate.test.js
```diff
@@ -173,5 +176,26 @@ describe('model', function() {
...
+    it('sets hydrated docs as populated (gh-15048)', async function() {
+      const userSchema = new Schema({
+        name: String
+      });
+      const companySchema = new Schema({
+        name: String,
+        users: [{ ref: 'User', type: Schema.Types.ObjectId }]
+      });
+
+      db.deleteModel(/User/);
```
hasezoey, line 189:
```suggestion
      db.deleteModel(/User/);
      db.deleteModel(/Company/);
```

lib/document.js
```diff
@@ -805,6 +805,14 @@ function init(self, obj, doc, opts, prefix) {
               reason: e
             }));
           }
+        } else if (opts.hydratedPopulatedDocs) {
```
hasezoey, line 808: i just noticed the option `hydratedPopulatedDocs` is not documented in the `init` function here.
(also no options are documented in `Document.prototype.init` which is public)

hasezoey (commented):
LGTM

Issue #15026, opened 2024-11-08
Random error when using refPath & strictPopulate on deep subdocs


BobBatard:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.7.1

### Node.js version

20.7.0

### MongoDB server version

8.0.1

### Description

Some context: I have a parent model, which contains a property `child` that stores an `ObjectId` pointing to either a `childA` or `childB` model, with the model being specified in a `childType` property. The value of `childType` is passed to the `refPath` of the `child` property.

`childA` has a property `subChildA`, and `subChildA` has a property `grandChildA`.  
`childB` has no properties.

```
parent  > childA > subChildA > grandChildA
        > childB
```

When performing a `find` on `parent` and populating the entire chain of `children`, the `populate` function sometimes mixes up the models, resulting in an error indicating that `child.subChildA.grandChildA` is not in the schema.

```
/node_modules/mongoose/lib/helpers/populate/getModelsMapForPopulate.js:50
    return new StrictPopulate(options._fullPath || options.path);
           ^

StrictPopulateError: Cannot populate path `child.subChildA.grandChildA` because it is not in your schema. Set the `strictPopulate` option to false to override.
    at getModelsMapForPopulate (/Users/bob/Documents/work/inmemori/crm_api/node_modules/mongoose/lib/helpers/populate/getModelsMapForPopulate.js:50:12)
    at populate (/Users/bob/Documents/work/inmemori/crm_api/node_modules/mongoose/lib/model.js:4245:21)
    at _populate (/Users/bob/Documents/work/inmemori/crm_api/node_modules/mongoose/lib/model.js:4205:5)
    at /Users/bob/Documents/work/inmemori/crm_api/node_modules/mongoose/lib/model.js:4177:5
    at new Promise (<anonymous>)
    at Function.populate (/Users/bob/Documents/work/inmemori/crm_api/node_modules/mongoose/lib/model.js:4176:10)
    at populate (/Users/bob/Documents/work/inmemori/crm_api/node_modules/mongoose/lib/model.js:4335:13)
    at _populate (/Users/bob/Documents/work/inmemori/crm_api/node_modules/mongoose/lib/model.js:4205:5)
    at /Users/bob/Documents/work/inmemori/crm_api/node_modules/mongoose/lib/model.js:4177:5
    at new Promise (<anonymous>) {
  path: 'child.subChildA.grandChildA'
[...]

PR #15082, opened 2024-12-09
fix(query): clone PopulateOptions when setting _localModel to avoid state leaking between subpopulate instances

vkarpov15:
**Summary**

#15026 is a very tricky bug where sometimes `_localModel` populate option leaks between subpopulate instances. I wasn't able to find the exact root cause to my satisfaction, nor was I able to come up with a consistent repro that doesn't involve repeatedly running the repro script from #15026. However, the more high level explanation is that `utils.populate()` doesn't clone instances of `PopulateOptions`, so PopulateOptions instances are shared between nested populate calls. This PR makes it so that we clone populate options before setting `_localModel`.

Also, to aid in debugging, I went ahead and refactored populate internals to use async/await instead of callbacks. Should be much cleaner.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -4248,17 +4248,9 @@ Model.populate = async function populate(docs, paths) {
   const _this = this;
   // normalized paths
   paths = utils.populate(paths);
-  // data that should persist across subPopulate calls
-  const cache = {};
 
-  return new Promise((resolve, reject) => {
-    _populate(_this, docs, paths, cache, (err, res) => {
-      if (err) {
-        return reject(err);
-      }
-      resolve(res);
-    });
-  });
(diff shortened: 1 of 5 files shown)


lib/model.js
```diff
@@ -4273,25 +4265,17 @@ Model.populate = async function populate(docs, paths) {
  * @api private
  */
 
-function _populate(model, docs, paths, cache, callback) {
-  let pending = paths.length;
+async function _populate(model, docs, paths) {
```
hasezoey, line 4268: jsdoc for this function needs to be updated (old parameters are still there)

hasezoey (approved):

Issue #14971, opened 2024-10-17
Connection timeout 10000ms when upgraded mongoose from 7.4.5 to 8.7.1


sukhminderx:
### Prerequisites

- [x] I have written a descriptive issue title

### Mongoose version

8.7.1

### Node.js version

20.x

### MongoDB version

6.9

### Operating system

None

### Issue

Last night, I upgraded my mongodb and mongoose from versions 5.8.1 to 6.9 and. 7.4.5 to 8.7.1 respectively. 

I tested on my local. Everything worked fine. I promoted the changes to test env - AWS amplify - things were fine. Then, I moved the changes to production.

My db is hosted on mongo atlas serverless cluster.

On production, sometimes same queries will work fine but other times give timeout error with 10000ms. I mean same queries behaving differently.

I have been scratching my head because both test and prod env are exactly similar except that prod is connected to my domain name.
Please assist if someone has knowledge about this.

sharing my connection code and one query that fails/passes abruptly.

```
import { ServerApiVersion } from "mongodb";
import mongoose from "mongoose";

const uri = process.env.MONGODB_URI as string;


const connection: {
  isConnected?: boolean;
} = {};

const dbConnect = async () => {
  if (connection.isConnected) return;
  mongoose.set("strictQuery", false);
  const db = await mongoose.connect(uri);

  connection.isConnected = Boolean(db.connections[0].readyState);
};
export default dbConnect;
```

```
export async function GET(req: Request) {
  try {
    await apiHandler(req, "");
    const countries: any = await Countries.find({}).exec();
    return NextResponse.json(countries, { status: 200 });
  } catch (error: any) {
    console.log("wb saas: countries");
    console.log(
      typeof error !== "string"
        ? error.message
          ? error.message
          : (error || {}).toString()
        : error
    );
    return errorHandler(error);
  }
}
```

garethlowriehive:
We're also experiencing a very similar issue.

Using MongoDB Server 8.0 in Atlas and locally in Docker.
Node: 20.11.1
Mongoose 8.7.1

Everything works intermittently, but lots of buffering error across lots of different find’s and aggregates in all of our microservices.


```
[Buffering timed out after 10000ms](https://www.mongodb.com/community/forums/t/buffering-timed-out-after-10000ms-8-7-0-mongoose-node-20-10-0/299932)
```

[...]

PR #15089, opened 2024-12-10
fix(connection): remove heartbeat check in load balanced mode

### LINKED ISSUE #15042: loadBalanced=true drops connection after a delay and never reconnects
### Prerequisites

### Mongoose version

8.7.0

### Node.js version

16.20.2

### MongoDB server version

7.0

_No response_

### Description

When trying to connect to a database using `loadBalanced=true` in the connection string, mongoose drops the connection after some time but never reconnects.

This happens in all versions >=8.7.0
8.6.4 doesn't have this issue and it reconnect successfully after the delay.

### Steps to Reproduce

Providing sample code in which we manually add a delay so the connection drops and we never reconnect to MongoDB.
```
const mongoose = require("mongoose")
const { ObjectId } = require("bson")

// some connection string using `loadBalanced=true`
const connectionString = ""

const delay = ms => new Promise(res => setTimeout(res, ms));

async function execTest() {
  mongoose.connection.on('connected', () => console.log('connected'));
  mongoose.connection.on('open', () => console.log('open'));
  mongoose.connection.on('disconnected', () => console.log('disconnected'));
  mongoose.connection.on('reconnected', () => console.log('reconnected'));
  mongoose.connection.on('disconnecting', () => console.log('disconnecting'));
  mongoose.connection.on('close', () => console.log('close'));

  mongoose.connection.on('error', () => console.log('close'));

  await mongoose.connect(connectionString, { maxPoolSize: 100 })
  console.log("MongoDB Connection State:", mongoose.connection.readyState)

  const drawingSchema = mongoose.Schema({
    prop: { type: String, required: true },
  });
  const someSchema = mongoose.model("someSchema", someSchema);

  await someSchema.insertMany([
    { name: "Name1", prop: "abc" },
    { name: "Name2", prop: "def" },
    { name: "Name3", prop: "ghi" },
    { name: "Name4", prop: "jkl" },
    { name: "Name5", prop: "mno" },
  ])

  await delay(20000)

  console.log("MongoDB Connection State:", mongoose.connection.readyState) // 0

  console.log("start session")
  const session = await mongoose.startSession(); // it hangs here, since it never reconnects
  console.log("start transaction")
  session.startTransaction();
  console.log("after start transaction")

  const x = await someSchema.findOne({ name: "Drawing1" }).lean()

  session.commitTransaction()

  session.endSession()
  console.log(x)

  console.log("after update")
  return
}

execTest()
```

Setting `heartbeatFrequencyMS` to a value higher that the delay makes the issue go away but I'm not sure if this would be a recommended workaround.

### Expected Behavior

Mongoose should reconnect to the MongoDB database and the query should work

vkarpov15:
cc @alexbevi 

Fix #15042
Revert #14812

**Summary**

We've gotten a few bug reports that indicate connectivity issues with Mongoose 8.7, including #14971 etc. Most notably, we've confirmed that #15042 is caused by the heartbeat check we added in #14812: we should not use that heartbeat check if the topology is `LoadBalanced` because the MongoDB Node driver doesn't send heartbeats in that case, so for LoadBalanced topology Mongoose currently loses connectivity after 20 seconds. Given this and all the issues we've gotten specifically referencing connectivity issues in Mongoose 8.7.0, I think it is prudent to remove this check for now.

**Examples**

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -106,15 +106,6 @@ Object.setPrototypeOf(Connection.prototype, EventEmitter.prototype);
 
 Object.defineProperty(Connection.prototype, 'readyState', {
   get: function() {
-    // If connection thinks it is connected, but we haven't received a heartbeat in 2 heartbeat intervals,
-    // that likely means the connection is stale (potentially due to frozen AWS Lambda container)
-    if (
-      this._readyState === STATES.connected &&
-      this._lastHeartbeatAt != null &&
-      typeof this.client?.topology?.s?.description?.heartbeatFrequencyMS === 'number' &&
-      Date.now() - this._lastHeartbeatAt >= this.client.topology.s.description.heartbeatFrequencyMS * 2) {
-      return STATES.disconnected;
-    }
     return this._readyState;
   },
   set: function(val) {
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -106,6 +106,18 @@ Object.setPrototypeOf(Connection.prototype, EventEmitter.prototype);
 
 Object.defineProperty(Connection.prototype, 'readyState', {
   get: function() {
+    // If connection thinks it is connected, but we haven't received a heartbeat in 2 heartbeat intervals,
+    // that likely means the connection is stale (potentially due to frozen AWS Lambda container)
+    if (
+      this._readyState === STATES.connected &&
+      this._lastHeartbeatAt != null &&
+      // LoadBalanced topology (behind haproxy, including Atlas serverless instances) don't use heartbeats,
+      // so we can't use this check in that case.
+      this.client?.topology?.s?.description?.type !== 'LoadBalanced' &&
+      typeof this.client?.topology?.s?.description?.heartbeatFrequencyMS === 'number' &&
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
Because #14812 main functionality is reverted, should we re-open #14727?

vkarpov15:
Based on convo with @alexbevi I'm thinking it's better to make the following 2 changes:

1. Avoid the stale heartbeat check in LoadBalanced mode. Apparently the MongoDB Node driver doesn't receive heartbeats in LoadBalanced mode, and that's the mode the driver runs in when connected to MongoDB behind haproxy or when connected to Atlas serverless instances.
2. Don't treat successful connection as a heartbeat. Part of the problem here is that Mongoose thinks it received a heartbeat because we set _lastHeartbeatAt when Mongoose successfully connects, but then no other heartbeats are ever received. With this change, if Mongoose never received an actual successful heartbeat event, it won't use the stale heartbeat check. This change is a precaution for other edge cases where Mongoose might not receive any heartbeats for some reason.

PR #15094, opened 2024-12-13
Fix aggregate cursor als

IchirokuXVI:
**Summary**

Aggregate cursors didn't automatically add session if it was present in AsyncLocalStorage. Also the options weren't the same for exec and cursor so I added a function to make them the same just like in QueryCursor.

**Examples**

```js
mongoose.set("transactionAsyncLocalStorage", true);

mongoose.connect("mongodb://127.0.0.1:27017/cursortest");

const test = async () => {
    const TestModel = mongoose.model(
      "Test",
      new mongoose.Schema({ name: String })
    );

    mongoose.set("debug", true);
    await TestModel.aggregate([{ $match: {} }]);
    const cursor1 = TestModel.aggregate([{ $match: {} }]).cursor();
    const cursor2 = TestModel.aggregate([{ $match: {} }])
      .session(session)
      .cursor();

    await cursor1.next();
    await cursor2.next();
    mongoose.set("debug", false);

    /*
        From the debug logs:
            Mongoose: TestModel.aggregate([ { '$match': {} } ], { session: ClientSession("bb843cfeb01d4ec3ad0373d2b7d3db20") })
            Mongoose: TestModel.aggregate([ { '$match': {} } ], { cursor: {}, session: ClientSession("bb843cfeb01d4ec3ad0373d2b7d3db20")})
            Mongoose: TestModel.aggregate([ { '$match': {} } ], { session: ClientSession("bb843cfeb01d4ec3ad0373d2b7d3db20"), cursor: {}})

        Before the fix the first cursor didn't have a session but aggregate and second cursor had a session.
    */
  });
};

test()
  .then(() => console.log("Finished"))
  .catch((error) => console.error(error))
  .finally(() => process.exit(0));
```

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -95,7 +95,7 @@ Aggregate.prototype.options;
  */
 
 Aggregate.prototype._optionsForExec = function() {
-  const options = clone(this.options || {});
+  const options = this.options || {};
 
   const asyncLocalStorage = this.model()?.db?.base.transactionAsyncLocalStorage?.getStore();
   if (!options.hasOwnProperty('session') && asyncLocalStorage?.session != null) {
@@ -932,6 +932,7 @@ Aggregate.prototype.option = function(value) {
  */
 
 Aggregate.prototype.cursor = function(options) {
+  this._optionsForExec();
   this.options.cursor = options || {};
   return new AggregationCursor(this); // return this;
 };
@@ -1040,7 +1041,7 @@ Aggregate.prototype.exec = async function exec() {
   applyGlobalMaxTimeMS(this.options, model.db.options, model.base.options);
   applyGlobalDiskUse(this.options, model.db.options, model.base.options);
 
-  const options = this._optionsForExec()
+  this._optionsForExec();
 
   if (this.options && this.options.cursor) {
     return new AggregationCursor(this);
@@ -1065,6 +1066,8 @@ Aggregate.prototype.exec = async function exec() {
   if (!this._pipeline.length) {
     throw new MongooseError('Aggregate has empty pipeline');
   }
+
+  const options = clone(this.options || {});
 
   let result;
   try {
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -88,6 +88,24 @@ function Aggregate(pipeline, model) {
 Aggregate.prototype.options;
 
 /**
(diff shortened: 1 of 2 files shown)


lib/aggregate.js
```diff
@@ -87,6 +87,24 @@ function Aggregate(pipeline, model) {
...
 
+/**
+ * Returns default options for this aggregate.
+ *
+ * @param {Model} model
+ * @api private
+ */
+
+Aggregate.prototype._optionsForExec = function() {
+  const options = clone(this.options || {});
```
vkarpov15, line 98: `clone()` shouldn't be necessary here, would be ideal to avoid cloning for performance purposes

lib/cursor/aggregationCursor.js
```diff
@@ -57,21 +58,25 @@ util.inherits(AggregationCursor, Readable);
 function _init(model, c, agg) {
   if (!model.collection.buffer) {
     model.hooks.execPre('aggregate', agg, function() {
+      Object.assign(c.options, agg._optionsForExec());
```
vkarpov15, line 61: Similar to previous comments, is there any need to set `c.options` at all? Could just call `aggregate(agg._pipeline, agg.optionsForExec())`

vkarpov15 (commented):
I had a couple of minor style suggestions, but looks good overall.

lib/aggregate.js
IchirokuXVI: The previous implementation also used clone(), so I retained it to ensure no existing functionality was unintentionally affected. However, I agree that removing clone() could improve performance. If it’s deemed unnecessary, I suggest calling the function directly in the constructor to initialize the options instead.

lib/cursor/aggregationCursor.js
IchirokuXVI: In QueryCursor, options is also set to a new object, so I followed the same approach here.
That said, QueryCursor uses the options object for some checks while AggregateCursor only uses it in _init, so it might make sense to remove it from AggregateCursor for now.

IchirokuXVI:
Removed the cloning in _optionsForExec and also sest the options before creating AggregationCursor in Aggregate.cursor(). Passed all tests with npm test.

vkarpov15 (approved):

Issue #15071, opened 2024-12-03
Update from 8.8.2 to 8.8.3 causes "Uncaught TypeError: Invalid `path`. Must be either string or array"


christian-schwaderer:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.8.3

### Node.js version

20.18.0

### MongoDB server version

7

### Description

I tried upgrading Mongoose from 8.8.2 to 8.8.3. But when trying to create a document I got this issue:

```
Uncaught TypeError: Invalid `path`. Must be either string or array
      at Object.exports.get (node_modules/mpath/lib/index.js:59:11)
      at Object.exports.getValue (Role node_modules/mongoose/lib/utils.js:624:16)
      at model.Document.$__getValue (Role node_modules/mongoose/lib/document.js:1771:16)
      at getSubdocs (Role node_modules/mongoose/lib/document.js:3695:23)
      at model.Document.$getAllSubdocs (Role node_modules/mongoose/lib/document.js:3723:3)
      at _getPathsToValidate (Role node_modules/mongoose/lib/document.js:2714:25)
      at model.Document.$__validate (Role node_modules/mongoose/lib/document.js:2980:25)
      at node_modules/kareem/index.js:497:25
      at processTicksAndRejections (node:internal/process/task_queues:77:11)
```

Tried to look a little bit and it seems that the problematic line is `for (const { path } of doc.$__schema.childSchemas) {`
It seems my schema has a `paths` property but not a `path` property and hence an `undefined` value is passed down. But that's where my knowledge ends. Cannot say how come.

### Steps to Reproduce

Sorry but I cannot paste hundreds of lines of company code in here. I have no idea where in my schema the problem is.

hansgen:
We're experiencing the same issue.

maximilianschmid:
Same here

vkarpov15:
Without a repro script there's no way for us to guarantee a fix, but in #15098 we added some logic to improve this error message and a potential (but unlikely) fix. Can anyone in this thread please provide a script that demonstrates this issue or some sample code that can give us some hints?

christian-schwaderer:
Boy, that took a few hours ... Here we go [https://github.com/christian-schwaderer/mongoose-8-8-3-invalid-path-demo](https://github.com/christian-schwaderer/mongoose-8-8-3-invalid-path-demo)
Just clone and run `node index.js`

[...]

PR #15098, opened 2024-12-15
fix(document+schema): add potential fix for #15071

vkarpov15:
**Summary**

Improve error message for #15071 to help clarify what the issue might be, and also a potential fix. The only possibility I can see for childSchemas `path` property to not be a string is if the schema's `paths` for some reason includes a symbol. Unlikely, but without a repro script in #15071 this is the best we can do.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -1779,11 +1779,7 @@ Document.prototype.$__set = function(pathToMark, path, options, constructing, pa
 Document.prototype.$__getValue = function(path) {
   if (typeof path !== 'string' && !Array.isArray(path)) {
     throw new TypeError(
-      'Invalid `path`. Must be either string or array. Got "' +
-      path +
-      '" (type ' +
-      typeof path +
-      ')'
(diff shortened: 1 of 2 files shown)


lib/document.js
```diff
@@ -1777,6 +1777,15 @@ Document.prototype.$__set = function(pathToMark, path, options, constructing, pa
  */
 
 Document.prototype.$__getValue = function(path) {
+  if (typeof path !== 'string' && !Array.isArray(path)) {
+    throw new TypeError(
+      'Invalid `path`. Must be either string or array. Got "' +
+      path +
+      '" (type ' +
+      typeof path +
+      ')'
```
hasezoey, line 1786: i think it would be a lot cleaner if template strings would be used here

hasezoey (approved):

lib/document.js
vkarpov15: Fixed in 7b0dba0

PR #15103, opened 2024-12-16
types: correctly handle union types in BufferToBinary and related helpers

### LINKED ISSUE #15057: Mongoose 8.7.2 breaks Union Types

vkarpov15:
Fix #15057
Replacement for #15072

**Summary**

This looks to be a more correct approach than #15072, which had some unexpected issues that @hasezoey pointed out. @hasezoey can you please double check whether this causes issues for Typegoose? 

**Examples**

### DIFF
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -206,3 +206,34 @@ async function gh13382() {
   const res = await Test.updateOne({}, { name: 'bar' }).lean();
   expectAssignable<{ matchedCount: number, modifiedCount: number }>(res);
 }
+
+async function gh15057() {
+  type Attachment =
+    | {
+        type: 'foo';
+        value?: undefined;
+      }
+    | {
+        type: 'string';
+        value?: string;
+      };
+
+  const TestSchema = new Schema<Attachment>({
+    type: { type: String, required: true },
+    value: { type: String }
+  });
+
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
LGTM, can confirm this version passes all typegoose tests and does not require modification.

PR #15112, opened 2024-12-17
fix(schema): avoid throwing duplicate index error if index spec keys have different order or index has a custom name

### LINKED ISSUE #15109: index setting error occurred at 8.9.1("Schema already has an index")
### Prerequisites

### Issue

# Error message
```
MongooseError: Schema already has an index on {"uid":1,"gid":1}
    at Schema.index (xxx/node_modules/mongoose/lib/schema.js:2151:13)
    at Object.<anonymous> (xx/schema/member.js:20:8)
```
# Schema 
```
/* eslint-disable */

const _schema = {
    uid: {type: mongoose.Schema.Types.ObjectId, required: true},
    gid: {type: mongoose.Schema.Types.ObjectId, required: true},
};
    _schema,
    {'versionKey': false, 'usePushEach': true, 'minimize': false, 'shardKey': {'uid': 1}}
);
schema.index({'gid': 1, 'uid': 1}, {background: false});
schema.index({'uid': 1, 'gid': 1}, {background: false});
```
# mongoDB
```
test> db.test.createIndex({uid: 1, gid:1})
uid_1_gid_1
test> db.test.createIndex({uid: 1, gid:1})
uid_1_gid_1
test> db.test.getIndexes()
[
  { v: 2, key: { gid: 1, uid: 1 }, name: 'gid_1_uid_1' },
  { v: 2, key: { uid: 1, gid: 1 }, name: 'uid_1_gid_1' }
]
```

In my opinion, the two indexes seem to work differently in MongoDB. 
why is this causing an error in mongoose? It works correctly in mongoose@8.8.x, but it behaves abnormally in mongoose@8.9.1.

vkarpov15:
Fix #15109

**Summary**

#15109 points out that if index spec has keys in different order, that is still valid. So `schema.index({ a: 1, b: 1 }); schema.index({ b: 1, a: 1 })` should work. Also, if there is a custom index `name`, then that is fine too, because MongoDB enforces index uniqueness based on `name`. It's just that two indexes with the same spec will have the same `name` unless an explicit `name` option is specified in the index options.

**Examples**

### DIFF
--- a/lib/helpers/indexes/isIndexSpecEqual.js
+++ b/lib/helpers/indexes/isIndexSpecEqual.js
@@ -1,0 +1,32 @@
+'use strict';
+
+/**
+ * Compares two index specifications to determine if they are equal.
+ *
+ * #### Example:
+ *     isIndexSpecEqual({ a: 1, b: 1 }, { a: 1, b: 1 }); // true
+ *     isIndexSpecEqual({ a: 1, b: 1 }, { b: 1, a: 1 }); // false
+ *     isIndexSpecEqual({ a: 1, b: -1 }, { a: 1, b: 1 }); // false
+ *
+ * @param {Object} spec1 - The first index specification to compare.
+ * @param {Object} spec2 - The second index specification to compare.
+ * @returns {Boolean} Returns true if the index specifications are equal, otherwise returns false.
+ */
+
+module.exports = function isIndexSpecEqual(spec1, spec2) {
+  const spec1Keys = Object.keys(spec1);
+  const spec2Keys = Object.keys(spec2);
+
+  if (spec1Keys.length !== spec2Keys.length) {
+    return false;
+  }
+
+  for (let i = 0; i < spec1Keys.length; i++) {
+    const key = spec1Keys[i];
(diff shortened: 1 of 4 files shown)


lib/helpers/indexes/isIndexSpecEqual.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Compares two index specifications to determine if they are equal.
+ *
+ * #### Example:
+ *     isIndexSpecEqual({ a: 1, b: 1 }, { a: 1, b: 1 }); // true
+ *     isIndexSpecEqual({ a: 1, b: 1 }, { b: 1, a: 1 }); // false
+ *     isIndexSpecEqual({ a: 1, b: -1 }, { a: 1, b: 1 }); // false
+ *
+ * @param {Object} spec1 - The first index specification to compare.
+ * @param {Object} spec2 - The second index specification to compare.
```
hasezoey, line 12:
```suggestion
 * @param {Object} spec1 The first index specification to compare.
 * @param {Object} spec2 The second index specification to compare.
```

i dont think the `-` is necessary here.

hasezoey (approved):

Issue #15117, opened 2024-12-18
Missing changelog / release for 7.8.3


felipeplets:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

Both the [changelog](https://github.com/Automattic/mongoose/blob/master/CHANGELOG.md) and the [releases page](https://github.com/Automattic/mongoose/releases) are missing details for version `7.8.3`

vkarpov15:
Fixed changelog entry, 7.8.3 is now in master: https://github.com/Automattic/mongoose/blob/master/CHANGELOG.md#783--2024-11-26 . Will look into fixing our automated release generation to see if we can apply that to 7.x too.

felipeplets:
Great, thank you @vkarpov15

vkarpov15:
GitHub releases fix should be in https://github.com/mongoosejs/backend/commit/505942cdaf7069899ed944d0ab26af9f3b2367a4, future 6.x and 7.x releases should end up in GitHub releases

PR #15119, opened 2024-12-19
move search endpoint to netlify

vkarpov15:
**Summary**

I refactored https://github.com/mongoosejs/backend to run on Netlify functions because I was having a lot of trouble with deploying to Azure Functions. Netlify's automatic CI/CD should make it much easier to maintain the backend going forward.

I will have to make this change in other branches. However not urgent - the Azure Functions endpoint is still running and up to date, I just had to make some changes to our GitHub webhook re: #15117 so that's the only code that's out of date.

**Examples**

### DIFF
--- a/docs/js/search.js
+++ b/docs/js/search.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const root = 'https://mongoosejs.azurewebsites.net/api';
+const root = 'https://mongoose-js.netlify.app/.netlify/functions';
 
 const defaultVersion = '8.x';
 const versionFromUrl = window.location.pathname.match(/^\/docs\/(\d+\.x)/);


hasezoey (commented):
Looks good to me, though is when going directly to that endpoint(`/.netlify/functions`) supposed to show "Page not found" (though adding `/search` gives a parameter error)?
Also when trying to run this locally, it seem to not work with the error:
```txt
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://mongoose-js.netlify.app/.netlify/functions/search?search=model&version=8.x. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 200.
```

vkarpov15:
Thanks for catching that, there was a CORS issue on mongoosejs/backend that I fixed. Seems to work fine now, I can `fetch()` the search endpoint from Chrome console on mongoosejs.com as shown in the following screenshot.

image

PR #15123, opened 2024-12-19
types: make BufferToBinary avoid Document instances

### LINKED ISSUE #15122: The error caused by Binary when using lean queries still exists in `8.9.2`
### Prerequisites

### Mongoose version

8.9.2

### Node.js version

20.10.0

### MongoDB server version

6.0.2

5.5.4

### Description

In version 8.9.2 of Mongoose, it seems that the error shown in this [issue](https://github.com/Automattic/mongoose/issues/15102) (and [this one](https://github.com/Automattic/mongoose/issues/15047)) is still not resolved.

In the reproduction link the query does not return IParent fields:
Image

Instead, in my project, the last part of the error updates from:
```
Type 'Binary | null | undefined' is not assignable to type 'Buffer | undefined'.
Type 'null' is not assignable to type 'Buffer | undefined'.
```
to
```
Type 'Binary | undefined' is not assignable to type 'Buffer | undefined'.
Type 'Binary' is missing the following properties from type 'Buffer': equals, compare, copy, slice, and 93 more.
```

### Steps to Reproduce

Reproduction link [here](https://stackblitz.com/edit/vitejs-vite-twnhseek?file=src%2Findex.ts).

### Expected Behavior

_No response_

vkarpov15:
Fix #15122

**Summary**

The problem in #15122 is that BufferToBinary is drilling into Document instances, and when a user does `Document & IChild` we can't reliably infer the raw document type, so we need to skip that case.

**Examples**

### DIFF
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -308,8 +308,7 @@ async function gh15122() {
         ref: 'Child',
         required: true
       }
-    },
-    { optimisticConcurrency: true }
+    }
   );
 
   parentSchema.virtual('fullName').get(function() {
--- a/test/types/lean.test.ts
+++ b/test/types/lean.test.ts
@@ -1,4 +1,4 @@
-import { Schema, model, Types, InferSchemaType, FlattenMaps } from 'mongoose';
+import { Schema, model, Types, InferSchemaType, FlattenMaps, HydratedDocument, Model, Document, PopulatedDoc } from 'mongoose';
 import { expectAssignable, expectError, expectType } from 'tsd';
 
 function gh10345() {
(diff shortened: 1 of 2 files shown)


test/types/lean.test.ts
```diff
@@ -237,3 +237,90 @@ async function gh15057() {
...
+        required: true,
+        trim: true
+      },
+      child: {
+        type: 'ObjectId',
+        ref: 'Child',
+        required: true
+      }
+    },
+    { optimisticConcurrency: true }
```
hasezoey, line 312:
```suggestion
    }
```

i think this can be omitted as it is not relevant to types, right?

hasezoey (approved):

test/types/lean.test.ts
vkarpov15: Correct, `optimisticConcurrency: true` has no impact on this test.

PR #15127, opened 2024-12-24
types: add cleanIndexes() to IndexManager interface

vkarpov15:
**Summary**

Without this PR, TypeScript complains about `Model.cleanIndexes()` not existing

**Examples**


hasezoey (approved):
LGTM, maybe also note what the `string[]` it returns is.

PR #15132, opened 2024-12-26
types(model+query): avoid stripping out virtuals when calling populate with paths generic

### LINKED ISSUE #15111: If I explicitly set the type of a PopulatedDoc variable using populate in a query, the virtual variables of the retrieved instances become nonexistent
### Prerequisites

### Mongoose version

8.9.1

### Node.js version

20.10.0

### MongoDB server version

6.0.2

5.5.4

### Description

It seems there is a typing conflict when executing a query using populate and explicitly setting the type of the PopulatedDoc variable. It's likely that the virtual variables are being overwritten.

As you can see from the example, in the query I explicitly declare the type of the PopulatedDoc variable "child". After that, if I try to add one of the retrieved instances to an array of type `ParentInstance`, I get a typing error. The same virtual variables no longer exist, and the linter returns an error.
```ts
const filteredParents: ParentInstance[] = [];
const parents = await Parent.find().populate<{ child: ChildInstance }>(
  'child'
);
parents.forEach((parent) => {
  const condition = parent.fullName === 'fake condition'; // error
  if (condition) {
    filteredParents.push(parent); // error
  }
});
```

The full error:
```
Argument of type 'Document<unknown, {}, MergeType<IParent, { child: Document<unknown, {}, IChild> & Omit<IChild & Required<{ _id: ObjectId; }> & { __v: number; }, "id"> & IChildVirtuals; }>> & Omit<...> & { ...; } & Required<...> & { ...; }' is not assignable to parameter of type 'Document<unknown, {}, IParent> & Omit<IParent & Required<{ _id: ObjectId; }> & { __v: number; }, keyof IParentVirtuals> & IParentVirtuals'.
  Property 'fullName' is missing in type 'Document<unknown, {}, MergeType<IParent, { child: Document<unknown, {}, IChild> & Omit<IChild & Required<{ _id: ObjectId; }> & { __v: number; }, "id"> & IChildVirtuals; }>> & Omit<...> & { ...; } & Required<...> & { ...; }' but required in type 'IParentVirtuals'.(2345)
index.ts(64, 3): 'fullName' is declared here.
```

`ParentInstance` is declared like this:
```ts
type ParentInstance = HydratedDocument<
  IParent,
  ParentDocumentOverrides & IParentVirtuals
>;
```

If I remove `& IParentVirtuals` from the declaration, the error disappears.

### Steps to Reproduce

Reproduction link [here](https://stackblitz.com/edit/vitejs-vite-cfnvs3zw?file=src%2Findex.ts).

### Expected Behavior

If I explicitly set the type of a PopulatedDoc variable, I should still be able to use the virtual variables of the retrieved instances.

vkarpov15:
Fix #15111

**Summary**

`.populate<Paths>` returns a hydrated document with no virtuals, because we're not passing `TVirtuals` into queries. This PR fixes that by replacing `TInstanceMethods` with `TDocOverrides` similar to what we do for `HydratedDocument` - globbing together instance methods and virtuals into one "overrides" type that we `&` on to get the fully hydrated document.

**Examples**

### DIFF
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -1,4 +1,4 @@
-import mongoose, { Schema, model, Document, PopulatedDoc, Types, HydratedDocument, SchemaTypeOptions } from 'mongoose';
+import mongoose, { Schema, model, Document, PopulatedDoc, Types, HydratedDocument, SchemaTypeOptions, Model } from 'mongoose';
 // Use the mongodb ObjectId to make instanceof calls possible
 import { ObjectId } from 'mongodb';
 import { expectAssignable, expectError, expectType } from 'tsd';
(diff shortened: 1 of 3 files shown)


AbdelrahmanHafez (approved):

test/types/populate.test.ts
```diff
@@ -459,3 +459,100 @@ async function gh14574() {
...
+    {
+      name: {
+        type: 'String',
+        required: true,
+        trim: true
+      }
+    },
+    {
+      optimisticConcurrency: true
+    }
```
hasezoey, line 497:
```suggestion
    }
```

test/types/populate.test.ts
```diff
@@ -459,3 +459,100 @@ async function gh14574() {
...
+        required: true,
+        trim: true
+      },
+      child: {
+        type: 'ObjectId',
+        ref: 'Child',
+        required: true
+      }
+    },
+    { optimisticConcurrency: true }
```
hasezoey, line 545:
```suggestion
    }
```

hasezoey (approved):
LGTM.

Some style changes to remove unrelated things from the types.

PR #15157, opened 2025-01-03
refactor: remove dotenv and axios from dev dependencies

vkarpov15:
**Summary**

Remove 2 quick dev deps. dotenv isn't used in Mongoose, I think it was accidentally installed by the TypeScript benchmark. axios is just used for one dev-only HTTP request, figure we can easily replace that with fetch(). Technically Mongoose 8 allows node engines down to v16.20.1 and fetch is only available starting in Node 17, but the script that uses axios is dev-only so we don't need to be as strict about engine support. If this can save us a dependabot PR per month, I think that's worth it.

**Examples**


AbdelrahmanHafez (approved):
Nice!

hasezoey (approved):

PR #15162, opened 2025-01-06
feat(model): add insertOne() function to insert a single doc 

### LINKED ISSUE #14843: Support for options argument when creating a single document in Model.create(): use Model.insertOne() instead
### Prerequisites

### 🚀 Feature Proposal

Enhance the `Model.create()` method to allow passing an `options` argument when creating a single document. Currently, this is only possible when documents are passed as an array, which is inconvenient for cases where only a single document needs to be created.

### Motivation

The current implementation of the `Model.create()` method requires documents to be passed as an array when using the `options` argument, even if only a single document is being created. This approach is less intuitive and can lead to unnecessary complexity, such as having to handle an array in return when only a single document is expected. Allowing the `options` argument to be used with a single document directly would result in cleaner, more straightforward code and avoid the need to deal with array structures when they aren't necessary.

### Example

Current Implementation:
```javascript
const [document] = await model.create(
  [
    {
      name: 'example',
    },
  ],
  {
    session: ClientSession,
  },
);
```

Proposed Implementation:
```javascript
const document = await model.create(
  {
    name: 'example',
  },
  {
    session: ClientSession,
  },
);
```

vkarpov15:
**Summary**

Fix #14843 

`insertOne()` is similar to `create()`, but only supports creating a single document. `insertOne()` is useful for 2 reasons:

1. Consistency with the [MongoDB CRUD API](https://www.mongodb.com/resources/products/fundamentals/crud). Mongoose models currently support all of the MongoDB CRUD API methods _except_ `insertOne()`, which is a bit of a gotcha.
2. `create()` with `options` has proven to be confusing for users, because `create(doc, { session })` attempts to create 2 documents, the 2nd with a `session` property. I don't want to drop support for `create(doc1, doc2)` (spread args) because that would just create the opposite hard-to-diagnose issue where `doc2` gets treated as options. `insertOne()` offers a neat alternative for users who just want to create 1 doc within a transaction without `new`.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2791,7 +2791,7 @@ Model.create = async function create(doc, options) {
  *     // Create a new character within a transaction.
  *     await Character.insertOne({ name: 'Jean-Luc Picard' }, { session });
  *
- * @param {Object|Document} docs Document to insert, as a POJO or Mongoose document
+ * @param {Object|Document} doc Document to insert, as a POJO or Mongoose document
  * @param {Object} [options] Options passed down to `save()`.
  * @return {Promise}
  * @api public
--- a/lib/model.js
+++ b/lib/model.js
@@ -2775,6 +2775,49 @@ Model.create = async function create(doc, options) {
 };
 
 /**
+ * Shortcut for saving one document to the database.
+ * `MyModel.insertOne(obj, options)` is almost equivalent to `new MyModel(obj).save(options)`.
+ * The difference is that `insertOne()` checks if `obj` is already a document, and checks for discriminators.
+ *
+ * This function triggers the following middleware.
+ *
+ * - `save()`
+ *
+ * #### Example:
+ *
+ *     // Insert one new `Character` document
+ *     await Character.insertOne({ name: 'Jean-Luc Picard' });
+ *
+ *     // Create a new character within a transaction.
+ *     await Character.insertOne({ name: 'Jean-Luc Picard' }, { session });
+ *
+ * @param {Object|Document} docs Document to insert, as a POJO or Mongoose document
+ * @param {Object} [options] Options passed down to `save()`.
+ * @return {Promise}
+ * @api public
+ */
+
+Model.insertOne = async function insertOne(doc, options) {
+  _checkContext(this, 'insertOne');
+
+  const discriminatorKey = this.schema.options.discriminatorKey;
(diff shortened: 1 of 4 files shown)


lib/model.js
```diff
@@ -2774,6 +2774,49 @@ Model.create = async function create(doc, options) {
...
+ *
+ * #### Example:
+ *
+ *     // Insert one new `Character` document
+ *     await Character.insertOne({ name: 'Jean-Luc Picard' });
+ *
+ *     // Create a new character within a transaction.
+ *     await Character.insertOne({ name: 'Jean-Luc Picard' }, { session });
+ *
+ * @param {Object|Document} docs Document to insert, as a POJO or Mongoose document
```
hasezoey, line 2794:
```suggestion
 * @param {Object|Document} doc Document to insert, as a POJO or Mongoose document
```

lib/model.js
```diff
@@ -2774,6 +2774,49 @@ Model.create = async function create(doc, options) {
...
+  const discriminatorKey = this.schema.options.discriminatorKey;
+  const Model = this.discriminators && doc[discriminatorKey] != null ?
+    this.discriminators[doc[discriminatorKey]] || getDiscriminatorByValue(this.discriminators, doc[discriminatorKey]) :
+    this;
+  if (Model == null) {
+    throw new MongooseError(
+      `Discriminator "${doc[discriminatorKey]}" not found for model "${this.modelName}"`
+    );
+  }
+  let toSave = doc;
```
hasezoey, line 2812: why do an alias?

types/models.d.ts
```diff
@@ -576,6 +576,8 @@ declare module 'mongoose' {
       Array<MergeType<THydratedDocumentType, Omit<DocContents, '_id'>>>
     >;
 
+    insertOne<DocContents = AnyKeys<TRawDocType>>(doc: DocContents | TRawDocType, options?: SaveOptions): Promise<THydratedDocumentType>;
```
hasezoey, line 579: consider duplicating some of the description to here

hasezoey (approved):

lib/model.js
vkarpov15: Copy/paste. Fixed in 439ec3d

types/models.d.ts
vkarpov15: Fixed in 439ec4d

PR #15163, opened 2025-01-06
feat(document): support schematype-level transform option

### LINKED ISSUE #15084: Custom toJSON and toObject Implementation Produces Object Instead of Plain Value
### Prerequisites

### Mongoose version

7.6.3

### Node.js version

22.10

### MongoDB version

5

### Operating system

macOS

_No response_

### Issue

I'm encountering an issue with implementing custom `toJSON` and `toObject` behavior for a custom type. My goal is to produce different outputs when converting the custom type's value—specifically, I want to output a plain number or string instead of the object representation.

However, despite my efforts, the result always includes the object representation, such as: `value: CustomType { value: 1 },`.

Below is a minimal reproduction of the issue:

```js
class SchemaCustomType extends mongoose.SchemaType {
  static schemaName = 'CustomType';

  constructor(key, options) {
    super(key, options, 'CustomType');
  }

  cast(value) {
    if (value === null) return null;
    return new CustomType(value);
  }
}

class CustomType {
  constructor(value) {
    this.value = value;
  }

  toObject() {
    return this.value;
  }

  $toObject() {
    return this.value;
  }

  toJSON() {
    return String(this.value);
  }
}

mongoose.Schema.Types.CustomType = SchemaCustomType;

const Model = mongoose.model(
  'Test',
  new mongoose.Schema({
    value: { type: mongoose.Schema.Types.CustomType },
  }),
);

const doc = new Model();
doc.value = 1;

console.log(doc.toJSON()); // expected { value: '1' }
console.log(doc.toObject()); // expected { value: 1 }
```

I would greatly appreciate any guidance on resolving this issue or understanding why the current implementation doesn't work as expected.

vkarpov15:
Fix #15084

**Summary**

With a little bit of work, we can support `transform` on a per-schematype basis, which can enable us to control how types end up being serialized to JSON. This is useful for custom schema types, as specified in #15084, also useful for controlling how buffers get serialized.

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -14225,6 +14225,47 @@ describe('document', function() {
     assert.strictEqual(duplicateKeyError.message, 'Email must be unique');
     assert.strictEqual(duplicateKeyError.cause.code, 11000);
   });
+
+  it('supports global transforms per schematype (gh-15084)', async function () {
+    class SchemaCustomType extends mongoose.SchemaType {
+      static schemaName = 'CustomType';
+
+      constructor(key, options) {
+        super(key, options, 'CustomType');
+      }
+
+      cast(value) {
+        if (value === null) return null;
+        return new CustomType(value);
+      }
+    }
+
+    class CustomType {
+      constructor(value) {
+        this.value = value;
+      }
+    }
+
+    mongoose.Schema.Types.CustomType = SchemaCustomType;
+
+    const Model = db.model(
+      'Test',
+      new mongoose.Schema({
+        value: { type: mongoose.Schema.Types.CustomType },
+      }),
+    );
+
+    const _id = new mongoose.Types.ObjectId('0'.repeat(24));
+    const doc = new Model({ _id });
+    doc.value = 1;
+
+    mongoose.Schema.Types.CustomType.set('transform', v => v == null ? v : v.value);
+
+    assert.deepStrictEqual(doc.toJSON(), { _id, value: 1 });
+    assert.deepStrictEqual(doc.toObject(), { _id, value: 1 });
+
+    delete mongoose.Schema.Types.CustomType;
+  });
 });
 
 describe('Check if instance function that is supplied in schema option is available', function() {
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -14226,10 +14226,8 @@ describe('document', function() {
     assert.strictEqual(duplicateKeyError.cause.code, 11000);
   });
 
-  it('supports global transforms per schematype (gh-15084)', async function () {
+  it('supports global transforms per schematype (gh-15084)', async function() {
     class SchemaCustomType extends mongoose.SchemaType {
-      static schemaName = 'CustomType';
-
       constructor(key, options) {
         super(key, options, 'CustomType');
       }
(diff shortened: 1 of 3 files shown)


hasezoey (approved):

test/document.test.js
```diff
@@ -14225,6 +14225,47 @@ describe('document', function() {
     assert.strictEqual(duplicateKeyError.message, 'Email must be unique');
     assert.strictEqual(duplicateKeyError.cause.code, 11000);
   });
+
+  it('supports global transforms per schematype (gh-15084)', async function () {
+    class SchemaCustomType extends mongoose.SchemaType {
+      static schemaName = 'CustomType';
```
hasezoey, line 14231: It looks like eslint needs a new option set: [`ecmaVersion: 2022`](https://stackoverflow.com/questions/60046847/eslint-does-not-allow-static-class-properties), this should be possible as the [target node (16.20), supports it](https://node.green/#ES2022).
Or alternatively use the old syntax to define statics.

test/document.test.js
```diff
@@ -14225,6 +14225,47 @@ describe('document', function() {
...
+      }
+    }
+
+    mongoose.Schema.Types.CustomType = SchemaCustomType;
+
+    const Model = db.model(
+      'Test',
+      new mongoose.Schema({
+        value: { type: mongoose.Schema.Types.CustomType },
+      }),
```
hasezoey, line 14255:
```suggestion
      })
```

test/document.test.js
```diff
@@ -14225,6 +14225,47 @@ describe('document', function() {
     assert.strictEqual(duplicateKeyError.message, 'Email must be unique');
     assert.strictEqual(duplicateKeyError.cause.code, 11000);
   });
+
+  it('supports global transforms per schematype (gh-15084)', async function () {
```
hasezoey, line 14229:
```suggestion
  it('supports global transforms per schematype (gh-15084)', async function() {
```

test/document.test.js
```diff
@@ -14225,6 +14225,47 @@ describe('document', function() {
...
+        this.value = value;
+      }
+    }
+
+    mongoose.Schema.Types.CustomType = SchemaCustomType;
+
+    const Model = db.model(
+      'Test',
+      new mongoose.Schema({
+        value: { type: mongoose.Schema.Types.CustomType },
```
hasezoey, line 14254:
```suggestion
        value: { type: mongoose.Schema.Types.CustomType }
```

hasezoey (commented):
Eslint fixes (when fixing the unsupported syntax)

test/document.test.js
vkarpov15: Fixed in [08113ef](https://github.com/Automattic/mongoose/pull/15163/commits/08113efa08ccec25265523a92a8075a48aabf057)

test/document.test.js
vkarpov15: Fixed in [08113ef](https://github.com/Automattic/mongoose/pull/15163/commits/08113efa08ccec25265523a92a8075a48aabf057)

test/document.test.js
vkarpov15: Fixed in [08113ef](https://github.com/Automattic/mongoose/pull/15163/commits/08113efa08ccec25265523a92a8075a48aabf057)

test/document.test.js
vkarpov15: In the interim I removed the offending code in [08113ef](https://github.com/Automattic/mongoose/pull/15163/commits/08113efa08ccec25265523a92a8075a48aabf057)

PR #15165, opened 2025-01-08
Set ESLint ECMAScript version to `2022`

hasezoey:
**Summary**

This PR sets ESLint's parser's ECMAScript version to `2022`, which is a version the targeted nodejs (16.20) supports, and could cause problems like in #15163


vkarpov15 (approved):
Thanks :+1:

Issue #15075, opened 2024-12-04
Mongoose 8.x: Discriminator schema validation breaks in static validation


skrtheboss:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.8.3 (also observed in 8.0.0)

### Node.js version

20.17.0

### MongoDB server version

7.0.15

### Description

There are significant inconsistencies in validation behavior between instance validation (`new Model().validate()`) and static validation (`Model.validate()`) when using discriminator schemas with nested document arrays:

#### Instance Validation
- Circle: Passes validation ✓
- Square: Fails validation as expected ✓

#### Static Validation
- Circle: Fails validation with "radius is required" ✗
- Square: Passes validation unexpectedly ✗

### Regression?
This issue appears to have changed in behavior between Mongoose versions:
- Mongoose 6.13.5: Both instance and static validation work consistently
- Mongoose 7.8.3: Both instance and static validation work consistently
- Mongoose 8.0.0: Inconsistent validation behavior emerges
- Mongoose 8.8.3: Inconsistent validation behavior persists

### Steps to Reproduce

```javascript
const mongoose = require('mongoose');

// Interfaces (conceptual, for TypeScript-like clarity)
// interface Circle { kind: 'Circle'; radius: number; }
// interface PropertyPath { property: string; path: string; }
// interface Square { kind: 'Square'; propertyPaths: PropertyPath[]; }

async function runValidationTest() {
    // Create the base shape schema
    const shapeSchema = new mongoose.Schema({ name: String }, {
        discriminatorKey: 'kind',
        _id: false
    });

    // Main schema with shape array
    const schema = new mongoose.Schema({
        shape: [shapeSchema]
    });

    // Circle discriminator
    schema
        .path('shape')
        .discriminator('Circle', new mongoose.Schema({
            radius: {
                type: mongoose.Schema.Types.Number,
                required: true
            }
        }, { _id: false }));

    // PropertyPath schema for Square
    const propertyPathSchema = new mongoose.Schema({
        property: {
            type: mongoose.Schema.Types.String,
            required: true,
        },
        path: {
            type: mongoose.Schema.Types.String,
            required: true,
        }
    }, { _id: false });

    // Square discriminator
    schema
        .path('shape')
        .discriminator(
            'Square',
            new mongoose.Schema({
                propertyPaths: {
                    type: [propertyPathSchema],
                    required: true
                }
[...]

PR #15169, opened 2025-01-08
fix(model): make Model.validate() static correctly cast document arrays

### LINKED ISSUE #15164: Mongoose 8.x: Required array schema validation breaks in static validation
### Prerequisites

### Mongoose version

8.9.3

### Node.js version

20.17.0

### MongoDB server version

7.0.14

5.5.4

### Description

When using a schema with a required array field, static validation using `Model.validate()` fails when the array is empty, even though instance validation using `new Model().validate()` passes.

### Steps to Reproduce

Run the following code with any 8.x.x version and it will fail the static validation:

```javascript

async function runValidationTest() {
    const barSchema = new mongoose.Schema({
        foo: {
            required: true,
        },
    });

    const fooSchema = new mongoose.Schema({
        bars: {
            type: [barSchema],
            required: true,
        },
    });

    const Model = mongoose.model('FooTest', fooSchema);

    const fooInstance = {
        bars: [],
    };

    console.log('--- Instance Validation (new Model().validate()) ---');
    try {
        await new Model(fooInstance).validate();
        console.log('Instance validation: PASSED ✓');
    } catch (error) {
        console.error('Instance validation: FAILED ✗', error);
    }

    console.log('\n--- Static Validation (Model.validate()) ---');
    try {
        await Model.validate(fooInstance);
        console.log('Static validation: PASSED ✓');
    } catch (error) {
        console.error('Static validation: FAILED ✗', error);
    }
}

// Run the test
runValidationTest().catch(console.error);

```

### Expected Behavior

Both instance validation (`new Model().validate()`) and static validation (`Model.validate()`) should pass when the bars array is empty because the array exists and satisfies the required constraint.

vkarpov15:
Fix #15164

**Summary**

#15075 fix unintentionally made it so that `Model.validate()` skips document arrays if the value being casted is an array. This PR fixes that.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3667,6 +3667,7 @@ Model.castObject = function castObject(obj, options) {
     const val = get(obj, path);
     pushNestedArrayPaths(paths, val, path);
   }
+  console.log('AB', obj, paths);
 
   let error = null;
 
@@ -3704,8 +3705,9 @@ Model.castObject = function castObject(obj, options) {
             Model.castObject.call(schemaType.caster, val)
           ];
         }
+
+        continue;
       }
-      continue;
     }
     if (schemaType.$isSingleNested || schemaType.$isMongooseDocumentArrayElement) {
       try {
--- a/lib/model.js
+++ b/lib/model.js
@@ -3667,7 +3667,6 @@ Model.castObject = function castObject(obj, options) {
     const val = get(obj, path);
     pushNestedArrayPaths(paths, val, path);
   }
-  console.log('AB', obj, paths);
 
   let error = null;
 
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -3667,6 +3667,7 @@ Model.castObject = function castObject(obj, options) {
     const val = get(obj, path);
     pushNestedArrayPaths(paths, val, path);
   }
+  console.log('AB', obj, paths);
```
hasezoey, line 3670:
```suggestion
```
Debug output?

hasezoey (approved):

PR #15172, opened 2025-01-09
Update lint dev-dependencies

hasezoey:
**Summary**

This PR updates some lint dev-dependencies to their latest version. ESLint 8.x is EOL, so we might as well use the latest version.
Should we upgrade ESLint to 9.x within mongoose 8.x (which requires node 18+), as we only run lint in higher node environments and also already have other things that require higher node versions?

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,8 @@
   "devDependencies": {
     "@babel/core": "7.26.0",
     "@babel/preset-env": "7.26.0",
-    "@typescript-eslint/eslint-plugin": "^8.18.0",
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

I have been looking to upgrade to ESLint 9, I think it is fine for Mongoose 8 to use that even with the Node 18+ requirement, because lint success/failure shouldn't depend on Node version. However, ESLint 9 has completely new configuration format, which has made upgrading tricky in my experience, so I haven't started that yet.

Issue #5714, opened 2017-10-15
connections with and without useMongoClient yield different NativeConnection objects 


Spown:
Mongoose: 4.12.2
OS: Win
Node 7.9.0
MongoDB: 3.2

``NativeConnection`` object from ``onConnected`` event, that was created with and without ``useMongoClient`` option has a slightly different structure. In particular with ``useMongoClient = true`` it has a ``.then()`` method and other promise-related fields, whereas otherwise it doesn't.

It becomes problematic in my case, since I pass it through a Q's promise. Q tries to "intelligently" (which means dumb as a brick, but anyway) resolve the whole promise chain, if it detects one and it does so by testing the resolved value and every chain-descendant value against ``isPromiseLike`` util, which in turn returns true if the said value has the ``.then()`` method.

Now, you can probably see the problem here already. If I depend my app flow on the first promise's triggering - the app would stall if the promise is resolved with such a ``NativeConnection`` object:

```js
var Q = require('q'),
        bootMongo = Q.defer(),
    mongoose = require('mongoose'),
        connOpts = {},
    GITHUB_ISSUE = `gh-5714`
;
mongoose.Promise = require('q').Promise;
connOpts.useMongoClient = true; // toggle for different output
mongoose.connect(`mongodb://localhost:27017/${GITHUB_ISSUE}`, connOpts)
mongoose.connection
.on('error', function(er) {
    bootMongo.reject(new Error(`failed to connect to DB! ${er}`));
})
.on('connected', function() {
    console.log( 'mongooseConnection connected' );
    // for useMongoClient = true ``this`` is a NativeConnection object with .then() method
    // without useMongoClient ``this`` is a NativeConnection object without .then() method
    bootMongo.resolve(this);
})
;
bootMongo.promise
.then(function (conn) {
    // will trigger without useMongoClient = true;
})
.catch(function (err) {
    // will trigger with useMongoClient = true; "promise.then is not a function"
});
```

Technically it is Q who is causing the problem, but Q was doing this in such manner all along and it is also the expected behavior for it. Mongoose however is the one who introduced such game-breaking change.

I can easily fix my problem by wrapping ``this`` from the event callback with another object and pass it into the promise. But I think you still have a problem of inconsistency...

Thank you.

vkarpov15:
Thanks for the detailed repro, will investigate ASAP

varunjayaraman:
[...]

PR #15173, opened 2025-01-09
test: remove testing for "q"

hasezoey:
**Summary**

The dependency `q` has been a long time deprecated and superseded by native JS Promise's and we only use it for 1 test.
Additionally, the way `q` was used in the tests can now be done natively by awaiting the connection.


vkarpov15 (approved):
I agree, the original issue #5714 wasn't even specific to Q, it was the same issue as #8810 of how to make `createConnection()` play nicely with async/await without making connections themselves thenable.

One fewer dev dep is always welcome :+1:

Issue #15178, opened 2025-01-13
createConnection not found when using mongoose in Deno


HerrTuring:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.9.4

### Node.js version

Deno 2.1.4

### MongoDB server version

4.0

### Typescript version (if applicable)

Deno 2.1.4

### Description

I have been using mongoose in Deno with relative success, but then I needed to change from "connect" to "createConnection" to allow for multiple databases. While the code seems to work, it doesn't seem to have the correct type declarations resulting in the IDE error:

```
Property 'createConnection' does not exist on type 'typeof import("mongoose")'.deno-ts(2339)
```

This does not happen using the connect function.

### Steps to Reproduce

Just import and execute this function:

```
import mongoose from "mongoose";

export const createMongoDBConnection = ( uri: string, db: string ) => mongoose.createConnection(uri + "/" + db);
```

Also tried with named import with similar result.

### Expected Behavior

Correct integration with IDE.

HerrTuring:
> Does this error happen when you compile your project in TypeScript or run the project, or does it just show up in your IDE? If so, what IDE are you using?

In this case there is no compile, since Deno runs typescript natively. And the code does run without issues.

I am using Visual Studio Code and the itellisense is being checked by the Deno's plugin which is properly configured since it both runs well with other codes and it does find other mongoose methods and exports.

I should also mention that when I click Control + Mouse on the mongoose import it does open a file with type declarations from mongoose and I failed to find any mention of "createConnection" in that file.

HerrTuring:
And I should add that I can import mongoose normally inside deno because I added this alias to my imports in deno.json:
```
"mongoose": "npm:mongoose@^8.9.4"
```

vkarpov15:
Try restarting your TypeScript server? It is very strange that `mongoose.connect()` shows up but `mongoose.createConnection()` does not since they are [defined right next to each other](https://github.com/Automattic/mongoose/blob/5af0a109fcf206eb765c02e0fffe54a78ab95174/types/connection.d.ts#L11-L16) in a file that's referenced [here](https://github.com/Automattic/mongoose/blob/5af0a109fcf206eb765c02e0fffe54a78ab95174/types/index.d.ts#L4).

I put in PR #15181 that may help, but shouldn't be necessary AFAIK

HerrTuring:
[...]

PR #15181, opened 2025-01-14
types(connection): explicitly export connect(), createConnection()

vkarpov15:
Re: #15178

**Summary**

Shouldn't be necessary to export these functions because they should be in global scope, but it shouldn't cause any problems and being explicit may help IDEs pick these up better.

**Examples**


hasezoey (approved):
LGTM. To my knowledge, if something is in a `declare module` block, it shouldnt matter if it has `export` or not and still be accessible.

vkarpov15:
Closing since this is unnecessary

PR #15183, opened 2025-01-15
chore: use Deno 2.1 in tests

vkarpov15:
**Summary**

We're using an out-of-date version of Deno, should test against latest

**Examples**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -108,7 +108,7 @@ jobs:
       - name: Setup Deno
         uses: denoland/setup-deno@v2
         with:
-          deno-version: v1.37.x
+          deno-version: v2.1.x
       - run: deno --version
       - run: npm install
       - name: Run Deno tests


hasezoey (commented):
LGTM, though it seems deno now requires "implicit js" files to have package.json `"type": "commonjs"` as seen in the failing tests:

```txt
     info: Deno supports CommonJS modules in .cjs files, or when the closest
          package.json has a "type": "commonjs" option.
    hint: Rewrite this module to ESM,
          or change the file extension to .cjs,
          or add package.json next to the file with "type": "commonjs" option,
          or pass --unstable-detect-cjs flag to detect CommonJS when loading.
    docs: https://docs.deno.com/go/commonjs
```

This *should* be a non-breaking change as (at least in node) a absense of the `type` field should imply `commonjs`.

vkarpov15:
We should be safe to add `"type": "commonjs"` because of [Node docs here](https://nodejs.org/api/packages.html#type):

image

In the interest of caution, I'll move this back to 8.10.

test/deno.js
```diff
@@ -16,8 +16,6 @@ const fixtures = require('./mocha-fixtures.js')
 await fixtures.mochaGlobalSetup();
 
 const child_args = [
-  // args is required to be set manually, because there is currently no way to get all arguments from deno
-  '--allow-env', '--allow-read', '--allow-net', '--allow-run', '--allow-sys', '--allow-write',
```
hasezoey, line 20: is this somehow not required anymore?

hasezoey (approved):

test/deno.js
vkarpov15: Yeah this looks to cause issues in Deno 2.x

PR #15199, opened 2025-01-23
fix(model): disallow `updateMany(update)` and fix TypeScript types re: `updateMany()`

### LINKED ISSUE #15190: Calling 'updateMany' without 'update' param leads to entire collection updated by the 'filter' param's data
### Prerequisites

### Mongoose version

6.11.3

### Node.js version

15.14

### MongoDB server version

4.4.10

4.3.5

### Description

A severe, undocumented behavior: `Model.updateMany` is switching between `filter` and `update` params if `update` is missing.

The result is unpredicted and very destructive as mongoose updates **ALL** records of the collection with the filter parameter.

The bug is clearly stated (yet undocumented) in file `query.js` (line 4038) 
```javascript
    doc = conditions;
    conditions = undefined;
```

Important to note that MongoDB **rejects** calls to `updateMany` without an `update` parameter .

This is only a mongoose undocumented feature, thus a (severe) bug as it updates the database without discretion.

### Steps to Reproduce

For a model named `PersonModel`, which represents a collection (`Person`) containing fields like `color` (a string)

Call this method:
```javascript
PersonModel.updateMany({ color : 'red' }); // should not compile + throw runtime error
```

**Expected Result:**
A runtime error should be thrown stating that 'update' must have a value (exactly as MongoDB rejects this issue)

**Actual Result:**
**All** records in collection `Person` are updated with their `color` set to "red"

### Expected Behavior

Mongoose should

1. Throw an error if this `updateMany` is called an undefined `update` paramter
2. Remove the optional mark (?) from parameter `update` in `updateMany`
3. document this behavior
4. It should never ever update the entire collection with the filter's data (!!)

vkarpov15:
Fix #15190

**Summary**

#15190 points out that `Model.updateMany(obj)` updates all documents using `obj`, which is risky, especially given that our TypeScript types as written imply that `update` parameter is optional, meaning `updateMany(obj)` treats `obj` as `filter` rather than `update`, but the runtime behavior is different.

This PR specifically blocks `Model.updateMany(obj, null)`, while allowing `Model.updateOne(obj)` and `Query.prototype.updateMany(obj, null)`.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3878,6 +3878,10 @@ Model.hydrate = function(obj, projection, options) {
  *     res.upsertedId; // null or an id containing a document that had to be upserted.
  *     res.upsertedCount; // Number indicating how many documents had to be upserted. Will either be 0 or 1.
  *
+ *     // Other supported syntaxes
+ *     await Person.find({ name: /Stark$/ }).updateMany({ isDeleted: true }); // Using chaining syntax
+ *     await Person.find().updateMany({ isDeleted: true }); // Set `isDeleted` on _all_ Person documents
+ *
  * This function triggers the following middleware.
  *
  * - `updateMany()`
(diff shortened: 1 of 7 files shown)


hasezoey (approved):
LGTM, though i think this will require changes for some users.

vkarpov15:
That is possible, however hopefully unlikely. The `updateMany(doc)` code path isn't in our current docs or tests. And this doesn't affect the chaining code path, just top-level `updateMany()`.

Issue #15188, opened 2025-01-19
$geoNear: Misleading error message when coordinates array is invalid


jl33-ai:
### Mongoose version
^8.0.0

### Description

The `$geoNear` aggregation stage produces a misleading error message when performing [GeoJSON](https://mongoosejs.com/docs/geojson.html) point queries with invalid coordinates. When the coordinates array in the near parameter is `undefined` or invalid (e.g. `[]`), instead of indicating the coordinate validation issue, it incorrectly suggests there's a problem with having multiple arguments:

```
MongoServerError: geo near accepts just one argument when querying for a GeoJSON point. Extra field found: $maxDistance: 1337.0
```

This error message is confusing because:

1. The `maxDistance` parameter is valid and supported by `$geoNear`
2. The actual issue is the invalid coordinates array in the query

### Steps to Reproduce
```javascript
import mongoose from 'mongoose';
import {MongoMemoryServer} from 'mongodb-memory-server';


const ExampleSchema = new mongoose.Schema({});
ExampleSchema.index({location: '2dsphere'});

const User = mongoose.model('User', ExampleSchema);

async function main() {
    let mongod;
    try {
        mongod = await MongoMemoryServer.create(); // same issue occurs with real connection
        const mongoUri = mongod.getUri();

        await mongoose.connect(mongoUri);

        // create a document
        await new User({
            location: {
                type: 'Point',
                coordinates: [0, 0]
            }
        }).save();

        await User.aggregate([
            {
                $geoNear: {
                    near: {
                        type: 'Point',
                        coordinates: [],
                    },
                    distanceField: 'distance',
                    spherical: true,
                    maxDistance: 1337,
                    key: 'location',
                },
            }
        ]);
    } catch (error) {
        console.error(error.stack);
    } finally {
        await mongoose.disconnect();
        if (mongod) {
            await mongod.stop();
        }
    }
}

main();
// will throw the following:
// MongoServerError: geo near accepts just one argument when querying for a GeoJSON point. Extra field found: $maxDistance: 1337.0

```
Copy and paste the code into any project with Typescript and mongoose 8.x, `npm i` and then run the script with Node. 

Note: The provided snippet uses an in-memory instance for ease of reproduction, but the behaviour is identical with a proper connection.

### Expected Behaviour

The error message should clearly indicate that the coordinates array is invalid instead of suggesting the removal of a valid parameter (`maxDistance`).

vkarpov15:
[...]

PR #15206, opened 2025-01-27
fix(aggregate): improve error when calling `near()` with invalid coordinates

vkarpov15:
**Summary**

#15188 points out a not-helpful error message from the MongoDB server. We can improve this message in the case where the user calls `near()`.

**Examples**

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -392,6 +392,7 @@ Aggregate.prototype.project = function(arg) {
  * @memberOf Aggregate
  * @instance
  * @param {Object} arg
+ * @param {Object|Array<Number>} arg.near GeoJSON point or coordinates array
  * @return {Aggregate}
  * @api public
  */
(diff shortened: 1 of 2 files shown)


lib/aggregate.js
```diff
@@ -397,6 +397,17 @@ Aggregate.prototype.project = function(arg) {
  */
 
 Aggregate.prototype.near = function(arg) {
+  if (arg == null) {
+    throw new MongooseError('Aggregate `near()` must be called with non-nullish argument');
+  }
+  if (arg.near == null) {
```
hasezoey, line 403: Maybe while you are at, document those properties in the jsdoc, or is the link to the mongodb docs enough?

hasezoey (approved):

lib/aggregate.js
vkarpov15: I added a note in jsdoc, since Mongoose is doing validation we should explain the validation step in the jsdoc.

PR #15214, opened 2025-01-30
8.10

vkarpov15:
**Summary**

**Examples**


hasezoey (approved):
LGTM, except the new Encryption tests.
Also this requires a minor change to exact type checks in typegoose (because of `Default__v` for `toObject` & `toJSON`).

PR #15216, opened 2025-01-30
fix: pin drivers-evergreen-tools version

aditi-khare-mongoDB:
**Summary**
The encryption tests are failing due to some failures on drivers-evergreen-tools. For now, we can pin the last stable commit.

**Examples**


vkarpov15 (approved):

vkarpov15:
@aditi-khare-mongoDB still having the same issue

PR #15229, opened 2025-02-03
feat(connection): make connection helpers respect bufferTimeoutMS

### LINKED ISSUE #15201: Model init can hang
### Prerequisites

### Mongoose version

8.9.5

### Node.js version

20.14.0

### MongoDB server version

8.0.1

_No response_

### Description

It is possible for model.init to fail to complete. This happens if the heartbeat expires when making the call; the code hangs inside "Connection.prototype.createCollection" - internally this function never returns from the call to "_waitForConnect".

### Steps to Reproduce

Run this self contained script - breakpoint the "init" line for long enough to expire the heartbeat.

```

async function testMongoose() {
    const db = await mongoose.connect('mongodb://127.0.0.1/testMongoose', {heartbeatFrequencyMS: 2500});
    let firstTime = true;

    // Wait for valid heartbeat to be registered.
    db.connection.db.client.on('serverHeartbeatSucceeded', () => {
        console.log(`heartbeat - readyState ${db.connection.readyState}`);

        if (firstTime) {
            firstTime = false;
            const ObjectId = Schema.ObjectId;

            const blogSchema = new Schema({author: ObjectId, title: String, body: String, date: Date});
            const blogModel = mongoose.model('blog', blogSchema);
            // Breakpoint here for long enough for the heartbeat to expire (2.5s in this case).
            blogModel.init().then(() => {
                console.log('initialised');
            }).catch((err) => {
                console.error(err);
            });
        }
    });
}

testMongoose().then(() => {});
```

I found it useful to update this mongoose function to show that the readyState becomes 0 and that causes the "end _waitForConnect" to never happen.

```
Connection.prototype.createCollection = async function createCollection(collection, options) {
  if (typeof options === 'function' || (arguments.length >= 3 && typeof arguments[2] === 'function')) {
    throw new MongooseError('Connection.prototype.createCollection() no longer accepts a callback');
  }

  console.log(`_waitForConnect ${this.readyState}`);
  await this._waitForConnect();
  console.log(`end _waitForConnect ${this.readyState}`);

  return this.db.createCollection(collection, options);
};
```

### Expected Behavior

model.init should never hang. From the test code above you can see that the heartbeat signal correctly continues to be emitted, and even the connection readyState gets updated, but _waitForConnect never returns.

console output when fails:

```
heartbeat - readyState 1
_waitForConnect 0
heartbeat - readyState 1
heartbeat - readyState 1
heartbeat - readyState 1
heartbeat - readyState 1
```

console output when succeeds:

```
heartbeat - readyState 1
_waitForConnect 1
end _waitForConnect 1
initialised
heartbeat - readyState 1
```

vkarpov15:
Fix #15201

**Summary**

#15201 pointed out an issue where we need to clear a connection's `_lastHeartbeatAt` when disconnecting, otherwise reconnecting after disconnecting can lead to the connection falsely thinking it is disconnected. This PR makes Mongoose set `_lastHeartbeatAt = null` when disconnecting _and_ when connecting, to ensure we can rely on `_lastHeartbeatAt` as an accurate representation of the last time we received a `serverHeartbeatSucceeded` on this connection.

As some followup work, I also made it so that `_waitForConnect()` respects `bufferTimeoutMS`. Right now `bufferTimeoutMS` applies to collection operations (`find()`, etc.) but not connection operations (`listCollections()`, `createCollection()`, etc.), which is very inconsistent because this means connection operations can hang forever. 

**Examples**


test/connection.test.js
```diff
@@ -1786,4 +1786,30 @@ describe('connections:', function() {
...
+  it('connection helpers buffering times out', async function() {
+    const m = new mongoose.Mongoose();
+    m.set('bufferTimeoutMS', 100);
+
+    // Delayed `assert.rejects()` pattern, because using the usual `await assert.rejects()` pattern _after_ the `setTimeout()`
+    // call below would lead to an unhandled promise rejection.
+    const promise = assert.rejects(m.connection.listCollections(), /Connection operation buffering timed out after 100ms/);
+
+    await new Promise(resolve => setTimeout(resolve, 200));
+    await promise;
```
hasezoey, line 1813: Couldnt the promise itself be awaited here until it times-out without the `setTimeout` in-between?

hasezoey (approved):

PR #15230, opened 2025-02-03
fix(bigint): throw error when casting BigInt that's outside of the bounds of what MongoDB can safely store

### LINKED ISSUE #15200: BigInt/Int64 negative value parse error
### Prerequisites

### Mongoose version

8.9.5

### Node.js version

22

### MongoDB server version

8.0.4

_No response_

### Description

BigInt field with negative value parse error in latest mongoose(8.9.5), bug works well in 7.8.4.

### Steps to Reproduce

Image

Model
```js
const UserRewardSchema = new Schema({
  chainId: {
    type: Number,
    required: true,
    immutable: true,
  },
  account: {
    type: String,
    required: true,
    index: true,
    immutable: true,
  },
  value: {
    type: BigInt,
    required: true,
    default: 0n,
  },
}, {
  versionKey: false,
  timestamps: true,
  collection: 'userRewards',
})
UserRewardSchema.index({ chainId: 1, account: 1 }, { unique: true })

export const UserReward = model('UserReward', UserRewardSchema)
```

Image

```js
import { connect } from "mongoose"
import { DB_URI } from "#common/env.js"

/**
 * @type {import('mongoose').Mongoose}
 */
let mongoose

export const connectDb = async () => {
  if (!mongoose) {
    mongoose = await connect(DB_URI, { useBigInt64: true })
    mongoose.set("transactionAsyncLocalStorage", true).set('strictQuery', true)
  }

  return mongoose
}
```

```js
const items = await UserReward.find({ account: 'xxxxx' })
    .select('-_id')
    .lean()

console.log(items)
```

### Expected Behavior

Image

vkarpov15:
Fix #15200

**Summary**

MongoDB stores BigInts as Longs under the hood, so you may lose precision if storing a BigInt that's too large in MongoDB. This PR makes Mongoose throw an error in that case, to avoid accidentally overwriting valid data.

Without `useBigInt64`, MongoDB can only store the range of a signed 64 bit integer, which is what this check uses. With `useBigInt64`, you can actually store up to the max _unsigned_ 64 bit integer (18446744073709551615), but that leads to some unexpected quirks with querying. For example, the following `$gt` query doesn't return any docs even though `18446744073709551615n > 9223372036854775807n`.

```javascript
'use strict';

mongoose.set('debug', true);

(async function() {
  await mongoose.connect('mongodb://127.0.0.1:27017/mongoose_test', { useBigInt64: true });

    myBigInt: BigInt
  });
  await Test.deleteMany({});

  const { _id } = await Test.create({ myBigInt: 18446744073709551614n });
  const doc = await Test.findById(_id);
  console.log('From db', doc);
  //doc.markModified('myBigInt');
  doc.myBigInt = 18446744073709551615n;
  await doc.save();

  console.log(await Test.findById(_id));
  console.log(await Test.findOne({ myBigInt: { $gt: 9223372036854775807n } })); // null
})();


```

Although this is more of a bug fix than anything, I think postponing it to a smaller minor release is prudent just in case.

**Examples**

### DIFF
--- a/lib/cast/bigint.js
+++ b/lib/cast/bigint.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const assert = require('assert');
 const { Long } = require('bson');
 
 /**
@@ -21,6 +20,9 @@ module.exports = function castBigInt(val) {
     return null;
   }
   if (typeof val === 'bigint') {
+    if (val > 9223372036854775807n || val < -9223372036854775808n) {
+      throw new Error('Mongoose only supports BigInts between -9223372036854775808n and 9223372036854775807n because MongoDB does not support arbitrary precision integers');
+    }
     return val;
   }
 
@@ -29,8 +31,12 @@ module.exports = function castBigInt(val) {
   }
 
   if (typeof val === 'string' || typeof val === 'number') {
-    return BigInt(val);
+    val = BigInt(val);
+    if (val > 9223372036854775807n || val < -9223372036854775808n) {
+      throw new Error('Mongoose only supports BigInts between -9223372036854775808n and 9223372036854775807n because MongoDB does not support arbitrary precision integers');
+    }
+    return val;
   }
 
-  assert.ok(false);
+  throw new Error(`Cannot convert value to BigInt: "${val}"`);
 };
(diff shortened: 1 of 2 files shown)


hasezoey (approved):
Looks good to me, though tests for this new validation error would be great.
Also consider using constants (or at least add a comment) to explain those "magic numbers" at a glance.

PR #15245, opened 2025-02-08
fix(connection): unref the "_waitForConnect"

### LINKED ISSUE #15241: Jest detect open handles in tests after upgrading from 8.9.6 to 8.10.0
### Prerequisites

### Mongoose version

8.10.0

### Node.js version

20

### MongoDB server version

6.0.20

_No response_

### Description

Jest detects open handles after upgrading from `8.9.6` to `8.10.0`.
Here are some logs:

```
Jest has detected the following 14 open handles potentially keeping Jest from exiting:

...

at node_modules/mongoose/lib/connection.js:838:19
at NativeConnection._waitForConnect (node_modules/mongoose/lib/connection.js:837:7)
at _createCollection (node_modules/mongoose/lib/model.js:1106:16)
at Function.init (node_modules/mongoose/lib/model.js:1119:16)
at Mongoose._model (node_modules/mongoose/lib/mongoose.js:667:9)
at Mongoose.model (node_modules/mongoose/lib/mongoose.js:622:27)
...
```

My code hasn't changed it was just a mongoose dependency upgrade. When I rollback the upgrade everything works fine again.

### Steps to Reproduce

upgrade from `8.9.6` to `8.10.0`. Run tests that use jest and mongoose connections.

### Expected Behavior

_No response_

hasezoey:
**Summary**

The recent change in #15229 has added a timeout, which can lead to having the node process around for longer than expected (ie until the timeout has fired), which can cause jest to detect open handles.
(I had also encountered this while upgrading typegoose)

Also the check of `unref` existing is made for alternative engines like deno and bun, where i had previously encountered that `unref` may not always be implemented.

fixes #15241

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -842,6 +842,11 @@ Connection.prototype._waitForConnect = async function _waitForConnect() {
           },
           bufferTimeoutMS
         );
+
+        // dont cause this timeout to keep the js process alive
+        if ('unref' in timeout) {
+          timeout.unref();
+        }
       })
     ]);
 


vkarpov15 (commented):
I'm not entirely convinced this is the right behavior because `unref()` here undermines the primary purpose of `_waitForConnect()`, which is to error out if the user tries to execute operations on a connection that is never opened. For example, with this change, the following script would exit immediately with no output:

```javascript
'use strict';

(async function() {
  await mongoose.connection.listCollections();
})();
```

I'm not saying we shouldn't merge this, but I would like to discuss further in #15241 to understand the use case here.

hasezoey:
This PR can be closed once #15251 lands, or should this be kept open for something?

vkarpov15:
No, let's close this in favor of #15251. Thanks for your help with investigating this issue, I appreciate it :+1:

vkarpov15:
Accidentally clicked "Merge" instead of "Close", sorry about that.

PR #15250, opened 2025-02-10
 perf: some more micro optimizations for find() and findOne() re: #14906 

vkarpov15:
**Summary**

Avoid some string concatenation and other small issues. No massive improvements, but hard to find any massive improvements other than removing `_executionStack`

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -65,6 +65,9 @@ const queryOptionMethods = new Set([
   'wtimeout'
 ]);
 
+// Map from operation name to the name of the function that executes the actual operation against MongoDB.
+// Called a thunk for legacy reasons, "thunk" means function that takes exactly 1 param, a callback.
+// Currently `_countDocuments()`, etc. are async functions that take no params.
 const opToThunk = new Map([
   ['countDocuments', '_countDocuments'],
   ['distinct', '__distinct'],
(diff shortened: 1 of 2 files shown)


lib/query.js
```diff
@@ -2366,7 +2382,7 @@ Query.prototype._find = async function _find() {
   }
 
   if (!mongooseOptions.populate) {
-    const versionKey = _this.schema.options.versionKey;
+    const versionKey = this.schema.options.versionKey;
```
hasezoey, line 2385: I dont exactly know why this change was made, but i think in this function `_this` can be completely removed and directly use `this`.

lib/query.js
```diff
@@ -65,6 +65,22 @@ const queryOptionMethods = new Set([
   'wtimeout'
 ]);
 
+const opToThunk = new Map([
```
hasezoey, line 68: Consider adding a comment to explain this a bit better.

hasezoey (approved):

lib/query.js
vkarpov15: Fixed in a41b821

lib/query.js
vkarpov15: Fixed in a41b821

PR #15251, opened 2025-02-11
fix(model): avoid adding timeout on `Model.init()` buffering to avoid unintentional dangling open handles

### LINKED ISSUE #15241: Jest detect open handles in tests after upgrading from 8.9.6 to 8.10.0

vkarpov15:
Fix #15241

**Summary**

Make `Model.init()` skip buffering timeouts introduced in #15229

Based on discussion from #15241, I think this is a reasonable approach which gives us the benefits of #15229 without the implicit dangling handles. The following Jest test won't complain about open handles:

```js

// Define a User schema and model
  name: String,
  email: String,
  age: Number
});

describe('Database connection', () => {
  it('test', function () {

  });
});
```

However, the following script will still error out with buffering timeout, as opposed to silently exiting with no error.

```js
'use strict';

(async function() {
  await mongoose.connection.listCollections();
})();
```

So explicit connection helper calls will be subject to buffer timeouts, but we will ignore buffer timeouts for `Model.init()` because `Model.init()` is called implicitly and there are use cases where you may want to create a model without opening the underlying connection, like testing document functionality.

**Examples**

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -818,32 +818,40 @@ Connection.prototype.dropCollection = async function dropCollection(collection) 
 /**
  * Waits for connection to be established, so the connection has a `client`
  *
+ * @param {Boolean} [noTimeout=false] if set, don't put a timeout on the operation. Used internally so `mongoose.model()` doesn't leave open handles.
  * @return Promise
  * @api private
  */
 
-Connection.prototype._waitForConnect = async function _waitForConnect() {
+Connection.prototype._waitForConnect = async function _waitForConnect(noTimeout) {
   if ((this.readyState === STATES.connecting || this.readyState === STATES.disconnected) && this._shouldBufferCommands()) {
     const bufferTimeoutMS = this._getBufferTimeoutMS();
     let timeout = null;
     let timedOut = false;
     // The element that this function pushes onto `_queue`, stored to make it easy to remove later
     const queueElement = {};
-    await Promise.race([
-      new Promise(resolve => {
+    if (noTimeout) {
+      await new Promise(resolve => {
         queueElement.fn = resolve;
         this._queue.push(queueElement);
-      }),
-      new Promise(resolve => {
-        timeout = setTimeout(
-          () => {
-            timedOut = true;
-            resolve();
-          },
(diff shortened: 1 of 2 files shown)


lib/connection.js
```diff
@@ -818,32 +818,40 @@ Connection.prototype.dropCollection = async function dropCollection(collection)
...
-        timeout = setTimeout(
-          () => {
-            timedOut = true;
-            resolve();
-          },
-          bufferTimeoutMS
-        );
-      })
-    ]);
+      });
```
hasezoey, line 837: I think this could be de-duplicated by storing the promise in a variable before the if, and just directly awaiting it here and in the other branch adding it to the race.

hasezoey (approved):
Can confirm this fixes the cases i had for typegoose without needing to set any options or other code changes.

lib/connection.js
vkarpov15: Fixed in https://github.com/Automattic/mongoose/pull/15251/commits/0a2da7115116fae133f8242965dcc58cc95549ff

PR #15259, opened 2025-02-13
feat(document): add schemaFieldsOnly option to toObject() and toJSON()

### LINKED ISSUE #15218: Document.prototype.toObject() with discriminators returns all fields and selectively applies getters
### Prerequisites

### Mongoose version

8.9.7

### Node.js version

20.18.0

### MongoDB server version

8.0.4

_No response_

### Description

After updating a discriminator key in a document with schema discriminators and getters, the Document includes the fields for the current discriminator, as expected. However, _Document.toObject()_ returns all fields, including those defined in a different discriminator schema, but only applies getters to the fields matching the current discriminator.

### Steps to Reproduce

The following script will fail when checking the discriminator fields. In particular, the _Document_ contains only the fields relevant to the discriminator, but the object returned by _Document.toObject()_ contains all fields, including those for the other discriminator schema.

```
      const conn = mongoose.createConnection(..);

      const baseType = 'baseType';
      const type1Key = 'Type1';
      const type2Key = 'Type2';

      const baseSchema = new mongoose.Schema({
        field: String,
        key: String,
      }, {discriminatorKey: 'key', toObject: {getters: true}});
      const dSchema1 = new mongoose.Schema({
        field1: {
          type: String,
          get(value) {
            return `${value} transformed`;
          },
        },
      });
      const dSchema2 = new mongoose.Schema({
        field2: {
          type: String,
          get(value) {
            return `${value} transformed`;
          },
        },
      });
      baseSchema.discriminator(type1Key, dSchema1);
      baseSchema.discriminator(type2Key, dSchema2);

      const TestModel = conn.model(baseType, baseSchema);

      const field = 'field val';
      const field1 = 'field1 val';
      const field1Transformed = `${field1} transformed`;
      const cdm = new TestModel({
        key: type1Key,
        field,
        field1,
      });
      let cd = await cdm.save();
      assert.equal(cd.field2, undefined);
      assert.equal(cd.field1, field1Transformed);
      cd = cd.toObject();
      assert.equal(cd.field, field);
      assert.equal(cd.key, type1Key);
      assert.equal(cd.field1, field1Transformed);
      assert.equal(cd.field2, undefined);

      const field2 = 'field2 val';
      const field2Transformed = `${field2} transformed`;
      await TestModel.updateOne(
        { _id: cd._id },
        {
          key: type2Key,
          field2,
        },
        {overwriteDiscriminatorKey: true},
      );
      let cd2 = await TestModel.findById(cd._id);
      assert.equal(cd2.field2, field2Transformed);
      assert.equal(cd2.field1, undefined);
      cd2 = cd2.toObject();
      assert.equal(cd2.field, field);
      assert.equal(cd2.key, type2Key);
      assert.equal(cd2.field2, field2Transformed);
      assert.equal(cd2.field1, undefined);

      conn.deleteModel(baseType);
```

### Expected Behavior

As shown in the script above, since a Document with a discriminator includes only the fields relevant to the currently set discriminator key (prior to calling _toObject()_), I would expect the object returned by _toObject()_ to also only include those fields defined in that discriminator schema. 

_If_ all fields must be returned by _toObject()_ (even those not relevant to the current discriminator schema) I would expect all getters to have been called.

vkarpov15:
Fix #15218

**Summary**

`toObject()` and `toJSON()` currently return an object with all fields from the document in the database, which may include fields that aren't in the schema. To work around that, we added a `schemaFieldsOnly` option to `toObject()` and `toJSON()`, which will loop over schema paths when getting values vs just cloning the entirety of `_doc`.

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -14325,8 +14325,8 @@ describe('document', function() {
     );
 
     const doc2 = await Discriminator.findById(doc).orFail();
-    assert.strictEqual(doc2.field1, undefined);
     assert.strictEqual(doc2.field2, 'test2');
+    assert.strictEqual(doc2.field1, undefined);
 
     const obj = doc2.toObject();
     assert.strictEqual(obj.field2, 'test2');
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -14296,6 +14296,93 @@ describe('document', function() {
 
     delete mongoose.Schema.Types.CustomType;
   });
+
+  it('supports schemaFieldsOnly option for toObject() (gh-15258)', async function() {
+    const schema = new Schema({ key: String }, { discriminatorKey: 'key' });
+    const subschema1 = new Schema({ field1: String });
+    const subschema2 = new Schema({ field2: String });
+
+    const Discriminator = db.model('Test', schema);
+    Discriminator.discriminator('type1', subschema1);
+    Discriminator.discriminator('type2', subschema2);
+
+    const type1Key = 'type1';
+    const type2Key = 'type2';
+
+    const doc = await Discriminator.create({
+      key: type1Key,
+      field1: 'test value'
+    });
+
+    await Discriminator.updateOne(
+      { _id: doc._id },
+      {
+        key: type2Key,
+        field2: 'test2'
+      },
+      { overwriteDiscriminatorKey: true }
+    );
+
+    const doc2 = await Discriminator.findById(doc).orFail();
+    assert.strictEqual(doc2.field1, undefined);
+    assert.strictEqual(doc2.field2, 'test2');
+
+    const obj = doc2.toObject();
+    assert.strictEqual(obj.field2, 'test2');
+    assert.strictEqual(obj.field1, 'test value');
+
+    const obj2 = doc2.toObject({ schemaFieldsOnly: true });
+    assert.strictEqual(obj.field2, 'test2');
+    assert.strictEqual(obj2.field1, undefined);
+  });
+
+  it('supports schemaFieldsOnly on nested paths, subdocuments, and arrays (gh-15258)', async function() {
+    const subSchema = new Schema({
+      title: String,
+      description: String
+    }, { _id: false });
+    const taskSchema = new Schema({
+      name: String,
+      details: {
+        dueDate: Date,
+        priority: Number
+      },
+      subtask: subSchema,
+      tasks: [subSchema]
+    });
+    const Task = db.model('Test', taskSchema);
+
+    const doc = await Task.create({
+      _id: '0'.repeat(24),
+      name: 'Test Task',
+      details: {
+        dueDate: new Date('2024-01-01'),
+        priority: 1
+      },
+      subtask: {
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -204,30 +204,32 @@ declare module 'mongoose' {
...
+    /** if true, convert Maps to POJOs. Useful if you want to `JSON.stringify()` the result of `toObject()`. */
+    flattenMaps?: boolean;
+    /** if true, convert any ObjectIds in the result to 24 character hex strings. */
+    flattenObjectIds?: boolean;
+    /** apply all getters (path and virtual getters) */
+    getters?: boolean;
     /** remove empty objects (defaults to true) */
     minimize?: boolean;
+    /** If true, the resulting object will only have fields that are defined in the document's schema. By default, `toJSON()` returns all fields in the underlying document from MongoDB, including ones that are not listed in the schema. */
+    schemaFieldsOnly?: boolean;
```
hasezoey, line 220:
```suggestion
    /** If true, the resulting object will only have fields that are defined in the document's schema. By default, `toJSON()` & `toObject()` returns all fields in the underlying document from MongoDB, including ones that are not listed in the schema. */
    schemaFieldsOnly?: boolean;
```

test/document.test.js
```diff
@@ -14296,6 +14296,93 @@ describe('document', function() {
...
+    const schema = new Schema({ key: String }, { discriminatorKey: 'key' });
+    const subschema1 = new Schema({ field1: String });
+    const subschema2 = new Schema({ field2: String });
+
+    const Discriminator = db.model('Test', schema);
+    Discriminator.discriminator('type1', subschema1);
+    Discriminator.discriminator('type2', subschema2);
+
+    const type1Key = 'type1';
+    const type2Key = 'type2';
```
hasezoey, line 14310: For consistency, this could also be used in the `.discriminator` calls

test/document.test.js
```diff
@@ -14296,6 +14296,93 @@ describe('document', function() {
...
+      {
+        key: type2Key,
+        field2: 'test2'
+      },
+      { overwriteDiscriminatorKey: true }
+    );
+
+    const doc2 = await Discriminator.findById(doc).orFail();
+    assert.strictEqual(doc2.field1, undefined);
+    assert.strictEqual(doc2.field2, 'test2');
```
hasezoey, line 14328:
```suggestion
    assert.strictEqual(doc2.field2, 'test2');
    assert.strictEqual(doc2.field1, undefined);
```

Align with the style of the following 2 checks

hasezoey (approved):
LGTM, some minor style suggestions.

test/document.test.js
vkarpov15: I went the other approach in [872a2e7](https://github.com/Automattic/mongoose/pull/15259/commits/872a2e7dc170fbfbbc4d10d52969ee90a8c9c62d) and removed `type1Key` and `type2Key` variables for consistency. I tend to dislike adding variable definitions for strings in tests because they add SLOC without improving readability IMO

PR #15261, opened 2025-02-17
types: make type inference logic resilient to no Buffer type due to missing `@types/node`

### LINKED ISSUE #15121: How to write a generic function to resolve PopulatedDoc type variables?
### Prerequisites

### Mongoose version

8.9.1

### Node.js version

20.10.0

### MongoDB version

6.0.2

### Operating system

macOS

_No response_

### Issue

In my project, I often find myself having to implement this code:

```ts
// Check if 'myPopulatedDocProp' has already been populated in a previous part of the code
if (myDocument.myPopulatedDocProp instanceof AnotherModel) {
  return myDocument.myPopulatedDocProp;
}

// If 'myPopulatedDocProp' was not populated, perform a database query to find and populate the document property
const result = await AnotherModel
  .findById(myDocument.myPopulatedDocProp) // Query the AnotherModel to find the document by ID
  .orFail(); // Throw an error if the document is not found
return result; // Return the populated document property

```

Is it possible to write a generic function so that I don't have to rewrite the same exact code but with different models?
I tried with this code but it doesn't work:

```ts
export const findOrReturnInstance = async <T extends Document>(
  populatedDoc: PopulatedDoc<T>,
  model: Model<T>,
) => {
  if (populatedDoc instanceof model) {
    return populatedDoc;
  }
  const result = await model
    .findById(populatedDoc)
    .orFail();
  return result;
};
```

Is it the right approach or am I doing something wrong?
Attention, with this generic function I want to ensure that any virtual variables and instance methods, etc., are preserved.

Thank you for your help. 🙏🏻

vkarpov15:
Fix #15121

**Summary**

#15121  points out an unpleasant edge case where our lean type inference breaks completely if `@types/node` isn't installed, because `Buffer === any`. Not something we can write tests for unfortunately. But this PR should make Mongoose more resilient to that case, I replaced all `extends Buffer` checks and we don't have any `: Buffer` or `?: Buffer` params or properties.

**Examples**

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -712,47 +712,53 @@ declare module 'mongoose' {
     [K in keyof T]: FlattenProperty<T[K]>;
   };
 
-  export type BufferToBinaryProperty<T> = T extends Buffer
-    ? mongodb.Binary
-    : T extends Types.DocumentArray<infer ItemType>
-      ? Types.DocumentArray<BufferToBinary<ItemType>>
-      : T extends Types.Subdocument<unknown, unknown, infer SubdocType>
-        ? HydratedSingleSubdocument<BufferToBinary<SubdocType>>
-        : BufferToBinary<T>;
+  export type BufferToBinaryProperty<T> = unknown extends Buffer
+    ? T
+    : T extends Buffer
+      ? mongodb.Binary
+      : T extends Types.DocumentArray<infer ItemType>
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM 👍

hasezoey (approved):

PR #15266, opened 2025-02-20
fix(model+connection): return MongoDB BulkWriteResult instance even if no valid ops

vkarpov15:
**Summary**

MongoDB's `BulkWriteResult` is a class, not just an interface, so we need to ensure we return an instance of that class in the "no valid ops" case.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -10,7 +10,6 @@ const Document = require('./document');
 const DocumentNotFoundError = require('./error/notFound');
 const EventEmitter = require('events').EventEmitter;
 const Kareem = require('kareem');
-const { MongoBulkWriteError } = require('mongodb');
 const MongooseBulkWriteError = require('./error/bulkWriteError');
 const MongooseError = require('./error/index');
 const ObjectParameterError = require('./error/objectParameter');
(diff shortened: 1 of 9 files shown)


lib/model.js
```diff
@@ -3399,7 +3398,13 @@ Model.bulkWrite = async function bulkWrite(ops, options) {
   const ordered = options.ordered == null ? true : options.ordered;
 
   if (ops.length === 0) {
-    return getDefaultBulkwriteResult();
+    const BulkWriteResult = this.base.driver.get().BulkWriteResult;
+    const bulkWriteResult = new BulkWriteResult(getDefaultBulkwriteResult(), false);
+    bulkWriteResult.n = 0;
+    bulkWriteResult.mongoose = bulkWriteResult.mongoose || {};
+    bulkWriteResult.mongoose.validationErrors = [];
+    bulkWriteResult.mongoose.results = [];
```
hasezoey, line 3406: This looks like a somewhat common operation, should this maybe be its own function?

hasezoey (approved):

lib/model.js
vkarpov15: Done in 633ea9d

PR #15277, opened 2025-02-26
fix(browser): expose "SchemaTypeOptions"

hasezoey:
**Summary**

This PR exposes `SchemaTypeOptions` on the browser build, like it is on the non-browser build.
As a workaround, it is already accessible via `mongoose.Schema.Types.Mixed.prototype.OptionsConstructor` (or similar) without this PR.

Btw, is the browser build still something that is supported?

### DIFF
--- a/lib/browser.js
+++ b/lib/browser.js
@@ -84,6 +84,15 @@ exports.VirtualType = require('./virtualType');
 exports.SchemaType = require('./schemaType.js');
 
 /**
+ * The constructor used for schematype options
+ *
+ * @method SchemaTypeOptions
+ * @api public
+ */
+
+exports.SchemaTypeOptions = require('./options/schemaTypeOptions');
+
+/**
  * Internal utils
  *
  * @property utils


vkarpov15:
The browser build is still supported, but I've been thinking about splitting it off into a plugin for Mongoose 9 for several reasons, including the following. What do you think?

1. Makes bundling difficult
2. Browser build accounts for around 30% of the Mongoose npm package size

hasezoey:
> but I've been thinking about splitting it off into a plugin for Mongoose 9 for several reasons, including the following. What do you think?

Sounds great to me, this would mean, aside from the points you already said, that there would likely be less confusion about mongoose normal build and browser build, right?
Though i am curious, this "plugin" way, would it continue to use the current `lib/browser*` code, or try to compile it whole (without a driver)?
Also, wouldnt this mean that now there would be 2 packages, ie `mongoose` and `mongoose-browser`, or did you think of a different way?

In any case, it would likely be good to have a discussion about this instead of doing it in this PR.

vkarpov15:
Yeah I'll move the discussion to a separate issue #15296. But yes, the idea would be to have a separate `@mongoosejs/browser` or `mongoose-browser` npm package that users would have to explicitly add to use the browser build.

vkarpov15 (approved):

PR #15293, opened 2025-03-02
fix(AggregationCursor): make next() error if schema pre('aggregate') middleware throws error

### LINKED ISSUE #15279: AggregationCursor will not emit errors like QueryCursor when errors thrown in pre hook.
### Prerequisites

### Mongoose version

8.10.0

### Node.js version

20.15

### MongoDB server version

5

_No response_

### Description

Using Model.aggregate.cursor() will not emit errors like QueryCursor when errors thrown in pre hook.

```
schema.pre('aggregate', function(next) {
  if (!this.options.allowed) {  
    throw new Error(`hey, that's nacho cheese!`)
  }
  next()
})

// -----------

// this would throw
const res = await Model.aggregate([{ $limit: 1 }], { allowed: false }).exec()
return res

// this will continue and stream results
const cursor = Model.aggregate([{ $limit: 1 }], { allowed: false }).cursor()
const aahsss = []
for await(const it of cursor) {
   aahsss.push(it)
}

```

QueryCursor seems to have this implemented. Is this by design, because aggregations aren't "ORMey"?

### Steps to Reproduce

- create a schema with a pre('aggregate') hook that throws an error
- use Model.aggregate([{$limit: 1 }]).cursor()
- see a document rather that an error thrown or emitted

### Expected Behavior

expect aggregation cursor to emit an error from a failed pre('aggregate') hook

vkarpov15:
Fix #15279

**Summary**

QueryCursor handles case where pre middleware throws an error, but AggregationCursor does not. For consistency, AggregationCursor `next()` should throw an error if pre('aggregate') middleware errored.

**Examples**

### DIFF
--- a/test/aggregate.test.js
+++ b/test/aggregate.test.js
@@ -1300,4 +1300,34 @@ describe('aggregate: ', function() {
       });
     }, /Aggregate `near\(\)` argument has invalid coordinates, got ""/);
   });
+
+  it('cursor() errors out if schema pre aggregate hook throws an error (gh-15279)', async function() {
+    const schema = new Schema({ name: String });
+
+    schema.pre('aggregate', function(next) {
+      if (!this.options.allowed) {
+        throw new Error('Unauthorized aggregate operation: only allowed operations are permitted');
+      }
+      next();
+    });
+
+    const Test = db.model('Test', schema);
+
+    await Test.create({ name: 'test1' });
+
+    assert.rejects(
+      async() => {
+        await Test.aggregate([{ $limit: 1 }], { allowed: false }).exec();
+      },
+      err => err.message === 'Unauthorized aggregate operation: only allowed operations are permitted'
+    );
+
+    const cursor = Test.aggregate([{ $limit: 1 }], { allowed: false }).cursor();
+    assert.rejects(
+      async() => {
+        await cursor.next();
+      },
+      err => err.message === 'Unauthorized aggregate operation: only allowed operations are permitted'
+    );
+  });
 });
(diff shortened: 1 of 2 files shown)


test/aggregate.test.js
```diff
@@ -1300,4 +1300,34 @@ describe('aggregate: ', function() {
...
+        throw new Error('Unauthorized aggregate operation: only allowed operations are permitted');
+      }
+      next();
+    });
+
+    const Test = db.model('Test', schema);
+
+    await Test.create({ name: 'test1' });
+
+    assert.rejects(
```
hasezoey, line 1318:
```suggestion
    await assert.rejects(
```

test/aggregate.test.js
```diff
@@ -1300,4 +1300,34 @@ describe('aggregate: ', function() {
...
+      async() => {
+        await Test.aggregate([{ $limit: 1 }], { allowed: false }).exec();
+      },
+      err => err.message === 'Unauthorized aggregate operation: only allowed operations are permitted'
+    );
+
+    const cursor = Test.aggregate([{ $limit: 1 }], { allowed: false }).cursor();
+    assert.rejects(
```
hasezoey, line 1326:
```suggestion
    await assert.rejects(
```

hasezoey (changes requested):
LGTM, but i think the tests are not actually awaited correctly.

AbdelrahmanHafez (approved):

test/aggregate.test.js
AbdelrahmanHafez: It's for this reason I prefer the `const err = await promise.then(()=> null, err => err);` way of asserting against errors.

hasezoey (approved):

PR #15310, opened 2025-03-10
feat(NODE-6506): Add support for encrypted schemas

baileympearson:
**Summary**

This PR is an exact copy of https://github.com/mongodb-js/mongoose/pull/5, opened here for ease of review.  @vkarpov15 

This PR adds support for declaring encrypted schemas, as the first step in adding first class support for CSFLE to mongoose.  Functionally:

1. A new schema option, `encryptionType`, has been added.  This is required for schemas that are declaring encrypted fields, and it determines whether the schema will be configured for 'csfle' or 'qe'.
2. Keys in schemas can now be configured with an additional `encrypt` option.  This option contains metadata for libmongocrypt to encrypt the field (which will be automatically included in a schemaMap or encryptedFieldsMap).  The contents of this document are exactly the same as the fields used to configure a field for csfle or qe, except that `bsonType` is not required (inferred from the schema type).

This PR also updates all the schema modifiers / cloning methods to account for updating encrypted fields as well.

**Examples**

Declare an encrypted schema:

```typescript
const encryptedSchema = new Schema({
  name: { type: String, encrypt: { keyId: '...' } },
  friend: { 
      // nested objects allowed
      name: { type: String, encrypt: { keyId: '...' } },
  },
  // arrays allowed - but becase of how CSFLE/QE encrypted arrays, the arrays are encrypted as a whole
  books: { type: [String], encrypt: { keyId: '...' } }
}, { encryptionType: 'qe' });
```

Modify / clone encrypted schemas

```typescript
encryptedSchema.clone();

encryptedSchema.add({ name: String }) // name is no longer encrypted
encryptedSchema.add({ newKey: { type: String, encrypt: { keyId: '...' } }); // new encrypted key added

encryptedSchema.remove('name'); // name no longer in schema at all

encryptedSchema.pick('friends'); // returns a new schema with only one field, `friends`, that is encrypted.

encryptedSchema.omit('friends'); // returns a new schema with all fields except `frields`
```

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -86,6 +86,7 @@ const numberRE = /^\d+$/;
  * - [pluginTags](https://mongoosejs.com/docs/guide.html#pluginTags): array of strings - defaults to `undefined`. If set and plugin called with `tags` option, will only apply that plugin to schemas with a matching tag.
  * - [virtuals](https://mongoosejs.com/docs/tutorials/virtuals.html#virtuals-via-schema-options): object - virtuals to define, alias for [`.virtual`](https://mongoosejs.com/docs/api/schema.html#Schema.prototype.virtual())
  * - [collectionOptions]: object with options passed to [`createCollection()`](https://www.mongodb.com/docs/manual/reference/method/db.createCollection/) when calling `Model.createCollection()` or `autoCreate` set to true.
+ * - [encryptionType]: the encryption type for the schema.  Valid options are `csfle` or `queryableEncryption`.  See https://mongoosejs.com/docs/field-level-encryption.
  *
  * #### Options for Nested Schemas:
  *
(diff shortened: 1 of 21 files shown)


lib/schema.js
```diff
@@ -495,7 +499,17 @@ Schema.prototype.pick = function(paths, options) {
   }
 
   for (const path of paths) {
-    if (this.nested[path]) {
+    if (path in this.encryptedFields) {
```
vkarpov15, line 502: Is this code path necessary? `newSchema.add({ [path]: schematype });` ideally should handle adding the full schematype to the new schema correctly. Is the issue that `this.encryptedFields` isn't copied over?

vkarpov15 (approved):
I had one very minor comment, but overall LGTM. I will merge this to 8.13 release branch

lib/schema.js
baileympearson: Yeah, because schematype doesn't include encryption information.

Its been awhile but I think I intentionally decided against storing encryption information on the SchemaType because it seemed like a separate concern.  SchemaTypes declare the type of the field, encryption configuration is declared on a schema.

I could reconsider though, if you think that is preferable.

vkarpov15:
Merging into csfle branch :+1:

PR #15320, opened 2025-03-18
feat(NODE-6507): generate encryption configuration on mongoose connect

baileympearson:
**Summary**

This PR contains the meat and potatoes of the automatic CSFLE/QE integration into mongoose.  With the changes included, users can now create encrypted schemas, instantiate models with them, insert and query using their model and have the resultant documents be encrypted and decrypted automatically.

This PR contains a lot of test changes.  I tried to be as exhaustive to ensure as much coverage as possible, and because the schema format is unlikely to change (changing the format of encryptedFieldsMap or schemaMap would break a lot of drivers and applications).  So ideally these tests are not a large maintenance burden.

**Examples**

See the integration tests in `encryption.test.ts` for an example of how the API works.


lib/schema/bigint.js
```diff
@@ -255,7 +255,7 @@ SchemaBigInt.prototype.toJSONSchema = function toJSONSchema(options) {
 };
 
 SchemaBigInt.prototype.autoEncryptionType = function autoEncryptionType() {
-  return 'int64';
+  return 'long';
```
baileympearson, line 258: The actual bson type strings were not the exact bson type strings that the server expects for the schemas modified in this PR.

scripts/configure-cluster-with-encryption.sh
```diff
@@ -4,11 +4,11 @@
 
 # this script downloads all tools required to use FLE with mongodb, then starts a cluster of the provided configuration (sharded on 8.0 server)
 
-export CWD=$(pwd);
-export DRIVERS_TOOLS_PINNED_COMMIT=35d0592c76f4f3d25a5607895eb21b491dd52543;
+export CWD=$(pwd)
+export DRIVERS_TOOLS_PINNED_COMMIT=4e18803c074231ec9fc3ace8f966e2c49d9874bb
```
baileympearson, line 8: This is the latest commit as of yesterday ~noon.  I have another ticket in this epic to better figure out how to integrate with drivers-evergreen-tools so it is maintainable for mongoose developers, but I haven't started it yet.  So for now, I'm sticking with this script.

docs/field-level-encryption.md
```diff
@@ -151,3 +151,68 @@ To declare a field as encrypted, you must:
...
+
+const connection = mongoose.createConnection();
+const UserModel = connection.model('User', encryptedUserSchema);
+```
+
+### Connecting and configuring encryption options
+
+CSFLE/QE in Mongoose work by generating the encryption schema that the MongoDB driver expects for each encrypted model on the connection.  This happens automatically the model's connection is established.
+
+Queryable encryption and csfle requires all the same configuration as outlined in <>, except for the schemaMap or encryptedFieldsMap options.
```
vkarpov15, line 169: Minor nitpick: consistently capitalize CSFLE vs csfle

docs/field-level-encryption.md
```diff
@@ -151,3 +151,68 @@ To declare a field as encrypted, you must:
 2. Choose an encryption type for the schema and configure the schema for the encryption type
 
 Not all schematypes are supported for CSFLE and QE.  For an overview of valid schema types, refer to MongoDB's documentation.
+
+### Registering Models
+
+Encrypted schemas must be registered on a connection, not the Mongoose global:
```
vkarpov15, line 157: Is there a particular reason for this? `mongoose.model('User', encryptedUserSchema)` is equivalent to `mongoose.connection.model('User', encryptedUserSchema)` so there shouldn't be any reason why you can't register an encrypted model on the Mongoose global. Unless there's some required configuration for `createConnection()`.

Also, should be "Encrypted models must be registered on a connection"

lib/drivers/node-mongodb-native/connection.js
```diff
@@ -343,6 +354,54 @@ NativeConnection.prototype.createClient = async function createClient(uri, optio
...
+  */
+NativeConnection.prototype._buildEncryptionSchemas = function() {
+  const qeMappings = {};
+  const csfleMappings = {};
+
+  // If discriminators are configured for the collection, there might be multiple models
+  // pointing to the same namespace.  For this scenario, we merge all the schemas for each namespace
+  // into a single schema.
+  // Notably, this doesn't allow for discriminators to declare multiple values on the same fields.
+  for (const model of Object.values(this.models)) {
```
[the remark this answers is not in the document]
vkarpov15: This idea of merging all discriminator schemas into a single top-level schema is a neat trick, I may have to use this in other parts of the code.

I'm not concerned with the performance implications of this code. Like you said, I don't expect this code to run in performance-sensitive contexts.

My only concern here is, as you mentioned in the docs above, "As a result, discriminators that declare the same key with different types are not supported." It would be great to add some logic to enforce this assumption. Right now you would just silently take the last discriminator type for a given path.

test/encryptedSchema.test.js
```diff
@@ -22,94 +22,174 @@ function schemaHasEncryptedProperty(schema, path) {
...
+          ]
+        }
+      }
+    };
 
-      it('then the schema has a nested property that is encrypted', function() {
-        assert.ok(schemaHasEncryptedProperty(schema, 'encrypted'));
-      });
-    });
+    for (const [description, { schemaFactory, predicate, schemaMap, encryptedFields }] of Object.entries(tests)) {
```
vkarpov15, line 184: I would prefer if you flattened out these tests into distinct `it()` calls. You aren't saving very much SLOC and you're adding some mental overhead for people reviewing this code.

vkarpov15 (changes requested):
I'd also like to see tests covering:

1. Subdocuments
2. Document arrays
3. Maps

If you can do CSFLE with nested paths, you should be able to also do CSFLE with subdocuments, [which are slightly different](https://mongoosejs.com/docs/subdocs.html#subdocuments-versus-nested-paths).

test/encryptedSchema.test.js
baileympearson: Sure, done

docs/field-level-encryption.md
baileympearson: The requirements for CSFLE are:

- auto encryption options must be provided to the mongoose connection creation method
- schemas and models must be declared before mongoose instantiates a mongo client

The workflow I imagined was:

```javascript
const connection = createConnection();
// declare encrypted schemas, initialize models on `connection`
await connection.openUri(<>, { autoEncryption: { ... } });
```

I guess this would also work with the global connection, like so?

```
// declare encrypted schemas, initialize models on the mongoose global
const connection = mongoose.connect(<>, { autoEncryption: { ... } });
```

Do you see any issues with this?  If not, I'll add tests just to confirm and update the docs.

docs/field-level-encryption.md
baileympearson: Good call.  Fixed

lib/drivers/node-mongodb-native/connection.js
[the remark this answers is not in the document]
baileympearson: I added errors that are thrown when:

1. the user has multiple discriminators declaring properties with the same key
2. the user has a discriminator with a key that exists in the base schema as well

Let me know what you think of my approach for the 2.  Adding validation during connection establishment seemed impossible because discriminator schemas have the parent schema merged into them.  That means that every child schema has all the same keys as the parent schema.  After the merging has happened there's no way to know if the duplicate keys were a result of the merge or if the duplicate keys were provided to the discriminator.

To solve this, I added logic in the schema merge function in the discriminator method that throws if the child schema has a key that exists in the parent.  This means that these errors are thrown at discriminator initialization time, whereas the errors outlined in 1. are thrown when the connection is established.

I think this is an acceptable difference, but let me know if you feel otherwise.

baileympearson:
> I'd also like to see tests covering:
> 
> - Subdocuments
> - Document arrays
> - Maps
> If you can do CSFLE with nested paths, you should be able to also do CSFLE with subdocuments, [which are slightly different](https://mongoosejs.com/docs/subdocs.html#subdocuments-versus-nested-paths).

I believe I now have tests for all of these.  Let me know if I'm still missing anything

test/encryption/encryption.test.js
```diff
@@ -185,6 +184,90 @@ describe('encryption integration tests', () => {
...
+
+          const [{ _id }] = await model.insertMany([{ a: {
+            name: 'bailey'
+          } }]);
+          const encryptedDoc = await utilClient.db('db').collection('schemas').findOne({ _id });
+
+          assert.ok(isBsonType(encryptedDoc.a, 'Binary'));
+          assert.ok(encryptedDoc.a.sub_type === 6);
+
+          const doc = await model.findOne({ _id });
```
vkarpov15, line 263: Minor question: can you actually query by encrypted map fields using queryable encryption?

lib/drivers/node-mongodb-native/connection.js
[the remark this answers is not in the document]
vkarpov15: The approach in [53d31a6](https://github.com/Automattic/mongoose/pull/15320/commits/53d31a6c59b8106dd517f47f2ea39439b5c0f211) works for me :+1:

test/encryption/encryption.test.js
baileympearson: No - queries are limited when using QE.  Here's the documentation on supported query types: https://www.mongodb.com/docs/manual/core/queryable-encryption/reference/supported-operations/#supported-and-unsupported-bson-types

lib/drivers/node-mongodb-native/connection.js
```diff
@@ -365,20 +369,30 @@ NativeConnection.prototype._buildEncryptionSchemas = function() {
...
-      qeMappings[namespace] ??= new Schema({}, { encryptionType: 'queryableEncryption' });
-      qeMappings[namespace].add(schema);
+    const mappings = schema.encryptionType() === 'csfle' ? csfleMappings : qeMappings;
+
+    mappings[namespace] ??= new Schema({}, { encryptionType: schema.encryptionType() });
+
+    const isNonRootDiscriminator = schema.discriminatorMapping && !schema.discriminatorMapping.isRoot;
+    if (isNonRootDiscriminator) {
+      const rootSchema = schema._baseSchema;
+      schema.eachPath((pathname) => {
```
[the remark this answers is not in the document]
baileympearson: Good catch - I wrote a test for this scenario and you were right.  

I've been trying out some different approaches and I can't find a better solution than something like this:

```typescript
  function* allPaths(schema, prefix) {
    for (const path of Object.keys(schema.paths)) {
      const fullPath = prefix != null ? `${prefix}.${path}` : path;
      if (schema.path(path).instance === 'Embedded') {
        yield* allPaths(schema.path(path).schema, fullPath);
      } else {
        yield fullPath;
      }
    }
  }

  const paths = new Set(allPaths(schema));

  for (const path of allPaths(model.schema)) {
    if (paths.has(path) && (model.schema._hasEncryptedField(path) || schema._hasEncryptedField(path))) {
      throw new Error(`cannot declare an encrypted field on child schema overriding base schema. key=${path}`);
    }
  }
```

This generates all possible paths in the schema (the above doesn't handle arrays, but encryption on fields in arrays isn't supported so that's out of scope here).  Not my first choice, because this feels brittle.  I'll keep looking at it but

1. Can you think of a better solution here than recursively iterating over all paths + nested schemas?  
2. If not, does a utility that I could use like `allPaths` exist somewhere?

docs/field-level-encryption.md
baileympearson: Yup, this works!  I added a test for this scenario.

lib/drivers/node-mongodb-native/connection.js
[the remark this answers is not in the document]
vkarpov15: I don't think recursively checking all paths is necessary here because you just want to find conflicts in the top-level paths. So if discriminator schema has a path `pathname` with an encrypted field, and root schema has a nested path with `rootSchema.nested[pathname.split('.')[0]]`, you can already call that a conflict and throw an error. Similarly, if discriminator schema has a nested path `pathname` but in root schema you have `rootSchema.paths[pathname]` then you can also throw an error.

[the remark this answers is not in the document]
baileympearson: Not necessarily, right?  That would throw an error in scenarios where the child schema provides a subdocument but doesn't modify the encrypted path.  ex:

```typescript
const schema = new Schema({
  name: {
    first: { type: String, encrypt: { keyId: [keyId], algorithm } }
  }
});

const discriminatorSchema = new Schema({
  name: new Schema({ age: Number }) // Different path, no encryption, stored as same field in MDB
});
```

I'd expect this to be fine, because there isn't a conflicting path for `name.first`.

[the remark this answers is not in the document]
baileympearson: I tried the approach of determining conflicting base paths, as you suggested but I ran into two complications:

1. It isn't enough to simply detect base path collisions, you need to check that something in the schema also has an encrypted field that uses the base path as well.  This is feasible but requires some logic to allow for asking "is there an encrypted field at some subpath of this path?"
2. It felt like there were more corner cases than not - schemas with only one level of nesting, one level of subdocuments, sub documents on parent but not child and vice versa, multiple layers of nesting of subdocuments, etc.  

I decided that the two above points complicated the approach.  The changes in this PR were much simpler for me to work with..  This approach works by:

1. finding the intersection of all keypaths for all fields (including subdocuments) between the parent and child
2. for each keypath, validates that this keypath isn't encrypted on the parent and on the child

Let me know what you think - I can rework it if you would prefer a different approach.

lib/helpers/model/discriminator.js
```diff
@@ -27,30 +27,32 @@ const CUSTOMIZABLE_DISCRIMINATOR_OPTIONS = {
...
+    if (parentSchema._hasEncryptedField(path) && childSchema._hasEncryptedField(path)) {
+      throw new Error(`encrypted fields cannot be declared on both the base schema and the child schema in a discriminator. path=${path}`);
+    }
+
+    if (parentSchema._hasEncryptedField(path) || childSchema._hasEncryptedField(path)) {
+      throw new Error(`encrypted fields cannot have the same path as a non-encrypted field for discriminators. path=${path}`);
     }
   }
 
+  function* allNestedPaths(schema) {
```
vkarpov15, line 47: This is a neat use case for generators, but I think doing [...Object.keys(schema.paths), ...Object.keys(schema.singleNestedPaths)] would be more concise and more consistent with the rest of the codebase

PR #15322, opened 2025-03-20
fix(document): avoid stripping out fields in discriminator schema after select: false field

### LINKED ISSUE #15308: Having `select: false` in a discriminated model causes later fields to be unselected as well
### Prerequisites

### Mongoose version

8.10.1

### Node.js version

22.8.0

### MongoDB server version

8.0.5

_No response_

### Description

Some fields are not selected when I use `find` without `lean`. The circumstances seem to be very specific:
1. It only seems to take effect when set up in discriminated schemas. If I merge the two schemas from the reproduction part into one, I do not observe the issue.
2. It needs a `select: false` field. If I remove the `select: false` from the `vin`, I do not observe the issue.
3. It only affects fields after the `select: false` field. If I move it to be after the `sunroof` field, then the `sunroof` shows up, but the `__v` does not.
4. If I add `.select('+vin')` to the query, I do not observe the issue.
5. If I add `.select('sunroof')` to the query, I do not observe the issue.
6. If I update the order of fields within the database, only the fields before the `vin` show up, regardless of which fields those are.

This issue started happening with v8.10.1 and is present in every version since, it is working fine with v8.10.0 and earlier.

### Steps to Reproduce

```js
const CarSchema = new mongoose.Schema(
  {
    make: {
      type: String,
      enum: ['mercedes'],
    },
  },
  {
    discriminatorKey: 'make',
  },
);
const MercedesSchema = new mongoose.Schema({
  vin: {
    type: String,
    select: false,
  },
  sunroof: Boolean,
});
const CarModel = mongoose.model('Car', CarSchema);
CarModel.discriminator('Car:mercedes', MercedesSchema, 'mercedes');

const newCar = await CarModel.create({
  make: 'mercedes',
  vin: 'someFakeVin',
  sunroof: true,
});

const car = await CarModel.findById(newCar._id);
const leanCar = await CarModel.findById(newCar._id).lean();

console.log('car', car);
console.log('leanCar', leanCar);
```

Results of `console.log`:
```js
car {
  _id: new ObjectId('67cc89d4c9e92c2ef7da58f8'),
  make: 'mercedes'
}
leanCar {
  _id: new ObjectId('67cc89d4c9e92c2ef7da58f8'),
  make: 'mercedes',
  vin: 'someFakeVin',
  sunroof: true,
  __v: 0
}
```

### Expected Behavior

Expected the `sunroof` (and `__v`) property to be present on the `car` object.

Expected results of `console.log`:
```js
car {
  _id: new ObjectId('67cc89d4c9e92c2ef7da58f8'),
  make: 'mercedes',
  sunroof: true,
  __v: 0
}
leanCar {
  _id: new ObjectId('67cc89d4c9e92c2ef7da58f8'),
  make: 'mercedes',
  vin: 'someFakeVin',
  sunroof: true,
  __v: 0
}
```

vkarpov15:
Fix #15308

**Summary**

When we refactored a `forEach()` into a `for` in `init()`, we left a `return` that should've become a `continue`

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -747,8 +747,8 @@ function init(self, obj, doc, opts, prefix) {
   for (let index = 0; index < len; ++index) {
     i = keys[index];
     // avoid prototype pollution
-    if (i === '__proto__' || i === 'constructor') {
-      return;
+    if (specialProperties.has(i)) {
+      continue;
     }
     path = prefix ? prefix + i : i;
     schemaType = docSchema.path(path);
@@ -1720,7 +1720,7 @@ Document.prototype.$__set = function(pathToMark, path, options, constructing, pa
     const last = next === l;
     cur += (cur ? '.' + parts[i] : parts[i]);
     if (specialProperties.has(parts[i])) {
-      return;
+      continue;
     }
 
     if (last) {
--- a/lib/document.js
+++ b/lib/document.js
@@ -756,7 +756,7 @@ function init(self, obj, doc, opts, prefix) {
     // necessary. This is *only* to catch the case where we queried using the
     // base model and the discriminated model has a projection
     if (docSchema.$isRootDiscriminator && !self.$__isSelected(path)) {
-      return;
+      continue;
     }
 
     const value = obj[i];
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -14422,6 +14422,39 @@ describe('document', function() {
       sinon.restore();
     }
   });
+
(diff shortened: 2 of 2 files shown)


Chee7ah:
I assume this `return` should become a `continue` as well then, right?
https://github.com/Automattic/mongoose/blob/47b88cd641f6ddd7b352bd36f133d594d4d134d1/lib/document.js#L747-L752

hasezoey (approved):
LGTM, i second @Chee7ah suggestion.

PR #15329, opened 2025-03-26
fix(populate): handle virtual populate on array of UUIDs

### LINKED ISSUE #15315: Virtual Populate foreignField is array in combination with UUIDs
Hello,

I have the exact same issue @Namchee had in this other issue. The query which Mongoose performs is correct, it would return the correct documents from the foreign schema. But it seems like it's not getting merged correctly afterwards...

Problem exists with Mongoose 8.12.0

P.S. I found out, it's only happening when using UUIDs instead of ObjectId.

> All of the cases here refers to `array of IDs` in virtualized-schema to `ID` in reference schema. But, what if I have the opposite case?
> 
> 
> 
> Example:
> 
> 
> 
> Announcement Schema
> 
> ```
> 
> const announcementSchema = new Schema({
> 
>   title: {
> 
>     type: String,
> 
>     trim: true,
> 
>     unique: true,
> 
>     required: true,
> 
>     maxlength: 25,
> 
>   },
> 
> 
> 
>   content: {
> 
>     type: String,
> 
>     trim: true,
> 
>     required: true,
> 
>     maxlength: 500,
> 
>   },
> 
> 
> 
>   validUntil: {
> 
>     type: Date,
> 
>     required: true,
> 
>   },
> 
> 
> 
>   important: {
> 
>     type: Boolean,
> 
>     default: false,
> 
>   },
> 
> 
> 
>   categories: {
> 
>     type: [Schema.Types.ObjectId],
> 
>     ref: 'Category',
> 
>     default: [],
> 
>   },
> 
> });
> 
> ```
> 
> 
> 
> Category Schema
> 
> ```
> 
> const categorySchema = new Schema({
> 
>   name: {
> 
>     type: String,
> 
>     trim: true,
> 
>     required: true,
> 
>     index: true,
> 
>     maxlength: 50,
> 
>   },
> 
> 
> 
>   desc: {
> 
>     type: String,
> 
>     trim: true,
> 
>     required: true,
> 
>     maxlength: 250,
> 
>   },
> 
> });
> 
> 
> 
> categorySchema.virtual('announcementCount', {
> 
>   ref: 'Announcement',
> 
>   localField: 'id',
> 
>   foreignField: 'categories',
> 
>   count: true,
> 
> });
> 
> ```
> 
> 
> 
> I've tried the above solution, but I keep getting `undefined` when I tried to access `announcementCount` after populating it. 

 _Originally posted by @Namchee in [#4585](https://github.com/Automattic/mongoose/issues/4585#issuecomment-569499509)_

vkarpov15:
Fix #15315

**Summary**

A bit of a hack to unblock #15315 and allow virtual populate on array of UUIDs. `String()` on a `Binary` value returns a unicode string that isn't properly indexable when doing `doc = resultDocs[sid];`.

Future improvement: make Mongoose UUID types use BSON UUID class

**Examples**

### DIFF
--- a/test/model.populate.test.js
+++ b/test/model.populate.test.js
@@ -11452,13 +11452,12 @@ describe('model: populate:', function() {
 
     const category = await Category.create({ name: 'Tech', desc: 'Technology News' });
 
-    const announcement = new Announcement({
+    const announcement = await Announcement.create({
       title: 'New Tech Release',
       content: 'Details about the new tech release',
       validUntil: new Date(),
       categories: [category._id]
     });
-    await announcement.save();
 
     const populatedCategory = await Category.findOne({ _id: category._id }).populate('announcements');
     assert.strictEqual(populatedCategory.announcements.length, 1);
--- a/test/model.populate.test.js
+++ b/test/model.populate.test.js
@@ -11450,8 +11450,7 @@ describe('model: populate:', function() {
     const Category = db.model('Category', categorySchema);
     const Announcement = db.model('Announcement', announcementSchema);
 
-    const category = new Category({ name: 'Tech', desc: 'Technology News' });
-    await category.save();
+    const category = await Category.create({ name: 'Tech', desc: 'Technology News' });
 
     const announcement = new Announcement({
       title: 'New Tech Release',
--- a/test/model.populate.test.js
+++ b/test/model.populate.test.js
@@ -7,6 +7,7 @@
 const start = require('./common');
 
 const assert = require('assert');
+const { randomUUID } = require('crypto');
 const utils = require('../lib/utils');
 const util = require('./util');
 const MongooseError = require('../lib/error/mongooseError');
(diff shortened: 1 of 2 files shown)


test/model.populate.test.js
```diff
@@ -11423,4 +11424,45 @@ describe('model: populate:', function() {
...
+      validUntil: { type: Date, required: true },
+      important: { type: Boolean, default: false },
+      categories: [{ type: 'UUID', ref: 'Category' }]
+    });
+
+    const Category = db.model('Category', categorySchema);
+    const Announcement = db.model('Announcement', announcementSchema);
+
+    const category = new Category({ name: 'Tech', desc: 'Technology News' });
+    await category.save();
```
hasezoey, line 11454:
```suggestion
    const category = await Category.create({ name: 'Tech', desc: 'Technology News' });
```

test/model.populate.test.js
```diff
@@ -11423,4 +11424,45 @@ describe('model: populate:', function() {
...
+
+    const announcement = new Announcement({
+      title: 'New Tech Release',
+      content: 'Details about the new tech release',
+      validUntil: new Date(),
+      categories: [category._id]
+    });
+    await announcement.save();
```
hasezoey, line 11462:
```suggestion
    const announcement = await Announcement.create({
      title: 'New Tech Release',
      content: 'Details about the new tech release',
      validUntil: new Date(),
      categories: [category._id]
    });
```

hasezoey (approved):

Issue #11247, opened 2022-01-19
Empty nested schemas are not minimized


aloune2:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Empty nested schemas are not minimized.

**If the current behavior is a bug, please provide the steps to reproduce.**

```js
const { model, Schema } = require('mongoose');

const nestedSchema = new Schema({
  bar: String
}, { _id: false });

const schema = new Schema({
  foo: nestedSchema
});

const MyModel = model('MyModel', schema);

const myModel = await MyModel.create({ foo: {} });

console.log(myModel); // { _id: 61e82a9247bc4b3b51231301, foo: {}, __v: 0 }
````
**What is the expected behavior?**
`foo` is stored in db . It should not and `myModel.foo` should not appears in the result of the `console.log`.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**

Node v16.10.0
Mongoose v5.13.9 or v6.1.7
MongoDB v4.4.7

**Comment by aloune2** [comment]
My mistake

**Comment by IslandRhythms** [comment]
```javascript
const mongoose = require('mongoose');

const nestedSchema = new mongoose.Schema({
  bar: String
}, { _id: false });

const schema = new mongoose.Schema({
  foo: nestedSchema
});

const MyModel = mongoose.model('MyModel', schema);


async function run() {
    await mongoose.connect('mongodb://localhost:27017/test');
    await mongoose.connection.dropDatabase();
    const myModel = await MyModel.create({ foo: {} });

    console.log(myModel); // { _id: 61e82a9247bc4b3b51231301, foo: {}, __v: 0 }
}

run();

```

PR #15336, opened 2025-04-01
feat(subdocument): support schematype-level minimize option to disable minimizing empty subdocuments

### LINKED ISSUE #15313: "minimize: false" is not working for it's own field
> No. It's a critical issue.
> 
> And you have to make a distinction here. “minimize: false” actually worked in mongoose5.x in such a way that it also affected the own object in the database and was not only allowed for all sub-objects. So you could also save the one field where minimize: false was directly as an empty object.
> 
> And secondly, the whole thing has been buggy since at least mongoose 7.x - possibly even 6.x. If the field is set to minimize: false and you try to save a new document (e.g. new MySchema(osbj).save()), then **no empty** object is created.
> 
> If, on the other hand, you save the field for an existing document (obj.save()), the empty object **is saved**! (as it was in mongoose5.x)
> 
> To make matters worse, it seems that you cannot read these empty objects in mongoose 7.x (in mongoose 5.x you can)
> For me, the minimize: false feature has been properly bugged since the use of mongoose 7.x, which currently has to be productively modified unnecessarily by means of a workaround. 

 _Originally posted by @kuehn-sba in [#15177](https://github.com/Automattic/mongoose/issues/15177#issuecomment-2720504403)_

vkarpov15:
Fix #15313

**Summary**

In #11247, we made it so that Mongoose would minimize empty subdocuments by default. However, there isn't a great way to configure that behavior right now: #14151 disabled that behavior for required subdocs, but you might not want to make the field required.

This PR lets you set a `minimize` option on the schematype itself, like `subdoc: { type: Schema, minimize: false }` to disable minimize on the full subdocument. Setting `minimize: false` on the schematype will also override the child schema's `minimize`.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -3796,6 +3796,8 @@ Document.prototype.$toObject = function(options, json) {
   let _minimize;
   if (options._calledWithOptions.minimize != null) {
     _minimize = options.minimize;
+  } else if (this.$__schemaTypeOptions?.minimize != null) {
+    _minimize = this.$__schemaTypeOptions.minimize;
   } else if (defaultOptions != null && defaultOptions.minimize != null) {
     _minimize = defaultOptions.minimize;
   } else {
--- a/lib/helpers/clone.js
+++ b/lib/helpers/clone.js
@@ -57,7 +57,6 @@ function clone(obj, options, isArrayChild) {
         return clonedDoc;
       }
     }
-    const isSingleNested = obj.$isSingleNested;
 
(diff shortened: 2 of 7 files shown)


hasezoey (changes requested):
LGTM, though i think this option still needs to be added to `SchemaSubDocumentOptions` (at least i did not see it in the chain yet)

vkarpov15:
Fixed in https://github.com/Automattic/mongoose/pull/15336/commits/5853a8b4bae983eab21dcc7fdb61b535a17668be

PR #15349, opened 2025-04-10
docs: adds Model#count to list of fns callback removed from

SethFalco:
**Summary**

I'm working on migrating from Mongoose v6 to Mongoose v7. Part of this includes dropping all callbacks in favor of Promises.

I noticed that `Model#count` and `Query#count` is not in the list though, which according to the types, had a callback param in v6, which was removed in v7. This should be in the list of functions as well that the `callback` param was removed from too.

_(Also fixes a petty typo as well, since I was modifying this file anyway.)_

**Examples**

N/A

### DIFF
--- a/docs/migrating_to_7.md
+++ b/docs/migrating_to_7.md
@@ -158,7 +158,7 @@ conn.startSession(function(err, session) {
 // After
 const session = await conn.startSession();
 // Or:
-conn.startSession().then(sesson => { /* ... */ });
+conn.startSession().then(session => { /* ... */ });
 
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Reasonable enough. I guess since `count()` was removed, we can also say that the callback was removed. Thanks :+1:

SethFalco:
Count was removed in v8, not v7! It was just the callback was removed in v7. (I'm still a bit behind on versions, but upgrading 1 major version at a time. ^-^')

PR #15356, opened 2025-04-15
feat: expose a ClientEncryption object on encrypted Models

baileympearson:
**Summary**

With the changes in this PR, users will be able to obtain a ClientEncryption object from encrypted models.  This allows users to use the Key Management API to manage data keys for their encrypted collections.

Additionally, I've added miscellaneous other integration tests that confirm that the encryption implementation works with other mongoose features, namely auto index creation and auto collection creation.


test/encryption/encryption.test.js
```diff
@@ -33,8 +45,8 @@ describe('encryption integration tests', () => {
     const regex = /^(?<key>.*): "(?<value>.*)"$/;
     const variables = Object.fromEntries(file.map((line) => regex.exec(line.trim()).groups).map(({ key, value }) => [key, value]));
 
-    process.env.CRYPT_SHARED_LIB_PATH ??= variables.CRYPT_SHARED_LIB_PATH;
-    process.env.MONGOOSE_TEST_URI ??= variables.MONGODB_URI;
+    process.env.CRYPT_SHARED_LIB_PATH = variables.CRYPT_SHARED_LIB_PATH;
```
baileympearson, line 48: I noticed locally that `MONGOOSE_TEST_URI` is often defined, resulting in the wrong URI being used.

lib/helpers/model/discriminator.js
```diff
@@ -44,10 +44,8 @@ function validateDiscriminatorSchemasForEncryption(parentSchema, childSchema) {
     }
   }
 
-  function* allNestedPaths(schema) {
-    const { paths, singleNestedPaths } = schema;
-    yield* Object.keys(paths);
-    yield* Object.keys(singleNestedPaths);
+  function allNestedPaths(schema) {
```
baileympearson, line 47: Just a follow up from https://github.com/Automattic/mongoose/pull/15320#discussion_r2044581203, which merged before I could make the change.

test/encryption/encryption.test.js
```diff
@@ -1307,4 +1157,302 @@ describe('encryption integration tests', () => {
...
+          const options = model.collection.conn.client.options.autoEncryption;
+          options.tlsOptions = {
+            tlsCAFile: 'some file'
+          };
+          assert.deepEqual(model.clientEncryption()._tlsOptions, {
+            tlsCAFile: 'some file'
+          });
+        });
+
+        it.skip('uses the same credentialProviders', async function() {
```
baileympearson, line 1246: This is a feature of 6.15+ drivers - is this something you plan to upgrade to soon?  If not, I can remove this test. If we do adopt 6.15, then I'll leave this test in and we can unskip it when the time comes.

lib/model.js
```diff
@@ -69,6 +69,7 @@ const minimize = require('./helpers/minimize');
 const MongooseBulkSaveIncompleteError = require('./error/bulkSaveIncompleteError');
 const ObjectExpectedError = require('./error/objectExpected');
 const decorateBulkWriteResult = require('./helpers/model/decorateBulkWriteResult');
+const { ClientEncryption } = require('mongodb');
```
vkarpov15, line 72: Can we move this over to the driver layer? Maybe make `lib/drivers/node-mongodb-native/index.js` export `ClientEncryption`?

We currently avoid importing the Node driver directly anywhere other than `lib/drivers/node-mongodb-native' (exception being `lib/index`) because of Mongoose's browser build.

lib/model.js
```diff
@@ -4880,6 +4881,29 @@ Model.compile = function compile(name, schema, collectionName, connection, base)
   return model;
 };
 
+Model.clientEncryption = function clientEncryption() {
+  /** @type(import('mongodb').MongoClient) */
```
vkarpov15, line 4885: Does this comment do anything? I don't think Mongoose supports this type definition syntax

test/encryption/encryption.test.js
```diff
@@ -33,8 +45,8 @@ describe('encryption integration tests', () => {
     const regex = /^(?<key>.*): "(?<value>.*)"$/;
     const variables = Object.fromEntries(file.map((line) => regex.exec(line.trim()).groups).map(({ key, value }) => [key, value]));
 
-    process.env.CRYPT_SHARED_LIB_PATH ??= variables.CRYPT_SHARED_LIB_PATH;
-    process.env.MONGOOSE_TEST_URI ??= variables.MONGODB_URI;
+    process.env.CRYPT_SHARED_LIB_PATH = variables.CRYPT_SHARED_LIB_PATH;
+    process.env.MONGOOSE_TEST_URI = variables.MONGODB_URI;
```
vkarpov15, line 49: I've had some bad experiences with code overwriting environment variables. Minor detail but can we make this throw an error if MONGOOSE_TEST_URI is specified rather than silently overwriting?

test/encryption/encryption.test.js
vkarpov15: I will ship a  Mongoose release that uses Node driver 6.16 this week.

lib/model.js
```diff
@@ -4880,6 +4881,29 @@ Model.compile = function compile(name, schema, collectionName, connection, base)
...
+
+  const {
+    keyVaultNamespace,
+    keyVaultClient,
+    kmsProviders,
+    credentialProviders,
+    proxyOptions,
+    tlsOptions
+  } = autoEncryptionOptions;
+  return new ClientEncryption(keyVaultClient ?? client,
```
vkarpov15, line 4902: GitHub Copilot makes a good point - is it worthwhile to cache this or is it inexpensive to create multiple ClientEncryption instances?

vkarpov15 (changes requested):
A couple of minor comments, mostly moving `ClientEncryption` into `lib/drivers`.

lib/model.js
baileympearson: I think I've moved the CE export to the Node driver layer.  Let me know if that isn't what you were looking for!

lib/model.js
```diff
@@ -4880,6 +4880,33 @@ Model.compile = function compile(name, schema, collectionName, connection, base)
   return model;
 };
 
+Model.clientEncryption = function clientEncryption() {
+  const ClientEncryption = driver.get().ClientEncryption;
```
vkarpov15, line 4884: You should be able to use `this.base.driver.get()` instead of `require('./driver')` directly, see `Model.bulkWrite()` for an example.

Technically drivers are scoped to Mongoose instances, so you can have multiple Mongoose instances using different drivers in one process.

vkarpov15 (changes requested):

lib/model.js
baileympearson: Didn't realize that - the getter/setters in `driver.js` seemed to be setting a module-scoped variable.  

Fixed!

vkarpov15: Yeah the module scoped variable is the default driver, but individual Mongoose instances can have their own driver which overrides the default.

vkarpov15 (approved):

PR #15357, opened 2025-04-16
Correctly recurse in InferRawDocType

### LINKED ISSUE #14954: Issue with InferRawDocType, not resulting in POJO, but instead as DocumentArray
### Prerequisites

### Mongoose version

8.7.1

### Node.js version

16.x

### MongoDB server version

8.0.0

5.2.2

### Description

Hello. I am trying to migrate from mongoose 7.x to 8.7.1 and I am running into some issues with using the new InferRawDocType. 
Specifically, I get an error when type inferring from objects with "double-nested" schemas, meaning schemas that are nested twice (see example below). The relatively simple example below illustrates how InferRawDocType doesn't work properly and doesn't translate the definitions to simple js arrays or POJOs. But when I omit the "double-nested" type and recreate it, it works fine (see Squad1 type). 

Mongo docs state that using InferRawDocType should result in raw document type like `doc.toObject()`. ([see docs here](https://mongoosejs.com/docs/typescript/schemas.html)), but this doesn't happen as the example illustrates

We also tried sticking with what we had before with InferSchemaType but it gives similar Typescript errors with object not containing DocumentArray props. 

This is a huge pain for us as we want to update to new MongoDB 8.0 and mongoose 8.x, but are stuck with typescript errors when migrating. 

```typescript

import { InferRawDocType, Schema } from "mongoose";

const roleDef = {
};

const tacticDef = {
  roles: [new Schema(roleDef)],
};

type Tactic = InferRawDocType<typeof tacticDef>;

const tacticSchema = new Schema(tacticDef);

const squad = {
  tactic: {
    type: tacticSchema,
  },
};

type Squad1 = Omit<InferRawDocType<typeof squad>, "tactic"> & {
  tactic: Tactic;
};

type Squad2 = InferRawDocType<typeof squad>;

// This code does not give any Typescript errors
const squad1: Squad1 = {
  tactic: {
    roles: [{ name: "name"}],
  },
};

// Code below gives typescript error, saying that roles need to have properties of document array
const squad2: Squad2 = {
  tactic: {
    roles: [{ name: "name" }],
  },
};
```

Thanks

### Steps to Reproduce

Copy and paste the code from the description into any project with Typescript and mongoose 8.7.1 and you should see typescript error on squad2, which shouldn't happen.

### Expected Behavior

Using InferRawDocType, this type should be POJOs or simple js arrays. Not DocumentArray, where props like $isSingleNested, $parent etc. is needed

JavaScriptBach:
**Summary**
Correctly recurse in InferRawDocType. When inferring the types of object arrays and objects, it should call itself, not ObtainDocumentType. The latter has been repurposed to obtain the type of full documents, not lean documents, and will infer DocumentArray instead of raw arrays.

**Examples**

Fixes #14954

### DIFF
--- a/types/inferrawdoctype.d.ts
+++ b/types/inferrawdoctype.d.ts
@@ -83,7 +83,7 @@ declare module 'mongoose' {
               IsItRecordAndNotAny<Item> extends true ?
                 Item extends Record<string, never> ?
                   ObtainRawDocumentPathType<Item, TypeKey>[] :
-                  Array<InferRawDocType<Item>>> :
+                  Array<InferRawDocType<Item>> :
                 ObtainRawDocumentPathType<Item, TypeKey>[]
         >:
         PathValueType extends StringSchemaDefinition ? PathEnumOrString<Options['enum']> :
--- a/types/inferrawdoctype.d.ts
+++ b/types/inferrawdoctype.d.ts
@@ -22,7 +22,7 @@ declare module 'mongoose' {
 
   /**
    * @summary Obtains schema Path type.
-   * @description Obtains Path type by separating path type from other options and calling {@link ResolvePathType}
+   * @description Obtains Path type by separating path type from other options and calling {@link ResolveRawPathType}
    * @param {PathValueType} PathValueType Document definition path type.
    * @param {TypeKey} TypeKey A generic refers to document definition.
    */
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1: merging into 8.14 branch

PR #15360, opened 2025-04-17
docs: update deleteOne & deleteMany API def

Elliot67:
**Summary**

The returned object of `deleteOne` & `deleteMany` comes directly from mongodb driver. 
The API definition was not updated.


vkarpov15 (approved):
Thanks!

PR #15366, opened 2025-04-21
ci(NODE-6919): Use `mongodb-runner` in encryption cluster setup

baileympearson:
**Summary**

This PR re-writes the existing encryption test setup tooling to use https://www.npmjs.com/package/mongodb-runner and https://www.npmjs.com/package/@mongodb-js/mongodb-downloader.  

These packages are tooling packages maintained by teams at MongoDB for use in testing.  Use of `npm` packages integrates more seamlessly into Mongoose's existing tooling.


test/encryption/encryption.test.js
```diff
@@ -22,43 +33,40 @@ function isEncryptedValue(object, property) {
...
-
-    // matches `key="value"` and extracts key and value.
-    const regex = /^(?<key>.*): "(?<value>.*)"$/;
-    const variables = Object.fromEntries(file.map((line) => regex.exec(line.trim()).groups).map(({ key, value }) => [key, value]));
+  /** @type { string } */
+  let cryptSharedLibPath;
+  /** @type { string } */
+  let clusterUri;
+  before(async function() {
+    const expansionFile = join(__dirname, '../..', 'fle-cluster-config.json');
```
baileympearson, line 41: Because we now use `npm` packages to launch clusters, it would be straightforward to setup and teardown the clusters as a part of the test suite.  I considered adding this and decided against it, just because the existing usage pattern works well and means you don't need to setup/teardown a cluster per test run locally, but I can add this functionality if it is desirable.

vkarpov15 (commented):
This looks great! Can you please merge in the latest changes from csfle branch so we can confirm the tests run correctly? Right now the tests aren't running in this PR because GitHub Actions dropped support for Ubuntu 20.04.

baileympearson:
@vkarpov15 Done.  I actually noticed the FLE tests weren't running, because they only run on PR to master, which I've fixed.  But they pass now 🙂

vkarpov15 (approved):
Good catch, thanks :+1:

PR #15373, opened 2025-04-25
8.14

vkarpov15:
**Summary**

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -3096,13 +3096,12 @@ function _handleSortValue(val, key) {
  *
  *     await Character.deleteOne({ name: 'Eddard Stark' });
  *
- * This function calls the MongoDB driver's [`Collection#deleteOne()` function](https://mongodb.github.io/node-mongodb-native/4.9/classes/Collection.html#deleteOne).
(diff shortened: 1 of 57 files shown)


hasezoey (approved):
LGTM

lib/query.js
```diff
@@ -4429,10 +4521,12 @@ Query.prototype.exec = async function exec(op) {
...
     }
     const err = new MongooseError('Query was already executed: ' + str);
-    err.originalStack = this._executionStack;
+    if (!this.model.base.options.skipOriginalStackTraces) {
+      err.originalStack = this._executionStack;
+    }
     throw err;
   } else {
-    this._executionStack = new Error().stack;
+    this._executionStack = this.model.base.options.skipOriginalStackTraces ? true : new Error().stack;
```
AbdelrahmanHafez, line 4529: Something about this line doesn't feel right to me. In one scenario we're assigning a boolean, in another scenario we're assigning a string. Is this right?

AbdelrahmanHafez (approved):
One minor comment, otherwise LGTM 👌

lib/query.js
vkarpov15: Yes it is expected unfortunately. The plan is to rename `_executionStack` and make it a boolean in 9.0. But for now this is a workaround to let people avoid `new Error().stack`, which can be expensive for apps that use source maps.

PR #15375, opened 2025-04-27
Async stack traces for common function calls, remove callbacks for `execPost()`.

vkarpov15:
**Summary**

Re: latest work in vkarpov15/kareem#39, remove callbacks from `execPost()` and rely on async functions instead. I also added a bunch of tests to confirm that we're now getting async stack traces for all common operations (asserting that error stack trace includes the test function name). And did some cleanup in `lib/model` to move `$__handleSave()` into `$__save()`.

**Examples**


lib/browserDocument.js
```diff
@@ -106,14 +106,7 @@ Document.prototype._execDocumentPreHooks = async function _execDocumentPreHooks(
  */
 
 Document.prototype._execDocumentPostHooks = function _execDocumentPostHooks(opName, error) {
```
hasezoey, line 108:
```suggestion
Document.prototype._execDocumentPostHooks = async function _execDocumentPostHooks(opName, error) {
```

lib/document.js
```diff
@@ -2911,14 +2911,7 @@ Document.prototype._execDocumentPreHooks = async function _execDocumentPreHooks(
  */
 
 Document.prototype._execDocumentPostHooks = function _execDocumentPostHooks(opName, error) {
```
hasezoey, line 2913:
```suggestion
Document.prototype._execDocumentPostHooks = async function _execDocumentPostHooks(opName, error) {
```

lib/model.js
```diff
@@ -3614,21 +3474,12 @@ async function buildPreSavePromise(document, options) {
 }
 
 function handleSuccessfulWrite(document) {
```
hasezoey, line 3476:
```suggestion
async function handleSuccessfulWrite(document) {
```

hasezoey (approved):
Looks good to me, some minor functions should likely be marked `async`.

PR #15378, opened 2025-04-28
BREAKING CHANGE: make UUID schema type return bson UUIDs

vkarpov15:
**Summary**

Currently, `type: 'UUID'` makes Mongoose internally store Mongoose Buffers, which get converted to strings via a built-in getter. This is inconsistent for a few reasons:

1. We have a separate `mongoose.Types.UUID` type, but `mongoose.Schema.Types.UUID` paths don't cast into `mongoose.Types.UUID` instances
2. Because UUID paths are stored as buffers and converted to string via getters, `lean()` means UUIDs are returned as instances of `mongodb.Binary` by default (unless mongoose-lean-getters plugin is enabled), so you end up having to handle 3 different types when working with UUIDs (Binary if lean, Buffer if toObject() with getters false or bypassing getters, string otherwise). Also adds some confusion for TypeScript. 
3. TypeScript types convert 'UUID' -> 'Buffer', which isn't correct by default, should be string.

Additional benefit: MongoDB `UUID` class has a `toJSON()` that always serializes the UUID as a hex string, which seems like an improvement. Currently UUIDs end up as `{ type: 'buffer', data: number[] }` in JSON if `getters: false` and `string` if `getters: true`.

**Examples**

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -32,6 +32,7 @@ declare module 'mongoose' {
   import events = require('events');
   import mongodb = require('mongodb');
   import mongoose = require('mongoose');
+  import bson = require('bson');
 
   export type Mongoose = typeof mongoose;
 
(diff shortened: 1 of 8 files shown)


types/index.d.ts
```diff
@@ -766,6 +767,25 @@ declare module 'mongoose' {
                     : BufferToBinary<T[K]>;
           } : T;
 
+    /**
+    * Converts any Buffer properties into { type: 'buffer', data: [1, 2, 3] } format for JSON serialization
```
hasezoey, line 771:
```suggestion
    * Converts any Buffer properties into "{ type: 'buffer', data: [1, 2, 3] }" format for JSON serialization
```
(or basically any other delimiter)

hasezoey (approved):
LGTM

Issue #15154, opened 2025-01-03
Use mongodb exported bson instead of mongoose to import bson separately?


zurmokeeper:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

When using chrome's memory to do a snapshot analysis of node.js references, I found that the bson library takes up a lot of memory, about 3-4M (estimated), especially if the project may reference multiple bson libraries.

I see that mongodb already has an API for exporting bson, so why does mongoose have to reference bson separately?

If I use mongodb's bson, then I can reference a mongoose, and I don't need to bring in mongodb and its dependent bson libraries.

It would be more memory efficient.

vkarpov15:
Mongoose imports bson directly rather than going through `mongodb.BSON` is for the browser build: we haven't tried recently, but importing the MongoDB driver makes it impossible for Mongoose to run in the browser.

We do try to keep the exact same version of bson as the MongoDB driver, so most package managers should automatically share the bson version between Mongoose and MongoDB.

Can you please run `npm list bson` and paste the output here? That would be helpful.

vkarpov15:
Option (1) is tricky because Webpack will pull in whatever we `require()` in, so it isn't as simple as `if (typeof window !== 'undefined') { require('bson') } else { require('mongodb') }` . And even a single import of `mongodb` anywhere in the browser fails with `Uncaught TypeError: can't convert undefined to object` because of an issue in whatwg-url. Even if we got mongodb working in the browser, the size difference between Mongoose's browser build with mongodb vs with just bson is 3MB vs 800KB.

![Image](https://github.com/user-attachments/assets/fca5565b-a56a-4466-864e-9c82a3e84a77)


What we could do is move all type imports into the driver layer, so `lib/drivers/browser/objectid` exports bson.ObjectId and `lib/drivers/node-mongodb-native/objectid` exports mongodb.ObjectId. This is a possibility, but would require substantial refactor because types would have to be scoped to a Mongoose instance, which means helpers that use types, like `lib/helpers/clone`, would also have to be scoped to a Mongoose instance. We've experimented with this course of action, but it's a long road.

In the meantime, have you considered just using `require('mongoose').mongo` instead of including mongodb as a top-level dependency? I know that's a pretty similar suggestion to yours, but that should knock out at least one bson instance.

zurmokeeper:
@vkarpov15 

[...]

PR #15385, opened 2025-04-29
BREAKING CHANGE: remove browser build, move to @mongoosejs/browser instead

### LINKED ISSUE #15296: Move Mongoose browser build to a separate plugin
### Prerequisites

### Issue

Re: discussion from #15277.

The browser build has increasingly become a source of issues with bundlers and edge runtimes. Not to mention the additional weight on our npm package bundle. We should consider splitting browser build and browser driver off into a separate repo like `@mongoosejs/browser`, etc.

vkarpov15:
Fix #15296

**Summary**

Moved the browser build and browser driver into this separate GitHub repo: https://github.com/mongoosejs/mongoose-browser. This lets us delete a lot of code and remove a lot of dependencies.

Outside of removing the browser lib, there's a couple of other noteworthy changes:

1. Changed `require('node:async_hooks')` to `require('async_hooks')` as a slight workaround for the browser lib. Webpack complains about not being able to handle `node:` imports.
2. Documents now have a `this.$__middleware` that is set in `$__setSchema` that we use for executing document middleware. Previously, we relied on `applyHooks()` to set `_middleware` on the `Model`, which would contain a filtered copy of `schema.s.hooks` that just contained document middleware. However, that caused trouble for the browser lib, and caused us trouble in some other places as well. The current approach of setting the filtered document middleware on the document (or document prototype) itself when `$__setSchema()` is called seems more robust because that ensures that a document has a `$__middleware` whenever it has a schema.

Still todo:

[ ] Move `browser.md` docs to mongoose-browser and make `browser.md` link out.

**Examples**

### DIFF
--- a/docs/browser.md
+++ b/docs/browser.md
@@ -1,37 +1,4 @@
 # Mongoose in the Browser
 
-Mongoose supports creating schemas and validating documents in the browser.
-Mongoose's browser library does **not** support saving documents, [queries](http://mongoosejs.com/docs/queries.html), [populate](http://mongoosejs.com/docs/populate.html), [discriminators](http://mongoosejs.com/docs/discriminators.html), or any other Mongoose feature other than schemas and validating documents.
-
-Mongoose has a pre-built bundle of the browser library. If you're bundling your code with [Webpack](https://webpack.js.org/), you should be able to import Mongoose's browser library as shown below if your Webpack `target` is `'web'`:
-
-```javascript
-import mongoose from 'mongoose';
-```
-
-You can use the below syntax to access the Mongoose browser library from Node.js:
-
-```javascript
-// Using `require()`
-const mongoose = require('mongoose/browser');
-
-// Using ES6 imports
-import mongoose from 'mongoose/browser';
-```
-
-## Using the Browser Library {#usage}
-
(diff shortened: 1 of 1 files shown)


vkarpov15:
Also worth noting there will be some followup work to this PR necessary for #15154.

hasezoey (approved):
Looks good to me, though this should likely also be noted in the migration guide.

Some other questions though:
- Do i understand correctly that the browser-build plugin is basically now just a different driver instead of having extra stuff like a `browserDocument`?
- Will the new documentation be on `https://mongoosejs.com/docs/` or as other things in `https://plugins.mongoosejs.io/`?

vkarpov15:
1. Yes, mongoose-browser is now structured as a separate driver with minimal support from Mongoose's core. There is slightly more than just a driver though because there is also a `dist/mongoose.umd.js` file that contains a compiled version of Mongoose with the driver, so you can either import Mongoose with a script tag or import Mongoose without having to compile it. 
2. For now, I plan on moving the browser docs to the mongoose-browser GitHub repo's README. Depending on demand, we may create a separate website for mongoose-browser docs, but right now we don't have any idea of how much the Mongoose browser build is used.

PR #15387, opened 2025-04-30
Eslint 9.0.0 & migrate to eslint flat configs

hasezoey:
**Summary**

This PR does a few eslint related things, mainly upgrading to eslint 9.0.0 and migrating the configs to the new "Flat Config" format. ~~Additionally, i have also migrated from `eslint-plugin-mocha-no-only` to `eslint-plugin-mocha`, please let met know if there are problems with this new style from the plugin change.~~
Due to those big changes, there are some more minor changes to list out:
- ~~renamed deno scripts to be `.mjs`, which makes it easier to configure eslint~~
- ~~removed some 100% duplicate tests~~
- ~~removed some disabled tests for removed functionality~~
- ~~removed checks for mongodb 2.x as since mongoose 5.x, [mongodb 3.0.0 or higher is required](https://mongoosejs.com/docs/compatibility.html)~~

Depends on #15385 to reduce duplication, as otherwise eslint fails with wrong `globals` versions (this PR is a DRAFT until that is merged)

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "devDependencies": {
     "typescript-eslint": "^8.31.1",
-    "acquit": "git+https://github.com/hasezoey/acquit.git#0e98e3292212dae9f25c889fb3a46f3d6688ca55",
+    "acquit": "1.4.0",
     "acquit-ignore": "0.2.1",
     "acquit-require": "0.1.1",
     "ajv": "8.17.1",


hasezoey:
Note CI `Test Generating Docs` currently fails because of `acquit`'s `esprima` not supporting ES2019's "optional catch parameter".
But https://github.com/vkarpov15/acquit/pull/32 works fine.

hasezoey:
Now that i look at this PR again, i think i should split it into 2 other (for a total of 3) PRs:
- one that can be already applied to master
- this one, upgrading & migrating eslint
- switch to `eslint-plugin-mocha`

Note that the latter 2 will happen(or be rebased) once the first one and the dependent PR (#15385) are merged into 9.0.

hasezoey:
Updated the PR to now only contain the upgrade to ESLint 9.0.0 and migration to eslint flat configs as stated earlier.

Because it is necessary or quick:
- fixed a typo in the browser docs
- changed the `acquit` dependency to use the `espree` PR to fix the "invalid syntax" errors in website generation, https://github.com/vkarpov15/acquit/pull/32 (this should be handled similar to the kareem dependency)

Alternatively, we could also just disable the new "recommended" rule to revert to the old behavior, if that is wanted over updating acquit.

vkarpov15 (approved):
Thanks :+1: I updated to acquit 1.4.0

PR #15388, opened 2025-04-30
fix(query): handle casting array filter paths underneath array filter paths with embedded discriminators

### LINKED ISSUE #15386: Arrayfilters does not work with nested array with descriminator
### Prerequisites

### Mongoose version

8.14.1

### Node.js version

20.18.0

### MongoDB server version

5

_No response_

### Description

If you have a array of documents on a discriminated schema array. mongoose fails to find the correct schema when casting array filters

### Steps to Reproduce

The following throws error `Could not find path "events.0.products.0._id" in schema`

```
  const eventSchema = new Schema(
    { message: String },
    { discriminatorKey: 'kind', _id: false },
  );

  const batchSchema = new Schema({ events: [eventSchema] });

  // `batchSchema.path('events')` gets the mongoose `DocumentArray`
  // For TypeScript, use `schema.path<Schema.Types.DocumentArray>('events')`
  const docArray = batchSchema.path<Schema.Types.DocumentArray>('events');

  // The `events` array can contain 2 different types of events, a
  // 'clicked' event that requires an element id that was clicked...
  const clickedSchema = new Schema(
    {
      element: {
        type: String,
        required: true,
      },
    },
    { _id: false },
  );
  // Make sure to attach any hooks to `eventSchema` and `clickedSchema`
  // **before** calling `discriminator()`.
  docArray.discriminator('Clicked', clickedSchema);

  const productSchema = new Schema({
    name: { type: String },
    price: { type: Number },
  });

  // ... and a 'purchased' event that requires the product that was purchased.
  docArray.discriminator(
    'Purchased',
    new Schema({
      products: {
        type: [productSchema],
        required: true,
      },
    }),
  );

  const Batch = db.model('EventBatch', batchSchema);

  const purchasedId = new mongoose.Types.ObjectId();
  const productId = new mongoose.Types.ObjectId();

  const doc = await Batch.create({
    events: [
      { kind: 'Clicked', element: '#hero', message: 'hello' },
      {
        kind: 'Purchased',
        _id: purchasedId,
        products: [{ name: 'productName1', price: 10, _id: productId }],
        message: 'world',
      },
    ],
  });

  console.log('Result', doc);

  const updatedBatch = await Batch.findByIdAndUpdate(
    doc._id,
    {
      'events.$[event].products.$[product].price': 20,
    },
    {
      arrayFilters: [
        {
          'event._id': purchasedId,
          'event.kind': 'Purchased', // Does not work with or without this
        },
        { 'product._id': productId },
      ],
      new: true,
    },
  ).exec();
```
### Expected Behavior

Should be able to use arrayFilters to update an array on nested array of a discriminated schema

vkarpov15:
Fix #15386

**Summary**

The issue in #15386 is that `events.$[event].products.$[product].price` updates don't quite work if `products` is a path on an embedded discriminator where the discriminator key is set in the `event` array filter.

To get that to work, we needed to first fix `castArrayFilters()` to handle this pattern when casting array filters, and then improve how we handle casting the update in `castUpdate()` (`getEmbeddedDiscriminatorPath()` changes).

**Examples**

### DIFF
--- a/lib/helpers/query/getEmbeddedDiscriminatorPath.js
+++ b/lib/helpers/query/getEmbeddedDiscriminatorPath.js
@@ -17,7 +17,7 @@ const updatedPathsByArrayFilter = require('../update/updatedPathsByArrayFilter')
  */
 
 module.exports = function getEmbeddedDiscriminatorPath(schema, update, filter, path, options) {
-  const parts = path.split('.');
+  const parts = path.indexOf('.') === -1 ? [path] : path.split('.');
   let schematype = null;
   let type = 'adhocOrUndefined';
 
@@ -26,9 +26,10 @@ module.exports = function getEmbeddedDiscriminatorPath(schema, update, filter, p
   const arrayFilters = options != null && Array.isArray(options.arrayFilters) ?
     options.arrayFilters : [];
   const updatedPathsByFilter = updatedPathsByArrayFilter(update);
+  let startIndex = 0;
 
   for (let i = 0; i < parts.length; ++i) {
-    const originalSubpath = parts.slice(0, i + 1).join('.');
+    const originalSubpath = parts.slice(startIndex, i + 1).join('.');
     const subpath = cleanPositionalOperators(originalSubpath);
     schematype = schema.path(subpath);
     if (schematype == null) {
@@ -89,6 +90,8 @@ module.exports = function getEmbeddedDiscriminatorPath(schema, update, filter, p
 
       const rest = parts.slice(i + 1).join('.');
       schematype = discriminatorSchema.path(rest);
+      schema = discriminatorSchema;
+      startIndex = i + 1;
       if (schematype != null) {
         type = discriminatorSchema._getPathType(rest);
         break;
--- a/lib/helpers/schema/getPath.js
+++ b/lib/helpers/schema/getPath.js
@@ -8,7 +8,7 @@ const numberRE = /^\d+$/;
  * @api private
  */
 
-module.exports = function getPath(schema, path) {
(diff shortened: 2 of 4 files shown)


lib/helpers/query/getEmbeddedDiscriminatorPath.js
```diff
@@ -17,7 +17,7 @@ const updatedPathsByArrayFilter = require('../update/updatedPathsByArrayFilter')
  */
 
 module.exports = function getEmbeddedDiscriminatorPath(schema, update, filter, path, options) {
-  const parts = path.split('.');
+  const parts = path.indexOf('.') === -1 ? [path] : path.split('.');
```
hasezoey, line 20: Is this a performance optimization? Because i think `.split()` will always return a array with a value, even if `separator` is not found.

lib/helpers/schema/getPath.js
```diff
@@ -26,10 +26,14 @@ module.exports = function getPath(schema, path) {
     schematype = schema.path(cur);
     if (schematype != null && schematype.schema) {
       schema = schematype.schema;
+      const currentPath = cur;
       cur = '';
```
hasezoey, line 30: Instead of assigning a new value, how about just moving the `cur = ''` reassignment to the end of the block?

hasezoey (commented):
LGTM, some minor nitpicks.

lib/helpers/schema/getPath.js
vkarpov15: Fixed in b91db2b, thanks for the suggestion :+1:

PR #15404, opened 2025-05-05
chore: address comments on CSFLE feature branch review

baileympearson:
This PR addresses comments from @hasezoey 's review of the CSFLE feature branch.  I believe I've addressed every comment.  Unfortunately, we'll lose some context for the few comments that I replied to - I'm not sure the best way to keep track of these discussions.

**Summary**

**Examples**


baileympearson:
RE https://github.com/Automattic/mongoose/pull/15390#pullrequestreview-2811530320

> This is one big PR, it mostly looks good to me, but there are some minor style nitpicks and general suggestions / questions.
> 
> Additional question not related to a specific file & line:
> 
> 1.I did not see SchemaType UUID get a autoEncryptionType or any tests as the field type, is UUID even possible to be used as such?
> 2.Should there be tests for encrypting a whole subdocument which contains types that dont have autoEncryptionType(eg Mixed)?
> 3. Should indexes on encrypted fields be handled or tested (though it does not seem to logical to me)?
> Also finally, shouldnt this PR target branch 8.15?

1. Yes - good catch, that was an oversight.  UUIDs are serialized as BSON binary objects, just like Buffers.  I've added tests and an encryption type to UUID's schemaType.
2. Can you elaborate on this?  I don't understand this scenario.  Mixed isn't supported as an encryption type.
3. I'm not sure this is necessary, but I can added tests if you want.  Manually creating indexes for queryable encryption isn't necessary, and indexes are only supported for deterministically encrypted CSFLE fields.  The index creation process shouldn't be different than for fields without CSFLE.  Let me know which you'd prefer.

As far as targeting 8.15 - according to the PR description, it probably should.  Either you or Val will need to change that though, I don't think I have permissions because it isn't my branch 🙂

vkarpov15 (approved):

PR #15408, opened 2025-05-07
C: add Node 24

stscoundrel:
Add Node 24 to version matrix, as it is the "current" and upcoming LTS version.

Required small change to two test cases:

The prevValue tests used "nulls" in expected values. This comparison fails in Node 24, as the actual value is "undefined".

However, the original value seems to be "undefined" in previous Node versions too. Therefore it does not seem like a change in behaviour, but only a change in how strict the assert is. Changing the expected value to "undefined" works in both Node 24 and previous Node versions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #15414, opened 2025-05-09
types: fix signature of DocumentArray.id

Sainan:
My main issue with this was that 'undefined' was a valid argument. Now it matches <https://mongoosejs.com/docs/api/documentarray.html#MongooseDocumentArray.prototype.id()>.


vkarpov15 (approved):
LGTM. In theory this PR wouldn't be correct, but right now `.id()` hard codes casting to ObjectId so you can only pass in ObjectIds or things that can cast to ObjectId

Issue #13907, opened 2023-09-29
When saving a model with bulkSave, the values in the discriminator schema are not updated.


0skgc:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.5.3

### Node.js version

18.16.0

### MongoDB server version

6.0.10

### Typescript version (if applicable)

5.2.2

### Description

When saving a model with bulkSave, the properties of the discriminator schema are not updated.
When I try the same update with model.prototype.save, it is updated, so is this a bug with bulkSave?

I followed the internal process of bulkSave in the debugger, and it looks like updates to paths that are not in the base schema (updates to the discriminator schema) are deleted in the walkUpdatePath function in castUpdate.js.

### Steps to Reproduce

Test code that can reproduce the problem
```
import mongoose from "mongoose"
import { beforeEach, test, expect } from "vitest"

const schema = new mongoose.Schema({ value: String }, { discriminatorKey: "type" })
const typeASchema = new mongoose.Schema({ aValue: String })
schema.discriminator("A", typeASchema)

type IBaseTest = mongoose.InferSchemaType<typeof schema>
type ITypeATest = IBaseTest & { type: "A" } & mongoose.InferSchemaType<typeof typeASchema>
type ITest = ITypeATest

const TestModel = mongoose.model<ITest>("Test", schema)

describe("mongoose buksSave test", () => {
  beforeEach(async () => {
    await TestModel.deleteMany({})
  })

  const testData: ITypeATest = { value: "initValue", type: "A", aValue: "initAValue" }

  test("bulkSave", async () => {
    // Arrange
    const doc = await TestModel.create<ITypeATest>(testData)

    // Act
    doc.value = "updatedValue1"
    doc.aValue = "updatedValue2"
    await TestModel.bulkSave([doc])

    // Assert (!!!aValue is not updated)
    const findDoc = await TestModel.findById(doc._id)
    expect(findDoc).toMatchInlineSnapshot(`
      {
        "__v": 0,
        "_id": "${doc._id.toString()}",
        "aValue": "initAValue",
        "type": "A",
        "value": "updatedValue1",
      }
    `)
  })

  test("save", async () => {
    // Arrange
    const doc = await TestModel.create<ITypeATest>(testData)

    // Act
    doc.value = "updatedValue1"
    doc.aValue = "updatedValue2"
    await doc.save()

    // Assert (both value and aValue are updated)
[...]

PR #15415, opened 2025-05-09
fix(model): make bulkSave() rely on document.validate() to validate docs and skip bulkWrite casting

### LINKED ISSUE #15410: `bulkSave` and `bulkWrite` ignore changes to fields belonging to subclasses of subdocuments
### Prerequisites

### Mongoose version

8.14.1

### Node.js version

20.10.0

### MongoDB server version

6.0.2

5.8.3

### Description

My model `Order` has a complex structure of embedded documents and subclasses: it has a first field called `rows` which is an embedded array, containing another field `rows` which is another embedded array, and finally inside this last `rows` there is `requirements`, an embedded array that is extended by two subclasses, `componentRequirement` and `toolRequirement`, which specify additional fields beyond those in `requirement`.

The problem I encountered is that if I modify a field of one of the subclasses, for example `order.rows.rows.requirements.set({ toolTest: new ObjectId() });`, and then run `Order.bulkSave(order)` or `Order.bulkWrite({ updateOne: { .... } })`, the change is not taken into account. However, I believe I have configured everything correctly, because if I run `order.save()`, everything works as expected.

`Order` model:
```ts
interface IOrder {
  _id: Types.ObjectId;
  code: string;
  rows: IOrderRow[];
  createdAt: Date;
  updatedAt: Date;
}
```

`OrderRow` subdocument:
```ts
interface IOrderRow {
  rows: ISubRow[]
}
```

`SubRow` subdocument and requirements configuration:
```ts
interface ISubRow {
  requirements: IRequirement[];
}

const requirements = schema.path<Schema.Types.DocumentArray>('requirements');
requirements.discriminator<IComponentRequirement, ComponentRequirementModelType>('ComponentRequirement', componentRequirementSchema);
requirements.discriminator<IToolRequirement, ToolRequirementModelType>('ToolRequirement', toolRequirementSchema);
```

`Requirement` subdocument:
```ts
interface IRequirement {
  kind: 'ComponentRequirement' | 'ToolRequirement';
  quantity: number;
  notes?: string;
}
```

`ComponentRequirement` subclass:
```ts
interface IComponentRequirement extends IRequirement {
  kind: 'ComponentRequirement';
  componentTest?: Types.ObjectId;
}
```

`ToolRequirement` subclass:
```ts
interface IToolRequirement extends IRequirement {
  kind: 'ToolRequirement';
  toolTest?: Types.ObjectId;
}
```

Repro script:
```ts
async function checkValue(
  Order: ReturnType<typeof registerOrderModel>,
  orderId: Types.ObjectId,
  newObjectId: Types.ObjectId,
) {
  const updatedOrder = await Order.findById(orderId).orFail();
  const updatedOrderValue = (updatedOrder.rows[0].rows[0].requirements[1] as ToolRequirementHydratedDocument).toolTest;
  assert(updatedOrderValue?.toString() === newObjectId.toString(), updatedOrder.code);
}

async function run() {
  const connection = await getConnection();

  const Order = registerOrderModel(connection);

  const [
    order1,
    order2,
    order3,
  ] = await Order.create(
    [1, 2, 3].map((index) => ({
      code: `test-${index}`,
      rows: [{
        rows: [{
          requirements: [
            {
              kind: 'ComponentRequirement',
              quantity: 1,
            },
            {
              kind: 'ToolRequirement',
              quantity: 1,
            },
          ]
        }]
      }],
    }))
  );

  const newObjectId = new Types.ObjectId();

  // ORDER 1
  order1.rows[0].rows[0].requirements[1].set({
    toolTest: newObjectId,
  });
  await order1.save();
  await checkValue(Order, order1._id, newObjectId);

  // ORDER 2
  order2.rows[0].rows[0].requirements[1].set({
    toolTest: newObjectId,
  });
  await Order.bulkSave([order2]);
  await checkValue(Order, order2._id, newObjectId);

  // ORDER 3
  await Order.bulkWrite([{
    updateOne: {
      filter: {
        _id: order3._id,
      },
      update: {
        $set: {
          'rows.0.rows.0.requirements.1.toolTest': newObjectId,
        }
      }
    }
  }]);
  await checkValue(Order, order3._id, newObjectId);

  await connection.close();
}

run();
```

Execution:

Image

### Steps to Reproduce

Reproduction link [here](https://stackblitz.com/edit/vitejs-vite-rp25amta?file=src%2Findex.ts).

### Expected Behavior

I expect the modification to be successfully applied in all three cases: `save`, `bulkSave`, and `bulkWrite`.

vkarpov15:
Fix #15410

**Summary**

@AbdelrahmanHafez please take a look at this one

The issue in #15410 is similar to the issue in #13907: `bulkSave()` currently relies on `bulkWrite()`'s `castBulkWrite()` logic to handle casting, which is inconsistent with `save()`. Further, it causes problems with discriminators (like in #13907  for top-level discriminators and #15410 for embedded discriminators) because you need to set the discriminator key on discriminator paths to tell `castBulkWrite()` which discriminator schema to use when casting.

We worked around the discriminator issue for top-level discriminators in #13907, but for embedded discriminators it is infeasible - needing to find every embedded discriminator in the doc will be complex and error prone. In the interest of simplifying things, we can just rely on `document.validate()` to ensure the document is valid before saving, which is consistent with how `save()` works. However, this change may also break user code because now async validators and `validate` middleware will run on `bulkSave()`.

Questions to consider:

1. Should we skip running `validate` middleware on `bulkSave()` for backwards compatibility?
2. Should we continue to use `validateSync()` instead of `validate()` for backwards compatibility?

**Examples**

### DIFF
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -6780,7 +6780,7 @@ describe('Model', function() {
       await users[2].save();
       users[2].name = 'I am the updated third name';
 
-      const writeOperations = User.buildBulkWriteOperations(users);
+      const writeOperations = await User.buildBulkWriteOperations(users);
 
       const desiredWriteOperations = [
         { insertOne: { document: users[0] } },
@@ -6795,7 +6795,7 @@ describe('Model', function() {
 
     });
 
-    it('throws an error when one document is invalid', () => {
+    it('throws an error when one document is invalid', async() => {
       const userSchema = new Schema({
         name: { type: String, minLength: 5 }
       });
@@ -6810,11 +6810,10 @@ describe('Model', function() {
 
       let err;
       try {
-        User.buildBulkWriteOperations(users);
+        await User.buildBulkWriteOperations(users);
       } catch (error) {
         err = error;
       }
-
 
       assert.ok(err);
     });
@@ -6827,10 +6826,8 @@ describe('Model', function() {
       const User = db.model('User', userSchema);
 
 
-      assert.throws(
-        function() {
-          User.buildBulkWriteOperations(null);
-        },
+      assert.rejects(
+        User.buildBulkWriteOperations(null),
         /bulkSave expects an array of documents to be passed/
       );
     });
@@ -6842,17 +6839,15 @@ describe('Model', function() {
       const User = db.model('User', userSchema);
 
 
-      assert.throws(
-        function() {
-          User.buildBulkWriteOperations([
-            new User({ name: 'Hafez' }),
-            { name: 'I am not a document' }
-          ]);
-        },
+      assert.rejects(
+        User.buildBulkWriteOperations([
+          new User({ name: 'Hafez' }),
+          { name: 'I am not a document' }
+        ]),
         /documents\.1 was not a mongoose document/
       );
     });
-    it('skips validation when given `skipValidation` true', () => {
+    it('skips validation when given `skipValidation` true', async() => {
       const userSchema = new Schema({
         name: { type: String, minLength: 5 }
       });
@@ -6865,7 +6860,7 @@ describe('Model', function() {
         new User({ name: 'b' })
       ];
 
-      const writeOperations = User.buildBulkWriteOperations(users, { skipValidation: true });
+      const writeOperations = await User.buildBulkWriteOperations(users, { skipValidation: true });
 
       assert.equal(writeOperations.length, 3);
     });
@@ -6904,7 +6899,7 @@ describe('Model', function() {
       userToUpdate.name = 'John Doe';
 
       // Act
-      const writeOperations = User.buildBulkWriteOperations([newUser, userToUpdate], { timestamps: false, skipValidation: true });
+      const writeOperations = await User.buildBulkWriteOperations([newUser, userToUpdate], { timestamps: false, skipValidation: true });
 
       // Assert
       const timestampsOptions = writeOperations.map(writeOperationContainer => {
@@ -6926,7 +6921,7 @@ describe('Model', function() {
       userToUpdate.name = 'John Doe';
 
       // Act
-      const writeOperations = User.buildBulkWriteOperations([newUser, userToUpdate], { timestamps: true, skipValidation: true });
+      const writeOperations = await User.buildBulkWriteOperations([newUser, userToUpdate], { timestamps: true, skipValidation: true });
 
       // Assert
       const timestampsOptions = writeOperations.map(writeOperationContainer => {
@@ -6948,7 +6943,7 @@ describe('Model', function() {
       userToUpdate.name = 'John Doe';
 
       // Act
-      const writeOperations = User.buildBulkWriteOperations([newUser, userToUpdate], { skipValidation: true });
+      const writeOperations = await User.buildBulkWriteOperations([newUser, userToUpdate], { skipValidation: true });
 
       // Assert
       const timestampsOptions = writeOperations.map(writeOperationContainer => {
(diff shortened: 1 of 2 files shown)


test/model.test.js
```diff
@@ -6827,10 +6826,8 @@ describe('Model', function() {
       const User = db.model('User', userSchema);
 
 
-      assert.throws(
-        function() {
-          User.buildBulkWriteOperations(null);
-        },
+      assert.rejects(
+        User.buildBulkWriteOperations(null),
```
hasezoey, line 6830:
```suggestion
      await assert.rejects(
        User.buildBulkWriteOperations(null),
```

test/model.test.js
```diff
@@ -6842,17 +6839,15 @@ describe('Model', function() {
...
 
-      assert.throws(
-        function() {
-          User.buildBulkWriteOperations([
-            new User({ name: 'Hafez' }),
-            { name: 'I am not a document' }
-          ]);
-        },
+      assert.rejects(
+        User.buildBulkWriteOperations([
```
hasezoey, line 6843:
```suggestion
      await assert.rejects(
        User.buildBulkWriteOperations([
```

test/model.test.js
```diff
@@ -7062,6 +7057,63 @@ describe('Model', function() {
...
+
+      const orderSchema = new Schema({
+        code: String,
+        rows: [rowSchema]
+      }, { timestamps: true });
+
+      const Requirement = requirementSchema;
+      Requirement.discriminators = {};
+      Requirement.discriminators['ComponentRequirement'] = componentRequirementSchema;
+      Requirement.discriminators['ToolRequirement'] = toolRequirementSchema;
```
hasezoey, line 7091:
```suggestion
      requirementSchema.discriminators = {};
      requirementSchema.discriminators['ComponentRequirement'] = componentRequirementSchema;
      requirementSchema.discriminators['ToolRequirement'] = toolRequirementSchema;
```

Why re-assign if not necessary?

hasezoey (commented):
Looks good to me.

> Should we continue to use validateSync() instead of validate() for backwards compatibility?

I think it ok for this small change as this would be more correct validation and also potentially faster because of being parallel.

> Should we skip running `validate` middleware on bulkSave() for backwards compatibility?

Was a different one run before?

vkarpov15:
`validateSync()` doesn't trigger `validate` middleware, so if we switch from `validateSync()` -> `validate()` then validation middleware will run. I changed back to `validateSync()` to maintain backwards compatibility, but we should switch to using `validate()` for Mongoose 9.

test/model.test.js
```diff
@@ -6904,7 +6894,7 @@ describe('Model', function() {
       userToUpdate.name = 'John Doe';
 
       // Act
-      const writeOperations = User.buildBulkWriteOperations([newUser, userToUpdate], { timestamps: false, skipValidation: true });
+      const writeOperations = await User.buildBulkWriteOperations([newUser, userToUpdate], { timestamps: false, skipValidation: true });
```
hasezoey, line 6897: `buildBulkWriteOperations` is now sync again, so `await` should likely be removed here and the other places too.
(though technically, `await` on sync is just a no-op; acting as if `await` was not there at all)

hasezoey (approved):
Still looks good to me, though now one minor style suggestions. Also maybe a test to check that it does not trigger `validate` middleware should be there. (if it is not already?)

test/model.test.js
vkarpov15: Fixed in 38c9b18

AbdelrahmanHafez:
@vkarpov15 Sorry I missed this since I was on vacation for the past couple of weeks.

I agree with the implemented solution here, maintaining backwards compatibility by using `validateSync` seems like the way to go, but we should introduce a breaking change in v9 in the interest of performance by using the async `validate`.

vkarpov15:
`validate()` isn't faster than `validateSync()`. If anything, `validate()` is slower because it runs both validate middleware and async validators.

Changing from `validateSync()` to `validate()` is more for consistency. `save()` calls `validate()` under the hood, so it would be reasonable to expect that `bulkSave()` does the same.

PR #15418, opened 2025-05-12
types: stricter projection typing with 1-level deep nesting

vkarpov15:
Re: #15327

**Summary**

@pshaddel please take a look as well.

Simplified alternative to #15327 that seems to compile with Typegoose. The one major tradeoff is that it only considers 1 level deep nested paths, as opposed to 3 in #15327. However, that should be sufficient for most use cases. There's also the added benefit that this PR reuses the existing `With1LevelDeepNestedPaths` type, with some fixes to handle arrays, which should reduce risk.

**Examples**

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -688,7 +688,7 @@ declare module 'mongoose' {
   type ExclusionProjection<T> = IsItRecordAndNotAny<T> extends true ? Projector<WithLevel1NestedPaths<T>, false | 0> & _IDType : AnyObject;
   type ProjectionUnion<T> = InclusionProjection<T> | ExclusionProjection<T>;
 
-  export type ProjectionType<T> = (ProjectionUnion<T> & AnyObject) | string | ((...agrs: any) => any);
+  export type ProjectionType<T> = (ProjectionUnion<T> & AnyObject) | string | ((...args: any) => any);
   export type SortValues = SortOrder;
 
   export type SortOrder = -1 | 1 | 'asc' | 'ascending' | 'desc' | 'descending';
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -686,9 +686,8 @@ declare module 'mongoose' {
   type _IDType = { _id?: boolean | 1 | 0 };
   type InclusionProjection<T> = IsItRecordAndNotAny<T> extends true ? Projector<WithLevel1NestedPaths<T>, true | 1> & _IDType : AnyObject;
   type ExclusionProjection<T> = IsItRecordAndNotAny<T> extends true ? Projector<WithLevel1NestedPaths<T>, false | 0> & _IDType : AnyObject;
-  type ProjectionUnion<T> = InclusionProjection<T> | ExclusionProjection<T>;
 
-  export type ProjectionType<T> = (ProjectionUnion<T> & AnyObject) | string | ((...args: any) => any);
+  export type ProjectionType<T> = (InclusionProjection<T> & AnyObject) | (ExclusionProjection<T> & AnyObject) | string;
   export type SortValues = SortOrder;
 
   export type SortOrder = -1 | 1 | 'asc' | 'ascending' | 'desc' | 'descending';
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -657,9 +657,38 @@ declare module 'mongoose' {
 
   export type ReturnsNewDoc = { new: true } | { returnOriginal: false } | { returnDocument: 'after' };
 
-  export type ProjectionElementType = number | string;
-  export type ProjectionType<T> = { [P in keyof T]?: ProjectionElementType } | AnyObject | string;
+  type ArrayOperators = { $slice: number | [number, number]; $elemMatch?: never } | { $elemMatch: Record<string, any>; $slice?: never };
+  /**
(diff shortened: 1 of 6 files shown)


pshaddel:
> Re: #15327
> 
> **Summary**
> 
> @pshaddel please take a look as well.
> 
> Simplified alternative to #15327 that seems to compile with Typegoose. The one major tradeoff is that it only considers 1 level deep nested paths, as opposed to 3 in #15327. However, that should be sufficient for most use cases. There's also the added benefit that this PR reuses the existing `With1LevelDeepNestedPaths` type, with some fixes to handle arrays, which should reduce risk.
> 
> **Examples**

Sure. I will also take a look at the original PR + Typegoose this week.

BTW. On this branch, I do not get any type suggestions for projection.

vkarpov15:
@pshaddel which IDE are you using and can you try restarting?

Type suggestions for projections seem to work well for me on this branch in zed@0.181.8

image

pshaddel:
> @pshaddel which IDE are you using and can you try restarting?
> 
> Type suggestions for projections seem to work well for me on this branch in zed@0.181.8
> 
> image

You are right, it needed a restart.
See this:
Screenshot 2025-05-12 at 23 17 14

These methods from here are also included:
```typescript
child?: PopulatedDoc<HydratedDocument<Child>>
```

I have a draft on my PR trying to remove these methods:

```typescript
  /**
   * Extract whatever is inside the PopulatedDoc
   * @example
   * type Doc = { name: string };
   * type Sample = PopulatedDoc<Doc>;
   * type DepopulatedDoc = DepopulatedDoc<Sample>; // Result: Doc
   */
  export type DepopulatedDoc<T> = T extends PopulatedDoc<infer U, infer RawID> ? Exclude<U, RawID> : T;
  /**
   * Extract the raw document from a hydrated document
   * @example
   * type Doc = { name: string };
   * type Sample = HydratedDocument<Doc>;
   * type ExtractedDoc = DehydrateDocument<Sample>; // Result: Doc
   */
  export type DehydrateDocument<T> = T extends HydratedDocument<infer U> ? U : T;
```

types/index.d.ts
```diff
@@ -657,9 +657,38 @@ declare module 'mongoose' {
...
+        ? {
+          [K in keyof T]?: T[K] extends Record<string, any> ? Projector<T[K], Element> | Element : Element;
+        }
+        : Element;
+  type _IDType = { _id?: boolean | 1 | 0 };
+  type InclusionProjection<T> = IsItRecordAndNotAny<T> extends true ? Projector<WithLevel1NestedPaths<T>, true | 1> & _IDType : AnyObject;
+  type ExclusionProjection<T> = IsItRecordAndNotAny<T> extends true ? Projector<WithLevel1NestedPaths<T>, false | 0> & _IDType : AnyObject;
+  type ProjectionUnion<T> = InclusionProjection<T> | ExclusionProjection<T>;
+
+  export type ProjectionType<T> = (ProjectionUnion<T> & AnyObject) | string | ((...agrs: any) => any);
```
hasezoey, line 691:
```suggestion
  export type ProjectionType<T> = (ProjectionUnion<T> & AnyObject) | string | ((...args: any) => any);
```

hasezoey (approved):
Can confirm this works with current typegoose types, thanks.

types/index.d.ts
vkarpov15: @hasezoey FYI I removed this type signature in f6545db because `((...args: any) => any);` was left in to support callbacks, which we no longer support at runtime. No sense in supporting `((...args: any) => any);` anymore because `Model.find(filter, fn)` throws an error now.

vkarpov15:
@pshaddel @hasezoey I successfully got my code to handle avoiding document methods, like `$assertPopulated()` and `toJSON()`, in projections. Please take another look :+1:

pshaddel:
> @pshaddel @hasezoey I successfully got my code to handle avoiding document methods, like `$assertPopulated()` and `toJSON()`, in projections. Please take another look 👍

LGTM! :)

msalafia:
@vkarpov15 I am not able to do something like this even though it should be allowed:

```t
const currentWeekDoc = await this.currentWeekModel
      .findOne(
        { _id: currentWeekId },
        {
          _id: 0,           // <----- error here because it should be allowed
          season: 1,
          seasonType: 1,
          week: 1,
          timestamp: 1,
        },
      )
      .lean()
      .exec();
```

DavideViolante:
@msalafia it is happening also into my project, but I found out the reason.
It seems like this is happening due to an "old definition" of our models in the code using an interface.

I had to do the following change to remove that error:
```diff
-const userSchema = new Schema<IUser>({
+const userSchema = new Schema({
```
and
```diff
-const User = model<IUser>('User', userSchema);
+const User = model('User', userSchema);
```

vkarpov15:
@msalafia we will have a fix in 8.15.1. As a workaround in the interim, you can do the following:

```ts
import { ProjectionType } from 'mongoose';

const currentWeekDoc = await this.currentWeekModel
      .findOne(
        { _id: currentWeekId },
        {
          _id: 0,
          season: 1,
          seasonType: 1,
          week: 1,
          timestamp: 1,
        } as ProjectionType<any>,
      )
      .lean()
      .exec();
```

msalafia:
@vkarpov15 Nice! Thank you!

Issue #15406, opened 2025-05-06
Autocomplete missing for $or, $elemMatch, $gte, and other MongoDB operators in JavaScript (VS Code)


Mehdi-Zarei:
### Prerequisites

- [x] I have written a descriptive issue title

### Mongoose version

8.12.1

### Node.js version

21

### MongoDB version

8.0.3

### Operating system

Windows

### Problem

When using Mongoose in **JavaScript** (not TypeScript), many MongoDB operators such as `$inc`, `$elemMatch`, and `$gte` are **not suggested in autocomplete** in VS Code.

This slows down development and creates confusion, especially for developers who are not using TypeScript.

In contrast, these suggestions **do appear correctly in JetBrains WebStorm**, so the issue seems related to how the type information or JSDoc is exposed to language servers.

---

### Example

Here is a real-world query:

```js
const result = await User.find({
  $or: [
    { userName: { $elemMatch: { name: "Mehdi" } } },
    { age: { $gte: 18 } }
  ]
});

```
In the above code, VS Code does not suggest $or, $elemMatch, or $gte in the autocomplete list.

What I Know
I understand that VS Code relies on the TypeScript language service even for JavaScript files. However, the lack of JSDoc or typings for MongoDB operators in JavaScript context seems to be the root cause here.

Feature Request
Would it be possible to:

Improve Mongoose's .d.ts or JSDoc annotations to make these operators more accessible to the language server for JavaScript users?

Thank you for your amazing work on Mongoose 🙌

vkarpov15:
`$or` does show up:

![Image](https://github.com/user-attachments/assets/3dc13dd8-efe7-4feb-bb4f-300c03a4e0ed)

However, the others do not show up because of #14510. That was a workaround for #14473 - TypeScript gets a bit messy with generics and `extends`. Because of #14473 we can't change this behavior without a breaking change.

PR #15422, opened 2025-05-13
BREAKING CHANGE: make FilterQuery properties no longer resolve to `any` in TypeScript

vkarpov15:
**Summary**

Right now `FilterQuery` types resolve to `any` for any query properties, which means query operators like $gt etc don't show up in intellisense (re: #15406). Making this change would break #14473, but that's an easy workaround - just add an `as`.

Bring back the changes reverted by #14510 - I think the benefit of autocomplete outweighs the friction of adding an `as` when you're using a pattern that you're not supposed to in TypeScript (at least according to https://stackoverflow.com/questions/56505560/how-to-fix-ts2322-could-be-instantiated-with-a-different-subtype-of-constraint).

**Examples**


hasezoey (approved):
LGTM

Issue #15431, opened 2025-05-23
Excluding _id and including another prop in projection returns TS error


DavideViolante:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.15.0

### Node.js version

22.x

### MongoDB server version

idk

### Typescript version (if applicable)

5.8.3

### Description

I've the same bug reported also [here](https://github.com/Automattic/mongoose/pull/15418#issuecomment-2897035479) by @msalafia
Note also my next comment there.
The bug is related to that PR changes.

In my project I have the error on this line:
```ts
const items = await User.find({}, { lastName: 1, _id: 0 });
```
Either if I remove lastName or _id it works, but is not the intended projection I want.

Error:
```
No overload matches this call.
  Overload 1 of 5, '(filter: RootFilterQuery<IUser>, projection?: ProjectionType<IUser>, options?: QueryOptions<IUser>): Query<...>', gave the following error.
    Type '0' is not assignable to type 'true | 1'.
  Overload 2 of 5, '(filter: RootFilterQuery<IUser>, projection?: ProjectionType<IUser>): Query<(Document<unknown, {}, IUser, {}> & IUser & Required<...> & { ...; })[], ... 4 more ..., {}>', gave the following error.
    Type '0' is not assignable to type 'true | 1'.ts(2769)
```

### Steps to Reproduce

```ts
const items = await User.find({}, { lastName: 1, _id: 0 });
```

Schema:
```ts
import { Document, model, Schema, Types } from 'mongoose';

const userSchema = new Schema<IUser>({
  username: { type: String, trim: true },
  firstName: { type: String, required: true, trim: true },
  lastName: { type: String, required: true, trim: true },
  // ...
});

export interface IUser extends Document {
  _id: Types.ObjectId;
  username: string;
  firstName: string;
  lastName: string;
  // ...
}

const User = model<IUser>('User', userSchema);

export default User;
```

### Expected Behavior
Ability to use _id: 0 and another prop.

PR #15432, opened 2025-05-23
types: correct handling of `_id` in ProjectionType

vkarpov15:
**Summary**

Quick fix for #15418: looks like we need to explicitly `Omit<T, '_id'>` to avoid TypeScript getting tripped up by `_id` in `InclusionProjection` and `ExclusionProjection`.

**Examples**

### DIFF
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -187,6 +187,8 @@ expectError(Test.find({}, { 'docs.profiles': { name: 'aa' } })); // should suppo
 expectError(Test.find({}, { endDate: { toString: 1 } }));
 expectError(Test.find({}, { tags: { trim: 1 } }));
 expectError(Test.find({}, { child: { toJSON: 1 } }));
+Test.find({}, { age: 1, _id: 0 });
+Test.find({}, { name: 0, age: 0, _id: 1 });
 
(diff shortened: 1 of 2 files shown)


vkarpov15:
This will fix #15431

hasezoey (commented):
LGTM

> Just a question about the tests, what's the difference between line 164, 167 and the newest 190-191? Seems like all of them check with _id: 0 and another prop: 1. Why line 164 or 167 passed the test before this PR?

At least line `191` is new as there previously was no `_id: 1` test. 

After some slight looking around and guess-work, it might be related to `Model.find` making use of `ProjectionType<TRawDocType>` where [in the tests `TRawDocType` (interface `ITest`)](https://github.com/Automattic/mongoose/blob/38f2da26a952e89e5274807aed0d31ca3caaeee2/test/types/queries.test.ts#L62-L70) does not include a `_id` property, but the example interface `IUser` in #15431 does. So my best guess is that in the tests it worked due to the `& AnyObject` clause in `ProjectionType`.

vkarpov15:
My best guess would be because existing tests have multiple fields with `: 1`, whereas the new test only has 1 field with `: 1`. TypeScript is full of surprising quirks and behaviors.

PR #15448, opened 2025-05-29
docs: fix typo

premdood:
**Summary**
The Document class is derived from Modal class


vkarpov15 (commented):
What makes you think this is a typo? In Mongoose, Model inherits from Document: https://github.com/Automattic/mongoose/blob/0c5f56f8bacc409e7b73f756196f34a307364647/lib/model.js#L116-L128

premdood:
After looking into this further, I realised this isn't actually a typo. The documentation is correct as written. Model inherits from Document by calling the Document constructor.
Apologies for the confusion.

PR #15464, opened 2025-06-04
fix: report default paths in VersionError message because they can can cause VersionError

vkarpov15:
**Summary**

I ran into a `VersionError: No matching document found for id "xyz" version 1 modifiedPaths "lastPaidAt, totalRevenue"` error - that shouldn't happen because `lastPaidAt` and `totalRevenue` are both primitives, shouldn't trigger versioning.

The root cause was an array default was triggering versioning: an array value wasn't set in the db. Paths that are in default status can trigger versioning, but VersionError doesn't currently list those paths.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -460,8 +460,8 @@ Model.prototype.$__handleSave = function(options, callback) {
     return;
   }
 
-  // store the modified paths before the document is reset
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
LGTM

Issue #15466, opened 2025-06-05
Nested schema in array causes subdocParent error when multiple Mongoose libraries are present


bootylactin:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

^8.13.2

### Node.js version

20.15.0

### MongoDB server version

6.0.23

### Description

When nesting an array of schemas inside another, Mongoose throws an error when attempting to validate.  A single, nested schema (not in an array) does not throw the error.  The error is:

```
\mongoose-issues\mongoose-shared-library\node_modules\mongoose\lib\document.js:2764
              !subdocParent.$__.activePaths.getStatePaths('modify').hasOwnProperty(fullPathToSubdoc) &&
                            ^

TypeError: Cannot read properties of undefined (reading '$__')
    at _getPathsToValidate (\mongoose-issues\mongoose-shared-library\node_modules\mongoose\lib\document.js:2764:29)
    at Document.$__validate (\mongoose-issues\mongoose-shared-library\node_modules\mongoose\lib\document.js:3013:25)
    at \mongoose-issues\nested-schema-test\node_modules\kareem\index.js:497:25
```

The following problem manifests at 8.13.2 and every version I tested afterward.  This example code runs fine at 8.13.1 and below.

This problem is specific to how our project is structured, as we are deploying lambda functions to AWS which all use a shared db-services module which are included via the "file: " syntax in the package.json.  To get this error to throw, both package.json files will need to be installed.  Note that this error **will not throw** when only one instance of the Mongoose library is present.  It's very specific to how we are using Mongoose inside our lambda functions to define schemas, but using a separate shared DB library to connect to our Mongo DBs.  It requires Mongoose be installed in two separate places.

I've attempted to create the simplest example that I can which follows our project structure.  The folder structure is as follows:
```
└── mongoose-issues/
    ├── mongoose-shared-library/
    │   ├── modules/
    │   │   └── db-service.js
    │   ├── index.js
    │   └── package.json   
    └── nested-schema-test/
        ├── package.json
        └── run-this-file.js
```

``` javascript
// run-this-file.js

const { Schema } = require( "mongoose" );
const dbService  = require( "mongoose-shared-library" ).dbService;

const nestedSchema = new Schema(
	{ name : String },
	{ _id : false }
);

const schema = new Schema( {
	subDoc   : nestedSchema,
[...]

PR #15471, opened 2025-06-09
fix(DocumentArray): correctly set parent if instantiated with schema from another Mongoose instance

vkarpov15:
**Summary**

#15466 points out that Mongoose throws an unreadable error if creating a document array that was defined with `Schema()` from a different Mongoose instance.

I added tests to cover this case automatically - historically we haven't tested the multiple `require()` instance setup, but we should since it is so common.

**Examples**


AbdelrahmanHafez (approved):
LGTM 👍

PR #15469, opened 2025-06-09
fix: fix validation for deeply nested maps of subdocuments

AbdelrahmanHafez:
This PR fixes #15447 

The issue was that mongoose was double validating nested maps of subdocuments, once as the correct path, but then trying to validate `.$*` which represents the map schema rather than the actual value of a path.

The bug was in the `_getPathsToValidate` function in `lib/document.js`. We were incorrectly treating map schema paths (paths ending with `.$*`) as actual document paths for validation. These `$*` paths represent the schema for Maps but are not actual document paths that should be validated.

This PR ignores the `.$*` "path" when getting paths to validate.

PR #15472, opened 2025-06-10
test: assert failing validation still throws an error in nested maps

AbdelrahmanHafez:
Add test case re #15469 to assert that validation still throws an error when the path fails the validation.


hasezoey (commented):
LGTM

PR #15474, opened 2025-06-10
fix(populate): handle populating deeply nested map paths 

vkarpov15:
**Summary**

There are a few issues that pop up when you try to populate a deeply nested virtual like `rubric.parameters.$*.evaluationPrompt`: getting the virtual, setting with `.$*`, etc.

**Examples**


hasezoey (commented):
LGTM

PR #15500, opened 2025-07-01
fix(model): avoid stripping out operations in bulkWrite() when ordered: false and _skipCastBulkWrite: true

### LINKED ISSUE #15495: Model.bulkSave() errors with `ordered: false`
### Prerequisites

### Mongoose version

8.16.1

### Node.js version

18.20.3

### MongoDB server version

6.17.0

_No response_

### Description

Calling `Model.bulkSave(documents, { ordered: false })` will always throw an error.

### Steps to Reproduce

You can repro by changing the "update documents" test in https://github.com/Automattic/mongoose/blob/master/test/model.test.js#L7040

```
-      await User.bulkSave(users);
+      await User.bulkSave(users, { ordered: false });
```

Running `npm test -- -g "updates documents"` should result in a test failure similar to the following:

```
  Model
    bulkSave() (gh-9673)
      1) updates documents


  0 passing (171ms)
  1 failing

  1) Model
       bulkSave() (gh-9673)
         updates documents:
     MongooseBulkSaveIncompleteError: User.bulkSave() was not able to update 3 of the given documents due to incorrect version or optimistic concurrency, document ids: 685da61806965d0afcb73d14, 685da61806965d0afcb73d15, 685da61806965d0afcb73d16
      at Function.bulkSave (lib/model.js:3616:11)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Context.<anonymous> (test/model.test.js:7654:7)
```

### Expected Behavior

_No response_

vkarpov15:
Fix #15495

**Summary**

As implemented, `bulkWrite()` will filter out all ops if `ordered: false` and `_skipBulkWriteCasting: true`. This PR fixes that, and also makes `validOps` vs `validOpsIndexes` more consistent

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3454,35 +3454,36 @@ Model.bulkWrite = async function bulkWrite(ops, options) {
       });
     }
   } else {
-    let remaining = validations.length;
-    let validOps = [];
+    let validOpIndexes = [];
     let validationErrors = [];
     const results = [];
-    await new Promise((resolve) => {
-      if (validations.length === 0) {
-        return resolve();
-      }
-      for (let i = 0; i < validations.length; ++i) {
-        validations[i]((err) => {
-          if (err == null) {
-            validOps.push(i);
-          } else {
-            validationErrors.push({ index: i, error: err });
-            results[i] = err;
-          }
-          if (--remaining <= 0) {
+    if (validations.length > 0) {
+      validOpIndexes = await Promise.all(ops.map((op, i) => {
+        if (i >= validations.length) {
+          return i;
+        }
+        return new Promise((resolve) => {
+          validations[i]((err) => {
+            if (err == null) {
+              resolve(i);
+            } else {
+              validationErrors.push({ index: i, error: err });
+              results[i] = err;
+            }
             resolve();
-          }
+          });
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
I don't completely understand the problem/solution, but we have tests that verify that it works, so LGTM! 👍

vkarpov15:
Here's where `validations` is set:

```js
const validations = options?._skipCastBulkWrite ? [] : ops.map(op => castBulkWrite(this, op, options));
```

Current `ordered: false` logic only adds `validOps` if the validation passes, skipping if there's no validation:

```js
await new Promise((resolve) => {
      if (validations.length === 0) {
        return resolve();
      }
      for (let i = 0; i < validations.length; ++i) {
        validations[i]((err) => {
          if (err == null) {
            validOps.push(i);
          } else {
            validationErrors.push({ index: i, error: err });
            results[i] = err;
          }
          if (--remaining <= 0) {
            resolve();
          }
        });
      }
    });
```

Which means that, if `_skipCastBulkWrite: true`, `validOps.push(i)` never executes. That's the TLDR of why this happens.

PR #15514, opened 2025-07-03
BREAKING CHANGE: remove background option from indexes

### LINKED ISSUE #15476: Background index creation is not longer supported by MongoDB
### Prerequisites

### Mongoose version

8.7.0

### Node.js version

22

### MongoDB server version

8

5.8.3

### Description

Hi dear reader,

I do not think this can be counted as a bug. But nonetheless I was not able to find a better category for it. So as you can see in [this Q&A](https://www.mongodb.com/community/forums/t/after-4-2-background-createindexes/15136/2) from MongoDB 4.2 onward we do not have an option to create indexes in the background or foreground.

So why is it that I can still specify a value for this when I call this `index` API?

```ts
import { model, Schema } from "mongoose";

const CommentSchema = new Schema({
  name: { type: String, default: "hahaha" },
  age: { type: Number, min: 18 },
  bio: { type: String, match: /[a-z]/ },
  date: { type: Date, default: Date.now },
  buff: Buffer,
});
const CommentModel = model("Comment", CommentSchema);

CommentSchema.index({ age: 1 }, { name: "commentAge", background: true });
```

### Steps to Reproduce

N/A

### Expected Behavior

`background` should not be a valid option 🙂.

vkarpov15:
Fix #15476

**Summary**

MongoDB no longer supports `background` option for indexes, that option is now a no-op. This PR makes Mongoose 9 no longer support that option.

**Examples**

### DIFF
--- a/docs/migrating_to_9.md
+++ b/docs/migrating_to_9.md
@@ -70,7 +70,7 @@ In Mongoose 9, `next(null, 'new arg')` doesn't overwrite the args to the next mi
 
 ## Removed background option for indexes
 
-[MongoDB no longer supports the `background` option for indexes as of MongoDB 4.2](https://www.mongodb.com/docs/manual/core/index-creation/#index-operations). Mongoose 9 will no longer set the background option by default.
+[MongoDB no longer supports the `background` option for indexes as of MongoDB 4.2](https://www.mongodb.com/docs/manual/core/index-creation/#index-operations). Mongoose 9 will no longer set the background option by default and Mongoose 9 no longer supports setting the `background` option on `Schema.prototype.index()`.
 
 ## Subdocument `deleteOne()` hooks execute only when subdocument is deleted
 


docs/migrating_to_9.md
```diff
@@ -68,6 +68,10 @@ schema.pre('save', function(next, arg) {
 
 In Mongoose 9, `next(null, 'new arg')` doesn't overwrite the args to the next middleware.
 
+## Removed background option for indexes
+
+[MongoDB no longer supports the `background` option for indexes as of MongoDB 4.2](https://www.mongodb.com/docs/manual/core/index-creation/#index-operations). Mongoose 9 will no longer set the background option by default.
```
hasezoey: > Mongoose 9 will no longer set the background option by default.

Shouldnt it also say that the option is now completely removed?

PR #15518, opened 2025-07-05
types: avoid FlattenMaps by default on toObject(), toJSON(), lean()

vkarpov15:
Re: #13523

**Summary**

**Examples**

### DIFF
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -256,23 +256,34 @@ declare module 'mongoose' {
     set(value: string | Record<string, any>): this;
 
     /** The return value of this method is used in calls to JSON.stringify(doc). */
+    toJSON(options: ToObjectOptions & { flattenMaps: true, flattenObjectIds: true, virtuals: true }): FlattenMaps<ObjectIdToString<Default__v<Require_id<DocType & TVirtuals>>>>;
+    toJSON(options: ToObjectOptions & { flattenMaps: true, flattenObjectIds: true }): FlattenMaps<ObjectIdToString<Default__v<Require_id<DocType>>>>;
+    toJSON(options: ToObjectOptions & { flattenMaps: true, virtuals: true }): FlattenMaps<Default__v<Require_id<DocType & TVirtuals>>>;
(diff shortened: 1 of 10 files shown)


types/document.d.ts
```diff
@@ -256,23 +256,34 @@ declare module 'mongoose' {
     set(value: string | Record<string, any>): this;
 
     /** The return value of this method is used in calls to JSON.stringify(doc). */
+    toJSON(options: ToObjectOptions & { flattenMaps: true, flattenObjectIds: true, virtuals: true }): FlattenMaps<ObjectIdToString<Default__v<Require_id<DocType & TVirtuals>>>>;
```
hasezoey: Should those options be listed in `ToObjectOptions` with documentation?

hasezoey (commented):
LGTM.

Requires minimal changes to typegoose's tests, once something that was written wrongly by typegoose and once because it explicitly tested against having `FlattenMaps` in a `toJSON()`.

Is it intentional to target branch `vkarpov15/schema-create` instead of `9.0`?

> Potential consolidation we could do: make toObject() and toJSON() convert Buffers to mongodb.Binary, in both JS runtime and TypeScript types. This would be a breaking change, and most likely cause some headache for application developers because Binary and Buffer end up with different representations in JSON. But this would reduce the number of types we need to support in TypeScript. WDYT?

I dont have particular feelings about that, but it seems like it `Binary` is base64 encoded and so requires less storage in json, but at the same time seemingly does not have a identifier marking it as a `Buffer` or `Binary` like `Buffer` has.
Maybe make it toggle-able via similar options to `flattenObjectIds` / `flattenMaps`?

vkarpov15:
Yes, intentional to branch off of `schema-create` branch. There's some stuff in this PR that relies on `schema-create` changes.

For now, I'll hold off on making `toObject()` and `toJSON()` convert buffer to binary.

david-loe:
looking forward to a consistent output of `lean()`, `toJSON()` and `toObject()` for `Buffer` 🚀

Issue #15522, opened 2025-07-08
Mismatch between getChanges, modifiedPaths, and isModified methods


nikzanda:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.16.2

### Node.js version

20.10.0

### MongoDB server version

6.0.2

### Typescript version (if applicable)

5.8.3

### Description

When updating a model schema by adding a new field with a default value, retrieving any record and then executing the `getChanges` method returns a MongoDB instruction to update the record with the new field and its default value (a very nice feature!).

However, if I execute the `modifiedPaths` or `isModified` methods, they return an empty array and false, respectively. Why do these results mismatch?

Reproduction link [here](https://stackblitz.com/edit/vitejs-vite-2qv6mtse?file=src%2Findex.ts).

<img width="616" height="104" alt="Image" src="https://github.com/user-attachments/assets/15a7db05-2d13-4f4b-bc6c-0b24bad52538" />

### Steps to Reproduce

1. Create a model  
1. Insert a document  
1. Add a field with a default value to the schema  
1. Query the inserted document  
1. Call the `getChanges`, `modifiedPaths`, and `isModified` methods  

### Expected Behavior

My goal is to automatically update a document when new fields are added to the schema. For example, I would like to check for any changes by calling the `isModified` method, and if it returns true, execute `save`.

vkarpov15:
Because, as written, `testField` and `booleanField` are not modified, they're set to `default` state in Mongoose's change tracking, which is slightly different.

Use `$isDefault()` to check if there are any fields in `default` state:

```javascript
console.log(setting.isModified()); // false
console.log(setting.$isDefault()); // true

setting.isModified() || setting.$isDefault(); // Check if there are modified or default paths
```

Does this help?

nikzanda:
`$isDefault` accepts the `path` parameter, so it cannot be executed without parameters. Currently, the fastest solution for me is to check if `Array.keys(setting.getChanges()).length` is greater than zero. If other methods are not intended for this scope, I’ll keep this solution.

vkarpov15:
You can call `$isDefault()` without `path`, we need to update TypeScript types for that.

PR #15528, opened 2025-07-10
types(document): allow calling `$isDefault()` with no args

vkarpov15:
Re: #15522

**Summary**

Calling `$isDefault()` with no args is supported by runtime, but TypeScript types don't allow this.

**Examples**


hasezoey (approved):
LGTM. Maybe a simple types test would be good (or modify one)?

vkarpov15:
I think I'll hold off on modifying the tests for this, it is a fairly trivial change.

PR #15530, opened 2025-07-11
types(inferschematype):  Infer Typescript string enums

ruiaraujo:
**Summary**

We are using InferSchemaType to derive types from the models from the schema definition.
For fields with Typescript enum, the types are currently `string` which creates unnecessary casting.

I had to use `expectAssignable` for the required case and it is not completely clear why, but it will solve my problem which is around assignment.


vkarpov15 (approved):
Thanks :+1:

PR #15538, opened 2025-07-15
types(schema): handle `required: string` in schema definitions

### LINKED ISSUE #15536: Required type property incorrect when using custom error messages
### Prerequisites

### Mongoose version

8.16.3

### Node.js version

20.19.0

### MongoDB server version

7.x

5.8.3

### Description

I have discovered some inconsistent typing in the inferred types for `required` properties. It works for a couple of different cases and breaks for a couple more. 

### Steps to Reproduce

There are a few different ways to document a property as being required, some of them can be correctly inferred by mongoose but at least one of them appears to be broken.

1. `required: true` - works! ✅

```ts

const UserModelNameRequired = mongoose.model('User', new mongoose.Schema(
  {
  },
));

const user1 = new UserModelNameRequired({ name: 'test' });

user1.name
//    ^? (property) name: string
```

2. `required: [true, 'This is a custom error message']` - works! ✅

```ts
const UserModelNameRequiredArrayCustomMessage = mongoose.model('User', new mongoose.Schema(
  {
    name: { type: String, required: [true, 'This is a custom error message'] }
  }
));

const user2 = new UserModelNameRequiredArrayCustomMessage({ name: 'test' });

user2.name
//    ^? (property) name: string
```

3. `required: 'This is a custom error message'` - does not work ❌

```ts
const UserModelNameRequiredCustom = mongoose.model('User', new mongoose.Schema(
  {
    name: { type: String, required: 'This is a custom error message' },
  },
));

const user3 = new UserModelNameRequiredCustom({ name: 'test' });

user3.name
//    ^? (property) name?: string | null | undefined
```

4. `required: () => 'This is a custom error message'` - not sure if it's possible ❓

```ts
const UserModelNameRequiredCustomFunction = mongoose.model('User', new mongoose.Schema(
  {
    name: { type: String, required: () => 'This is a custom error message' },
  },
));

const user4 = new UserModelNameRequiredCustomFunction({ name: 'test' });

user4.name
//    ^? (property) name?: string | null | undefined
```

### Expected Behavior

Case 1 and 2 are working as expected. Case 3 above should be properly typed as `(property) name: string`. I'm not sure if case 4 could ever be fixed because it's impossible to know what the function returns with statically?

vkarpov15:
Fix #15536

**Summary**

`required: 'my custom message'` makes a path required with a custom error message in JS, I updated TypeScript types to reflect that.

**Examples**

### DIFF
--- a/test/types/schema.test.ts
+++ b/test/types/schema.test.ts
@@ -1881,3 +1881,14 @@ function gh15516() {
     expectType<HydratedUserDoc>(this);
   });
 }
+
+function gh15536() {
+  const UserModelNameRequiredCustom = model('User', new Schema(
+    {
+      name: { type: String, required: 'This is a custom error message' }
+    }
+  ));
+
+  const user3 = new UserModelNameRequiredCustom({ name: null });
+  expectType<string>(user3.name);
+}
--- a/types/inferschematype.d.ts
+++ b/types/inferschematype.d.ts
@@ -107,7 +107,7 @@ type IsPathDefaultUndefined<PathType> = PathType extends { default: undefined } 
  * @param {TypeKey} TypeKey A generic of literal string type."Refers to the property used for path type definition".
  */
 type IsPathRequired<P, TypeKey extends string = DefaultTypeKey> =
-  P extends { required: true | [true, string | undefined] | { isRequired: true } } | ArrayConstructor | any[]
+  P extends { required: true | string | [true, string | undefined] | { isRequired: true } } | ArrayConstructor | any[]
     ? true
     : P extends { required: boolean }
       ? P extends { required: false }


domharrington (approved):
I think this will definitely work for my case! Thanks again. Now I can see how you're doing this type of change, I'll look to going straight to submitting a PR of my own in future if that's welcomed?

test/types/schema.test.ts
```diff
@@ -1881,3 +1881,14 @@ function gh15516() {
...
+
+function gh15536() {
+  const UserModelNameRequiredCustom = model('User', new Schema(
+    {
+      name: { type: String, required: 'This is a custom error message' }
+    }
+  ));
+
+  const user3 = new UserModelNameRequiredCustom({ name: null });
+  expectType<string>(user3.name);
```
hasezoey, line 1893: Shouldnt this test against something like `InferSchemaType`?

hasezoey (approved):
LGTM

test/types/schema.test.ts
vkarpov15: No, I think we should test against `new Model()`, that's more important IMO. Ideally users don't have to use `InferSchemaType`.

vkarpov15:
@domharrington sure, you're always welcome to put in PRs.

PR #15571, opened 2025-08-04
fix: trim long strings in minLength and maxLength error messages and display the string length

### LINKED ISSUE #15550: Have max length string validation return the number of characters for string that caused the validation error
### Prerequisites

### 🚀 Feature Proposal

Currently, when the string characters exceed the max length set on the schema, it will tell you the maximum length you set and what the string is that caused it, but it does not tell you what the number of characters for the errant string are. It would be a nice QOL improvement if it did that so its easier to see what the adjustment needs to be.

### Motivation

_No response_

### Example

_No response_

vkarpov15:
Fix #15550

**Summary**

minLength and maxLength validators currently display the full string output, which can be problematic in the case of very long strings. This PR trims the output if more than 30 chars, and also displays the length of the string.

**Examples**


hasezoey (approved):
LGTM, maybe add a test to know that >30 characters is actually shortened?

PR #15574, opened 2025-08-07
feat(schema): support for union types

### LINKED ISSUE #10894: Different types for field in schema with a setter

I have a birthday field that the client send a string like "03/02/2000" and I convert it via set to a Date
Is it possible to define two types the first one String (the first validation by mongoose to check if it's a string) and a second type of Date (when I read it from the DB)
```typescript
birthday: {
    type: SchemaTypes.Date
    set: (value: string) => {
        return value ? startOfDay(parse(value, 'DD/MM/YYYY', new Date())).toISOString() : null;
    }
}
```

Right now I get the following error
`mongoose validation error Cast to Date failed for value "03/02/2000" (type string) at path "birthday" `

Currently I use the "latest version" 6.0.11

vkarpov15:
Fix #10894. There's still work to do on this PR (TypeScript types, options testing, thinking about how to handle getters/setters on unions) but I wanted to get this out there for review.

**Summary**

`type: 'Union'` means a given path can be one of multiple types.

For example, in the following schema, `purchased` can be either a Boolean or a Date.

```javascript
const schema = new Schema({
  purchased: {
    type: 'Union',
    of: [Boolean, Date]
  }
});
```

Order matters in `of`: Mongoose will loop through the `Of` schema types and take the first one that casts successfully, _unless_ one of the schema types returns the original value, in which case Mongoose just takes the original value.

For example, in the above case, `purchased: 1` would get converted to a boolean (true) because Mongoose can cast 1 -> true. But `purchased: 2` would get converted to a Date, because `2` is not a value Mongoose can cast to a Boolean by default.

**Examples**

### DIFF
--- a/lib/options/schemaUnionOptions.js
+++ b/lib/options/schemaUnionOptions.js
@@ -1,0 +1,32 @@
+'use strict';
+
+const SchemaTypeOptions = require('./schemaTypeOptions');
+
+/**
+ * The options defined on a Union schematype.
+ *
+ * @api public
+ * @inherits SchemaTypeOptions
+ * @constructor SchemaUnionOptions
+ */
+
+class SchemaUnionOptions extends SchemaTypeOptions {}
+
+const opts = require('./propertyOptions');
+
+/**
+ * If set, specifies the type of this map's values. Mongoose will cast
+ * this map's values to the given type.
+ *
+ * If not set, Mongoose will not cast the map's values.
+ *
+ * @api public
+ * @property of
+ * @memberOf SchemaUnionOptions
+ * @type {Function|string}
+ * @instance
+ */
+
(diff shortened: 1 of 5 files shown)


hasezoey (approved):
LGTM, though lint would need to be fixed and as Copilot already pointed out, the documentation for `of` still refers to `map`s.

types/schematypes.d.ts
```diff
@@ -188,7 +188,7 @@ declare module 'mongoose' {
     _id?: boolean;
 
     /** If set, specifies the type of this map's values. Mongoose will cast this map's values to the given type. */
```
hasezoey, line 190: This comment should likely mention that the property is also valid for `Union` now.

PR #15576, opened 2025-08-10
BREAKING CHANGE: remove bson as direct dependency, use mongodb/lib/bson instead

### LINKED ISSUE #15154: Use mongodb exported bson instead of mongoose to import bson separately?

vkarpov15:
Fix #15154

**Summary**

The primary purpose of bson as a top-level dependency is to allow importing mongoose without the mongodb node driver for the browser build (now in https://github.com/mongoosejs/mongoose-browser). Right now, you can require() in `mongoose/lib/mongoose` and get the core Mongoose global without the MongoDB node driver, but the problem is how to handle types like ObjectId and Decimal128.

I worked through a couple of alternatives, including:

1. Using `driver.SchemaTypes` property to make Mongoose's mongodb driver adapter define ObjectId, Decimal128, etc. schema types. That approach isn't great because mongoose-browser would have to duplicate those implementations.
2. Using a new `driver.Types` property to make Mongoose's mongodb driver adapter instead add `ObjectId` etc. to `Types`. That would mean ObjectId and Decimal128 SchemaTypes would error out if using a driver that doesn't define ObjectId and Decimal128 types

All of those have unfortunate tradeoffs. The reason why we didn't want to use `mongodb/lib/bson` was we were worried about potential breaking changes with the MongoDB node driver, but since the BSON export is part of their TypeScript types I'm comfortable making this change and removing bson as a top-level dependency.

**Examples**


hasezoey (approved):
LGTM.
Just to confirm, with the approach of directly importing from `mongodb/lib/bson` it does not import the whole `mongodb` package and break the browser build, or does the new browser package do something special?

vkarpov15:
Correct, importing mongodb/lib/bson just imports bson, not the rest of the mongodb node driver. I tested this change with mongoose-browser using npm pack and it seemed to work fine.

I still want to double check deno though, looks like there's some deno test failures

AbdelrahmanHafez (approved):

PR #15587, opened 2025-08-16
BREAKING CHANGE: make create() and insertOne() params more strict, remove generics to prevent type inference

### LINKED ISSUE #15355: Ditch `AnyKeys` in `Model.create` and `Model.insertOne`
### Prerequisites

### 🚀 Feature Proposal

Only reference I can find to this is #11148, which is now several years old.

I initially suggested this be a feature in typegoose (see [here](https://github.com/typegoose/typegoose/issues/995#issuecomment-2807041242)), but was informed that mongoose now has the features necessary to infer enough type information to properly type `Model.create` and `Model.insertOne`.

I mention in the issue linked above that `AnyKeys` destroys type information, and is equivalent to the following type, which only preserves keys and not any information about their values, ie the use of `T[P] | any` is equivalent to `any`.

```ts
// given type
type AnyKeys<T> = { [P in keyof T]?: T[P] | any };
// equivalent to
type AnyKeys<T> = Partial<Record<keyof T, any>>
```

Likewise, if you put this type in union as `T | AnyKeys<T>` (as is default in `Model.create`), it is equivalent to `AnyKeys<T>`, as typescript favors the least restrictive type in the union if types overlap.

In other words, `TRawDocType | AnyKeys<TRawDocType>` means `Partial<Record<keyof TRawDocType, any>>`.

### Motivation

Suppose you're storing something of the following shape;

```txt
{
  foo: {
    bar: string
    baz: number
  }
}
```
`foo` is a required property. It must be an object. It must have the key `bar` of type `string`, and it must have the key `baz` of type `number.

It is unhelpful for the type inference in the first argument to `model.create` to propose the shape;

```txt
{
  foo?: any
}
```
`foo` is an optional property. It can be anything at all.

The only way to get around this is to explicitly specify the type on every call to `model.create` or `model.insertOne`. This should not be needed. The correct type is already stored in the model via `TRawDocType`, specifying it again on every call is redundant.

### Example

```diff
export interface Model/*...*/ {
  //...
-  create<DocContents = AnyKeys<TRawDocType>>(docs: Array<TRawDocType | DocContents>, options: CreateOptions & { aggregateErrors: true }): Promise<(THydratedDocumentType | Error)[]>;
-  create<DocContents = AnyKeys<TRawDocType>>(docs: Array<TRawDocType | DocContents>, options?: CreateOptions): Promise<THydratedDocumentType[]>;
-  create<DocContents = AnyKeys<TRawDocType>>(doc: DocContents | TRawDocType): Promise<THydratedDocumentType>;
-  create<DocContents = AnyKeys<TRawDocType>>(...docs: Array<TRawDocType | DocContents>): Promise<THydratedDocumentType[]>;
+  create(docs: Array<TRawDocType>, options: CreateOptions & { aggregateErrors: true }): Promise<(THydratedDocumentType | Error)[]>;
+  create(docs: Array<TRawDocType>, options?: CreateOptions): Promise<THydratedDocumentType[]>;
+  create(doc: TRawDocType): Promise<THydratedDocumentType>;
+  create(...docs: Array<TRawDocType>): Promise<THydratedDocumentType[]>;
  //...
-  insertOne<DocContents = AnyKeys<TRawDocType>>(doc: DocContents | TRawDocType, options?: SaveOptions): Promise<THydratedDocumentType>;
+  insertOne(doc: TRawDocType, options?: SaveOptions): Promise<THydratedDocumentType>;
}
```

Or, if it is still the case that `TRawDocType` is unreliable for whether a property is optional, `Partial<TRawDocType>` could be used. If its unreliable for nested keys too, a `DeepPartial` type (eg, [ts-essentials DeepPartial](https://github.com/ts-essentials/ts-essentials/tree/master/lib/deep-partial)) could be created and used.

vkarpov15:
Fix #15355

**Summary**

Supporting `create<DocContents>(doc: DocContents)` means that you can pass any value to `doc` due to TypeScript type inference. This PR removes that while retaining support for most common cases: passing strings for ObjectIds, POJOs for maps and subdocs, plain arrays for document arrays, strings for buffers and uuids. Also uses `Partial<>` to allow excluding properties that may be set by middleware or defaults.

This makes autocomplete for `create()` and `insertOne()` much better:

image

There are some definite downsides to this change: custom getters/setters and custom casting logic can transform types, and there's no way to plug in custom types into `ApplyBasicCreateCasting`. But those are problems you can work around with `as`.

**Examples**


types/models.d.ts
```diff
@@ -260,6 +260,14 @@ declare module 'mongoose' {
     hint?: mongodb.Hint;
   }
 
+  type ApplyBasicCreateCasting<T> = {
```
hasezoey, line 273: Maybe some documentation for this type would be good?

docs/migrating_to_9.md
```diff
@@ -294,6 +294,24 @@ function findById<ModelType extends {_id: Types.ObjectId | string}>(model: Model
...
+const schema = new Schema({ age: Number });
+const TestModel = mongoose.model('Test', schema);
+
+// Worked in Mongoose 8, TypeScript error in Mongoose 9
+const doc = await TestModel.create({ age: 'not a number', someOtherProperty: 'value' });
+```
+
+In Mongoose 9, `create()` and `insertOne()` no longer accept a generic parameter. Instead, they accept `Partial<RawDocType>` with some additional query casting applied that allows objects for maps, strings for ObjectIds, and POJOs for subdocuments and document arrays.
+
+```ts
```
hasezoey, line 364: Maybe a comment explaining the following code block would be good instead of just having it directly after this paragraph?

hasezoey (commented):
This would likely be a good idea, but after trying it in typegoose, i came across the following problems:
- `number` does not seem to be valid for a `Date` property (ie `time: Date.now()` is now invalid, even though default casting would work)
- `array` is not valid anymore for maps and have to be a object now (instead of `[['key1', val1], ['key2', val2]]`)
- somehow `new mongoose.Types.ObjectId()` is not valid for `_id` (or any ObjectId field?)

The only big problem is that somehow, as mentioned above, `new ObjectId()` does not seem to be valid for `ObjectIdQueryTypeCasting`. Due to this PR i also had to add `mongoose.Require_id<T>` to `Model->TRawDocType`.

Aside from that, there are only expected changes, like intentionally providing invalid data for runtime validation testing, and some type incompatibilities due to typegoose likely not using the correct mapping and instead passing hydrated document types for `TRawDocType`  (and so typescript complains about missing properties like `$pop, $shift, addToSet, isMongooseArray`)

<details>
<summary>For context, the full diff of the changes i had already done</summary>

Note that patch is not "full" as i have not figured out the subdocument situation

```diff
diff --git a/src/types.ts b/src/types.ts
index 4e37283a..905f65dc 100644
--- a/src/types.ts
+++ b/src/types.ts
@@ -29,7 +29,7 @@ export type ArraySubDocumentType<T, QueryHelpers = BeAnObject> = DocumentType<T,
  * Used Internally for ModelTypes
  */
 export type ModelType<T, QueryHelpers = BeAnObject> = mongoose.Model<
-  T, // raw doc type
+  mongoose.Require_id<T>, // raw doc type
   QueryHelpers, // query helpers
   IObjectWithTypegooseFunction, // instance methods
   DefaultIdVirtual // virtuals
diff --git a/test/tests/arrayValidator.test.ts b/test/tests/arrayValidator.test.ts
index a7961c67..152e6573 100644
--- a/test/tests/arrayValidator.test.ts
+++ b/test/tests/arrayValidator.test.ts
@@ -68,7 +68,7 @@ it('should respect enum [String]', async () => {
       enumedString: [
         'not in the enum', // string not in the enum
       ],
-    });
+    } as any);
 
     fail('Expected to throw ValidationError!');
   } catch (err) {
@@ -96,7 +96,7 @@ it('should respect enum [Number]', async () => {
       enumedNumber: [
         5, // number not in the enum
       ],
-    });
+    } as any);
 
     fail('Expected to throw ValidationError!');
   } catch (err) {
diff --git a/test/tests/numberValidator.test.ts b/test/tests/numberValidator.test.ts
index 7138b736..a482c1f5 100644
--- a/test/tests/numberValidator.test.ts
+++ b/test/tests/numberValidator.test.ts
@@ -23,7 +23,7 @@ it('should respect enum', async () => {
   try {
     await NumberValidatorsModel.create({
       enumed: 5, // number not in the enum
-    });
+    } as any);
 
     fail('Expected to throw ValidationError!');
   } catch (err) {
diff --git a/test/tests/ref.test.ts b/test/tests/ref.test.ts
index b3446458..b3130714 100644
--- a/test/tests/ref.test.ts
+++ b/test/tests/ref.test.ts
@@ -409,10 +409,14 @@ it('Reference-Maps should work and be populated', async () => {
   const dummy2 = await RefMapDummyModel.create({ dummy: '2' });
 
   const doc1 = await RefMapModel.create({
-    mapped: [
-      ['1', dummy1],
-      ['2', dummy2],
-    ],
+    // mapped: [
+    //   ['1', dummy1],
+    //   ['2', dummy2],
+    // ],
+    mapped: {
+      '1': dummy1,
+      '2': dummy2,
+    },
   });
 
   const found = await RefMapModel.findById(doc1).orFail().exec();
diff --git a/test/tests/shouldRun.test.ts b/test/tests/shouldRun.test.ts
index 48061907..ea6efa42 100644
--- a/test/tests/shouldRun.test.ts
+++ b/test/tests/shouldRun.test.ts
@@ -1147,7 +1147,7 @@ it('should validate Decimal128', async () => {
   try {
     await CarModel.create({
       carModel: 'Tesla',
-      price: 'NO DECIMAL',
+      price: 'NO DECIMAL' as any,
     });
     fail('Validation must fail.');
   } catch (e) {
@@ -1181,7 +1181,7 @@ it(`should Validate Map`, async () => {
   try {
     await InternetUserModel.create({
       projects: {
-        p1: 'project',
+        p1: 'project' as any,
       },
     });
     fail('Validation should Fail');
```

</details>

vkarpov15:
@hasezoey I added support for array of arrays for maps and strings+numbers for dates. Also added `Require_id`. Does this help?

vkarpov15:
@hasezoey I fixed the ` _id: new mongoose.Types.ObjectId(),` issue, I needed to use `DeepPartial<>` to make subdoc and document array subpaths optional. However I'm still getting a bunch of these errors in typegoose:

```
    test/tests/biguser.test.ts:56:55 - error TS2339: Property 'id' does not exist on type 'Document<unknown, BeAnObject, User, BeAnyObject, {}> & Omit<User & { _id: ObjectId; } & { __v: number; }, "typegooseName"> & IObjectWithTypegooseFunction'.

    56       const foundUser = await UserModel.findById(user.id).populate('car previousCars').orFail().exec();

```

Do you have any idea what might be causing those?

vkarpov15:
1. Re: `create()` with no args, I fixed that in 4936971
2. I see that making discriminator model instances from the base model is inconvenient now, I had to do the following, which is a bit messy. However, how often do you think people create discriminator model instances from the base model vs using the discriminator model? I imagine if you're able to pass the discriminator model hydrated doc type to `create<>`, you should also have a reference to the discriminator model.

image

PR #15593, opened 2025-08-18
BREAKING CHANGE: consolidate RootQuerySelector, Condition, etc. types with MongoDB driver's

vkarpov15:
**Summary**

`QuerySelector`, `RootQuerySelector`, and `Condition` are almost exactly duplicates of types the MongoDB Node driver exports. With a little work, we can make it so that Mongoose uses the MongoDB driver's types rather than maintaining its own (which may be out of date).

**Examples**


vkarpov15:
@hasezoey can you please take another look and see if you still get the same error with typegoose?

vkarpov15:
I checked and you're right, removing the NonNullable<> in the following fixes typegoose tests.

```ts
export function isDocumentArray<T, S extends RefType>(
  docs: mongoose.Types.Array<Ref<T, S>> | null | undefined
): docs is mongoose.Types.Array<DocumentType<NonNullable<T>>>;
```

I'll take a look as to why.

vkarpov15:
@hasezoey can you please take another look? Typegoose tests seem to pass on latest `feature/13.0` branch and this latest PR

image


I also ended up stripping out a bunch of unnecessary generics to shave off some instantiations for performance. Will need to make some changelog notes for that.

types/models.d.ts
```diff
@@ -916,21 +916,21 @@ declare module 'mongoose' {
...
       filter: QueryFilter<TRawDocType>,
       update: UpdateQuery<TRawDocType> | UpdateWithAggregationPipeline,
       options?: (mongodb.UpdateOptions & MongooseUpdateQueryOptions<TRawDocType>) | null
-    ): QueryWithHelpers<UpdateWriteOpResult, ResultDoc, TQueryHelpers, TRawDocType, 'updateOne', TInstanceMethods & TVirtuals>;
-    updateOne<ResultDoc = THydratedDocumentType>(
+    ): QueryWithHelpers<UpdateWriteOpResult, THydratedDocumentType, TQueryHelpers, TRawDocType, 'updateOne', TInstanceMethods & TVirtuals>;
+    updateOne(
       update: UpdateQuery<TRawDocType> | UpdateWithAggregationPipeline
-    ): QueryWithHelpers<UpdateWriteOpResult, ResultDoc, TQueryHelpers, TRawDocType, 'updateOne', TInstanceMethods & TVirtuals>;
+    ): QueryWithHelpers<UpdateWriteOpResult, THydratedDocumentType, TQueryHelpers, TRawDocType, 'updateOne', TInstanceMethods & TVirtuals>;
```
hasezoey, line 933: Why remove the ability for `updateOne` to specify / overwrite the result type?

Edit: in case github still shows this later, somehow in the PR-thread interface it show the completely wrong lines, but shows it correctly in the PR-files interface.

hasezoey (approved):
Looks good to me now in its current state. Though the generics removal should arguably be a separate PR.

types/models.d.ts
vkarpov15: Better tsc performance due to fewer instantiations is the primary driver.

I also don't see `updateOne<ResultType>` being especially useful because `updateOne()` always returns `UpdateOneResult`, the generic parameter doesn't update the ResultType virtual. Also, `updateOne` not having a generic is more consistent with `deleteOne`, which similarly always returns `DeleteOneResult`.

PR #15605, opened 2025-08-30
Fix types for a string of enums.

ruiaraujo:
**Summary**
When using an array of enum, the types inferred was an array string.

The reason is that when resolving the type, we drop the options information by using `ObtainDocumentPathType`. 
The fix was for the Scalar case to recursively use ResolvePathType again and jam in the Options by forwarding the enum value. While not elegant, it works so here is my proposal.


vkarpov15 (approved):
Thanks :+1:

PR #15619, opened 2025-09-02
types: improve performance on inferSchemaType re: #15588

vkarpov15:
**Summary**

Various performance improvements related to #15588 to make it so that top-line instantiations and memory usage for #15588 are roughly equivalent to master.

1. Consolidate array and readonly array paths
2. Minimize use of `IfEquals`: `IfEquals<>` is only necessary for `String` because of TypeScript's implicit boxing/unboxing (see below screenshot), there's no risk of TypeScript inferring container types for ObjectIds or Decimal128. There is a chance with `Boolean` and `Number`, but only if someone does `type: true` or `type: 42`, I don't think it is worth checking for that case though.
3. Consolidate `UnionToType`

@ssalbdivad can you also please take a look?

image

**Examples**

### DIFF
--- a/types/inferschematype.d.ts
+++ b/types/inferschematype.d.ts
@@ -1,4 +1,5 @@
 import {
+  AnyArray,
   BooleanSchemaDefinition,
   DateSchemaDefinition,
   DefaultSchemaOptions,
@@ -246,12 +247,9 @@ type IsSchemaTypeFromBuiltinClass<T> =
   : T extends Types.Decimal128 ? true
   : T extends NativeDate ? true
   : T extends typeof Schema.Types.Mixed ? true
-  : IfEquals<T, Schema.Types.ObjectId, true, false> extends true ? true
   : unknown extends Buffer ? false
   : T extends Buffer ? true
   : false;
-
-type UnionToType<T extends readonly any[]> = T[number] extends infer U ? ResolvePathType<U> : never;
 
 /**
  * @summary Resolve path type by returning the corresponding type.
@@ -269,12 +267,10 @@ type ResolvePathType<
   TypeHint,
   never,
   PathValueType extends Schema ? InferSchemaType<PathValueType>
-  : PathValueType extends (infer Item)[] ?
-    IfEquals<
-      Item,
-      never,
-      any[],
-      Item extends Schema ?
+  : PathValueType extends AnyArray<infer Item> ?
+    IfEquals<Item, never> extends true
+      ? any[]
+      : Item extends Schema ?
         // If Item is a schema, infer its type.
         Types.DocumentArray<InferSchemaType<Item>>
       : Item extends Record<TypeKey, any> ?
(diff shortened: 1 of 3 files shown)


hasezoey (approved):

types/inferschematype.d.ts
```diff
@@ -291,54 +287,23 @@ type ResolvePathType<
...
   : IfEquals<PathValueType, String> extends true ? PathEnumOrString<Options['enum']>
   : PathValueType extends NumberSchemaDefinition ?
     Options['enum'] extends ReadonlyArray<any> ?
       Options['enum'][number]
     : number
-  : IfEquals<PathValueType, Schema.Types.Number> extends true ? number
   : PathValueType extends DateSchemaDefinition ? NativeDate
-  : IfEquals<PathValueType, Schema.Types.Date> extends true ? NativeDate
-  : PathValueType extends typeof Buffer | 'buffer' | 'Buffer' | typeof Schema.Types.Buffer ? Buffer
+  : PathValueType extends typeof Buffer | 'buffer' | 'Buffer' | typeof Schema.Types.Buffer | Schema.Types.Buffer ? Buffer
```
ssalbdivad: I would use the aliased versions of these types like `BufferSchemaDefinition` that are abstracted out so TS can better cache the comparisons rather than inlining these unions. They should be in `index.d.ts` after my last commit.

types/inferschematype.d.ts
```diff
@@ -291,54 +287,23 @@ type ResolvePathType<
...
+  : PathValueType extends 'uuid' | 'UUID' | typeof Schema.Types.UUID | Schema.Types.UUID ? Buffer
   : PathValueType extends 'double' | 'Double' | typeof Schema.Types.Double ? Types.Double
-  : IfEquals<PathValueType, Schema.Types.UUID> extends true ? Buffer
-  : PathValueType extends MapConstructor | 'Map' ? Map<string, ObtainDocumentPathType<Options['of']>>
-  : IfEquals<PathValueType, typeof Schema.Types.Map> extends true ? Map<string, ObtainDocumentPathType<Options['of']>>
+  : PathValueType extends MapConstructor | 'Map' | typeof Schema.Types.Map | Schema.Types.Map ? Map<string, ObtainDocumentPathType<Options['of']>>
   : PathValueType extends 'Union' | 'union' | typeof Schema.Types.Union ?
-    Options['of'] extends readonly any[] ?
-      UnionToType<Options['of']>
+    Options['of'] extends AnyArray<infer U> ? ResolvePathType<U>
```
ssalbdivad, line 313: It looks like the definition of `AnyArray` is this:

```ts
  export type AnyArray<T> = T[] | ReadonlyArray<T>;
```

You're safe to just use an expressions like `Options['of'] extends ReadonlyArray<infer U>` in situations like this because a mutable array is assignable to a readonly array so this will always work.

You're right to make sure this case is handled carefully though because the opposite is not true- why I'm assuming this utility exists.

After you've been bitten by this a few times it definitely gets ingrained that if you ever see an `extends __[]` without a `readonly` or `ReadonlyArray<>`, it's a red flag.

types/inferschematype.d.ts
vkarpov15: Fixed in [c9d3d37](https://github.com/Automattic/mongoose/pull/15619/commits/c9d3d37f59a693555314380a5c9afe82b1f248e2)

types/inferschematype.d.ts
vkarpov15: Believe me, I've tried to make this change countless times, but for some reason changing AnyArray -> ReadonlyArray makes tsd blow up with:

```ts
$ npm run test-tsd

> mongoose@8.18.0 test-tsd
> node ./test/types/check-types-filename && tsd

Error running tsd:
Error: Debug Failure. False expression.
    at Object.isUndefinedIdentifierExpression (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:55016:15)
    at typeFromExpression (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:137328:22)
    at serializeTypeOfExpression (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:137063:20)
    at Object.serializeTypeOfDeclaration (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:137087:16)
    at serializeTypeForDeclaration (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:57354:41)
    at addPropertyToElementList (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:56233:43)
    at createTypeNodesFromResolvedType (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:56128:11)
    at createTypeNodeFromObjectType (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:55863:25)
    at visitAndTransformType (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:55786:24)
    at createAnonymousTypeNode (/home/v/Desktop/MongoDB/mongoose/node_modules/@tsd/typescript/typescript/lib/typescript.js:55726:20)
```

I'm not certain why this happens and figuring it out hasn't been a priority.

ssalbdivad: Maybe an issue with how the types are displayed for TSD since I think it uses a string comparison?

Obviously this particular issue is not a huge deal and wouldn't have much effect on performance, but may be worth looking into `@ark/attest` for type testing as well as benchmarking since it has a much more powerful API for this sort of thing:

https://github.com/arktypeio/arktype/tree/main/ark/attest#readme

vkarpov15: I'll try out attest for testing in a separate PR, looking forward to trying it out. In the meantime, I'll merge this PR for 8.19.

hasezoey: Alternatively, typegoose makes use of [`tstyche`](https://www.npmjs.com/package/tstyche), which is more similar to the current `tsd` test style in use (at least from my quick look at `@ark/attest`)

ssalbdivad: @hasezoey that is probably true, and tstyche seems like a very good solution if you only need static analysis (i.e. what you can assert with TS types alone).

If you want to test for specific type errors, completions, jsdoc/go to definition, type display (snapshotting) etc., an approach like attest's that can interact with the typescript compiler directly offers a lot of advantages.

PR #15656, opened 2025-09-30
BREAKING CHANGE: remove skipId parameter to Document() and Model(), always use as options

### LINKED ISSUE #8862: Remove `skipId` param for documents
Similar to #4625 . This option is only used to avoid setting a default id on `replaceOne()` and `findOneAndReplace()`, and it would be cleaner to just delete `_id` before sending the operation to the server as opposed to passing this option everywhere.

vkarpov15:
Fix #8862

**Summary**

The `skipId` parameter was taking up some prime real estate as the 3rd arg to `new MyModel()`, replace that with `options`. We already supported passing `options`, this PR just drops support for passing a boolean as 3rd arg to `new MyModel()`.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -79,6 +79,7 @@ const VERSION_ALL = VERSION_WHERE | VERSION_INC;
  * @param {Object} [fields] optional object containing the fields which were selected in the query returning this document and any populated paths data
  * @param {Object} [options] various configuration options for the document
  * @param {Boolean} [options.defaults=true] if `false`, skip applying default values to this document.
+ * @param {Boolean} [options.skipId=false] By default, Mongoose document if one is not provided and the document's schema does not override Mongoose's default `_id`. Set `skipId` to `true` to skip this generation step.
  * @inherits NodeJS EventEmitter https://nodejs.org/api/events.html#class-eventemitter
  * @event `init`: Emitted on a document after it has been retrieved from the db and fully hydrated by Mongoose.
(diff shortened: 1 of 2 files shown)


lib/document.js
```diff
@@ -85,12 +85,12 @@ const VERSION_ALL = VERSION_WHERE | VERSION_INC;
  * @api private
  */
 
-function Document(obj, fields, skipId, options) {
-  if (typeof skipId === 'object' && skipId != null) {
-    options = skipId;
-    skipId = options.skipId;
+function Document(obj, fields, options) {
```
hasezoey, line 89: Consider adding `skipId` option to JSDoc

lib/model.js
```diff
@@ -104,7 +104,8 @@ const saveToObjectOptions = Object.assign({}, internalToObjectOptions, {
  *
  * @param {Object} doc values for initial set
  * @param {Object} [fields] optional object containing the fields that were selected in the query which returned this document. You do **not** need to set this parameter to ensure Mongoose handles your [query projection](https://mongoosejs.com/docs/api/query.html#Query.prototype.select()).
- * @param {Boolean} [skipId=false] optional boolean. If true, mongoose doesn't add an `_id` field to the document.
+ * @param {Object} [options] optional object containing the options for the document.
+ * @param {Boolean} [options.defaults=true] if `false`, skip applying default values to this document.
```
hasezoey, line 108: Consider adding `skipId` option to JSDoc

PR #15679, opened 2025-10-05
perf: Avoid needless path translation

orgads:
when arrayFilters is not set.


orgads:
@vkarpov15

vkarpov15 (approved):
Thanks :+1:

PR #15684, opened 2025-10-10
ci: add publish script with provenance, with --dry-run for now

vkarpov15:
**Summary**

Automate publishing of mongoose, with provenance enabled. I have `--dry-run` set for now for testing purposes, but once we test that out we should be able to publish automatically.

**Examples**

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,6 +28,4 @@ jobs:
         run: npm run build-browser
 
       - name: Dry run publish with provenance
-        run: npm publish --provenance --access public --dry-run
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npm publish --provenance --access public


.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+          node-version: '20.x'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Dry run publish with provenance
+        run: npm publish --provenance --access public --dry-run
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
hasezoey: If [trusted publishing](https://docs.npmjs.com/trusted-publishers) is configured, this shouldnt be necessary anymore, or is this something to do later?

vkarpov15: I was not aware of trusted publishing, I will do some research and try that out

PR #15697, opened 2025-10-23
upgrade to pre-release of mongodb node driver 7

vkarpov15:
**Summary**

I needed to make a couple of small adjustments: no more `noResponse` option, and numbers are no longer valid as ObjectIds, but the rest seems to work.

**Examples**

### DIFF
--- a/test/docs/validation.test.js
+++ b/test/docs/validation.test.js
@@ -381,8 +381,10 @@ describe('validation docs', function() {
     err.errors['numWheels'].message;
     // acquit:ignore:start
     assert.equal(err.errors['numWheels'].name, 'CastError');
-    assert.equal(err.errors['numWheels'].message,
-      'Cast to Number failed for value "not a number" (type string) at path "numWheels"');
+    assert.match(
+      err.errors['numWheels'].message,
+      /^Cast to Number failed for value "not a number" \(type string\) at path "numWheels"/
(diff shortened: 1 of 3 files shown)


test/types/base.test.ts
```diff
@@ -57,8 +57,8 @@ function gh10139() {
 }
 
 function gh12100() {
-  mongoose.syncIndexes({ continueOnError: true, noResponse: true });
-  mongoose.syncIndexes({ continueOnError: false, noResponse: true });
+  mongoose.syncIndexes({ continueOnError: true, sparse: true });
```
hasezoey, line 60: This should likely be in the migration guide.

AbdelrahmanHafez (commented):
Other than the comment from @hasezoey, LGTM 👍

test/types/base.test.ts
vkarpov15: I disagree. For one thing, removing the `noResponse` option is a change in the MongoDB Node driver, not Mongoose; and we typically don't list out MongoDB Node driver changes in our changelog. For another, the purpose of this test is not specifically to test the `noResponse` option, but to test that you can specify MongoDB Node driver index options when calling `syncIndexes()` - `noResponse` was just the option we originally chose, but `sparse` achieves the same result.

PR #15700, opened 2025-10-27
refactor: pass parent schema to SchemaType constructors in interpretAsType to make implementing custom container types easier

vkarpov15:
**Summary**

Implementing your own container type is a pain unless the container type happens to extend from Array, because otherwise you need to handle interpreting `of` by yourself. For example, imagine `{ type: Set, of: { type: String, required: true } }` - right now anyone implementing a set schematype would have to re-implement interpretAsType semantics themselves. We had to do a bit of a hack to avoid having to re-implement interpretAsType for unions, but with this fix we can remove the hack as well. 

**Examples**

### DIFF
--- a/lib/types/map.js
+++ b/lib/types/map.js
@@ -1,8 +1,6 @@
 'use strict';
 
-const Mixed = require('../schema/mixed');
 const MongooseError = require('../error/mongooseError');
-const assert = require('assert');
 const clone = require('../helpers/clone');
 const deepEqual = require('../utils').deepEqual;
 const getConstructorName = require('../helpers/getConstructorName');


lib/schema/union.js
```diff
@@ -14,12 +14,12 @@ const firstValueSymbol = Symbol('firstValue');
  */
 
 class Union extends SchemaType {
-  constructor(key, options, schemaOptions = {}) {
+  constructor(key, options, schemaOptions, parentSchema) {
```
hasezoey, line 25: This change should likely be documented somehwere in the code, though i dont know if this already exists somehwere (and `SchemaType` itself seems to have different parameters, so likely not there).
Also the "Custom Schema Types" guide seems to be somewhat outdated, and this change should likely be documented there as a available parameter (`/docs/customschematypes.md`).

vkarpov15: Fixed, although I did a lot of refactor to make this consistent

vkarpov15:
@hasezoey can you please take another look? I did a bunch more refactor to make the constructor signatures more consistent - I was able to make it so that all schema types take in a `parentSchema` param, although there's still some inconsistencies in function signatures that I would like to fix in the 9.0 branch.

lib/schema/documentArrayElement.js
```diff
@@ -14,18 +14,20 @@ const getConstructor = require('../helpers/discriminator/getConstructor');
  *
  * @param {String} path
  * @param {Object} options
+ * @param {Object} _schemaOptions
```
hasezoey:
```suggestion
 * @param {Object} schemaOptions
```

The other schematypes dont refer to it with underscore in the jsdoc when unused (re `SchemaBoolean`, `SchemaBuffer`, etc)

lib/schema/map.js
```diff
@@ -8,15 +8,21 @@ const MongooseMap = require('../types/map');
...
  * ignore
  */
 
 class SchemaMap extends SchemaType {
-  constructor(key, options) {
-    super(key, options, 'Map');
+  constructor(key, options, _schemaOptions, parentSchema) {
+    super(key, options, 'Map', parentSchema);
     this.$isSchemaMap = true;
+
```
hasezoey:
```suggestion

```

lib/schema/map.js
```diff
@@ -8,15 +8,21 @@ const MongooseMap = require('../types/map');
 const SchemaMapOptions = require('../options/schemaMapOptions');
 const SchemaType = require('../schemaType');
 const createJSONSchemaTypeDefinition = require('../helpers/createJSONSchemaTypeDefinition');
+const MongooseError = require('../error/mongooseError');
+const Schema = require('../schema');
+const utils = require('../utils');
 
 /*!
  * ignore
  */
```
hasezoey: `SchemaMap` is public right? This should likely also have a proper jsdoc then like the other schema types (though not directly related to this PR)

lib/schema/map.js
```diff
@@ -8,15 +8,21 @@ const MongooseMap = require('../types/map');
...
 
 class SchemaMap extends SchemaType {
-  constructor(key, options) {
-    super(key, options, 'Map');
+  constructor(key, options, _schemaOptions, parentSchema) {
```
hasezoey:
```suggestion
  constructor(key, options, schemaOptions, parentSchema) {
```

Its not unused

lib/schema/map.js
```diff
@@ -8,15 +8,21 @@ const MongooseMap = require('../types/map');
...
+    super(key, options, 'Map', parentSchema);
     this.$isSchemaMap = true;
+    
+    // Create the nested schema type for the map values
+    this._createNestedSchemaType(parentSchema, key, options, _schemaOptions);
```
hasezoey:
```suggestion
    this._createNestedSchemaType(parentSchema, key, options, schemaOptions);
```

Re: its not unused

lib/schema/map.js
```diff
@@ -12,23 +12,48 @@ const MongooseError = require('../error/mongooseError');
...
   }
 
+  /**
+   * Sets a default option for all Map instances.
+   *
+   * @param {String} option The option you'd like to set the value for
+   * @param {Any} value value for option
+   * @return {undefined}
+   * @function set
+   * @static
```
hasezoey, line 41: From what i can tell, `set` is not marked `static`

vkarpov15: Good catch, I fixed this in the 8.20 branch

Issue #15389, opened 2025-05-01
Inability to Migrate Primitive String Field to Subdocument Schema via get or init Hook


timheerwagen:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.14.0

### Node.js version

22.13.1

### MongoDB server version

"mongodb-memory-server": "^10.1.4"

### Typescript version (if applicable)

5.8.2

### Description

I'm attempting to migrate a time field stored as a primitive string (e.g., `"12:34"`) to a subdocument with `hours` and `minutes` keys. This transformation needs to apply to multiple fields in a dynamic, reusable way. So there cant be a `getter` or `init` hook on the main document.

I tried two approaches:

Using a `get` function on the schema field.

Adding a `pre("init")` hook within the subdocument schema.

However, neither strategy correctly transforms the old string value. It seems the validator strips the string before the getter or hook is executed.

### Steps to Reproduce

1. Create and save a model with a string time field.

```ts
const oldUserSchema = new Schema<{ unknownKey: string }>(
  {
    unknownKey: {
      type: String,
      required: true,
    },
  },
);

const UserModelName = "User";

const OldUser = mongoose.model(UserModelName, oldUserSchema);

await OldUser.create({ unknownKey: "12:34" });
```

2. Delete old model
```ts
mongoose.deleteModel(UserModelName);
```

3. Redefine it using a subdocument `timeSchema` and attempt to transform the string value using a getter or pre("init") hook.
```ts
type Time = { hours: number; minutes: number };

const timeStringToObject = (time: string | Time): Time => {
  if (typeof time !== "string") return time;

  const [hours, minutes] = time.split(":");

  return { hours: parseInt(hours), minutes: parseInt(minutes) };
};

const timeSchema = new Schema(
  {
    hours: { type: Number, required: true },
    minutes: { type: Number, required: true },
  },
);

const userSchema = new Schema<{ unknownKey: Time }>({
  unknownKey: {
    type: timeSchema,
    get: timeStringToObject, // try getter
    required: true,
  },
});

// try also pre init hook
timeSchema.pre("init", function (rawDoc) {
  console.log(rawDoc);
  if (typeof rawDoc === "string") {
    rawDoc = timeStringToObject(rawDoc);
  }
});

const User = mongoose.model(UserModelName, userSchema);
```

4. Query documents and inspect their output
```ts
[...]

PR #15702, opened 2025-10-30
Allow pre() hooks to overwrite arguments to the wrapped function

vkarpov15:
**Summary**

#15389 pointed out a case that Mongoose doesn't handle very well: suppose you have a `time` field in the database that is a string, but you want to convert it to be an object when hydrating from the db. There's no way to do this with getters, but it should be possible if you overwrite the arguments to `init` in a pre('init') hook.

In general, it is useful to be able to overwrite the arguments passed in to the middleware function, it's just something we should do through an explicit function call similar to `skipWrappedFunction()` and `overwriteResult()`.

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -6574,7 +6574,7 @@ describe('document', function() {
 
       const User = db.model('Test', userSchema);
       const doc = new User({});
-      doc.$init({ time: '12:30' });
+      doc.init({ time: '12:30' });
       await doc.validate();
     });
 


test/document.test.js
```diff
@@ -14907,6 +14951,56 @@ describe('document', function() {
...
+    });
+    userSchema.pre('validate', function(options) {
+      if (options == null) {
+        return mongoose.overwriteMiddlewareArguments({ pathsToSkip: ['test'] });
+      }
+    });
+    const User = db.model('Test', userSchema);
+    const user = new User();
+    await user.validate();
+    await assert.rejects(() => user.validate({}), /Path `test` is required/);
```
AbdelrahmanHafez: Nitpick:
Took me a minute to understand the difference here, using an explicit `null` would make it clearer:
```suggestion
    await user.validate(null);
    await assert.rejects(() => user.validate({}), /Path `test` is required/);
```

lib/document.js
```diff
@@ -849,11 +853,11 @@ function init(self, obj, doc, opts, prefix) {
  * @instance
  */
 
-Document.prototype.updateOne = function updateOne(doc, options, callback) {
+Document.prototype.updateOne = function updateOne(doc, options) {
```
hasezoey, line 856: Removal of callback should likely be noted in the migration guide.

lib/mongoose.js
```diff
@@ -1327,6 +1327,43 @@ Mongoose.prototype.skipMiddlewareFunction = Kareem.skipWrappedFunction;
...
+ *     const User = mongoose.model('User', userSchema);
+ *     const doc = new User({});
+ *     doc.$init({ time: '12:30' });
+ *
+ * @method overwriteMiddlewareArguments
+ * @param {...any} args The new arguments to be passed to the next middleware. Pass multiple arguments as a spread, **not** as an array.
+ * @api public
+ */
+
+Mongoose.prototype.overwriteMiddlewareArguments = Kareem.overwriteArguments;
```
hasezoey, line 1365: This still needs to be added to the types.

test/document.test.js
```diff
@@ -14907,6 +14951,56 @@ describe('document', function() {
...
+
+    const userSchema = new Schema({
+      unknownKey: {
+        type: timeSchema,
+        required: true
+      }
+    });
+    const User = db.model('Test', userSchema);
+    await User.collection.insertOne({ unknownKey: '12:34' });
+    const user = await User.findOne();
```
hasezoey: Just to ensure no flaky tests (residual data) happen, maybe consider finding by id instead.

test/document.test.js
hasezoey: Or a comment would likely also be enough

hasezoey (approved):

PR #15705, opened 2025-11-01
fix(model+plugins): correctly apply shard key on deleteOne()

### LINKED ISSUE #15701: Mongoose not sending shard key on deleteOne()
### Prerequisites

### Mongoose version

8.19.*

### Node.js version

24.*

### MongoDB server version

6.*

_No response_

### Description

This bug is similar to https://github.com/Automattic/mongoose/issues/7393, but the fix was applied using `'remove'` when it should now be using `'deleteOne'`.

https://github.com/Automattic/mongoose/blob/master/lib/plugins/sharding.js#L19-L22

### Steps to Reproduce

See the related issue: https://github.com/Automattic/mongoose/issues/7393

### Expected Behavior

The query to delete the document using `deleteOne` contains the shard key.

vkarpov15:
Fix #15701

**Summary**

Couple of issues preventing shard key from getting set on `deleteOne()`:

1. Plugin still uses `pre('remove')` not `pre('deleteOne')`, even though `pre('remove')` is no longer supported
2. `deleteOne()` needs to handle custom `$where` property on documents similar to how `save()` does. That's what sharding plugin uses to indicate what the shard key is.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -802,8 +802,18 @@ Model.prototype.deleteOne = function deleteOne(options) {
     }
   }
 
-  query.pre(function queryPreDeleteOne(cb) {
-    self.constructor._middleware.execPre('deleteOne', self, [self], cb);
+  query.pre(async function queryPreDeleteOne() {
+    await new Promise((resolve, reject) => {
+      self.constructor._middleware.execPre('deleteOne', self, [self], err => {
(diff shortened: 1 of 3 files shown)


hasezoey (commented):
LGTM

PR #15725, opened 2025-11-06
Fix: cast id parameter based on schema _id type in DocumentArray.id()

### LINKED ISSUE #15724: DocumentArray.id() doesn't properly cast id parameter for custom _id types (String, Number, etc.)
### Prerequisites

### Mongoose version

8.19.3

### Node.js version

v24.2.0

### MongoDB server version

6.0

_No response_

### Description

The `DocumentArray.id()` method always attempts to cast the `id` parameter to ObjectId, but it should cast based on the schema's `_id` type. This causes issues when using custom `_id` types like String, Number, or Object.

**Location:** `lib/types/documentArray/methods/index.js:124` - There's a TODO comment indicating this needs to be fixed: `@TODO cast to the _id based on schema for proper comparison`

**Current Behavior:** When calling `array.id(someId)` on a DocumentArray where documents have String or Number `_id` types, the method may not find matching documents correctly because it tries to cast everything to ObjectId first.

**Expected Behavior:** The `id()` method should:
1. Get the schema's `_id` type from the document array schema
2. Cast the input `id` parameter to match that type
3. Then perform the comparison

This is a bug because the method doesn't work correctly with non-ObjectId `_id` types, even though Mongoose supports them.

### Steps to Reproduce

1. Create a schema with custom String `_id`:pt
const CustomSchema = new Schema({
  _id: { type: String, required: true },
  name: String
});
const ParentSchema = new Schema({ children: [CustomSchema] });
const Parent = mongoose.model('Parent', ParentSchema);2. Create document and try to find child by id:
const parent = new Parent({ children: [{ _id: 'abc123', name: 'test' }] });
await parent.save();

const found = parent.children.id('abc123');
// May return null because method tries to cast to ObjectId instead of String

## Expected Behavior
The `id()` method should:
1. Detect the schema's `_id` type (String, Number, Object, ObjectId, etc.)
2. Cast the input `id` parameter to match that type using the schema's cast method
3. Perform the comparison using the properly casted value

This would make the method work correctly with all `_id` types that Mongoose supports, not just ObjectId.

**Note:** There's already a TODO comment in the code at line 124 of `lib/types/documentArray/methods/index.js` that acknowledges this needs to be fixed.

Lex-Ashu:
Closes #15724

## Summary

The `DocumentArray.id()` method was always attempting to cast the `id` parameter to ObjectId, which caused issues when using custom `_id` types like String, Number, or Object. This fix implements schema-based casting by:

1. Getting the `_id` schema type from the document array schema
2. Casting the input `id` parameter to match that type before comparison
3. Maintaining backward compatibility with ObjectId fallback

This resolves the TODO comment at line 124 and ensures the method works correctly with all `_id` types that Mongoose supports.

## Example demonstrating the fix:

```javascript
// Schema with custom String _id
const CustomSchema = new Schema({
  _id: { type: String, required: true },
  name: String,
});

const ParentSchema = new Schema({ children: [CustomSchema] });
const Parent = mongoose.model("Parent", ParentSchema);

// Create a parent document with a child
const parent = new Parent({ children: [{ _id: "abc123", name: "test" }] });
await parent.save();

// Before this fix, parent.children.id('abc123') would fail to find the document
// because 'abc123' would be incorrectly cast to an ObjectId.
// After this fix, it correctly casts 'abc123' to a String and finds the document.
const found = parent.children.id("abc123");
console.log(found); // Will now correctly output the embedded document: { _id: 'abc123', name: 'test' }
```

### DIFF
--- a/lib/types/documentArray/methods/index.js
+++ b/lib/types/documentArray/methods/index.js
@@ -137,7 +137,7 @@ const methods = {
       idSchemaType = schemaType.schema.path('_id');
     } else if (schemaType && schemaType.casterConstructor && schemaType.casterConstructor.schema) {
       idSchemaType = schemaType.casterConstructor.schema.path('_id');
-    } else if (this.length > 0 && this[0] && this[0].$__schema) {
+    } else if (this.length > 0 && this[0] != null && this[0].$__schema) {
       idSchemaType = this[0].$__schema.path('_id');
     }
 


lib/types/documentArray/methods/index.js
```diff
@@ -121,23 +121,48 @@ const methods = {
...
+      return null;
+    }
+
+    const schemaType = this[arraySchemaSymbol];
+    let idSchemaType = null;
+
+    if (schemaType && schemaType.casterConstructor && schemaType.casterConstructor.schema) {
+      idSchemaType = schemaType.casterConstructor.schema.path('_id');
+    } else if (this.length > 0 && this[0] && this[0].$__schema) {
+      idSchemaType = this[0].$__schema.path('_id');
```
vkarpov15, line 141: Better to do `schemaType.schema.path('_id')` here in case of discriminators

lib/types/documentArray/methods/index.js
```diff
@@ -121,23 +121,48 @@ const methods = {
...
+    }
+
+    let castedId = id;
+    if (idSchemaType) {
+      try {
+        castedId = idSchemaType.cast(id);
+      } catch (e) {
+        castedId = id;
```
vkarpov15: Why `= id` when in L160 we do `casted = null`?

lib/types/documentArray/methods/index.js
```diff
@@ -121,23 +121,48 @@ const methods = {
...
+      }
+    }
 
+    let casted;
     try {
-      casted = castObjectId(id).toString();
+      if (isBsonType(castedId, 'ObjectId')) {
```
vkarpov15: I think we should get rid of this extra ObjectId code path. Shouldn't be necessary since we're already casting to the `_id` schematype, and the only thing this code path does differently is convert `toString()`, which isn't necessary because of `utils.deepEqual()`.

lib/types/documentArray/methods/index.js
```diff
@@ -153,10 +178,14 @@ const methods = {
           return val;
         }
       } else if (!isBsonType(id, 'ObjectId') && !isBsonType(_id, 'ObjectId')) {
```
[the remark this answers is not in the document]
vkarpov15: :+1: for this

lib/types/documentArray/methods/index.js
Lex-Ashu: I updated that block so it now uses `schemaType.schema.path('_id')`. That should ensure discriminators get the right schematype.

lib/types/documentArray/methods/index.js
Lex-Ashu: I removed that extra ObjectId branch and rely on the schema-based cast now—`deepEqual()` handles the comparison without needing the extra `toString()` path.

PR #15727, opened 2025-11-07
Performance: avoid cloning when saving document if minimize: false #15694

### LINKED ISSUE #15694: Performance: avoid cloning when saving document if `minimize: false`
### Prerequisites

### Issue

In `$__handleSave()` we have the following:

```javascript
  if (this.$isNew) {
    // send entire doc
    const obj = this.toObject(saveToObjectOptions);
```

However, if `minimize: false`, we don't need to run `toObject()`, we can just do `const obj = this._doc` because the MongoDB driver will clone and apply `toBSON()` anyway. This should significantly help performance.

Re: #8024, let's implement this functionality first and make sure it works as intended before considering making `minimize` default to false.

mahi656:
## Description
Optimize `$__handleSave()` performance by skipping `toObject()` call when `minimize: false`

## Problem
When saving a new document (`isNew: true`), Mongoose always calls `toObject(saveToObjectOptions)` to prepare the document for insertion. However, when `minimize: false`, this call is unnecessary because:
- The MongoDB driver clones the document internally before insertion
- The driver automatically applies `toBSON()` transformations
- The `toObject()` call adds significant overhead by cloning, transforming, and processing the entire document structure

## Solution
Skip the `toObject()` call when `schema.options.minimize === false` and directly use `this._doc` instead. The MongoDB driver will handle all necessary cloning and transformations.

## Performance Impact
This optimization significantly improves save performance for schemas with `minimize: false` by eliminating an expensive document transformation step. The performance gain is especially noticeable for:
- Large documents with many fields
- Documents with nested subdocuments
- High-throughput applications performing many inserts

## Testing
✅ All existing tests pass:
- Save-related tests: **118 passing**
- Minimize-related tests: **17 passing**
- isNew-related tests: **15 passing**

Verified the optimization works correctly:
- ✅ MongoDB driver handles cloning correctly
- ✅ Tested with `minimize: false` schemas - optimization applied
- ✅ Tested with `minimize: true` schemas - no behavior change
- ✅ Tested with schemas where `minimize` is undefined - no behavior change

## Backward Compatibility
✅ This change is fully backward compatible:
- Only affects documents being inserted (`isNew: true`)
- Only applies optimization when `minimize: false`
- All other cases use existing `toObject()` logic

## Issue
Fixes #15694

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -358,11 +358,11 @@ Model.prototype.$__handleSave = function(options, callback) {
   if (this.$isNew) {
     // send entire doc
     let obj;
-    if (this.$__schema.options.minimize === false) {
-      obj = this._doc;
-    } else {
-      obj = this.toObject(saveToObjectOptions);
-    }
+    // Always use toObject for serialization; control minimize via options
+    saveToObjectOptions = Object.assign({}, saveToObjectOptions, {
+      minimize: this.$__schema.options.minimize
+    });
+    obj = this.toObject(saveToObjectOptions);
     if ((obj || {})._id === void 0) {
       // documents must have an _id else mongoose won't know
       // what to update later if more changes are made. the user
@@ -5199,4 +5199,4 @@ function _getContexts(hook) {
  * Module exports.
  */
 
-module.exports = exports = Model;
+module.exports = Model;


mahi656 (commented):
done this export correctly

vkarpov15:
closing this re: https://github.com/Automattic/mongoose/issues/15694#issuecomment-3513758147

PR #15731, opened 2025-11-09
docs: add detailed loadClass() TypeScript usage guide and new loadcla…

### LINKED ISSUE #12813: Docs :: Add guidance on using loadClass with Typescript

Necro-Rohan:
**PR Title:**
`Docs: Add guidance on using loadClass with TypeScript (fixes #12813)`

**PR Description:**

**Summary**

This PR addresses [[issue #12813](https://github.com/Automattic/mongoose/issues/12813)](https://github.com/Automattic/mongoose/issues/12813), which pointed out that it's confusing for developers to figure out how to use `schema.loadClass()` correctly with TypeScript.

To fix this, I've added a new documentation section to **`docs/typescript/statics-and-methods.md`** that provides clear, official guidance.

This new section covers the complete pattern discussed in the issue thread, including:
* **The Basic Pattern:** How to manually combine types for the Schema, Class, Model (`Model<Combined> & typeof Class`), and Document (`Document & Combined`).
* **Typing `this`:** The correct way to type the `this` parameter in both instance and static methods.
* **Known Limitations:** Explicitly documents the main caveats:
    * That `this` parameters aren't supported in getters/setters (ts(2784)).
    * The "lean" caveat: how `toObject()`, `toJSON()`, or `.lean()` return plain objects without the class methods, which can lead to runtime errors.

To validate this new documentation and prevent future regressions, this PR also adds a new `tsd` test file at **`test/types/docs-loadClass.test.ts`**.

**Examples**

The new documentation page itself contains all the code examples:
* See the new `## Using schema.loadClass()` section in **`docs/typescript/statics-and-methods.md`**.

The new `tsd` test file also serves as a practical, working example of these patterns:
* See **`test/types/docs-loadClass.test.ts`**.

---
✅ Adds documentation
✅ Adds type tests
✅ Fixes #12813

### DIFF
--- a/docs/typescript/statics-and-methods.md
+++ b/docs/typescript/statics-and-methods.md
@@ -81,12 +81,12 @@ const doc = new User({ name: 'test' });
 doc.updateName('foo');
 ```
 
-# Using `loadClass()` with TypeScript
+## Using `loadClass()` with TypeScript
 
-Mongoose supports applying ES6 classes to a schema using [`schema.loadClass()`](../api/schema.html#Schema.prototype.loadClass()).
+Mongoose supports applying ES6 classes to a schema using [`schema.loadClass()`](../api/schema.html#Schema.prototype.loadClass()) as an alternative to defining statics and methods in your schema.
 When using TypeScript, there are a few important typing details to understand.
 
-## Basic Usage
+### Basic Usage
 
 `loadClass()` copies static methods, instance methods, and ES getters/setters from the class onto the schema.
 
@@ -147,7 +147,7 @@ doc.myVirtual;
 doc.property1;     
 ```
 
-## Typing `this` Inside Methods
+### Typing `this` Inside Methods
 
 You can annotate `this` in methods to enable full safety, using the [Model](../api/model.html) and [HydratedDocument](../typescript.html) types you defined.
 Note that this must be done for **each method individually**; it is not possible to set a `this` type for the entire class at once.
@@ -166,7 +166,7 @@ class MyClass {
 }
 ```
 
-## Getters / Setters Limitation
+### Getters / Setters Limitation
 
 TypeScript currently does **not** allow `this` parameters on getters/setters:
 
@@ -191,17 +191,7 @@ get myVirtual() {
 }
 ```
 
-## Feature Support Summary
-
-| Behavior                                       | Supported |
-| ---------------------------------------------- | --------- |
-| Copy instance / static methods                 | ✅         |
-| Copy getters/setters                           | ✅         |
-| Automatic TS merging                           | ❌         |
-| `this` typing in methods                       | ✅         |
-| `this` typing in getters/setters               | ❌         |
-
-## Full Example Code
+### Full Example Code
 
 ```ts
 import { Model, Schema, model, HydratedDocument } from 'mongoose';
@@ -253,13 +243,10 @@ MyModel.myStatic();
 console.log(doc.myVirtual); 
 ```
 
-## When Should I Use `loadClass()`?
+### When Should I Use `loadClass()`?
 
-`loadClass()` is useful when organizing logic in ES6 classes.
+`loadClass()` is useful for defining methods and statics in classes.
+If you have a strong preference for classes, you can use `loadClass()`; however, we recommend defining `statics` and `methods` in schema options as described in the first section.
 
-However:
-
-* ✅ works fine
-* ⚠ requires manual Typescript Types
-
-If you want better type inference, using schema options [`methods`](../guide.html#methods) and [`statics`](../guide.html#statics) instead of `loadClass` are recommended.
+The major downside of `loadClass()` in TypeScript is that it requires manual TypeScript types.
+If you want better type inference, you can use schema options [`methods`](../guide.html#methods) and [`statics`](../guide.html#statics).


docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
 // Compiles correctly
 doc.updateName('foo');
 ```
+
+
+
+# Using `loadClass()` with TypeScript
```
hasezoey:
```suggestion

# Using `loadClass()` with TypeScript
```

Only one is enough, otherwise please put it into another file.

Maybe it would be a good idea to put this whole section into <https://mongoosejs.com/docs/advanced_schemas.html> instead.

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+
+Mongoose supports applying ES6 classes to a schema using `schema.loadClass()`.
```
hasezoey: A link to the `loadClass` function would likely be good here.
<https://mongoosejs.com/docs/api/schema.html#Schema.prototype.loadClass()>

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+When using TypeScript, there are a few important typing details to understand.
+
+```ts
+import { Schema, Model, model, Document } from 'mongoose';
+```
```
hasezoey: I dont think this is necessary. Also in its current state, this is a somewhat undocumented codeblock.

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+type MyCombinedModel = Model<MyCombined> & typeof MyClass;
+type MyCombinedDocument = Document & MyCombined;
+
+const MyModel = model<MyCombinedDocument, MyCombinedModel>(
+  'MyClass',
+  schema as any
+);
+```
+
+```ts
```
hasezoey: Consider combining the codeblocks.

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+MyModel.myStatic();    // static
+const doc = new MyModel();
+doc.myMethod();        // instance
+doc.myVirtual;         // getter
+doc.property1;         // schema prop
+```
+
+---
```
hasezoey: There are headers already, is this really necessary? (and all the other places)

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+
+## Typing `this` Inside Methods
+
+You can annotate `this` in methods to enable full safety.
```
hasezoey: Maybe it would be worth to note that it is not possible to overwrite it for all methods at once and requires it being set for each method individually.

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+class MyClass {
+  // error TS2784
+  get myVirtual(this: MyCombinedDocument) {
+    return this.property1;
+  }
+}
+```
+
+This is a TypeScript limitation.
+See: [https://github.com/microsoft/TypeScript/issues/52923](https://github.com/microsoft/TypeScript/issues/52923)
```
hasezoey:
```suggestion
See: [TypeScript issue #52923](https://github.com/microsoft/TypeScript/issues/52923)
```

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+```
+
+TypeScript:
+
+```ts
+pojo.myMethod; // compiles (unsafe)
+```
+
+This is a known limitation:
+TS cannot detect when class methods are dropped.
```
hasezoey: Due to mongoose type changes, this *could* be prevented, but would require a filter type to filter out methods, though this would not really work for getters & setters (at least not cleanly).

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+| Copy getters/setters                           | ✅         |
+| Automatic TS merging                           | ❌         |
+| `this` typing in methods                       | ✅         |
+| `this` typing in getters/setters               | ❌         |
+| Methods preserved in `toObject()` / `toJSON()` | ❌         |
+| Methods preserved with `.lean()`               | ❌         |
+
+---
+
+## Recommended Pattern
```
hasezoey:
```suggestion
## Full example Code
```

I dont think `Recommended Pattern` quite fits this section, though i dont know if my suggestion is much better.

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+## When Should I Use `loadClass()`?
+
+`loadClass()` is useful when organizing logic in ES6 classes.
+
+However:
+✅ works fine
+⚠ requires manual TS merging
+⚠ methods lost in `toObject()` / `toJSON()` / `lean()`
+
+If you want better type inference, `methods` & `statics` on schema are recommended.
```
hasezoey: A link to the `methods` and `statics` options would likely be good here too
<https://mongoosejs.com/docs/guide.html#statics>
<https://mongoosejs.com/docs/guide.html#methods>

test/types/loadclass.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  type MyCombined = MySchema & MyClass;
+
+  // The model type must include statics from the class
+  type MyCombinedModel = Model<MyCombined> & typeof MyClass;
+
+  // A document must combine Mongoose Document + class + schema
+  type MyCombinedDocument = Document & MyCombined;
+
+  // Cast schema to satisfy TypeScript
+  const MyModel = model<MyCombinedDocument, MyCombinedModel>('MyClass', schema as any);
```
hasezoey, line 32: Maybe it would be worth to include those comments in the documentation too.

test/types/loadclass.test.ts
```diff
@@ -0,0 +1,178 @@
...
+    static myStatic(this: MyCombinedModel) {
+      return 42;
+    }
+
+
+    // TypeScript does NOT allow `this` parameters in getters/setters.
+    // So we show an example error here.
+    get myVirtual() {
+      // @ts-expect-error: getter does not support `this` typing
+      return this.property1;
```
hasezoey, line 73: `expectError` should likely be used in addition to the comment.

Also it would likely be worth noting that the following can just be done:
```ts
const self = this as WhateverType;
```
Though it might require a `as any` in between.

test/types/loadclass.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  const virtual = doc.myVirtual;
+  expectType<any>(virtual);
+}
+
+
+// ----------------------------------------------------------
+// Test that `toObject()` / `toJSON()` lose class behavior.
+// But TypeScript does NOT warn you about this.
+//
+// This matches the behavior described in the issue:
```
hasezoey:
```suggestion
// This matches the behavior described in #12813:
```

test/types/loadclass.test.ts
```diff
@@ -0,0 +1,178 @@
...
+      return this.name;
+    }
+  }
+
+  interface TestDoc extends TestGetter, Omit<Document, '_id'> {
+    _id: Types.ObjectId;
+  }
+}
+
+export {};
```
hasezoey:
```suggestion
```

This shouldnt be necessary in the test file.

docs/typescript/statics-and-methods.md
Necro-Rohan: Thanks! Good catch, I've removed the duplicate heading and I kept this section in statics-and-methods.md because it integrates with the surrounding TS guidance.
If maintainers prefer, I can relocate it to advanced_schemas.md.

docs/typescript/statics-and-methods.md
Necro-Rohan: Good suggestion — thank you!
I’ve added the link to the schema.loadClass() API in the updated commit.

docs/typescript/statics-and-methods.md
Necro-Rohan: Sure, i have removed the isolated import example to avoid duplication and keep the section clearer.

docs/typescript/statics-and-methods.md
Necro-Rohan: Good suggestion, thanks! I've combined those code blocks.

docs/typescript/statics-and-methods.md
Necro-Rohan: You're right, the headings are sufficient. I've removed the commented horizontal lines to clean up the formatting. Thanks!

docs/typescript/statics-and-methods.md
Necro-Rohan: Good point. I've added a note explaining that `this` must be typed for each method individually.

docs/typescript/statics-and-methods.md
Necro-Rohan: Done. I've updated the link format as you suggested.

docs/typescript/statics-and-methods.md
Necro-Rohan: Sounds good. I've renamed that section heading to `## Full Example`.

docs/typescript/statics-and-methods.md
Necro-Rohan: Agreed, adding the links will let users quickly navigate to the methods and statics docs. I’ll update it.

test/types/loadclass.test.ts
Necro-Rohan: Sure, I've copied those explanatory comments from the test file into the documentation's code block.

test/types/loadclass.test.ts
Necro-Rohan: Thanks, I've added a note to the documentation explaining the workaround for the getter limitation `(using const self = this as ...;)`.

test/types/loadclass.test.ts
Necro-Rohan: Done, I've updated the comment to reference the issue number.

test/types/loadclass.test.ts
Necro-Rohan: OK, I've removed the empty `export {}`.

Necro-Rohan:
Thanks again for all the detailed feedback @hasezoey! I've just pushed an update that addresses all the suggestions and changes you asked for.

I've also fixed those `markdownlint` errors, so the CI should be passing now.

Please let me know what you think when you get a chance! I'm happy to make any other adjustments if needed.

docs/typescript/statics-and-methods.md
hasezoey: FYI, from what i can tell, the section markings / Horizontal Rule still exists in the current version.

Necro-Rohan:
Hi @hasezoey!
Sorry about the linting issue. I’ve reviewed the file and added the missing trailing newline to comply with MD047. The problem should now be resolved.
Thank you for your patience!

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,196 @@ const doc = new User({ name: 'test' });
 // Compiles correctly
 doc.updateName('foo');
 ```
+
+# Using `loadClass()` with TypeScript
+
+Mongoose supports applying ES6 classes to a schema using [`schema.loadClass()`](https://mongoosejs.com/docs/api/schema.html#Schema.prototype.loadClass()).
```
hasezoey:
```suggestion
Mongoose supports applying ES6 classes to a schema using [`schema.loadClass()`](../api/schema.html#Schema.prototype.loadClass()).
```

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,196 @@ const doc = new User({ name: 'test' });
...
+## When Should I Use `loadClass()`?
+
+`loadClass()` is useful when organizing logic in ES6 classes.
+
+However:
+✅ works fine
+⚠ requires manual TS merging
+⚠ methods lost in `toObject()` / `toJSON()` / `lean()`
+
+If you want better type inference, [`methods`](https://mongoosejs.com/docs/guide.html#methods) & [`statics`](https://mongoosejs.com/docs/guide.html#statics) on schema are recommended.
```
hasezoey:
```suggestion
If you want better type inference, [`methods`](../guide.html#methods) & [`statics`](../guide.html#statics) on schema are recommended.
```

docs/typescript/statics-and-methods.md
hasezoey: I still dont think these Horizontal Rule's (`<hr>`) are necessary when we have headers.

test/types/loadclass.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    static myStatic(this: MyCombinedModel) {
+      return 42;
+    }
+
+
+    // TypeScript does NOT allow `this` parameters in getters/setters.
+    // So we show an example error here.
+    get myVirtual() {
+      // @ts-expect-error: getter does not support `this` typing
+      return this.property1;
```
hasezoey, line 73:
```suggestion
    get myVirtual() {
      expectError(this.property1);
      // @ts-expect-error: getter does not support `this` typing
      return this.property1;
```

I still think `expectError `should also be added in this case.

docs/typescript/statics-and-methods.md
Necro-Rohan: Thank you for the helpful suggestion.
I’ve updated the links to use the correct relative Markdown paths.
I appreciate it — I wasn’t sure with this pattern before, and it was good to learn.

test/types/loadclass.test.ts
Necro-Rohan: You're right, that should be in there. My mistake for overlooking it and the `Horizontal Rule's (<hr>)`. i have fixed it now.

test/types/loadclass.test.ts
```diff
@@ -0,0 +1,177 @@
+import { Schema, model, Document, Model, Types } from 'mongoose';
+import { expectType } from 'tsd';
```
hasezoey:
```suggestion
import { expectType, expectError } from 'tsd';
```

Forgot it needs to be imported instead of being a global.

Necro-Rohan:
@hasezoey I’ve reviewed everything again and addressed the remaining suggested changes.
Please feel free to share any additional suggestions or feedback.
Thank you!

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+
+// `loadClass()` does NOT update TS types automatically.
+// So we must manually combine schema fields + class members.
+type MyCombined = MySchema & MyClass;
+
+// The model type must include statics from the class
+type MyCombinedModel = Model<MyCombined> & typeof MyClass;
+
+// A document must combine Mongoose Document + class + schema
+type MyCombinedDocument = Document & MyCombined;
```
vkarpov15: Use HydratedDocument<> instead here

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+interface MySchema {
+  property1: string;
+}
+
+// `loadClass()` does NOT update TS types automatically.
+// So we must manually combine schema fields + class members.
+type MyCombined = MySchema & MyClass;
+
+// The model type must include statics from the class
+type MyCombinedModel = Model<MyCombined> & typeof MyClass;
```
vkarpov15: You should do Model<MySchema> here, you don't want the raw doc interface to include methods etc. That's why you're running into the toJSON() issue

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+
+const schema = new Schema({ property1: String });
+schema.loadClass(MyClass);
+```
+
+Mongoose does **not** automatically update TypeScript types for class members.
+To get full type support, you must manually merge types.
+
+```ts
+interface MySchema {
```
vkarpov15: Prefer to call this `RawDocType` or something similar since this isn't a mongoose Schema

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+}
+```
+
+## `toObject()` / `toJSON()` Caveat
+
+`loadClass()` attaches methods at runtime.
+However, `toObject()` and `toJSON()` return **plain JavaScript objects** without these methods.
+
+```ts
+const doc = new MyModel({ property1: 'test' });
```
vkarpov15: Wouldn't happen if you passed in the raw document type

vkarpov15 (commented):
Overall I think something more like the following would be better, @hasezoey WDYT?

```ts
interface RawDocType {
  property1: string;
}

// `loadClass()` does NOT update TS types automatically.
// So we must manually combine schema fields + class members.
// The model type must include statics from the class
type MyCombinedModel = Model<RawDocType, {}, Pick<MyClass, 'myMethod'>, Pick<MyClass, 'myVirtual'>> & Pick<typeof MyClass, 'myStatic'>;

// A document must combine Mongoose Document + class + schema
type MyCombinedDocument = mongoose.HydratedDocument<RawDocType, Pick<MyClass, 'myMethod'>, {}, Pick<MyClass, 'myVirtual'>>;

const schema = new Schema({ property1: String });
schema.loadClass(MyClass);

// Cast schema to satisfy TypeScript
const MyModel = model<MyCombinedDocument, MyCombinedModel>(
  'MyClass',
  schema 
);
```

Another alternative I'd like to consider is to is recommending using `extends` on model instances as a way of handling methods, statics, and virtuals. The following works correctly for methods and statics, but has some caveats with virtuals.

```ts
class MyClass extends TestModel {
  myMethod() {
    return 'hello from myMethod ' + this.name;
  }

  static myStatic() {
    return 'hello from myStatic';
  }

  get myVirtual() {
    return 'hello from myVirtual';
  }
}

const doc = new MyClass({ name: 'test' });
// All work
console.log(doc.myMethod());
console.log(MyClass.myStatic());
console.log(doc.myVirtual);

// Caveat: output does NOT include myVirtual because myVirtual technically isn't a virtual here
console.log(doc.toObject());
```

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+`loadClass()` is useful when organizing logic in ES6 classes.
+
+However:
+- ✅ works fine
+- ⚠ requires manual TS merging
+- ⚠ methods lost in `toObject()` / `toJSON()` / `lean()`
+
+If you want better type inference, [`methods`](../guide.html#methods) & [`statics`](../guide.html#statics) on schema are recommended.
+
+---
```
hasezoey:
```suggestion

```

hasezoey (approved):
Looks good to me in its current state as a initial guide. I agree with @vkarpov15's suggestions though.

FYI:
> `type MyCombinedModel = Model<RawDocType, {}, Pick<MyClass, 'myMethod'>, Pick<MyClass, 'myVirtual'>> & Pick<typeof MyClass, 'myStatic'>;`

Those `Pick`s are what i meant with "would require a Filter type", or a lot of manual typing.

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,186 @@ const doc = new User({ name: 'test' });
...
+
+type MyCombined = MySchema & MyClass;
+type MyCombinedModel = Model<MyCombined> & typeof MyClass;
+type MyCombinedDocument = Document & MyCombined;
+
+const MyModel = model<MyCombinedDocument, MyCombinedModel>(
+  'MyClass',
+  schema as any
+);
+```
```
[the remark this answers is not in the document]
Necro-Rohan: This is a great point regarding readability.
I explored reordering the declarations (as Copilot suggested), but it unfortunately creates a circular-dependency issue.
The combined types (e.g., MyCombined) depend on the MyClass definition, so they must be declared after the class. If moved before, TypeScript fails because MyClass is not yet defined.
Given that, I think the current order is the most correct.If helpful, I can add a clarifying comment like:
```
interface MySchema {
  property1: string;
}

// 1. Declare the class first.
// It's OK to reference types (MyCombinedDocument) that aren't defined yet
// in 'this' annotations because of how TypeScript handles hoisting.
class MyClass {
  myMethod(this: MyCombinedDocument) {
    return this.property1;
  }
  static myStatic(this: MyCombinedModel) {
    return 42;
  }
}

// 2. Now define the combined types.
// This works because 'MyClass' has been declared.
type MyCombined = MySchema & MyClass;
type MyCombinedModel = Model<MyCombined> & typeof MyClass;
type MyCombinedDocument = Document & MyCombined;

// 3. Continue with the schema and model.
const schema = new Schema<MySchema>({ property1: String });
schema.loadClass(MyClass);

const MyModel = model<MyCombinedDocument, MyCombinedModel>(
  'MyClass',
  schema as any
);
```
@vkarpov15  Please let me know what you think — happy to update!

docs/typescript/statics-and-methods.md
Necro-Rohan: Thanks, I appreciate the correction! I'll update the example to use `HydratedDocument` as suggested.

docs/typescript/statics-and-methods.md
Necro-Rohan: Ah, that makes perfect sense. Thanks for the clarification, I'll make that change

docs/typescript/statics-and-methods.md
Necro-Rohan: Done. Switched to `RawDocType`.

docs/typescript/statics-and-methods.md
Necro-Rohan: You're absolutely right. I've updated the docs to reflect this, now the types work as expected now. Thanks!

Necro-Rohan:
Thank you @vkarpov15 and @hasezoey for the detailed review!
I’ve updated the documentation based on your suggestions:

- Switched to RawDocType
- Used `HydratedDocument` and `Model generics`
- Updated `model()` to use `RawDocType` as the first generic
- Revised the `toObject()` section so the types behave as expected
- Fixed the markdown linting issue

Everything should be up to date now.
Whenever you get a moment, I’d appreciate it if you could take another look and let me know if anything else needs adjusting.

Thanks again!

docs/typescript/statics-and-methods.md
Necro-Rohan: done

docs/typescript/statics-and-methods.md
```diff
@@ -185,31 +191,20 @@ get myVirtual() {
...
 ```
 
-## `toObject()` / `toJSON()` Caveat
+## `toObject()` / `toJSON()`
 
-`loadClass()` attaches methods at runtime.
-However, `toObject()` and `toJSON()` return **plain JavaScript objects** without these methods.
+When using the correct [Model](../api/model.html) and [HydratedDocument](../typescript.html) generics (as shown above), Mongoose's types for `toObject()` and `toJSON()` work as expected. They will correctly return a type that includes **only the raw data** (e.g., `RawDocType`), not the methods or virtuals.
+
+This is a feature, as it accurately reflects the plain object returned at runtime and prevents you from trying to access methods that don't exist.
```
hasezoey: `This is a feature,` reads somewhat weird here.

docs/typescript/statics-and-methods.md
```diff
@@ -225,41 +220,63 @@ TS cannot detect when class methods are dropped.
...
   property1: string;
 }
 
 class MyClass {
   myMethod(this: MyCombinedDocument) {
     return this.property1;
   }
-  static myStatic(this: MyCombinedModel) {
+
+  static myStatic() {
```
hasezoey: No `this` overwrite?

docs/typescript/statics-and-methods.md
```diff
@@ -225,41 +220,63 @@ TS cannot detect when class methods are dropped.
...
 `loadClass()` is useful when organizing logic in ES6 classes.
 
 However:
-- ✅ works fine
-- ⚠ requires manual TS merging
-- ⚠ methods lost in `toObject()` / `toJSON()` / `lean()`
 
-If you want better type inference, [`methods`](../guide.html#methods) & [`statics`](../guide.html#statics) on schema are recommended.
+* ✅ works fine
+* ⚠ requires manual TS merging
```
hasezoey:
```suggestion
* ⚠ requires manual Typescript Types
```

docs/typescript/statics-and-methods.md
```diff
@@ -225,41 +220,63 @@ TS cannot detect when class methods are dropped.
...
+* ⚠ methods lost in `toObject()` / `toJSON()` / `lean()`
```
hasezoey: This is not accurate anymore. (and with the current writing, why would methods be "not lost" be good?)

docs/typescript/statics-and-methods.md
```diff
@@ -225,41 +220,63 @@ TS cannot detect when class methods are dropped.
...
 
----
+If you want better type inference, [`methods`](../guide.html#methods) & [`statics`](../guide.html#statics) on schema are recommended.
```
hasezoey:
```suggestion
If you want better type inference, using schema options [`methods`](../guide.html#methods) and [`statics`](../guide.html#statics) instead of `loadClass` are recommended.
```

docs/typescript/statics-and-methods.md
Necro-Rohan: Now that you’ve pointed it out, it does read a bit awkward. I’ll remove that part.

docs/typescript/statics-and-methods.md
Necro-Rohan: Ah, you're right. I missed that. Fixing it now, thanks!

docs/typescript/statics-and-methods.md
Necro-Rohan: done

docs/typescript/statics-and-methods.md
Necro-Rohan: You're absolutely right. That bullet point is a leftover from the old example and contradicts the new behavior we're using now. I'll remove it. Thanks for the catch!

docs/typescript/statics-and-methods.md
Necro-Rohan: done

Necro-Rohan:
Hi @vkarpov15 @hasezoey, pls review this PR when convenient. Thanks!

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,202 @@ const doc = new User({ name: 'test' });
...
+MyModel.myStatic();
+const doc = new MyModel();
+doc.myMethod();
+doc.myVirtual;
+doc.property1;     
+```
+
+## Typing `this` Inside Methods
+
+You can annotate `this` in methods to enable full safety, using the [Model](../api/model.html) and [HydratedDocument](../typescript.html) types you defined.
```
vkarpov15, line 152: Do we need to set the type of `this` in methods and statics? In my brief experiment it seems like Mongoose sets these correctly.

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,202 @@ const doc = new User({ name: 'test' });
...
+
+```ts
+get myVirtual() {
+  // Workaround: cast 'this' to your document type
+  const self = this as MyCombinedDocument;
+  return `Name: ${self.property1}`;
+}
+```
+
+## `toObject()` / `toJSON()`
```
vkarpov15: I don't think this section is necessary anymore given that methods are no longer on `toObject()` output. This section effectively says "toObject() and toJSON() behave the same as everywhere else" which isn't useful IMO.

docs/typescript/statics-and-methods.md
```diff
@@ -80,3 +80,202 @@ const doc = new User({ name: 'test' });
...
+## Limitations
+
+| Behavior                                       | Supported |
+| ---------------------------------------------- | --------- |
+| Copy instance / static methods                 | ✅         |
+| Copy getters/setters                           | ✅         |
+| Automatic TS merging                           | ❌         |
+| `this` typing in methods                       | ✅         |
+| `this` typing in getters/setters               | ❌         |
+| Methods preserved in `toObject()` / `toJSON()` | ❌         |
```
vkarpov15: I don't think this is a limitation. Neither is `lean()`. In general I'd just remove this section.

docs/typescript/statics-and-methods.md
Necro-Rohan: I've double-checked this, and you're absolutely right — Mongoose correctly handles the `this` context at runtime. However, the explicit `this` annotation (e.g., `this: MyCombinedDocument`) is still important for TypeScript during compile time. Without it, TypeScript has no way to infer that `this.property1` is valid inside the class, since property1 isn’t actually defined on MyClass itself.
So I think it’s best to keep this section in place, as it provides important context and maintains type safety.
Thanks again for pointing this out!

docs/typescript/statics-and-methods.md
Necro-Rohan: You're absolutely right. With the latest changes, that part ended up being redundant — it was just describing standard Mongoose behavior and adding extra clutter. I've removed it. Thanks for pointing it out!

docs/typescript/statics-and-methods.md
Necro-Rohan: Excellent point, The `toObject() and lean()` behaviors aren’t actually “limitations” of `loadClass()` they’re just how Mongoose works. I’ve removed those rows to avoid any confusion, and I’m also renaming the table to “**Feature Support Summary**” since “Limitations” felt a bit misleading.

Necro-Rohan:
@vkarpov15 I’ve made the updates based on your feedback.

- Removed the redundant toObject() / toJSON() section
- Renamed Limitations to Feature Support Summary and removed the toObject / lean rows

The PR has been updated and is ready for another look whenever you have a moment. Thanks again!

PR #15742, opened 2025-11-14
Add benchmark for large nested array documents (related to #9588)

### LINKED ISSUE #9588: nodejs CPU usage spike when saving large documents with mongoose
Bug probably

When inserting many large documents experiencing a spike in CPU

Here is the schema I use:

```
{
  ticker: String,
  asks: [[Number]],
  bids: [[Number]],
  timestamp: String,
}
```

Asks and bids field can have up to 1000 arrays inside the outer array.
Both .save and insertMany cause this behavior.

CPU spike should not occur.

Node v12.19.0
Mongoose 5.10.1
MongoDB 4.4.0

Thank you for help!

Kundan-CR7:
### Summary

This PR adds a new benchmark `benchmarks/nestedArrayLarge.js` that measures
Mongoose’s performance when constructing documents with large nested arrays,
matching the structure used in issue #9588 (`asks: [[Number]]`, `bids: [[Number]]`).

This benchmark also helps verify that the performance issue described in #9588
is no longer reproducible in current Mongoose versions.

While the original issue showed a significant CPU spike in Mongoose 5.x,
modern Mongoose (v8.x) performs well. This benchmark provides ongoing regression
coverage for nested-array casting performance.

### What this benchmark does

- Follows the same structure and conventions used in existing Mongoose benchmarks

- Creates a schema with nested arrays (`[[Number]]`)
- Generates 10,000 nested arrays in the document
- Measures construction performance using Benchmark.js
- Logs results when `MONGOOSE_DEV=1` or `PULL_REQUEST=1`
- Matches benchmark formatting used in the Mongoose repo

### Why this is useful

Issue #9588 was originally caused by heavy casting and atomic tracking in nested arrays.
Multiple optimizations since then have improved performance significantly.

This benchmark ensures:

- Nested-array casting remains fast in future releases  
- No regressions occur in deep array performance  
- Mongoose maintains visibility into a historically problematic case  

### Related Issue

Closes #9588 (performance regression appears resolved in Mongoose 8.x)

### DIFF
--- a/benchmarks/nestedArrayLarge.js
+++ b/benchmarks/nestedArrayLarge.js
@@ -45,6 +45,6 @@ async function run() {
         'save() time ms': +(saveEnd - saveStart).toFixed(2)
     };
 
-    console.log(JSON.stringify(results, null, ' '));
+    console.log(JSON.stringify(results, null, '  '));
     process.exit(0);
 }


vkarpov15 (changes requested):
Our current benchmarks don't use the `benchmark` npm package, see `benchmarks/createDeepNestedDocArray.js` for an example of how we build benchmarks now

Kundan-CR7:
@vkarpov15  Thanks for the clarification! I've updated the benchmark to follow the current style used in `benchmarks/createDeepNestedDocArray.js`, Let me know if you'd like any further adjustments

vkarpov15 (approved):

PR #15743, opened 2025-11-15
Added a new FAQ entry right before the final “Something to add?” sect…

Mario5T:
### [docs/faq.md](cci:7://file:///Users/adityasingh/Documents/axios/mongoose/docs/faq.md:0:0-0:0)

- **Added** a new FAQ entry right before the final “Something to add?” section.
- **Anchor ID**: `divergent-array-error`
- **Content covers:**

  - **When [DivergentArrayError](cci:2://file:///Users/adityasingh/Documents/axios/mongoose/lib/error/divergentArray.js:15:0-29:1) occurs:**
    - Array selected using `$elemMatch`.
    - Array populated with `populate()` plus `skip` / `limit` / query conditions or excluding `_id`.
    - `save()` would cause a `$set` / `$pop` on the **entire** array.

  - **Why it happens:**
    - The array is only **partially loaded** in memory.
    - Updating it via `save()` would risk overwriting elements that weren’t loaded, so Mongoose throws [DivergentArrayError](cci:2://file:///Users/adityasingh/Documents/axios/mongoose/lib/error/divergentArray.js:15:0-29:1) to prevent data loss.

  - **How to fix:**
    - **Option 1: Re-load the full array and then call `save()`:**
      ```js
      const doc = await BlogPost.findById(_id);
      doc.comments.id(commentId).text = 'Updated';
      await doc.save();
      ```

    - **Option 2: Use `updateOne()` / `updateMany()` with positional operators / `arrayFilters`:**
      ```js
      await BlogPost.updateOne(
        { _id, 'comments._id': commentId },
        { $set: { 'comments.$.text': 'Updated' } }
      );
      ```

  - **Note**:
    - The same guidance applies if you populated an array with `skip`, `limit`, query conditions, or excluded `_id`:  
      avoid calling `save()` on that partially loaded array; instead, re-query without those options or use an update operation as above.

---

### [lib/error/divergentArray.js](cci:7://file:///Users/adityasingh/Documents/axios/mongoose/lib/error/divergentArray.js:0:0-0:0)

- **Updated the error message** to include a direct link to the new FAQ entry:
  - Appended:
    ```text
    See [https://mongoosejs.com/docs/faq.html#divergent-array-error](https://mongoosejs.com/docs/faq.html#divergent-array-error) for more information.
    ```
- **Removed** the old comment:
  ```js
  // TODO write up a docs page (FAQ) and link to it

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -456,7 +456,7 @@ await doc.save();
 
 To fix this error, either:
 
-1) Load the full array before modifying and saving:
+(1) Load the full array before modifying and saving:
 
 ```javascript
 const doc = await BlogPost.findById(_id); 
@@ -464,7 +464,7 @@ doc.comments.id(commentId).text = 'Updated';
 await doc.save();
 ```
 
-2) Or use `updateOne()` / `updateMany()` with positional operators or `arrayFilters` so MongoDB can update the array atomically without requiring the full array on the document:
+(2) Or use `updateOne()` / `updateMany()` with positional operators or `arrayFilters` so MongoDB can update the array atomically without requiring the full array on the document:
 
 ```javascript
 await BlogPost.updateOne(


Mario5T:
@vkarpov15 plz review the changes made

vkarpov15 (approved):
Thanks :+1:

Issue #3935, opened 2016-03-02
Automatic cast to date from boolean is made instead of throwing CastError.


dailypixie:
Doesn't throw a cast error when a date field is set as true or false. This happens from version 4.4.0

var Schema = mongoose.Schema({
  acquired: {type: Date, required: false},
...
});

var elements= [{
 acquired: true,
...
  }];
var model = mongoose.model('app', Schema , 'Apps');
var instances = new model(elements); //this should throw a cast error.

PR #15747, opened 2025-11-16
Optimize test efficiency in schema.test.js

Vegapunk-debug:
**Summary**

This PR optimizes the test **"booleans cause cast error for date (gh-3935)"** in `test/schema.test.js` by removing redundant calls to `test.validateSync()`.  
Previously, the test invoked `validateSync()` twice within its assertions, which was unnecessary.  
The updated approach stores the result in `validationResult` and reuses it, improving test efficiency.

**Examples**

This is a test-only optimization and does not introduce new features or modify runtime behavior.  
No additional examples are required, and the existing test continues to pass with the updated implementation.


test/schema.test.js
```diff
@@ -1645,12 +1645,11 @@ describe('schema', function() {
...
     const test = new Test({ test: true });
 
-    assert.ok(test.validateSync());
-    assert.equal(test.validateSync().errors.test.name, 'CastError');
+    const validationResult = test.validateSync();
+    assert.ok(validationResult);
+    assert.equal(validationResult.errors.test.name, 'CastError');
 
 
-  });
```
vkarpov15, line 1652: Removing this closing brace seems to cause a syntax error?

PR #15749, opened 2025-11-17
Fixed Global autoCreate option not applied to models on createConnection()

### LINKED ISSUE #15748: Global autoCreate option not applied to models on createConnection()
### Prerequisites

### Mongoose version

6.x.x

### Node.js version

14.x,

### MongoDB server version

5.x

Not using TypeScript for this reproduction (JavaScript only)

### Description

When mongoose.set('autoCreate', true) is used, models registered on the default connection correctly auto-create their collections.
However, models registered on a new connection created via mongoose.createConnection() do not inherit this global option.

This results in inconsistent behaviour between connections.

### Steps to Reproduce

```import mongoose from 'mongoose';

mongoose.set('autoCreate', true);

const uri = 'mongodb://localhost:27017/test';

const User1 = mongoose.model('User1', new mongoose.Schema({ name: String }));

const conn = mongoose.createConnection(uri);
const User2 = conn.model('User2', new mongoose.Schema({ name: String }));

await mongoose.connection.asPromise();
await conn.asPromise();

await User1.init(); // ✓ collection created
await User2.init(); // ✗ collection NOT created
```
Expected: Both collections should be created.
Actual: Only the model on the default connection respects autoCreate.

**Root Cause**
createConnection() initialises its own connection.options without inheriting mongoose.options, so global settings like autoCreate never reach models created on additional connections.

### Expected Behavior

**Proposed Fix**

Merge global and connection-level options inside the Connection constructor:

```this.options = Object.assign({}, base.options, options);```

I’ve already implemented and tested this locally.
If this behavior is confirmed as unintended, I’d be happy to open a PR with the fix and accompanying test.

KirtiGautam620:
Ensure global Mongoose options like `autoCreate` are inherited by additional connections

**Related Issue:**  
Fixes #15748 – Global `autoCreate` option not applied to models on `createConnection()`

**Description:**  
This PR addresses an issue where global Mongoose options, such as `autoCreate`, were not inherited by connections created via `mongoose.createConnection()`. Previously, only models on the default connection respected these global settings, causing inconsistent behavior between default and additional connections.

**Changes Made:**  
- Updated the `Connection` constructor (`lib/connection.js`) to merge global options from `base.options` into `this.options` using:  
  ```js
  this.options = Object.assign({}, base.options);

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -68,7 +68,7 @@ function Connection(base) {
   this.models = {};
   this.config = {};
   this.replica = false;
-  this.options = Object.assign({}, base.options);
+  this.options = Object.assign({}, base?.options);
   this.otherDbs = []; // FIXME: To be replaced with relatedDbs
   this.relatedDbs = {}; // Hashmap of other dbs that share underlying connection
   this.states = STATES;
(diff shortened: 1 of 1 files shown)


vkarpov15:
I'm just going to close this PR, it is a mess - wrong branch, unnecessary changes, doesn't even seem like it solves the original issue.

PR #15751, opened 2025-11-17
types: fix auto inferred objectid

vkarpov15:
**Summary**

Little gotcha I found with Mongoose 9 types: `myId: { type: mongoose.Types.ObjectId }` fails. While we don't recommend that sort of type definition, we do support it, so this definition should also work in TypeScript.

**Examples**


hasezoey:
Maybe add a test for it as well?

PR #15754, opened 2025-11-18
perf(query): skip recursive casting for simple equality filters

manak-sharma20:
## Summary

This PR introduces a conservative performance optimization in `Query.prototype._castConditions`.  
For simple equality filters (e.g. `{ name: "Alice" }`, `{ age: 20 }`), Mongoose currently performs the full recursive casting pipeline, which adds noticeable overhead compared to the native MongoDB driver.

This PR adds a fast-path that skips recursive casting when the filter meets all of the following conditions:
- plain object (not an array)
- no operator keys (`$...`)
- values are primitives (`string`, `number`, `boolean`), `RegExp`, or `null`
- the matching schema paths are **not** ObjectId fields (to avoid skipping required ObjectId casting)

If all conditions are satisfied, `_castConditions` returns early and avoids the expensive casting process.

---

## Motivation

Mongoose users have reported slower query performance compared to the native driver, especially for frequently-used simple lookups such as:
```js
Model.find({ name: "Alice" })
Model.findOne({ email: "abc@test.com" })

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -2368,7 +2368,37 @@ Query.prototype._castConditions = function() {
 
   if (sanitizeFilterOpt) {
     sanitizeFilter(this._conditions);
+
   }
+  function _isSimpleEqualityFilter(filter, model) {
+    if (!filter || typeof filter !== 'object' || Array.isArray(filter)) return false;
+    const keys = Object.keys(filter);
+    if (keys.length === 0) return false;
+
+    for (let i = 0; i < keys.length; ++i) {
+      const k = keys[i];
+      if (k.charAt(0) === '$') return false;
+
+      const v = filter[k];
+      if (v === null) continue;
+
+      const t = typeof v;
+      if (t === 'string' || t === 'number' || t === 'boolean' || v instanceof RegExp) {
+        const schemaType = model && model.schema ? model.schema.path(k) : null;
+        if (schemaType && schemaType.instance && /ObjectId/i.test(schemaType.instance)) return false;
+        continue;
+      }
+
+      return false;
+    }
+
+    return true;
+  }
+
+  if (_isSimpleEqualityFilter(this._conditions, this.model)) {
+    return;
+  }
+
 
   try {
     this.cast(this.model);

### CODE REVIEW

**Comment by manak-sharma20** [comment]
@vkarpov15 can you approve this please.

File: lib/query.js [hunk 1]
```diff
@@ -2368,7 +2368,37 @@ Query.prototype._castConditions = function() {
 
   if (sanitizeFilterOpt) {
     sanitizeFilter(this._conditions);
+
   }
+  function _isSimpleEqualityFilter(filter, model) {
+    if (!filter || typeof filter !== 'object' || Array.isArray(filter)) return false;
+    const keys = Object.keys(filter);
```
billouboq, line 2375: isn't that expensive to create an array instead of a for loop ?

manak-sharma20: @billouboq 
This approach is crucial for correctly validating the filter structure because it offers two key safety guarantees:
It guarantees iteration only over the object's own enumerable properties.
It automatically excludes inherited properties (which a standard for...in loop would include without an extra hasOwnProperty check).

lib/query.js
```diff
@@ -2368,7 +2368,37 @@ Query.prototype._castConditions = function() {
...
+    for (let i = 0; i < keys.length; ++i) {
+      const k = keys[i];
+      if (k.charAt(0) === '$') return false;
+
+      const v = filter[k];
+      if (v === null) continue;
+
+      const t = typeof v;
+      if (t === 'string' || t === 'number' || t === 'boolean' || v instanceof RegExp) {
+        const schemaType = model && model.schema ? model.schema.path(k) : null;
```
vkarpov15, line 2387: This doesn't check if the `t` lines up with `schemaType`. If you pass in `email: 42` this will skip casting number `42` to a string.

lib/query.js
```diff
@@ -2368,7 +2368,37 @@ Query.prototype._castConditions = function() {
...
+        if (schemaType && schemaType.instance && /ObjectId/i.test(schemaType.instance)) return false;
```
vkarpov15, line 2388: Also wouldn't handle if `schemaType` is an object, array, or map.

vkarpov15 (changes requested):
This PR skips all casting if you have only string/number/boolean properties in your filter, which is incorrect behavior.

Can you please put together a benchmark script which demonstrates a case where this change would make for a significant performance improvement?

vkarpov15:
I'm going to close this PR since I don't think it is salvageable.

PR #15765, opened 2025-11-19
fix(docs): prevent parser crash when prop.tags is undefined

### LINKED ISSUE #15764: Docs parser crashes when prop.tags is undefined
### Prerequisites

### Mongoose version

7.6.3

### Node.js version

18.x

### MongoDB server version

6.0.x

Not using TypeScript

### Description

While working on Mongoose documentation generation, I found that the docs parser crashes when a property does not contain the `tags` array.

In `docs/source/api.js`, the loop assumes `prop.tags` always exists.  
However, in some parsed doc blocks, `prop.tags` is **undefined**, which causes a runtime error:

TypeError: Cannot read properties of undefined (reading 'Symbol(Symbol.iterator)')

The crash happens before documentation pages finish building.

### Steps to Reproduce

1. Clone the mongoose repository
2. Install dependencies: `npm install`
3. Run the docs build script: `npm run docs:build`
4. Build fails with:

   TypeError: Cannot read properties of undefined (reading 'Symbol(Symbol.iterator)')
      at docs/source/api.js:XXX

### Expected Behavior

The docs parser should safely skip properties without `tags` instead of crashing.  
Documentation generation should complete without errors.

THE-Amrit-mahto-05:
### Summary

The docs parser crashed when it encountered documentation blocks where `prop.tags` was missing or undefined.  
This caused the following runtime error during docs generation:

TypeError: Cannot read properties of undefined (reading 'Symbol(Symbol.iterator)')

This PR adds a guard:

```js
if (!prop.tags) continue;
```

to safely skip entries without tags and prevent the crash.
Fixes #15764

### DIFF
--- a/docs/source/api.js
+++ b/docs/source/api.js
@@ -269,7 +269,7 @@ function processFile(props) {
     if ('constructor' in ctx && ctx.constructor === undefined) {
       ctx.constructorWasUndefined = true;
     }
-
+    if (!prop.tags) continue;
     for (const __tag of prop.tags) {
       // the following has been done, because otherwise no type could be applied for intellisense
       /** @type {TagObject} */


THE-Amrit-mahto-05:
@vkarpov15  please review it

vkarpov15 (approved):
Thanks :+1:

PR #15782, opened 2025-11-22
fix(bulkWrite): pass overwriteImmutable option to castUpdate fixes #15781

### LINKED ISSUE #15781: Timestamp `createdAt` can't be updated using `bulkWrite`
### Prerequisites

### Mongoose version

7.x.x, 8.x.x

### Node.js version

20.x

### MongoDB server version

8.x

_No response_

### Description

I'm trying to update the `createdAt` field of some documents using `bulkWrite` and it's not possible. It works if I use `updateOne` directly as a function but if I do it through a `bulkWrite` it doesn't work.

This only affects `createdAt`, `updatedAt` can be updated.

So far I have tested it using Mongoose v7 and v8 and both have the same issue, I would appreciate it if this could be fixed in both versions 🙇🏽‍♂️.

### Steps to Reproduce

```js
const personSchema = new mongoose.Schema(
  {
  },
  { timestamps: true }
);

const PersonModel = mongoose.model("Person", personSchema);

const person = new PersonModel({
  name: "John",
});
await person.save();

// With `bulkWrite` it doesn't work
await PersonModel.bulkWrite([
  {
    updateOne: {
      filter: { name: "John" },
      update: {
        createdAt: new Date(0),
      },
      options: {
        timestamps: false,
        overwriteImmutable: true,
      },
    },
  },
]);

// Using `updateOne` directly it works
// await PersonModel.updateOne(
//   { name: "John" },
//   { createdAt: new Date(0) },
//   { timestamps: false, overwriteImmutable: true }
// );
```

### Expected Behavior

I expect to be able to override the `createdAt` field if needed.

jhaayushkumar:
Fixes #15781

**Summary**

When using `bulkWrite` with `updateOne` or `updateMany` operations, the `overwriteImmutable` option was not being passed through to `castUpdate`. This prevented users from updating immutable timestamp fields like `createdAt` even when explicitly setting `overwriteImmutable: true` and `timestamps: false`.

The issue was that in `castBulkWrite.js`, the `overwriteImmutable` option from the operation was not included in the options object passed to `castUpdate`, causing `handleImmutable` to remove the immutable field from the update.

This PR fixes the issue by:
1. Passing `overwriteImmutable` option from the operation to `castUpdate` in both `castUpdateOne` and `castUpdateMany`
2. Passing `timestamps` option to `applyTimestampsToUpdate` to prevent deletion of `createdAt` when `timestamps: false` is set

**Examples**

```javascript
const personSchema = new mongoose.Schema(
  { name: { type: String, required: true } },
  { timestamps: true }
);
const PersonModel = mongoose.model('Person', personSchema);

const person = await PersonModel.create({ name: 'John' });

// Before this fix: createdAt would NOT be updated
// After this fix: createdAt IS updated to new Date(0)
await PersonModel.bulkWrite([{
  updateOne: {
    filter: { name: 'John' },
    update: { createdAt: new Date(0) },
    timestamps: false,
    overwriteImmutable: true
  }
}]);

### DIFF
--- a/lib/helpers/model/castBulkWrite.js
+++ b/lib/helpers/model/castBulkWrite.js
@@ -118,7 +118,9 @@ module.exports.castUpdateOne = function castUpdateOne(originalModel, updateOne, 
   if (model.schema.$timestamps != null && doInitTimestamps) {
     const createdAt = model.schema.$timestamps.createdAt;
     const updatedAt = model.schema.$timestamps.updatedAt;
-    applyTimestampsToUpdate(now, createdAt, updatedAt, update, {});
+    applyTimestampsToUpdate(now, createdAt, updatedAt, update, {
+      timestamps: updateOne.timestamps
+    });
   }
 
   if (doInitTimestamps) {
@@ -150,7 +152,8 @@ module.exports.castUpdateOne = function castUpdateOne(originalModel, updateOne, 
     strict: strict,
     upsert: updateOne.upsert,
     arrayFilters: updateOne.arrayFilters,
-    overwriteDiscriminatorKey: updateOne.overwriteDiscriminatorKey
+    overwriteDiscriminatorKey: updateOne.overwriteDiscriminatorKey,
+    overwriteImmutable: updateOne.overwriteImmutable
   }, model, updateOne['filter']);
 
   return updateOne;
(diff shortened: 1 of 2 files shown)


jhaayushkumar:
@vkarpov15  please review and suggest changes if requires

gaastonsr:
Ty!

PR #15787, opened 2025-11-24
Optimized lib/utils.js and Updated test/utils.test.js

yogesh968:
#### 1️⃣ Improved `lib/utils.js`
- **Removed** a redundant `for` loop that checked whether two objects had keys in the *exact same order*.
- This check introduced an unnecessary **O(N)** time complexity.
- In standard deep equality comparisons, **key order does not matter**.  
  Example:  
  ```js
  { a: 1, b: 2 } === { b: 2, a: 1 }
  ```
- By removing the loop:
  - >The function became faster
  - > The logic became cleaner
  - > Behavior now aligns with typical deep equality expectations

#### 2️⃣ Updated `test/utils.test.js`
- A test previously expected `deepEqual` to return **false** when key order differed.
- Since key order is no longer a factor, `deepEqual` now correctly returns **true**.
- The test was updated to reflect the correct behavior.
#### 2️⃣ Updated `test/utils.test.js`
- I have done better formatting of code adding some whitespaces and remove the extra-spaces.

### 🚀 Why This Matters
- Eliminates unnecessary computation
- Avoids misleading equality results
- Makes the utility more predictable and consistent
- Improves overall performance without changing any functional outcomes

###  Result
- **Correct deep equality behavior**
- **Reduced time complexity**
- **Cleaner, more maintainable code**

---

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -44,7 +44,7 @@ const manySpaceRE = /\s+/;
  * @api private
  */
 
-exports.toCollectionName = function(name, pluralize) {
+exports.toCollectionName = function (name, pluralize) {
   if (name === 'system.profile') {
     return name;
   }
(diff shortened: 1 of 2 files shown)


yogesh968:
@wvl @svperfecta @Sija @lackac @vkarpov15 can you please see it ..

yogesh968:
Hi team 👋

I’ve submitted an optimization to `deepEqual` that improves performance and aligns with standard object equality behavior. All tests have been updated accordingly.

Whenever you get a chance, could someone please review and approve the workflows so the checks can run?

Thank you!

vkarpov15:
1. There are certain cases where key order does matter in Mongoose, for example when defining indexes: `schema.index({ name: 1, age: 1 })` is a different index than `schema.index({ age: 1, name: 1 })`
2. Please don't commit all these formatting changes, that makes code review more difficult

PR #15788, opened 2025-11-25
fix(test): stop retry loop after successful clearTestData() drop

yogesh968:
# PR: fix(test): stop retry loop after successful clearTestData() drop

## Summary
`clearTestData()` wrapped `db.dropDatabase()` in a retry loop to swallow intermittent "operation was interrupted" write-concern errors from replica-set tests. The original loop continued iterating even after a retry succeeded because it lacked a `return` when `_inner()` resolved. A later iteration could run `dropDatabase()` again and fail, making cleanups appear flaky.

This PR adds an early `return` immediately after a successful `_inner()` resolution so the function exits as soon as a drop succeeds. The existing conditional logging and rethrow behavior is preserved:
- If the retryable write-concern error occurs **and** attempts remain → log & continue.
- Any non-retryable error (or the retryable error with no attempts left) → rethrow immediately.

## Problem
- Successful retry could be overwritten by a subsequent failed iteration.
- `afterEach()` cleanup became non-deterministic; tests sporadically failed.

## Fix
- Add `return` right after `_inner()` resolves.
- Keep existing retry/backoff, logging, and rethrow semantics unchanged.

## Why this is safe
- Behavior on errors is unchanged.
- Only prevents unnecessary additional `dropDatabase()` calls after success.
- Restores deterministic cleanup behavior for tests in replica-set environments.

## Suggested commit message
`fix(test): return early in clearTestData retry loop to avoid overwriting a successful drop`
/cc @vkarpov15

### DIFF
--- a/test/util.js
+++ b/test/util.js
@@ -19,8 +19,10 @@ exports.clearTestData = async function clearTestData(db) {
     retries -= 1;
     try {
       await _inner();
+      return;
     } catch (err) {
-      if (err instanceof mongoose.mongo.MongoWriteConcernError && /operation was interrupted/.test(err.message)) {
+      const retryable = err instanceof mongoose.mongo.MongoWriteConcernError && /operation was interrupted/.test(err.message);
+      if (retryable && retries > 0) {
         console.log('DropDB operation interrupted, retrying'); // log that a error was thrown to know that it is going to re-try
         continue;
       }


yogesh968:
can you please review it @vkarpov15 ..

yogesh968:
please merge it @vkarpov15

yogesh968:
Thank you very much!

PR #15792, opened 2025-11-25
Fix docs parser crash when prop.tags is undefined (#15764)

Jeet-Srivastava:
Fixed docs parser crash when prop.tags is undefined (#15764)

### Problem
The documentation parser in `docs/source/api.js` crashes with a `TypeError: Cannot read properties of undefined (reading 'Symbol(Symbol.iterator)')` when processing JSDoc comments that don't have tags. This causes the entire documentation build process to fail.

### Root Cause
At line 273 in `docs/source/api.js`, the code attempts to iterate over `prop.tags` without checking if it exists:
When the `dox` parser encounters JSDoc comments without tags, `prop.tags` is `undefined`, causing the iterator to fail.

### Solution
Added a defensive check before the iteration to skip properties without tags:

This allows the parser to gracefully skip entries without tags while continuing to process properly documented code.

### DIFF
--- a/docs/source/api.js
+++ b/docs/source/api.js
@@ -270,6 +270,12 @@ function processFile(props) {
       ctx.constructorWasUndefined = true;
     }
 
+    // Skip properties without tags to prevent iteration errors
+    // See: https://github.com/Automattic/mongoose/issues/15764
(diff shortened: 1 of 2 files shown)


vkarpov15:
This was already fixed in #15765

PR #15793, opened 2025-11-25
Fix docs parser crash when prop.tags is undefined (#15764)

Jeet-Srivastava:
**Summary**

**Examples**


vkarpov15:
This is a duplicate of #15793 please stop spamming

Issue #15780, opened 2025-11-21
All methods required by TS when setting statics via Schema.static with object parameter


I-Enderlord-I:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.20.1

### Node.js version

20.x

### MongoDB server version

5.x

### Typescript version (if applicable)

5.9.3

### Description

Using the `Schema.static` method to set static model methods requires all methods to be passed. The current type definition is
(https://github.com/Automattic/mongoose/blob/830bf730ec3c5521e7a1f50401a63edc6b254525/types/index.d.ts#L626
As seen there, the keys of TStaticMethods are all required.

This also causes problems with writing a generic Schema type.

### Steps to Reproduce

Test:
```typescript
interface IDoc {
}

interface DocStatics {
	m1(): void;
	m2(): void;
}

const schema = new Schema<IDoc, Model<IDoc>, {}, {}, {}, DocStatics>({});
schema.static({ m1() {} }); //TS Error. Property m2 missing
```

For the second observation consider
```typescript
schema satisfies Schema<any, Model<any>, {}, {}, {}, {}>;
```
This might fail for other reasons as well, but part of the problem is that `{}` is missing the properties in `DocStatics`, making the `static` function unassignable to it's "generic" counterpart. Even when using an index signature e.g. `{ [name: string]: () => never }` the check will still fail because index signature properties are considered optional. If the 

### Expected Behavior

No error.

The should allow defining the statics one by one as with `Schema.method` where the parameter is `obj: Partial<TInstanceMethods>`.
In this case the type thus should be
```typescript
static(obj: { [F in keyof TStaticMethods]?: TStaticMethods[F] } & { [name: string]: (this: TModelType, ...args: any[]) => any }): this;
```
notice the added `?` to make the keys optional.

vkarpov15:
@I-Enderlord-I WDYT about #15794?

I-Enderlord-I:
I think having to do an incorrect type cast in order to use this is absolutely not the right solution.
#15794 changed nothing about this problem as the properties defined in TStatics are still all required.
The test just avoids it by casting the partial object to the full one, which is of course not safe and should **never** be done without very good reason.

I-Enderlord-I:
[...]

PR #15794, opened 2025-11-25
types(schema): allow calling schema.static() with `as TStatics`

vkarpov15:
Re: #15780

**Summary**

My suggested workaround for #15780 would be to use `as DocStatics`, but that currently fails with the following error:

```
Type 'DocStatics' is not assignable to type '{ [name: string]: (this: Model<IDoc, {}, {}, {}, Document<unknown, {}, IDoc, {}, {}> & IDoc & Required<{ _id: unknown; }> & { __v: number; }, any>, ...args: any[]) => any; }'. Index signature for type 'string' is missing in type 'DocStatics'.
```

This PR adds an override to `Schema.prototype.static()` to support calling `static()` with exactly the `TStatics` type

**Examples**

### DIFF
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -801,6 +801,7 @@ function gh15671() {
 
 async function gh15786() {
   interface IDoc {
+    nmae: string;
   }
 
   interface DocStatics {


test/types/queries.test.ts
```diff
@@ -826,3 +798,16 @@ function gh15671() {
     };
   };
 }
+
+async function gh15786() {
+  interface IDoc {
+  }
```
hasezoey, line 805: If i remember correctly, empty interfaces can act weird sometimes, maybe consider adding a dummy property?

test/types/queries.test.ts
```diff
@@ -826,3 +798,16 @@ function gh15671() {
...
+
+  interface DocStatics {
+          m1(): void;
+          m2(): void;
+  }
```
hasezoey, line 810:
```suggestion
  interface DocStatics {
    m1(): void;
    m2(): void;
  }
```
I dont know why lint didnt catch this.

PR #15796, opened 2025-11-26
feat(plugins): auto-convert $set undefined to $unset

wcr-karan:
## Tasks

- Add built-in `convertSetUndefinedToUnset` plugin that turns `$set` or top-level `undefined` values into `$unset` in update queries.
- Document this behavior in `docs/tutorials/findoneandupdate.md` with short examples.
- Add regression tests in `test/plugins.convertSetUndefinedToUnset.test.js` covering:
  - `$set` with `undefined`
  - Top-level `undefined` assignments
  - Merging with existing `$unset`
- Run tests with  
  `npm test -- test/plugins.convertSetUndefinedToUnset.test.js`  
  (won’t run here due to MongoMemoryServer port limits — rerun elsewhere)

### DIFF
--- a/docs/tutorials/findoneandupdate.md
+++ b/docs/tutorials/findoneandupdate.md
@@ -7,6 +7,7 @@ However, there are some cases where you need to use [`findOneAndUpdate()`](https
 * [Atomic Updates](#atomic-updates)
 * [Upsert](#upsert)
 * [The `includeResultMetadata` Option](#includeresultmetadata)
+* [Removing Fields with `undefined`](#removing-fields-with-undefined)
 * [Updating Discriminator Keys](#updating-discriminator-keys)
 
 ## Getting Started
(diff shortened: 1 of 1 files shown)


vkarpov15 (changes requested):
how would someone use this plugin?

wcr-karan:
[vkarpov15](https://github.com/vkarpov15)

Thanks for the feedback! I've added documentation in `docs/tutorials/findoneandupdate.md` explaining how users interact with this feature.

**How users use it:**

The plugin is built-in and automatically applied to all schemas—no setup required. Users can remove fields by setting them to `undefined` in update operations. Mongoose automatically converts these `undefined` values to `$unset` operations behind the scenes, so the fields are properly removed from the document. This works with all update methods (`updateOne`, `updateMany`, `findOneAndUpdate`, etc.), and supports both `$set` with `undefined` and direct field assignment. It also works with nested fields using dot notation.

The documentation now includes a comprehensive section with examples. Let me know if you'd like any changes!

vkarpov15:
Unfortunately this is too much of a backwards breaking PR for us to merge. Thanks for the PR though.

PR #15799, opened 2025-11-26
Fix: Robust Null Safety for Internal Utility Functions

deepesh224-ux:
# Fix: Preventing crashes on empty inputs in utils

### Summary!

I noticed a few spots in the internal utility functions ([lib/utils.js](cci:7://file:///Users/deepeshdey/Desktop/mongoose/lib/utils.js:0:0-0:0)) that were a bit fragile. If you accidentally passed `null` or `undefined` to them, they would immediately crash with a `TypeError` (like trying to read `.length` of null).

**What I fixed:**
I added some simple guard checks to these three functions so they handle empty inputs gracefully instead of blowing up:
*   **`utils.last`**: Now returns `undefined` if the array is missing.
*   **`utils.merge`**: Now just returns the original target object if there's nothing to merge from.
*   **`utils.object.vals`**: Now returns an empty array `[]` if the input is empty.

**Testing:**
I added a few regression tests in [test/utils.test.js](cci:7://file:///Users/deepeshdey/Desktop/mongoose/test/utils.test.js:0:0-0:0) to confirm these edge cases are covered and won't regress. All tests are passing locally!

## Thank You!

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -190,6 +190,9 @@ exports.deepEqual = function deepEqual(a, b) {
  */
 
 exports.last = function(arr) {
+  if (arr == null) {
+    return void 0;
+  }
   if (arr.length > 0) {
     return arr[arr.length - 1];
   }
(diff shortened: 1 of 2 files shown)


deepesh224-ux:
> please undo all the unrelated formatting changes otherwise this PR is hard to review

okay done @vkarpov15

vkarpov15 (approved):
Thanks :+1:

Issue #12798, opened 2022-12-14
Merge select/projection of subdocument tree


Paso:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

When combining select from schema and query, remove parts that are overlapping and causes an invalid query.

### Motivation

When using `select: false` in a subdocuments schema the entire subdocument can not be unselected in a query as it results in a "Path collision" error from mongodb. It would be very useful if the select could resolve the selects on a subdocument so that the extra projection is removed.

### Example

```
//Bar model with subdocument subd where raw is rarely needed
const BarSchema = new Schema({
  name: String,
  subd: {
    raw: {
      type: String,
      select: false,
    },
    clean: String,
  },
})
const Bar = mongoose. Model('Bar', BarSchema)

// This will throw a path collision error
Bar.find({}).select('-subd');
```

As I understand things the example above results in a query with `select: { subd: 0, 'subd.raw': 0}` which is not allowed even though there is no logical ambiguity.

Himanshu197200:
@vkarpov15 I made a PR for this feature Plz review it

PR #15807, opened 2025-11-28
fix: remove conflicting child projections when parent path excluded

Himanshu197200:
---

# Fix: Prevent Parent–Child Projection Collisions in Mongoose (#12798 )

## Overview

This pull request fixes an issue where Mongoose generates **invalid MongoDB projections** when both a parent path and its nested child path are excluded. This occurs when combining:

* Schema-level `select: false` fields
* User-level `.select('-parent')` projections

MongoDB does **not** allow excluding both a parent and its child paths at the same time, resulting in a **Path collision error**.

---

##  The Problem

Given the following schema:

```js
  name: String,
  subd: {
    raw: { type: String, select: false },
    clean: String
  }
});
```

Running this query:

```js
Bar.find().select('-subd');
```

Currently results in this projection:

```js
{ 
  subd: 0,
  'subd.raw': 0   
}
```

MongoDB error:

```
Path collision: Cannot project both 'subd' and 'subd.raw'
```

---

## The Fix

This PR ensures that **nested child exclusions are removed automatically** when the parent path is already excluded.

###  Correct behavior after fix:

```js
{
  subd: 0
}
```

A new helper — `removeConflictingProjections()` — cleans up the projection before it is returned and forwarded to MongoDB.

---

## Implementation Details

* Updated `parseProjection.js`
* Added `removeConflictingProjections()`:

  * Detects excluded parent paths
  * Removes child-level exclusions that would cause collisions
* Ensures compatibility with:

  * Schema-level `select: false`
  * Query-level `.select()` usage
  * Existing projection logic

---

##  Tests Added

A new test validates that:

* Excluding a parent path removes nested exclusions
* Schema-level defaults do not conflict with user projections
* No MongoDB path collision occurs
* All existing projection behavior remains unchanged

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -1782,7 +1782,7 @@ Connection.prototype.syncIndexes = async function syncIndexes(options = {}) {
 };
 
 /**
- * Switches to a different database using the same [connection pool](https://mongoosejs.com/docs/api/connectionshtml#connection_pools).
+ * Switches to a different database using the same [connection pool](https://mongoosejs.com/docs/connections.html#connection_pools).
  *
  * Returns a new connection object, with the new db.
  *
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -190,6 +190,9 @@ exports.deepEqual = function deepEqual(a, b) {
  */
 
 exports.last = function(arr) {
+  if (arr == null) {
+    return void 0;
+  }
   if (arr.length > 0) {
     return arr[arr.length - 1];
   }
@@ -271,6 +274,10 @@ exports.clonePOJOsAndArrays = function clonePOJOsAndArrays(val) {
 
 exports.merge = function merge(to, from, options, path) {
   options = options || {};
+
+  if (from == null) {
+    return to;
+  }
 
   const keys = Object.keys(from);
   let i = 0;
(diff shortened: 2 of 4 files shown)


Himanshu197200:
@vkarpov15  Plz review this PR

lib/helpers/projection/parseProjection.js
```diff
@@ -29,5 +29,20 @@ module.exports = function parseProjection(v, retainMinusPaths) {
     ret[field] = include;
   }
 
+  removeConflictingProjections(ret);
+
   return ret;
 };
+
+function removeConflictingProjections(projection) {
```
vkarpov15: Need to test that this only removes schema selected paths - if a user where to say `select({ subd: 1, 'subd.raw': 0 })` we wouldn't want to strip out the user specified `subd.raw`.

Himanshu197200:
@vkarpov15 I made the changes according to the error , please review it.

Himanshu197200:
Hi @vkarpov15 
I’ve made all the requested changes and updated the PR accordingly.
When you get a moment, could you please take another look and let me know if anything else is needed from my side?
Thanks for your time and guidance!

vkarpov15:
I'm just going to close this PR. For one, there's a bunch of unrelated style changes that make it impossible to review in its entirety. For another, it adds an O(n^2) loop on every query, which can cause significant performance degradation if the user has a large projection.

PR #15810, opened 2025-11-28
Fix/document init crash

Vegapunk-debug:
# Fix: Prevent crash in `Document.prototype.init()` when called with null/undefined

## Summary
Calling `Document.prototype.init()` with `null` or `undefined` currently throws a low-level
TypeError (`Cannot read properties of null (reading '_id')`). This PR adds an explicit
null/undefined check and throws a proper `ObjectParameterError` instead.

## Motivation and Context
`init()` is part of the public API. While normally used internally, custom scripts or plugins
may pass invalid input. The new error message improves developer experience and prevents an
unexpected crash.

## Reproduction
```js
const doc = new mongoose.Document({}, new mongoose.Schema({ name: String }));

// Before:
// TypeError: Cannot read properties of null (reading '_id')
// doc.init(null);

// After:
// ObjectParameterError: Parameter "doc" to init() must be an object.
doc.init(null);
```
## Testing
- I have verified locally using the reproduction script above.
- Also, ran `npm test` to ensure no regressions. 

## Thank you!

### DIFF
--- a/scripts/tsc-diagnostics-check.js
+++ b/scripts/tsc-diagnostics-check.js
@@ -3,7 +3,7 @@
 const fs = require('fs');
 
 const stdin = fs.readFileSync(0).toString('utf8');
-const maxInstantiations = isNaN(process.argv[2]) ? 350000 : parseInt(process.argv[2], 10);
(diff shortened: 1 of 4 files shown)


Vegapunk-debug:
Closing in favor of clean PR #15812

PR #15811, opened 2025-11-28
refactor: simplify Query.prototype.near argument parsing

AyushCoder9:
Summary

Refactors Query.prototype.near in lib/query.js to simplify argument parsing logic and remove a TODO refactor comment.

The previous implementation used a complex chain of if/else statements to handle various argument signatures, leading to code duplication. The new implementation normalizes the arguments first (extracting path if present) and then handles the remaining arguments (val or x, y), making the logic cleaner, more readable, and easier to maintain.

Examples

This refactor maintains support for all existing signatures of near. Verified with a reproduction test case covering:

// Supported signatures:
query.near([x, y]);
query.near(x, y);
query.near(path, [x, y]);
query.near(path, x, y);
query.near({ center: [x, y], maxDistance: d });
query.near(path, { center: [x, y], maxDistance: d });

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -5339,42 +5339,31 @@ Query.prototype.near = function() {
   const params = [];
   const sphere = this._mongooseOptions.nearSphere;
 
-  // TODO refactor
+  const args = Array.from(arguments);
 
-  if (arguments.length === 1) {
-    if (Array.isArray(arguments[0])) {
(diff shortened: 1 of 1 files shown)


vkarpov15:
There's a bunch of unrelated formatting changes and the changes to near() argument parsing are potentially backwards breaking. I'm closing.

PR #15816, opened 2025-11-29
types(schema): fixed #15780 by admitting calls to Schema.static with partial object

### LINKED ISSUE #15780: All methods required by TS when setting statics via Schema.static with object parameter

I-Enderlord-I:
Fixed #15780

**Summary**

Actual fix for #15780 this time, rather than forcing an unsafe type cast as suggested in #15794.

**Examples**

See #15780 and for bad unsafe type cast think about:

```ts
interface IDoc {
  name: string;
}

interface DocStatics {
  m1(): void;
  m2(): void;
}

const schema = new Schema<IDoc, Model<IDoc>, {}, {}, {}, DocStatics>({});
schema.static({ m1(a: string) { console.log(a.slice(1)); } } } as DocStatics);
```

### DIFF
--- a/test/types/queries.test.ts
+++ b/test/types/queries.test.ts
@@ -863,5 +863,5 @@ async function gh15786() {
   }
 
   const schema = new Schema<IDoc, Model<IDoc>, {}, {}, {}, DocStatics>({});
-  schema.static({ m1() {} } as DocStatics);
+  schema.static({ m1() {} });
 }


I-Enderlord-I:
@vkarpov15 
To your comment in #15786 which also relates to this here:

Making types more strict is only useful when the runtime is as strict as well. In this case, passing a partial object is deliberately allowed by the runtime code. If you do not like the types to allow that, you should change the JS code, not hinder corrections of the types.

As for using `as` to work around this: Doing type casts is **never** a soluion or even a workaround. It completely overrules the type checking and should only be done to silence errors which are known to be undesired behaviour. Even then, it's not safe to do so and needs to be thought about from the perspective of maintainability. Technically, every even remotely related type cast needs to be checked every time the code changes, because the argument why it was correct might no longer be valid.

Valid code that needs type casts to avoid errors is always an issue and excluding some long running TS design problems like the "correspondence problem" should be seen as a bug to be fixed by the developer responsible for the type.

I-Enderlord-I:
Also, don't forget that the same partial definition is already a thing for `Schema.method`. No one is complaining there and by symmetry, no one should here. There is no reason to treat statics different to methods.

I-Enderlord-I:
Oops, problem with github here. Accidentally created two branches. Fixed it.

vkarpov15:
Superseded by acd7f24cacb8bf8d2096b083e24484a62922a330, which is going out in 8.20.3. 

Regarding "It completely overrules the type checking and should only be done to silence errors which are known to be undesired behaviour." I agree - the thinking was that calling `schema.static(obj)` without defining all the statics at once was undesired behavior. That leaves the unfortunate possibility that a static defined in your types isn't defined at runtime, which is still possible using `schema.static(name, fn)`, but is an issue that we could protect against in the `schema.static(obj)` case. But I think it is reasonable to be flexible here.

PR #15819, opened 2025-11-30
fix(model): fix overwriteImmutable not working with timestamps: true, add `overwriteImmutable` types re #15781

AbdelrahmanHafez:
- Adds types for `overwriteImmutable` re #15781 and https://github.com/Automattic/mongoose/pull/15782
- Fix bug where `overwriteImmutable` would only work on `createdAt` with  `timestamps: false`.

### DIFF
--- a/test/model.updateOne.test.js
+++ b/test/model.updateOne.test.js
@@ -2680,7 +2680,7 @@ describe('model: updateOne: ', function() {
     assert.equal(doc.age, 20);
   });
 
-  describe.only('bulkWrite overwriteImmutable option (gh-15781)', function() {
+  describe('bulkWrite overwriteImmutable option (gh-15781)', function() {
     it('updateOne can update immutable field with overwriteImmutable: true', async function() {
       // Arrange
       const { User } = createTestContext();


package.json
```diff
@@ -99,7 +99,7 @@
     "test-tsd": "node ./test/types/check-types-filename && tsd --full",
     "setup-test-encryption": "node scripts/setup-encryption-tests.js",
     "test-encryption": "mocha --exit ./test/encryption/*.test.js",
-    "tdd": "mocha ./test/*.test.js --inspect --watch --recursive --watch-files ./**/*.{js,ts}",
+    "tdd": "npx mocha ./test/*.test.js --inspect --watch --recursive --watch-files 'test/**/*.js' --watch-files 'lib/**/*.js'",
```
hasezoey: As copilot already pointed out, why is this change necessary?

AbdelrahmanHafez: I created the `tdd` script long time ago, but it's broken now. I'm guessing this is because of a change in the type-testing approach, IIRC we didn't start with `tsd` as our type testing tool.

Anyway, the new script fixes the errors from the old script. It's not really used as part of any automation so it's safe to change.

PR #15820, opened 2025-11-30
fix: allow timestamps and overwriteImmutable in TS and JS for 7.x

AbdelrahmanHafez:
## PR Summary

**Add TypeScript types and per-operation timestamps support for bulkWrite (gh-15782)**

- Added mongoose `AnyBulkWriteOperation` type with `timestamps`, `overwriteImmutable`, `setDefaultsOnInsert`, and `skipValidation` options
- Fix current solution requiring `timestamps: false` to be able to update timestamps with `overwriteImmutable`.
- Added JS and TS tests

### DIFF
--- a/test/model.updateOne.test.js
+++ b/test/model.updateOne.test.js
@@ -2788,46 +2788,47 @@ describe('model: updateOne: ', function() {
         assert.equal(updatesDocs[0].createdAt.valueOf(), createdAt.valueOf());
         assert.equal(updatesDocs[1].createdAt.valueOf(), createdAt.valueOf());
       });
-    }
 
-    it('can not update immutable fields without overwriteImmutable: true', async function() {
+      it(`can not update immutable fields without overwriteImmutable: true and timestamps: ${timestamps}`, async function() {
       // Arrange
-      const { User } = createTestContext();
-      const users = await User.create([
-        { name: 'Bob', ssn: '222-22-2222' },
-        { name: 'Eve', ssn: '333-33-3333' }
-      ]);
-      const newCreatedAt = new Date('2020-01-01');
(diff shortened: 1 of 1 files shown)


lib/helpers/model/castBulkWrite.js
```diff
@@ -190,7 +190,7 @@ module.exports = function castBulkWrite(originalModel, op, options) {
 
       // set `skipId`, otherwise we get "_id field cannot be changed"
       const doc = new model(op['replaceOne']['replacement'], strict, true);
-      if (model.schema.options.timestamps) {
+      if (model.schema.options.timestamps && getTimestampsOpt(op['replaceOne'], options)) {
```
vkarpov15: I'd really prefer to inline this so we don't have an extra function here, could just do `op['replaceOne'].timestamps ?? options.timestamps ?? true`

types/models.d.ts
```diff
@@ -142,6 +142,105 @@ declare module 'mongoose' {
 
   interface RemoveOptions extends SessionOption, Omit<mongodb.DeleteOptions, 'session'> {}
 
+  export type AnyBulkWriteOperation<TSchema = AnyObject> = {
+    insertOne: InsertOneModel<TSchema>;
```
vkarpov15: Can we reuse the MongoDB driver's `InsertOneModel` etc. types here? I'd rather not duplicate them because the types can then fall out of sync.

AbdelrahmanHafez:
@vkarpov15 The original solution required `timestamps: false` to work with updating timestamps. This PR also fixes that to make `overwriteImmutable` take effect on `createdAt` regardless of the `timestamps` option value. Also fixed in the PR that targets master #15819

PR #15828, opened 2025-12-01
fix: expose validation errors in insertMany with ordered: false

### LINKED ISSUE #15771: Unordered insertMany drops per-document validation errors when rawResult is false
### Prerequisites

### Mongoose version

6.13.1

### Node.js version

14.21.3

### MongoDB server version

 5.0.26

4.8.4

### Description

Calling insertMany() with ordered=false and rawResult=false silently drops the per-document validation errors. You only get the successfully inserted docs (or a driver error), so there’s no way to tell which documents failed validation without re-running the entire batch.

### Steps to Reproduce

const schema = new Schema({ name: { type: String, required: true }, year: Number });
const Movie = mongoose.model('Movie', schema);

await Movie.insertMany([
  { foo: 'invalid 1' },
  { name: 'valid movie' },
  { bar: 'invalid 2' }
], { ordered: false });
// Promise resolves with `[ { name: 'valid movie' } ]`
// but there is no information about the two invalid docs

### Expected Behavior

Even when rawResult=false, the resolved docs (or thrown error) should expose mongoose.validationErrors/results so callers can identify which indexes failed validation, similar to the rawResult behavior.

maynkxx:
Fixes #15771

**Summary**

Previously, when calling `Model.insertMany()` with `{ ordered: false, rawResult: false }`, any per-document validation errors were silently dropped. The method would return only the successfully inserted documents, making it impossible for the caller to know which documents failed validation or why.

This PR modifies `insertMany` to ensure that when `ordered: false` is used, validation errors are captured and attached to the result array as a property named `validationErrors`. This behavior is consistent with other error handling paths in Mongoose and allows developers to handle partial failures gracefully.

**Examples**

```javascript
const schema = new Schema({ name: { type: String, required: true } });

// Insert one valid document and one invalid document
const docs = [{ name: 'Valid' }, { foo: 'Invalid' }];

// Before this fix: validation errors were lost, and we only got the valid doc back.
// After this fix: result.validationErrors contains the error info.
const result = await Model.insertMany(docs, { 
  ordered: false, 
  rawResult: false 
});

console.log(result.length); 
// Output: 1 (The valid document)

console.log(result.validationErrors); 
// Output: [ValidationError: Path `name` is required.]

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3152,15 +3152,27 @@ Model.insertMany = async function insertMany(arr, options) {
   }
 
   if (options.populate != null) {
-    return this.populate(docAttributes, options.populate).catch(err => {
+    res = await this.populate(docAttributes, options.populate).catch(err => {
       if (err != null) {
         err.insertedDocs = docAttributes;
       }
       throw err;
     });
+
+    if (ordered === false && validationErrors.length > 0) {
+      res.validationErrors = validationErrors;
+    }
+
+    return res;
   }
 
-  return await this._middleware.execPost('insertMany', this, [docAttributes]).then(res => res[0]);
+  res = await this._middleware.execPost('insertMany', this, [docAttributes]).then(res => res[0]);
+
+  if (ordered === false && validationErrors.length > 0) {
+    res.validationErrors = validationErrors;
+  }
+
+  return res;
 };
 
 /*!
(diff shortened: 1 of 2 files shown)


maynkxx:
hi [vkarpov15](https://github.com/vkarpov15) this PR is ready for review! Let me know if any changes are needed.
Thanks!

maynkxx:
@hasezoey ,Thanks! I just cleaned up the style changes and pushed the fix.

vkarpov15:
Backwards breaking

PR #15829, opened 2025-12-01
refactor: remove custom toString from ValidatorError

maynkxx:
This PR addresses a TODO in 
lib/error/validator.js
 by removing the custom toString() method from the 
ValidatorError
 class.

Changes:

Removed the custom toString() method in 
lib/error/validator.js
 which was overriding the default Error.prototype.toString() behavior.
ValidatorError
 instances now inherit the standard toString() behavior, resulting in output like "ValidatorError: message" instead of just "message".
Updated unit tests in 
test/schema.validation.test.js
 to reflect this change in string representation.
Verification:

Ran npm run lint to ensure code style compliance.
Ran npx mocha test/schema.validation.test.js and npx mocha test/docs/validation.test.js to verify that all tests pass with the updated behavior.

### DIFF
--- a/lib/error/validator.js
+++ b/lib/error/validator.js
@@ -34,15 +34,6 @@ class ValidatorError extends MongooseError {
   }
 
   /**
-   * toString helper
-   * TODO remove? This defaults to `${this.name}: ${this.message}`
-   * @api private
-   */
-  toString() {
-    return this.message;
-  }
-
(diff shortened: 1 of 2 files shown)


vkarpov15:
There have been multiple PRs like this - please stop spamming AI slop or I will report you to GitHub.

PR #15830, opened 2025-12-01
refactor: remove redundant ValidatorError.prototype.toString()

Srishtyshree:
Summary

Removes the redundant `toString() `method from the `ValidatorError class` in [lib/error/validator.js](https://github.com/Automattic/mongoose/blob/master/lib/error/validator.js).
A TODO comment in the code noted that this method was likely unnecessary as it duplicates the default behavior of `Error.prototype.toString() (returning ${name}: ${message})`, with the minor difference that the custom method returned only the message.

### DIFF
--- a/lib/error/validator.js
+++ b/lib/error/validator.js
@@ -33,14 +33,6 @@ class ValidatorError extends MongooseError {
     this.reason = properties.reason;
   }
 
-  /**
-   * toString helper
-   * TODO remove? This defaults to `${this.name}: ${this.message}`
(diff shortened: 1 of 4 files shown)


Srishtyshree:
@AbdelrahmanHafez  @hasezoey Can you please review this PR and let me know if any changes are needed.
Thank you

Srishtyshree:
@vkarpov15 @AbdelrahmanHafez Please review the changes and suggest improvements if any!

PR #15846, opened 2025-12-02
fix(schema): Add enumValues support to SchemaBuffer (consistency with String/Number)

### LINKED ISSUE #15845: SchemaBuffer does not expose enumValues like String and Number enums
### Prerequisites

### Mongoose version

v9.0.0

### Node.js version

v20.19.4

### MongoDB server version

6.0

_No response_

### Description

`SchemaBuffer` does not expose an `enumValues` property, even though it accepts enum definitions and performs enum validation. This makes buffer enums inconsistent with `SchemaString` and `SchemaNumber`, both of which expose `enumValues` and support additive `.enum()` calls.

The result is that applications cannot introspect allowed enum values for buffer fields in the same way they can for string or number enums.

### Steps to Reproduce

```js
const { Schema } = mongoose;

const schema = new Schema({
  buf: { type: Buffer, enum: [Buffer.from('a'), Buffer.from('b')] }
});

console.log(schema.path('buf').enumValues);
// Expected:
// [<Buffer 61>, <Buffer 62>]

// Actual:
// undefined
```
SchemaString and SchemaNumber both correctly set enumValues when enum options are provided. SchemaBuffer does not, even though validation works.

### Expected Behavior

```md
`SchemaBuffer` should provide an `enumValues` array containing the allowed values (cast to `Buffer`), just like `SchemaString` and `SchemaNumber`.

It should also support:
- `path.enumValues` introspection
- additive `.enum(v1, v2)`
- object syntax (`{ values, message }`)
- consistent enum validation behavior

This keeps all built-in schema types aligned.

Anuj-Chhajed:
Fixes #15845

This PR adds `enumValues` support to SchemaBuffer so that buffer enum behavior becomes consistent with SchemaString and SchemaNumber.

### Background

SchemaBuffer previously allowed enum validation but did not expose an `enumValues` array or fully support the `.enum()` API pattern used by other schema types. This created an inconsistency where buffer enums could validate values, but the allowed values could not be introspected or extended using additional `.enum()` calls.

### What this PR changes

- Initializes `this.enumValues = []` in the SchemaBuffer constructor.
- Implements `SchemaBuffer.prototype.enum()` following the same structure as `SchemaNumber.prototype.enum()`:
  - Removes the previous enum validator if one exists.
  - Supports object syntax: `{ values, message }`.
  - Casts enum inputs using `this.cast(...)` before storing.
  - Pushes casted values into `this.enumValues`.
  - Creates a new enum validator that compares buffers using `Buffer.compare`.
  - Stores the validator and associated metadata in `this.validators`.
  - Returns `this` for method chaining.
- Ensures message fallback uses a Buffer-specific default string format.
- Does not alter existing casting or validation logic outside the enum pathway.

### Tests

A new test has been added to `test/schema.validation.test.js` that mirrors the structure of the existing string and number enum tests and verifies:

- `enumValues` is correctly populated on schema paths.
- Additive `.enum()` behavior works (`enum('c')` appends).
- Object syntax (`{ values, message }`) is supported.
- Valid buffer values pass validation; invalid ones fail.
- `null` and `undefined` behave consistently with other SchemaTypes.

### Why this change is needed

This is an API consistency improvement. String and Number schema types already support:

- `enumValues` exposure
- additive `.enum()`
- object syntax
- getter behavior when calling `.enum()` with no args

This PR brings SchemaBuffer up to the same standard without introducing breaking changes.

### Notes

- All changes are self-contained within `schema/buffer.js`.
- No behavior outside enum handling is modified.
- No breaking changes.

### DIFF
--- a/lib/schema/buffer.js
+++ b/lib/schema/buffer.js
@@ -26,6 +26,7 @@ const CastError = SchemaType.CastError;
  */
 
 function SchemaBuffer(key, options, _schemaOptions, parentSchema) {
+  this.enumValues = [];
   SchemaType.call(this, key, options, 'Buffer', parentSchema);
 }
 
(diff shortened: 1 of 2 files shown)


Anuj-Chhajed:
@vkarpov15 @AbdelrahmanHafez pls review my pr

vkarpov15:
I'm going to close this PR since I'm not sure that having `enum` support for buffers is useful - we haven't received any requests for this functionality and I'm having a hard time thinking of a case where this would be useful.

PR #15847, opened 2025-12-02
Fix typo in text search error path and variable bug in double casting

lomesh2312:
- Fix typo in lib/schema/operators/text.js: '' -> '' This ensures error messages correctly reference the  field

- Fix variable reference bug in lib/cast/double.js: use 'tempVal' instead of 'val' When casting objects with valueOf()/toString() methods, the extracted string value should be used instead of the original object

**Summary**

**Examples**


lomesh2312:
Closing to create a fresh PR with updated information.

PR #15850, opened 2025-12-02
fix: pathsToSave should allow saving parent paths of whitelisted paths

Bitshifter-9:
This PR fixes an issue where pathsToSave would skip updates when a modified path was a parent of one of the whitelisted paths. As a result, valid updates were being ignored even though they should have been saved.

What Was Changed

In lib/model.js, the save method now checks whether the modified key is either:

a parent of a whitelisted path, or

a child of it.

If either condition matches, the path is correctly included in the delta.

Verification

Added a dedicated test file test/gh-paths-to-save.test.js to reproduce the issue and validate the fix.

Ran all existing tests (test/model.test.js) to confirm no regressions.

Verified that the new test passes:

gh-paths-to-save
  ✔ should allow saving parent paths of whitelisted paths (gh-issue)

This ensures that parent paths are now properly saved when related whitelisted paths are updated.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -409,7 +409,8 @@ Model.prototype.$__save = async function $__save(options) {
         for (const key in delta[1]['$set']) {
           if (options.pathsToSave.includes(key)) {
             continue;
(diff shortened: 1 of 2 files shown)


vkarpov15:
Closing in favor of #15896, this PR is good as a bug report but isn't close to mergeable unfortunately.

PR #15856, opened 2025-12-03
fix: allow custom properties on QueryFilter objects (gh-15854)

### LINKED ISSUE #15854: Can't access custom props of `QueryFilter` objects
### Prerequisites

### Mongoose version

9.0.0

### Node.js version

24.11.1

### MongoDB server version

N/A

5.9.3

### Description

Trying to access properties of a `QueryFilter` in TypeScript fails with "Property '\<name\>" does not exist".

### Steps to Reproduce

```ts
const filter = { foo: 'hi' } as mongoose.QueryFilter<{ foo: string }>;
filter.foo; // error: Property 'foo' does not exist on type '_QueryFilter<{ foo: string; }>'.
```

### Expected Behavior

In v8, the above code worked without producing an error. I would expect it to still do so.

shalini-saloni:
Fixes #15854
## Description
This PR fixes an issue where TypeScript would throw an error when trying to access custom properties on `QueryFilter` objects.
## Changes
- Modified `_QueryFilter` type definition in `types/query.d.ts` to include `& Record<string, any>`, allowing custom properties
- Added test case `gh15854()` in `test/types/queries.test.ts` to verify the fix
## Testing
- All TypeScript type tests pass (`npm run test-tsd`)
- The new test case verifies that custom properties can be accessed without TypeScript errors

### DIFF
--- a/types/query.d.ts
+++ b/types/query.d.ts
@@ -9,18 +9,18 @@ declare module 'mongoose' {
   type QueryTypeCasting<T> = T extends string
     ? StringQueryTypeCasting
     : T extends Types.ObjectId
-      ? ObjectIdQueryTypeCasting
-      : T extends Types.UUID
-        ? UUIDQueryTypeCasting
-        : T extends Buffer
-          ? BufferQueryCasting
-          : T extends NativeDate
-            ? DateQueryTypeCasting
-            : T;
+    ? ObjectIdQueryTypeCasting
+    : T extends Types.UUID
+    ? UUIDQueryTypeCasting
+    : T extends Buffer
+    ? BufferQueryCasting
+    : T extends NativeDate
+    ? DateQueryTypeCasting
+    : T;
 
   export type ApplyBasicQueryCasting<T> = QueryTypeCasting<T> | QueryTypeCasting<T[]> | (T extends (infer U)[] ? QueryTypeCasting<U> : T) | null;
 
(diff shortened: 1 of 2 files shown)


types/query.d.ts
```diff
@@ -9,18 +9,18 @@ declare module 'mongoose' {
...
+    : T extends Buffer
+    ? BufferQueryCasting
+    : T extends NativeDate
+    ? DateQueryTypeCasting
+    : T;
 
   export type ApplyBasicQueryCasting<T> = QueryTypeCasting<T> | QueryTypeCasting<T[]> | (T extends (infer U)[] ? QueryTypeCasting<U> : T) | null;
 
-  type _QueryFilter<T> = ({ [P in keyof T]?: mongodb.Condition<ApplyBasicQueryCasting<T[P]>>; } & mongodb.RootFilterOperators<{ [P in keyof T]?: ApplyBasicQueryCasting<T[P]>; }>);
+  type _QueryFilter<T> = ({ [P in keyof T]?: mongodb.Condition<ApplyBasicQueryCasting<T[P]>>; } & mongodb.RootFilterOperators<{ [P in keyof T]?: ApplyBasicQueryCasting<T[P]>; }> & Record<string, any>);
```
simon-abbott, line 23: This is a bad approach as it allows arbitrary access into what should be a strongly typed object. It feels like trying to solve the issue of a locked door by taking a sledgehammer to the adjacent wall. Sure it works, but there must be a better way.

vkarpov15:
This doesn't have anything to do with the reported issue, I'm closing

PR #15860, opened 2025-12-03
chore(workflows/test): drop nodejs 18

hasezoey:
**Summary**

Mongoose already requires 20.19 since 9.0.0. But the Node 18 tests still pass (though reporting unsupported engine)

https://github.com/Automattic/mongoose/blob/5f872fcc78fde68f491fca5e326d82cb08c36371/package.json#L109-L111

Noticed while doing #15859


vkarpov15 (approved):
Thanks :+1:

PR #15862, opened 2025-12-03
fix: prevent negative slice index in populate subdocPath calculation

### LINKED ISSUE #15861: Bug: Negative slice index in populate subdocPath calculation
## Description
When calculating `subdocPath` in populate operations with nested paths, if `schema.path.length + 1` is greater than or equal to `options.path.length`, the slice operation results in a negative index. This causes `slice()` to count from the end of the string instead of the beginning, leading to incorrect path calculations.

## Affected Code
`lib/helpers/populate/getModelsMapForPopulate.js` - Lines 242, 304, and 325

## Current Behavior
```javascript
const subdocPath = options.path.slice(0, options.path.length - schema.path.length - 1);
```

When `schema.path.length + 1 >= options.path.length`, the second argument becomes negative or zero, causing incorrect subdocPath extraction.

## Expected Behavior
The slice index should never be negative. The subdocPath should be correctly extracted even in edge cases where the schema path is longer than or equal to the options path.

## Proposed Fix
Add `Math.max(0, ...)` to prevent negative indices:
```javascript
const subdocPath = options.path.slice(0, Math.max(0, options.path.length - schema.path.length - 1));
```

## Impact
This bug affects populate operations with nested paths and could cause populate to fail or use incorrect paths in edge cases.

## Testing
All existing populate tests pass with this fix (305 tests in `test/model.populate.test.js`).

maniktyagi04:
## Description
This PR fixes a bug where calculating `subdocPath` in populate operations could result in negative slice indices, causing incorrect path calculations.

Fixes #15861

## Problem
In `lib/helpers/populate/getModelsMapForPopulate.js`, lines 242, 304, and 325 use:
```javascript
const subdocPath = options.path.slice(0, options.path.length - schema.path.length - 1);
```

If `schema.path.length + 1 >= options.path.length`, the second argument becomes negative, causing `slice()` to count from the end instead of the beginning.

## Solution
Added `Math.max(0, ...)` to ensure the slice index never goes negative:
```javascript
const subdocPath = options.path.slice(0, Math.max(0, options.path.length - schema.path.length - 1));
```

## Changes
- Fixed line 242: Added `Math.max(0, ...)` wrapper
- Fixed line 304: Added `Math.max(0, ...)` wrapper
- Fixed line 325: Added `Math.max(0, ...)` wrapper

## Testing
- All existing populate tests pass (305 tests in `test/model.populate.test.js`)
- Lint passes (`npm run lint`)

## Impact
This fix prevents incorrect subdocPath calculations in edge cases where the schema path is longer than or equal to the options path, ensuring populate operations work correctly in all scenarios.

### DIFF
--- a/lib/helpers/populate/getModelsMapForPopulate.js
+++ b/lib/helpers/populate/getModelsMapForPopulate.js
@@ -239,7 +239,7 @@ module.exports = function getModelsMapForPopulate(model, docs, options) {
       modelNames = [modelNameFromQuery]; // query options
     } else if (refPath != null) {
       if (typeof refPath === 'function') {
-        const subdocPath = options.path.slice(0, options.path.length - schema.path.length - 1);
+        const subdocPath = options.path.slice(0, Math.max(0, options.path.length - schema.path.length - 1));
         const vals = mpath.get(subdocPath, doc, lookupLocalFields);
         const subdocsBeingPopulated = Array.isArray(vals) ?
           utils.array.flatten(vals) :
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez (changes requested):
Can you please write a test case that fails without this change, so that we can prevent regressions in the future?
Also helps understand with code what the issue you're facing looks like.

PR #15864, opened 2025-12-04
feat(document): add flattenUUIDs option to `toObject()` and `toJSON()`

### LINKED ISSUE #15021: flattenUUIDs
Re: https://github.com/Automattic/mongoose/discussions/14969#discussioncomment-11005437

AbdelrahmanHafez:
## Summary
  - Add `flattenUUIDs` option to `toObject()` and `toJSON()` that converts UUID instances to 36-character hex strings, similar to `flattenObjectIds`
  - Add TypeScript support with `UUIDToString<T>` type helper and method overloads
  - Rename `UUIDToJSON` to `UUIDToString` for consistency with `ObjectIdToString`, keeping `UUIDToJSON` as a deprecated alias for backwards compatibility

Fixes #15021

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -7163,6 +7163,27 @@ describe('document', function() {
     }
 
 
+    it('converts UUIDs inside Maps when both flattenMaps and flattenUUIDs are true', function() {
+      // Arrange
+      const { User, UUID } = createTestContext();
+      const user = new User({
+        _id: new UUID('00000000-0000-0000-0000-000000000000'),
+        uuidMap: new Map([
+          ['first', { refId: new UUID('11111111-1111-1111-1111-111111111111') }],
+          ['second', { refId: new UUID('22222222-2222-2222-2222-222222222222') }]
+        ])
+      });
+
+      // Act
+      const userObj = user.toObject({ flattenMaps: true, flattenUUIDs: true });
+
+      // Assert
+      assert.deepStrictEqual(userObj.uuidMap, {
+        first: { refId: '11111111-1111-1111-1111-111111111111' },
+        second: { refId: '22222222-2222-2222-2222-222222222222' }
+      });
+    });
+
     function createTestContext() {
       const UUID = mongoose.Types.UUID;
       const userSchema = new Schema({
@@ -7172,7 +7193,8 @@ describe('document', function() {
           uuid: 'UUID'
         },
         subdocument: new Schema({ _id: 'UUID' }),
-        documentArray: [new Schema({ _id: 'UUID' })]
+        documentArray: [new Schema({ _id: 'UUID' })],
+        uuidMap: { type: Map, of: new Schema({ refId: 'UUID' }, { _id: false }) }
       }, { versionKey: false });
 
       const User = db.model('User', userSchema);


hasezoey (approved):

types/document.d.ts
```diff
@@ -287,6 +287,10 @@ declare module 'mongoose' {
     toJSON(options: ToObjectOptions & { flattenMaps: true, virtuals: true }): FlattenMaps<Require_id<DocType & TVirtuals>>;
     toJSON(options: ToObjectOptions & { flattenMaps: true }): FlattenMaps<Require_id<DocType>>;
 
+    // flattenUUIDs overloads
+    toJSON(options: ToObjectOptions & { flattenUUIDs: true, virtuals: true }): UUIDToString<Require_id<DocType & TVirtuals>>;
```
vkarpov15: Unfortunately we would also need to handle all other combinations:

- `flattenUUIDs: true, flattenMaps: true`
- `flattenUUIDs: true, flattenObjectIds: true`
- `flattenUUIDs: true, flattenObjectIds: true, flattenMaps: true`
- `flattenUUIDs: true, flattenObjectIds: true, flattenMaps: true, virtuals: true`
- `flattenUUIDs: true, flattenObjectIds: true, virtuals: true`
- `flattenUUIDs: true, flattenMaps: true, virtuals: true`

Plus versionKey variants for completeness' sake.

Honestly this is why I don't like adding new options to toObject() and toJSON() unless absolutely necessary: they cause combinatorial explosion in TypeScript. There isn't a good way to handle all the different cases without explicitly listing out every single override.

types/index.d.ts
```diff
@@ -927,23 +929,33 @@ declare module 'mongoose' {
           } : T;
 
     /**
-    * Converts any Buffer properties into "{ type: 'buffer', data: [1, 2, 3] }" format for JSON serialization
+    * Converts any UUID properties into strings for JSON serialization
     */
-    export type UUIDToJSON<T> = T extends mongodb.UUID
+    export type UUIDToString<T> = T extends Types.UUID
```
vkarpov15: Why is rename necessary?

vkarpov15 (changes requested):
Need to make sure the TypeScript types are complete for handling all possible options overrides, which is a pain but it's the right thing to do. Adding all these overrides makes me wonder whether adding this feature is even worthwhile :worried:

types/index.d.ts
AbdelrahmanHafez: The rename isn't strictly necessary. I did it for consistency with `ObjectIdToString` (since we have `flattenObjectIds` -> `ObjectIdToString`, it felt natural to have `flattenUUIDs` -> `UUIDToString`). The old `UUIDToJSON` name is preserved as a deprecated alias for backwards compatibility.

That said, I'm happy to drop the rename entirely if you'd prefer to keep things simpler.

types/document.d.ts
AbdelrahmanHafez: I reworked this to use compositional conditional types instead of explicit overloads for every combination:

```typescript
type ApplyVirtuals<T, TVirtuals, O> = O extends { virtuals: true } ? T & TVirtuals : T;
type ApplyVersionKey<T, O, TSchemaOptions> = O extends { versionKey: false } ? Omit<T, VersionKeyName> : T;
type ApplyFlattenTransforms<T, O> = // handles flattenMaps, flattenObjectIds, flattenUUIDs in one pass
```

These feed into `ToObjectReturnType<DocType, TVirtuals, O, TSchemaOptions>` which composes them all together. Adding a new option now just means adding one more transform type instead of doubling the overload count, so future options won't blow up the type definitions.

The flatten options (`flattenMaps`, `flattenObjectIds`, `flattenUUIDs`) are handled by a single recursive type that processes all three in one traversal. I initially had separate recursive types for each, but nesting them caused TypeScript's "union type too complex" error. The single-pass approach avoids that and also correctly handles edge cases like ObjectIds nested inside Map values.

test/types/schema.test.ts
```diff
@@ -2015,7 +2015,7 @@ function testInferHydratedDocTypeFromSchema() {
     map?: Map<string, string> | null | undefined
   }, { id: string }, {}, { id: string }>;
 
-  expectType<Expected>({} as HydratedDocType);
+  expectAssignable<Expected>({} as HydratedDocType);
```
AbdelrahmanHafez: Changed to `expectAssignable` due to how TypeScript compares type identity.

This test compares two Document types:
- `HydratedDocType`: inferred from the schema, has `TSchemaOptions = ResolveSchemaOptions<{}>`
- `Expected`: manually constructed, has `TSchemaOptions = DefaultSchemaOptions`

Even though this test never calls `toJSON`, `expectType` checks for type **identity** by comparing the entire structure of both types, including all method signatures.

Before this PR, `TSchemaOptions` wasn't used in any Document method signatures:
```typescript
toJSON(options?: ToObjectOptions): Require_id<DocType>;
```

After this PR, `TSchemaOptions` appears in return types:
```typescript
toJSON(options?: ToObjectOptions): Default__v<Require_id<DocType>, TSchemaOptions>;
```

So now TypeScript sees different method signatures on the two Document types:
```typescript
// HydratedDocType's toJSON:
toJSON(): Default__v<..., ResolveSchemaOptions<{}>>

// Expected's toJSON:
toJSON(): Default__v<..., DefaultSchemaOptions>
```

`ResolveSchemaOptions<{}>` and `DefaultSchemaOptions` are structurally identical (they resolve to the same shape), but they're different type expressions. This makes the entire Document types non-identical, even though they're fully compatible.

`expectAssignable` is correct here because the types ARE compatible for all practical purposes.

types/inferschematype.d.ts
```diff
@@ -112,7 +112,12 @@ declare module 'mongoose' {
       }[alias]
     : unknown;
 
-  type ResolveSchemaOptions<T> = MergeType<DefaultSchemaOptions, T>;
+  type TypesAreEqual<A, B> = [A] extends [B] ? [B] extends [A] ? true : false : false;
+
+  type ResolveSchemaOptions<T> =
```
AbdelrahmanHafez, line 115: Updated `ResolveSchemaOptions` to preserve type identity when `T` is `{}` or `DefaultSchemaOptions`.

When creating a schema without options (`new Schema({...})`), TypeScript infers `TSchemaOptions` as `{}`. Previously, `ResolveSchemaOptions<{}>` returned `MergeType<DefaultSchemaOptions, {}>`, which is structurally identical to `DefaultSchemaOptions` but a different type expression. This caused `expectType` checks to fail when comparing inferred document types against manually constructed ones.

Now `ResolveSchemaOptions` returns `DefaultSchemaOptions` directly in these cases, so the types are truly identical.

Without this change we need to modify existing tests, see https://github.com/Automattic/mongoose/pull/15864#discussion_r2640762977

test/types/document.test.ts
```diff
@@ -652,6 +660,175 @@ async function gh15578() {
...
+function testObjectIdsInsideMaps() {
+  // Test that ObjectIds/UUIDs nested inside Map values are correctly converted
+  interface DocWithMapOfObjectIds {
+    _id: Types.ObjectId;
+    userRefs: Map<string, { oderId: Types.ObjectId }>;
+    uuidRefs: Map<string, { refId: Types.UUID }>;
+  }
+
+  const schema = new Schema<DocWithMapOfObjectIds>({
+    userRefs: { type: Map, of: { oderId: Schema.Types.ObjectId } },
```
vkarpov15: Minor typo - this could be `orderId` or `otherId`?

types/index.d.ts
vkarpov15: I'd prefer to keep the old name, at least unless we're going to drop support for the old name in the next major release. I'd prefer to avoid adding aliased names unless there's a strong reason to - keep the API consistent

types/inferschematype.d.ts
```diff
@@ -112,7 +112,12 @@ declare module 'mongoose' {
       }[alias]
     : unknown;
 
-  type ResolveSchemaOptions<T> = MergeType<DefaultSchemaOptions, T>;
+  type TypesAreEqual<A, B> = [A] extends [B] ? [B] extends [A] ? true : false : false;
```
vkarpov15: This helper should be equivalent to `IfEquals<>`, can we reuse IfEquals to avoid adding a separate helper?

types/index.d.ts
```diff
@@ -1011,14 +1023,85 @@ declare module 'mongoose' {
...
+    // Handle regular objects - recurse into properties
+    : T extends Record<string, any>
+      ? { [K in keyof T]: ApplyFlattenTransforms<T[K], O> }
+    : T;
+
+  /**
+   * Computes the return type of toObject/toJSON based on the provided options.
+   * Uses a single-pass transform for flatten operations to correctly handle all combinations.
+   */
+  export type ToObjectReturnType<DocType, TVirtuals, O extends ToObjectOptions, TSchemaOptions = {}> =
```
vkarpov15, line 1086: image

AbdelrahmanHafez: 🫡

vkarpov15:
Merging into 9.2.

PR #15865, opened 2025-12-04
perf: optimize Object.keys() checks and optional chaining

AbdelrahmanHafez:
## Summary

- Add `utils.hasOwnKeys()` helper that checks for empty objects without allocating an array via `Object.keys()`
- Replace ~50 instances of `Object.keys(obj).length` pattern with `hasOwnKeys()`, reading the first key only instead of creating an array for all keys.
- Expand optional chaining conversions across ~78 files, including:
  - `x && x.y` → `x?.y` patterns
  - `x != null && x.y` → `x?.y` patterns
  - `x && x.y && x.z` → `x?.y?.z` deep access patterns
  - `x != null ? x : default` → `x ?? default`
  - Remove redundant null checks before `Array.isArray()` calls (Array.isArray already returns false for null/undefined, and performs the same)
- Cache repeated property lookups (e.g., `this.$__schema.options`, `model.db.options`)
- Optimize `indexOf()` calls by caching the result instead of calling twice
- Convert `arguments` to rest parameters in `Query.prototype.slice()`
- Optimize `Set` operations in `document.overwrite()` and `getModelsMapForPopulate()`

### DIFF
--- a/lib/helpers/populate/getModelsMapForPopulate.js
+++ b/lib/helpers/populate/getModelsMapForPopulate.js
@@ -625,7 +625,7 @@ function _getLocalFieldValues(doc, localField, model, options, virtual, schema) 
   const localFieldPath = localFieldPathType === 'real' ?
     model.schema.path(localField) :
     localFieldPathType.schema;
-  const localFieldGetters = localFieldPath.getters || [];
+  const localFieldGetters = localFieldPath?.getters || [];
 
   localField = localFieldPath?.instance === 'Embedded' ? localField + '._id' : localField;
 


lib/utils.js
```diff
@@ -436,10 +436,26 @@ exports.isNativeObject = function(arg) {
...
+};
+
+/**
+ * Determines if `obj` has any own keys. Assumes obj is already an object.
+ * Faster than Object.keys(obj).length > 0.
+ * @param {Object} obj
+ */
+
+exports.hasOwnKeys = function hasOwnKeys(obj) {
+  for (const key in obj) {
```
vkarpov15, line 453: Why is this a `for` loop instead of just checking `Object.keys(obj).length`? [quick benchmark](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJBN1lSaG94NUF0b3hBTDQxY1ZSUGEiLCJjb2RlIjoiaWYgKE9iamVjdC5rZXlzKERBVEEpLmxlbmd0aCA-IDApIHJldHVybiB0cnVlOyIsIm5hbWUiOiJGaW5kIDk5IiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6IkZ1b3lKdzZlMjVpTTFvOTBFZzFNdyIsImNvZGUiOiJmb3IgKGNvbnN0IGtleSBpbiBEQVRBKSBpZiAoT2JqZWN0LnByb3RvdHlwZS5oYXNPd25Qcm9wZXJ0eShEQVRBLCBrZXkpKSByZXR1cm4gdHJ1ZTsiLCJuYW1lIjoiRmluZCAxOTkiLCJkZXBlbmRlbmNpZXMiOltdfV0sImNvbmZpZyI6eyJuYW1lIjoiQmFzaWMgZXhhbXBsZSIsInBhcmFsbGVsIjp0cnVlLCJnbG9iYWxUZXN0Q29uZmlnIjp7ImRlcGVuZGVuY2llcyI6W119LCJkYXRhQ29kZSI6InJldHVybiB7IGE6IDEgfTsifX0) shows that Object.keys() is much faster

image

lib/cast.js
```diff
@@ -126,10 +126,7 @@ module.exports = function cast(schema, obj, options, context) {
           const pathFirstHalf = split.slice(0, j).join('.');
           const pathLastHalf = split.slice(j).join('.');
           const _schematype = schema.path(pathFirstHalf);
-          const discriminatorKey = _schematype &&
-            _schematype.schema &&
-            _schematype.schema.options &&
-            _schematype.schema.options.discriminatorKey;
+          const discriminatorKey = _schematype?.schema?.options?.discriminatorKey;
```
vkarpov15, line 129: Optional chaining is much cleaner code-wise, but performance impact is often mixed: https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJBN1lSaG94NUF0b3hBTDQxY1ZSUGEiLCJjb2RlIjoicmV0dXJuIERBVEEgJiYgREFUQS5hICYmIERBVEEuYS5iICYmIERBVEEuYS5iLmM7IiwibmFtZSI6IkZpbmQgOTkiLCJkZXBlbmRlbmNpZXMiOltdfSx7ImlkIjoiRnVveUp3NmUyNWlNMW85MEVnMU13IiwiY29kZSI6InJldHVybiBEQVRBPy5hPy5iPy5jOyIsIm5hbWUiOiJGaW5kIDE5OSIsImRlcGVuZGVuY2llcyI6W119XSwiY29uZmlnIjp7Im5hbWUiOiJCYXNpYyBleGFtcGxlIiwicGFyYWxsZWwiOnRydWUsImdsb2JhbFRlc3RDb25maWciOnsiZGVwZW5kZW5jaWVzIjpbXX0sImRhdGFDb2RlIjoicmV0dXJuIHsgYTogeyBiOiB7IGM6IDEgfSB9IH07In19

image

lib/connection.js
```diff
@@ -1250,8 +1250,11 @@ Connection.prototype._close = async function _close(force, destroy) {
   const conn = this;
   switch (this.readyState) {
     case STATES.disconnected:
-      if (destroy && this.base.connections.indexOf(conn) !== -1) {
-        this.base.connections.splice(this.base.connections.indexOf(conn), 1);
+      if (destroy) {
+        const index = this.base.connections.indexOf(conn);
```
vkarpov15, line 1254: I'm not sure what the purpose of this change is - is this intended to improve readability? I'm not sure adding an intermediate variable helps anything.

lib/document.js
```diff
@@ -1015,17 +1015,21 @@ Document.prototype.$timestamps = function $timestamps(value) {
  */
 
 Document.prototype.overwrite = function overwrite(obj) {
-  const keys = Array.from(new Set(Object.keys(this._doc).concat(Object.keys(obj))));
+  const keys = new Set(Object.keys(this._doc));
+  for (const key of Object.keys(obj)) {
```
vkarpov15, line 1026: At first glance this seems strange, but you're right the loop approach looks faster:

https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJBN1lSaG94NUF0b3hBTDQxY1ZSUGEiLCJjb2RlIjoiQXJyYXkuZnJvbShuZXcgU2V0KFsnYScsICdjJywgJ2QnXS5jb25jYXQoREFUQSkpKSIsIm5hbWUiOiJGaW5kIDk5IiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6IkZ1b3lKdzZlMjVpTTFvOTBFZzFNdyIsImNvZGUiOiJjb25zdCB4ID0gbmV3IFNldChbJ2EnLCAnYycsICdkJ10pOyBmb3IgKGNvbnN0IHYgb2YgREFUQSkgeyB4LmFkZCh2KTsgfSIsIm5hbWUiOiJGaW5kIDE5OSIsImRlcGVuZGVuY2llcyI6W119XSwiY29uZmlnIjp7Im5hbWUiOiJCYXNpYyBleGFtcGxlIiwicGFyYWxsZWwiOnRydWUsImdsb2JhbFRlc3RDb25maWciOnsiZGVwZW5kZW5jaWVzIjpbXX0sImRhdGFDb2RlIjoicmV0dXJuIFsnYScsICdhJywgJ2InLCAnYyddIn19

image

vkarpov15 (changes requested):
Some of the changes look slower than what was originally there, most notably optional chaining and checking if Object.keys(obj).length.

lib/utils.js
AbdelrahmanHafez: There's a typo in the benchmark: `Object.prototype.hasOwnProperty` should be `Object.prototype.hasOwnProperty.call`. When [fixed](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJBN1lSaG94NUF0b3hBTDQxY1ZSUGEiLCJjb2RlIjoiaWYgKE9iamVjdC5rZXlzKERBVEEpLmxlbmd0aCA-IDApIHJldHVybiB0cnVlOyIsIm5hbWUiOiJPYmplY3Qua2V5cyguLi4pIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6IkZ1b3lKdzZlMjVpTTFvOTBFZzFNdyIsImNvZGUiOiJmb3IgKGNvbnN0IGtleSBpbiBEQVRBKSBpZiAoT2JqZWN0LnByb3RvdHlwZS5oYXNPd25Qcm9wZXJ0eS5jYWxsKERBVEEsIGtleSkpIHJldHVybiB0cnVlOyIsIm5hbWUiOiJmb3IuLi5pbiIsImRlcGVuZGVuY2llcyI6W119XSwiY29uZmlnIjp7Im5hbWUiOiJCYXNpYyBleGFtcGxlIiwicGFyYWxsZWwiOnRydWUsImdsb2JhbFRlc3RDb25maWciOnsiZGVwZW5kZW5jaWVzIjpbXX0sImRhdGFDb2RlIjoicmV0dXJuIHsgYTogMSB9OyJ9fQ) it shows that `for...in` is ~2x faster.

Image

Also testing on Node and the browser give different results, I ran benchmarks locally with different object sizes in Node and they show `for...in` is always faster with varying degrees, depending on the size of the object, which makes sense because we don't need to look up all the object keys, create an array with it and then check its length. We just look at the first key.


Here's the benchmark results if you're interested in looking deeper into this:
### Results on Node 22
| Scenario | `Object.keys` | `hasOwnProperty.call` |
|----------|-------------:|----------------------:|
| 0 keys | 25.2M ops/s | 25.7M ops/s (+2%) |
| 1 key | 24.6M ops/s | 24.7M ops/s (~0%) |
| 2 keys | 24.7M ops/s | 25.4M ops/s (+3%) |
| 5 keys | 23.8M ops/s | 25.7M ops/s (+8%) |
| 20 keys | 23.5M ops/s | 24.7M ops/s (+5%) |
| 100 keys | 16.8M ops/s | 25.5M ops/s (+52%) |

Results show even bigger gains in Node 24.

### Benchmark
```js
'use strict';

const { Bench } = require('tinybench');

const testCases = {
  '0 keys': {},
  '1 key': { a: 1 },
  '2 keys': { _id: '507f1f77bcf86cd799439011', name: 'Test' },
  '5 keys': { _id: '507f1f77bcf86cd799439011', name: 'Test', age: 25, email: 'test@test.com', active: true },
  '20 keys': Object.fromEntries(Array.from({ length: 20 }, (_, i) => [`key${i}`, i])),
  '100 keys': Object.fromEntries(Array.from({ length: 100 }, (_, i) => [`key${i}`, i]))
};

function withObjectKeys(obj) {
  return Object.keys(obj).length > 0;
}

function withHasOwnProperty(obj) {
  for (const key in obj) {
    if (Object.prototype.hasOwnProperty.call(obj, key)) return true;
  }
  return false;
}

function formatOps(n) {
  if (n >= 1e6) return (n / 1e6).toFixed(1) + 'M';
  if (n >= 1e3) return (n / 1e3).toFixed(1) + 'K';
  return n.toFixed(0);
}

function formatDiff(val, base) {
  const diff = ((val - base) / base) * 100;
  if (Math.abs(diff) < 1) return '~0%';
  return (diff > 0 ? '+' : '') + diff.toFixed(0) + '%';
}

async function run() {
  const bench = new Bench({ time: 500 });

  for (const [name, obj] of Object.entries(testCases)) {
    bench.add(`Object.keys - ${name}`, () => withObjectKeys(obj));
    bench.add(`hasOwnProperty - ${name}`, () => withHasOwnProperty(obj));
  }

  await bench.run();

  console.log('# hasOwnKeys Benchmark\n');
  console.log('| Scenario | `Object.keys` | `hasOwnProperty.call` |');
  console.log('|----------|-------------:|----------------------:|');

  for (const name of Object.keys(testCases)) {
    const ok = bench.tasks.find(t => t.name === `Object.keys - ${name}`).result.throughput.mean;
    const hp = bench.tasks.find(t => t.name === `hasOwnProperty - ${name}`).result.throughput.mean;

    const hpDiff = formatDiff(hp, ok);

    console.log(`| ${name} | ${formatOps(ok)} ops/s | ${formatOps(hp)} ops/s (${hpDiff}) |`);
  }
}

run();
```

On a different topic, interesting find re https://github.com/Automattic/mongoose/pull/15875 is that `Object.hasOwn(...)` is slower than `Object.prototype.hasOwnProperty`, we probably should bring back `hop` (below) in utils and use it everywhere instead of relying on `Object.hasOwn(...)`, but that's not directly related to this PR.

```js
const hop = Object.prototype.hasOwnProperty;

/**
 * Safer helper for hasOwnProperty checks
 *
 * @param {Object} obj
 * @param {String} prop
 */

exports.object.hasOwnProperty = function(obj, prop) {
  return hop.call(obj, prop);
};
```

lib/cast.js
AbdelrahmanHafez: Ineresting, your benchmark shows identical results on my machine:
optional chaining

I benchmarked in Node and the good news is that in addition to making code cleaner, optional chaining is faster in most scenarios.
### Results
| Scenario | `&&` chaining | `?.` chaining |
|----------|-------------:|-------------:|
| 3L exists | 26.3M ops/s | 35.3M ops/s (+34%) |
| 3L null root | 31.4M ops/s | 30.6M ops/s (-3%) |
| 3L missing .a | 35.0M ops/s | 34.9M ops/s (~0%) |
| 3L missing .b | 34.2M ops/s | 35.2M ops/s (+3%) |
| 3L missing .c | 34.3M ops/s | 34.2M ops/s (~0%) |
| 6L exists | 29.3M ops/s | 35.1M ops/s (+20%) |
| 6L null root | 35.2M ops/s | 35.0M ops/s (~0%) |
| 6L missing .a | 32.5M ops/s | 35.2M ops/s (+8%) |
| 6L missing .c | 33.7M ops/s | 34.8M ops/s (+3%) |
| 6L missing .f | 29.4M ops/s | 34.2M ops/s (+16%) |

### optional-chaining-benchmark.js
```js
'use strict';

const { Bench } = require('tinybench');

// Test cases with different scenarios (3 levels deep)
const testCases3 = {
  '3L exists': { a: { b: { c: 1 } } },
  '3L null root': null,
  '3L missing .a': {},
  '3L missing .b': { a: {} },
  '3L missing .c': { a: { b: {} } }
};

// 6 levels deep (common in Mongoose)
const testCases6 = {
  '6L exists': { a: { b: { c: { d: { e: { f: 1 } } } } } },
  '6L null root': null,
  '6L missing .a': {},
  '6L missing .c': { a: { b: {} } },
  '6L missing .f': { a: { b: { c: { d: { e: {} } } } } }
};

const testCases = { ...testCases3, ...testCases6 };

// 3-level functions
function withAnd3(obj) {
  return obj && obj.a && obj.a.b && obj.a.b.c;
}

function withOptional3(obj) {
  return obj?.a?.b?.c;
}

// 6-level functions
function withAnd6(obj) {
  return obj && obj.a && obj.a.b && obj.a.b.c && obj.a.b.c.d && obj.a.b.c.d.e && obj.a.b.c.d.e.f;
}

function withOptional6(obj) {
  return obj?.a?.b?.c?.d?.e?.f;
}

function formatOps(n) {
  if (n >= 1e6) return (n / 1e6).toFixed(1) + 'M';
  if (n >= 1e3) return (n / 1e3).toFixed(1) + 'K';
  return n.toFixed(0);
}

function formatDiff(val, base) {
  const diff = ((val - base) / base) * 100;
  if (Math.abs(diff) < 1) return '~0%';
  return (diff > 0 ? '+' : '') + diff.toFixed(0) + '%';
}

async function run() {
  const bench = new Bench({ time: 500 });

  // Add 3-level tests
  for (const [name, obj] of Object.entries(testCases3)) {
    bench.add(`&& - ${name}`, () => withAnd3(obj));
    bench.add(`?. - ${name}`, () => withOptional3(obj));
  }

  // Add 6-level tests
  for (const [name, obj] of Object.entries(testCases6)) {
    bench.add(`&& - ${name}`, () => withAnd6(obj));
    bench.add(`?. - ${name}`, () => withOptional6(obj));
  }

  await bench.run();

  console.log('# Optional Chaining vs AND Checking Benchmark\n');
  console.log('| Scenario | `&&` chaining | `?.` chaining |');
  console.log('|----------|-------------:|-------------:|');

  for (const name of Object.keys(testCases)) {
    const andResult = bench.tasks.find(t => t.name === `&& - ${name}`).result.throughput.mean;
    const optResult = bench.tasks.find(t => t.name === `?. - ${name}`).result.throughput.mean;

    const diff = formatDiff(optResult, andResult);

    console.log(`| ${name} | ${formatOps(andResult)} ops/s | ${formatOps(optResult)} ops/s (${diff}) |`);
  }
}

run();
```

lib/connection.js
AbdelrahmanHafez: That's a tiny optimization that honestly doesn't matter much, it's to avoid running `this.base.connections.indexOf(conn)` twice. Also as a nice plus it slightly improves readability.

AbdelrahmanHafez:
Changing this to draft since I found a bunch of other opportunities to refactor for optional chaining and simplify code in general.

lib/schema/buffer.js
```diff
@@ -146,12 +146,12 @@ SchemaBuffer.prototype.checkRequired = function(value, doc) {
...
   let ret;
   if (SchemaType._isRef(this, value, doc, init)) {
-    if (value && value.isMongooseBuffer) {
+    if (value?.isMongooseBuffer) {
       return value;
     }
 
     if (Buffer.isBuffer(value)) {
-      if (!value || !value.isMongooseBuffer) {
+      if (!value?.isMongooseBuffer) {
```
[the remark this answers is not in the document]
AbdelrahmanHafez: Unrelated to the changes in the PR, PR is already large enough as it is, don't wanna introduce more changes to it.

lib/schema/buffer.js
```diff
@@ -175,16 +175,16 @@ SchemaBuffer.prototype.cast = function(value, doc, init, prev, options) {
...
   }
 
-  if (value && value.isMongooseBuffer) {
+  if (value?.isMongooseBuffer) {
     return value;
   }
 
   if (Buffer.isBuffer(value)) {
-    if (!value || !value.isMongooseBuffer) {
+    if (!value?.isMongooseBuffer) {
```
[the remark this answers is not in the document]
AbdelrahmanHafez: Unrelated to PR.

lib/schema/union.js
```diff
@@ -24,7 +24,7 @@ class Union extends SchemaType {
    */
   constructor(key, options, schemaOptions, parentSchema) {
     super(key, options, 'Union', parentSchema);
-    if (!options || !Array.isArray(options.of) || options.of.length === 0) {
+    if (!Array.isArray(options?.of) || options.of.length === 0) {
```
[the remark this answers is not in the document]
AbdelrahmanHafez: The changes are actually equivalent. The key is that `Array.isArray(undefined)` returns `false`, so `!Array.isArray(undefined)` returns `true`.

When `options` is null/undefined, `options?.of` evaluates to `undefined`, and `!Array.isArray(undefined)` is `true`, same as the original `!options` check.

Same applies when `options` exists but has no `of` property or a non-array `of`. In all cases `!Array.isArray(options?.of)` correctly evaluates to `true`.

AbdelrahmanHafez:
@vkarpov15 This PR grew a bit larger than I initially intended, there were many places where optional chaining could improve readability and/or performance. I've reviewed it a couple of times myself, had Copilot validate it as well, and all tests and lint are passing, which should give us good confidence in these changes.

That said, if it would make things easier for you, I'm happy to split this into smaller PRs. Let me know if there's anything I can do to make the review process smoother.

lib/cast.js
vkarpov15: You're right, I double checked your results and optional chaining vs && is extremely close on Node and on Chrome. The results I posted in my original screenshot were on Firefox, which isn't relevant here.

vkarpov15:
This PR's great, no need for separate PRs. Thanks :+1:

Issue #15858, opened 2025-12-03
The $isDeleted is not set automatically as the documentation states.


djereg:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.20.1

### Node.js version

24.5.0

### MongoDB server version

8.0.3

### Typescript version (if applicable)

5.9.3

### Description

As the [documentation](https://mongoosejs.com/docs/8.x/docs/api/document.html#Document.prototype.$isDeleted()) says:

> The Model.prototype.deleteOne() method sets $isDeleted if the delete operation succeeded.

But it doesn't. The `$isDeleted()` method returns `false` after calling the `deleteOne()` method and the document has been deleted successfully.

The example given in documentation cannot be run because the product variable cannot be redeclared.

> TS2451: Cannot redeclare block-scoped variable `product`

The `deleteOne()` method returns [DeleteResult](https://github.com/mongodb/node-mongodb-native/blob/a96fa26dea3060828af747f0f2974229018de730/src/operations/delete.ts#L24-L29) not the document itself.

The way I achived the expected result is:

```typescript
await product.deleteOne();
product.$isDeleted(true);
```
And after I can check whether the document has been deleted or not.

### Steps to Reproduce

Follow the example on the  [documentation](https://mongoosejs.com/docs/8.x/docs/api/document.html#Document.prototype.$isDeleted())  page.

### Expected Behavior

The `$isDeleted()` method returns `true` after calling `deleteOne()` on the document and it has been successfully deleted from the database.

djereg:
As I see, this changed in v8. Since then, the `deleteOne` also returns a different value and although it takes into account the value of the `isDeleted` variable, but no longer sets it.

PR #15872, opened 2025-12-04
fix(document): ensure  is set to true after deleteOne()

jagadish-555:
Summary

This PR fixes issue #15858, where Model.prototype.deleteOne() was not setting the $isDeleted property to true on the document instance after a successful deletion.

This behavior contradicted the documentation, which states that calling deleteOne() should set $isDeleted to true when the operation succeeds.
 

Examples
```js
const doc = await Model.create({ name: 'test' });
await doc.deleteOne();

console.log(doc.$isDeleted()); // Now prints 'true' (used to be 'false')
```

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -788,6 +788,7 @@ Model.prototype.deleteOne = function deleteOne(options) {
     return Promise.all(self.$getAllSubdocs().map(subdoc => subdoc.constructor._middleware.execPost('deleteOne', subdoc, [subdoc])));
   });
   query.post(function queryPostDeleteOne() {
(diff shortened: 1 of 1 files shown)


AbdelrahmanHafez (changes requested):
Welcome @jagadish-555
Can you add a test-case that does not pass without your fix to prove that the bug exists?
Also tests are failing in the CI pipeline, can you fix them?

vkarpov15:
I did some digging, this PR misses a lot of key details and isn't close to mergeable. I'm closing.

PR #15877, opened 2025-12-05
fix: pass options to deleteOne middleware hooks

### LINKED ISSUE #15870: Difference in post-middleware arguments between .remove() and .deleteOne()
### Prerequisites

### Mongoose version

6.13.8

### Node.js version

22.21.1

### MongoDB version

8.x

### Operating system

Windows

11

### Issue

I needed some help understanding the difference between the options passed to `doc.remove()` in Mongoose 6.x vs `doc.deleteOne()`. From what I know, "remove" worked similar to "save", in that any options passed when calling `doc.remove(options)` were available in the post middleware. However, since I have switched to using "deleteOne", and attaching document middleware on it, the options object is no longer available as the second argument in post middleware. In fact, the document instance itself is the second argument.

I am maintaining some code which used to pass custom options (for example, request identifiers) when deleting the document, that I need to access in the post-middleware. Is there no way to access the options object similar to remove/save?

shalini-saloni:
Fixes #15870
### Summary
This PR enables access to custom options in `deleteOne()` middleware hooks by passing the `options` parameter to pre/post middleware, matching the behavior of `save()` middleware.
### Changes
- Modified `lib/model.js` to pass `options` instead of `self` to `execPre` and `execPost` calls for deleteOne middleware
- Added test case to verify options are accessible in middleware
- All existing deleteOne tests continue to pass (34 passing)
### Breaking Changes
None. This is a backward-compatible enhancement.
### Example Usage
\`\`\`javascript
schema.pre('deleteOne', { document: true, query: false }, function(options) {
  console.log(options.customOption); // Now accessible!
});
await doc.deleteOne({ customOption: 'value' });
\`\`\`

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -631,7 +631,7 @@ Model.prototype.$save = Model.prototype.save;
  * @instance
  */
 
-Model.prototype.$__version = function(where, delta) {
+Model.prototype.$__version = function (where, delta) {
   const key = this.$__schema.options.versionKey;
   if (where === true) {
     // this is an insert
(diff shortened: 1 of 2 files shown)


vkarpov15:
There's a bunch of unrelated style changes in this PR, I'm closing. Please create a clean PR if you want us to review.

PR #15879, opened 2025-12-08
docs: improve colors on dark mode

AbdelrahmanHafez:
# Summary

This PR fixes a few issues with dark mode, mostly contrasts or areas not covered by the original dark mode version.

I'm colorblind so I particularly found that the logo on the left menu had poor contrast and fails [WCAG contrast test](https://webaim.org/resources/contrastchecker/). I'm not a designer so I used AI to decide the correct color for the mongoose brand on dark mode, and it calculated `#FFB4B4` as the mongoose color for dark backgrounds, which passes WCAG contrast test. I modified the CSS code to be parameterized, so that if we decide to change the color from #FFB4B4 to something else, we just need to change it in one place and re-generate the logo with that new color.

### WCAG Contrast Test

| Before | After |
|--------|-------|
| WCAG before | WCAG after |

### Homepage & GitHub Buttons

| Before | After |
|--------|-------|
| Homepage before | Homepage after |

### Logo & Sidebar Menu

| Before | After |
|--------|-------|
| Logo before | Logo after |

### Version Dropdown

| Before | After |
|--------|-------|
| Version dropdown before | Version dropdown after |

### Enterprise Page Buttons

| Before | After |
|--------|-------|
| Enterprise buttons before | Enterprise buttons after |

### Jobs Panel

| Before | After |
|--------|-------|
| Jobs panel before | Jobs panel after |

### Consolidated Theme Variables

Extracted all hardcoded color values into CSS custom properties defined in `mongoose5.css`:
- `--brand-primary` - Mongoose brand color (`#800` light, `#FFB4B4` dark)
- `--btn-bg`, `--btn-text-shadow` - Button styling
- `--tagline-text-shadow` - Homepage tagline effect
- `--focus-ring`, `--focus-ring-shadow` - Accessibility focus indicators

### SVG Logo with CSS-controlled Color

Replaced PNG logo swap approach with a single SVG logo using CSS `mask-image`:
- Modified `mongoose.svg` to use `currentColor` instead of hardcoded `#880000`
- Logo element uses `background-color: var(--brand-primary)` with SVG as mask
- Eliminates need for separate light/dark PNG files
- Logo color automatically follows theme via CSS variable

### No-JS / System Preference Support

Added `@media (prefers-color-scheme: dark)` fallbacks alongside `[data-theme="dark"]` selectors:
- Prevents white flash on initial page load for dark mode users
- Supports users with JavaScript disabled
- Uses `:root:not([data-theme="light"])` pattern to allow manual override

### Consolidated Dark Mode Selectors

- Replaced `.code-theme-dark` class with `[data-theme="dark"]` attribute selectors
- Removed redundant class toggling from `theme-toggle.js`
- All dark mode styles now consistently use `[data-theme="dark"]`

### Other Improvements

- Fixed squished text in homepage code blocks (`line-height: 1.5`)
- Removed CSS variable fallbacks for consistency (all variables defined in `:root`)
- Added `.mongoose-btn-outline` and `.mongoose-btn-solid` classes for brand buttons
- Fixed various hardcoded colors in Jobs panel, version dropdown, search input

### DIFF
--- a/docs/css/mongoose5.css
+++ b/docs/css/mongoose5.css
@@ -34,15 +34,15 @@
 @media (prefers-color-scheme: dark) {
   :root:not([data-theme="light"]) {
     --bg-primary: #1a1a1a;
-    --bg-secondary: #2d2d2d;
+    --bg-secondary: #252525;
     --bg-tertiary: #252525;
     --text-primary: #f0f0f0;
     --text-secondary: #d8d8d8;
     --text-muted: #a0a0a0;
-    --link-color: #FFB4B4;
-    --link-hover: #FFCECE;
+    --link-color: #7BC3FF;
+    --link-hover: #8AD0FF;
     --border-color: #444444;
-    --code-bg: #2d2d2d;
+    --code-bg: #252525;
     --code-text: #e8e8e8;
     --menu-bg: #252525;
     --menu-hover: rgba(255, 255, 255, 0.1);
@@ -51,8 +51,8 @@
     --focus-ring: #FFB4B4;
     --focus-ring-shadow: rgba(255, 180, 180, 0.2);
     /* Brand colors */
-    --brand-primary: #FFB4B4;
-    --brand-primary-hover: #FFCECE;
+    --brand-primary: #FF5E5E;
+    --brand-primary-hover: #F55;
     --brand-primary-subtle: rgba(255, 180, 180, 0.1);
     /* Button colors */
     --btn-bg: #444;
@@ -70,10 +70,10 @@
   --text-primary: #f0f0f0;
   --text-secondary: #d8d8d8;
   --text-muted: #a0a0a0;
-  --link-color: #FFB4B4;
-  --link-hover: #FFCECE;
+  --link-color: #7BC3FF;
+  --link-hover: #8AD0FF;
   --border-color: #444444;
-  --code-bg: #2d2d2d;
+  --code-bg: #252525;
   --code-text: #e8e8e8;
   --menu-bg: #252525;
   --menu-hover: rgba(255, 255, 255, 0.1);
@@ -82,8 +82,8 @@
   --focus-ring: #FFB4B4;
   --focus-ring-shadow: rgba(255, 180, 180, 0.2);
   /* Brand colors */
-  --brand-primary: #FFB4B4;
-  --brand-primary-hover: #FFCECE;
+  --brand-primary: #FF5E5E;
+  --brand-primary-hover: #F55;
   --brand-primary-subtle: rgba(255, 180, 180, 0.1);
   /* Button colors */
   --btn-bg: #444;
@@ -221,7 +221,7 @@ h4 a:focus-visible {
   padding-bottom: 0px;
 }
 
-.pure-menu-link:hover, 
+.pure-menu-link:hover,
 .pure-menu-link.selected {
   background-color: var(--menu-hover);
 }


docs/css/style.css
```diff
@@ -1,52 +1,28 @@
 body {
   font-family: 'Open Sans', Helvetica, Arial, FreeSans;
-  color: var(--text-secondary, #333);
+  color: var(--text-secondary);
   -webkit-font-smoothing: antialiased;
   -webkit-text-size-adjust: 100%;
   padding: 0;
   margin: 0;
-  background-color: var(--bg-primary, #ffffff);
+  background-color: var(--bg-primary);
```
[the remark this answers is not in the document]
AbdelrahmanHafez: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Automattic/mongoose/pull/15879#discussion_r2596864188)

docs/css/style.css
```diff
@@ -288,6 +248,15 @@ h2 a {
...
   top: -4px;
 }
+/* GitHub buttons iframe - use color-scheme to make iframe bg transparent in dark mode */
+.github-btn {
+  background: transparent;
+}
+[data-theme="dark"] .github-btn {
+  color-scheme: light;
+  filter: invert(0.85) hue-rotate(180deg);
+  background: transparent;
```
hasezoey: Isnt this property already inherited by the definition above?

docs/images/mongoose5_62x30_transparent_light.png
hasezoey, line 1: Maybe we should switch always use the SVG instead and color that?

AbdelrahmanHafez:
@hasezoey addressed feedback:

> white flash when moving between pages due to removal of @media query

Fixed - restored `@media (prefers-color-scheme: dark)` queries using the `:root:not([data-theme="light"])` pattern alongside `[data-theme="dark"]` selectors. This handles both no-JS and the initial page load flash.

> code blocks still look way too blue / API pages also look way too blue

Changed dark mode `--link-color` from `#5aabff` to `#FFB4B4` (brand color) - matches what you showed in your screenshot.

> couldn't class `code-theme-dark` be combined into `data-theme`?

Done - replaced all `.code-theme-dark` with `[data-theme="dark"]` in github.css and removed the class toggling from theme-toggle.js.

> squished text in homepage code blocks is back

Added line-height: 1.6 to pre code in github.css - the code element wasn't inheriting line-height from its parent pre.

> Maybe we should switch to always use the SVG instead and color that?

Good call, changed `mongoose.svg` to use `currentColor` and switched to CSS `mask-image`:
```css
#logo {
  background-color: var(--brand-primary);
  mask-image: url('/docs/images/mongoose.svg');
}
```

> Isn't this property already inherited by the definition above?

Removed the redundant `background: transparent`.

vkarpov15:
Minor nitpick but I think the new dark color `--brand-primary` and `--brand-primary-hover` are a bit too different from current, WDYT about #FF4747 and #F55? I think this is more faithful while still passing contrast check

image

image

Also, I think I'd prefer to keep the previous link color #5aabff

hasezoey:
FYI [`#FF4747` is `5.17:1`](https://webaim.org/resources/contrastchecker/?fcolor=FF4747&bcolor=1A1A1A) contrast and [`#FF5555` is `5.53:1`](https://webaim.org/resources/contrastchecker/?fcolor=FF5555&bcolor=1A1A1A) contrast, not `6.24:1` as the background is `#1A1A1A` not `#000000` (though that does not change the listed Pass/Fails). (please note that i have no clue about colors or contrast)
Personally, this "fuller color" looks better to me.

> Also, I think I'd prefer to keep the previous link color #5aabff

If that should be done, i would like to split link color and code color as otherwise i really think the API page (and any code block) will look too-blue. (and may be confusing as most of the code block things are the link color, but cannot actually be clicked).

AbdelrahmanHafez:
> Minor nitpick but I think the new dark color `--brand-primary` and `--brand-primary-hover` are a bit too different from current, WDYT about #FF4747 and #F55?

I checked the contrast ratios - `#FF4747` comes in at 5.17:1 against `#1A1A1A` (as @hasezoey pointed out). That passes WCAG AA but fails WCAG AAA (requires 7:1). About 8% of men have some form of red-green color blindness, and lower contrast ratios can make text difficult to read for this group.

I'm not attached to `#FFB4B4` specifically - that was just what an AI suggested. Any color that clears WCAG AAA works. If there's something closer to the original brand feel while hitting 7:1, happy to use it.

> Also, I think I'd prefer to keep the previous link color #5aabff

Can do. @hasezoey's suggestion about splitting link color from code color makes sense if you want to avoid the "too blue" API pages - I'll leave that decision to you two.

vkarpov15:
For #FF4747, I'm suggesting using that for the `brand-primary`, which is only for large text - #FF4747 still passes for large text. Admittedly that color choice is more about style than readability, but I want to retain more of a red color than drifting into light pink. I also updated the links to #7BC3FF, which has good contrast with the dark background and more closely matches the light theme links.

vkarpov15 (approved):

PR #15882, opened 2025-12-09
fix(model): trigger error post hook on bulkwrite when pre-hook throws an error

### LINKED ISSUE #15881: [Bug] bulkWrite pre-hook errors do not trigger post error hooks
### Prerequisites

### Mongoose version

9.0.1

### Node.js version

N/A

### MongoDB server version

N/A

N/A

### Description

When a `bulkWrite` pre-hook throws an error, post-hooks are not called. This is inconsistent with other middleware like `save()` and `insertMany()`, which correctly call post-hooks when pre-hooks throw.

### Steps to Reproduce

```javascript
'use strict';

run().catch(err => {
  console.error(err);
  process.exit(1);
});

async function run() {

  // Test 1: insertMany correctly calls post-hooks (WORKS)
  {

    let postErrorHookCalled = false;
    schema.pre('insertMany', function() {
      throw new Error('insertMany pre-hook error');
    });
    schema.post('insertMany', function(error, docs, next) {
      postErrorHookCalled = true;
      next(error);
    });

    const Model = mongoose.model('InsertManyTest', schema);

    const err = await Model.insertMany([{ name: 'test' }]).catch(e => e);
    assert.equal(err.message, 'insertMany pre-hook error');
    assert.equal(postErrorHookCalled, true, 'insertMany: post-hook should be called');
    console.log('insertMany: PASS - post-hook was called');
  }

  // Test 2: bulkWrite does NOT call post-hooks (BUG)
  {

    let postErrorHookCalled = false;
    schema.pre('bulkWrite', function() {
      throw new Error('bulkWrite pre-hook error');
    });
    schema.post('bulkWrite', function(error, res, next) {
      postErrorHookCalled = true;
      next(error);
    });

    const Model = mongoose.model('BulkWriteTest', schema);

    const err = await Model.bulkWrite([{ insertOne: { document: { name: 'test' } } }]).catch(e => e);
    assert.equal(err.message, 'bulkWrite pre-hook error');
    assert.equal(postErrorHookCalled, true, 'bulkWrite: post-hook should be called');
    console.log('bulkWrite: PASS - post-hook was called');
  }

  console.log('All tests passed!');
  await mongoose.disconnect();
}
```

**Actual output:**

```
insertMany: PASS - post-hook was called
AssertionError [ERR_ASSERTION]: bulkWrite: post-hook should be called
```

### Expected Behavior

**Expected output:**
```
insertMany: PASS - post-hook was called
bulkWrite: PASS - post-hook was called
All tests passed!
```

Both `insertMany` and `bulkWrite` should call post-hooks when pre-hooks throw an error. This is consistent with:

1. Kareem's `wrap()` function, which calls `execPost` with `{ error }` when pre-hooks throw
2. Document middleware like `save()`, which calls post-hooks when pre-hooks throw
3. `insertMany()`, which correctly calls post-hooks when pre-hooks throw

AbdelrahmanHafez:
## Summary

Fix `bulkWrite()` to call error post hooks when a pre-hook throws an error. Fixes #15881 

## Changes

- Call `execPost` with `{ error }` when `bulkWrite` pre-hook throws, matching `insertMany` behavior

## Before

```javascript
[ops, options] = await this.hooks.execPre(...).catch(err => {
  if (err instanceof Kareem.skipWrappedFunction) return [err];
  throw err;  // error post hooks not called
});
```

## After

```javascript
try {
  [ops, options] = await this.hooks.execPre(...);
} catch (err) {
  if (err instanceof Kareem.skipWrappedFunction) {
    ops = err;
  } else {
    await this.hooks.execPost('bulkWrite', this, [null], { error: err });
  }
}
```

### DIFF
--- a/test/model.insertMany.test.js
+++ b/test/model.insertMany.test.js
@@ -739,7 +739,7 @@ describe('insertMany()', function() {
     assert.ok(err.stack.includes('postInsertManyError'));
   });
 
-  describe.only('pre-hook errors should propagate', function() {
+  describe('pre-hook errors should propagate', function() {
     it('insertMany() should throw when pre-hook throws an error', async function() {
       // Arrange
       const preHookError = new Error('Pre-hook error - should stop insertMany');
--- a/test/model.middleware.test.js
+++ b/test/model.middleware.test.js
@@ -625,7 +625,7 @@ describe('model middleware', function() {
       assert.strictEqual(res, 'skipMiddlewareFunction test');
     });
 
-    describe.only('pre-hook errors should propagate (gh-15881)', function() {
+    describe('pre-hook errors should propagate (gh-15881)', function() {
       for (const ordered of [true, false]) {
         it(`bulkWrite() should throw when pre-hook throws an error (ordered: ${ordered})`, async function() {
           // Arrange


lib/model.js
```diff
@@ -3286,12 +3286,15 @@ Model.bulkWrite = async function bulkWrite(ops, options) {
...
 
-  [ops, options] = await this.hooks.execPre('bulkWrite', this, [ops, options]).catch(err => {
+  try {
+    [ops, options] = await this.hooks.execPre('bulkWrite', this, [ops, options]);
+  } catch (err) {
     if (err instanceof Kareem.skipWrappedFunction) {
-      return [err];
+      ops = err;
+    } else {
+      await this.hooks.execPost('bulkWrite', this, [null], { error: err });
```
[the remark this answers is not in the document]
AbdelrahmanHafez: Kareem re-throws on `execPost`

hasezoey (commented):
LGTM

test/model.insertMany.test.js
```diff
@@ -738,4 +738,77 @@ describe('insertMany()', function() {
...
+    });
+
+    function createTestContext(options) {
+      const schema = new Schema({ name: String });
+
+      schema.pre('insertMany', function() {
+        throw options.preHookError;
+      });
+
+      if (options.postHook) {
```
vkarpov15, line 802: `createTextContext()` is fine, but in this case it feels like you're building a mini-framework for this test suite, which hurts readability and doesn't meaningfully reduce SLOC. Fine for now, but let's avoid creating too many abstractions in the future.

AbdelrahmanHafez: @vkarpov15 
My aim is to have the `it('...')` blocks be as minimal as possible and only showing the important parts to understand what a test does, with building the "common" blocks in `createTestContext`, which I think you find acceptable.
Do you think instead of having a generic `createTestContext`, we should try to have multiple `createTestContext`s specific to each use-case even if they're only slightly different?

Issue #8768, opened 2020-04-04
[Feature] option to disable running pre/post hooks


AbdelrahmanHafez:
I am not sure whether this is present or not, but it would be great if we have a way to optionally disable running middlewares, a couple of use cases that I could think of is for internal use in the library, this would help us avoid situations like #8739.

Other developers could also use this in their applications if they need to run a query without triggering middlewares, or inside the logic of the middleware in itself to avoid an infinite loop such as the custom option in https://github.com/Automattic/mongoose/issues/8756#issuecomment-607605315.

Considering the following API:
```js
User.findOne({ name: 'John' }, null, { middlewares: false })
```

Later on, we could make it so that the options accept an object
```js
{ middlewares: { pre: true, post: false } }
```

vkarpov15:
I like this idea, definitely something to consider for a future minor release. Right now the only easy workaround is to just use `User.collection.findOne()`, but then that bypasses all of Mongoose.

I think I'd call the option `middleware` though, not `middlewares`.

AbdelrahmanHafez:
@vkarpov15 I've been trying to implement this, but I couldn't find where the middleware-execution-related code is.

I see that we're defining pre/post hooks using Kareem, and in `schema.s.hooks`. Could you point me in the right direction?

vkarpov15:
@AbdelrahmanHafez Kareem's `createWrapper()` function is probably where these changes should end up. Mongoose calls that function to wrap Mongoose's built-in functions with middleware for [documents](https://github.com/Automattic/mongoose/blob/master/lib/helpers/model/applyHooks.js) and [queries](https://github.com/Automattic/mongoose/blob/master/lib/helpers/query/applyQueryMiddleware.js)

ahmedshuhel:
@AbdelrahmanHafez did you make any progress on this one? It could be a very useful feature. In our case, we have a bunch of scripts written, mostly for data migration, etc where we want the pre/post hooks to be disabled by default. 
I was thinking of a global setting like `mongoose.set('disableMiddlewares', true)` with `false` by default. Can we consider that?

madaher-dev:
We have found a workaround for this by adding disableMiddlewares field in the query options and then check for it in the middleware itself
```js
 let options = {
      new: true,
      upsert: true,
      setDefaultsOnInsert: true,
      rawResult: true,
      disableMiddlewares: true, // can be checked in middleware with this.options.disableMiddlewares
    }; 

```
```js
[...]

PR #15883, opened 2025-12-09
feat: add option to skip middleware

AbdelrahmanHafez:
## Summary

This PR adds a `middleware` option to Mongoose operations that allows skipping pre and/or post hooks selectively. This addresses #8768.

### The Problem

Users needed a way to skip middleware for scenarios like:

- Data migrations where hooks interfere with bulk processing
- Performance-critical paths where hook overhead matters  
- Administrative operations that should bypass business logic
- Avoiding infinite loops when hooks need to perform the same operation

Existing workarounds required manual effort:
- Pass custom options and check them inside each hook (`this.getOptions().disableMiddlewares`)
- Use `Model.collection.findOne()` which bypasses Mongoose entirely
- Use `Model.bulkWrite()` for save operations (skips all middleware)

This PR provides a first-class option that works consistently across all operations.

### The Solution

The new `middleware` option accepts three forms:

```javascript
// Skip all hooks
await Model.find({}, null, { middleware: false });
await doc.save({ middleware: false });

// Skip only pre hooks
await Model.updateOne({}, { name: 'x' }, { middleware: { pre: false } });

// Skip only post hooks  
await Model.insertMany([...], { middleware: { post: false } });
```

### Key Implementation Details

**Supported operations:** `find`, `findOne`, `findOneAndUpdate`, `findOneAndDelete`, `findOneAndReplace`, `updateOne`, `updateMany`, `deleteOne`, `deleteMany`, `countDocuments`, `replaceOne`, `save`, `validate`, `insertMany`, `bulkWrite`, `aggregate`, and document-level `updateOne`.

**Subdocument cascading:** When saving a parent document with `middleware: false`, the option cascades to all subdocuments - their hooks are also skipped.

**Validation is NOT affected:** The `middleware` option on `save()` only skips save hooks - validation still runs. This is intentional because validation is a safety feature. To skip validation, use the existing `validateBeforeSave: false` option.

To support this, I moved validation logic out of the `validateBeforeSave` plugin (which was a pre-save hook) into a new `$__validateBeforeSave()` method called directly in `$__save()`. This ensures validation runs regardless of the middleware option.
This is probably worth discussing whether or not it's a safe change to introduce, currently all tests pass, and even though if it was common knowledge that `validateBeforeSave` was a pre-hook, I can't think of any way it being moved to a normal method call would break users applications, it's an implementation detail that's hidden.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "type": "commonjs",
   "license": "MIT",
   "dependencies": {
-    "kareem": "github:AbdelrahmanHafez/kareem#feat/add-filter-option-to-exec-pre-sync",
+    "kareem": "3.2.0",
     "mongodb": "~7.0",
     "mpath": "0.9.0",
     "mquery": "6.0.0",


lib/model.js
```diff
@@ -360,15 +361,54 @@ function _createSaveOptions(doc, options) {
   return saveOptions;
 }
 
+/*!
+ * ignore
+ */
+
+Model.prototype.$__validateBeforeSave = async function $__validateBeforeSave(options) {
```
hasezoey: If this is now a function on `Model`, maybe it should get some JSDoc.

AbdelrahmanHafez:
Note to self:
Find out if we need to do something with this code block in `model.js` around line 3650, and all other `.execPre('` parts in the codebase.
The skip-middlewares feature should not change the behavior for anything internal and should only enable/disable user-defined middlewares.
```js
function buildPreSavePromise(document, options) {
  return new Promise((resolve, reject) => {
    document.schema.s.hooks.execPre('save', document, [options], (err) => {
      if (err) {
        reject(err);
        return;
      }
      resolve();
    });
  });
}
```

lib/helpers/query/shouldSkipMiddleware.js
```diff
@@ -0,0 +1,19 @@
...
+ * @param {Object} options - The options object
+ * @param {String} phase - Either 'pre' or 'post'
+ * @returns {Boolean} - True if middleware should be skipped
+ */
+
+module.exports = function shouldSkipMiddleware(options, phase) {
+  if (options?.middleware === false) {
+    return true;
+  }
+  if (typeof options?.middleware === 'object') {
```
vkarpov15: Need to check if options.middleware == null, otherwise the next line will throw. Remember `typeof null === 'object'`

lib/helpers/shouldSkipMiddleware.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
```
vkarpov15: This isn't strictly a query helper because documents use it too

lib/document.js
```diff
@@ -857,10 +859,10 @@ Document.prototype.updateOne = function updateOne(doc, options) {
   const query = this.constructor.updateOne({ _id: this._doc._id }, doc, options);
   const self = this;
   query.pre(async function queryPreUpdateOne() {
-    const res = await self._execDocumentPreHooks('updateOne', self);
+    const res = await self._execDocumentPreHooks('updateOne', options, self);
```
vkarpov15: This looks like a breaking change: `pre('updateOne', query => { ... })` will assume that the first parameter to pre('updateOne') is a query, not options.

lib/helpers/shouldSkipMiddleware.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Check if middleware should be skipped for a given phase.
+ *
+ * @param {Object} options - The options object
+ * @param {String} phase - Either 'pre' or 'post'
+ * @returns {Boolean} - True if middleware should be skipped
+ */
+
+module.exports = function shouldSkipMiddleware(options, phase) {
```
vkarpov15: I'm not a fan of these sort of micro-helpers, they're verbose and obfuscate logic that's sufficiently concise to inline. Could easily be one-liner as `const shouldRunPreMiddleware = options?.middleware?.pre ?? options?.middleware ?? true`

lib/document.js
```diff
@@ -2938,15 +2940,24 @@ function _pushNestedArrayPaths(val, paths, path) {
  * ignore
  */
 
-Document.prototype._execDocumentPreHooks = async function _execDocumentPreHooks(opName, ...args) {
-  return this.$__middleware.execPre(opName, this, [...args]);
+Document.prototype._execDocumentPreHooks = async function _execDocumentPreHooks(opName, options, ...args) {
+  if (shouldSkipMiddleware(options, 'pre')) {
+    return [options, ...args];
+  }
+  return this.$__middleware.execPre(opName, this, [options, ...args]);
```
vkarpov15: similarly potentially backwards breaking here - adding `options` to all document pre hooks.

lib/document.js
AbdelrahmanHafez: I changed `_execDocumentPreHooks` to explicitly accept `argsForHooks` which are then passed to `this.$__middleware.execPre`, this should maintain the arguments the users receive on their hooks as is.

lib/helpers/shouldSkipMiddleware.js
AbdelrahmanHafez: I'd prefer to keep this as a separate helper. The `if (shouldSkipMiddleware(options, 'pre'))` reads clearly without needing to parse the condition, without a separate variable, and if someone wants to understand the logic they can easily jump to the function.
Inlining would mean repeating the ternary 17 times (across document.js, query.js, aggregate.js, model.js) and creating a local variable at each call site that the reader still has to mentally parse.

If you feel strongly about it, I'm happy to change it, but I think it makes our lives slightly easier to have it wrapped in a function.

AbdelrahmanHafez:
@vkarpov15 I moved `validateBeforeSave` out of middleware into a direct call, but there are more internal middlewares that would be skipped with `middleware: false`:

- **saveSubdocs** - triggers subdoc pre/post hooks
- **sharding** - adds shard key to `$where` (breaks sharded clusters if skipped)
- **trackTransaction** - stores doc state for transaction rollback
- **timestamps** - sets createdAt/updatedAt
- **virtual populate** - hydrates populated virtuals on init

I can think of the following options:

1. Extend `builtInMiddleware` symbol, already used for timestamp query hooks, could extend to filter internal hooks from being skipped

2. Separate internal hooks registry, two Kareem instances:
   ```js
   this.s.hooks = new Kareem();         // user hooks (skippable)
   this.s.internalHooks = new Kareem(); // always run
   ```

3. Call logic directly, move internal plugin logic out of hooks entirely (like I did with validateBeforeSave)

How do you think we should proceed? I personally would like to get rid of internal middlewares in favor of direct calls, or to separate them in a new Kareem instance to avoid accidentally interacting with user-defined hooks. What do you think?

lib/document.js
AbdelrahmanHafez: Please check current implementation with `argsForHooks`, should fix this.

lib/helpers/query/shouldSkipMiddleware.js
AbdelrahmanHafez: Good catch 👍

lib/model.js
```diff
@@ -360,15 +361,64 @@ function _createSaveOptions(doc, options) {
...
+
+  let shouldValidate;
+  if (hasValidateBeforeSaveOption) {
+    shouldValidate = !!options.validateBeforeSave;
+  } else {
+    shouldValidate = this.$__schema.options.validateBeforeSave;
+  }
+
+  if (shouldValidate) {
+    const hasValidateModifiedOnlyOption = options &&
```
vkarpov15: This would be a good candidate for optional chaining. Not a blocker but nice to fix if you have a minute

lib/helpers/shouldSkipMiddleware.js
vkarpov15: We can keep this as a helper, but let's make the helper more concise - this helper should be a one-liner using optional chaining and `??`.

lib/helpers/filterBuiltInMiddleware.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
```
AbdelrahmanHafez: This file won't be necessary if we publish https://github.com/mongoosejs/kareem/pull/43

lib/plugins/saveSubdocs.js
```diff
@@ -1,75 +1,92 @@
...
+
+  // Invalidate subdocs cache because subdoc pre hooks can add new subdocuments
+  if (this.$__.saveOptions) {
+    this.$__.saveOptions.__subdocs = null;
+  }
+}
+
+async function saveSubdocsPreDeleteOne() {
+  const removedSubdocs = this.$__.removedSubdocs;
+  if (!removedSubdocs || !removedSubdocs.length) {
```
AbdelrahmanHafez:
```suggestion
  if (!removedSubdocs?.length) {
```

AbdelrahmanHafez:
This PR is blocked by https://github.com/mongoosejs/kareem/pull/43

lib/helpers/filterBuiltInMiddleware.js
vkarpov15: @AbdelrahmanHafez I released kareem@3.1.0

lib/plugins/saveSubdocs.js
```diff
@@ -1,76 +1,94 @@
...
-
-    const promises = [];
-    for (const subdoc of subdocs) {
-      promises.push(subdoc._execDocumentPostHooks('save'));
-    }
-
-    await Promise.all(promises);
-  }, null, unshift);
+
+  saveSubdocsPreSave[symbols.builtInMiddleware] = true;
```
AbdelrahmanHafez: The diff in this file looks more complex than it is:
1. Adds `[symbols.builtInMiddleware] = true`.
2. Passes `options` to `_execDocumentPreHooks`, and `_execDocumentPostHooks`.
3. Rearranges functions to be `preSave, postSave, preDeleteOne, postDeleteOne`.

lib/aggregate.js
```diff
@@ -800,13 +801,20 @@ Aggregate.prototype.explain = async function explain(verbosity) {
...
   try {
-    await model.hooks.execPre('aggregate', this);
+    await model.hooks.execPre('aggregate', this, [], { filter: preFilter });
   } catch (error) {
-    return await model.hooks.execPost('aggregate', this, [null], { error });
+    return await model.hooks.execPost('aggregate', this, [null], { error, filter: postFilter });
   }
 
-  const cursor = model.collection.aggregate(this._pipeline, this.options);
+  const cursor = await model.collection.aggregate(this._pipeline, options);
```
AbdelrahmanHafez, line 817: The missing `await` here was causing flaky behavior when running before a connection was established.

lib/helpers/model/applyHooks.js
```diff
@@ -84,7 +86,16 @@ function applyHooks(model, schema, options) {
...
+    createWrapperSync('init', objToDecorate.$__init, null, {
+      ...kareemOptions,
+      getOptions: (args) => {
+        const opts = args[1];
+        return {
+          pre: { filter: buildMiddlewareFilter(opts, 'pre') },
+          post: { filter: buildMiddlewareFilter(opts, 'post') }
+        };
+      }
+    });
```
AbdelrahmanHafez, line 98: This is necessary because we need to pass the options all the way to the underlying `execPreSync` and `execPostSync`

AbdelrahmanHafez:
This is finally ready for review, builds on top of https://github.com/mongoosejs/kareem/pull/44, I tested locally with the Kareem version in the PR and all tests are passing. @vkarpov15

lib/helpers/buildMiddlewareFilter.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {Object} options - Options object that may contain `middleware` setting
+ * @param {String} phase - Either 'pre' or 'post'
+ * @returns {Function|null} - null runs all middleware, isBuiltInMiddleware skips user middleware
+ */
+function buildMiddlewareFilter(options, phase) {
+  const shouldRun = options?.middleware?.[phase] ?? options?.middleware ?? true;
+  return shouldRun ? null : isBuiltInMiddleware;
+}
+
+module.exports = {
```
vkarpov15, line 22: Minor nit: can we rename this file buildMiddlewareFilter.js? Most other places a file that exports only 1 function has the same name as the function for easier search

vkarpov15 (commented):
LGTM minus the test failures - looks like skipping `init` middleware isn't working as expected. Maybe the `middleware` option for `findOne()` isn't making it down to `init`?

AbdelrahmanHafez:
Test failure would be fixed with the PR in Kareem, there's [a comment](https://github.com/Automattic/mongoose/pull/15883#issuecomment-3754739243) before yours that's probably hidden by GitHub "Load more items" button.

lib/helpers/buildMiddlewareFilter.js
AbdelrahmanHafez: Makes sense.

lib/helpers/timestamps/setupTimestamps.js
```diff
@@ -42,14 +42,17 @@ module.exports = function setupTimestamps(schema, timestamps) {
 
   schema.add(schemaAdditions);
 
-  schema.pre('save', function timestampsPreSave() {
+  timestampsPreSave[symbols.builtInMiddleware] = true;
+  schema.pre('save', timestampsPreSave);
+
+  function timestampsPreSave() {
```
hasezoey: This might work, but it looks weird to define something on something that is only defined later.

lib/plugins/saveSubdocs.js
hasezoey: Shouldnt this symbol be applied statically instead of each function(saveSubdocs) call?
(this applies to basically all changed files, i will not make a comment each time)

lib/aggregate.js
```diff
@@ -893,6 +901,7 @@ Aggregate.prototype.session = function(session) {
  * @param {Boolean} [options.allowDiskUse] boolean if true, the MongoDB server will use the hard drive to store data during this aggregation
  * @param {Object} [options.collation] object see [`Aggregate.prototype.collation()`](https://mongoosejs.com/docs/api/aggregate.html#Aggregate.prototype.collation())
  * @param {ClientSession} [options.session] ClientSession see [`Aggregate.prototype.session()`](https://mongoosejs.com/docs/api/aggregate.html#Aggregate.prototype.session())
+ * @param {Boolean|Object} [options.middleware] set to `false` to skip all user-defined middleware, or `{ pre: false }` / `{ post: false }` to skip only pre or post hooks
```
hasezoey: Maybe it might be worth to add doc-comments for `options.middleware.pre/post`?

AbdelrahmanHafez:
All good suggestions @hasezoey, addressed.
Also great idea re: package.json change.

The builtInMiddleware static setting was a bit tricky for `setupTimestamps` since it relied on closure, but I refactored it to look up the properties from context instead and extracted it out of the function.

lib/helpers/timestamps/setupTimestamps.js
```diff
@@ -85,32 +78,56 @@ module.exports = function setupTimestamps(schema, timestamps) {
...
-      now,
-      createdAt,
-      updatedAt,
-      this.getUpdate(),
-      this._mongooseOptions,
-      replaceOps.has(this.op)
-    );
-    applyTimestampsToChildren(now, this.getUpdate(), this.model.schema);
+function timestampsPreSave() {
+  const timestampOption = get(this, '$__.saveOptions.timestamps');
```
vkarpov15: We should really avoid using `get()` unless necessary because `get()` is significantly slower than `this.$__?.saveOptions?.timestamps`

vkarpov15: I fixed this in [a85fa2c](https://github.com/Automattic/mongoose/pull/15883/commits/a85fa2c4f1cd55c16dda98863a27f032de688e1c)

PR #15886, opened 2025-12-11
ci(benchmark): use shallow clone for faster checkout

AbdelrahmanHafez:
The benchmark workflow was using `fetch-depth: 0` which fetches the entire git history. This is unnecessary since the TypeScript benchmark just runs type-checking on the current code and doesn't need any git history.

`actions/checkout` defaults to `fetch-depth: 1` (shallow clone), so removing the explicit setting lets it use the faster default. All other workflows in the repo already use this default.

Also bumps the TS instantiation limit from 390k to 400k since CI was already failing (398250 instantiations).

### DIFF
--- a/scripts/tsc-diagnostics-check.js
+++ b/scripts/tsc-diagnostics-check.js
@@ -3,7 +3,7 @@
 const fs = require('fs');
 
 const stdin = fs.readFileSync(0).toString('utf8');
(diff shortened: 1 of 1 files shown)


hasezoey (approved):

vkarpov15 (approved):
Nice, thanks :+1:

PR #15889, opened 2025-12-13
types(schema): avoid treating paths with default: null as required

### LINKED ISSUE #15878: TypeScript: Optional fields with `default: null` incorrectly remove `| null` from inferred type
### Prerequisites

### Mongoose version

9.0.1

### Node.js version

24.11.0

### MongoDB server version

n/a

_No response_

### Description

When an optional field has `default: null` in the schema definition, the TypeScript type inference removes `| null` from the field's type. This creates a type mismatch where the field can actually be `null` at runtime, but TypeScript thinks it cannot be.

This appears to be caused by the `IsPathRequired` type logic treating any field with a non-undefined default as "required", which prevents `| null` from being added to the type, even when the default value itself is `null`.

---

ALSO, as a side note, I think that users should be allowed to configure whether they want inferred types to include `|  null` for ANY type (even optional ones) unless explicitly included. Maybe an option inside `interface DefaultSchemaOptions`?

My rational is that when users [create separate interfaces](https://mongoosejs.com/docs/typescript/schemas.html#separate-document-interface-definition) and use them to define their schema instead of using [automatic type inference](https://mongoosejs.com/docs/typescript/schemas.html#automatic-type-inference), I very rarely see people include `| null` for EVERY OPTIONAL TYPE. Despite it being possible for `null` values to be stored, most users do not want to deal with it in their types. I guess this relates to #14421. Let me know if I should make a separate issue for this thought.

### Steps to Reproduce

```typescript
  name: { type: String, default: null }
});

type RawUser = InferRawDocTypeFromSchema<typeof UserSchema>;
// type RawUser = {
//     name: string;
// }

type HydratedUser = InferHydratedDocTypeFromSchema<typeof UserSchema>;
// type HydratedUser = mongoose.Document<unknown, {}, {
//     name: string;
// }, {
//     id: string;
// }, mongoose.ResolveSchemaOptions<mongoose.DefaultSchemaOptions>> & Omit<{
//     user: string;
// } & {
//     _id: mongoose.Types.ObjectId;
// } & {
//     __v: number;
// }, "id"> & {
//     id: string;
// }

const UserModel = mongoose.model('User', UserSchema);

const user = new UserModel({});
console.log(user.name)
// null
```

### Expected Behavior

The inferred types should include `| null`.

```typescript
type RawUser = {
    name: string | null;
}
```

vkarpov15:
Fix #15878

**Summary**

Currently, defining a path as `{ type: String, default: undefined }` will correctly treat the path as not required in TypeScript... but `{ type: String, default: null }` will treat the path as required, which is incorrect.

**Examples**

### DIFF
--- a/scripts/tsc-diagnostics-check.js
+++ b/scripts/tsc-diagnostics-check.js
@@ -3,7 +3,7 @@
 const fs = require('fs');
 
 const stdin = fs.readFileSync(0).toString('utf8');
-const maxInstantiations = isNaN(process.argv[2]) ? 300000 : parseInt(process.argv[2], 10);
+const maxInstantiations = isNaN(process.argv[2]) ? 350000 : parseInt(process.argv[2], 10);
 
 console.log(stdin);
 
--- a/types/inferrawdoctype.d.ts
+++ b/types/inferrawdoctype.d.ts
@@ -97,7 +97,7 @@ declare module 'mongoose' {
       ResolveRawPathType<Options['of'] extends ReadonlyArray<infer Item> ? Item : never>
     : PathValueType extends ArrayConstructor ? any[]
     : PathValueType extends typeof Schema.Types.Mixed ? any
-    : IfEquals<PathValueType, ObjectConstructor> extends true ? any
+    : PathValueType extends ObjectConstructor ? any
     : IfEquals<PathValueType, {}> extends true ? any
     : PathValueType extends typeof SchemaType ? PathValueType['prototype']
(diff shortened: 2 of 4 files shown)


AbdelrahmanHafez (approved):

vkarpov15:
I was able to make some small improvements to reduce number of typescript instantiations, but not enough to get back down to 300k. Which is fine because the ts benchmark is failing on 8.x right now with 338k instantiations, and this PR knocks that down to 334k. Small improvement, and I increased the instantiations limit to make the benchmark pass.

PR #15893, opened 2025-12-15
fix(document): use bitwise OR to accumulate version mode flags

### LINKED ISSUE #15888: [Bug] Version mode can be downgraded when modifying multiple paths
### Prerequisites

### Mongoose version

9.0.

### Node.js version

N/A

### MongoDB server version

N/A

N/A

### Description

When modifying multiple paths in a single save, the version mode (`this.$__.version`) can be incorrectly downgraded. This happens because the versioning logic uses direct assignment instead of accumulating with bitwise OR.

For example, if you:
1. Push to a top-level array (should set VERSION_INC = 2)
2. Set a value on an array index path (should set VERSION_WHERE = 1)

The second operation overwrites the first, resulting in VERSION_WHERE (1) instead of VERSION_ALL (3). This means `__v` is added to the query filter but NOT incremented, even though we pushed to an array.

### Steps to Reproduce

```javascript

run();

async function run() {

  const Model = mongoose.model('Test', new mongoose.Schema({
    items: [{ name: String }],
    tags: [String]
  }));

  const doc = await Model.create({
    items: [{ name: 'item1' }, { name: 'item2' }, { name: 'item3' }],
    tags: ['tag']
  });

  // Two clients load the same document
  const clientA = await Model.findById(doc._id);
  const clientB = await Model.findById(doc._id);

  // Client A modifies the document
  clientA.items.pull(clientA.items[0]._id); // pull (VERSION_INC)
  clientA.tags[0] = 'modified'; // set index (VERSION_WHERE overwrites it)
  await clientA.save();
  // items is now ['item2', 'item3'], __v should be 1

  const afterA = await Model.findById(doc._id);
  assert.strictEqual(afterA.__v, 1, '__v should be incremented after pull');

  // Client B has stale view, thinks items[1] is "item2"
  clientB.items[1].name = 'item2-updated';
  const err = await clientB.save().then(() => null, err => err);
  assert.ok(err, 'save should throw VersionError due to stale __v');
  assert.strictEqual(err.name, 'VersionError');
}
```

### Expected Behavior

When modifying multiple paths that trigger different version modes, the modes should be combined using bitwise OR:
- VERSION_WHERE (1) + VERSION_INC (2) = VERSION_ALL (3)

This ensures that:
1. `__v` is included in the query filter (for index safety)
2. `__v` is incremented (to signal array position changes to other clients)

AbdelrahmanHafez:
Fixes #15888

**The problem:** When you modify multiple array paths in a single save, the version mode can get overwritten instead of accumulated. For example:

```js
doc.items.pull(doc.items[0]._id);  // sets VERSION_INC
doc.tags[0] = 'modified';          // overwrites with VERSION_WHERE
```

The second operation overwrites the first, so `__v` gets added to the query filter but never incremented. This breaks optimistic locking since concurrent clients won't see the version bump from the pull.

**The fix:** Changed two assignments from `=` to `|=` so version flags accumulate properly:

```diff
- self.$__.version = VERSION_INC;
+ self.$__.version |= VERSION_INC;

- self.$__.version = VERSION_WHERE;
+ self.$__.version |= VERSION_WHERE;
```

Now `VERSION_INC (2) | VERSION_WHERE (1) = VERSION_ALL (3)`, which means both the increment and the query filter happen as expected.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -5174,7 +5174,7 @@ function operand(self, where, delta, data, val, op) {
     if (/\.\d+\.|\.\d+$/.test(data.path)) {
       self.$__.version = VERSION_ALL;
     } else {
-      self.$__.version = VERSION_INC;
+      self.$__.version |= VERSION_INC;
     }
   } else if (/^\$p/.test(op)) {
     // potentially changing array positions
@@ -5185,7 +5185,7 @@ function operand(self, where, delta, data, val, op) {
   } else if (/\.\d+\.|\.\d+$/.test(data.path)) {
     // now handling $set, $unset
     // subpath of array
-    self.$__.version = VERSION_WHERE;
+    self.$__.version |= VERSION_WHERE;
   }
 }
 


.gitignore
```diff
@@ -76,3 +76,4 @@ fle-cluster-config.json
 
 # @ark/attest (type testing)
 .attest
+.worktrees/
```
hasezoey, line 79: What program creates that?

PR #15897, opened 2025-12-15
test: add util.isDeepStrictEqual coverage alongside lodash.isequal

shash-hq:
Thanks for the feedback!

I've updated the PR to retain `lodash.isequal` for backward compatibility testing. I added `util.isDeepStrictEqual` as an additional test case to ensure we cover both the legacy usage and the native Node.js method. 

We can look into fully dropping the dependency in a future major release.

### DIFF
--- a/test/model.findOneAndUpdate.test.js
+++ b/test/model.findOneAndUpdate.test.js
@@ -20,16 +20,16 @@ const isEqualWith = require('lodash.isequalwith');
 const util = require('./util');
 const uuid = require('uuid');
 
-describe('model: findOneAndUpdate:', function () {
+describe('model: findOneAndUpdate:', function() {
   let Comments;
   let BlogPost;
   let db;
 
-  before(function () {
+  before(function() {
     db = start();
   });
 
-  after(async function () {
+  after(async function() {
     await db.close();
   });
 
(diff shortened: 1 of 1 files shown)


vkarpov15:
No, the whole point is to test integration with `lodash.isequal()`, removing it is potentially backwards breaking, although something we should consider for a future release.

Instead, can you please add separate tests covering util.isDeepStrictEqual() and retain the lodash.isequal tests?

shash-hq:
makes sense @vkarpov15
I failed to consider that we want to explicitly test compatibility with lodash.isequal. I've updated the PR to keep the existing tests and added util.isDeepStrictEqual as additional test coverage.

vkarpov15 (changes requested):
There's a bunch of unrelated formatting changes in this PR, can you please undo them? It looks like your editor added spaces between `function` and `()`

shash-hq:
Sorry about that @vkarpov15
I will revert those formatting changes now and push an update so the diff only shows the relevant test additions.

shash-hq:
@vkarpov15 Done. I've gone through and reverted those formatting changes to match the project's style. The diff should be much cleaner now. Let me know if everything looks good on your end. 
Thanks

PR #15898, opened 2025-12-15
fix(model): ensure $isDeleted is set after calling `doc.deleteOne()` successfully

### LINKED ISSUE #15858: The $isDeleted is not set automatically as the documentation states.

vkarpov15:
Fix #15858

**Summary**

Using `query.transform()` instead of `query.post()` because Mongoose doesn't pass the result to `query.post()`, and adding the result to `query.post()` would be backwards breaking for cb-based post hooks.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -830,7 +830,7 @@ Model.prototype.deleteOne = function deleteOne(options) {
     self.constructor._middleware.execPost('deleteOne', self, [self], {}, cb);
   });
   query.transform(function setIsDeleted(result) {
-    if (result && result.deletedCount != null && result.deletedCount > 0) {
+    if (result?.deletedCount > 0) {
       self.$isDeleted(true);
     }
     return result;


hasezoey (commented):
LGTM, aside from what copilot brought up about the issue id in test name

lib/model.js
```diff
@@ -829,6 +829,12 @@ Model.prototype.deleteOne = function deleteOne(options) {
   query.post(function queryPostDeleteOne(cb) {
     self.constructor._middleware.execPost('deleteOne', self, [self], {}, cb);
   });
+  query.transform(function setIsDeleted(result) {
+    if (result && result.deletedCount != null && result.deletedCount > 0) {
```
AbdelrahmanHafez: Nitpicky suggestion:
```suggestion
    if (result?.deletedCount > 0) {
```

Take it with a grain of salt for now, will be looking more into the benchmarks of optional chaining later on my PR 👍

Neha-Sharmaaa:
Implemented in #16137

`query.post()` middleware now receives the query result as the first parameter.

**Usage:**
```js
query.post(function(result) {
  if (result.deletedCount > 0) {
    // Handle successful delete
  }
});
```

**Supported signatures:**
- `query.post(function() {})` - backward compatible, no params
- `query.post(function(result) {})` - receives result
- `query.post(function(result, next) {})` - receives result + next for async

Works with find(), findOne(), updateOne(), deleteOne(), and other query operations. Fully backward compatible.

PR #15904, opened 2025-12-17
fix(document): when cloning a doc with subdocs, make sure the subdocs parent is the cloned doc

### LINKED ISSUE #15901: [Bug]: `$clone()` subdocuments have wrong `$parent()` reference
### Prerequisites

### Mongoose version

9.0.1

### Node.js version

N/A

### MongoDB server version

N/A

N/A

### Description

When calling `$clone()` on a document with subdocument arrays, the cloned subdocuments' `$parent()` method returns the **original** parent document instead of the **cloned** parent document.

### Steps to Reproduce

```javascript
'use strict';
const { Schema } = mongoose;

run();

function run() {
  const addressSchema = new Schema({
    street: String,
    city: String
  });

    name: String,
    addresses: [addressSchema]
  });

  const user = new User({
    name: 'Test User',
    addresses: [{ street: '123 Main St', city: 'Boston' }]
  });

  const clonedUser = user.$clone();

  assert.strictEqual(
    clonedUser.addresses[0].$parent(),
    clonedUser,
    'cloned subdocument $parent() should return cloned document'
  );
  console.log('All tests passed');
}
```

### Expected Behavior

After `$clone()`, the cloned subdocuments' `$parent()` should return the cloned parent document, not the original parent document.

```javascript
const clonedUser = user.$clone();
clonedUser.addresses[0].$parent() === clonedUser; // should be true
clonedUser.addresses[0].$parent() === user; // should be false
```

vkarpov15:
Fix #15901

**Summary**

Good find @AbdelrahmanHafez  :+1:

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -12353,6 +12353,60 @@ describe('document', function() {
     assert.deepEqual(doc.arr, [99]);
   });
 
+  it('updates subdocument parents when cloning (gh-15901)', async function() {
+    const addressSchema = new Schema({
+      street: String,
+      city: String,
+      image: new Schema({ url: String })
+    });
+
+    const userSchema = new Schema({
+      name: String,
+      addresses: [addressSchema],
+      bestFriend: { type: Schema.Types.ObjectId, ref: 'User' }
+    });
+
+    const User = db.model('User', userSchema);
+
+    const bestFriend = new User({
+      name: 'Best Friend'
+    });
+
+    const user = new User({
+      name: 'Test User',
+      addresses: [{ street: '123 Main St', city: 'Boston', image: { url: 'google.com' } }],
+      bestFriend: bestFriend._id
+    });
+
+    user.bestFriend = bestFriend;
+
+    assert.ok(user.populated('bestFriend'));
+    assert.ok(user.bestFriend instanceof User);
+
+    const clonedUser = user.$clone();
+
+    // Check cloned subdoc parent pointers
+    assert.ok(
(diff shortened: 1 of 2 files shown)


test/document.test.js
```diff
@@ -12348,11 +12348,65 @@ describe('document', function() {
     assert.ok(clonedDoc.arr.isMongooseArray);
     assert.ok(!clonedDoc.arr.isMongooseDocumentArray);
 
-    assert.deepEqual(doc.subdocArray[0], clonedDoc.subdocArray[0]);
-    assert.deepEqual(doc.subdoc, clonedDoc.subdoc);
+    assert.deepEqual(doc.subdocArray[0].toObject(), clonedDoc.subdocArray[0].toObject());
+    assert.deepEqual(doc.subdoc.toObject(), clonedDoc.subdoc.toObject());
```
AbdelrahmanHafez, line 12352: Is this change necessary? Does this fix break current behavior somehow?

AbdelrahmanHafez (commented):
One minor comment, otherwise LGTM 👍

hasezoey (commented):
LGTM, agree with previous comment from Hafez.

test/document.test.js
vkarpov15: This change is necessary because this `deepEqual()` compared parent documents :) That didn't process when we first wrote this test, but in hindsight this test asserted on incorrect behavior.

Issue #15894, opened 2025-12-15
Feature Proposal: Schema-level Auto-lean option


aclarembeau:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Hello, I'd like to suggest a new feature: "Add a schema-level lean option that automatically applies .lean() to all query operations for that schema". 

When set to true, all queries would return plain JavaScript objects by default instead of Mongoose documents, eliminating the need to manually call .lean() on every query.

### Motivation

**Performance Benefits**

Many applications have schemas that represent read-heavy data where Mongoose's document hydration overhead is unnecessary. 
Currently, developers must remember to append .lean() to every query:

```
await User.findById(id).lean();
await User.find({ active: true }).lean();
await User.findOne({ email }).lean();
```

This becomes particularly burdensome in codebases with hundreds of queries, leading to:

- Inconsistent performance: Forgetting .lean() on a single query causes unexpected overhead
- Code verbosity: Every query requires an extra method call
- Maintenance burden: Refactoring requires updating each query individually

**Use Cases**

- Read-only data models: Logs, analytics, historical records
- API responses: Data that's immediately serialized to JSON
- High-throughput services: Microservices where every millisecond counts
- Aggregation pipelines: Data that goes through significant transformation
- Cache layers: Objects stored in Redis or similar

### Example

**Proposed API**
```

const userSchema = new mongoose.Schema({
  name: String,
  email: String,
  role: String,
  createdAt: Date
}, {
  lean: true  // Enable auto-lean for all queries
});

const User = mongoose.model('User', userSchema);
```

Behavior

```
// All queries automatically return plain objects
const user = await User.findById(id);
// user is a plain object, not a Mongoose document

const users = await User.find({ role: 'admin' });
// users is an array of plain objects

// Override when needed - explicit opt-out
const fullDoc = await User.findById(id).lean(false);
// fullDoc is a Mongoose document with all methods
```

Query Types Affected

- find(), findOne(), findById()
- findOneAndUpdate(), findOneAndDelete(), etc. (when returning documents)
- count(), estimatedDocumentCount() (no-op, already return primitives)

Backwards Compatibility

- Default remains false to maintain current behavior
- Explicit .lean(false) overrides schema setting
- Chaining .lean() without arguments remains equivalent to .lean(true)

vkarpov15:
[...]

Issue #10090, opened 2021-04-01
Add lean schema option


kyrylkov:
To be able to configure: `schema.set('lean', { getters: true, virtuals: false });`

vkarpov15:
Fixed in #12081, will be release in 6.5.

Issue #15583, opened 2025-08-14
set lean transform at schema level


y-nk:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I have found in issues and PR that `.lean()` supports a `transform: (obj: any) => void` function which is used to mutate the object retrieved from db post lean. The problem being that we need to put this function in every `.lean()` call.

I think it'd be nicer if we could have a `leanTransform: (obj: any) => void` option right into the schema to avoid spreading the same transform function (or rather, set as base)

vkarpov15:
The recommended way to do this is via middleware, `schema.pre(['find', 'findOne', 'findOneAndUpdate', 'findOneAndReplace', 'findOneAndDelete'])`. Check for `lean === true`, and, if so, apply lean transform as follows:

```javascript
  mongoose.plugin(schema => schema.pre(['find', 'findOne', 'findOneAndUpdate', 'findOneAndReplace', 'findOneAndDelete'], function() {
    if (this.mongooseOptions().lean === true) {
      this.lean({
        transform: (doc) => {
          doc.transformed = true;
          return doc;
        }
      });
    }
  }));
```

The above will apply your transform for all schemas, but you can also apply that middleware to individual schemas. Does this help?

y-nk:
@vkarpov15 this does help indeed :) tho typewise it's a bit tricky and overall i feel like it's a bit weird that there are transform defaults for some methods (toObject/toJSON) but not all of them / at the same time i understand that the api could be clogged if too many options as well.

any opinion against this as options?

vkarpov15:
Yeah I think it makes sense to support default `lean` options in schema options, maybe something like the following?

```javascript
const schema = new Schema({ /* schema definition */ }, {
  lean: {
    transform: (doc) => {
          doc.transformed = true;
          return doc;
        }
  }
});
```

That way, if you apply `lean()` on a query, Mongoose would apply your `lean` options by default.

y-nk:
@vkarpov15 yes it could be done at schema level, it makes more sense indeed.

vkarpov15:
We already have a `lean` option for schemas that was added in #10090. #15646 will add this to the TypeScript types. Couple of caveats related to what we discussed:

1. Setting `new Schema(obj, { lean: { virtuals: true } })` will set `lean` by default on all queries
[...]

PR #15907, opened 2025-12-21
docs(schema): add lean to list of schema options

vkarpov15:
Re: #15894
Re: #10090
Re: #12081

**Summary**

`lean` isn't on our list of schema options on our docs 😿 

**Examples**


vkarpov15:
@AbdelrahmanHafez correct, see #15583 for how we added support for lean by default based on schema options.

PR #15908, opened 2025-12-21
feat(document+model): pass options to pre('deleteOne') and update+options to pre('updateOne') hooks

### LINKED ISSUE #15870: Difference in post-middleware arguments between .remove() and .deleteOne()

vkarpov15:
Fix #15870

**Summary**

Make it possible to access parameters from document middleware - currently not possible in Mongoose 8 or 9. Can't make this change for post middleware because that would be backwards breaking.

With this PR, you can do:

```javascript
schema.pre('deleteOne', { document: true, query: false }, function(doc, options) {
  // Check options
});
```

**Examples**

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -6850,7 +6850,7 @@ describe('document', function() {
     let docRegexCount = 0;
     let docPostRegexCount = 0;
 
-    schema.pre('updateOne', { document: false, query: true }, function () {
+    schema.pre('updateOne', { document: false, query: true }, function() {
       ++queryCount;
       assert.strictEqual(this.options.testOption, 'newValue');
     });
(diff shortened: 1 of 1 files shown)


hasezoey (commented):
LGTM

PR #15911, opened 2025-12-22
chore(ci): use min reporter for testing

AbdelrahmanHafez:
## Summary

- Switch mocha test reporter from `spec` to `min` in `.mocharc.yml`

## Motivation

The default `spec` reporter logs output for every test that runs. When tests fail, you need to scroll through ~5k lines to find what's actually failing. The per-test output for passing tests adds no value.

This is particularly problematic when running `npm test` from AI agents, where the verbose output pollutes the LLM context with unnecessary information.

With `min` reporter, we only see:
- Failing tests (with details)
- A summary of test status


vkarpov15 (commented):
Is there a way to add a separate script to use reporter: min? I personally prefer spec because I like seeing the test progress

AbdelrahmanHafez:
Sure, what do you have on mind? Should we make CI use the one with `spec` or `min` in that case?

vkarpov15:
I think spec is fine for CI. I'm not even sure we need a separate script because you can run `npm test -- --reporter=min`, or instruct your LLM to do so if you prefer, if the default reporter is causing you trouble.

My experience: I've never had trouble with Claude Code getting confused with too much output from npm test. The most common error I've seen is that Claude Code tends to assume that a script running for more than 1.5 minutes with no output is a script that succeeded, and using reporter=min would make that problem worse.

AbdelrahmanHafez:
Fair point re: local-testing.

It's not that Claude Code gets confused; it's that it uses more tokens than it needs to, making the auto-compact happen a lot faster. An AI agent would need to read 5k lines (75k tokens[1]) unnecessarily, that's more half of the context (Claude uses 45k tokens for  autocompact buffer, 17k for system tools, and other miscellaneous data loaded in context by default), this makes 1 `npm test` leave me with ~50k tokens in the context.

This actually affects my workflow, I would ask Claude about a failing GH run, which then would run a CLI command `gh` to get the logs, it'd be a lot more efficient to not output the successful tests and only show the failing ones.

I think it makes sense to use min as the default for CI, or at least `dot` if you'd like to see progress in real time. Those tokens are real expensive. 😄
As a workaround, I can manually copy the failing tests from the gh run and paste it in Claude Code, but it slows me down, and I really doubt anyone reads the entire output.
What do you think?

[1]https://claude-tokenizer.vercel.app, [ci run](https://github.com/Automattic/mongoose/actions/runs/20440290150/job/58731419332)

vkarpov15:
Yeah ok, using `min` as the reporter for CI tests can work. Would it be easy enough for you to tell Claude to use `npm run ci` instead of `npm test`?

As long as we keep spec as the default reporter for `npm test` I'm fine with it, I prefer spec for local dev to 1) make sure tests are running and not hanging, 2) check that the tests I expected to run actually ran because most of the time I'm testing with `-g`. FWIW I typically instruct Claude Code to run only specific tests with `-g` so Claude doesn't have to wait for the full test suite to run.

AbdelrahmanHafez:
I agree that `spec` is sensible for local development, and using `--grep` makes sense for Claude Code locally.

I'll revert the default back to `spec` and add a `test:ci` script instead. Using `ci` as the script name would be too easy to confuse with `npm ci`.

AbdelrahmanHafez:
A CI test output now went down from 75k tokens to 1k tokens on success.

vkarpov15 (approved):
Thanks :+1:

Issue #15913, opened 2025-12-23
[Bug] Enum casting broken for numeric typescript enums


mjfwebb:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

9.0.1

### Node.js version

x

### MongoDB server version

x

### Typescript version (if applicable)

5.9.3

### Description

Works in 8x, and 9.0.0.

Fairly sure this is related to https://github.com/Automattic/mongoose/pull/15824

Given this setup:

```ts
enum MyEnum {
  Hello = 0,
  GoodBye,
}

myEnumProperty: { type: Number, enum: MyEnum, required: true, default: 0 },
```

Getting an error like this:

> Cast to Number failed for value "Hello" (type string) at path "myEnumProperty"

### Steps to Reproduce

Schema with something like this:

```ts
import mongoose, { Schema } from 'mongoose';

enum MyEnum {
  Hello = 0,
  GoodBye,
}

const schema = new Schema<User>({
  name: { type: String, required: true },
  email: { type: String, required: true },
  myEnumProperty: { type: Number, enum: MyEnum, required: true, default: 0 },
  avatar: String,
});

interface User {
  name: string;
  email: string;
  avatar?: string;
}

const UserModel = mongoose.model<User>('User', schema);

const doc = new UserModel({ name: 'test', email: 'test', myEnumProperty: MyEnum.GoodBye });
```

### Expected Behavior

No errors

mjfwebb:
The issue is here:

https://github.com/Automattic/mongoose/blob/4c133006be54c4f12ca817c310b60a2332c73055/lib/schema/number.js#L337-L339

PR #15914, opened 2025-12-23
fix: add support for typescript style enums

mjfwebb:
Re: #15913


vkarpov15:
Squashing and merging so I can more easily backport to 8.x.

Issue #15915, opened 2025-12-24
[Bug] `optimisticConcurrencyControl: { exclude: boolean}` is documented but not implemented


AbdelrahmanHafez:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

re https://github.com/Automattic/mongoose/issues/10591#issuecomment-3690238711

AbdelrahmanHafez:
@vkarpov15 How was `exclude` intended to behave? Does it exclude some fields and enable OCC for the rest of the schema?

vkarpov15:
Yes exclude some fields, but include the rest. The use case is you want OCC but you want to disable 1-2 potentially problematic fields.

PR #15916, opened 2025-12-24
fix(versioning): make `optimisticConcurrency: string[]` add more fields to default behavior instead of replacing default

### LINKED ISSUE #15912: [Docs] `optimisticConcurrency` with array of fields disables default array versioning
### Prerequisites

### Mongoose version

9.0.2

### Node.js version

N/A

### MongoDB server version

N/A

N/A

## Current Behavior

When you set `optimisticConcurrency: ['balance']` on a schema, Mongoose will only bump `__v` when the `balance` field is modified. This completely replaces the default array versioning behavior.

```javascript
const schema = new Schema({
  balance: Number,
  tags: [String]
}, { optimisticConcurrency: ['balance'] });
```

With this config:
- Modifying `balance` bumps `__v`
- Modifying `tags` (an array) does NOT bump `__v`

### Steps to Reproduce

```javascript
const schema = new Schema({
  balance: Number,
  items: [{ name: String }]
}, { optimisticConcurrency: ['balance'] });

// Two requests load the same document
const docA = await Model.findById(id);  // items: ['a', 'b'], __v: 0
const docB = await Model.findById(id);  // items: ['a', 'b'], __v: 0

// Both modify the array
docA.items.pull('a');      // wants to remove index 0
docB.items.set(1, 'x');    // wants to update index 1

await docA.save();  // succeeds, items: ['b'], __v still 0
await docB.save();  // ALSO succeeds, overwrites with wrong data
```

## Why This Is a Problem

The default array versioning exists to prevent data corruption from concurrent array modifications. 

Without array versioning, docB's update targeting index 1 succeeds even though that index no longer exists after docA's change. The user thought they were protecting `balance` with optimistic concurrency, but accidentally removed protection from arrays.

### Expected Behavior

`optimisticConcurrency: ['field']` should add to the default array versioning, not replace it. Users opting into optimistic concurrency for specific fields probably don't expect to lose the safety features they had before.

AbdelrahmanHafez:
## Summary

Fixes #15912 

When `optimisticConcurrency` is set to an array of fields (e.g., `['balance']`), the default array versioning was being completely bypassed.

The array form of `optimisticConcurrency` should *add* versioning for specific fields, not *replace* the default array versioning. Users opting into optimistic concurrency for specific fields probably don't expect to lose the safety features they had before.

### Before (bug)
- `optimisticConcurrency: ['balance']` + modify `balance` → `__v` increments
- `optimisticConcurrency: ['balance']` + modify array field → `__v` does NOT increment

### After (fix)
- `optimisticConcurrency: ['balance']` + modify `balance` → `__v` increments
- `optimisticConcurrency: ['balance']` + modify array field → `__v` increments

Also updated docs to clarify that the string array form adds optimistic concurrency "in addition to the default array versioning".

I also noticed that `{ exclude: string[] }` was documented and has TS types but isn't implemented in JS, created #15915 for it, removed the docs where possible for that option without breaking TS contract, and will fix it in a future PR.

### DIFF
--- a/test/versioning.test.js
+++ b/test/versioning.test.js
@@ -733,7 +733,7 @@ describe('versioning', function() {
     }
   });
 
-  describe('optimisticConcurrency with array of fields preserves default array versioning (gh-15809)', function() {
+  describe('optimisticConcurrency with array of fields preserves default array versioning (gh-15912)', function() {
     const VERSION_INC = mongoose.Document.VERSION_INC;
     const VERSION_WHERE = mongoose.Document.VERSION_WHERE;
     const VERSION_ALL = mongoose.Document.VERSION_ALL;


lib/document.js
```diff
@@ -4980,7 +4980,7 @@ Document.prototype.$__delta = function $__delta() {
     if (Array.isArray(optimisticConcurrency)) {
       const optCon = new Set(optimisticConcurrency);
       const modPaths = this.modifiedPaths();
-      if (modPaths.find(path => optCon.has(path))) {
+      if (modPaths.some(path => optCon.has(path))) {
```
AbdelrahmanHafez, line 4983: Not strictly necessary change, but it took me a minute to understand this line, and `.some(...)` is more intention-revealing here.

PR #15944, opened 2026-01-05
feat(model): allow passing `strict` option to `hydrate()`

vkarpov15:
Fix #15940

**Summary**

#15940 suggested a `strip` option to `hydrate()`, but I think just adding `strict` as an option to `hydrate()`, that lines up with existing document strict mode semantics, would be more consistent with the existing API. WDYT @AbdelrahmanHafez ?

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -783,7 +783,7 @@ function init(self, obj, doc, opts, prefix) {
         self[i] = value;
       } else if (opts?.virtuals && (i in docSchema.virtuals)) {
         self[i] = value;
-      } else if (strict === 'throw' && opts?.strict === 'throw') {
+      } else if (opts?.strict === 'throw') {
         // Only use strict: 'throw' semantics if explicit `strict: 'throw'` option
         // passed in, like via `MyModel.hydrate(obj, null, { strict: 'throw' })`
         // This is for backwards compatibility - strict: 'throw' at the schema level


test/model.hydrate.test.js
```diff
@@ -318,5 +318,64 @@ describe('model', function() {
...
+
+      // Test with strict: 'throw' - should throw on extra fields
+      assert.throws(() => {
+        const doc3 = Model.hydrate({
+          _id: '000000000000000000000003',
+          name: 'Bob',
+          age: 35,
+          extraField: 'should throw'
+        }, null, { strict: 'throw' });
+        doc3.set('anotherExtraField', 'value');
```
AbdelrahmanHafez: Is it the Model.hydrate that throws here or the `doc3.set(...)`? Is this line needed? Or are these intended to be two different `assert.throws(` that are unintentionally merged in one?

test/model.hydrate.test.js
```diff
@@ -318,5 +318,64 @@ describe('model', function() {
 
       assert.ok(!hydrated.stories[1].article);
     });
+
+    it('handles strict option to control non-schema properties', function() {
```
AbdelrahmanHafez, line 322: Something about synchronous tests feels so satisfying!

AbdelrahmanHafez (commented):
One question other than Copilot's comments, otherwise LGTM.
Also, impressive speed on shipping this. 😁

test/types/models.test.ts
```diff
@@ -1121,3 +1121,49 @@ async function gh15910() {
...
+
+  // Test with strict: false
+  const doc1 = TestModel.hydrate({
+    _id: new mongoose.Types.ObjectId(),
+    name: 'John',
+    age: 30,
+    extraField: 'value'
+  }, undefined, { strict: false });
+
+  expectType<ReturnType<(typeof TestModel)['hydrate']>>(doc1);
```
AbdelrahmanHafez, line 1141: Does this implicitly assert that `extraField` is accessible in `doc1`?

lib/model.js
```diff
@@ -4004,7 +4004,7 @@ Model.hydrate = function(obj, projection, options) {
     }
     obj = applyProjection(obj, projection);
   }
-  const document = require('./queryHelpers').createModel(this, obj, projection);
+  const document = require('./queryHelpers').createModel(this, obj, projection, projection, options);
```
hasezoey, line 4008: This function should now also list `strict` in the JSDoc.

test/model.hydrate.test.js
vkarpov15: Fixed in [882a325](https://github.com/Automattic/mongoose/pull/15944/commits/882a325df306385c360221a203aadbb9d9f39b0e)

test/types/models.test.ts
vkarpov15: No it does not. As a matter of fact, this assertion asserts the opposite: `doc1.extraField` won't be defined in TypeScript. I presume that is a problem for the use case you described in #15940 right @AbdelrahmanHafez ?

AbdelrahmanHafez: It would be nice to have TS support for fields defined in rawObject but not in schema when `strict: false`.
I can imagine it might be challenging to achieve, if it is, we can open an issue and look into it later. 👍

vkarpov15: I tinkered with adding a type-widening override if `strict: false` is set, but adding the following adds a lot of type errors, presumably because we make use of `ReturnType<(typeof Model)['hydrate']>` to get the hydrated doc type.

```ts
hydrate<TResultType>(
      obj: any,
      projection: ProjectionType<TRawDocType> | null | undefined,
      options: HydrateOptions & { strict: false }
    ): TResultType;
```

vkarpov15: You can always just do `const doc = Model.hydrate(obj, null, { strict: false }) as HydratedDoc & { extraField: string }`. Not ideal DX of course, but that's at least a workaround.

AbdelrahmanHafez: Sounds good, I think it's fine to open an issue and park the type thing for now, I assume it's a broader issue with the `virtuals` option as well.

lib/document.js
```diff
@@ -781,6 +783,12 @@ function init(self, obj, doc, opts, prefix) {
         self[i] = value;
       } else if (opts?.virtuals && (i in docSchema.virtuals)) {
         self[i] = value;
+      } else if (strict === 'throw' && opts?.strict === 'throw') {
+        // Only use strict: 'throw' semantics if explicit `strict: 'throw'` option
+        // passed in, like via `MyModel.hydrate(obj, null, { strict: 'throw' })`
+        // This is for backwards compatibility - strict: 'throw' at the schema level
+        // does not apply to documents loaded from the db.
+        throw new StrictModeError(i);
```
hasezoey, line 791: Unless i missed something, this does not seem to be tested. (from what i can tell, only `schema.opts.strict = throw && opts == undefined` is tested)

vkarpov15: I confirmed this is tested by the following:

```javascript
      // Test with strict: 'throw' - should throw on extra fields
      assert.throws(() => {
        Model.hydrate({
          _id: '000000000000000000000003',
          name: 'Bob',
          age: 35,
          extraField: 'should throw'
        }, null, { strict: 'throw' });
      }, /Field `extraField` is not in schema/);
```

although admittedly it can be hard to spot because of the levels of indirection and the `strict && opts?.strict` call.

PR #15948, opened 2026-01-07
types(model): apply basic type casting to paths underneath subdocuments

### LINKED ISSUE #15947: 'create' incorrectly rejects string values for ObjectId fields in subdocuments
### Prerequisites

### Mongoose version

9.1.2

### Node.js version

24.8.0

### MongoDB server version

 8.0.17

5.9.2

### Description

As the title says, the 'create' function incorrectly rejects string values for ObjectId fields in nested subdocuments.

### Steps to Reproduce

```ts
import mongoose, { Schema } from 'mongoose';

import { type SchemaTypeWithId } from '../utils/types.js';

const nestedSchema = new Schema(
  {
    name: {
      type: String,
      required: true,
      trim: true,
    },
    subdocument: {
      type: {
        _id: {
          type: Schema.Types.ObjectId,
          required: true,
        },
        field1: {
          type: String,
          required: true,
          trim: true,
        },
        field2: {
          type: Number,
          required: false,
        },
      },
      required: true,
    },
  },
  { timestamps: true },
);

export type Nested = SchemaTypeWithId<typeof nestedSchema>;
export const NestedModel = mongoose.model('Nested', nestedSchema);

setTimeout(async () => {
  await NestedModel.create({
    name: 'test',
    subdocument: {
       // Long TS error that boils down to `Type 'string' is not assignable to type 'ObjectId'`
      _id: '6951265a11a2b0976013be20',
      // Replacing ^ with the following line fixes the problem
      // _id: new mongoose.Types.ObjectId('6951265a11a2b0976013be20'),
      field1: 'test',
      field2: 1,
    },
  });
}, 5000);
```

Note that when the code runs, it works. A document is created and `subdocument._id` is an ObjectId as expected.

### Expected Behavior

No TypeScript error, it should accept string values for subdocuments just as it accepts string values for the root document

vkarpov15:
Fix #15947

**Summary**

Quick fix to allow `TestModel.create({ subdoc: { _id: '0'.repeat(24) } })`

**Examples**

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -1168,19 +1168,19 @@ async function gh15947() {
       type: new Schema({
         _id: {
           type: Schema.Types.ObjectId,
-          required: true,
+          required: true
         },
         field1: {
           type: String,
           required: true,
-          trim: true,
+          trim: true
         },
         field2: {
           type: Number,
-          required: false,
-        },
+          required: false
+        }
       }),
-      required: true,
+      required: true
     },
     docArr: [{ _id: 'ObjectId' }]
   });


hasezoey (changes requested):
LGTM, but definitely needs to fix lint.

PR #15950, opened 2026-01-09
Refactor Cursor class and update type definitions

Nubicusit:
**Summary**

**Examples**

### DIFF
--- a/types/cursor.d.ts
+++ b/types/cursor.d.ts
@@ -1,8 +1,12 @@
 declare module 'mongoose' {
+  import { Readable } from 'stream';
 
-  import stream = require('stream');
-
-  type CursorFlag = 'tailable' | 'oplogReplay' | 'noCursorTimeout' | 'awaitData' | 'partial';
+  type CursorFlag =
+    | 'tailable'
+    | 'oplogReplay'
+    | 'noCursorTimeout'
+    | 'awaitData'
+    | 'partial';
 
   interface EachAsyncOptions {
     parallel?: number;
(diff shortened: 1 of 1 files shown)


types/cursor.d.ts
```diff
@@ -11,57 +15,75 @@ declare module 'mongoose' {
...
 
     /**
-     * Registers a transform function which subsequently maps documents retrieved
-     * via the streams interface or `.next()`
+     * Iterate documents in batches
      */
-    map<ResultType>(fn: (res: DocType) => ResultType): Cursor<ResultType, Options>;
+    eachAsync(
+      fn: (docs: DocType[], index: number) => unknown | Promise<unknown>,
+      options: EachAsyncOptions & { batchSize: number }
```
vkarpov15, line 69: batchSize override needs to come before the `options?: EachAsyncOptions` override, TypeScript checks overrides in order so this override will never be used.

types/cursor.d.ts
```diff
@@ -11,57 +15,75 @@ declare module 'mongoose' {
...
+   */
+  class Cursor<
+    DocType = unknown,
+    Options extends Record<string, any> = Record<string, never>,
+    NextResultType = DocType | null
+  > extends Readable {
+    /**
+     * Async iterator support
+     */
+    [Symbol.asyncIterator](): AsyncIterator<DocType>;
```
vkarpov15, line 29: [Symbol.asyncIterator] still needs to return a `Cursor`, otherwise compiling fails with `Type AsyncIterator<DocType, any, any> is missing the following properties from type AsyncIterator<any, undefined, any>: [Symbol.asyncIterator], [Symbol.asyncDispose] `

vkarpov15:
Closing since this PR is inactive and needs a bunch of work

PR #15951, opened 2026-01-09
Remove tsd in favor of tsc + test utilities

vkarpov15:
**Summary**

tsd is nice, but with the ts-go rewrite the future of tsd is uncertain. Either way, using TypeScript directly should be more accurate in terms of catching errors, and will hopefully allow us to test with multiple different TypeScript configs as well.

**Examples**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,35 @@
-9.1.2 / 2025-01-05
+8.21.1 / 2026-01-23
+===================
+* fix(clone): fix parent doc for map subdocuments and array subdocuments #15958 [AbdelrahmanHafez](https://github.com/AbdelrahmanHafez)
+* fix(document): when cloning a doc with subdocs, make sure the subdocs parent is the cloned doc #15904 #15901
+* fix: respect currentTime schema option in bulkWrite updates #15976 [sderrow](https://github.com/sderrow)
+* types(models): support Mongoose query casting in AnyBulkWriteOperation filter property #15910
+* types: add toBSON() to documents #15927
+
+9.1.5 / 2026-01-20
+==================
+* fix(map): validate map subdocument when loaded with init #15960 #15957 [AbdelrahmanHafez](https://github.com/AbdelrahmanHafez)
+* fix(discriminator): prevent indexes and callQueue duplication with shared nested schemas #15974 #15966 [AbdelrahmanHafez](https://github.com/AbdelrahmanHafez)
+* fix(subdocuments): do not pass parent path on init #15970 #15969 #15682 [AbdelrahmanHafez](https://github.com/AbdelrahmanHafez)
+* types(inferrawdoctype): correct handling for subdocs and doc arrays #15967 #13772
+* docs: improve grammar and clarity in TypeScript schema comments #15971 [harshsinghpujari](https://github.com/harshsinghpujari)
+
+9.1.4 / 2026-01-15
+==================
+ * fix: attach sessions to docs retrieved by cursor #15953 #15949 [mjfwalsh](https://github.com/mjfwalsh)
+ * fix(model): make hydrate() handle nested schema arrays #15964 #15956
+ * fix(clone): fix parent doc for map subdocuments and array subdocuments #15958 #15954 [AbdelrahmanHafez](https://github.com/AbdelrahmanHafez)
+ * fix: prevent crash when accessing nested paths on prototype #15962 #15961 [som14062005](https://github.com/som14062005)
+
+9.1.3 / 2026-01-09
+==================
+ * fix(model): support timestamps option to insertMany() as both boolean and QueryTimestampsConfig #15941 #15938
+ * fix(query): include preview of current and incoming update in error when merging normal update with pipeline #15939 #15928
+ * types(model): apply basic type casting to paths underneath subdocuments #15948 #15947
+ * types(utility): make WithLevel1NestedPaths correctly handle PopulatedDoc and other TypeScript unions with Document members #15942 #15923
+ * docs(schema): expose "DocumentArrayElement" #15590 [hasezoey](https://github.com/hasezoey)
+
+9.1.2 / 2026-01-05
 ==================
  * fix(subdocs): pass options to pre-save hooks for subdocs #15921 #15920 [AbdelrahmanHafez](https://github.com/AbdelrahmanHafez)
  * perf(model): select only _id when checking document existence during save() #15919 [AbdelrahmanHafez](https://github.com/AbdelrahmanHafez)
--- a/docs/typescript/schemas.md
+++ b/docs/typescript/schemas.md
@@ -68,9 +68,9 @@ If automatic type inference doesn't work for you, you can define a separate raw 
 ```typescript
 import { Schema } from 'mongoose';
 
-// Raw document interface. Contains the data type as it will be stored
-// in MongoDB. So you can ObjectId, Buffer, and other custom primitive data types.
-// But no Mongoose document arrays or subdocuments.
+// Raw document interface. Represents the data types as they are stored in MongoDB.
+// This allows the use of ObjectId, Buffer, and other custom primitive data types,
+// but excludes Mongoose document arrays and subdocuments.
 interface User {
   name: string;
   email: string;
--- a/docs/typescript/virtuals.md
+++ b/docs/typescript/virtuals.md
@@ -28,6 +28,9 @@ const schema = new Schema(
 );
 ```
 
+If you are using automatic schema inference, you should define virtuals using the `virtuals` option in the schema constructor as shown above.
+Mongoose will not automatically infer any virtuals you define using `Schema.prototype.virtual()`.
+
 Note that Mongoose does **not** include virtuals in the returned type from `InferSchemaType`.
 That is because `InferSchemaType` returns a value similar to the raw document interface, which represents the structure of the data stored in MongoDB.
 
@@ -83,6 +86,9 @@ schema.virtual('fullName').get(function() {
   return `${this.firstName} ${this.lastName}`;
 });
 ```
+
+If you explicitly define the `UserVirtuals` interface, you should define your virtuals using `schema.virtual()` as shown above.
+We recommend `schema.virtual()` over the `virtuals` option to the Schema constructor shown in the Automatically Inferred Types section above because the `virtuals` option won't allow you to access other virtuals on `this`.
 
 ## Override the Type of `this` in Your Virtual
 
--- a/lib/cursor/queryCursor.js
+++ b/lib/cursor/queryCursor.js
@@ -599,6 +599,9 @@ function _nextDoc(ctx, doc, pop, callback) {
     if (err != null) {
       return callback(err);
     }
+    if (options.session != null) {
+      doc.$session(options.session);
+    }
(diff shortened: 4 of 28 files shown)


vkarpov15:
cc @chriskrycho

test/types/lean.test.ts
```diff
@@ -43,7 +45,7 @@ async function gh11761() {
   {
     // make sure _id has been added to the type
     const { _id, ...thing1 } = (await ThingModel.create({ name: 'thing1' })).toObject();
-    expectType<Types.ObjectId>(_id);
+    Expect<Equal<typeof _id, Types.ObjectId>>();
```
AbdelrahmanHafez: Why are we not using `ExpectType` in this file?

test/types/middleware.test.ts
```diff
@@ -128,7 +129,7 @@ function gh11480(): void {
   const UserSchema = new Schema<IUserSchema>({ name: { type: String } });
 
   UserSchema.pre('save', function() {
-    expectNotType<any>(this);
+    ExpectAssignable<HydratedDocument<IUserSchema>>()(this);
```
AbdelrahmanHafez, line 132: Does `ExpectType` break here?
Edit: never mind, I misread this change.

test/types/PipelineStage.test.ts
```diff
@@ -1,6 +1,5 @@
 import { ObjectId } from 'mongodb';
 import type { PipelineStage } from 'mongoose';
-import { expectError, expectType } from 'tsd';
```
AbdelrahmanHafez, line 3: Interesting, I would've thought that ESLint would catch this. Maybe an opportunity to have ESLint work on our TS tests if it's not already.

test/types/helpers.ts
```diff
@@ -0,0 +1,14 @@
...
+export declare function Expect<T extends true>(): void;
+
+export declare const ExpectType: <Expected>() =>
+  <Actual>(value: Equal<Actual, Expected> extends true ? Actual : never) => void;
+
+export declare const ExpectAssignable: <Expected>() =>
+  <Actual extends Expected>(value: Actual) => void;
+
+export declare const ExpectNotType: <Expected>() =>
+  <Actual>(value: Equal<Actual, Expected> extends false ? Actual : never) => void;
```
hasezoey: Could this extra (empty) function invocation be somehow avoided?

Maybe make it more explicit like `expect-type` could provide via `expectType(TYPE).toEqualType<ACTUAL_TYPE>()`?

test/types/util/assertions.ts
hasezoey, line 1: Might it be worth moving this file into a directory like `utils`? Otherwise it might be confusing to also have things like `helpers.test.ts` alongside this.

test/types/helpers.ts
```diff
@@ -0,0 +1,14 @@
+export type Equal<A, B> =
+  (<T>() => T extends A ? 1 : 2) extends
+  (<T>() => T extends B ? 1 : 2) ? true : false;
+
+export declare function Expect<T extends true>(): void;
+
+export declare const ExpectType: <Expected>() =>
+  <Actual>(value: Equal<Actual, Expected> extends true ? Actual : never) => void;
```
hasezoey:
```suggestion
export declare const ExpectType: <Expected>(value: Expected) => void;
```

Wouldnt this definition make more sense? This would remove the extra function call added and give better errors.

Previously this would give for example:
```txt
Argument of type '_QueryFilter<{ [x: string]: any; [x: number]: any; }>' is not assignable to parameter of type 'never'.ts(2345)
```
(For `ExpectTypeOld<boolean | undefined>()({} as UpdateManyModel['filter']);`)

Whereas with this is gives now:
```txt
Argument of type '_QueryFilter<{ [x: string]: any; [x: number]: any; }>' is not assignable to parameter of type 'boolean | undefined'.ts(2345)
```
(For `ExpectTypeNew<boolean | undefined>({} as UpdateManyModel['filter']);`)

test/types/PipelineStage.test.ts
hasezoey: It does not as "unused imports" is bundled into rule `no-unused-vars` (or for typescript `@typescript-eslint/no-unused-vars`), this cannot be individually warned, without warning on all other unused variables (ie unused functions). At least not without adding another extra eslint plugin.

My recommendation to allow this: enable the rule again, but this would require to modify the tests by:
- calling all unused functions (simply `gh32312()`)
- moving all "global scope" tests to be in a function
- adding some kind of ignore pattern (like variables / args starting with `_`)

I am not sure which kind of approach is wanted here; personally i would be for a mix of "calling all unused functions" and "adding some kind of ignore pattern".

mrazauskas:
> Might it be worth to investigate switching to `tstyche` instead

Thanks for mentioning TSTyche! (I am its author.) Yes, TypeScript 7 support is in the plans. I will start working on this as soon as APIs will be available.

---

Let me add some doubts regarding the following statement:

> using TypeScript directly should be more accurate in terms of catching errors

In my opinion, `// @ts-expect-error` does a good job in code that will be executed, but not in type tests. The problem of `// @ts-expect-error`: it hides errors of missing imports, duplicate identifier, renamed APIs and anything else. These errors are introduced during refactoring and `tsc` cannot catch them. (A sad fact, I have seen a repository where removed APIs are still passing tests marked with `// @ts-expect-error`.)

In contrary, `tsd`s `expectError()` is only accepting a predefined [list of errors](https://github.com/mcollina/tsd/blob/c46b372130ecad265b1e1e807578e45f912046bf/source/lib/compiler.ts#L17). Although on surface it might look similar to `// @ts-expect-error`.

TSTyche does even more. It is able to [check errors](https://tstyche.org/reference/config-file#check-suppressed-errors) suppressed by `// @ts-expect-error` (try it on [StackBlitz](https://stackblitz.com/edit/tstyche-suppressed-errors?file=examples%2Fexpect-error.tst.ts)).

There are many other differences between compiler (`tsc`) and type test runners. You know the requirements of your project. So use what suits you best.

chriskrycho:
Just saw this! Will try to give it a gander in the next couple of days—thanks for the ping, and hooray for doing the work!

test/types/PipelineStage.test.ts
vkarpov15: I would be ok with leaving this as-is. Extraneous imports in TypeScript tests is not likely to cause any user-facing issues. So while it would be nice to fix, I also don't think it is worthwhile to make TypeScript tests fail if there's an extraneous import.

test/types/lean.test.ts
vkarpov15: Likely because I added ExpectType after I started working on this file. Fixed in [d567931](https://github.com/Automattic/mongoose/pull/15951/commits/d567931d36c5cb8ef36a280d47b5f5e4566cbb3c)

test/types/helpers.ts
vkarpov15: I just removed this function since we don't use ExpectNotType<>. It is better to use `ExpectType<>` instead because ExpectNotType is not very helpful :)

test/types/helpers.ts
vkarpov15: I applied in [d567931](https://github.com/Automattic/mongoose/pull/15951/commits/d567931d36c5cb8ef36a280d47b5f5e4566cbb3c)

test/types/util/assertions.ts
vkarpov15: Good call fixed in [6235ce7](https://github.com/Automattic/mongoose/pull/15951/commits/6235ce7b6b670375729e3c4c521772ed3d3a2182)

test/types/helpers.ts
hasezoey: This comment was not specifically about `ExpectNotType`, but all 3 shown functions to require a extra function invocation (`Expect*<E>()(TYPE)`).

EDIT: ~~nevermind~~, due to https://github.com/Automattic/mongoose/pull/15951#discussion_r2694879830, this now only applies to `ExpectAssignable`

vkarpov15:
@mrazauskas I tried out tstyche, it is very cool. I like how you can assert on the error message with `// @ts-expect-error`.

Couple of questions:

1. How does tstyche work under the hood? From what I've been able to gather, it spawns tsc processes directly to compile individual test files and check the output, is that accurate?
2. How do I pass in test files to run? It seems to pick up tests I put in `typetests` dir just fine, but without a typetests then `./node_modules/.bin/tstyche ./test/types/*.test.ts` fails with `Error: No test files were selected using current configuration.`
3. Can you bring your own assertions? `expect<x>().toBe()` is not my favorite syntax. Minor point, I'm more just curious.

mrazauskas:
@vkarpov15 Thanks! Glad to hear that.

> 1. How does tstyche work under the hood? From what I've been able to gather, it spawns tsc processes directly to compile individual test files and check the output, is that accurate?

TSTyche uses TypeScript programmatically via language service API. This is the entry point used by IDEs and it is optimised to open / check a single file. Also the `references` of a project get handled. It works at any scale. In contrary, running `tsc --noEmit` does not work with `references` and always checks all files although lots of that job can be skipped.

`.toBe()`, `.toBeAssignableTo()` and `.toBeAssignableFrom()` compare types programmatically. Same as in `tsd`, but without patching TypeScript.

`.toBeCallableWith()`, `.toBeConstructableWith()`, `.toHaveProperty()` and other ability matchers are implemented via deleted syntax. A file gets open and checked for errors, if there are none, a virtual files gets open with deleted syntax and are errors get mapped to matchers. Here is how deleted syntax roughly looks like:

```ts
expect(getPerson).type.not.toBeCallableWith({ name: false })
;     (getPerson)                          ({ name: false })         
```

Having an error means the expression is not callable. (The matcher also makes sure that the expression has call signatures.)

Same with `// @ts-expect-error`. Errors are checked in a virtual file.

As you noticed, there are no generics in the matchers by design. Having any generics means the limitations of conditional types become limitations of the library.

> 2. How do I pass in test files to run? It seems to pick up tests I put in `typetests` dir just fine, but without a typetests then `./node_modules/.bin/tstyche ./test/types/*.test.ts` fails with `Error: No test files were selected using current configuration.`

You can configure this via the [`testFileMatch`](https://tstyche.org/reference/config-file#test-file-match) option. Every file matched by the patterns is a test file. You can also run a single file: `tstyche query-params`.

By the way, `test()` and `describe()` helpers are provided. They are optional. `.skip`, `.only` can be used with the helpers as well as with assertions: `expect.skip`, `expect.only`.

> 3. Can you bring your own assertions? `expect<x>().toBe()` is not my favorite syntax. Minor point, I'm more just curious.

In case if it feels better, you can use TSTyche as a runner and bring in your own generic based assertions. That works. `// @ts-expect-error` are checked, you can run a single file or test against several versions of TypeScript.

I went with `expect()` style, because it allows passing expression or type on both sides of assertion. This means `{} as unknown as SomeType` casting is never needed.

At some point I plan to add `assertType()` (https://github.com/tstyche/tstyche/issues/467). Would that look better? I need help / feedback regarding the syntax. The classical generic `assertType<Expected>(actual)` is somewhat odd, because expected goes before actual. Perhaps I am too conservative (; If it sounds interesting, lets talk in the issue.

vkarpov15:
@hasezoey that's my bad, I saw the "re-request review" icon on "Reviewers" so I just re-requested without reading your review comment, sorry

image

PR #15962, opened 2026-01-12
fix: prevent crash when accessing nested paths on prototype (#15961)

### LINKED ISSUE #15961: [Bug] Unsafe prototype getter generated for ALL nested paths (breaks Jest/Introspection)
### Prerequisites

### Mongoose version

9.1.2

### Node.js version

22.17.1

### MongoDB server version

6.21.0

_No response_

### Description

In Mongoose 9.x, defining **any** nested path in a Schema generates a getter on `Model.prototype` that is unsafe for static introspection.

When tools like `Jest` (specifically `jest.mock` or `automock`) inspect the Model class, they traverse the prototype chain. The generated getter attempts to access `this.$__.getters`. However, when accessing the property directly on the prototype (e.g., `Model.prototype.myField`), `this` refers to the Prototype object, not a Document instance. Consequently, `this.$__` is undefined, causing a crash.

### Steps to Reproduce

Run the following standalone script (no external dependencies required):

```javascript

// 1. Define ANY schema with a nested path
  randomWord: { nested: String }
});

console.log('Checking prototype...');

try {
  // 2. Accessing the property on the prototype triggers the crash.
  // This simulates what `jest.mock` does when inspecting modules.
  const val = Model.prototype.randomWord;
} catch (err) {
  console.error('CRASH REPRODUCED:');
  console.error(err.message); 
  console.error(err.stack);
}
```

**Actual Output**
```text
CRASH REPRODUCED:
Cannot read properties of undefined (reading 'getters')
TypeError: Cannot read properties of undefined (reading 'getters')
    at Model.get [as randomWord] (node_modules/mongoose/lib/helpers/document/compile.js:80:23)
```

### Expected Behavior

**Additional Context**
This breaks generic unit tests that rely on `jest.mock('./my.model.js')`, as Jest iterates over exports and touches their prototype properties, triggering this crash immediately upon test suite load.

som14062005:
Fixes #15961

## Summary
Prevents crashes when tools like Jest or automock inspect Model prototypes by accessing nested path properties.

## The Problem
When `jest.mock()` inspects a Mongoose Model, it accesses `Model.prototype.myField`. The getter tried to access `this.$__.getters`, but `this.$__` was undefined on the prototype, causing:
TypeError: Cannot read properties of undefined (reading 'getters')

## The Solution
Added a guard clause to return `undefined` when `this.$__` doesn't exist (prototype access).

## Changes
- Added safety check in `lib/helpers/document/compile.js` (3 lines)
- Added regression test in `test/document.test.js`

## Testing
- All existing tests pass
- New test validates the fix
- Backward compatible

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -328,7 +328,6 @@ describe('document', function() {
         nested: { field: String }
       });
       const TestModel = mongoose.model('gh15961', schema);
-      
       // Should not throw when Jest or other tools inspect the prototype
       assert.doesNotThrow(() => {
         const val = TestModel.prototype.nested;


som14062005:
Thank you for the review and for merging the fix!

PR #15964, opened 2026-01-14
fix(model): make hydrate() handle nested schema arrays

### LINKED ISSUE #15956: [Bug] hydrate() with virtuals: true does not set virtuals on nested array subdocuments
### Prerequisites

### Mongoose version

9.1.3

### Node.js version

N/A

### MongoDB server version

N/A

N/A

### Description

re #15627 #15638 
When using `Model.hydrate(obj, null, { virtuals: true })` with a schema that has nested arrays (array of document arrays, e.g. `[[subdocSchema]]`), virtuals on the innermost subdocuments are not set from the raw data.

The `virtuals: true` option works correctly for:
- Top-level virtuals
- Single-level document arrays
- Single nested subdocuments

But fails for nested arrays like `[[innerSchema]]`.

### Steps to Reproduce

```javascript

const innerSchema = new mongoose.Schema({ name: String });
innerSchema.virtual('computed');

  matrix: [[innerSchema]]
});

const Model = mongoose.model('Matrix', schema);

const raw = {
  _id: new mongoose.Types.ObjectId(),
  matrix: [
    [
      { name: 'a', computed: 'virtual-a' },
      { name: 'b', computed: 'virtual-b' }
    ]
  ]
};

const doc = Model.hydrate(raw, null, { virtuals: true });

console.log(doc.matrix[0][0].name);     // 'a' (correct)
console.log(doc.matrix[0][0].computed); // undefined (should be 'virtual-a')
```

### Expected Behavior

`doc.matrix[0][0].computed` should be `'virtual-a'`.

The `virtuals: true` option should propagate through nested arrays so that subdocuments at any nesting level have their virtuals set from the raw data.

vkarpov15:
Fix #15956

**Summary**

Just needed to pass down an extra option

**Examples**

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -398,6 +398,9 @@ SchemaArray.prototype.cast = function(value, doc, init, prev, options) {
           if (options.hydratedPopulatedDocs) {
             opts.hydratedPopulatedDocs = options.hydratedPopulatedDocs;
           }
+          if (options.virtuals) {
+            opts.virtuals = options.virtuals;
+          }
           rawValue[i] = caster.applySetters(rawValue[i], doc, init, void 0, opts);
         }
       } catch (e) {
--- a/test/model.hydrate.test.js
+++ b/test/model.hydrate.test.js
@@ -318,5 +318,43 @@ describe('model', function() {
 
       assert.ok(!hydrated.stories[1].article);
     });
+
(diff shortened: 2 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks 👍

PR #15968, opened 2026-01-17
fix(test): add missing Test.init() to fix flaky ordered create test

AbdelrahmanHafez:
The test for `Model.create()` with `ordered: true` (gh-4038) was flaky because it wasn't waiting for the unique index to be created before inserting documents. Without the index ready, all 11 docs get inserted instead of stopping at 5 when the first duplicate occurs.

Added `await Test.init()` to fix.

Examples of this flaky failure:
- https://github.com/Automattic/mongoose/actions/runs/21071975907/job/60603989655
- https://github.com/Automattic/mongoose/actions/runs/20978547122/job/60298605107

### DIFF
--- a/test/model.create.test.js
+++ b/test/model.create.test.js
@@ -189,6 +189,7 @@ describe('model', function() {
         });
 
         const Test = db.model('gh4038Test', testSchema);
+        await Test.init();
         const data = [];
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

Issue #15980, opened 2026-01-23
Migrate typescript tests to use tstyche


vkarpov15:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

Re: discussion from https://github.com/Automattic/mongoose/pull/15951, tstyche looks like a great tool for running the sort of TypeScript tests that we used tsd for. I'm opening up a separate issue for this since I don't want to merge #15951 first

mrazauskas:
Happy to see this idea. I can open a PR next week unless you have already started to work on it.

vkarpov15:
I haven't started working on this, so feel free to take a pass at it if you have the time :+1:

AbdelrahmanHafez:
While working on #15965 I noticed that the current `ExpectType` setup doesn't catch `any`, which means type regression tests can silently pass.

For example, if a function is supposed to return `{ title: string }` but regresses to `{ [key: string]: any }`, this test passes without errors:

```typescript
ExpectType<string>(obj.title); // passes, because obj.title is `any` and `any` is assignable to `string`
```

We only caught the regression by using this pattern instead:

```typescript
// @ts-expect-error title should be string, not any
const _check: number = obj.title;
```

This works because when `obj.title` is `string`, the assignment to `number` fails (and `@ts-expect-error` suppresses it). But when `obj.title` is `any`, the assignment succeeds, making the `@ts-expect-error` directive unused, which is a compile error.

mrazauskas:
I also saw few places where `ExpectType` is not catching `never`. It can be I mixed up something with setup or so.

For the beginning I wanted to introduce TSTyche for checking `// @ts-expect-error` messages. (Working on a PR.) I think that would be a good first step.

PR #15982, opened 2026-01-26
chore: type tests cleanup

mrazauskas:
Following up #15951

This PR replaces few more `tsd`s `expectType` with `ExpectType`. These are simply the leftovers.

I also updated contributing guidelines with the reference to `test:types`.


test/model.middleware.preposttypes.test.js
```diff
@@ -38,13 +38,13 @@ function getTypeName(obj) {
...
+/* Helper for generating type test test/types/middleware.preposttypes.test.ts,
+ * only active when "GEN_TYPE_TEST" is true
  */
-const GEN_TSD = false;
-const tsdOut = [];
-if (GEN_TSD) {
-  tsdOut.push('/* The following tests were generated by test/model.middleware.preposttypes.test.js (with GEN_TSD hardcoded set to true) */');
+const GEN_TYPE_TEST = true;
+const typeTestText = [];
+if (GEN_TYPE_TEST) {
```
mrazauskas, line 46: This script does not produce a correct type test. Unfortunately. My change is only removing the presence of `tsd`.

An interesting solution could be TSTyche’s [template test file](https://tstyche.org/guides/template-test-files). As a test runner, TSTyche can execute the code (by loading it via dynamic `import()`) and run the exported `typeTestText` as a type test.

I will play with that. This is a very interesting case which can be a part of #15980.

vkarpov15: Nice, template test files look very cool

PR #15992, opened 2026-01-30
Clean up npm audit warnings

vkarpov15:
**Summary**

Lots of security warnings due to broken-link-checker, nyc, and acquit-ignore. For acquit-ignore I just needed to bump version, but I replaced BLC with linkinator and nyc with c8 

**Examples**


vkarpov15:
There are still npm audit warnings from mongodb-js/mongodb-downloader, but I don't want to switch off of that for now.

hasezoey (approved):
Looks good to me.

The following may not be audit problems, but maybe should also be done at some point:
```txt
npm warn deprecated whatwg-encoding@3.1.1: Use @exodus/bytes instead for a more spec-conformant and faster implementation
npm warn deprecated lodash.isequal@4.5.0: This package is deprecated. Use require('node:util').isDeepStrictEqual instead.
npm warn deprecated eslint-plugin-markdown@5.1.0: Please use @eslint/markdown instead
npm warn deprecated tar@6.2.1: Old versions of tar are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exhorbitant rates) by contacting i@izs.me
```

TL;DR:
- change `eslint-plugin-markdown` to `@eslint/markdown`
- try upgrading `cheerio` to the latest version
- try upgrading `@mongodb-js/mongodb-downloader` to the latest version
- dont use `lodash.isequal`

Maybe i will put in a PR, but not today anymore.

vkarpov15:
Looks like we don't use eslint-plugin-markdown anymore re: #13156, `npm run lint:md` uses a different package, so we can remove that. I bumped cheerio but still get the whatwg-encoding warning.

lodash.isequal we will keep because we keep that explicitly for testing lodash.isequal. 

the tar warning is unfortunately tied to mongodb-runner and @mongodb-js/mongodb-downloader, which we can't really get away from right now.

Issue #15972, opened 2026-01-19
[Chore] Deprecate `new` and `returnOriginal` in favor of `returnDocument`


AbdelrahmanHafez:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

`new`, and `returnDocument` are deprecated. Docs, JSDocs, and TS types should reflect that

PR #16008, opened 2026-02-03
feat(base): add `returnDocument` global option, deprecate `returnOriginal` and `new`

AbdelrahmanHafez:
Re #15972 
This PR:
1. Shows a deprecation warning when users use `new` or `returnOriginal` in `findOneAndUpdate`, `findOneAndReplace,` and `findByIdAndUpdate`.
2. Shows a deprecation warning when users use `mongoose.set('returnOriginal')`.
3. Adds a new global option `mongoose.set('returnDocument')`.

In a future version (maybe 11?) we should be able to clean up `new` and `returnOriginal` completely from the codebase.

**Edit:**
I noticed that this addition polluted the tests with a bunch of unnecessary warnings, I'll try to figure out a way in a future PR to make warnings not appear at all in test logs.


lib/model.js
```diff
@@ -2296,7 +2296,7 @@ Model.$where = function $where() {
  * @param {Object} [conditions]
  * @param {Object} [update]
  * @param {Object} [options] optional see [`Query.prototype.setOptions()`](https://mongoosejs.com/docs/api/query.html#Query.prototype.setOptions())
- * @param {String} [options.returnDocument='before'] Has two possible values, `'before'` and `'after'`. By default, it will return the document before the update was applied.
+ * @param {'before'|'after'} [options.returnDocument='before'] Has two possible values, `'before'` and `'after'`. By default, it will return the document before the update was applied.
```
AbdelrahmanHafez, line 2335: This currently renders as `[object Object] | [object Object]` in the docs, working on a fix

AbdelrahmanHafez: https://github.com/Automattic/mongoose/pull/16009

lib/mongoose.js
```diff
@@ -233,7 +233,8 @@ Mongoose.prototype.setDriver = function setDriver(driver) {
...
  * - `objectIdGetter`: `true` by default. Mongoose adds a getter to MongoDB ObjectId's called `_id` that returns `this` for convenience with populate. Set this to false to remove the getter.
  * - `overwriteModels`: Set to `true` to default to overwriting models with the same name when calling `mongoose.model()`, as opposed to throwing an `OverwriteModelError`.
- * - `returnOriginal`: If `false`, changes the default `returnOriginal` option to `findOneAndUpdate()`, `findByIdAndUpdate`, and `findOneAndReplace()` to false. This is equivalent to setting the `new` option to `true` for `findOneAndX()` calls by default. Read our [`findOneAndUpdate()` tutorial](https://mongoosejs.com/docs/tutorials/findoneandupdate.html) for more information.
+ * - `returnDocument`: Set to `'before'` or `'after'` to set the default value for the `returnDocument` option to `findOneAndUpdate()`, `findByIdAndUpdate()`, and `findOneAndReplace()`. Defaults to `'before'`, which returns the document before the update was applied.
+ * - `returnOriginal`: **Deprecated.** Use `returnDocument` instead.
```
hasezoey: Maybe the original description should be preserved, in addition to the deprecated warning.

AbdelrahmanHafez:
```suggestion
 * - `returnOriginal`: **Deprecated.** Use `returnDocument` instead. If `false`, changes the default `returnOriginal` option to `findOneAndUpdate()`, `findByIdAndUpdate`, and `findOneAndReplace()` to false. This is equivalent to setting the `new` option to `true` for `findOneAndX()` calls by default. Read our [`findOneAndUpdate()` tutorial](https://mongoosejs.com/docs/tutorials/findoneandupdate.html) for more information.
```

lib/mongoose.js
```diff
@@ -313,6 +314,12 @@ Mongoose.prototype.set = function getsetOptions(key, value) {
           _mongoose.connections.shift();
         }
       }
+    } else if (optionKey === 'returnOriginal') {
```
vkarpov15, line 317: We should throw an error if calling `mongoose.set('returnOriginal', v)` when a global `returnDocument` is already set and vice versa - since `set('returnOriginal', v)` is a no-op if there's already a global `returnDocument`.

test/model.findOneAndUpdate.test.js
```diff
@@ -2229,4 +2230,155 @@ describe('model: findOneAndUpdate:', function() {
...
+      function createTestContext() {
+        sinon.stub(utils, 'warn');
+        const userSchema = new Schema({ name: String });
+        const User = db.model('User', userSchema);
+        return { User };
+      }
+    });
+
+    describe('mongoose.set() options', function() {
+      const originalReturnDocument = mongoose.get('returnDocument');
```
vkarpov15: I know in some other places we use this pattern of remembering the original global option value. However, I'd prefer if we tested global options by creating `const m = new mongoose.Mongoose();` and setting global options on the new Mongoose instance `m`. This prevents any accidental state leakage.

PR #16009, opened 2026-02-04
fix(docs): fix displaying string literal types

AbdelrahmanHafez:
JSDoc types like `'before'|'after'` were showing up as `[object Object]|[object Object]` in the generated API docs. This happens because `dox` uses `jsdoctypeparser` internally, which returns `STRING_VALUE` nodes for string literals, but `dox` doesn't handle those nodes properly and ends up calling `.toString()` on them.

The fix uses `typesDescription` (a correctly-formatted HTML string that `dox` provides) as a fallback when the parsed types contain `[object Object]`.

Also updated JSDoc types to use string literals instead of generic `String` where our TypeScript definitions already specify the exact values: `strict` (`Boolean|'throw'`), `explain` verbosity (`'queryPlanner'|'executionStats'|'allPlansExecution'`), `readConcern` level, and `addCursorFlag`.

<details>
<summary>Screenshots</summary>

**Before:**
before

**After:**
after

</details>

This PR builds on top of #16008 because it modifies some JSDoc params to `'before' | 'after'`  and I didn't want to re-do the work here.

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -719,7 +719,7 @@ Aggregate.prototype.read = function(pref, tags) {
  *
  *     await Model.aggregate(pipeline).readConcern('majority');
  *
- * @param {'local'|'available'|'majority'|'snapshot'|'linearizable'|'l'|'a'|'m'|'s'} level one of the listed read concern level or their aliases
+ * @param {'local'|'available'|'majority'|'snapshot'|'linearizable'|'l'|'a'|'m'|'s'|'lz'} level one of the listed read concern level or their aliases
  * @see mongodb https://www.mongodb.com/docs/manual/reference/read-concern/
  * @return {Aggregate} this
  * @api public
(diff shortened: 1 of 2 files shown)


lib/helpers/query/castUpdate.js
```diff
@@ -41,7 +41,7 @@ const mongodbUpdateOperators = new Set([
  * @param {Schema} schema
  * @param {Object} obj
  * @param {Object} [options]
- * @param {Boolean|String} [options.strict] defaults to true
+ * @param {Boolean|'throw'} [options.strict] defaults to true
```
vkarpov15, line 44: This is more of a discussion starter than a concrete suggestion: WDYT about changing this to lowercase "b" `{boolean|'throw'}` so we match TypeScript types? Given the popularity of TypeScript and the fact that this as written is technically valid TypeScript, I worry this comment could be misinterpreted as "strict is either a Boolean object or 'throw'"

PR #16011, opened 2026-02-05
test(document): enable and fix skipped pathsToSkip test

Divyapahuja31:
**Summary**

This PR enables a test case for `Model.validate()` with the `pathsToSkip` option that was previously skipped with a "TODO" comment. 

The test was originally skipped because the implementation was incorrect; it attempted to await `User.validate()` without catching the rejected Promise that occurs during a validation error. This caused the test to crash instead of asserting the error contents.

I have updated the test to correctly handle the Promise rejection and assert that the correct validation errors are returned, verifying that the `pathsToSkip` feature works as intended for `Model.validate()`.

**Examples**

**Before (Skipped):**
```javascript
it.skip('support `pathsToSkip` option for `Model.validate()`', async function() {
  const User = getUserModel();
  const err1 = await User.validate({}, { pathsToSkip: ['age'] });
  assert.deepEqual(Object.keys(err1.errors), ['name']);
});
```

**After (Enabled & Fixed):**
```javascript
it('support `pathsToSkip` option for `Model.validate()`', async function() {
  const User = getUserModel();
  const err1 = await User.validate({}, { pathsToSkip: ['age'] }).then(() => null, err => err);
  assert.deepEqual(Object.keys(err1.errors), ['name']);
  
  const err2 = await User.validate({}, { pathsToSkip: ['name'] }).then(() => null, err => err);
  assert.deepEqual(Object.keys(err2.errors), ['age']);
});
```

### DIFF
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -10858,7 +10858,6 @@ describe('document', function() {
       assert.deepEqual(Object.keys(err2.errors), ['age']);
     });
 
-    // skip until gh-10367 is implemented
     it('support `pathsToSkip` option for `Model.validate()`', async function() {
       const User = getUserModel();
       const err1 = await User.validate({}, { pathsToSkip: ['age'] }).then(() => null, err => err);


Divyapahuja31:
apologies for the unrelated formatting changes in the previous commit, i have fixed that formatting changes now it will be much easier to review it

test/document.test.js
```diff
@@ -10859,12 +10859,12 @@ describe('document', function() {
     });
 
     // skip until gh-10367 is implemented
```
AbdelrahmanHafez:
```suggestion
```

AbdelrahmanHafez (approved):
LGTM, thanks! 👍

PR #16014, opened 2026-02-07
chore: suppress mongoose warnings in tests

AbdelrahmanHafez:
re https://github.com/Automattic/mongoose/pull/16008#:~:text=I%20noticed%20that%20this%20addition%20polluted%20the%20tests%20with%20a%20bunch%20of%20unnecessary%20warnings%2C%20I%27ll%20try%20to%20figure%20out%20a%20way%20in%20a%20future%20PR%20to%20make%20warnings%20not%20appear%20at%20all%20in%20test%20logs https://github.com/Automattic/mongoose/pull/15911

This PR:
* Suppresses logs in tests to avoid polluting test outputs with unnecessary logs, `utils.warn` can still be asserted against, and we still validate that any invalid calls to `utils.warn` error out.
* Sets `GEN_TYPE_TEST` to `false` in `test/model.middleware.preposttypes.test.js` to avoid logging auto-generated TS file in all the CI runs, same argument as #15911. I suspect this was left as `true` accidentally.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -7,19 +7,6 @@
 const mongoose = require('../index');
 const Collection = mongoose.Collection;
 const assert = require('assert');
-// Suppress MONGOOSE warnings in tests to avoid polluting test output.
-// Tests that need to assert on warnings use sinon.stub(utils, 'warn') or
-// sinon.stub(process, 'emitWarning') which intercept before this handler.
-const _originalEmitWarning = process.emitWarning;
-process.emitWarning = function suppressedEmitWarning(message, options) {
-  if (options?.code === 'MONGOOSE') {
-    if (typeof message !== 'string') {
-      throw new Error('`utils.warn` should be called with a string, got: ' + typeof message);
-    }
-    return;
-  }
-  return _originalEmitWarning.apply(this, arguments);
-};
 
 const collectionNames = new Map();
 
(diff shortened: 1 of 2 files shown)


hasezoey (approved):

test/common.js
```diff
@@ -7,6 +7,19 @@
 const mongoose = require('../index');
 const Collection = mongoose.Collection;
 const assert = require('assert');
+// Suppress MONGOOSE warnings in tests to avoid polluting test output.
+// Tests that need to assert on warnings use sinon.stub(utils, 'warn') or
+// sinon.stub(process, 'emitWarning') which intercept before this handler.
+const _originalEmitWarning = process.emitWarning;
+process.emitWarning = function suppressedEmitWarning(message, options) {
+  if (options?.code === 'MONGOOSE') {
```
vkarpov15: I don't love stubbing out `emitWarning()` globally, this seems a bit hacky.

What do you think about running tests with `--disable-warning=MONGOOSE` instead? That tells Node.js to suppress warnings with code `MONGOOSE` but emit other warnings normally.

vkarpov15 (commented):
Also can you please update this PR to be against master branch?

test/common.js
AbdelrahmanHafez: Good call, I preferred the stubbing over `--disable-warning=MONGOOSE` to avoid adding it in multiple places, but we could add it to `.mocharc.yml`.

PR #16020, opened 2026-02-10
types(query): allow assigning QueryFilter<DocType> to QueryFilter<any>

### LINKED ISSUE #16006: TypeScript Error: FilterQuery with string _id not assignable to PipelineStage $match in Mongoose 9.1.5
**Description**
When using `FilterQuery<T>` where `T` is an interface with `_id: string` (not `ObjectId`), passing this filter object to an aggregation pipeline's `$match` stage causes a TypeScript error in Mongoose 9.1.5.
The error suggests that `RootFilterOperators` expects `_id` to be compatible with `ObjectId` (or `ApplyBasicQueryCasting<ObjectId>`), but the interface explicitly defines it as `string`, causing an incompatibility even if `_id` is not present in the filter object.

**Steps to Reproduce**
1. Define an interface with `_id: string`.
2. Create a `FilterQuery` object for that interface.
3. specific the filter query to a `$match` stage in an aggregation pipeline.

**Reproduction Script**
```typescript
import { FilterQuery, PipelineStage } from 'mongoose';

interface MyDoc {
  _id: string; // strict string, not ObjectId
  name: string;
}

const matchCriteria: FilterQuery<MyDoc> = {
  name: 'test'
};

// Error occurs here
const pipeline: PipelineStage[] = [
  {
    $match: matchCriteria
  }
];
```

**Expected Behavior**
The `FilterQuery` should be assignable to the `$match` stage input, as `_id` as a string is a valid use case (e.g. user-defined string IDs).

**Error Output**
```
Type '_QueryFilter<...>' is not assignable to type 'RootFilterOperators<{ [x: string]: any; _id?: ApplyBasicQueryCasting<ObjectId> | undefined; }>'
...
Types of property '_id' are incompatible.
  Type 'Condition<string | RegExp | string[] | null> | undefined' is not assignable to type 'Condition<((string | Types.ObjectId | ObjectId | ObjectId[] | undefined) & (string | Types.ObjectId | ObjectId | ObjectId[])) | null> | undefined'.
```

**Versions**
- Mongoose: 9.1.5
- TypeScript: 5.9.3

vkarpov15:
Fix #16006

**Summary**

The problem in #16006 is that `WithId<>` in `QueryFilter<any>` makes `QueryFilter<any>` act like a query filter with `_id: ObjectId`, which is potentially problematic because you would intuitively assume that `QueryFilter<{ _id: string }>` is assignable to `QueryFilter<any>`. 

I also removed a duplicate `UUIDToJSON` definition

**Examples**

### DIFF
--- a/test/types/PipelineStage.test.ts
+++ b/test/types/PipelineStage.test.ts
@@ -1,5 +1,5 @@
 import { ObjectId } from 'mongodb';
-import type { PipelineStage } from 'mongoose';
+import type { PipelineStage, QueryFilter } from 'mongoose';
 
 /**
  * $addFields:
@@ -574,3 +574,21 @@ const vectorSearchStages: PipelineStage[] = [
     }
   }
 ];
+
+function gh16006() {
+  interface MyDoc {
+    _id: string;
+    name: string;
+  }
+
+  const matchCriteria: QueryFilter<MyDoc> = {
+    _id: 'foobar',
+    name: 'test'
+  };
+
+  const pipeline: PipelineStage[] = [
+    {
+      $match: matchCriteria
+    }
+  ];
+}
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -945,29 +945,6 @@ declare module 'mongoose' {
                     : BufferToBinary<T[K]>;
           } : T;
 
-    /**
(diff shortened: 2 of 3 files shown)


AbdelrahmanHafez (approved):
LGTM 👍

PR #16022, opened 2026-02-11
fix(document): fix pathToSave crashing with no updates, and ignoring non-`$set` operators

AbdelrahmanHafez:
re https://github.com/Automattic/mongoose/pull/15850 https://github.com/Automattic/mongoose/pull/15896
This PR fixes two issues:
* When calling `Document#save`  with `pathsToSave` on a document with no changes, mongoose crashes with `TypeError: Cannot read properties of null`.
* Updates that would use `$push` or any other non-`$set` operator would silently be ignored.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -407,11 +407,14 @@ Model.prototype.$__save = async function $__save(options) {
       this.$__.inserting = false;
       const delta = this.$__delta();
 
-      if (options.pathsToSave && delta) {
+      if (options.pathsToSave && delta?.[1]) {
         const update = delta[1];
+        const versionKey = this.$__schema.options.versionKey;
         for (const op of Object.keys(update)) {
           for (const key in update[op]) {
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez:
@vkarpov15 I got into a bit of a rabbit hole with this feature. One thing I noticed is that `$__reset()` clears all dirty state after `save()`, even when using `pathsToSave`:

```js
const user = await User.create({ name: 'John', age: 30 });
user.name = 'Jane';
user.age = 31;
await user.save({ pathsToSave: ['name'] });

user.isModified('age'); // false
user.directModifiedPaths(); // []
```

So the user ends up with data in memory that's out of sync with the DB, and no way to know about it. I think when using `pathsToSave` we should only clear the paths that were actually saved, so a subsequent `save()` would persist the remaining fields. What do you think?

AbdelrahmanHafez:
Closing this in favor of #16027

PR #16024, opened 2026-02-11
chore: introduce TSTyche for type testing

mrazauskas:
**Summary**

Reference #15980

This PR is introducing TSTyche for type testing. Documentation: https://tstyche.org

For the beginning I only used the tool to check errors suppress by `// @ts-expect-error`.

The next possible step could be to test agains several TypeScript versions. For example, `tstyche --target '>=5.6'`. I would recommend adding that for CI only.

Since assertions are not touched, TS7 could be used in CI as well. TSTyche does not support TS7, because it does not have programmatic API yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "test-deno:ci": "npm run test-deno -- --reporter min",
     "test-rs": "START_REPLICA_SET=1 mocha --timeout 30000 --exit ./test/*.test.js",
     "test-rs:ci": "npm run test-rs -- --reporter min",
-    "test:types": "tsc --project test/types/tsconfig.json && tstyche",
+    "test:types": "tstyche",
     "setup-test-encryption": "node scripts/setup-encryption-tests.js",
     "test-encryption": "mocha --exit ./test/encryption/*.test.js",
     "test-encryption:ci": "npm run test-encryption -- --reporter min",


tstyche.config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://tstyche.org/schemas/config.json",
+  "checkDeclarationFiles": false,
```
mrazauskas: Please the remove the line with `checkDeclarationFiles` and run `npx tstyche`. Duplicate identifier gets reported in `index.d.ts`. Is that expected?

test/types/maps.test.ts
```diff
@@ -1,4 +1,4 @@
-import { Schema, model, Document, Model, Types } from 'mongoose';
```
mrazauskas, line 1: Not used.

test/types/middleware.preposttypes.test.ts
```diff
@@ -22,7 +22,7 @@ schema.pre('save', function() {
 
 schema.post('save', function(res) {
   ExpectAssignable<HydratedDocument<IDocument>>()(this);
-  // @ts-expect-error res should be a document, no query type
+  // @ts-expect-error  Type 'Document...' is missing the following properties from type 'Query...'
```
mrazauskas, line 25: The messages can be truncated using `...`. This is a good example.

package.json
```diff
@@ -98,7 +99,7 @@
     "test-deno:ci": "npm run test-deno -- --reporter min",
     "test-rs": "START_REPLICA_SET=1 mocha --timeout 30000 --exit ./test/*.test.js",
     "test-rs:ci": "npm run test-rs -- --reporter min",
-    "test:types": "tsc --project test/types/tsconfig.json",
+    "test:types": "tsc --project test/types/tsconfig.json && tstyche",
```
vkarpov15: Minor question - is the tsc still required here? I think this script can just be `"tstyche"`

vkarpov15 (approved):
One minor question but otherwise I love it. This is awesome 😃

package.json
mrazauskas: Right. In this case, `tsc` doesn’t do anything additional. I simplified the script.

vkarpov15: Great thank you! Merging :+1: :rocket:

PR #16025, opened 2026-02-11
feat(setDefaultsOnInsert): pass query as context to default functions

vkarpov15:
**Summary**

Validators get query as context, but default functions still get `null` as context, which makes writing default functions that depend on other fields tricky.

**Examples**


hasezoey (commented):
LGTM

PR #16027, opened 2026-02-14
fix(document): make pathsToSave filter all update operators and preserve unsaved state

AbdelrahmanHafez:
re https://github.com/Automattic/mongoose/pull/15850 https://github.com/Automattic/mongoose/pull/15896 https://github.com/Automattic/mongoose/pull/16022#issuecomment-3882060185

The current `pathsToSave` behavior had a few gaps:
- `save({ pathsToSave })` could crash when there are no local changes.
- Filtering only applied to `$set`, so non-`$set` updates like `$inc`, `$addToSet`, `$push`, `$pull`, `$unset`, etc. could still get persisted even when excluded.
- Version handling did not consistently reflect the filtered update set, especially with `optimisticConcurrency` modes.
- After partial saves, excluded changes could be marked clean, so a later plain `save()` could no-op even though in-memory data was still dirty.

This PR fixes that by moving `pathsToSave` filtering to `Document#$__delta()`, so filtering happens at the dirty-path to update-op boundary.

This PR:
- Applies `pathsToSave` filtering before operator generation, so all update operators are handled consistently, not just `$set`.
- Preserves no-op behavior correctly, `save({ pathsToSave })` no longer crashes when there are no matching changes.
- Keeps version mode tied to filtered changes, including `optimisticConcurrency` array/exclude modes.
- Preserves unsaved state after partial saves by restoring excluded dirty/default paths after reset, so subsequent `save()` persists remaining changes.
- Preserves unsaved array atomics (`$push`, `$addToSet`, etc.) so follow-up saves keep atomic operators instead of degrading to `$set`.
- Uses a compiled matcher (`createPathsToSaveMatcher`) with a `Set` for exact-path checks plus prefix checks for dot descendants, which avoids repeatedly scanning `pathsToSave` arrays in multiple loops.

Also added focused tests for:
- no-change saves,
- multi-operator filtering,
- dot-notation parent/child behavior,
- empty filtered update behavior,
- dirty-state preservation across saves,
- array atomic preservation,
- custom `versionKey`,
- `optimisticConcurrency` array/exclude interactions with `pathsToSave`,
- existing validation behavior where `pathsToSave` remains default validation scope.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -5027,17 +5027,19 @@ Document.prototype.$__delta = function $__delta(pathsToSave, pathsToSaveSet) {
     if (Array.isArray(optimisticConcurrency)) {
       const optCon = new Set(optimisticConcurrency);
       const modPaths = this.modifiedPaths();
-      const relevantModPaths = pathsToSave == null ? modPaths :
-        modPaths.filter(path => isInPathsToSave(path, pathsToSaveSet, pathsToSave));
-      if (relevantModPaths.some(path => optCon.has(path))) {
+      const hasRelevantModPaths = pathsToSave == null ?
+        modPaths.find(path => optCon.has(path)) :
+        modPaths.find(path => optCon.has(path) && isInPathsToSave(path, pathsToSaveSet, pathsToSave));
+      if (hasRelevantModPaths) {
         this.$__.version = dirty.length ? VERSION_ALL : VERSION_WHERE;
       }
     } else if (Array.isArray(optimisticConcurrency?.exclude)) {
       const excluded = new Set(optimisticConcurrency.exclude);
       const modPaths = this.modifiedPaths();
-      const relevantModPaths = pathsToSave == null ? modPaths :
-        modPaths.filter(path => isInPathsToSave(path, pathsToSaveSet, pathsToSave));
-      if (relevantModPaths.some(path => !excluded.has(path))) {
+      const hasRelevantModPaths = pathsToSave == null ?
+        modPaths.find(path => !excluded.has(path)) :
+        modPaths.find(path => !excluded.has(path) && isInPathsToSave(path, pathsToSaveSet, pathsToSave));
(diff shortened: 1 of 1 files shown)


lib/document.js
```diff
@@ -4998,26 +4999,35 @@ Document.prototype.getChanges = function() {
...
+  const dirty = pathsToSave == null ? allDirty : allDirty.filter(data => isPathIncluded(data.path));
   const optimisticConcurrency = this.$__schema.options.optimisticConcurrency;
   if (optimisticConcurrency) {
     if (Array.isArray(optimisticConcurrency)) {
       const optCon = new Set(optimisticConcurrency);
       const modPaths = this.modifiedPaths();
-      if (modPaths.some(path => optCon.has(path))) {
+      const relevantModPaths = pathsToSave == null ? modPaths :
+        modPaths.filter(path => isPathIncluded(path));
+      if (relevantModPaths.some(path => optCon.has(path))) {
```
AbdelrahmanHafez: I suppose we can merge the `filter` and the `some` here, I'll do that when I have some time.

lib/model.js
```diff
@@ -526,6 +522,53 @@ Model.prototype.$__save = async function $__save(options) {
...
+  }
+
+  for (const path of unsavedState.unsavedDefaultPaths) {
+    doc.$__.activePaths.default(path);
+  }
+}
+
+/*!
+ * ignore
+ */
```
hasezoey:
```suggestion
```

This looks like there was a function once, but is not anymore

lib/model.js
```diff
@@ -526,6 +522,53 @@ Model.prototype.$__save = async function $__save(options) {
...
+
+  return {
+    unsavedDirty,
+    unsavedDefaultPaths
+  };
+}
+
+/*!
+ * ignore
+ */
```
hasezoey, line 537: Instead of plain ignored, the function should likely have a little bit of documentation

PR #16029, opened 2026-02-14
refactor: remove duplicate definition of the `UUIDToJSON` type

mrazauskas:
**Summary**

Is PR removes duplicate remove duplicate definition of the `UUIDToJSON` type. Seem like the type is rename to `UUIDToString`.

`UUIDToJSON` is also exported here as an alias of `UUIDToString`:

https://github.com/Automattic/mongoose/blob/7a73025db14ccdbd2691cb36ceede4776d27fadb/types/index.d.ts#L971-L975

---

By the way, TSTyche detected this duplication while I was putting together #16024.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -945,29 +945,6 @@ declare module 'mongoose' {
                     : BufferToBinary<T[K]>;
           } : T;
 
-    /**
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #16034, opened 2026-02-20
fix(connection): handle calling watch() on disconnected connection

vkarpov15:
**Summary**

Currently calling `connection.watch()` on a connection that you haven't called `openUri()` on throws an error "cannot read property watch of undefined", which is inconsistent with other connection helpers. Other connection helpers use `_waitForConnect()` to wait for the connection to connect subject to buffering.

**Examples**


AbdelrahmanHafez (changes requested):
Seems like this broke [a test](https://github.com/Automattic/mongoose/actions/runs/22209198987/job/64242096011?pr=16034) for `watch()`

vkarpov15:
Merging into 9.3 in an abundance of caution

Issue #16018, opened 2026-02-10
[Docs] JSDoc `boolean` vs `Boolean`


AbdelrahmanHafez:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

re https://github.com/Automattic/mongoose/pull/16009#discussion_r2776469187

vkarpov15:
WDYT, is this worth doing? On one hand updating Boolean -> boolean in our JSdoc types isn't an urgent priority, but I do think it would be nice for our JSdoc types to line up with TypeScript types for clarity.

AbdelrahmanHafez:
Yeah I checked earlier today and `boolean` is actually valid JSDoc, I went ahead and did some work on this locally and aligned our JSDoc with TS types as much as possible, I'll submit a draft PR just to take a look there.

PR #16036, opened 2026-02-20
docs: use lowercase primitive types re #16018

AbdelrahmanHafez:
re #16018

This PR lowercases `Boolean`, `String`, `Number`, and `Object` to `boolean`, `string`, `number`, and `object` in JSDoc annotations across `lib/`. This aligns our JSDoc with TypeScript primitive types so there's no ambiguity between e.g. `boolean` (the primitive) and `Boolean` (the wrapper object).

`Array` and `Function` were left as-is since we don't know what the types of array elements are, and `Function` as a class reference is more conventional in JSDoc. Class names like `Schema`, `Document`, `Model`, `Date`, `Buffer`, `Promise`, etc. were also left untouched.

### DIFF
--- a/lib/cast/bigint.js
+++ b/lib/cast/bigint.js
@@ -7,7 +7,7 @@ const { Long } = require('mongodb/lib/bson');
  * cannot be casted. `null` and `undefined` are considered valid.
  *
  * @param {any} value
- * @return {number}
+ * @return {bigint|null|undefined}
  * @throws {Error} if `value` is not one of the allowed values
  * @api private
  */


lib/schema/bigint.js
```diff
@@ -263,7 +263,7 @@ SchemaBigInt.prototype._castNullish = function _castNullish(v) {
  *
  * @param [options]
  * @param [options.useBsonType=false] If true, return a representation with `bsonType` for use with MongoDB's `$jsonSchema`.
```
hasezoey: This looks invalid. (missing type)
Note that this is like this on all SchemaType's `toJSONSchema` functions.

Though not the main purpose of this PR.

lib/document.js
```diff
@@ -1060,11 +1060,11 @@ Document.prototype.overwrite = function overwrite(obj) {
...
  * Alias for `set()`, used internally to avoid conflicts
  *
- * @param {String|Object} path path or object of key/vals to set
- * @param {Any} val the value to set
- * @param {Schema|String|Number|Buffer|*} [type] optionally specify a type for "on-the-fly" attributes
- * @param {Object} [options] optionally specify options that modify the behavior of the set
- * @param {Boolean} [options.merge=false] if true, setting a [nested path](https://mongoosejs.com/docs/subdocs.html#subdocuments-versus-nested-paths) will merge existing values rather than overwrite the whole object. So `doc.set('nested', { a: 1, b: 2 })` becomes `doc.set('nested.a', 1); doc.set('nested.b', 2);`
+ * @param {string|object} path path or object of key/vals to set
+ * @param {any} val the value to set
+ * @param {Schema|string|number|Buffer|*} [type] optionally specify a type for "on-the-fly" attributes
```
hasezoey: The asterisk here should likely be changed to `any`
(and all the other locations)

lib/cursor/aggregationCursor.js
```diff
@@ -293,10 +293,10 @@ AggregationCursor.prototype.next = async function next() {
  * Returns a promise that resolves when done.
  *
  * @param {Function} fn
- * @param {Object} [options]
- * @param {Number} [options.parallel] the number of promises to execute in parallel. Defaults to 1.
- * @param {Number} [options.batchSize=null] if set, Mongoose will call `fn` with an array of at most `batchSize` documents, instead of a single document
- * @param {Boolean} [options.continueOnError=false] if true, `eachAsync()` iterates through all docs even if `fn` throws an error. If false, `eachAsync()` throws an error immediately if the given function `fn()` throws an error.
+ * @param {object} [options]
```
vkarpov15, line 296: I checked and lowercase "o" object is actually correct here, because "object" is how you exclude primitives.

lib/schema/array.js
```diff
@@ -248,7 +248,7 @@ SchemaArray.prototype.checkRequired = function checkRequired(value, doc) {
  * Adds an enum validator if this is an array of strings or numbers. Equivalent to
  * `SchemaString.prototype.enum()` or `SchemaNumber.prototype.enum()`
  *
- * @param {...String|Object} [args] enumeration values
+ * @param {...string|object} [args] enumeration values
```
vkarpov15, line 251: Any ideas how we can better represent this case? `...string|object` is not valid TypeScript, so I worry this is inconsistent and unclear. Would `{string|object} [...args]` be more clear?

lib/schema/bigint.js
vkarpov15: Yeah should add types here, useBsonType is a boolean and options is an object

lib/schema/string.js
```diff
@@ -204,7 +204,7 @@ SchemaString.checkRequired = SchemaType.checkRequired;
  *     await m.save();
  *     // success
  *
- * @param {...String|Object} [args] enumeration values
+ * @param {...string|object} [args] enumeration values
```
vkarpov15, line 207: Same here, I don't love this syntax because it isn't valid TypeScript, but I don't have a better alternative right now.

lib/types/documentArray/methods/index.js
```diff
@@ -230,7 +230,7 @@ const methods = {
   /**
    * Pulls items from the array atomically.
    *
-   * @param {...Object} [args]
+   * @param {...object} [args]
```
vkarpov15, line 233: In this case, the spread operator looks more clear, even though this still isn't valid TypeScript

lib/document.js
```diff
@@ -1695,14 +1695,14 @@ Document.prototype.$__shouldModify = function(pathToMark, path, options, constru
 /**
  * Handles the actual setting of the value and marking the path modified if appropriate.
  *
- * @param {String} pathToMark
- * @param {String|Symbol} path
- * @param {Object} options
- * @param {Any} constructing
+ * @param {string} pathToMark
+ * @param {string|Symbol} path
```
vkarpov15: I think lowercase "s" symbol is more correct here because "Symbol" refers to the wrapper class for the symbol primitive, like "Boolean" vs "boolean"

lib/document.js
```diff
@@ -1919,12 +1919,12 @@ Document.prototype.$__setValue = function(path, val) {
...
  *     doc.get('age', String) // "47"
  *
- * @param {String} path
- * @param {Schema|String|Number|Buffer|*} [type] optionally specify a type for on-the-fly attributes
- * @param {Object} [options]
- * @param {Boolean} [options.virtuals=false] Apply virtuals before getting this path
- * @param {Boolean} [options.getters=true] If false, skip applying getters and just get the raw value
- * @return {Any}
+ * @param {string} path
+ * @param {Schema|string|number|Buffer|*} [type] optionally specify a type for on-the-fly attributes
```
vkarpov15: "any" here instead of *

vkarpov15 (commented):
A couple of small comments. Overall looks great, I just want to think a little more about how we represent spreads in our docs

lib/schema/array.js
hasezoey: It may not be valid typescript, but this is *JSDoc*, where this is how the spread operator is documented.
See [tags-param](https://jsdoc.app/tags-param), section `Allows a parameter to be repeated`.
(infact `[...args]` is *invalid* JSDoc, to my knowledge)

Edit:
<details>
<summary>Example</summary>

Valid:
correct

Invalid:
invalid

Though note that if i change it to `string|object` instead of just `string` it becomes `any[] | undefined`(instead of the previous `string[] | undefined`) in the valid case, due to typescripts's type unification.

</details>

lib/document.js
hasezoey: Just for context, `*` is the correct JSDoc type, but we practically use `any` everywhere else.
(And typescript's JSDoc parser does not care if it is `*` or `any`, unless `any` is a object that is defined, which it is not in mongoose)

lib/schema/array.js
hasezoey: If this is meant to make it clearer what options are available for the `object` case, you can use `@typedef` to define a object in JSDoc (similar to typescripts `interface`) and use it there.

<details>
<summary>Example</summary>

With a defined object:
typedef

</details>

vkarpov15: Ok yeah this works for me, if `...string` is valid for jsdoc then the current approach is fine. I haven't found an approach that I like better. I'll resolve this comment.

AbdelrahmanHafez:
@vkarpov15 @hasezoey Feedback addressed 👍

lib/stateMachine.js
```diff
@@ -180,8 +179,7 @@ StateMachine.prototype._iter = function _iter(iterMethod) {
  * this.forEach(state1, state2, fn); // iterates over all paths in state1 or state2
  * this.forEach(fn);                 // iterates over all paths in all states
  *
- * @param {String} [state]
- * @param {String} [state]
+ * @param {...string} [state] one or more state names to filter by
  * @param {Function} callback
  * @api private
```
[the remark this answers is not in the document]
vkarpov15: I'll mark this one as resolved since I think since, while not technically valid, there's no valid way to express the syntactic constraints of this function. So I think we should just redo how this function works, but leave the jsdoc as is for now.

vkarpov15 (approved):

PR #16037, opened 2026-02-20
Automatically map mongodb driver urls to the current driver version

hasezoey:
**Summary**

I had once again noticed that we have a bunch of links to the driver documentation, but never actually keep them up-to-date properly, resulting in outdated links (ex `4.9`, `6.15`), potentially leading to outdated documentation.
This PR changes the website generation to automatically change the links to the current driver's version.
Downside is that this link replacer does not check the links, so they might be broken links then (ex renamed interface).

Also fixes a warning of `SchemaType.prototype.toJSON` being both `instance` and `static`.

Example:
`https://mongodb.github.io/node-mongodb-native/4.9/classes/Collection.html#countDocuments` -> `https://mongodb.github.io/node-mongodb-native/7.0/classes/Collection.html#countDocuments`

### DIFF
--- a/scripts/website.js
+++ b/scripts/website.js
@@ -369,19 +369,27 @@ const mongooseComRegex = /(?:href=")(https:\/\/mongoosejs\.com\/?)/g;
 const versionedDocs = /docs\/\d/;
 
 /**
- * Map urls (https://mongoosejs.com/) to local paths
+ * Map urls the given urls to their latest or local paths
  * @param {String} block The String block to look for urls
  * @param {String} currentUrl The URL the block is for (non-versioned)
  */
 function mapURLs(block, currentUrl) {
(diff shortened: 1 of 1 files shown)


vkarpov15 (commented):
I'm thinking I would prefer an automated check that updates the links in our code, rather than transforming in scripts/website.js. scripts/website would just update the website, but that doesn't ensure the links stay consistent in our `docs/` markdown files or our jsdoc, and people may be reading those directly instead of going through https://mongoosejs.com.

hasezoey:
> I'm thinking I would prefer an automated check that updates the links in our code, rather than transforming in scripts/website.js

Sounds better. In what way would you like to see this being updated, another `script/` that is *manually* run, or a CI action that is *automatically* ran on driver update detection?
Personally, i think at least at first a manual approach would be better.

PR #16043, opened 2026-02-23
Validation error stack trace improvements and some performance improvements for saving $isNew

vkarpov15:
**Summary**

In order to make ValidationErrors show less Mongoose internals, I moved `new ValidationError()` calls up a couple of levels so now a save error stack trace looks like this:

```
ValidationError: User validation failed: age: Age must be at least 18
    at model.validate (/home/v/Desktop/MongoDB/mongoose/lib/document.js:2893:36)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async model.validateBeforeSavePreSave (/home/v/Desktop/MongoDB/mongoose/lib/plugins/validateBeforeSave.js:39:5)
    at async Kareem.execPre (/home/v/Desktop/MongoDB/mongoose/node_modules/kareem/index.js:70:24)
    at async model.$__save (/home/v/Desktop/MongoDB/mongoose/lib/model.js:376:5)
    at async model.save (/home/v/Desktop/MongoDB/mongoose/lib/model.js:642:5)
    at async Function.create (/home/v/Desktop/MongoDB/mongoose/lib/model.js:2729:5)
    at async run (/home/v/Desktop/MongoDB/troubleshoot-mongoose/test.js:23:16)
```

No more anonymous functions or Mongoose internals. Plus I updated `create()` to avoid using `Promise.all()` in the case where there's only 1 doc, which also helps.

I also started iterating on the `saveSimple` benchmark to see if we can close the gap between the MongoDB Node driver and Mongoose. I found a couple of quick wins:

**before**

```
{
  "Average mongoose save time ms": 2.59,
  "Average driver insertOne time ms": 0.77
}
```

**after**

```
{
  "Average mongoose save time ms": 2.29,
  "Average driver insertOne time ms": 0.76
}
```

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -4157,12 +4157,11 @@ Query.prototype._updateMany = async function _updateMany() {
   }
 
   const options = this._optionsForExec(this.model);
-  removeUnusedArrayFilters(this._update, options);
-
   this._update = this._castUpdate(this._update);
   if (this._update == null || utils.hasOwnKeys(this._update) === false) {
     return { acknowledged: false };
   }
+  removeUnusedArrayFilters(this._update, options);
 
   this._update = setDefaultsOnInsert(
     this._conditions,
@@ -4203,12 +4202,11 @@ Query.prototype._updateOne = async function _updateOne() {
   }
 
   const options = this._optionsForExec(this.model);
-  removeUnusedArrayFilters(this._update, options);
-
   this._update = this._castUpdate(this._update);
   if (this._update == null || utils.hasOwnKeys(this._update) === false) {
     return { acknowledged: false };
   }
+  removeUnusedArrayFilters(this._update, options);
 
   this._update = setDefaultsOnInsert(
     this._conditions,
@@ -4249,9 +4247,8 @@ Query.prototype._replaceOne = async function _replaceOne() {
   }
 
   const options = this._optionsForExec(this.model);
-  removeUnusedArrayFilters(this._update, options);
-
   this._update = new this.model(this._update, null, { skipId: true });
+  removeUnusedArrayFilters(this._update, options);
 
   if (_getOption(this, 'runValidators', false)) {
     await this.validate(this._update, options, true);
--- a/lib/model.js
+++ b/lib/model.js
@@ -373,9 +373,12 @@ function _createSaveOptions(doc, options) {
 
 Model.prototype.$__save = async function $__save(options) {
   try {
-    const shouldValidateBeforeSave = options?.validateBeforeSave == null ?
-      this.$__schema.options.validateBeforeSave :
-      !!options.validateBeforeSave;
+    const hasValidateBeforeSaveOption = options &&
+      (typeof options === 'object') &&
+      ('validateBeforeSave' in options);
+    const shouldValidateBeforeSave = hasValidateBeforeSaveOption ?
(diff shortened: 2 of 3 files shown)


test/model.skip.middleware.test.js
```diff
@@ -639,7 +639,7 @@ describe('middleware option to skip hooks (gh-8768)', function() {
       // Assert - all hooks should be built-in (no user hooks were added)
       const builtInHooks = allHooks.filter(hook => hook.fn[builtInMiddleware]);
 
-      assert.ok(builtInHooks.length >= 18, 'Expected at least 18 built-in hooks'); // 18 is current count, may increase in future
+      assert.ok(builtInHooks.length >= 17, 'Expected at least 17 built-in hooks'); // 18 is current count, may increase in future
```
hasezoey:
```suggestion
      assert.ok(builtInHooks.length >= 17, 'Expected at least 17 built-in hooks'); // 17 is current count, may increase in future
```

lib/query.js
hasezoey, line 1: I dont quite like how we now have practically the same code in 3 different places, but if this gives enough performance, i guess it is fine?

hasezoey (commented):
Looks good to me, though due to it modifying functions directly on some classes (even if they are marked internal), i would recommend moving this to the next minor version.

lib/query.js
vkarpov15: I removed `_updateThunk()` primarily for cleaner stack traces, as well as cleaning up a somewhat clunky abstraction - there was a little more discrepancy between `replaceOne()` path and `updateX()` path than I preferred. I reduced some of the boilerplate in [20f6496](https://github.com/Automattic/mongoose/pull/16043/commits/20f6496301a67d0ddaf740861cbbd459aabeb5f8), but these methods are roughly consistent with the amount of complexity in the other `_find()`, `_findOneAndUpdate()` etc. methods.

lib/query.js
```diff
@@ -4206,7 +4196,39 @@ Query.prototype._updateMany = async function _updateMany() {
...
+    this._update,
+    options,
+    this._mongooseOptions
+  );
+
+  if (_getOption(this, 'runValidators', false)) {
+    await this.validate(this._update, options, false);
+  }
+
+  if (this._update.toBSON) {
```
AbdelrahmanHafez: For consistency with the rest of the conditions
```suggestion
  if (typeof this._update.toBSON === 'function') {
```

lib/document.js
```diff
@@ -2699,7 +2698,234 @@ Document.prototype.validate = async function validate(pathsToValidate, options)
...
+          // Make sure cast errors persist
+          if (!this[documentArrayParent] &&
+              error.errors[key] instanceof MongooseError.CastError) {
+            this.invalidate(key, error.errors[key]);
+          }
+        }
+      }
+
+      await this._execDocumentPostHooks('validate', options, error);
+      return;
```
AbdelrahmanHafez: Those ~25 lines look like they're duplicated again later at line 2897.

Deduplicating this logic in a private function would help maintainability and would still preserve the stack trace benefits since the function would be called from within the `validate` method.

AbdelrahmanHafez (commented):
Couple of suggestions, otherwise LGTM. 👍

PR #16044, opened 2026-02-24
chore: make the TSConfig stricter

mrazauskas:
**Summary**

This PR makes the TSConfig stricter, by removing `strictPropertyInitialization: false` and `noImplicitAny: false`. Only few minor changes had to be included in the type test files.

TSTyche is using this `tsconfig.json` file and it checks `.d.ts` declarations based on it (as an example, #16029 was detected). So I think it is better to avoid settings like `noImplicitAny: false`.

### DIFF
--- a/test/types/tsconfig.json
+++ b/test/types/tsconfig.json
@@ -4,15 +4,11 @@
     "esModuleInterop": false,
     "strict": true,
     "allowSyntheticDefaultImports": true,
-    "strictPropertyInitialization": false,
-    "noImplicitAny": false,
-    "strictNullChecks": true,
     "module": "commonjs",
     "target": "ES2022",
     "noEmit": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "types": ["node"]
   },
-  "include": [
-    "*.test.ts"
-  ]
+  "include": ["**/*"]
 }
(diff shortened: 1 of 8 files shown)


test/types/tsconfig.json
```diff
@@ -4,15 +4,11 @@
...
     "allowSyntheticDefaultImports": true,
-    "strictPropertyInitialization": false,
-    "noImplicitAny": false,
-    "strictNullChecks": true,
     "module": "commonjs",
     "target": "ES2022",
     "noEmit": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "types": ["node"]
```
mrazauskas, line 11: This is needed to make `tstyche --target next` pass. Will be required in TypeScript 6.

test/types/tsconfig.json
```diff
@@ -4,15 +4,11 @@
...
   },
-  "include": [
-    "*.test.ts"
-  ]
+  "include": ["**/*"]
```
mrazauskas, line 13: Almost the same as before, because only files from `‎test/types` are included.

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "strict": true,
-    "strictNullChecks": true,
```
mrazauskas, line 4: `strictNullChecks` is enabled by `strict: true` above.

Issue #16012, opened 2026-02-06
TypeScript: Regression in allowing Virtuals with Setters in Document creation/initialization after upgrading to v9


nikzanda:
### Prerequisites

- [x] I have written a descriptive issue title

### Mongoose version

9.1.6

### Node.js version

24.11.0

### MongoDB version

6.0.2

### Operating system

macOS

### Issue

After upgrading from Mongoose 8 to Mongoose 9, I've encountered a strict typing issue when initializing documents that contain subdocuments with virtual setters.

In Mongoose 8, it was possible to pass a value to a virtual field (which has a setter) during document creation. In Mongoose 9, the TypeScript compiler throws an error, forcing the use of the raw database paths instead of the virtual path.

### Schema setup
I have a decimalSchema used as a subdocument. It stores a number as base and exponent but exposes a value virtual for ease of use:
```ts
import {
  HydratedSingleSubdocument,
  Model,
  Schema,
  SchemaOptions,
  SchemaTypes,
} from 'mongoose';

export interface IDecimal {
  base: number;
  exponent: number;
}

export interface IDecimalVirtuals {
  value: number;
}

export type DecimalInstance = HydratedSingleSubdocument<IDecimal, IDecimalVirtuals>;

type DecimalModelType = Model<IDecimal, {}, {}, IDecimalVirtuals, DecimalInstance>;

type DecimalSchemaOptions = SchemaOptions<
  IDecimal,
  {},
  {},
  {},
  IDecimalVirtuals,
  DecimalInstance,
  DecimalModelType
>;

const options: DecimalSchemaOptions = {
  _id: false,
  toJSON: { virtuals: true },
  toObject: { virtuals: true },
};

const decimalSchema = new Schema<
  IDecimal,
  DecimalModelType,
  {},
  {},
  IDecimalVirtuals,
  {},
  DecimalSchemaOptions,
  IDecimal,
  DecimalInstance
>({
  base: {
    type: SchemaTypes.Number,
    required: true,
  },
  exponent: {
    type: SchemaTypes.Number,
    required: true,
  },
}, options);

decimalSchema.virtual('value')
  .get((_value, _virtual, doc) => fromDecimal(doc)) // returns number
  .set((value: number, _virtual, doc) => doc.set(toDecimal(value))); // converts number to IDecimal

export default decimalSchema;
```

### The issue
When I try to create a parent document containing this subdocument:
```ts
// Parent Schema usage
const parentSchema = new Schema({
[...]

PR #16047, opened 2026-02-25
types: fix ApplyBasicCreateCasting and DeepPartial to better support VSCode intellisense

vkarpov15:
**Summary**

While working on #16012 I spotted that intellisense for `create()` wasn't working at all 

image

image

image

With this change, seems to work much better:

image

image

image

image

image

I borrowed heavily from the [Vercel AI SDK's DeepPartial type](https://github.com/vercel/ai/blob/4a893418b05e4aa7c6be95984aeea82894ab3f96/packages/ai/src/util/deep-partial.ts#L32) for this implementation - their DeepPartial seems to work better with intellisense.

For some reason Boolean is still showing up as `any` in intellisense, I'm looking into that, but this change should still improve intellisense across the board.

@AbdelrahmanHafez @hasezoey can you check whether this helps intellisense on your IDE?

**Examples**

### DIFF
--- a/types/models.d.ts
+++ b/types/models.d.ts
@@ -189,9 +189,6 @@ declare module 'mongoose' {
    * - vanilla arrays of POJOs for document arrays
    * - POJOs and array of arrays for maps
    */
-  /* type ApplyBasicCreateCasting<T> = {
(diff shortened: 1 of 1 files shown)


vkarpov15:
Checking what @hasezoey , it looks like intellisense works fine without this PR too :sob: Likely some strange flake with my local dev environment. Next time I'll have to make sure to restart Zed and reboot TypeScript server a few times just in case.

PR #16048, opened 2026-02-26
types: allow unknown keys in subdocs (retain autocomplete suggestions) and use expect-type for type assertions

### LINKED ISSUE #16012: TypeScript: Regression in allowing Virtuals with Setters in Document creation/initialization after upgrading to v9

vkarpov15:
Fix #16012

**Summary**

Allows passing in extra keys to subdocuments while retaining autocomplete:

image

I decided to allow adding extra keys to subdocuments because adding the full virtuals to subdocuments would require more switches to `InferRawDocType`, `InferSchemaType`, etc. and that would be unwieldy. Allowing unknown fields while retaining autocomplete seems like a reasonable middle ground.

While working on this, I also found that our current `ExpectType` assertions don't handle `never` well: `never` is assignable to any type so `ExpectType<number>({} as never)` unexpectedly succeeds:

image

I wasn't able to come up with a reasonable workaround in ExpectType<> to correctly handle `never`, so I'm pulling in [expect-type npm package](https://www.npmjs.com/package/expect-type) to help with that. I made current `assertions.ts` a wrapper for expect-type for now to avoid massive diffs, but removing assertions.ts in favor of using expect-type directly is a candidate for future work.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -1102,7 +1102,7 @@ Query.prototype.projection = function(arg) {
  * @method select
  * @memberOf Query
  * @instance
- * @param {object|string|string[]} arg
+ * @param {object|string|Array<string>} arg
  * @return {Query} this
  * @see SchemaType https://mongoosejs.com/docs/api/schematype.html
  * @api public
@@ -3070,7 +3070,7 @@ Query.prototype.distinct = function(field, conditions, options) {
  *
  * Cannot be used with `distinct()`
  *
- * @param {object|string|(string | number)[][]} arg
+ * @param {object|string|Array<Array<(string | number)>>} arg
  * @param {object} [options]
  * @param {boolean} [options.override=false] If true, replace existing sort options with `arg`
  * @return {Query} this
@@ -4942,7 +4942,7 @@ Query.prototype._castUpdate = function _castUpdate(obj) {
  * a response for each query has also been returned, the results are passed to
  * the callback.
  *
- * @param {object|string|string[]} path either the path(s) to populate or an object specifying all parameters
+ * @param {object|string|Array<string>} path either the path(s) to populate or an object specifying all parameters
  * @param {object|string} [select] Field selection for the population query
  * @param {Model} [model] The model you wish to use for population. If not specified, populate will look up the model by the name in the Schema's `ref` field.
  * @param {object} [match] Conditions for the population query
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -20,6 +20,7 @@
   "type": "commonjs",
   "license": "MIT",
   "dependencies": {
+    "expect-type": "^1.3.0",
```
hasezoey: As copilot already pointed out, this should be a dev-dependency

test/types/create.test.ts
```diff
@@ -224,3 +225,66 @@ async function gh15902() {
...
+    toObject: { virtuals: true }
+  });
+
+  decimalSchema.virtual('value').set((value: number, _virtual, doc) => doc.set({ base: value, exponent: 1 }));
+
+  const ParentModel = mongoose.model('Gh16012', new mongoose.Schema({
+    quantity: decimalSchema
+  }));
+
+  // @ts-expect-error  Type 'number' does not satisfy the constraint '"Expected: number, Actual: never"'.
```
hasezoey:
```suggestion
  // @ts-expect-error Type 'number' does not satisfy the constraint '"Expected: number, Actual: never"'.
```

PR #16049, opened 2026-02-26
fix: avoid casting subdocs with query as context when running setDefaultsOnInsert

vkarpov15:
#16025 made it so that tests fail on 9.3 because we were creating subdocuments with a query instance as the parent. This PR makes it so that setDefaultsOnInsert explicitly pass the query as a `context` option, and renames it so that `schemaType.getDefault`'s first parameter is `parentDoc` to reflect the fact that it should be document. 

**Summary**

**Examples**

### DIFF
--- a/lib/schemaType.js
+++ b/lib/schemaType.js
@@ -33,9 +33,9 @@ const setOptionsForDefaults = { _skipMarkModified: true };
  *     const schema = new Schema({ name: String });
  *     schema.path('name') instanceof SchemaType; // true
  *
- * @param {String} path
+ * @param {string} path
  * @param {SchemaTypeOptions} [options] See [SchemaTypeOptions docs](https://mongoosejs.com/docs/api/schematypeoptions.html)
- * @param {String} [instance]
+ * @param {string} [instance]
  * @param {Schema} [parentSchema]
  * @api public
  */
@@ -215,9 +215,9 @@ SchemaType.prototype.validators;
 SchemaType.prototype.isRequired;
 
 /**
- * Split the current dottet path into segments
+ * Split the current dotted path into segments
  *
- * @return {String[]|undefined}
+ * @return {string[]|undefined}
  * @api private
  */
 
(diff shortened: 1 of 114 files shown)


lib/schemaType.js
```diff
@@ -1183,26 +1183,32 @@ SchemaType.prototype.ref = function(ref) {
 /**
  * Gets the default value
  *
- * @param {Object} scope the scope which callback are executed
+ * @param {Object} parentDoc
  * @param {Boolean} init
+ * @param {Object} options
  * @return {Any} The Stored default value.
```
hasezoey:
```suggestion
 * @param {object} options
 * @param {object} [options.context]
 * @param {boolean} [options.skipCast]
 * @return {any} The Stored default value.
```

AbdelrahmanHafez: Maybe instead it should be the following?
```suggestion
 * @param {object} options
 * @param {object} [options.context]
 * @param {boolean} [options.skipCast]
* @return {any} The Stored default value.
```

Is the uppercase `O` in `{Object} options` and `Any` intentional? @hasezoey

test/model.updateOne.test.js
```diff
@@ -624,6 +624,35 @@ describe('model: updateOne:', function() {
 
     });
 
+    it('setDefaultsOnInsert does not run setters on default values (gh-16025)', async function() {
```
AbdelrahmanHafez, line 627: I suspect this might cause some undesired behavior, for example if I have a setter that always removes special characters from strings, I'd expect that field to always not have special characters.

I'll look a bit deeper into this and finish my review today.

lib/schemaType.js
hasezoey: > Is the uppercase O in {Object} options and Any intentional? @hasezoey

At least for my part, i only added things, and let the other things stay the way they were (uppercase), only the new ones i added i had made lowercase. (though changed it anyway now)

test/model.updateOne.test.js
AbdelrahmanHafez: So turns out that this is a pre-existing issue that's unrelated to this change, I'll open a separate issue for it.

PR #16052, opened 2026-02-27
Fix issues related to defining timestamps and virtuals with methods and/or statics in schema options

### LINKED ISSUE #16046: TypeScript `this` Context Error When Calling Static Method on Schema Using `timestamps` and `SchemaOptions.virtuals`
### Prerequisites

### Mongoose version

9.2.2

### Node.js version

24.13.1

### MongoDB server version

8.2

5.9.3

### A TypeScript error occurs when calling a **static method** on a Mongoose model whose schema includes:

- `timestamps: true`
- a **virtual populate defined via `SchemaOptions.virtuals`**
- statics defined inside `SchemaOptions.statics`

The following TypeScript error occurs **when calling a static method**.  
The error originates from an incompatibility in the inferred **`this` context** of the static method.

```ts
The 'this' context of type
'Model<
  { placeholder?: unknown } & DefaultTimestampProps,
  {},
  {},
  { votes: unknown } & { id: string },
  Document<
    unknown,
    {},
    { placeholder?: unknown } & DefaultTimestampProps,
    { votes: unknown } & { id: string }
  >
>'
is not assignable to method's 'this' of type
'Model<
  { placeholder?: string | null | undefined },
  {},
  {},
  { votes: unknown },
  Document<
    unknown,
    {},
    { placeholder?: string | null | undefined },
    { votes: unknown } & { id: string },
    DefaultSchemaOptions
  >
>'.

The types of 'castObject(...).placeholder' are incompatible between these types.

Type 'unknown' is not assignable to type 'string | null | undefined'.
```

### Steps to Reproduce

```ts

const issueOneSchema = new mongoose.Schema(
  { placeholder: String },
  {
    timestamps: true,
    virtuals: {
      votes: {
        options: {
          ref: 'IssueTwo',
          localField: '_id',
          foreignField: 'issueOneId'
        }
      }
    },
    statics: {
      myStaticMethod: function () {
        console.log('placeholder');
      }
    }
  }
);

const IssueOne = mongoose.model('IssueOne', issueOneSchema);

IssueOne.myStaticMethod();
```

vkarpov15:
**Summary**

Fix #16046 

There's 2 issues fixed in this PR:

1. The original issue reported in #16046: if `timestamps`, `statics`, and `virtuals` are both defined in Schema constructor options, calling the static function causes TypeScript to fail with a context compatibility error due to DocType properties resolving as `unknown`. This was due to circular references - via trial and error I was able to break the circular reference by making virtuals rely primarily on `THydratedDocumentType` instead of `DocType`, which is more correct anyway: the `this` for virtual functions should be `THydratedDocumentType` not `HydratedDocument<DocType>`.
2. I wrote tests for the #16046 case for both `new Schema()` and `Schema.create()`, and also added a separate test that added `timestamps`, `virtuals`, `statics`, **and** `methods`. There was an issue with how `Schema.create()` was handling this case: because of how `TSchemaOptions` was inferred, the `this` value for methods was missing the default `id` virtual. I fixed this by explicitly omitting the `this` type in methods and statics when inferred at the function definition, because the `TSchemaOptions` context was inferring `this` before the full `THydratedDocumentType` is computed.

**Examples**

### DIFF
--- a/types/inferschematype.d.ts
+++ b/types/inferschematype.d.ts
@@ -119,6 +119,11 @@ declare module 'mongoose' {
     TypesAreEqual<T, DefaultSchemaOptions> extends true ? DefaultSchemaOptions :
     MergeType<DefaultSchemaOptions, T>;
 
+  /*!
+   * ResolveVirtuals resolves the virtuals option from the schema options.
+   * Adds the default `id` virtual if it is not disabled in schema options or
+   * overwritten in the RawDocType.
+   */
   type ResolveVirtuals<TSchemaOptions, RawDocType> =
     ResolveSchemaOptions<TSchemaOptions> extends { virtuals: infer V }
       ? AddDefaultId<RawDocType, V, ResolveSchemaOptions<TSchemaOptions>>


types/inferschematype.d.ts
```diff
@@ -119,6 +119,11 @@ declare module 'mongoose' {
     TypesAreEqual<T, DefaultSchemaOptions> extends true ? DefaultSchemaOptions :
     MergeType<DefaultSchemaOptions, T>;
 
+  type ResolveVirtuals<TSchemaOptions, RawDocType> =
```
hasezoey, line 127: Maybe consider a little documentation

test/types/schema.test.ts
```diff
@@ -2179,3 +2179,60 @@ function gh15915() {
...
+        }
+      },
+      methods: {
+        myMethod: function() {
+          ExpectType<string | null | undefined>(this.placeholder);
+        }
+      },
+      statics: {
+        myStaticMethod: function() {
+          console.log('placeholder', this.modelName);
```
hasezoey, line 2230: This should likely be a type assertion, as copilot pointed out.

hasezoey (commented):
LGTM

PR #16063, opened 2026-03-02
upgrade to eslint 10.x

vkarpov15:
**Summary**

Quick upgrade that required adding some additional packages, the dependabot PR failed. I had to pin @eslint/js to 9.x because 10.x has a bunch of backwards breaking changes that I didn't want to fix right now.

**Examples**


vkarpov15:
@AbdelrahmanHafez sure that would be helpful. Probably just disabling some of the new recommended rule changes.

PR #16066, opened 2026-03-04
chore: update type test setup

mrazauskas:
**Summary**

To draw your attention, #16048 added `expect-type` with a hope to handle `never` types. The assertions from `assertions.ts` were turned into wrappers for `expect-type` methods. Unfortunately that does not work with static analysis, because functions are not called. TypeScript only checks if arguments are assignable to parameter types, but does not pass them through the logic of the function. Try including `ExpectType<number>({} as never)` somewhere, it does not error.

I also noticed type errors in updated `assertions.ts`:

Screenshot 2026-03-04 at 09 17 09

Those were not caught by TSTyche, because it does not check `.ts` files. Assuming the `.ts` files are source files, most likely these are checked already. In contrary, `.d.ts` files are checked.

Based on above, I propose the following:

- rename `assertions.ts` to `assertions.d.ts` (it will be checked and will indicated its non-executable nature by disallowing function implementations)
- remove `expect-type` since it does nothing in current setup


test/types/util/assertions.ts
```diff
@@ -1,12 +0,0 @@
-import { expectTypeOf } from 'expect-type';
-
-export declare function Expect<T extends true>(): void;
```
mrazauskas, line 3: `Expect` is not used anywhere. So I removed it as well.

vkarpov15:
You're right. I didn't test sufficiently thoroughly, I saw `expectTypeOf({} as never).toEqualTypeOf<{} as number>()` worked as expected, but didn't test what happened when I moved it into a helper.

I presume what we need to do going forward is to switch to using `expect-type` directly? Making our current `ExpectType<>` handle `never` correctly has proven to be a deep rabbit hole.

PR #16067, opened 2026-03-04
chore: use TSTyche assertions

mrazauskas:
Part of #15980

**Summary**

This PR migrates type test to TSTyche assertions in the following files:

```
test/types/aggregate.test.ts
test/types/base.test.ts
test/types/connect.test.ts
test/types/connection.test.ts
test/types/create.test.ts
test/types/discriminator.test.ts
test/types/docArray.test.ts
test/types/esm.test.ts
test/types/helpers.test.ts
```

The key difference from similar libraries: the TSTyche assertions *always* check types programmatically. This means that all flavours of `any` and `never` can be detected and reported. (I honestly admire the effort to develop generic type assertions, but even the most advanced implementations like `expect-type` cannot detect all `any`s correctly ([playground](https://tsplay.dev/Wo2zem)).)

I must agree that generic assertions with instant response are attractive while writing tests. Although reading their errors in CI isn’t the greatest experience. With TSTyche I went the other direction: no limitations of generics and clear errors in CI, but no instant red squiggles in IDE. (Same as with any JavaScript test runner, by the way.)

Another aspect is TypeScript 7 support. Currently it is not possible to compare types programatically using TS7, because there is no programmatic API yet. Also passing file with TSTyche assertions through `tsgo` does nothing, because the assertions are not generic.

In the other hand, as you already know TSTyche checks errors suppressed by `@ts-expect-error`. This PR makes a steps forward by replacing the `@ts-expect-error` directives with `.not.toHaveProperty()`, `.not.toBeCallableWith()` and `.not.toBeConstructableWith()` matchers. (The last two are only available in TSTyche.)

Please look through. If you are happy with this change, I can continue migrating test in batches.


test/types/connection.test.ts
```diff
@@ -1,95 +1,87 @@
...
-ExpectType<Promise<Connection>>(conn.openUri('mongodb://127.0.0.1:27017/test', { bufferCommands: true }));
+expect(conn.openUri('mongodb://127.0.0.1:27017/test')).type.toBe<Promise<Connection>>();
+expect(conn.openUri('mongodb://127.0.0.1:27017/test', { bufferCommands: true })).type.toBe<Promise<Connection>>();
 
 conn.readyState === 0;
 conn.readyState === 99;
 
-// @ts-expect-error  Cannot assign to 'readyState' because it is a read-only property.
-conn.readyState = 0;
+expect(pick(conn, 'readyState')).type.toBe<{ readonly readyState: ConnectionStates }>();
```
mrazauskas, line 21: The shape of the type is tested instead of triggering an error. A `readonly` property is not assignable. The test would fail if property would not be marked `readonly`.

vkarpov15:
It is a bit of a bummer that the assertions don't work with TS7, but given that we're not actively testing against TS7 right now I think this is a reasonable direction to take. The fact that our current `ExpectType` helper struggles with `never` is a more immediate problem.

PR #16068, opened 2026-03-04
chore: use TSTyche assertions in the `document` type test

mrazauskas:
**Summary**

Same as #16067, but with some mismatch of expected types. Therefore, I think this needs an extra attention (see inline comments).


test/types/document.test.ts
```diff
@@ -177,7 +174,7 @@ function gh11435() {
   const ItemSchema = new Schema<Item>({ name: String });
 
   ItemSchema.pre('validate', function preValidate() {
-    ExpectType<Model<unknown>>(this.$model('Item1'));
+    expect(this.$model('Item1')).type.toBe<unknown>();
```
mrazauskas: `Model<unknown>` does not pass. The type seems to be plain `unknown`. Is that correct?

test/types/document.test.ts
```diff
@@ -306,15 +303,15 @@ function gh13878() {
   });
   const User = model('User', schema);
   const user = new User({ name: 'John', age: 30 });
-  ExpectType<typeof User>(user.$model());
-  ExpectType<typeof User>(user.model());
+  expect(user.$model()).type.toBe<unknown>();
+  expect(user.model()).type.toBe<unknown>();
```
mrazauskas: `typeof User` or simply `.type.toBe(User)` does not pass. The type seems to be plain `unknown`. Is that correct?

test/types/document.test.ts
vkarpov15: Probably some type widening quirk - `$model` uses generics. I think we could just expect assignable here, or test `const model: Model<unknown> = this.$model('Item1')`, what do you think? I don't like expecting that the type is strictly `unknown` here

test/types/document.test.ts
vkarpov15: Similar type widening. Maybe we just check assignability?

mrazauskas:
@vkarpov15 Thanks for feedback.

I took a better look and found out the following. The argument of `expect()` is typed as `unknown` and both `Document#$model()` and `Document#model()` methods do not type argument constraints. Looks like TypeScript decides to pass `unknown` as type argument (since that is allowed). The resulting type becomes `unknown`.

Two possible solutions:

1. Move the function call outside of the `expect`:

   ```ts
   const x = this.$model('Item1');
   expect(x).type.toBe<Model<unknown>>(); // pass
   ```

2. Or add constraints to both signatures and the resulting type will be inferred correctly. See #16081

What sounds better for you?

---

Another detail. Right now `ReturnType<typeof this.$model>` is inferred as `unknown`. After adding the constraints (as suggested in #16081) it becomes `Model<DocType>` (the return type of last overload).

I try to show that other unrelated APIs infer the `unknown` as well. TSTyche does not use `ReturnType` anyhow.

vkarpov15:
I merged #16081

mrazauskas:
> I merged #16081

Thanks! Seems like all is working as expect. The `unknown`s are gone. See updated tests in https://github.com/Automattic/mongoose/pull/16068/commits/d5cfb0a4d3fc485b77bfd2b6137dd74d40b0e7fd

PR #16069, opened 2026-03-04
feat(schema): support `discriminators` option inline for better TypeScript support

vkarpov15:
**Summary**

Our current pattern of declaring embedded discriminators doesn't work very well with TypeScript: there's no way for TypeScript to infer that subdocuments with `kind === 'Circle'` have a `radius` property in the following example.

```ts
const schema = Schema({ shape: shapeSchema });

// For TypeScript, use `schema.path<Schema.Types.Subdocument>('shape').discriminator(...)`
schema.path('shape').discriminator('Circle', Schema({ radius: String }));
schema.path('shape').discriminator('Square', Schema({ side: Number }));
```

This PR adds a new `discriminators` schematype option and allows TypeScript to infer the discriminator types as an enum from the following:

```ts
    const schema = new Schema({
      shape: {
        type: shapeSchema,
        discriminators: {
          Circle: circleSchema,
          Square: squareSchema
        }
      }
    });
```

**Examples**

### DIFF
--- a/types/schematypes.d.ts
+++ b/types/schematypes.d.ts
@@ -213,7 +213,7 @@ declare module 'mongoose' {
     _id?: boolean;
 
     /** Embedded discriminators for this path. Only allowed for subdocuments and subdocument arrays. */
-    discriminators?: Record<string, Schema<any, any, any>>;
+    discriminators?: Record<string, Schema<unknown>>;
 
     /** If set, specifies the type of this map's values. Mongoose will cast this map's values to the given type. */
     of?: Function | SchemaDefinitionProperty<any> | (Function | SchemaDefinitionProperty<any>)[];


types/schematypes.d.ts
```diff
@@ -212,6 +212,9 @@ declare module 'mongoose' {
     /** If set, overrides the child schema's `_id` option. Only allowed for subdocuments and subdocument arrays. */
     _id?: boolean;
 
+    /** Embedded discriminators for this path. Only allowed for subdocuments and subdocument arrays. */
+    discriminators?: Record<string, Schema<any, any, any>>;
```
AbdelrahmanHafez: Seeing three `any`s here looks like something that can be improved.
@copilot are there any types improvements we can make here?

test/types/schema.test.ts
```diff
@@ -2237,3 +2237,116 @@ function gh16046() {
...
+    |(Omit<InferSchemaType<typeof shapeSchema>, keyof InferSchemaType<typeof circleSchema>> & InferSchemaType<typeof circleSchema>)
+      | (Omit<InferSchemaType<typeof shapeSchema>, keyof InferSchemaType<typeof squareSchema>> & InferSchemaType<typeof squareSchema>)
+      >({} as ShapeElementType);
+
+  const Model = model('EmbeddedDiscriminatorPathOptionDocArray', schema);
+  const circleDoc = new Model({ shapes: [{ kind: 'Circle', radius: 5, name: { baseName: 'a' } }] });
+  const squareDoc = new Model({ shapes: [{ kind: 'Square', side: 10, name: 'b' }] });
+
+  for (const doc of [circleDoc, squareDoc]) {
+    for (const shape of doc.shapes ?? []) {
```
hasezoey:
```suggestion
    for (const shape of doc.shapes!) {
```

I think this would be more concise.

types/schematypes.d.ts
vkarpov15: I just made it `Schema<unknown>` in [e94e3d7](https://github.com/Automattic/mongoose/pull/16069/commits/e94e3d7bfeb9a68aa150019d6dcbd05f164108c8), that's marginally better but I don't think there's any way to further type the discriminator schemas here.

vkarpov15:
@hasezoey looks like we added support for this at runtime in #11600, but correct not well documented.

PR #16077, opened 2026-03-08
types(InferRawDocType): fall back to using InferRawDocType instead of pulling non-raw inferred doc type if EnforcedDocType not set

### LINKED ISSUE #16053: InferRawDocTypeFromSchema missing `_id`
### Prerequisites

### Mongoose version

9.2.3

### Node.js version

22.14

### MongoDB server version

8.2.3

5.7.3

### Description

I expect the `InferRawDocTypeFromSchema` type to be the same as what's returned from `.lean()`. As far as I understand it, it should result in the same output as `InferRawDocType` which does have `_id`.

### Steps to Reproduce

```js
const schemaDefinition = new Schema({
  email : { 
    type : String ,
    trim: true,
    required : true,
    unique: true,
    lowercase: true
  },
  password : {
    type : String ,
    required : true,
  },
  dateOfBirth: {
    type: Date,
    required: true
  }
})

const TestModel = mongoose.model('Test', schemaDefinition);
type TestType = mongoose.InferRawDocTypeFromSchema<typeof schemaDefinition>;

let test: TestType | null;

async function queryById(id: string ){
    let test: TestType | null ;
    TestModel.findById(id).lean().then(result=>{
        if (result){
            test = result
            console.log(result._id)
            console.log(test._id) // This throws a type error since _id doesn't exist on the type when it should
        }
    })
}
```

### Expected Behavior

I would expect the _id to exist here since it's technically part of the schema. I would expect it to behave the same as `InferRawDocType`.

vkarpov15:
Fix #16053

**Summary**

`InferRawDocTypeFromSchema` does not actually use `InferRawDocType` as currently written, which leads to inconsistencies like #16053. I suspect the issue is that we wrote `InferRawDocTypeFromSchema` before we had a `TSchemaDefinition` generic on schemas, so we couldn't use that. But now, using `TSchemaDefinition` to infer the raw doc type makes a lot more sense.

**Examples**

### DIFF
--- a/test/types/inferrawdoctype.test.ts
+++ b/test/types/inferrawdoctype.test.ts
@@ -442,7 +442,7 @@ async function gh16053() {
   const schemaDefinition = new mongoose.Schema({
     email: {
       type: String,
-      required: true,
+      required: true
     },
     password: {
       type: String,


vkarpov15:
Fixes from #16076 also ended up in here due to some merging, please disregard those changes when reviewing since those changes should already be in master and 9.3.

test/types/inferrawdoctype.test.ts
```diff
@@ -437,3 +437,35 @@ function gh15988() {
...
 }
+
+async function gh16053() {
+  const schemaDefinition = new mongoose.Schema({
+    email: {
+      type: String,
+      trim: true,
+      required: true,
+      unique: true,
+      lowercase: true
```
hasezoey:
```suggestion
      type: String,
      required: true,
```

To my knowledge, those options dont actually change the types.

hasezoey (approved):

PR #16081, opened 2026-03-10
fix: add type constraints for `Document#model()`

mrazauskas:
**Summary**

This PR adds type constraints for type arguments of `Document#$model()` and `Document#model()` methods.

Currently, the type arguments have default types only. This means users can pass each and every type (`unknown`, `boolean`, etc.) as a type argument. That is not a problem, but it feels like replacing the default with constraints can slightly improve the behaviour.


vkarpov15 (approved):
I think this is a reasonable change, we shouldn't allow `model<string>('FooModel')` for instance.

PR #16083, opened 2026-03-10
chore: use TSTyche assertions

mrazauskas:
**Summary**

Following up #16067.

This PR migrates type test to TSTyche assertions in the following files:

```
test/types/lean.test.ts
test/types/loadclass.test.ts
test/types/maps.test.ts
test/types/mongo.test.ts
test/types/populate.test.ts
test/types/sanitizeFilter.test.ts
test/types/schemaTypeOptions.test.ts
test/types/subdocuments.test.ts
test/types/utility.test.ts
test/types/virtuals.test.ts
```

Similar like in previous PR, the `@ts-expect-error` directives are replaced with `.not.toHaveProperty()` matchers. Also type casting like `{} as X` became unnecessary, because two types can be compared directly. For example:

```ts
expect<MergeType<A, B>['c']>().type.toBe<number>();
```

### DIFF
--- a/test/types/virtuals.test.ts
+++ b/test/types/virtuals.test.ts
@@ -91,11 +91,13 @@ async function autoTypedVirtuals() {
   type AutoTypedSchemaType = InferSchemaType<typeof testSchema>;
   type VirtualsType = { domain: string } & { id: string };
   type InferredDocType = AutoTypedSchemaType & ObtainSchemaGeneric<typeof testSchema, 'TVirtuals'>;
+  // Domain is unknown because of circular dep: this is the type of virtuals in the domain getter and setter
+  type ThisVirtualsType = { domain: unknown } & { id: string };
   type HydratedDocType = mongoose.HydratedDocument<
     AutoTypedSchemaType,
-    VirtualsType,
+    ThisVirtualsType,
     {},
-    VirtualsType
+    ThisVirtualsType
   >;
 
   const testSchema = new Schema({


test/types/subdocuments.test.ts
```diff
@@ -144,8 +144,8 @@ function gh14601() {
   });
 
   const obj = item.toObject();
+  expect(obj).type.toBeAssignableTo<IMain>();
 
   const obj2 = item.f2.toObject();
-
-  ExpectAssignable<{ _id: Types.ObjectId, field1: string }>()(obj2);
+  expect(obj2).type.toBeAssignableTo<ISub>();
```
mrazauskas, line 150: Is it fine, or did I over simplify?

test/types/utility.test.ts
```diff
@@ -43,5 +43,4 @@ type NestedPaths = WithLevel1NestedPaths<SchemaWithPopulatedRef>;
 
 // The refField type should be a union that includes RefSchema and ObjectId
 // This ensures that PopulatedDoc fields can be queried with both the populated document and the raw ID
-ExpectAssignable<NestedPaths['refField']>()({ name: 'test' } as RefSchema);
-ExpectAssignable<NestedPaths['refField']>()(new Types.ObjectId());
+expect<NestedPaths['refField']>().type.toBe<RefSchema | Types.ObjectId>();
```
mrazauskas, line 46: I adjusted the expected type to be exactly the same as described in the comment above.

test/types/virtuals.test.ts
```diff
@@ -101,25 +101,24 @@ async function autoTypedVirtuals() {
     virtuals: {
       domain: {
         get() {
-          ExpectType<Document<any, any, { email: string }> & AutoTypedSchemaType>(this);
+          expect(this).type.toBe<Document<unknown, {}, { email: string }> & AutoTypedSchemaType & { _id: Types.ObjectId, __v: number, domain: unknown, id: string }>();
```
mrazauskas: This and the following were puzzling cases. `this.domain` is defined on the type, but `ExpectType` does not require the `domain` property.

vkarpov15: The problem is that the type of `domain` isn't known until the return type of this function is resolved - this is the getter function for `domain` :)

test/types/subdocuments.test.ts
vkarpov15: Nope this is fine.

PR #16084, opened 2026-03-10
Make typescript fail when providing unknown query operator

vkarpov15:
**Summary**

Right now providing an unknown query operator like `$wrong` passes in TypeScript, but throws a runtime error. 937eeb1 fixes a related issue where the error message is inconsistent for numbers vs other schematypes.

269afd9 updates it so Mongoose doesn't use MongoDB Node driver's `Condition` type, which intentionally allows string signature. Instead, we strip out the string signature so only known query operators are allowed. 

**Examples**

### DIFF
--- a/lib/schema/number.js
+++ b/lib/schema/number.js
@@ -473,7 +473,7 @@ SchemaNumber.prototype.castForQuery = function($conditional, val, context) {
   if ($conditional != null) {
     handler = this.$conditionalHandlers[$conditional];
     if (!handler) {
-      throw new CastError('number', val, this.path, null, this);
+      throw new Error('Can\'t use ' + $conditional + ' with Number.');
     }
     return handler.call(this, val, context);
   }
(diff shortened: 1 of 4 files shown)


lib/schema/number.js
```diff
@@ -473,7 +473,7 @@ SchemaNumber.prototype.castForQuery = function($conditional, val, context) {
   if ($conditional != null) {
     handler = this.$conditionalHandlers[$conditional];
     if (!handler) {
-      throw new CastError('number', val, this.path, null, this);
+      throw new Error('Can\'t use ' + $conditional + ' with Number.');
```
AbdelrahmanHafez:
```suggestion
      throw new MongooseError('Can\'t use ' + $conditional + ' with Number.');
```

AbdelrahmanHafez (commented):
One suggestion, otherwise LGTM

hasezoey (commented):
LGTM

PR #16089, opened 2026-03-12
types(populate): retain populated paths in toObject() and toJSON() unless depopulate: true set

### LINKED ISSUE #16085: `toObject` does not keep populated types
### Prerequisites

### Mongoose version

9.2.2

### Node.js version

24.13.1

### MongoDB server version

n/a

5.8.3

### Description

calling `toObject` after `populate` does not keep the populated type in the plain object type

### Steps to Reproduce

```ts
const childSchema = new Schema({
    name: String
});
const Child = mongoose.model('Child', childSchema);

    title: String,
    children: [{ type: Schema.Types.ObjectId, ref: 'Child' }]
});
const Parent = mongoose.model('Parent', parentSchema);

async function reproduce() {
    const parentDoc = new Parent({ title: 'My Parent', children: [] });

    const populatedDoc = await parentDoc.populate<{
        children: Types.DocumentArray<HydratedDocFromModel<typeof Child>>;
    }>('children');

    const childNameFromHydrated = populatedDoc.children[0]?.name; // OK

    const plainObject = populatedDoc.toObject();
    const childNameFromPlain = plainObject.children[0].name; // Property 'name' does not exist on type 'ObjectId'. ts(2339)
}
```

### Expected Behavior

_No response_

vkarpov15:
Fix #16085

**Summary**

The populate typing fix now preserves populated path types through `toObject()` / `toJSON()` for document, query, and model `populate()` calls, while still returning the original raw path types for `toObject({ depopulate: true })`. 

The primary change is a new `PopulateDocumentResult` type that returns a document with a `mongoosePopulatedDocumentMarker` marker type that stores the populated type and depopulated type. I switched to using a marker for performance reasons - the previous type approach lead to too many instantiations and made our ts benchmark blow up. `toObject()` and `toJSON()` have overrides that detect `mongoosePopulatedDocumentMarker` on `this` and pull the types from the marker.

**Examples**

### DIFF
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -453,6 +453,37 @@ function gh14441() {
       ExpectType<string>(plainObject.children[0].name);
       const depopulatedObject = populatedDoc.toObject({ depopulate: true });
       ExpectType<Types.ObjectId>(depopulatedObject.children![0]);
+
+      const objectWithVirtuals = populatedDoc.toObject({ virtuals: true });
+      ExpectType<string>(objectWithVirtuals.children![0].name);
+
+      const objectWithFlattenObjectIds = populatedDoc.toObject({ flattenObjectIds: true });
+      ExpectType<string>(objectWithFlattenObjectIds.children![0].name);
+
+      const depopulatedAndFlattened = populatedDoc.toObject({ depopulate: true, flattenObjectIds: true });
+      ExpectType<string>(depopulatedAndFlattened.children![0]);
+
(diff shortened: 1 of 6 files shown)


hasezoey (approved):
LGTM, maybe consider Copilot's suggestion of also adding `toJSON` tests.

vkarpov15:
I reworked this PR to address the benchmark issues, changed to use a marker type instead. Please re-review.

hasezoey (commented):
LGTM

PR #16092, opened 2026-03-13
perf(projection): split by space instead of regex in parseProjection

### LINKED ISSUE #16087: Make parseProjection only split by spaces
### Prerequisites

### Issue

Splitting by `' '` is about 5x faster than splitting by `/\s+/`:

Image

Further, checking whether a string contains a space is 10x faster than checking if a string contains any whitespace, so our usual optimization of "check if split will do anything before splitting" would be significantly faster if we could just use spaces.

Image

mahmoodhamdi:
## Problem

`parseProjection` currently uses `v.split(/\s+/)` to split string projections. Splitting by regex is significantly slower than splitting by a plain string.

As noted in #16087, splitting by `' '` is ~5x faster than splitting by `/\s+/`.

## Solution

Changed `v.split(/\s+/)` to `v.split(' ')` in `parseProjection`.

This is a safe change because the existing `if (!field) continue` guard on line 22 already handles empty strings that may result from consecutive spaces. In practice, projection strings use regular spaces as delimiters, so there's no need to match tabs or other whitespace characters.

## Testing

All existing tests pass. Verified the following cases manually:

```js
parseProjection('name age email')       // { name: 1, age: 1, email: 1 }
parseProjection('name  age   email')    // { name: 1, age: 1, email: 1 } (consecutive spaces)
parseProjection(' name age ')           // { name: 1, age: 1 } (leading/trailing spaces)
parseProjection('-name age')            // { name: 0, age: 1 } (exclusion)
parseProjection('-name age', true)      // { '-name': 0, age: 1 } (retainMinusPaths)
```

Fixes #16087

### DIFF
--- a/lib/helpers/projection/parseProjection.js
+++ b/lib/helpers/projection/parseProjection.js
@@ -9,7 +9,7 @@ module.exports = function parseProjection(v, retainMinusPaths) {
   const type = typeof v;
 
   if (type === 'string') {
-    v = v.split(/\s+/);
+    v = v.split(' ');
   }
   if (!Array.isArray(v) && Object.prototype.toString.call(v) !== '[object Arguments]') {
     return v;


mahmoodhamdi:
Thanks for the review @AbdelrahmanHafez! That makes sense — I can see how changing from `/\s+/` to `' '` could technically be considered a breaking change, even though in practice projection strings only ever use spaces.

Happy to target this for 10.0 whenever the timing is right. Just let me know if you'd like me to rebase onto a different branch when one is created, or if there's anything else I should adjust.

lib/helpers/projection/parseProjection.js
```diff
@@ -9,7 +9,7 @@ module.exports = function parseProjection(v, retainMinusPaths) {
   const type = typeof v;
 
   if (type === 'string') {
```
vkarpov15, line 11: Let's also add a check here `indexOf(' ') !== -1`

vkarpov15 (changes requested):
one minor comment but looks good overall. As @AbdelrahmanHafez said this is a breaking change, we'll merge this into 10.0

PR #16093, opened 2026-03-13
fix(setDefaultsOnInsert): run setters on default values during upsert

### LINKED ISSUE #16051: [Bug] setters do not run on setDefaultsOnInsert
### Prerequisites

### Issue

re https://github.com/Automattic/mongoose/pull/16049#discussion_r2864733843

Setters should run with `setDefaultsOnInsert`

```js
'use strict';

const { Schema } = mongoose;

run().catch(err => {
  console.error(err);
  process.exit(1);
});

async function run() {
  const mongod = await MongoMemoryServer.create();

  try {
    await mongoose.connect(mongod.getUri(), { dbName: 'test' });

    const schema = new Schema({
      name: String,
      slug: {
        type: String,
        default: 'Hello World!',
        set: function(v) {
          return v.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/(^-|-$)/g, '');
        }
      }
    });

    // Normal document creation: setter runs on default
    const doc = new Test({ name: 'test' });
    assert.equal(doc.slug, 'hello-world');

    // Upsert: setter does NOT run on default
    await Test.updateOne(
      { name: 'test2' },
      { $set: { name: 'test2' } },
      { upsert: true, setDefaultsOnInsert: true }
    );
    const upserted = await Test.findOne({ name: 'test2' });
    assert.equal(upserted.slug, 'hello-world');

    console.log('PASS');
  } finally {
    await mongoose.disconnect();
    await mongod.stop();
  }
}
```

mahmoodhamdi:
## Problem

When using `setDefaultsOnInsert` with upsert operations, default values are not passed through setters. This means a schema like:

```js
const schema = new Schema({
  name: String,
  slug: {
    type: String,
    default: 'Hello World!',
    set: function(v) {
      return v.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/(^-|-$)/g, '');
    }
  }
});
```

will store `slug` as `'Hello World!'` when upserted, instead of `'hello-world'` (what you get with normal document creation).

## Root Cause

In `lib/helpers/setDefaultsOnInsert.js`, `schemaType.getDefault()` was called with `init = true`. This causes `_applySetters` to skip all setters, since `init` is meant to signal that the value is being initialized and shouldn't go through setter transformations.

## Solution

Changed `init` from `true` to `false` in the `getDefault()` call so setters run on default values, matching the behavior of normal document creation via `new Model()`.

Also updated the existing test from gh-16025 which was asserting the old (buggy) behavior, and added a unit test in `test/helpers/setDefaultsOnInsert.test.js`.

## How to Test

```js
const schema = new Schema({
  name: String,
  slug: {
    type: String,
    default: 'Hello World!',
    set: v => v.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/(^-|-$)/g, '')
  }
});

await Test.updateOne(
  { name: 'test' },
  { $set: { name: 'test' } },
  { upsert: true, setDefaultsOnInsert: true }
);

const doc = await Test.findOne({ name: 'test' });
assert.equal(doc.slug, 'hello-world'); // was 'Hello World!' before this fix
```

Fixes #16051

### DIFF
--- a/lib/helpers/setDefaultsOnInsert.js
+++ b/lib/helpers/setDefaultsOnInsert.js
@@ -84,7 +84,7 @@ module.exports = function(filter, schema, castedDoc, options, queryMongooseOptio
     if (schemaType.path === '_id' && schemaType.options.auto) {
       return;
     }
-    const def = schemaType.getDefault(null, true, { context });
+    const def = schemaType.getDefault(null, false, { context });
     if (typeof def === 'undefined') {
       return;
     }
--- a/test/helpers/setDefaultsOnInsert.test.js
+++ b/test/helpers/setDefaultsOnInsert.test.js
@@ -107,6 +107,25 @@ describe('setDefaultsOnInsert', function() {
     assert.equal(update.$setOnInsert['referralConversion'], 'bar');
   });
 
+  it('runs setters on default values (gh-16051)', function() {
+    const schema = new Schema({
+      name: String,
+      slug: {
+        type: String,
+        default: 'Hello World!',
+        set: function(v) {
+          return v.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/(^-|-$)/g, '');
+        }
+      }
+    });
+
+    const opts = { upsert: true, setDefaultsOnInsert: true };
+    let update = { $set: { name: 'test' } };
+    update = setDefaultsOnInsert({}, schema, update, opts);
+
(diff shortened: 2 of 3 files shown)


vkarpov15 (approved):
LGTM although for caution I'll merge into 9.4. I don't expect this to be backwards breaking because setters already run on non-default values so I think it is reasonable to assume that setters already work on queries.

PR #16094, opened 2026-03-13
fix(document): handle nested paths in optimisticConcurrency include/exclude

### LINKED ISSUE #16054: [Bug]
### Prerequisites

### Issue

`optimisticConcurrency` option needs to handle nested paths properly, it currently doesn't.

```js
const profileSchema = new Schema({ firstName: String, lastName: String }, { _id: false });
  profile: profileSchema,
  balance: Number
}, { optimisticConcurrency: { exclude: ['profile'] } });

const user = await User.create({ profile: { firstName: 'Alice', lastName: 'Smith' }, balance: 100 });

user.profile.firstName = 'Bob';
user.$__delta();

// BUG: user.$__.version === VERSION_ALL (3)
// EXPECTED: user.$__.version === undefined (profile is excluded, so profile.firstName should be too)
```

mahmoodhamdi:
## Problem

When using `optimisticConcurrency` with include arrays or exclude objects, nested paths are not handled properly.

**Exclude case:** Excluding `profile` does not exclude `profile.firstName`. Modifying `profile.firstName` still triggers versioning even though its parent is excluded.

```js
  profile: profileSchema,
  balance: Number
}, { optimisticConcurrency: { exclude: ['profile'] } });

user.profile.firstName = 'Bob';
user.$__delta();
// BUG: user.$__.version === VERSION_ALL
// EXPECTED: user.$__.version === undefined
```

**Include case:** Including `profile.address.country` does not trigger versioning when setting `profile.address` (the parent).

## Root Cause

The existing code uses `Set.has()` for exact path matching, which doesn't account for parent-child path relationships.

## Solution

Added two helper functions:

- `_isExcludedByOptCon(path, excludeSet)`: a path is excluded if it matches exactly OR any of its parent prefixes is in the exclude set (e.g. `profile` excludes `profile.firstName`)
- `_isIncludedByOptCon(path, includeSet)`: a path is included if it matches exactly, OR any parent prefix is included, OR any included path is a child of the modified path (e.g. modifying `profile.address` triggers when `profile.address.country` is included)

## Testing

Added 5 new test cases in `test/versioning.test.js` covering:
1. Exclude: parent exclusion cascades to children
2. Exclude: non-excluded paths still trigger version alongside excluded children
3. Include: parent inclusion cascades to children
4. Include: modifying parent triggers when child is included
5. Include: non-included nested paths don't trigger

All 22 optimisticConcurrency tests pass.

Fixes #16054

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -5055,8 +5055,8 @@ Document.prototype.$__delta = function $__delta(pathsToSave, pathsToSaveSet) {
       const optCon = new Set(optimisticConcurrency);
       const modPaths = this.modifiedPaths();
       const hasRelevantModPaths = pathsToSave == null ?
-        modPaths.find(path => optCon.has(path)) :
-        modPaths.find(path => optCon.has(path) && isInPathsToSave(path, pathsToSaveSet, pathsToSave));
+        modPaths.find(path => _isIncludedByOptCon(path, optCon)) :
+        modPaths.find(path => _isIncludedByOptCon(path, optCon) && isInPathsToSave(path, pathsToSaveSet, pathsToSave));
       if (hasRelevantModPaths) {
         this.$__.version = dirty.length ? VERSION_ALL : VERSION_WHERE;
       }
@@ -5064,8 +5064,8 @@ Document.prototype.$__delta = function $__delta(pathsToSave, pathsToSaveSet) {
       const excluded = new Set(optimisticConcurrency.exclude);
       const modPaths = this.modifiedPaths();
       const hasRelevantModPaths = pathsToSave == null ?
-        modPaths.find(path => !excluded.has(path)) :
-        modPaths.find(path => !excluded.has(path) && isInPathsToSave(path, pathsToSaveSet, pathsToSave));
+        modPaths.find(path => !_isExcludedByOptCon(path, excluded)) :
+        modPaths.find(path => !_isExcludedByOptCon(path, excluded) && isInPathsToSave(path, pathsToSaveSet, pathsToSave));
       if (hasRelevantModPaths) {
         this.$__.version = dirty.length ? VERSION_ALL : VERSION_WHERE;
       }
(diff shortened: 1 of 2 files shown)


lib/document.js
```diff
@@ -5588,6 +5588,56 @@ Document.prototype._applyVersionIncrement = function _applyVersionIncrement() {
...
+ * Check if a path is included by an optimisticConcurrency include list,
+ * handling nested paths in both directions.
+ */
+
+function _isIncludedByOptCon(path, includeSet) {
+  if (includeSet.has(path)) {
+    return true;
+  }
+  // Check if any parent prefix is included (e.g. 'profile' includes 'profile.name')
+  const pieces = path.split('.');
```
vkarpov15, line 5601: Need to make sure to check if path contains `'.'` before splitting for performance

lib/document.js
```diff
@@ -5588,6 +5588,56 @@ Document.prototype._applyVersionIncrement = function _applyVersionIncrement() {
   }
 };
 
+/*!
+ * Check if a path is included by an optimisticConcurrency include list,
+ * handling nested paths in both directions.
+ */
+
+function _isIncludedByOptCon(path, includeSet) {
```
vkarpov15, line 5596: I'd like to see if there is a way to reuse split values in this function similar to `hasIncludedChildren` for projections in `lib/document.js`. This check is very expensive - lots of `split()` calls, loops. It would be ideal to avoid this by precomputing the split values from `this.$__schema.options.optimisticConcurrency`

mahmoodhamdi:
Thanks for the review @vkarpov15\! Good points on the performance — the repeated `split('.')` calls and O(N) scans would add up on large documents. I'll rework this to precompute the split values from `this.$__schema.options.optimisticConcurrency` up front (similar to how `hasIncludedChildren` handles projections) and add the `'.'` check before splitting. Will open a new PR with the fixes.

Issue #16098, opened 2026-03-13
`Model.useConnection()` lacks validation for Connection instance, causing silent state corruption


mrsandy1965:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

9.3.0

### Node.js version

20.x

### MongoDB server version

7.x

### Description

The guard only checks `if (!connection)`, but passes any truthy value through. Passing a plain object, a string, or an unrelated class instance causes confusing downstream errors when `connection.collection(...)` or `connection.models[...]` is accessed.

### Steps to Reproduce

```javascript
const Model = mongoose.model('Test', new Schema({ name: String }));
// Silently succeeds but corrupts state
Model.useConnection({ not: 'a connection' }); 
// Downstream error
try {
  await Model.findOne();
} catch (err) {
  console.log(err); // TypeError: connection.collection is not a function
}

### Expected Behavior

`Model.useConnection()` should throw a `TypeError` if the provided argument is not an instance of `mongoose.Connection`

chakri11-git:
Hi, I'd like to work on this issue!

My proposed fix:
- Find the `useConnection()` method in `lib/model.js`
- Add an `instanceof mongoose.Connection` check
- Throw a `TypeError` with a clear message if the argument 
  is not a valid Connection instance

Something like:
if (!(connection instanceof mongoose.Connection)) {
  throw new TypeError(
    '`connection` argument must be an instance of mongoose.Connection, ' +
    'got "' + typeof connection + '"'
  );
}

Would this approach work? Happy to also add a test case for this in the test suite.

Issue #10404, opened 2021-06-29
this.$__isSelected is not a function error when timestamps set to true in the Mongoose Schema


LawP09:
Good morning.
I have exactly the same issue as @soubhikchatterjee with mongoose 5.13.0, NodeJS 14.17.1

Schema creation : 
```import { Schema } from 'mongoose';

export const receiptBatchImportRequestLockSchema: Schema<IReceiptBatchRequestLock> = new Schema(
  {
    _id: String
  },
  { timestamps: true }
);

export interface IReceiptBatchRequestLock {
  _id: string;
  createdAt?: Date;
  updatedAt?: Date;
}
```

Model creation : 
``` 
import * as mongoose from 'mongoose';

private db: mongoose.Connection;
private receiptBatchImportRequestLockModel: mongoose.Model<IReceiptBatchRequestLock>;

this.receiptBatchImportRequestLockModel = this.db.model<IReceiptBatchRequestLock>(
    'receiptBatchImportRequestLock',
    receiptBatchImportRequestLockSchema,
    'receiptBatchImportRequestLock'
);
```

Save in MongoDB (v4.2) : 
```
  public async createReceiptBatchImportRequestLock(lock: IReceiptBatchRequestLock, session: mongoose.ClientSession) {
    logger.info('#createReceiptBatchImportRequestLock ...');
    try {
      this.initialize();
      const result = new this.receiptBatchImportRequestLockModel(lock);
      await result.save({ session });
      return result;
    } catch (error) {
      logger.error({ error }, 'Error when creating a Receipt Batch Import Request Lock.');
      throw error;
    }
  }
``` 

- When I put `{ timestamps: false }` it works correctly and don't save createdAt and updatedAt fields

- When I remove `this.$__isSelected(createdAt)` from the setupTimestamps.js file it works correctly and save createdAt and updatedAt fields into the DB.

What the $__isSelected function is supposed to do. I tried to debug but I found nothing.

Thanks in advance for your investigation.

_Originally posted by @LawP09 in https://github.com/Automattic/mongoose/issues/10259#issuecomment-870676547_

LawP09:
I downgraded the mongoose version to 5.9.29 and it works well.

IslandRhythms:
it seems like you're dealing with transactions, in which case you cannot test this locally and must use a live db. I created a script based off what you have given and am unable to reproduce. Give a more detailed script and I can try again.

LawP09:
[...]

PR #16099, opened 2026-03-13
Strictly validate Model.useConnection() argument and update tests

mrsandy1965:
**Summary**
This PR adds input validation to `Model.useConnection()` to ensure the passed `connection` is a valid Mongoose Connection instance. This prevents silent state corruption and improves developer experience with clearer error messages.
**Proposed Changes**
- Added `const Connection = require('./connection')` to `lib/model.js`.
- Updated `Model.useConnection` validation guard to use `instanceof Connection`.
- Changed the error thrown from a generic `Error` to a `TypeError` with a more descriptive message.

Fixes Issue #16098 

**Examples**
```javascript
const Model = db.model('Test', schema);
// Before: silently accepted {}
// After: throws TypeError: [useConnection()] requires a Mongoose Connection instance.
Model.useConnection({});
```
**Verification Results**
Added comprehensive tests in test/model.test.js covering both missing and invalid connection inputs.
Manual verification script confirmed the fix works as intended.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -189,7 +189,7 @@ Model.prototype.db;
  */
 
 Model.useConnection = function useConnection(connection) {
-  if (!(connection instanceof Connection)) {
+  if (connection == null || typeof connection.model !== 'function' || typeof connection.collection !== 'function' || connection.models == null) {
     throw new TypeError('`useConnection()` requires a Mongoose Connection instance.');
   }
   if (this.db) {
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -9168,7 +9168,7 @@ describe('Model', function() {
       assert.throws(() => db.model('Test'), /MissingSchemaError/);
     });
 
-    it('should throw an error if no connection is passed (gh-14802)', async function() {
+    it('should throw an error if no connection is passed (gh-16098)', async function() {
       const schema = new mongoose.Schema({
         name: String
       });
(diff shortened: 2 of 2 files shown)


lib/model.js
```diff
@@ -188,8 +189,8 @@ Model.prototype.db;
  */
 
 Model.useConnection = function useConnection(connection) {
-  if (!connection) {
-    throw new Error('Please provide a connection.');
+  if (!(connection instanceof Connection)) {
```
vkarpov15: I'm considering whether  this should rely on duck typing check rather than a strict instance of. People frequently have multiple copies of the mongoose module floating around and I could see a case where a model defined in a shared lib would useConnection() on a connection from a different mongoose module. WDYT @AbdelrahmanHafez @hasezoey

hasezoey: Personally, i think duplicated (non-deduped) copies should be avoided, and not especially supported, because this can lead to very confusing errors, problems or straight up weird and unexpected behavior. (and on a performance level, would you really want multiple different instances being loaded & running?)

vkarpov15: Personally, I agree wholeheartedly - importing schemas created in one copy of the module and using them in another copy is dubious. But we get a lot of issues like #15466, #10404, etc. related to npm resolving multiple copies of the Mongoose module when using shared libs. 

One alternative we could consider to minimize risk would be version checking. Check if `conn` is a connection and `connection.base.version === this.db.base.version` - that way if you have 2 copies of the Mongoose module on the exact same version this works fine.

mrsandy1965:
Hi @vkarpov15 — just a quick update regarding PR #16099.

I accidentally closed it while the discussion was still ongoing — I misread the conversation as a rejection. I wasn't able to reopen it due to a branch conflict (I had multiple fixes on the same `master` branch).

I've since cleaned up my branches and I'm opening a new PR that implements the version-checking approach you suggested:

```js
if (!conn || conn.base?.version !== this.db.base.version) {
  throw new TypeError('`useConnection()` requires a Mongoose Connection instance.');
}
```

Sorry for the noise — I'll make sure to use dedicated branches going forward.

PR #16100, opened 2026-03-14
replace ms with ms-tiny

ofershap:
ms has no TypeScript types and no ESM exports.

ms-tiny has the same API, ships ESM + CJS with built-in types. 833 bytes gzipped, zero deps.

One behavioral difference: ms-tiny throws on invalid input instead of returning undefined. Added try-catch where needed to preserve existing behavior.

https://npmgraph.js.org/?q=ms-tiny

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -5,7 +5,7 @@
  */
 
 const UUID = require('mongodb/lib/bson').UUID;
-const ms = require('ms');
+const ms = require('ms-tiny');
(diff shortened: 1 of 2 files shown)


vkarpov15:
I'm going to close this PR for now, I'm concerned about replacing a stable well established dependency with a brand new one, especially given that mongoose only uses ms for a very minimal legacy case.

PR #16125, opened 2026-03-19
feat(document): add $getChanges() alias, deprecate getChanges()

### LINKED ISSUE #15959: `getChanges` should be renamed to `$getChanges` in mongoose 10
### Prerequisites

### Issue

Basically the title. We should also go over the rest of the internal properties/methods and make sure they all start with `$`, if anyone uses `getChanges` as a method or a property name they'll now be unable to use `getChanges`.

Maybe we can even introduce an alias `$getChanges` now, and for the rest of the properties, if any, and make the original `getChanges` emit a deprecation warning.

techcodie:
Closes #15959

Hey! This PR adds `$getChanges()` as the preferred way to get pending document changes, and deprecates the old `getChanges()` with a `process.emitWarning`.

**Why?**

Right now `getChanges` sits on the document prototype without the `$` prefix, which means if someone defines a schema method or virtual called `getChanges`, it just gets silently overridden. That's confusing and hard to debug. The `$` prefix convention exists exactly to avoid this kind of collision.

**What changed:**

- `$getChanges()` — new primary method, same logic that `getChanges()` had before
- `getChanges()` — still works, but now delegates to `$getChanges()` and fires a deprecation warning via `utils.warn()` (which uses `process.emitWarning`)
- Updated TypeScript types — added `$getChanges()` and marked `getChanges()` with `@deprecated`

**Backward compat:**

Nothing breaks. `getChanges()` still works the same way, it just warns you to switch over. The warning only fires once per process thanks to how Node handles `emitWarning` with the same message.

As mentioned in the issue, this could be a first step before a full rename or removal of other non-prefixed internal methods in a future major version.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -4908,22 +4908,42 @@ Document.prototype.$__fullPath = function(path) {
  *     });
  *
  *     // returns an empty object, no changes happened yet
- *     user.getChanges(); // { }
+ *     user.$getChanges(); // { }
  *
  *     user.country = undefined;
  *     user.age = 26;
  *
- *     user.getChanges(); // { $set: { age: 26 }, { $unset: { country: 1 } } }
+ *     user.$getChanges(); // { $set: { age: 26 }, { $unset: { country: 1 } } }
  *
  *     await user.save();
  *
- *     user.getChanges(); // { }
+ *     user.$getChanges(); // { }
  *
- * Modifying the object that `getChanges()` returns does not affect the document's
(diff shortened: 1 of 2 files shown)


techcodie:
Hi @vkarpov15 

Just a gentle ping on this PR when you get a chance.

This adds `$getChanges()` and deprecates `getChanges()` to prevent method name collisions, while keeping full backward compatibility. All checks are passing.

Happy to make any changes if needed. Thanks!

vkarpov15 (approved):

PR #16149, opened 2026-03-22
fix(model): use duck-typing with version check to validate the argument to useConnection() is actually a connection

### LINKED ISSUE #16098: `Model.useConnection()` lacks validation for Connection instance, causing silent state corruption

vkarpov15:
Fix #16098

**Summary**

Throw an error if the argument to `useConnection()` doesn't have `model()` and `collection()` functions - sufficient to avoid many common potentially incorrect arguments. Also check that the version is an exact match. This avoids erroring out in the case of multiple Mongoose modules presuming the versions line up exactly. We do not want to support mixed versions - using a connection from version 8.1.4 with a model from version 8.2.1 is not something we can reasonably write tests for.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -193,7 +193,7 @@ Model.useConnection = function useConnection(connection) {
     throw new MongooseError('`useConnection()` requires a Mongoose connection.');
   }
   if (this.db.base?.version && this.db.base?.version !== connection.base?.version) {
-    throw new MongooseError('The connection passed to `useConnection()` has a different version of Mongoose than the model you are using.');
+    throw new MongooseError(`The connection passed to \`useConnection()\` has a different version of Mongoose (${connection.base?.version}) than the model you are using (${this.db.base?.version}).`);
   }
   if (this.db) {
     delete this.db.models[this.modelName];
--- a/lib/model.js
+++ b/lib/model.js
@@ -189,7 +189,7 @@ Model.prototype.db;
  */
 
 Model.useConnection = function useConnection(connection) {
-  if (!connection || typeof connection.model !== 'function' || typeof connection.collection !== 'function') {
+  if (typeof connection?.model !== 'function' || typeof connection.collection !== 'function') {
     throw new MongooseError('`useConnection()` requires a Mongoose connection.');
   }
   if (this.db.base?.version && this.db.base?.version !== connection.base?.version) {
--- a/lib/model.js
+++ b/lib/model.js
@@ -189,7 +189,7 @@ Model.prototype.db;
  */
 
 Model.useConnection = function useConnection(connection) {
-  if (typeof connection?.model !== 'function' || typeof connection.collection !== 'function') {
+  if (typeof connection?.model !== 'function' || typeof connection.collection !== 'function' || typeof connection.base?.version !== 'string') {
     throw new MongooseError('`useConnection()` requires a Mongoose connection.');
   }
   if (this.db.base?.version && this.db.base?.version !== connection.base?.version) {
--- a/lib/model.js
+++ b/lib/model.js
@@ -189,8 +189,11 @@ Model.prototype.db;
  */
 
 Model.useConnection = function useConnection(connection) {
-  if (!connection) {
-    throw new MongooseError('Please provide a connection.');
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -189,8 +189,11 @@ Model.prototype.db;
  */
 
 Model.useConnection = function useConnection(connection) {
-  if (!connection) {
-    throw new MongooseError('Please provide a connection.');
+  if (!connection || typeof connection.model !== 'function' || typeof connection.collection !== 'function') {
```
AbdelrahmanHafez:
```suggestion
  if (typeof connection?.model !== 'function' || typeof connection.collection !== 'function') {
```

AbdelrahmanHafez (approved):

lib/model.js
hasezoey:
```suggestion
  if (typeof connection?.model !== 'function' || typeof connection?.collection !== 'function') {
```

~~As this is a OR, it may error on the second condition~~

Nevermind, forgot `(typeof undefined !== 'function') === true`.

test/model.test.js
```diff
@@ -9173,9 +9173,42 @@ describe('Model', function() {
...
-      assert.throws(() => {
-        Model.useConnection();
-      }, { message: 'Please provide a connection.' });
+      assert.throws(
+        () => Model.useConnection(),
+        { name: 'MongooseError', message: '`useConnection()` requires a Mongoose connection.' }
+      );
+    });
+
+    it('should throw a MongooseError if a non-connection object is passed (gh-16098)', async function() {
```
hasezoey:
```suggestion
    it('should throw a MongooseError if a non-connection object is passed (gh-16098)', function() {
```

test/model.test.js
```diff
@@ -9173,9 +9173,42 @@ describe('Model', function() {
...
+      const schema = new mongoose.Schema({ name: String });
+      const Model = db.model('Test', schema);
+      assert.throws(
+        () => Model.useConnection({}),
+        { name: 'MongooseError', message: '`useConnection()` requires a Mongoose connection.' }
+      );
+    });
+
+    it('should throw a MongooseError if null is passed (gh-16098)', async function() {
```
hasezoey:
```suggestion
    it('should throw a MongooseError if null is passed (gh-16098)', function() {
```

test/model.test.js
```diff
@@ -9173,9 +9173,42 @@ describe('Model', function() {
...
+      const schema = new mongoose.Schema({ name: String });
+      const Model = db.model('Test', schema);
+      assert.throws(
+        () => Model.useConnection(null),
+        { name: 'MongooseError', message: '`useConnection()` requires a Mongoose connection.' }
+      );
+    });
+
+    it('should throw a MongooseError if mismatched Mongoose version (gh-16098)', async function() {
```
hasezoey:
```suggestion
    it('should throw a MongooseError if mismatched Mongoose version (gh-16098)', function() {
```

lib/model.js
```diff
@@ -189,8 +189,11 @@ Model.prototype.db;
...
+    throw new MongooseError('`useConnection()` requires a Mongoose connection.');
+  }
+  if (this.db.base?.version && this.db.base?.version !== connection.base?.version) {
+    throw new MongooseError('The connection passed to `useConnection()` has a different version of Mongoose than the model you are using.');
```
hasezoey: For clarity, this message could include both versions used.

hasezoey (approved):
LGTM, minor nitpicks.

PR #16174, opened 2026-03-24
perf(document+model): avoid parallel save error instantiation, simplify resetting atomics, streamline validation and collection handling

vkarpov15:
**Summary**

Some performance improvements to optimize the `saveSimple` benchmark path

1. Separate `$__resetAtomics()` function to avoid the overhead of `$__dirty()`
2. Avoid instantiating `ParallelSaveError` every time. ParallelSaveError is substantially less useful now that we have async stack traces.
3. Minimize `delete()` on the `paths` set in `_getPathsToValidate()`: avoid adding the paths to begin with.
4. Avoid converting `arguments` -> `Array.from(arguments)` and avoid instantiating `opId` unless the user actually uses them. `Array.from(arguments)` and `opId` have non-trivial performance costs.

Benchmark on master (sample run):

```
{
  "Average mongoose save time ms": 2.37,
  "Average driver insertOne time ms": 0.72
}

```

With this change:

```
{
  "Average mongoose save time ms": 2.1,
  "Average driver insertOne time ms": 0.86
}
```

So we're getting a decent speedup of maybe 10-15% on `saveSimple`, which is helpful

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2981,7 +2981,37 @@ function _getPathsToValidate(doc, pathsToValidate, pathsToSkip, isNestedValidate
       // Skip $* paths - they represent map schemas, not actual document paths
       return;
     }
+
+    const _pathType = doc.$__schema.path(p);
+
+    // Optimization: if primitive path with no validators, or array of primitives
+    // with no validators, skip validating this path entirely.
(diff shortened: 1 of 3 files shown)


hasezoey (commented):
LGTM

PR #16182, opened 2026-03-26
fix(schema): support toJSONSchema on unions

vkarpov15:
Fix #16179

**Summary**

Add `toJSONSchema()` to union types, with the note that `required` on the union propagates down to the individual union members by default. Needed to make some changes to how `toJSONSchema` works to support that - a `_defaultRequired` option to createJSONSchemaTypeDefinition sets `isRequired` to a value unless the schematype has its own required.

**Examples**


hasezoey (changes requested):
Mostly looks good to me, though the provided option in `Number` should use the new name (as Copilot already pointed out), and if what Copilot is correct regarding `oneOf` and multiple `null`, `null` should be de-duplicated.

PR #16186, opened 2026-03-28
chore: use TSTyche assertions

mrazauskas:
**Summary**

This PR migrates type tests to TSTyche assertions in the following files:

```
test/types/middleware.preposttypes.test.ts
test/types/middleware.test.ts
test/types/model.skip.middleware.test.ts
```

As mentioned before, it would be nice to generate `middleware.preposttypes.test.ts` using the [template test file](https://tstyche.org/guides/template-test-files) feature. I can work on this, but first let’s agree wherever the shape of the test is correct.

And I also updated TSTyche to the latest version ([release notes](https://tstyche.org/releases/tstyche-7)). It kept me busy with that lately.


test/types/middleware.preposttypes.test.ts
```diff
@@ -1,881 +1,871 @@
-import { Schema, model, Model, Document, SaveOptions, Query, Aggregate, HydratedDocument, PreSaveMiddlewareFunction } from 'mongoose';
-import { ExpectAssignable, ExpectType } from './util/assertions';
+import { Schema, Document, Query, HydratedDocument } from 'mongoose';
+import { expect } from 'tstyche';
 
 interface IDocument extends Document {
   name?: string;
 }
 
-const preMiddlewareFn: PreSaveMiddlewareFunction<Document> = function(opts) {
```
mrazauskas, line 8: Repeats in `test/types/middleware.test.ts‎`.

test/types/middleware.preposttypes.test.ts
```diff
@@ -1,881 +1,871 @@
...
 schema.pre('save', { document: false, query: true }, function() {
-  ExpectType<never>(this);
+  expect(this).type.toBe<never>();
 });
 
 schema.post('save', { document: false, query: true }, function(res) {
-  ExpectType<never>(this);
-  ExpectType<unknown>(res);
+  expect(this).type.toBe<never>();
+  expect(res).type.toBe<never>();
```
mrazauskas, line 45: Here and in similar tests below, `res` is of type `never` not `unknown`.

tstyche.json
mrazauskas, line 1: The default config file name got changed to `tstyche.json`.

test/types/middleware.preposttypes.test.ts
vkarpov15: It looks like Expect<> was similarly not good about handling `never` vs `unknown`, good catch. Glad tstyche handles this.

vkarpov15 (approved):
This is great, thanks :+1: If you want to try using the template file to generate the middleware test file that would be cool to see

Issue #9534, opened 2020-11-10
Invalid schema inheritance in discriminator model in case of base schema property is an array


oopyrj:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**

(node:12548) UnhandledPromiseRejectionWarning: ValidationError: BModel validation failed: prop.reqProp: Path `reqProp` is required.

Properties are incorrectly inherited in discriminator in case of base property being an array and discriminator property being non-array

**If the current behavior is a bug, please provide the steps to reproduce.**
```
const mongoose = require('mongoose');

const { Schema } = mongoose;


const AModel = mongoose.model('AModel', new mongoose.Schema({
	prop: { type: [new mongoose.Schema({
		reqProp: {
			type: Date,
			required: true
		},
	})]},
}, {
	discriminatorKey: 'discriminator'
}));

const BModel = AModel.discriminator('BModel', new Schema({
	prop: {
		type: new Schema({
			hash: {
				type: String,
			}
		})
	}
}));



(async () => {
	await mongoose.connect(`mongodb://127.0.0.1:27017/test`)
	const c = new BModel({
		prop: {
			hash: '123',
		}
	})

	await c.save({ validateBeforeSave: true });
})();

```

**What is the expected behavior?**

'prop' property wouldn't inherit base property props but rather override it, after all 'prop' types in schemas are different (nested vs array), aren't they? Or is there a way to omit base schema properties?

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Node: 12.16.3
Mongoose: 5.10.13
MongoDB: 3.6

oopyrj:
For now as a workaround i resorted to:
BModel.schema._baseSchema.remove('prop');

oopyrj:
Also, 'prop' ends up in singleNestedPaths in BModel (which probably means it's a nested, right?) unless i specify { typePojoToMixed: true } which contradicts documentation that states that if I specify type: new Schema({}) then it'll be subdoc.

oopyrj:
awesome, thanks

PR #16187, opened 2026-03-28
fix(utils): wrap discriminator merge check in parentheses to fix precedence

Necro-Rohan:
**Summary**:-

While going through `lib/utils.js`, I noticed a regression in `merge()` that seems to have been introduced with the changes from gh-9534. The issue comes down to JavaScript operator precedence — the `isDiscriminatorSchemaMerge` check was only affecting part of the condition, not the whole thing.

**The issue**:-

Right now the condition is effectively evaluated as:

`(options.isDiscriminatorSchemaMerge && A) || B`

Because of that, whenever Condition B is true (the array vs. single nested mismatch case), the merge gets skipped even if `isDiscriminatorSchemaMerge` is false. That doesn’t seem intentional, and in practice it can cause merges to silently drop data in normal (non-discriminator) scenarios that happen to hit this edge case.

**The fix**:-

I added parentheses to group the branches, ensuring the guard covers the entire mismatch logic.

`options.isDiscriminatorSchemaMerge && (ConditionA || ConditionB)`

This keeps the behavior introduced in gh-9534, but limits it to actual discriminator merges, which seems to be the original intent.
While I was here, I also checked `hasKey` and `mergeClone` for similar patterns, but didn’t see the same issue there.

**Tests**:-

Added a couple of regression tests in `test/utils.test.js` to make sure this only triggers when `isDiscriminatorSchemaMerge` is true, and to prevent this from slipping back in later.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -313,8 +313,8 @@ exports.merge = function merge(to, from, options, path) {
         // base schema has a given path as a single nested but discriminator schema
         // has the path as a document array, or vice versa (gh-9534)
         if (options.isDiscriminatorSchemaMerge &&
-            (from[key].$isSingleNested && to[key].$isMongooseDocumentArray) ||
-            (from[key].$isMongooseDocumentArray && to[key].$isSingleNested)) {
+            ((from[key].$isSingleNested && to[key].$isMongooseDocumentArray) ||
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
LGTM good find

Necro-Rohan:
> LGTM good find

Always Happy to Help.

PR #16197, opened 2026-03-30
docs: fix typos in code comments

Goldyvaiiii:
This PR fixes a few minor typographical errors in comments to improve code readability. Found via codespell.

### DIFF
--- a/lib/schemaType.js
+++ b/lib/schemaType.js
@@ -380,10 +380,10 @@ SchemaType.get = function(getter) {
  * #### Example:
  *
  *     // values are cast:
- *     const schema = new Schema({ a number: { type: Number, default: 4.815162342 }})
+ *     const schema = new Schema({ num: { type: Number, default: 4.815162342 }})
  *     const M = db.model('M', schema)
  *     const m = new M;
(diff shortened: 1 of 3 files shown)


lib/schemaType.js
```diff
@@ -380,10 +380,10 @@ SchemaType.get = function(getter) {
  * #### Example:
  *
  *     // values are cast:
- *     const schema = new Schema({ aNumber: { type: Number, default: 4.815162342 }})
+ *     const schema = new Schema({ a number: { type: Number, default: 4.815162342 }})
  *     const M = db.model('M', schema)
  *     const m = new M;
- *     console.log(m.aNumber) // 4.815162342
+ *     console.log(m.a number) // 4.815162342
```
AbdelrahmanHafez: This is wrong, should stay as `aNumber`

AbdelrahmanHafez (changes requested):
One comment, otherwise LGTM

lib/schemaType.js
vkarpov15: I updated it to `num` because LLM slop will keep tripping up on this if we leave it as `aNumber`

PR #16203, opened 2026-04-01
fix(connection): snapshot Date in heartbeat handler and flush queue on recovery

andreialecu:
## Summary

Fixes #16183. Two related bugs in the heartbeat staleness check:

- **Missing Date snapshot**: `drivers/node-mongodb-native/connection.js` used the global `Date.now()` in the `serverHeartbeatSucceeded` handler instead of a snapshotted copy (like `lib/connection.js` already does). When test frameworks fake `globalThis.Date`, `_lastHeartbeatAt` gets a fake timestamp while the `readyState` getter uses the real `Date.now()`, causing false staleness detection.

- **No recovery path for buffered operations**: When a stale heartbeat causes `readyState` to report `disconnected`, operations buffer into `_queue`. When the next heartbeat refreshes `_lastHeartbeatAt`, `readyState` returns `connected` again, but the queue is never flushed because `_readyState` never actually changed — so no state change event fires and `onOpen()` is never called. This extracts `_flushQueue()` from `onOpen()` and calls it from the heartbeat handler when the connection is no longer stale.

## Test plan

- [x] Added test: heartbeat handler uses snapshotted Date, not affected by faked `globalThis.Date`
- [x] Added test: buffered operations flush when heartbeat refreshes stale connection
- [x] Added test: child db (`useDb()`) queues also flush on heartbeat recovery
- [x] Existing `connection pool sharing` tests all pass
- [x] ESLint passes on changed files

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -1005,10 +1005,7 @@ Connection.prototype.error = function error(err, callback) {
 Connection.prototype.onOpen = function() {
   this.readyState = STATES.connected;
 
-  for (const d of this._queue) {
-    d.fn.apply(d.ctx, d.args);
-  }
-  this._queue = [];
+  this._flushQueue();
 
   // avoid having the collection subscribe to our event emitter
   // to prevent 0.3 warning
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
Good catch, thanks :+1:

PR #16204, opened 2026-04-01
chore: ignore scripts when running npm install for security

vkarpov15:
**Summary**

Avoid running preinstall and postinstall scripts when running `npm install` as an additional layer of security against supply chain attacks like the recent one against Axios

**Examples**


AbdelrahmanHafez (approved):

hasezoey (commented):
LGTM, though usages of `mongodb-client-encryption` might need a slight extra review to see that they still work as expected (like CI), as Copilot pointed out already.

PR #16205, opened 2026-04-01
fix(clone): support cloning objects that have an own __proto__ property

vkarpov15:
Fix #16202

**Summary**

I want to consider this PR carefully before merging because of potential security ramifications (cloning `__proto__` is risky).

General idea: Mongoose's `clone()` avoids copying `__proto__`, which can be problematic for objects that have an own `__proto__` property. However, we want to avoid the two common workarounds for __proto__ copying: we don't want to make `clone()` return a null prototype object because that would be backwards breaking, and we do not want to use `Object.defineProperty()` in `clone()` because that can be slow. Hence the alternative trick of spreading a pre-existing object with an `__proto__` property, which seems to do what we want: create a plain object with ordinary prototype and __proto__ as a property much faster than using `Object.defineProperty()`

image

**Examples**

### DIFF
--- a/lib/helpers/clone.js
+++ b/lib/helpers/clone.js
@@ -2,7 +2,6 @@
 
 const Decimal = require('../types/decimal128');
 const ObjectId = require('../types/objectid');
-const specialProperties = require('./specialProperties');
 const isMongooseObject = require('./isMongooseObject');
 const getFunctionName = require('./getFunctionName');
 const isBsonType = require('./isBsonType');
@@ -15,6 +14,8 @@ const BSON = require('mongodb/lib/bson');
 const UUID = BSON.UUID;
 
 const Binary = BSON.Binary;
+
+const objWith__proto__ = { ['__proto__']: null };
 
 /**
  * Object clone with Mongoose natives support.
@@ -161,7 +162,7 @@ function cloneObject(obj, options, isArrayChild) {
   const minimize = options?.minimize;
   const omitUndefined = options?.omitUndefined;
   const seen = options?._seen;
-  const ret = {};
+  let ret = {};
   let hasKeys;
 
   if (seen && seen.has(obj)) {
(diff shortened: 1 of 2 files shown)


lib/helpers/clone.js
```diff
@@ -175,10 +176,13 @@ function cloneObject(obj, options, isArrayChild) {
 
   const keys = Object.keys(obj);
   const len = keys.length;
+  if (keys.includes('__proto__')) {
+    ret = { ...objWith__proto__ };
```
hasezoey, line 180: Maybe some documentation would be good here as to why this specific way is required.

hasezoey (commented):
Mostly looks good to me, though the Copilot suggestions should be investigated and documentation should be made why a specific way is required (as that is not obvious at a glance)
Also, just to be sure, maybe consider this for at the very least a minor version, if not a major version.

lib/helpers/clone.js
AbdelrahmanHafez: Seconded, ~~also not quite sure whether the `includes` check is broader than it should be.~~
**Edit:** That's the right check, `keys` is an array, not a string.

Potential perf improvement instead of using `keys.includes('__proto__')` is to use `Object.hasOwn('__proto__')`

lib/helpers/clone.js
```diff
@@ -2,7 +2,6 @@
 
 const Decimal = require('../types/decimal128');
 const ObjectId = require('../types/objectid');
-const specialProperties = require('./specialProperties');
```
AbdelrahmanHafez, line 5: BTW `specialProperties` (which includes `__proto__`) is used in 7 files, not just `clone()`. I quickly scanned through them and I think some of them have the same problem of data loss as well and should be updated.

AbdelrahmanHafez (changes requested):
A couple of comments, overall LGTM and I don't see anything that can be a breaking change.

vkarpov15:
I'm going to close this one for now since I came up with an alternative fix for the original issue, and the risk of security issues isn't worth the risk.

PR #16222, opened 2026-04-07
chore: use TSTyche assertions

mrazauskas:
**Summary**

This PR migrates type tests to TSTyche assertions in the following files:

```
test/types/inferrawdoctype.test.ts
test/types/models.test.ts
test/types/queries.test.ts
test/types/querycursor.test.ts
```

While working on these, I found a bug in TSTyche logic. Index signatures were required to be defined in the same order, but that is not correct behaviour. Fixed that and updated the TSTyche version.

Also, the new `.toBeInstantiableWith()` matcher is used here once.

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -8,6 +8,7 @@ import mongoose, {
   InferSchemaType,
   InsertManyResult,
   Model,
+  ModifyResult,
   Query,
   Schema,
   Types,
@@ -20,8 +21,8 @@ import mongoose, {
   UpdateManyModel
 } from 'mongoose';
 import { AutoTypedSchemaType, autoTypedSchema } from './schema.test';
-import { ModifyResult, UpdateOneModel as MongoUpdateOneModel, ChangeStreamInsertDocument, ObjectId } from 'mongodb';
-import { ExpectAssignable, ExpectType } from './util/assertions';
+import { UpdateOneModel as MongoUpdateOneModel, ChangeStreamInsertDocument, ObjectId } from 'mongodb';
+import { expect } from 'tstyche';
 
 function rawDocSyntax(): void {
   interface ITest {
@@ -40,7 +41,7 @@ function rawDocSyntax(): void {
 
   const Test = connection.model<ITest, TestModel>('Test', TestSchema);
 
-  ExpectType<Model<ITest, {}, ITestMethods, {}>>(Test);
+  expect(Test).type.toBe<Model<ITest, {}, ITestMethods, {}>>();
 
   const doc = new Test({ foo: '42' });
   console.log(doc.foo);
@@ -81,10 +82,10 @@ async function insertManyTest() {
   });
 
   const res = await Test.insertMany([{ foo: 'bar' }], { rawResult: true });
-  ExpectType<Types.ObjectId>(res.insertedIds[0]);
+  expect(res.insertedIds[0]).type.toBe<Types.ObjectId>();
 
   const res2 = await Test.insertMany([{ foo: 'bar' }], { ordered: false, rawResult: true });
-  ExpectAssignable<Error | Object | ReturnType<(typeof Test)['hydrate']>>()(res2.mongoose.results[0]);
+  expect(res2.mongoose.results[0]).type.toBe<Error | Object | ReturnType<(typeof Test)['hydrate']>>();
 }
 
 function gh13930() {
@@ -139,10 +140,10 @@ async function gh10359() {
 
   async function foo(model: Model<User, {}, {}, {}>) {
     const doc = await model.findOne({ groupId: 'test' }).orFail().lean().exec();
-    ExpectType<string>(doc.firstName);
-    ExpectType<string>(doc.lastName);
-    ExpectType<Types.ObjectId>(doc._id);
-    ExpectType<string>(doc.groupId);
+    expect(doc.firstName).type.toBe<string>();
+    expect(doc.lastName).type.toBe<string>();
+    expect(doc._id).type.toBe<Types.ObjectId>();
+    expect(doc.groupId).type.toBe<string>();
     return doc;
   }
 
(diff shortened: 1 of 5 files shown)


test/types/models.test.ts
```diff
@@ -664,31 +663,31 @@ async function gh13705() {
...
-  ExpectType<ExpectedLeanDoc | null>(findOneAndReplaceRes);
+  expect(findOneAndReplaceRes).type.toBe<ExpectedLeanDoc | null>();
 
   const findOneAndUpdateRes = await TestModel.findOneAndUpdate({}, {}, { lean: true });
-  ExpectType<ExpectedLeanDoc | null>(findOneAndUpdateRes);
+  expect(findOneAndUpdateRes).type.toBe<ExpectedLeanDoc | null>();
 
   const findOneAndUpdateResWithMetadata = await TestModel.findOneAndUpdate({}, {}, { lean: true, includeResultMetadata: true });
-  ExpectAssignable<ModifyResult<ExpectedLeanDoc>>()(findOneAndUpdateResWithMetadata);
+  expect(findOneAndUpdateResWithMetadata).type.toBe<ModifyResult<{ name?: string | null | undefined }>>();
```
mrazauskas, line 690: This is the only pace where `ModifyResult` is used. Importing `ModifyResult` from `mongoose` instead of `mongodb`, makes `.toBe()` pass. Is that fine, or the intention was different?

test/types/queries.test.ts
```diff
@@ -549,33 +522,30 @@ async function gh13142() {
...
-  // @ts-expect-error  Property 'age' does not exist on type
-  u3!.age;
-  ExpectAssignable<Function>()(u3.toObject);
+  expect(u3.name).type.toBe<string | undefined>();
+  expect(u3).type.not.toHaveProperty('age');
+  expect(u3.toObject).type.toBeAssignableTo<Function>();
 
-  // @ts-expect-error  Type '{ notInSchema: string; }' has no properties in common with type '{ name?: any; age?: any; _id?: any; __v?: any; }'.
-  await UserModel.findOne().select<{ notInSchema: string }>(['name']).orFail();
+  expect(UserModel.findOne().select).type.not.toBeInstantiableWith<[{ notInSchema: string }]>();
```
mrazauskas, line 548: Here `.not.toBeInstantiableWith()` checks that `UserModel.findOne().select` rejects `{ notInSchema: string }` type argument.

test/types/models.test.ts
vkarpov15: Yes that is correct, Mongoose's ModifyResult is slightly different from the MongoDB driver's for backwards compatibility reasons

test/types/models.test.ts
```diff
@@ -1106,23 +1098,19 @@ async function gh15693() {
 
   const schema = new Schema<IUser, Model<IUser>, UserMethods>({ name: { type: String, required: true } });
   schema.method('printNamePrefixed', function printName(this: IUser, prefix: string) {
-    // @ts-expect-error  Property 'isModified' does not exist on type 'IUser'.
-    this.isModified('name');
-    // @ts-expect-error  Property 'doesNotExist' does not exist on type 'IUser'.
-    this.doesNotExist();
-    ExpectType<string>(this.name);
+    expect(this).type.not.toHaveProperty('isModified');
```
vkarpov15, line 1101: Is there anything wrong with using `// @ts-expect-error` for these sort of assertions or is this just a style choice? Not saying it's wrong or I disagree, just want to know.

test/types/queries.test.ts
```diff
@@ -389,11 +370,8 @@ function gh14397() {
 
   const id = 'Test Id';
 
-  let idCondition: Condition<WithId<TestUser>['id']>;
-  let filter: QueryFilter<WithId<TestUser>>;
-
-  ExpectAssignable<typeof idCondition>()(id);
-  ExpectAssignable<typeof filter>()({ id });
+  expect<Condition<WithId<TestUser>['id']>>().type.toBeAssignableFrom(id);
```
vkarpov15, line 373: Is there a reason to prefer toBeAssignableFrom vs toBeAssignableTo here? Again just want to know for my own learning, not saying it is a problem.

test/types/queries.test.ts
vkarpov15: Could you also use `toBeCallableWith` here as well?

vkarpov15 (approved):
LGTM just had a couple of minor questions

test/types/models.test.ts
mrazauskas: In a way `.not.toHaveProperty()` is expecting an error as well, so it is somewhat equivalent to `@ts-expect-error`. In both cases syntax is erased, file is checked again and newly found errors mapped to either assertions or `@ts-expect-error` comments. These are two separate checks, so sticking to one or another style improves performance.

I prefer assertions, because they are more expressive and do not depend on error message that can change between TypeScript versions.

test/types/queries.test.ts
mrazauskas: I try to keep the type under test on the left hand side of the assertion. So I thought `id` is not under test here. Maybe `id` could be replaced with `string`:

```ts
expect<Condition<WithId<TestUser>['id']>>().type.toBeAssignableFrom<string>();
expect<QueryFilter<WithId<TestUser>>>().type.toBeAssignableFrom<{ id: string }>();
```

In a way, `.toExtend()` could perhaps read better? Easy to add an alias. In the other hand, it will make it even harder to explain which assertion to choose.

test/types/queries.test.ts
mrazauskas: Hm.. Does not work with `.toBeCallableWith()`, I just tried. The thing is that `UserModel.findOne().select<{ notInSchema: string }>` errors with or without final `()`. Therefore it is not possible pass it to `expect()`, but we can test it as an [instantiation expression](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#instantiation-expressions) using `.not.toBeInstantiableWith()`.

PR #16227, opened 2026-04-12
#16226  Extend defaults query option to find

vkarpov15:
**Summary**

Fixed up the discriminator issue from #16226 that copilot pointed out

cc @sderrow 

**Examples**


AbdelrahmanHafez (approved):
LGTM 👍

lib/queryHelpers.js
```diff
@@ -90,7 +90,17 @@ exports.createModel = function createModel(model, doc, fields, userProvidedField
...
+        skipId: true,
+        isNew: false,
+        willInit: true
+      };
+      if (options?.defaults != null) {
+        _opts.defaults = options.defaults;
+      }
+      if (options?.strict != null) {
+        _opts.strict = options.strict;
+      }
```
sderrow: This same block exists on lines 108-118 below. Perhaps worth collapsing and defining once, right before line 79? That way it's consistent between discriminators and non-discriminators, and there's no risk of forgetting to add future options to both paths.

PR #16231, opened 2026-04-18
feat: add deep option to omitUndefined helper

IshitaSingh0822:
Currently `mongoose.omitUndefined()` only strips undefined values 
one level deep, so nested undefined keys are silently left behind.

This PR adds an optional `{ deep: true }` flag that recursively 
removes undefined from all nested objects and arrays.

Fully backward compatible — existing behaviour is unchanged.

Changes:
- lib/helpers/omitUndefined.js — added deep recursive support
- test/helpers/omitUndefined.test.js — 29 tests covering all cases

### DIFF
--- a/lib/helpers/omitUndefined.js
+++ b/lib/helpers/omitUndefined.js
@@ -1,20 +1,59 @@
 'use strict';
 
-module.exports = function omitUndefined(val) {
+/**
+ * Remove undefined values from an object or array.
+ *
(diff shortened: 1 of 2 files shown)


vkarpov15:
This feature isn't necessary because `omitUndefined()` is primarily useful for top-level properties: avoiding `$set: { name: undefined }` setting `name: null` or `name: undefined` ending up as a query for `name: null`. Not as useful for nested properties.

Issue #15687, opened 2025-10-15
Add llms.txt & llms-full.txt for better documentation inference by AI-assisted development


RayyanNafees:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

This will increase accuracy for the models to quickly be made from , eg, Zod Schema or JSON schema, by quickly converting them to mongoose.Schema and creating models at once

There's also a [Standardised Guide to create llms.txt](https://llms-txt.io/blog/how-to-create-llms-full-txt)

### Example

Like nextjs offeres one here : [nextjs.org/docs/llms-full.txt](https://nextjs.org/docs/llms-full.txt)

Albez0-An7h:
Hi @RayyanNafees, I’d like to work on this feature. The proposal looks clear, and I’m interested in implementing it. Could you please assign the issue to me so I can get started?

PR #16232, opened 2026-04-21
New mongoosejs.com layout

vkarpov15:
**Summary**

It's been over 8 years since we last redesigned the docs site. I wanted to first spruce up the layout a bit to look more like Vue's docs

image

image

Dark mode:

image

image

Mobile:

image

image

image

image

**Examples**


docs/layout.pug
```diff
@@ -15,177 +15,223 @@ html(lang='en')
...
+              each pastVersion in versions.pastVersions
+                li
+                  a.version-picker-item(role="menuitem" href=pastVersionsHref(pastVersion.path)) v#{pastVersion.listed}
+
+          nav.sidebar-nav(aria-label="Docs navigation")
+            .sidebar-section
+              h3.sidebar-section-title Getting Started
+              ul.sidebar-list
+                li.sidebar-item
+                  a.sidebar-link(href=`${versions.versionedPath}/docs/index.html`, class=outputUrl === `${versions.versionedPath}/docs/index.html` ? 'active' : '') Quick Start
```
hasezoey: I dont know how feasible it is for pug, but maybe we could put the "active outputurl class" check as a function to reduce duplication?

hasezoey (changes requested):
Home issues i found:
- the homepage still makes use of variables that are now unset (some i found, but might not be all: `--text-secondary`, `--btn-bg`, `--btn-text-shadow`); practically making all buttons disappear (see screenshots)
- likely due to the same issue, light/dark switch does not work on the homepage
- the TOC sidebar can be opened in mobile mode, even if there is no content present (ex `docs/index.html`)
- some headers in the API page (likely other *could* too) get too long and offset the headerbar (see screenshots)
- i dont know how intentional it is, but on mobile the "edit page / copy" buttons are below the header instead of on the right side, even if there is enough space for most headers
- on higher space screens, the TOC sidebar is still rather small, commonly cutting names off in the API page (see screenshots)

<details>
<summary>Issue Screenshots</summary>

homepage without buttons (invisible `read the docs`, `discover plugins` and `get professionally supported mongoose`): 
Image

long header names:
Image

small toc, cutting off names, even if there would be space to expand:
Image

</details>

Otherwise, this looks really great; i really do like [docusaurus](https://docusaurus.io/docs/installation)-ish layout style.

PS: unrelated to this issue, it seems like the [sponsors page](https://mongoosejs.com/docs/sponsors.html) seems to be empty.

docs/layout.pug
vkarpov15: Done in https://github.com/Automattic/mongoose/pull/16232/commits/e753c17b68c4aba66cbd1ae4d036ef538586f9c0, looks like [Pug mixins](https://pugjs.org/language/mixins.html) do this quite nicely.

hasezoey (approved):
Looks good to me now. Though the TOC entries are still cut short in wide screens.

AbdelrahmanHafez (approved):
LGTM 👍

vkarpov15:
@hasezoey which TOC entries are you referring to? In the API docs, I made it so that the headers have `overflow-y-auto` so at least they don't break the page width.

hasezoey:
> @hasezoey which TOC entries are you referring to? In the API docs, I made it so that the headers have `overflow-y-auto` so at least they don't break the page width.

The same as i had in my original review screenshot, this time without a "details/spoiler":

Image

<details>
<summary>For context, the full visible space</summary>

full

</details>

vkarpov15:
I see. What do you think we can do about that? A couple of possibilities:

1. Replace `.prototype.` with `#`, so `Mongoose#now()`. `#` has historically been used as a symbol for class members, but I'm not sure how recognizable that is.
2. Strip out `Mongoose.protototype.` entirely in the right side ToC, and just leave `now()`. It is likely clear from context that you're looking at `Mongoose.prototype.`, so `now()` may be sufficient.

hasezoey:
> I see. What do you think we can do about that?

The easiest band-aid fix i can currently think of is to just allow overflow scrolling.
The better alternative i can think of is to somehow allow it to take more space without shifting the other content so that it can make use of the empty space.

Aside from that, i think replacing `.prototype.` with `#` sounds good, though i fear there will still be longer text that gets cut off.

docs/css/mongoose9.css
```diff
@@ -0,0 +1,1261 @@
...
+  border-radius: 3px;
+}
+
+/* Table of contents */
+.toc {
+  margin-bottom: 32px;
+}
+
+.right-sidebar .toc {
+  height: calc(100vh - 310px);
```
hasezoey: I investigated a bit for the toc width adjustment, and found that this, at least on lower height screens, practically reduces the TOC space to less than half the space.
Consider dropping it this rule or reducing by `vh` instead of `px`.

Screenshot:
low_height_toc

(the screenshot is the full artificial height, excluding the header)

docs/css/mongoose9.css
```diff
@@ -0,0 +1,1261 @@
...
+
+.toc-link {
+  display: block;
+  padding: 4px 8px;
+  font-size: 13px;
+  color: var(--text-secondary);
+  line-height: 1.4;
+  border-left: 2px solid transparent;
+  transition: color 0.15s ease, border-color 0.15s ease;
+  overflow: hidden;
```
hasezoey, line 984: This is the main issue, maybe consider dropping this so that the full list is scrollable (at once instead of each individual link being scrollable)

hasezoey (commented):
I investigated the "toc cut short" issue and what could be done about it, but i am not a css / html expert, so there are always trade-offs.
My best try is to reduce the padding between content and toc (and toc itself) to gain *a little* more character space.
And/or changing the list to be `width: fit-content` with the container of that list being `overflow: scroll` and removing the `overflow: hidden` from the `.toc-link`.

PS: maybe also consider using [`grid-template-area`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/grid-template-areas) over `grid-columns` to be easier to change. Example real-world usage is MDN's sidebar itself.

RaschidJFR:
This new look looks great, very in line with modern JS community's taste.
I tested it found no functional issues. I have some feedback purely on aesthetics to make the experience even more appealing:

1. On the left-side menu, headers are not very evident. They could stand out more and/or become collapsible to follow the industry standards. Something like this:

    Screenshot 2026-04-30 at 6 57 02 PM

2. It's so cool to have dark mode. Nitpick: we could adapt the code block's colour pallet to match the preferred schemes. [This list](https://www.gitkraken.com/blog/10-best-vs-code-color-themes-2025) has some of the most popular.
3. The light mode's colour scheme is ok. I don't have any specific recommendation or list to pick from, I also understand it is following Mongoose's original colour pallet. I would suggest updating the default link's colour to remove the remaining "legacy" feeling. Mongoose is up to any modern JS stack and I believe it's fair to show so.
4. And then there's the story of making docs agent-friendly, adding `llms.txt` but that can be a different issue. Just noting it here for the record.

vkarpov15:
@RaschidJFR thanks for the feedback! I implemented (1), I'm not really sure what you mean for (2) or (3) - do you mean you want to change the color of the links in the following screenshot?

image

(4) is tracked in #15687

vkarpov15:
@hasezoey I did the `.prototype.` -> `#` fix. The text overflow ellipsis in the right sidebar toc I think we can live with, Vue docs do the same thing:

image

hasezoey:
> @hasezoey I did the `.prototype.` -> `#` fix. The text overflow ellipsis in the right sidebar toc I think we can live with, Vue docs do the same thing:

Definitely looks better now and there is way less overflow, so i think the things that still overflow are readable / identifiable enough, even when cut-off.

docs/js/sidebar-collapse.js
```diff
@@ -0,0 +1,37 @@
...
+    const toggle = section.querySelector('.sidebar-section-toggle');
+    if (!toggle) return;
+
+    const storageKey = STORAGE_PREFIX + section.id;
+
+    function setCollapsed(collapsed) {
+      section.classList.toggle('collapsed', collapsed);
+      toggle.setAttribute('aria-expanded', collapsed ? 'false' : 'true');
+      try {
+        localStorage.setItem(storageKey, collapsed ? '1' : '0');
```
hasezoey: Should this really be saved across reloads? I think something more temporary would be fine.

docs/layout.pug
```diff
@@ -90,15 +90,19 @@ html(lang='en')
               block
 
           nav.sidebar-nav(aria-label="Docs navigation")
-            .sidebar-section
-              h3.sidebar-section-title Getting Started
-              ul.sidebar-list
+            .sidebar-section.sidebar-section-collapsible#sidebar-section-getting-started
+              h3.sidebar-section-title.sidebar-section-toggle(role="button" tabindex="0" aria-expanded="true" aria-controls="sidebar-list-getting-started")
+                span Getting Started
+                <svg class="sidebar-section-caret" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.5" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true"><polyline points="6 9 12 15 18 9"/></svg>
```
hasezoey: Once again, think this should be de-duplicated.

docs/js/sidebar-collapse.js
vkarpov15: You're right, I removed that.

docs/layout.pug
vkarpov15: Fixed :+1:

hasezoey (approved):

snehil002:
Subject: UI Bug - Sidebar scroll position resets on navigation

Hi @vkarpov15,

Thanks for the great work on the new layout! The site looks much more modern.

I noticed a small UX regression with the new design: when clicking a link in the sidebar, the sidebar scroll position resets to the top instead of persisting. This makes it a bit difficult to navigate through long lists of documentation because you lose your place after every click.

Ideally, the sidebar should either:
- Maintain its scroll position across page loads.
- Automatically scroll the "active" link into view upon navigation.

Are you open to a fix for this, or should I open a separate issue to track it?

vkarpov15:
@snehil002 good suggestion, I opened up PR #16279 :+1:

PR #16233, opened 2026-04-21
types(models): propagate inferred Model type to QueryWithHelpers to support methods on docs returned from statics

### LINKED ISSUE #15532: Document method types are missing methods when queried from statics
### Prerequisites

### Mongoose version

8.16.3

### Node.js version

20.19.0

### MongoDB server version

7.0.15

5.8.3

### Description

Typescript method types are getting dropped if they are returned from a query called from a static method.

### Steps to Reproduce

I have the following code (heavily inspired by [the docs](https://mongoosejs.com/docs/typescript/statics-and-methods.html)):

```ts

const userSchema = new mongoose.Schema(
  {
  },
  {
    methods: {
      updateName(name: string) {
        this.name = name;
        return this.save();
      }
    },
    statics: {
      findByName(name: string) {
        return this.findOne({ name });
      }
    }
  }
);
const UserModel = mongoose.model('User', userSchema);

const doc = new UserModel({ name: 'test' });
// Compiles correctly
doc.updateName('foo');

const foundDoc = await UserModel.findOne({ name: 'test' }).orFail();
// Compiles correctly
foundDoc.updateName('foo');

const foundDocFromStatic = await UserModel.findByName('test').orFail();
// Method types are broken
foundDocFromStatic.updateName('foo');
```

The `foundDoc` variable has the `updateName` method typed correctly. The `foundDocFromStatic` variable seems to have lost the method typing information, even though `findByName` also just does a `findOne()`:

Image

### Expected Behavior

I would expect method types to be returned when the query is happening from a static.

Am I doing something wrong here? Thanks in advance.

vkarpov15:
Fix #15532

**Summary**

#15532 pointed out that if you have a model static that returns a query, the resulting document will not have methods if you are using automatic schema inference. This PR makes all model functions that return queries infer the model type from `this`, which seems to solve the issue. By inferring `this`, Mongoose models can get the "real model type" when creating the query.

**Examples**

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -286,6 +286,98 @@ async function gh15532() {
   expect(foundDocsAfterDistinct[0].updateName('baz')).type.toBe<Promise<UserModel>>();
 }
 
+async function gh15532ExplicitSchemaGeneric() {
+  interface IUser {
+    name: string;
+  }
+
+  const userSchema = new Schema<IUser>({
+    name: { type: String, required: true }
+  });
+
+  const BaseUserModel = model<IUser>('ExplicitGenericUser', userSchema);
+  class UserModel extends BaseUserModel {
+    updateName(name: string) {
+      this.name = name;
+      return this.save();
+    }
+
+    static findByName(name: string) {
+      return this.findOne({ name });
+    }
+  }
+
+  const foundDoc = await UserModel.findOne({ name: 'test' }).orFail();
+  expect(foundDoc.name).type.toBe<string>();
+  expect(foundDoc.updateName('foo')).type.toBe<Promise<UserModel>>();
+
+  const foundById = await UserModel.findById('0'.repeat(24)).orFail();
+  expect(foundById.updateName('foo')).type.toBe<Promise<UserModel>>();
+
+  const foundDocs = await UserModel.find({ name: 'test' });
+  expect(foundDocs[0].updateName('foo')).type.toBe<Promise<UserModel>>();
+
+  const updatedDoc = await UserModel.findOneAndUpdate({ name: 'test' }, { name: 'foo' }, { new: true }).orFail();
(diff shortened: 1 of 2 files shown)


hasezoey (changes requested):
This at first glance this looks good, but this somehow changes the return type of `Model.findOne()`(and likely others) to be `Class` instead of `Document<Class>`. This happens when the model type is manually created by using `Model</**/> & typeof Class` for adding statics onto the model. This is used by typegoose currently. Example Code:

```ts
async function repro() {
  class TestClass {
    public someProp!: string;

    // the following functions do not affect the type, but are here for illustrative purposes
    someFunc(this: HydratedDocument<TestClass>) {
      return this.someProp;
    }

    static someStaticFunc(this: TestClassModel) {
      return this.findOne();
    }
  }

  const schema = new Schema({ someProp: String });
  schema.loadClass(TestClass);

  type TestClassModel = Model<TestClass> & typeof TestClass;
  const model = mongoose.model('test', schema) as TestClassModel;

  // Type: const foundDoc: TestClass | null
  // Should be: const foundDoc: Document<TestClass> | null
  const foundDoc = await model.findOne();

  // Type: const foundDoc: TestClass | null
  // Should be: const foundDoc: Document<TestClass> | null
  const anotherDoc = await model.someStaticFunc();
}
```

Edit: i found a workaround (might also be recommended, if i look at [typescript: statics](https://mongoosejs.com/docs/typescript/statics-and-methods.html#basic-usage)) changing `& typeof Class` to `Pick<typeof Class, keyof typeof Class>`

vkarpov15:
I'm going to close this PR for now and consider alternative approaches. [Copilot's comment](https://github.com/Automattic/mongoose/pull/16233#discussion_r3120049713) is valid: as written, this PR does nothing for `new Schema({}, { methods, statics })`. Only helps in the `extends BaseUserModel` case, which is not our #1 recommended approach right now. We will take another pass at #15532 later.

PR #16237, opened 2026-04-22
feat(schematype): support `allowNull` option to disallow null values even if not `required`

### LINKED ISSUE #15905: All non-required inferred fields are nullable
### Prerequisites

### Mongoose version

9.0.1

### Node.js version

24.11.0

### MongoDB version

n/a

### Operating system

None

_No response_

### Issue

I think that users should be allowed to configure whether they want inferred types to include `| null` for ANY type (even optional ones) unless explicitly included. Maybe an option inside interface DefaultSchemaOptions?

My rational is that when users [create separate interfaces](https://mongoosejs.com/docs/typescript/schemas.html#separate-document-interface-definition) and use them to define their schema instead of using [automatic type inference](https://mongoosejs.com/docs/typescript/schemas.html#automatic-type-inference), I very rarely see people include `| null` for EVERY OPTIONAL TYPE. Despite it being possible for null values to be stored, most users do not want to deal with it in their types. I guess this relates to https://github.com/Automattic/mongoose/issues/14421.

I also mentioned this in https://github.com/Automattic/mongoose/issues/15878 but presume I should of split this request into its own issue.

Thanks!

vkarpov15:
Fix #15905

**Summary**

Mongoose has historically used `null` and `undefined` interchangeably, but TypeScript sometimes makes working with potentially `null` values cumbersome, leading to a lot of `?? undefined`. This PR adds a `allowNull: false` option which allows a value to be `undefined`, but does not allow said value to be set to `null`, including automatic type inference in TypeScript.

**Examples**

### DIFF
--- a/lib/error/messages.js
+++ b/lib/error/messages.js
@@ -30,6 +30,7 @@ msg.DocumentNotFoundError = null;
 msg.general = {};
 msg.general.default = 'Validator failed for path `{PATH}` with value `{VALUE}`';
 msg.general.required = 'Path `{PATH}` is required.';
+msg.general.allowNull = 'Path `{PATH}` does not allow null values.';
 
 msg.Number = {};
 msg.Number.min = 'Path `{PATH}` ({VALUE}) is less than minimum allowed value ({MIN}).';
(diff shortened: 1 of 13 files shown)


hasezoey (approved):

AbdelrahmanHafez (approved):
LGTM 👍

PR #16238, opened 2026-04-22
fix(schema): always pass raw string value to error validators, only trim to 30 chars for maxlength validator

### LINKED ISSUE #16236: value passed in `props` to the `message` for a custom validator are truncated as of 8.18/#15571
### Prerequisites

### Mongoose version

>=8.18

### Node.js version

24

### MongoDB server version

7.x

_No response_

### Description

Based on some digging, it appears that as of Mongoose v8.18, _all_ `props.value` strings are truncated at 30 characters after the change in #15571 that trimmed strings. The text of the PR seems to indicate that the intent of the change was to trim strings for `maxLength` validators, but the implementation affects all validators including custom validators. 

### Steps to Reproduce

Here is a simple schema that validates a property is a valid URL:

```
const schema = new Schema({
  url: {
    type: String,
    validate: {
      validator: function(val) {
        return validateUrlWithProtocol(val) // assume some lib function that does what we want
      },
      message: (props) => {
        // this could contain _any_ logic that needs to manipulate the failing value
        return `${props.value} is not a valid url`
      }
    }
  }
});
```

### Expected Behavior

* for an valid input string `https://hello.short-enough.com` (31 chars), this functions as expected
* for an invalid input string `hello.short-enough.com` (23 chars), this functions as expected
* for an invalid input string `long-long-long.short-enough.com` (32 chars), the message field can no longer perform any business logic on the preserved input value since `props.value` is truncated to `long-long-long.short-enough.co...`

The previous behavior allowed the `message` function to perform business logic on the full input value unchanged.

vkarpov15:
Fix #16236
Re: #15550
Re: #15571

**Summary**

Original bug report was about trimming long strings in the `maxlength` validator, but the fix in #15571 also leaks the 30 character max limit to user-specified validators, which is incorrect. With this PR, the validator props will always have the raw untrimmed value, and we will just apply the 30 character limit to the error message for the `maxlength` validator.

**Examples**

### DIFF
--- a/lib/schema/string.js
+++ b/lib/schema/string.js
@@ -736,7 +736,11 @@ module.exports = SchemaString;
 
 function formatMaxLengthValidatorMessage(msg) {
   return function(props, doc) {
-    if (typeof props.value === 'string' && props.value.length > 30) {
+    if (typeof msg === 'function') {
+      return MongooseError.ValidatorError.prototype.formatMessage(msg, props, doc);
+    }
+
+    if (typeof msg === 'string' && typeof props.value === 'string' && props.value.length > 30) {
       props = Object.assign({}, props, {
         value: props.value.slice(0, 30) + '...'
       });
--- a/lib/schema/string.js
+++ b/lib/schema/string.js
@@ -474,12 +474,14 @@ SchemaString.prototype.maxlength = function(value, message) {
 
   if (value !== null && value !== undefined) {
     let msg = message || MongooseError.messages.String.maxlength;
-    msg = msg.replace(/{MAXLENGTH}/, value);
+    if (typeof msg !== 'function') {
+      msg = msg.replace(/{MAXLENGTH}/, value);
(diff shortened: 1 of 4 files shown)


hasezoey (approved):

lib/schema/string.js
```diff
@@ -736,7 +736,11 @@ module.exports = SchemaString;
 
 function formatMaxLengthValidatorMessage(msg) {
   return function(props, doc) {
-    if (typeof props.value === 'string' && props.value.length > 30) {
+    if (typeof msg === 'function') {
+      return MongooseError.ValidatorError.prototype.formatMessage(msg, props, doc);
+    }
```
hasezoey, line 741: Considering that the following `if` also added a `typeof == "string"` check and is followed by a unconditional call to `formatMessage`, this seems unnecessary; unless it is necessary for performance?

PR #16244, opened 2026-04-27
feat(query): add lean virtuals option

### LINKED ISSUE #12684: Move lean({ virtuals: true }) support from mongoose-lean-virtuals into core Mongoose

vkarpov15:
**Summary**

Fix #12684: moving mongoose-lean-virtuals into Mongoose core. Mostly leveraging existing `applyVirtuals()` infrastructure with a couple of minor tweaks.

Postponing this to 10.0 release because there isn't a very good way to make this change work without breaking existing code that uses mongoose-lean-virtuals

**Examples**


lib/helpers/document/applyVirtuals.js
```diff
@@ -46,7 +64,34 @@ function applyVirtuals(schema, obj, virtuals) {
  * @param {string[]} [virtuals] optional whitelist of virtuals to apply
  */
 
-function applyVirtualsToChildren(schema, res, virtuals) {
+function addToParentMap(res, options, applyVirtualsToResult) {
```
hasezoey, line 69: The JSDoc above this likely needs to be moved or updated.

lib/helpers/document/applyVirtuals.js
```diff
@@ -87,17 +135,17 @@ function applyVirtualsToChildren(schema, res, virtuals) {
  *
  * @param {Schema} schema
  * @param {object} doc
- * @param {string[]} [virtuals] optional whitelist of virtuals to apply
+ * @param {(string|string[])[]} [virtuals] optional whitelist of virtuals to apply
  * @returns
```
hasezoey: `options` is missing.

Also `returns` is present but empty.

lib/helpers/lean/parent.js
hasezoey, line 1: The functions in this file likely should have some JSDoc.

hasezoey (commented):
Mostly LGTM

AbdelrahmanHafez (approved):

PR #16249, opened 2026-04-28
perf(document): speed up $__reset() in saveSimple insert case

vkarpov15:
**Summary**

Creating `$__.backup` and running `$__resetAtomics()` add a non-trivial amount of overhead that can be avoided:

1. `$__resetAtomics()` is not necessary on schemas that only have primitives
2. `$__.backup` isn't necessary in the case where we're inserting the new document, because we don't use change tracking to determine which fields to save when inserting a new document in `save()`. Because `save()` currently doesn't call `undoReset()` if inserting a new document fails, `$__.backup` isn't used.

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -3603,7 +3603,7 @@ Document.prototype.$isValid = function(path) {
  * @instance
  */
 
-Document.prototype.$__reset = function reset() {
+Document.prototype.$__reset = function reset(options) {
   let _this = this;
 
   const onlyPrimitiveValues = this.$__hasOnlyPrimitiveValues();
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (commented):
I think worth adding tests to prevent accidentally regressing these perf gains in the future, a matrix of tests with sinon spies that assert the behavior with only primitives, schemas with non-primitives, new docs (with subdocs re Copilot's comment), and existing docs.

Also curious about the amount of perf gains here if you've run some benchmarks.
Other than that, and Copilot's comments, LGTM.

hasezoey (commented):
LGTM, though i agree with Hafez and Copilot already suggested

vkarpov15:
Closing this in favor of #16331

PR #16251, opened 2026-04-29
types(objectid): fix _id getter helper on ObjectId

noseworthy:
**Summary**

When we removed the dependency on `bson` in
https://github.com/Automattic/mongoose/pull/15576, we inadvertently broke the augmentation added in
https://github.com/Automattic/mongoose/pull/13515 to capture the default option that adds a helper `_id` property on `ObjectId` values.

Updating the augmentation to reference the `ObjectId` type exported from `mongodb` instead of `bson` fixes the augmentation.

### DIFF
--- a/test/types/mongo.test.ts
+++ b/test/types/mongo.test.ts
@@ -1,5 +1,5 @@
 import * as mongoose from 'mongoose';
-import * as bson from 'bson';
+import * as mongodb from 'mongodb';
 import { expect } from 'tstyche';
 
 function gh12537() {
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
This might fix this issue, but in runtime, if i recall correctly, mongoose actually *does* modify the bson `ObjectId` as mongodb just re-exports from bson.

noseworthy:
Sort of?

https://github.com/Automattic/mongoose/blob/master/lib/types/objectid.js#L13

Since https://github.com/Automattic/mongoose/commit/1556421ed04466e09d9e32f85297607ddf4ab775, `bson` was removed as a direct dependency, so it's modifying the `bson` `ObjectId` through `mongodb` and re-exports it from `mongodb` not from `bson` anymore.

Or am I wrong here?

vkarpov15 (approved):

noseworthy:
Confirmed `9.6.1` fixes this issue for me! Thanks folks! 🙏

PR #16263, opened 2026-05-02
chore(deps): add Dependabot cooldown

AbdelrahmanHafez:
This PR adds a 2-day cooldown for dependabot updates, matching the `.npmrc` `min-release-age=2` policy.

Node 24 CI uses npm 11, which enforces `min-release-age`. A freshly published package can be selected by Dependabot before npm will install it, causing dependency update PRs like #16262 to fail until the cooldown expires.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
   directory: "/"
   schedule:
     interval: "monthly"
+  cooldown:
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Works for me :+1: Just need to remember to bump this if we bump min-release-age

PR #16274, opened 2026-05-07
BREAKING CHANGE: require passing both filter and update to Query.prototype.updateOne(), make Document.prototype.init() fully sync

### LINKED ISSUE #16269: Remove support for `query.updateOne(update)` in favor of `query.updateOne(filter, update)` for consistency with TypeScript types
### Prerequisites

### Mongoose version

9.x.x

### Node.js version

22.x

### MongoDB server version

8.x

6.x.x

### Description

Types for using Query modification methods with a single arg (`Query.updateMany(doc)` etc.) were removed in 5dc43e0c222b049864b0caddde4efeeac8247c05, but these signatures still seem to be supported and in some cases documented.

### Steps to Reproduce

N/A

### Expected Behavior

Ideally signatures should be typed, or documentation updated to reflect these not being recommended/supported.

vkarpov15:
Fix #16269

**Summary**

Remove some more callback-based internals: `Document.prototype.init()` has no reason to support callbacks it is fully sync. Likewise, logic that uses it, like `completeOne()`, doesn't necessarily need to be async unless there's `populate()` involved.

Also, for consistency with our TypeScript types and Model.updateX(); `Query.prototype.updateOne()`, `updateMany()`, and `replaceOne()` now require both a `filter` and `update`/`replacement` argument. To avoid any confusion with our legacy behavior of `Query.prototype.updateOne(obj)` treating `obj` as a `update` not a `filter`.

**Examples**

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -4315,7 +4315,7 @@ Query.prototype._replaceOne = async function _replaceOne() {
  */
 
 Query.prototype.updateMany = function updateMany(filter, update, options) {
-  if (typeof filter === 'function' || typeof update === 'function' || typeof options === 'function') {
+  if (typeof filter === 'function' || typeof update === 'function' || typeof options === 'function' || typeof arguments[3] === 'function') {
     throw new MongooseError('Query.prototype.updateMany() no longer accepts a callback');
   }
   if (arguments.length < 2) {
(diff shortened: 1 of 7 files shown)


lib/query.js
```diff
@@ -4363,32 +4314,15 @@ Query.prototype._replaceOne = async function _replaceOne() {
...
-  } else if (typeof conditions === 'object' && !doc && !options && !callback) {
-    // .update(doc)
-    doc = conditions;
-    conditions = undefined;
-    options = undefined;
-    callback = undefined;
-  }
-
-  return _update(this, 'updateMany', conditions, doc, options, callback);
+Query.prototype.updateMany = function updateMany(conditions, doc, options) {
```
hasezoey: The argument names should likely match the argument names in the JSDOC

hasezoey (changes requested):
Sounds like a good change, but the Copilot comments should likely be investigated, also slight style suggestion.

AbdelrahmanHafez (approved):

hasezoey (approved):

PR #16275, opened 2026-05-07
feat: Add tracing channel support

### LINKED ISSUE #16105: `TracingChannel` Proposal for observability
### Prerequisites

### 🚀 Feature Proposal

I'd like to propose adding first-class [`TracingChannel`](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel) support to `mongoose`, following the pattern established by [`undici`](https://github.com/nodejs/undici) in Node.js core.

### Motivation

Mongoose already has a powerful pre/post middleware (hook) system that covers document, query, aggregate, and model operations. 

However, APM instrumentation libraries **cannot use these hooks** for span-based tracing today. Instead, they resort to monkey-patching `Query.prototype.exec`, `Aggregate.prototype.exec`, `Model.prototype.save`, and 12+ other methods via IITM/RITM. Here's why:

1. **No async context propagation.** Mongoose hooks run in the middleware chain's execution context, not the caller's `AsyncLocalStorage` context. This is a documented pain point: [#10020](https://github.com/Automattic/mongoose/issues/10020) reported post-hooks not working with `AsyncLocalStorage`, and [#10478](https://github.com/Automattic/mongoose/issues/10478) identified the root cause as an underlying library issue where async context is lost across the hook chain. While these issues have been resolved, they illustrate that async context propagation through the middleware system has been a persistent challenge, and it's exactly the problem `TracingChannel` was designed to solve at the platform level.

2. **Per-schema registration.** Hooks must be registered on each schema individually. APM tools need global instrumentation that covers all models without requiring user configuration. There's no single attachment point for cross-cutting tracing concerns.

3. **No lifecycle correlation.** Pre and post hooks are separate middleware functions. To build a span, APM tools would need to correlate the pre-hook (span start) with the post-hook (span end) and manage span storage manually. `TracingChannel` provides a unified lifecycle for the operation, so correlation is built-in.

Beyond these APM-specific gaps, the current monkey-patching approach has broader ecosystem concerns:

- **Runtime lock-in:** RITM and IITM rely on Node.js-specific module loader internals (`Module._resolveFilename`, `module.register()`). They don't work on Bun or Deno, which implement the Node.js API surface but not the module loader internals.
- **ESM fragility:** IITM is built on Node.js's module customization hooks, which are still evolving and have been a persistent source of breakage in the OTEL JS ecosystem.
- **Initialization ordering:** Both require instrumentation to be set up before `mongoose` is first `require()`'d / `import`'d.
- **Bundling:** Users must ensure instrumented modules are externalized, which is increasingly difficult as frameworks bundle server-side code into single executables or deployment files.

`TracingChannel` solves all of these. It provides structured lifecycle events (`start`, `end`, `asyncStart`, `asyncEnd`, `error`) with built-in async context propagation, zero-cost when no subscribers are attached, and a standardized subscription model that requires no monkey-patching.

### Example

I propose implementing the following tracing channels:

| TracingChannel | Tracks | Context fields |
|---|---|---|
| `mongoose:query` | All `Query.exec()` operations — find, update, delete, count, distinct, etc. | `operation`, `collection`, `query`, `fields`, `options`, `database`, `serverAddress`, `serverPort` |
| `mongoose:aggregate` | `Aggregate.exec()` pipeline execution | `pipeline`, `collection`, `options`, `database`, `serverAddress`, `serverPort` |
| `mongoose:save` | `Model.prototype.save()` — document insert or update | `operation`, `collection`, `database`, `serverAddress`, `serverPort` |
| `mongoose:model` | Model-level statics — `insertMany`, `bulkWrite` | `operation`, `collection`, `database`, `serverAddress`, `serverPort` |

The usage will look something like this:

```ts
const dc = require('node:diagnostics_channel');

// Subscribe to query operations, covers ALL find/update/delete/count variants
dc.tracingChannel('mongoose:query').subscribe({
  start(ctx) {
    // TracingChannel automatically propagates this span as the active context
    // through the entire async operation, no manual context.with() needed
    ctx.span = tracer.startSpan(`${ctx.operation} ${ctx.collection}`, {
      attributes: {
        'db.system': 'mongodb',
        'db.operation.name': ctx.operation,
        'db.collection.name': ctx.collection,
        'db.query.text': sanitize(ctx.query),
        'db.namespace': ctx.database,
        'server.address': ctx.serverAddress,
        'server.port': ctx.serverPort,
      },
    });
  },
  asyncEnd(ctx) {
    // asyncEnd fires after the promise resolves — this is where the span ends
    ctx.span?.end();
  },
  error(ctx) {
    ctx.span?.setStatus({ code: SpanStatusCode.ERROR, message: ctx.error?.message });
    ctx.span?.recordException(ctx.error);
  },
});
```

The context payload can match what [OTEL currently has](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-mongoose).

I'm happy to help spec this out in a PR and see what you folks think.

## Prior Art

This approach follows the same pattern already adopted or in progress by other major libraries:

- **`undici`** (Node.js core) — ships `TracingChannel` support since Node 20.12: [`undici:request`](https://nodejs.org/api/diagnostics_channel.html#undici-channels)
- **`node-redis`** — [redis/node-redis#3195](https://github.com/redis/node-redis/pull/3195) (`node-redis:command`, `node-redis:connect`)
- **`ioredis`** — [redis/ioredis#2089](https://github.com/redis/ioredis/pull/2089) (`ioredis:command`, `ioredis:connect`)
- **`pg` / `pg-pool`** — [brianc/node-postgres#3624](https://github.com/brianc/node-postgres/pull/3624) (`pg:query`, `pg:connection`, `pg:pool:connect`)
- **`mysql2`** — [sidorares/node-mysql2#4178](https://github.com/sidorares/node-mysql2/pull/4178) (`mysql2:query`, `mysql2:execute`, `mysql2:connect`, `mysql2:pool:connect`)

logaretm:
This PR explores adding tracing channels to enable agnostic observability for users and APMs alike without having to resort to monkey patching or module patchers.

Each operation type has its own dedicated `TracingChannel`, so subscribers only pay overhead for the operations they care about. The channels are:

| Channel | Operations |
|---|---|
| `mongoose:query` | `find`, `findOne`, `updateOne`, `deleteOne`, `deleteMany`, `count`, `distinct`, etc. |
| `mongoose:aggregate` | `aggregate` |
| `mongoose:model:save` | `save` |
| `mongoose:model:insertMany` | `insertMany` |
| `mongoose:model:bulkWrite` | `bulkWrite` |
| `mongoose:cursor:next` | Cursor iteration (both query and aggregation cursors) |

### Payload

All channels emit the following context:

| Field | Type | Description |
|---|---|---|
| `operation` | `string` | The operation name (e.g. `find`, `updateOne`, `aggregate`) |
| `collection` | `string` | MongoDB collection name |
| `database` | `string` | Database name |
| `serverAddress` | `string` | DB host |
| `serverPort` | `number` | DB port |
| `args` | `Record<string, any>` | Operation-specific arguments |

I also exported the types for use by consumers.

### Example Consumer

```ts
const dc = require('node:diagnostics_channel');

// Subscribe to query operations only
const queryChannel = dc.tracingChannel('mongoose:query');

queryChannel.subscribe({
  start({ operation, collection, database, serverAddress, serverPort, args }) {
    console.log('[query:start]', operation, collection, args);
  },
  asyncEnd({ operation, collection, result }) {
    console.log('[query:end]', operation, collection);
  },
  error({ operation, collection, error }) {
    console.error('[query:error]', operation, collection, error);
  }
});
```

this is what it looks like after consuming the events in an APM, which is inline with current instrumentations today.

CleanShot 2026-05-07 at 13 08 23@2x

closes #16105

### DIFF
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -15,6 +15,7 @@ const { buildMiddlewareFilter } = require('./helpers/buildMiddlewareFilter');
 const stringifyFunctionOperators = require('./helpers/aggregate/stringifyFunctionOperators');
 const utils = require('./utils');
 const { modelSymbol } = require('./helpers/symbols');
+const { traceAggregate } = require('./tracing');
 const read = Query.prototype.read;
 const readConcern = Query.prototype.readConcern;
 
@@ -1090,32 +1091,42 @@ Aggregate.prototype.exec = async function exec() {
   prepareDiscriminatorPipeline(this._pipeline, this._model.schema);
   stringifyFunctionOperators(this._pipeline);
 
-  const preFilter = buildMiddlewareFilter(this.options, 'pre');
-  const postFilter = buildMiddlewareFilter(this.options, 'post');
+  const _this = this;
+  return traceAggregate(async function() {
+    const preFilter = buildMiddlewareFilter(_this.options, 'pre');
+    const postFilter = buildMiddlewareFilter(_this.options, 'post');
 
-  try {
-    await model.hooks.execPre('aggregate', this, [], { filter: preFilter });
-  } catch (error) {
-    return await model.hooks.execPost('aggregate', this, [null], { error, filter: postFilter });
-  }
+    try {
+      await model.hooks.execPre('aggregate', _this, [], { filter: preFilter });
+    } catch (error) {
+      return await model.hooks.execPost('aggregate', _this, [null], { error, filter: postFilter });
+    }
 
-  if (!this._pipeline.length) {
-    throw new MongooseError('Aggregate has empty pipeline');
-  }
+    if (!_this._pipeline.length) {
+      throw new MongooseError('Aggregate has empty pipeline');
+    }
 
-  const options = clone(this.options || {});
-  delete options.middleware;
+    const options = clone(_this.options || {});
+    delete options.middleware;
 
-  let result;
-  try {
-    const cursor = await collection.aggregate(this._pipeline, options);
-    result = await cursor.toArray();
-  } catch (error) {
-    return await model.hooks.execPost('aggregate', this, [null], { error, filter: postFilter });
-  }
+    let result;
+    try {
+      const cursor = await collection.aggregate(_this._pipeline, options);
+      result = await cursor.toArray();
+    } catch (error) {
+      return await model.hooks.execPost('aggregate', _this, [null], { error, filter: postFilter });
+    }
 
-  await model.hooks.execPost('aggregate', this, [result], { error: null, filter: postFilter });
-  return result;
+    await model.hooks.execPost('aggregate', _this, [result], { error: null, filter: postFilter });
+    return result;
+  }, () => ({
+    pipeline: _this._pipeline,
+    collection: collection.name,
+    options: _this.options,
+    database: model.db?.name,
+    serverAddress: model.db?.host,
+    serverPort: model.db?.port
+  }));
 };
 
 /**
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -15,7 +15,8 @@ const { buildMiddlewareFilter } = require('./helpers/buildMiddlewareFilter');
 const stringifyFunctionOperators = require('./helpers/aggregate/stringifyFunctionOperators');
 const utils = require('./utils');
 const { modelSymbol } = require('./helpers/symbols');
-const { aggregateChannel } = require('./tracing');
+const { createTracedChannel } = require('./tracing');
+const { trace: traceAggregate } = createTracedChannel('mongoose:aggregate');
 const read = Query.prototype.read;
 const readConcern = Query.prototype.readConcern;
 
@@ -1073,7 +1074,7 @@ Aggregate.prototype.exec = async function exec() {
     this._optionsForExec();
 
     const _this = this;
-    return aggregateChannel.trace(async function maybeTracedConnectionAggregate() {
+    return traceAggregate(async function maybeTracedConnectionAggregate() {
       const cursor = await _this._connection.client.db().aggregate(_this._pipeline, _this.options);
       return await cursor.toArray();
     }, () => ({
@@ -1104,7 +1105,7 @@ Aggregate.prototype.exec = async function exec() {
   stringifyFunctionOperators(this._pipeline);
 
   const _this = this;
-  return aggregateChannel.trace(async function maybeTracedAggregateExec() {
+  return traceAggregate(async function maybeTracedAggregateExec() {
     const preFilter = buildMiddlewareFilter(_this.options, 'pre');
     const postFilter = buildMiddlewareFilter(_this.options, 'post');
 
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -1092,7 +1092,7 @@ Aggregate.prototype.exec = async function exec() {
   stringifyFunctionOperators(this._pipeline);
 
   const _this = this;
-  return trace(async function() {
+  return trace(async function maybeTracedAggregateExec() {
     const preFilter = buildMiddlewareFilter(_this.options, 'pre');
     const postFilter = buildMiddlewareFilter(_this.options, 'post');
 
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -1072,8 +1072,20 @@ Aggregate.prototype.exec = async function exec() {
 
     this._optionsForExec();
 
-    const cursor = await this._connection.client.db().aggregate(this._pipeline, this.options);
-    return await cursor.toArray();
+    const _this = this;
+    return trace(async function maybeTracedConnectionAggregate() {
+      const cursor = await _this._connection.client.db().aggregate(_this._pipeline, _this.options);
+      return await cursor.toArray();
+    }, () => ({
+      operation: 'aggregate',
+      database: _this._connection.name,
+      serverAddress: _this._connection.host,
+      serverPort: _this._connection.port,
+      args: {
+        pipeline: _this._pipeline,
+        options: _this.options
+      }
+    }));
   }
 
   const model = this._model;
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -15,7 +15,7 @@ const { buildMiddlewareFilter } = require('./helpers/buildMiddlewareFilter');
 const stringifyFunctionOperators = require('./helpers/aggregate/stringifyFunctionOperators');
 const utils = require('./utils');
 const { modelSymbol } = require('./helpers/symbols');
-const { traceAggregate } = require('./tracing');
+const { trace } = require('./tracing');
 const read = Query.prototype.read;
 const readConcern = Query.prototype.readConcern;
 
@@ -1092,7 +1092,7 @@ Aggregate.prototype.exec = async function exec() {
   stringifyFunctionOperators(this._pipeline);
 
   const _this = this;
-  return traceAggregate(async function() {
+  return trace(async function() {
     const preFilter = buildMiddlewareFilter(_this.options, 'pre');
     const postFilter = buildMiddlewareFilter(_this.options, 'post');
 
@@ -1120,12 +1120,15 @@ Aggregate.prototype.exec = async function exec() {
     await model.hooks.execPost('aggregate', _this, [result], { error: null, filter: postFilter });
     return result;
   }, () => ({
-    pipeline: _this._pipeline,
+    operation: 'aggregate',
     collection: collection.name,
-    options: _this.options,
     database: model.db?.name,
     serverAddress: model.db?.host,
-    serverPort: model.db?.port
+    serverPort: model.db?.port,
+    args: {
+      pipeline: _this._pipeline,
+      options: _this.options
+    }
   }));
 };
 
--- a/lib/aggregate.js
+++ b/lib/aggregate.js
@@ -15,7 +15,7 @@ const { buildMiddlewareFilter } = require('./helpers/buildMiddlewareFilter');
 const stringifyFunctionOperators = require('./helpers/aggregate/stringifyFunctionOperators');
 const utils = require('./utils');
 const { modelSymbol } = require('./helpers/symbols');
-const { queryChannel } = require('./tracing');
+const { aggregateChannel } = require('./tracing');
 const read = Query.prototype.read;
 const readConcern = Query.prototype.readConcern;
 
@@ -1073,7 +1073,7 @@ Aggregate.prototype.exec = async function exec() {
     this._optionsForExec();
 
     const _this = this;
-    return queryChannel.trace(async function maybeTracedConnectionAggregate() {
+    return aggregateChannel.trace(async function maybeTracedConnectionAggregate() {
       const cursor = await _this._connection.client.db().aggregate(_this._pipeline, _this.options);
       return await cursor.toArray();
     }, () => ({
@@ -1104,7 +1104,7 @@ Aggregate.prototype.exec = async function exec() {
   stringifyFunctionOperators(this._pipeline);
 
   const _this = this;
-  return queryChannel.trace(async function maybeTracedAggregateExec() {
+  return aggregateChannel.trace(async function maybeTracedAggregateExec() {
     const preFilter = buildMiddlewareFilter(_this.options, 'pre');
     const postFilter = buildMiddlewareFilter(_this.options, 'post');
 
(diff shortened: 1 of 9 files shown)


lib/query.js
```diff
@@ -4778,42 +4779,56 @@ Query.prototype.exec = async function exec(op) {
...
 
-  res = await _executePostHooks(this, res, error);
+    res = await _executePostHooks(_this, res, error);
 
-  await this._hooks.execPost('exec', this, []);
+    await _this._hooks.execPost('exec', _this, []);
 
-  return res;
+    return res;
+  }, () => ({
```
vkarpov15, line 4821: Given that `tracing()` just calls `contextFactory()`, why is `contextFactory` a function? I don't see any reason why we can't pass a `context` object to `tracing()`

lib/query.js
```diff
@@ -4778,42 +4779,56 @@ Query.prototype.exec = async function exec(op) {
...
-  let skipWrappedFunction = null;
-  try {
-    await this._hooks.execPre('exec', this, []);
-  } catch (err) {
-    if (err instanceof Kareem.skipWrappedFunction) {
-      skipWrappedFunction = err;
-    } else {
-      throw err;
+  const _this = this;
+  return trace(async function() {
```
vkarpov15: We prefer to avoid anonymous functions for more informative stack traces.

logaretm:
Copilot deciding not to wipe the floor with this PR with endless comments is a rare one.

lib/aggregate.js
vkarpov15: We would also need to wrap connection level aggregate

vkarpov15, line 1213: How do you recommend we handle cursors for the purposes of tracing channels? [Cursors](https://thecodebarbarian.com/cursors-in-mongoose-45) are effectively query iterators: they let you call `next()` to get the next batch of results to avoid loading the full result set into memory at once. Cursors can also be used as streams or with `for/await/of`. Should we wrap cursor next() in `trace()` as well?

lib/model.js
```diff
@@ -3371,13 +3394,26 @@ function _setIsNew(doc, val) {
  * @api public
  */
 
-Model.bulkWrite = async function bulkWrite(ops, options) {
+Model.bulkWrite = function bulkWrite(ops, options) {
   _checkContext(this, 'bulkWrite');
 
   if (typeof options === 'function' ||
```
vkarpov15, line 3403: What are the trade-offs of including these sort of error checks within the trace vs outside of the trace? For example, if user passes in a callback, as written that error would bypass tracing.

lib/tracing.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const dc = process.getBuiltinModule('node:diagnostics_channel');
+
+const hasTracingChannel = typeof dc.tracingChannel === 'function';
+
+function shouldTrace(channel) {
+  return !!channel && channel.hasSubscribers !== false;
```
vkarpov15, line 17: `channel.hasSubscribers !== false` is a little strange because it would pick up `channel.hasSubscribers === undefined`. What is the reasoning for making this `!== false`?

types/tracing.d.ts
```diff
@@ -0,0 +1,10 @@
+declare module 'mongoose' {
+  interface TracingContext {
+    operation: string;
+    collection: string;
+    database: string;
+    serverAddress: string;
+    serverPort: number;
+    args: Record<string, any>;
```
vkarpov15, line 8: Is there any benefit to including the result, or whether this operation reported an error?

vkarpov15 (commented):
This is very neat! With a little bit of work I was able to get some basic spans working with HyperDX:

Image

and desktop-otel-viewer:

Image

I just had some high level questions/concerns. Sorry if any of them seem nitpicky, I'm just trying to understand as much of the potential consequences as I can because I'm not an expert on APM.

lib/query.js
logaretm: It's an optimization, by taking advantage of the `hasSubscribers` flag we can lazily allocate the object only if there are subscribers actively listening. We worked with some other libs and this did affect performance significantly.

May not be worth it here, but generally it is a safe practice we just do anyways. WDYT?

lib/tracing.js
logaretm: It's a compatibility thing. So Node 18 doesn't ship the `hasSubscribers` flag, nor does `bun` which means we cannot tell if there is a subscriber.

So it boils down to what you want to do on those runtimes or versions:

- Treat undefined as `true` and just trace these calls, this means those runtimes won't have true zero overhead when no subscribers attached but it is still negligible.
- OR, treat undefined as `false` and basically we effectively only trace in the runtimes that do support it. But because of `bun` this option is very aggressive.

There is an alternative if you prefer. We can check the individual channels for subscribers:

```
// Bun-only fallback
return ['start', 'end', 'asyncStart', 'asyncEnd', 'error'].some(ch => {
  return queryChannel[ch].hasSubscribers;
})
```

Some libs did that, but I find the trade off here not that worth it. Especially with Node 20 being EOL, and the overhead being close to zero when there are no subs.

I can document it just in case so it doesn't change later.

types/tracing.d.ts
logaretm: It is [done automatically in both cases](https://arc.net/l/quote/kkrkxixs), so `end` and `asyncEnd` channels report the resolved result of the promise on their own. For errors, the `error` channel will fire with the error object as well. 

We have assertions for `error` being reported. Would you like me to add `result` assertions as well?

From a typing perspective, `error` is fine because it is always an error in the async paths. For `result` I don't think APMs can reliably type that since we won't be able to tell what query ran reliably, so they can sniff it with type checks in runtimes which will be needed anyways.

lib/aggregate.js
logaretm: Thanks! addressed.

logaretm: Maybe @qard will have some insights here, tracing channels lifecycle model doesn't work with async iterators because they don't exactly have a known start/finish signals and both the start and finish may happen in different async contexts.

so we have two options, either:

- Trace `next` calls, allowing APMs to create a span for each call or async tick in this case.
- Wrap cursors with a trace to cover the whole thing, but there is no guarantee the iterator will be fully consumed, so if an API creates a span for it, it may never end and will block any child spans from being reported depending on the implementation.

This is very similar to streams, tracing them is tricky because u either create a span right away and patch its end time, but u risk extending your buffer duration for an unknown time, blocking child spans. Or you create a span later when it is done and override the start time, but u risk creating a span without any trace relationship. Both aren't great options.

I would say we can set them aside for now, and once we figure out a way for it we can create a new channel for them. @qard, thoughts on this?

vkarpov15:
Thanks! I have a couple of more questions: https://github.com/Automattic/mongoose/pull/16275#discussion_r3236803960 , https://github.com/Automattic/mongoose/pull/16275#discussion_r3236811037 . Otherwise I think this should be ready to go for 9.7 release.

lib/model.js
logaretm: I think this is a design choice, I saw those as API validation so no actual work is being done but maybe you see it differently and want APMs to report those.

Up to you, I don't have strong opinions on this one.

logaretm:
Thanks for the review and comments, I replied to the last two!

lib/aggregate.js
vkarpov15: Given this, I think the best approach for now would be to wrap `next()`. So given `const cursor = await User.find().cursor()`, you could see how long each `next()` call took. Is there any downside to tracing `next()`?

Ideally, it would be great if the cursor itself was a span, and `next()` calls as sub-spans of the cursor itself. But you are right that there's no guarantee that the cursor will be fully consumed. For example, change streams are effectively cursors, and those can last for the entire application lifecycle.

logaretm: I don't think there are any downsides, and it really drives the point of all of this effort here: libraries should own their observability and expose what they believe is useful, APMs can then pick and choose.

I think we could create a specific channel for the `next` calls to allow APMs to be register their interest or skip it. Since I think cursors can be noisy, right?

Like a `mongoose:cursor:next` which leaves the door open to another channel in the future that overarches the entire cursor lifecycle if we figure out a way to do it. I will add the `next` tracing right now.

Qard: Yeah, iterators are not yet a solved problem for diagnostics_channel. There is work in nodejs/node#62641 to address that need now that async iterators are becoming a common interface for AI SDKs, and that can apply similarly to tracing libraries. If you _do_ want to do something with that now, I would suggest looking at that Node.js core PR and try to align with that.

lib/model.js
```diff
@@ -3008,13 +3019,25 @@ Model.startSession = function() {
  * @api public
  */
 
-Model.insertMany = async function insertMany(arr, options) {
+Model.insertMany = function insertMany(arr, options) {
```
Qard: This would result in the input validation error being thrown rather than rejected. This should probably be left as it was.

lib/model.js
Qard: You'll want these either inside the traced function or restore the async function marker as otherwise these will throw rather than reject.

lib/tracing.js
```diff
@@ -0,0 +1,41 @@
...
+function createTracedChannel(name) {
+  const ch = hasTracingChannel ? dc.tracingChannel(name) : undefined;
+
+  function trace(fn, contextFactory) {
+    if (!shouldTrace(ch)) {
+      return fn();
+    }
+
+    const traced = ch.tracePromise(fn, contextFactory());
+    traced.catch(noop);
```
Qard: Why are you catching and ignoring the errors here? Are you trying to silence the unhandled rejection on that for some reason?

Qard (commented):
Mostly seems fine, but I would suggest undoing the removal of the async keyword in a few spots, and you may want to consider splitting up to more channels by source so subscribers can be a bit more granular.

lib/model.js
logaretm: Good catch, I think I dropped it while refactoring from anon functions...

lib/model.js
logaretm: Done! good catch!

lib/tracing.js
logaretm: I noticed that `tracePromise` returns a new promise, so there is a chance that `unhandledRejection` is fired randomly if the promise rejects. Especially if there is more async work to be done after the traced call microtask, it's a defensive mechanism.

It doesn't silence the error as it is still returned to the user, if they handle it then it works, if they don't then it produces the warning as it would in their untraced path.

What do you think?

logaretm:
> you may want to consider splitting up to more channels by source so subscribers can be a bit more granular.

Do you have channels in mind? the original proposal had:

- `mongoose:query` 
- `mongoose:aggregate` 
- `mongoose:save` 
- `mongoose:model` (`insertMany` and `bulkWrite`)

Which felt like too much and vague sometimes on what the grouping needed to be for which operations. Do you have ideas on mind how to split them up? Right now there is a `operation` field which APMs can use to filter out what they don't want to handle but will still incur the overhead.

lib/tracing.js
Qard: Personally, I'd leave it out, but I'm not against it. In my opinion the `unhandledRejection` event handling is a bit fragile, so I would not be against trying to silence it.

Qard:
I'd basically just raise the operation (whatever it may be) to be the channel name and allow listening to whichever operations one cares about. It's slightly more effort to listen to more channels, but it's a lot cheaper than branching/filtering on every message. The diagnostics_channel design with channel objects is intended to encourage making lots of granular channels and letting you subscribe to many different channels with the same subscriber if you want to. The subscriber receives the channel name as the second parameter, so you can filter with that when you use the same subscriber function on many channels. You only pay for as many channels as you subscribe to though, and you don't need to pay _any_ cost if you only subscriber to _one_ channel, which is presumed to be the default.

logaretm:
I just split operations into dedicated `TracingChannels` so subscribers only pay overhead for what they
  listen to:

  - `mongoose:query` Query.exec (find, findOne, updateOne, deleteOne, etc.)
  - `mongoose:aggregate` Aggregate.exec
  - `mongoose:model:save` document save
  - `mongoose:model:insertMany` insertMany
  - `mongoose:model:bulkWrite` bulkWrite
  - `mongoose:cursor:next` cursor iteration (query + aggregation)

@vkarpov15 Do you prefer this current split or did you think the previous state was better? The argument here is with fine grained channels, overhead is only incurred for the channels that the APM is interested in.

lib/tracing.js
vkarpov15: I would also remove this `.catch(noop)` because we're returning the full trace promise to the user, so the user can handle any promise rejections themselves. `catch(noop)` is only useful if the final promise isn't accessible by the user.

lib/aggregate.js
vkarpov15: Is the `wrapIterator()` function the logic you're talking about in this PR? If so, that looks like it's doing something similar, treating each next promise as a separate event.

vkarpov15 (approved):
Looks good to merge, I'm looking forward to trying this out in prod.

PR #16278, opened 2026-05-11
fix(update): preserve $setOnInsert.createdAt with overwriteImmutable: true

AbdelrahmanHafez:
`applyTimestampsToUpdate` only checked top-level and `$set` for a user-provided `createdAt`, so bulkWrite upserts with `$setOnInsert: { createdAt }` and `overwriteImmutable: true` silently overwrote the value with `now`. Including `$setOnInsert` in the check preserves it.

Also cleans up #15781 tests added in #15819:
- `timestamps` was placed outside the op descriptor in the loop tests, so all four `[true, false, null, undefined]` iterations ran with the default. Moved inside.
- Loop tests now assert `updatedAt` actually varies with `timestamps`. Previously assertions were invariant, so a regression would still pass.
- Removes a duplicate of the describe block.


hasezoey (commented):
LGTM

vkarpov15 (approved):

PR #15958, opened 2026-01-11
fix(clone): fix parent doc for map subdocuments and array subdocuments

AbdelrahmanHafez:
Fixes #15954

re https://github.com/Automattic/mongoose/issues/15901 https://github.com/Automattic/mongoose/pull/15904

PR #16281, opened 2026-05-13
fix(clone): isolate cloned arrays from source documents

AbdelrahmanHafez:
re #15958

`$clone()` was still leaving one parent pointer connected to the source document. We fixed the parent pointers on cloned subdocs, but retained Mongoose arrays were still constructed with the source document as their internal array parent. That meant cloning could mark the original document modified while populating the clone, and later writes to a cloned array could bubble `markModified()` back to the original.

This PR:
- uses the cloned parent doc when constructing retained Mongoose arrays
- populates cloned document arrays without marking either document modified
- restamps cloned array subdoc indexes while doing that, so stale indexes from prior array edits don't leak into the clone
- adds regression coverage for document arrays and primitive arrays

### DIFF
--- a/lib/helpers/clone.js
+++ b/lib/helpers/clone.js
@@ -202,10 +202,15 @@ function cloneArray(arr, options) {
 
   let ret = null;
   if (options?.retainDocuments) {
+    // Use the cloned parent doc (options.parentDoc) when available so the new
+    // array is wired up to the clone, not the source. Falling back to the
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
LGTM

vkarpov15 (approved):
This is great, thanks :+1:

vkarpov15:
I also cherry-picked this into 8.x branch.

PR #16282, opened 2026-05-13
fix(documentarray): reindex subdocs after reordering

AbdelrahmanHafez:
Document array operations that remove or reorder entries can leave surviving subdocs with stale `__index` values. After operations like `pull()`, `shift()`, `splice()`, `$shift()`, `unshift()`, positioned `push()`, `sort()`, or `reverse()`, a subdoc can still think it lives at its previous array index. A later nested change then marks the wrong path, and since the delta serializes the value from whatever subdoc currently sits at that stale index, the update writes a no-op to the wrong element and the user's change is silently lost:

```js
const user = await User.create({
  addresses: [{ city: 'Boston' }, { city: 'Chicago' }, { city: 'Denver' }]
});
user.addresses.pull(user.addresses[0]._id);
await user.save();

user.addresses[0].city = 'New York';
user.$getChanges(); // { $set: { 'addresses.1.city': 'Denver' } }, 'New York' never reaches the DB
```

This PR:
- restamps document-array subdoc indexes after operations that can move existing subdocs: `pull()`, `shift()`, `$shift()`, `splice()`, `unshift()`, positioned `push()`, `sort()`, `reverse()`
- keeps append-only `push()` on the fast path, the new subdoc is already cast with the correct appended index and existing indexes can't move, so repeated appends don't become O(n²)
- refreshes the parent populated cache in operations that previously skipped it: `$shift()`, `unshift()`, `sort()`, `reverse()`, `pop()`, `$pop()`
- makes document-array `reverse()` register a full-array `$set`, matching `sort()`/`splice()`/`unshift()` and avoiding stale pending atomics after a prior `$push`
- adds regression coverage asserting the exact `$getChanges()` and `modifiedPaths()` after each operation, persistence after save, populated-cache order, and reverse atomics
- fixes the MDN links in the touched doc comments, `shift()` was pointing at the `unshift` article

`addToSet()`, `nonAtomicPush()`, and `set()` have the same populated cache gap, I'll send that as a follow-up PR once this one lands.


hasezoey (commented):
LGTM

lib/types/documentArray/methods/index.js
```diff
@@ -389,6 +483,26 @@ const methods = {
...
+/*!
+ * ignore
+ */
+
+function _updateSubdocIndexes(arr) {
+  const rawArray = utils.isMongooseArray(arr) ? arr.__array : arr;
+
+  for (let i = 0; i < rawArray.length; ++i) {
+    if (typeof rawArray[i]?.$setIndex === 'function') {
+      rawArray[i].$setIndex(i);
```
vkarpov15, line 495: There's a related error with `$setIndex()`: `$setIndex()` adds new validation error keys, but retains existing ones. For example:

```
user.addresses[1].invalidate('city', 'invalid city');
user.addresses.sort((a, b) => a.sequence - b.sequence);
user.$__.validationError.errors; // Has both `addresses.1.city` and `addresses.0.city` presuming that the invalid city was sorted into position 0
```

vkarpov15 (approved):
I spotted one issue to follow up on, but otherwise LGTM this is a meaningful improvement

vkarpov15:
Gonna squash this to make it easier to backport to 8.x.

PR #16285, opened 2026-05-14
chore: update `queries` type test

mrazauskas:
**Summary**

While I was working on type tests, a "Module '"...mongodb"' has no default export." error surfaced unexpectedly. (Seems like accidentally removing `"allowSyntheticDefaultImports": true` from TSConfig triggered that.)

Interestingly all other type tests either use start (`import * as mongodb from 'mongodb'`) or named import. So perhaps it is good idea to refactor `queries` type test in similar fashion.


vkarpov15 (approved):
Thanks :+1:

PR #16295, opened 2026-05-25
docs: update default maxPoolSize on createConnection()

torbiak-samdesk:
**Summary**

Looks like this reference to the old default maxPoolSize of 5 was missed when the docs were updated for v6.

**Examples**

N/A


lib/mongoose.js
```diff
@@ -398,7 +398,7 @@ Mongoose.prototype.get = Mongoose.prototype.set;
...
+ * @param {number} [options.maxPoolSize=100] The maximum number of sockets the MongoDB driver will keep open for this connection. Keep in mind that MongoDB only allows one operation per socket at a time, so you may want to increase this if you find you have a few slow queries that are blocking faster queries from proceeding. See [Slow Trains in MongoDB and Node.js](https://thecodebarbarian.com/slow-trains-in-mongodb-and-nodejs).
  * @param {number} [options.minPoolSize=1] The minimum number of sockets the MongoDB driver will keep open for this connection. Keep in mind that MongoDB only allows one operation per socket at a time, so you may want to increase this if you find you have a few slow queries that are blocking faster queries from proceeding. See [Slow Trains in MongoDB and Node.js](https://thecodebarbarian.com/slow-trains-in-mongodb-and-nodejs).
```
AbdelrahmanHafez: Looks good, the driver default has been 100 since the v4 driver landed in v6. Verified locally that `createConnection()` actually opens with `maxPoolSize: 100`.

One thing while we're here: the `minPoolSize=1` on the next line of the same block is also stale. The real default is `0` (I checked, and it matches the `connect()` JSDoc just below which already says `minPoolSize=0`). Looks like the same v6 update that missed `maxPoolSize` missed this one too. Mind fixing it in the same PR?

AbdelrahmanHafez (changes requested):

lib/mongoose.js
torbiak-samdesk: Yeah, sounds good. Updated.

AbdelrahmanHafez (approved):
LGTM.
Thanks for improving mongoose!

PR #16298, opened 2026-05-26
docs: rewrite documents docs

vkarpov15:
**Summary**

I've always felt like our existing [Documents guide](https://mongoosejs.com/docs/documents.html) is one of the weakest pages on the documentation, so I went ahead and rewrote it. I tried to highlight some more common pain points and points of confusion:

1. Mongoose documents are not POJOs - what is a hydrated document?
2. Cast errors being stored internally - cast errors don't surface until validation
3. More info on how `save()` works especially with change tracking
4. More guidance on middleware, especially with regards to `updateOne()` etc. not triggering `save()` middleware

**Examples**


docs/documents.md
```diff
@@ -130,81 +170,182 @@ doc4.nested.subdoc; // Empty object
...
-There are 2 different ways to overwrite a document (replacing all keys in the
-document). One way is to use the
-[`Document#overwrite()` function](api/document.html#document_Document-overwrite)
-followed by `save()`.
+Document [middleware](https://mongoosejs.com/docs/middleware.html) lets you run code during key parts of a document's lifecycle.
+The most common document middleware hooks are for `validate()` and `save()`.
+
+At a high level, saving a document looks like this:
+
+<img src="https://res.cloudinary.com/drfhhq8wu/image/upload/v1779818614/bf1e3e29-7554-4ff9-9564-f631ed3eef98_ac2cb2.png" alt="save document lifecycle diagram">
```
hasezoey: The inline-codeblocks in this image still show backticks, is this intentional?

docs/documents.md
```diff
@@ -130,81 +170,182 @@ doc4.nested.subdoc; // Empty object
...
+However, use middleware sparingly.
+Middleware is usually not a good place for complex application logic.
+In particular, we recommend avoiding code in middleware that:
+
+* makes network calls to unrelated services
+* depends heavily on request-specific context
+* performs expensive work that should be explicit
+
+Also note that document middleware only runs for document operations.
+Query updates like `updateOne()` and `findOneAndUpdate()` do **not** run `save()` middleware.
```
hasezoey, line 345: Maybe a link to all the event names and where they get triggered would be good here.

docs/documents.md
```diff
@@ -1,87 +1,127 @@
...
 </ul>
 
-## Documents vs Models {#documents-vs-models}
+## What is a Document?
 
-[Document](api/document.html#Document) and [Model](api/model.html#Model) are distinct
-classes in Mongoose. The Model class is a subclass of the Document class.
-When you use the [Model constructor](api/model.html#Model), you create a
-new document.
+A Mongoose [document](api/document.html#Document) is an instance of a [Model](api/model.html#Model).
```
hasezoey:
```suggestion
A Mongoose [document](https://mongoosejs.com/docs/api/document.html#Document) is an instance of a [Model](https://mongoosejs.com/docs/api/model.html#Model).
```

Links should be changed to be consistently relative (`/api/...`) or always start with the root (`https://mongoosejs.com/...`)

hasezoey (changes requested):

docs/documents.md
vkarpov15: What do you mean by this - list all middleware functions that developers can hook into? I'm not sure the documents documentation is the right place for that.

docs/documents.md
vkarpov15: Not intentional, fixed

image

docs/documents.md
```diff
@@ -1,87 +1,127 @@
...
+Mongoose documents represent individual documents stored in a MongoDB collection.
+For example, documents created from the `User` model in the above example are stored in the `users` collection by default.
+
+You can create a new document using `new User()` or `await User.create()`; or load an existing document from MongoDB using queries like `findOne()`.
+
+```javascript
+const doc = new User({ name: 'John Smith' });
+// Persist the document to MongoDB. Calling `new User()` creates a new document in memory:
+// the document is not persisted to MongoDB until you call `save()`.
+await doc.save();
```
AbdelrahmanHafez, line 46: I think adding a separate comment first before `new User` reads a bit easier.
```suggestion
// Create a new document in memory
const doc = new User({ name: 'John Smith' });
// Persist the document to MongoDB:
// the document is not persisted to MongoDB until you call `save()`.
await doc.save();
```

docs/documents.md
```diff
@@ -130,81 +170,182 @@ doc4.nested.subdoc; // Empty object
...
+
 const Person = mongoose.model('Person', schema);
 
-const p = new Person({ name: 'foo', age: 'bar' });
-// Cast to Number failed for value "bar" at path "age"
-await p.validate();
+const doc = new Person({ age: -1 });
+
+await doc.validate();
+// Path `age` (-1) is less than minimum allowed value (0)
```
AbdelrahmanHafez: Moved comment on top, and prefixed with `Throws an error:`
```suggestion
// Throws an error "Path `age` (-1) is less than minimum allowed value (0)"
await doc.validate();
```

AbdelrahmanHafez (changes requested):
Dumping my thought process as I try to read this from a fresh reader's perspective:

----

> Mongoose documents are JavaScript objects backed by MongoDB data

I'm worried this could be misinterpreted as POJOs (the actual POJOs with the data in the mongo server). The next sentence and the _Hydrated Documents vs Lean Documents_ section clarify it, but I think it's worth avoiding the misunderstanding in the first place. I'd change `JavaScript objects` to `Document class instances`. They're technically the same thing, but one can be mistaken for POJOs.

----

> They are instances of Mongoose's Document class
>
> ...later, in "What is a Document":
>
> A Mongoose document is an instance of a Model. When you instantiate a model, you create a new document.

This is a bit confusing on first read: are they instances of a Document or a Model? How do Model and Document relate?

Do we assume the reader of this page has already read the Models doc, or should this page stand on its own without having to go somewhere else to understand the relationship?

----

> You can use the lean() method to make Mongoose queries return POJOs instead of hydrated documents.

Would be nice to briefly explain when I'd reach for `lean()` over hydrated docs, the performance benefit for read-only cases, and the drawbacks. If it's covered in detail elsewhere we can link to it, or just re-explain here.

----

Nitpick on the _Setting Nested Properties_ code example:

```js
const schema = new Schema({
  nested: {
    subdoc: new Schema({
      name: String
    })
  }
});
const TestModel = mongoose.model('Test', schema);

const doc = new TestModel();
doc.set('nested.subdoc.name', 'John Smith');
doc.nested.subdoc.name; // 'John Smith'
```

A newcomer might think `subdoc` and `nested` have meaningful distinction rather than just being generic path names. I'd rewrite this with a realistic example. More generally: using different realistic examples across the file helps people who learn by pattern recognition see what's constant and what's variable.

----

On this snippet, we tell readers not to use nullish coalescing assignments without explaining why it works in the first case but not the second:

```js
// The following works fine
const doc3 = new TestModel();
doc3.nested.subdoc ??= {};
doc3.nested.subdoc.name = 'John Smythe';

// The following does **NOT** work.
// Do not use the following pattern with Mongoose documents.
const doc4 = new TestModel();
(doc4.nested.subdoc ??= {}).name = 'Charlie Smith';
doc4.nested.subdoc; // Empty object
doc4.nested.subdoc.name; // undefined.
```

I honestly haven't seen this pattern used in the wild. We should either explain more (why it works in the first case, why not in the second, ideally the root cause), or drop it entirely given how rare it is. I'm leaning towards keeping it but explaining more.

----

Very nitpicky: the `save()` lifecycle being an image looks a bit off, especially on the dark theme. Probably fine for now, but might be worth figuring out how to inject custom HTML into the docs later if we don't have an easy way to do it.

docs/documents.md
hasezoey: Now that you mention it, yeah it likely is not the correct place. Also the link i meant was to `https://mongoosejs.com/docs/middleware.html`, which is already covered above, just below the section header.

hasezoey:
> Very nitpicky: the save() lifecycle being an image looks a bit off, especially on the dark theme. Probably fine for now, but might be worth figuring out how to inject custom HTML into the docs later if we don't have an easy way to do it.

Maybe we could consider switching the image to use [Mermaid](https://mermaid.live/edit)? (it is natively supported in github comments and github markdown display, though we would likely need another dependency for the website OR statically generate the html once)

vkarpov15:
"A newcomer might think subdoc and nested have meaningful distinction rather than just being generic path names." <-- they do have a meaningful distinction, but just not in this context, so you're right to call that out as a red herring.

AbdelrahmanHafez:
> Maybe we could consider switching the image to use [Mermaid](https://mermaid.live/edit)? (it is natively supported in github comments and github markdown display, though we would likely need another dependency for the website OR statically generate the html once)

My preference here would be some native HTML, gives us a lot more control over the styling, but Mermaid also works.
For the sake of experimenting, here's how Mermaid diagram would look like:
```mermaid
---
title: "save() Lifecycle"
---
flowchart TB
A("`**🛡️ 1. Pre-Validate Middleware**
Runs pre('validate') middleware.
Use this to prepare or modify the document before validation.`")
B("`**🔍 2. Validate**
Mongoose casts values and runs built-in validators and custom validators.
_If validation fails, save() is aborted and no changes are written._`")
C("`**✅ 3. Post-Validate Middleware**
Runs post('validate') middleware.
Runs whether validation succeeds or fails.
Use this for logging, metrics, etc.`")
D("`**⚙️ 4. Pre-Save Middleware**
Runs pre('save') middleware.
Use this for tasks that should run right before the document is persisted (e.g. hashing a password, setting timestamps).`")
E[("`**🗄️ 5. Write to MongoDB**
Mongoose generates the appropriate write operation:
• New document: insertOne()
• Existing document: updateOne() with only the modified paths`")]
F("`**📥 6. Post-Save Middleware**
Runs post('save') middleware.
Runs after the write attempt.
The document is now up to date with what's in MongoDB.`")

A --> B --> C --> D --> E --> F

classDef purple fill:#EDE7F6,stroke:#7E57C2,stroke-width:2px,color:#311B92;
classDef green fill:#E8F5E9,stroke:#43A047,stroke-width:2px,color:#1B5E20;
classDef orange fill:#FFF3E0,stroke:#FB8C00,stroke-width:2px,color:#E65100;
classDef blue fill:#E3F2FD,stroke:#1E88E5,stroke-width:2px,color:#0D47A1;

class A,C,F purple;
class B green;
class D orange;
class E blue;
```

vkarpov15:
@AbdelrahmanHafez how do you like this dark theme image?

image

docs/documents.md
```diff
@@ -1,210 +1,371 @@
...
+* change tracking
+* validation
+* middleware
+* document methods
+* `save()`
+* getters, setters, and virtuals
 
-When you load documents from MongoDB using model functions like [`findOne()`](api/model.html#model_Model-findOne),
-you get a Mongoose document back.
+That means some JavaScript object operations behave differently with documents.
```
IslandRhythms: Might be helpful to list what exact operations are and are not supported.

docs/documents.md
```diff
@@ -1,210 +1,371 @@
...
+
+**Validation** is a separate step that runs when you call the document's `validate()` method.
+The `save()` method calls `validate()` internally, so `save()` also triggers validation.
+Validation checks whether the document satisfies schema rules, which includes checking for cast errors, but also:
+
+* [required](https://mongoosejs.com/docs/api/schematype.html#SchemaType.prototype.required())
+* [min](https://mongoosejs.com/docs/api/schemanumber.html#SchemaNumber.prototype.min())
+* [max](https://mongoosejs.com/docs/api/schemanumber.html#SchemaNumber.prototype.max())
+* [enum](https://mongoosejs.com/docs/api/schemastring.html#SchemaString.prototype.enum())
+* custom validators
```
IslandRhythms: why is this not elaborated on?

IslandRhythms (approved):
Left a couple comments but LGTM otherwise

AbdelrahmanHafez (approved):

PR #16308, opened 2026-05-28
feat(model): add standard schema adapter for models

### LINKED ISSUE #16280: Standard Schema support
### Prerequisites

### 🚀 Feature Proposal

https://standardschema.dev/schema

### Motivation

Got a request for this at an in-person meetup. A good idea - make it easier to reuse Zod etc. schemas for Mongoose or vice versa.

### Example

Need to figure this out first, hopefully it would be similar to our existing `toJSONSchema()` methods, maybe a `toStandardSchema()` method?

vkarpov15:
Fix #16280

cc @ssalbdivad

**Summary**

Support for `~standard` validate adapter. See [standard schema docs](https://standardschema.dev/schema).

Outstanding questions:

1. **Casting:** is `User['~standard'].validate({ age: '42' })` ok given that Mongoose will convert string `'42'` to number `42` and return the result in `result`?

**Examples**


types/models.d.ts
```diff
@@ -108,6 +108,41 @@ declare module 'mongoose' {
    */
   type pathsToValidate = PathsToValidate;
 
+  export namespace StandardSchemaV1 {
+    export interface Props<Output = unknown> {
+      readonly version: 1;
+      readonly vendor: 'mongoose';
+      readonly validate: (
+        value: unknown,
+        options?: Options | undefined
```
hasezoey, line 117: The extra `| undefined` shouldnt be necessary due to the `?`, or did i mis-remember?

types/models.d.ts
```diff
@@ -108,6 +108,41 @@ declare module 'mongoose' {
...
+    export type Result<Output> = SuccessResult<Output> | FailureResult;
+
+    export interface SuccessResult<Output> {
+      readonly value: Output;
+      readonly issues?: undefined;
+    }
+
+    export interface FailureResult {
+      readonly issues: ReadonlyArray<Issue>;
+    }
```
hasezoey, line 134: Considering that `SuccessResult` has `issues: undefined`, maybe this one should have `value: undefined`?

test/model.validate.test.js
```diff
@@ -61,6 +61,82 @@ describe('model: validate: ', function() {
...
+    }, {
+      libraryOptions: { pathsToSkip: ['age'] }
+    });
+
+    assert.deepStrictEqual(skippedResult, {
+      value: {
+        name: 'Val',
+        age: 16
+      }
+    });
```
hasezoey, line 122: Maybe there should also be a test where paths are skipped but casting would be necessary?

hasezoey (approved):
Looks good to me from a mongoose perspective, i dont know if it is correct according to Standard Schema though.

types/models.d.ts
vkarpov15: Strange, you're right, I would have expected this to throw a compiler error under `exactOptionalPropertyTypes`. But nope.

However, I think it is worth keeping for consistency because that's how the [Standard Schema schema defines it](https://standardschema.dev/schema).

types/models.d.ts
vkarpov15: I agree (although maybe `value: never` would be better?) but again that's Standard Schema v1's definition, so keep for consistency.

image

test/model.validate.test.js
vkarpov15: That's a good point, I added a test which demonstrates that if `age` is string `'16'` here then Mongoose will still cast. And, similarly, if `age` is string `'not a number'`, then `validate()` will fail even though `pathsToSkip` has `age`.

However, given that `pathsToSkip` is a Mongoose-specific option that is explicitly tied to validation not casting, I don't think it is strictly necessary for us to change this behavior to be compliant with Standard Schema. Users could always strip out the `age` if they don't want Mongoose to try to cast it. WDYT @hasezoey @AbdelrahmanHafez ?

hasezoey: I think current behavior is fine, as long as it is documented.

lib/model.js
```diff
@@ -4302,6 +4313,8 @@ Model.validate = async function validate(obj, pathsOrOptions, context) {
 
 /**
  * Standard Schema adapter for this model.
+ * Calls [`Model.validate()`](#Model.validate()) internally with the provided `libraryOptions`
```
hasezoey, line 4316: ~~This link got messed up.~~

Nevermind, didnt see the closing `)`, still all links should start with `mongoosejs.com` as they are replaced automatically and so can work in jsdoc too.

Issue #16291, opened 2026-05-21
validateSync middleware cycle


AbdelrahmanHafez:
Is this intended or a bug?
```js
schema.pre('validate', () => ++pre);
schema.post('validate', () => ++post);

doc.validateSync(); // pre/post stay 0
await doc.validate(); // pre/post run
```

`doc.validateSync()` has a `middleware: false` option added in #15883 , but there's not validateSync middleware cycle at all right now.
I should either remove the option from the docs/TS, or add a validateSync synchronous middleware akin to `init`.

Thoughts? @vkarpov15 @hasezoey

vkarpov15:
What do you think about removing `validateSync()` altogether?

My original rationale for adding `validateSync()` was primarily validation against a schema in the browser. The secondary concern was more concise syntax (no callbacks) in cases where that was desirable. Both are less of a concern now - `doc.validateSync()` vs `await doc.validate()` isn't a big difference complexity-wise; and since we split off the Mongoose browser library into a [separate package](https://www.npmjs.com/package/@mongoosejs/browser) in Mongoose 9 the npm download numbers tell us that Mongoose usage in the browser is practically 0.

For now, I think `doc.validateSync()` not running middleware is intentional. We can remove that `middleware` option for consistency. But I'm starting to think that we should consider deprecating `validateSync()` precisely because it adds potential confusion for the validation API. For example, users may wonder whether `save()` uses `validate()` or `validateSync()`; or what other tradeoffs there are between `validate()` and `validateSync()`.

AbdelrahmanHafez:
Yeah sounds good. I think the plan would be three-fold:
1. Remove the middleware option for consistency for now.
2. Add deprecation warning and marker to the JSDoc in TS types and whatnot.
3. In 10.0 delete it altogether.

WDYT?

PR #16310, opened 2026-05-28
fix(document): deprecate validateSync re #16291

AbdelrahmanHafez:
re #16291
This PR:
- removes `middleware` support from the `validateSync()` TypeScript surface, because `validateSync()` does not run validate middleware
- marks `Document.prototype.validateSync()` as deprecated in the runtime JSDoc and TypeScript declarations
- emits a deprecation warning when `validateSync()` is called
- updates the deprecation docs and focused type/runtime coverage

Runtime validation behavior is otherwise unchanged.

### DIFF
--- a/lib/schema/union.js
+++ b/lib/schema/union.js
@@ -115,9 +115,6 @@ class Union extends SchemaType {
     if (value != null && typeof value.$__validateSync === 'function') {
       return value.$__validateSync();
     }
-    if (value != null && typeof value.validateSync === 'function') {
-      return value.validateSync();
-    }
   }
 
   clone() {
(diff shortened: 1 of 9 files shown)


lib/schema/union.js
```diff
@@ -112,6 +112,9 @@ class Union extends SchemaType {
         return schemaTypeError;
       }
     }
+    if (value != null && typeof value.$__validateSync === 'function') {
+      return value.$__validateSync();
+    }
     if (value != null && typeof value.validateSync === 'function') {
```
vkarpov15: Shouldn't we delete the validateSync() call?

vkarpov15 (commented):
One minor comment but looks good overall

vkarpov15 (approved):

PR #16322, opened 2026-06-04
Drop `lodash.isEqual` dev-dependency

hasezoey:
**Summary**

This PR drops the `lodash.isEqual` dependency in favor of nodejs native `deepStrictEqual` due to the package itself reporting to not be used anymore:

> npm warn deprecated lodash.isequal@4.5.0: This package is deprecated. Use require('node:util').isDeepStrictEqual instead.

Also changes:
- the `assert.ok(util.deepStrictEqual())` to just `assert.deepStrictEqual` as that should provide better errors on fail.
- Drop double compares of the same value

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "glob": "^13.0.6",
     "highlight.js": "11.11.1",
     "linkinator": "7.x",
-    "lodash.isequal": "4.5.0",
     "lodash.isequalwith": "4.4.0",
     "markdownlint-cli2": "0.22.1",
     "marked": "18.0.4",
--- a/test/model.findOneAndUpdate.test.js
+++ b/test/model.findOneAndUpdate.test.js
@@ -14,7 +14,6 @@ const utils = require('../lib/utils');
 const Schema = mongoose.Schema;
 const ObjectId = Schema.Types.ObjectId;
 const DocumentObjectId = mongoose.Types.ObjectId;
-const isEqual = require('lodash.isequal');
 const { isDeepStrictEqual } = require('util');
 const isEqualWith = require('lodash.isequalwith');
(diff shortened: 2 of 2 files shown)


AbdelrahmanHafez (approved):

vkarpov15:
This has been brought up a couple of times before: we have `lodash.isequal` as dependency specifically to test `lodash.isequal` because deprecated or not, that package gets 18M weekly downloads. I'm open to merging this PR if @hasezoey and @AbdelrahmanHafez feel strongly about it; but I've kept this dev dependency in an abundance of caution.

hasezoey:
> I'm open to merging this PR if @hasezoey and @AbdelrahmanHafez feel strongly about it; but I've kept this dev dependency in an abundance of caution.

I dont feel strongly about it, i just saw the deprecation warning and the git log about node 0.10's `deepStrictEqual` having had problems, but everything seemed to work now.

Due to it being deprecated and you wanting to keep if for caution, how about merging this for mongoose 10.0?

AbdelrahmanHafez:
I don't feel strongly about it, I'm slightly leaning towards keeping it. It's harmless and guarantees the integration.
Since it's a dev dependency, the warning only shows up for us installation and in our CIs, downstream users never see this warning.

Also removing it on 10.x is fine, since the native assertion should provide equivalent guarantees. Side-note: should probably treat `lodash.isequalwith` the same way we treat `lodash.isequal`, whichever way we decide to go.

vkarpov15 (approved):
I'll merge this into 10.0 branch in the interest of caution. Given that lodash.isequal is deprecated, we can't keep supporting it forever, so 10.0 is a good opportunity to drop support.

PR #16324, opened 2026-06-05
`validateFilter` function

### LINKED ISSUE #15581: { enum: true } option for queries or perhaps Model.validate
### Prerequisites

### 🚀 Feature Proposal

Currently, when you query for values not included in an enum, no warnings or errors are thrown. We have `strictQuery` which checks the keys, but we don't have something similar for values.

### Motivation

Sometimes when working on multiple project with similar designs, names can be confused i.e. `done` versus `completed` or `in_progress` vs `in-progress`

### Example

`TestModel.find(filter, { enum: true });`

IslandRhythms:
closes #15581

### DIFF
--- a/lib/cursor/queryCursor.js
+++ b/lib/cursor/queryCursor.js
@@ -98,7 +98,20 @@ function QueryCursor(query) {
     }
   };
 
-  model.hooks.execPre('find', query).then(() => onPreComplete(null), err => onPreComplete(err));
+  model.hooks.execPre('find', query).then(() => {
+    if (!query.options.validateFilter) {
+      onPreComplete(null);
+      return;
+    }
+
+    return query._castConditions().then(() => {
+      if (query.error()) {
+        onPreComplete(query.error());
+        return;
+      }
+      onPreComplete(null);
+    });
+  }, err => onPreComplete(err));
 }
 
 util.inherits(QueryCursor, Readable);
--- a/lib/cursor/queryCursor.js
+++ b/lib/cursor/queryCursor.js
@@ -104,7 +104,7 @@ function QueryCursor(query) {
       return;
     }
 
-    return query._castConditions().then(() => {
+    return query._validateFilterConditions().then(() => {
       if (query.error()) {
         onPreComplete(query.error());
         return;
--- a/lib/model.js
+++ b/lib/model.js
@@ -46,6 +46,7 @@ const createPopulateQueryFilter = require('./helpers/populate/createPopulateQuer
 const decorateUpdateWithVersionKey = require('./helpers/update/decorateUpdateWithVersionKey');
 const getDefaultBulkwriteResult = require('./helpers/getDefaultBulkwriteResult');
 const getSchemaDiscriminatorByValue = require('./helpers/discriminator/getSchemaDiscriminatorByValue');
+const isOperator = require('./helpers/query/isOperator');
 const discriminator = require('./helpers/model/discriminator');
 const each = require('./helpers/each');
 const get = require('./helpers/get');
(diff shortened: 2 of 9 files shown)


lib/query.js
```diff
@@ -2386,6 +2387,9 @@ Query.prototype.mongooseOptions = function(v) {
 };
 
 /**
+ * Casts query filter values and optionally validates them when the
```
vkarpov15: Casting and validation are separate steps in Mongoose, so I don't like having `castConditions()` also run `validateFilter()`. I'd rather have each individual function have a separate check for `validateFilter` option and call `validateFilter()` if necessary.

lib/model.js
```diff
@@ -4350,6 +4351,138 @@ Model.validate = async function validate(obj, pathsOrOptions, context) {
...
+
+    if (key === '$where' || key === '$expr' || key === '$text' || key === '$comment') {
+      continue;
+    }
+
+    const val = obj[key];
+
+    if (utils.isPOJO(val) && !Object.keys(val).some(isOperator)) {
+      for (const nestedKey of Object.keys(val)) {
+        const nestedPath = key + '.' + nestedKey;
```
vkarpov15, line 4439: We should run this check _after_ `const schematype = schema.path(key);` . This logic is meant to capture nested paths, which do not have a schematype, but it would also run on subdocuments, which do have a schematype. See [subdocs vs nested paths](https://mongoosejs.com/docs/subdocs.html#subdocuments-versus-nested-paths).

lib/model.js
```diff
@@ -4350,6 +4351,138 @@ Model.validate = async function validate(obj, pathsOrOptions, context) {
...
+  if (schemaType == null) {
+    return;
+  }
+
+  try {
+    await schemaType.doValidate(value, context, { path });
+  } catch (err) {
+    const validationError = new ValidationError();
+    validationError.addError(path, err);
+    throw validationError;
```
vkarpov15: As written, this will only get the first validator error, not all validator errors. For example query on `{ status: 'invalid', priority: -3.1415 }` would only get the error on `status` but not include the error on `priority`. We should create a validation error and continue to validate all fields in the query, making sure we return all validation errors. Just be careful not to create a ValidationError if there aren't any errors - that is bad for perf.

vkarpov15 (changes requested):
A few comments.

Also, at a high level it would make sense to create a new FilterValidationError class so it is easier to distinguish between validation errors in the filter vs validation errors in the update.

Also, a point to discuss: how should `validateFilter` work with the existing `runValidators` option? The `runValidators` option runs validation on the update portion of the query, but not the filter. Maybe worthwhile to deprecate `runValidators` in favor of a separate `validateUpdate` option, WDYT @AbdelrahmanHafez @hasezoey ?

AbdelrahmanHafez:
>Maybe worthwhile to deprecate runValidators in favor of a separate validateUpdate option

Agreed, that makes things clearer

lib/model.js
```diff
@@ -4350,6 +4351,138 @@ Model.validate = async function validate(obj, pathsOrOptions, context) {
...
+ * @param {object} conditions query filter object
+ * @param {object} [context] context passed to validators
+ * @return {Promise<object>} the query filter if validation succeeded
+ * @api public
+ */
+
+Model.validateFilter = async function validateFilter(conditions, context) {
+  if (typeof context === 'function' || typeof arguments[2] === 'function') {
+    throw new MongooseError('Model.validateFilter() no longer accepts a callback');
+  }
```
hasezoey: This is a new function, so it shouldnt need this guard.

lib/model.js
hasezoey, line 1: The new private functions should also have JSDoc

lib/cursor/queryCursor.js
```diff
@@ -98,7 +98,20 @@ function QueryCursor(query) {
...
+    }
+
+    return query._castConditions().then(() => {
+      if (query.error()) {
+        onPreComplete(query.error());
+        return;
+      }
+      onPreComplete(null);
+    });
+  }, err => onPreComplete(err));
```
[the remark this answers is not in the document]
IslandRhythms: outdated

hasezoey (commented):
LGTM

PR #16327, opened 2026-06-10
docs(home): center version and GitHub buttons on homepage

AbdelrahmanHafez:
The "Version X" text on the home page wasn't actually centered between the Star and Fork buttons, the row was positioned with magic offsets (`margin-left: 109px`, asymmetric margins, a `top: -4px` hack). The ghbtns iframes were also 100px wide while the Star button with the current star count needs 107px, so the count box was getting clipped.

This PR centers the row with flexbox instead, and widens the iframes to fit their rendered content (109px Star, 103px Fork) so the counts aren't clipped. Visual-only change.

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -228,22 +228,18 @@ h2 a {
 }
 #follow {
   margin-bottom: 36px;
-  margin-left: 109px;
 }
 #follow ul {
+  display: flex;
+  align-items: center;
+  justify-content: center;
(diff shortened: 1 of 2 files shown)


vkarpov15 (approved):
Looking good :+1: 

Image

vkarpov15 (approved):

PR #16329, opened 2026-06-10
docs: remove stale docs/llms.txt and gitignore generated llms artifacts

AbdelrahmanHafez:
Since 3c5d395 the llms.txt generator writes to the top level instead of `docs/llms.txt`, but the old `docs/llms.txt` was committed and never removed, so master now carries a stale copy the script will never update again.

Also, running `npm run docs:generate` leaves the new top-level `llms.txt` and ~100 `docs/api/*.md` files (markdown versions of the API pages for llms.txt and the copy-markdown button) as untracked clutter. Their `.html` siblings are already gitignored, the `.md` flow just never got the same treatment.

This PR:

- removes the stale `docs/llms.txt`
- gitignores `docs/api/*.md` and `/llms.txt`
- fixes an outdated JSDoc comment in `generateLLMsTXT.js` that still mentioned the old path

gh-pages keeps these ignore rules commented out so the generated files stay trackable there, #16328 adds the two new patterns to that commented block. When merging master into gh-pages, keep gh-pages's commented versions if `.gitignore` conflicts.


vkarpov15 (approved):
Thanks :+1:

PR #16331, opened 2026-06-11
perf: various improvements to speed up saveSimple

vkarpov15:
**Summary**

1. Avoid repeated deletes when calling StateMachine `clear()`
2. Early return in `applyDefaults()`
3. Skipping $__resetAtomics for primitive-only docs (similar to #16249)
4. Avoid unnecessary IIFE
5. Avoid hooks that are unnecessarily marked as async functions - for example, saveSubdocs exits early if there's no subdocs, we shouldn't allocate an extra promise in that case
6. Only one set in `getKeysInSchemaOrder.js`

Before:

```
{
  "Average mongoose save time ms": 2.41,
  "Average driver insertOne time ms": 1.45
}
{
  "Average mongoose save time ms": 2.09,
  "Average driver insertOne time ms": 1.4
}
{
  "Average mongoose save time ms": 2.05,
  "Average driver insertOne time ms": 1.42

```

After:

```
{
  "Average mongoose save time ms": 1.99,
  "Average driver insertOne time ms": 1.36
}
{
  "Average mongoose save time ms": 1.7,
  "Average driver insertOne time ms": 1.24
}
{
  "Average mongoose save time ms": 1.73,
  "Average driver insertOne time ms": 1.38
}

```

**Examples**

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -562,27 +562,26 @@ function $applyDefaultsToNested(val, path, doc) {
 Document.prototype.$__buildDoc = function(obj, fields, skipId, exclude, hasIncludedChildren) {
   const doc = {};
 
-  const paths = Object.keys(this.$__schema.paths).
-    // Don't build up any paths that are underneath a map, we don't know
-    // what the keys will be
-    filter(p => !p.includes('$*'));
+  const paths = Object.keys(this.$__schema.paths);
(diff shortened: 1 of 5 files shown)


AbdelrahmanHafez (approved):
Nice!

PR #16333, opened 2026-06-12
fix(populate): split into separate populate per document if resulting $in has more than 50k elements to reduce risk of BSON errors

### LINKED ISSUE #5890: Running into 16MB BSON limit with Mongoose Populate
I am doing a really large Mongoose call with nested population, and then doing some processing on the data.

However, because of the nested populate call I am hitting the 16MB limit for a BSON document in Mongoose.

What's a solid way around this? 

    let allDocuments = await exampleModel.find({
        condition: true
      }).populate({path: 'tvShows', populate: {path: 'views'}});

[![enter image description here][1]][1]

Maybe I can break into into multiple calls? But I'm not sure a logical way to do that. Thanks.

  [1]: https://i.stack.imgur.com/zUVO7.png

vkarpov15:
Fix #5890

**Summary**

The crux of the issue in #5890 is that `populate()` can lead to BSON size limit errors in the case of multiple documents. About 800k ObjectIds fit in the BSON size limit - however, `populate()` on multiple documents can try to stuff more ids than that into a query. For example, if we have a relationship `User` -> `TVShow` -> `View`, the `populate()` call for `View` can include the `views` from multiple `TVShow` documents, and lead to BSON overflow.

The practical solution I came up with is to switch to `perDocumentLimit` behavior (one query per document vs per model) if the resulting query has more than 50k elements. That limit is intentionally not very rigorous, because calculating the BSON serialized size for a large array can be performance intensive; but the 50k limit is enough to avoid BSON overflows for ObjectIds, uuids, dates, and numbers.

The key insight behind splitting by document is that documents themselves must fit within the BSON size limit, so an individual document's `populate()` cannot trigger a BSON size limit presuming that document was already saved in MongoDB (of course possible to create a document in memory with an ObjectId array of length 1M and try to populate it but not very realistic because you could never save this document).

**Examples**

### DIFF
--- a/lib/helpers/populate/createPopulateQueryFilter.js
+++ b/lib/helpers/populate/createPopulateQueryFilter.js
@@ -25,7 +25,15 @@ module.exports = function createPopulateQueryFilter(ids, _match, _foreignField, 
     for (let i = 0; i < _parentPaths.length - 1; ++i) {
       const cur = _parentPaths[i];
       if (match[cur] != null && match[cur].$elemMatch != null) {
-        match[cur].$elemMatch[foreignField.slice(cur.length + 1)] = trusted({ $in: ids });
+        // Copy rather than mutate so the user's `match` stays unchanged and split populate
+        // queries (gh-5890) each get their own `$in` rather than sharing one object
+        match[cur] = {
+          ...match[cur],
+          $elemMatch: {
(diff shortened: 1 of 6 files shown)


hasezoey (commented):
LGTM

PR #16336, opened 2026-06-18
feat(model): support named parameters in Model.findOne

swayamsingh2004:
## Summary

Adds support for calling `Model.findOne` with a single named-parameters object, in addition to the existing positional arguments. Addresses #10367, scoped to `Model.findOne` as a first step (per discussion in the issue).

## Approach

Per the discussion in #10367, named parameters are detected using an internal `namedParamsSymbol` (added to `lib/helpers/symbols.js`, following the same pattern as existing internal symbols like `modelCollectionSymbol`). This avoids any ambiguity with real schema field names like `filter`, `projection`, or `options`, since a Symbol key can never collide with a string key.

## Example

```js
// existing positional style (unchanged)
User.findOne({ email: 'x@y.com' }, 'name', { lean: true })

// new named-parameters style
User.findOne({
  [namedParamsSymbol]: true,
  filter: { email: 'x@y.com' },
  projection: 'name',
  options: { lean: true }
})
```

## Testing

- Added a test in `test/model.querying.test.js` covering both the old positional style and the new named-parameters style.
- Ran the full test suite locally (`npm test`): 4378 passing, 0 failures (one unrelated pre-existing flaky timeout in `timestamps.test.js` that passes in isolation).

### DIFF
--- a/lib/helpers/symbols.js
+++ b/lib/helpers/symbols.js
@@ -18,3 +18,4 @@ exports.schemaTypeSymbol = Symbol('mongoose#schemaType');
 exports.sessionNewDocuments = Symbol('mongoose#ClientSession#newDocuments');
 exports.scopeSymbol = Symbol('mongoose#Document#scope');
 exports.validatorErrorSymbol = Symbol('mongoose#validatorError');
+exports.namedParamsSymbol = Symbol('mongoose#Model#namedParams');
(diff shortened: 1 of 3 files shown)


AbdelrahmanHafez:
closing in favor of #16343 as discussed in #10367

PR #16343, opened 2026-06-26
feat(model): proof-of-concept named parameters for findOne

AbdelrahmanHafez:
re #10367

Proof of concept for RORO-style named parameters, scoped to `Model.findOne`:

```js
User.findOne({ $filter: { email: 'x@y.com' }, $projection: 'name', $options: { lean: true } });
```

`$`-prefixed keys can't collide with a positional filter (the server rejects unknown `$`-prefixed top-level keys as operators), so detection is unambiguous.

- `getNamedParams` helper: single object arg, every key a known sentinel, throws on mixed/unknown keys or a stray positional arg. Positional calls are unchanged.
- Branded TS overload: `$filter` required (`$filter: null` for options-only). Branding the positional overloads with `& { $filter?: never; ... }` stops malformed calls from silently falling through to the positional filter, for ~free (~120 instantiations).

### DIFF
--- a/lib/helpers/query/getNamedParams.js
+++ b/lib/helpers/query/getNamedParams.js
@@ -1,0 +1,51 @@
+'use strict';
+
+const MongooseError = require('../../error/mongooseError');
+
+/*!
+ * Detects and validates mongoose "named parameters" (RORO style), e.g.
+ * `Model.findOne({ $filter, $projection, $options })`.
+ *
+ * A call is treated as named parameters when a single object argument is passed
+ * and at least one of its top-level keys is a recognized sentinel for the
+ * method. `$`-prefixed sentinels can never collide with a positional filter,
+ * since the server rejects unknown `$`-prefixed keys as top-level operators.
+ *
+ * @param {String} method the calling method name, used in error messages
+ * @param {String[]} sentinels recognized `$`-prefixed keys, e.g. `['$filter', '$projection', '$options']`
+ * @param {Array|arguments} args the arguments passed to the method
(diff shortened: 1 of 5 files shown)


lib/helpers/query/getNamedParams.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+module.exports = function getNamedParams(method, sentinels, args) {
+  const first = args[0];
+  if (first == null || typeof first !== 'object' || Array.isArray(first)) {
+    return null;
+  }
+
+  const keys = Object.keys(first);
+  const usesNamedParams = keys.some(key => sentinels.indexOf(key) !== -1);
```
vkarpov15, line 28: This means passing an object with `$projection` would be recognized as named params even if `$filter` is not set, which is inconsistent.

lib/helpers/query/getNamedParams.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  if (args.length > 1) {
+    throw new MongooseError(
+      `${method}() named parameters must be passed as a single argument`
+    );
+  }
+
+  const params = {};
+  for (const key of keys) {
```
vkarpov15, line 40: I'd also get rid of this check since Mongoose is typically permissive about extraneous object properties.

vkarpov15 (commented):
Couple of comments. I'd recommend removing `getNamedParams` entirely in favor of inlining a small check.

Also worth mentioning: using named params doesn't prevent using query chaining.

PR #16344, opened 2026-06-27
types: correct this parameter for methods with versionKey

### LINKED ISSUE #16046: TypeScript `this` Context Error When Calling Static Method on Schema Using `timestamps` and `SchemaOptions.virtuals`

vkarpov15:
Fix #16046

**Summary**

Omit `this` from methods because of `versionKey` option handling mismatch. `Schema.create()` already does this with the original fix for #16046 [here](https://github.com/automattic/mongoose/commit/19d3b4293f591cbefe1062ac55699c4814f1c754).

Mongoose adds `this` parameter to methods with `AddThisParameter<>` so the method body gets type checked correctly in `new Schema()` (`this` is correctly set to the inferred document), but then Mongoose strips `this` when the methods are added to the actual hydrated document type, because at that point TypeScript already knows the correct type of `this` (the actual document).

**Examples**

### DIFF
--- a/types/utility.d.ts
+++ b/types/utility.d.ts
@@ -163,6 +163,9 @@ declare module 'mongoose' {
 
   type OmitThisParameterIfFunction<T> = T extends (...args: any[]) => any ? OmitThisParameter<T> : T;
 
+  // Strip explicit `this` parameter from methods in schema type options. The `this` parameter
+  // is just for type-checking the function body. Keeping the explicit 'this' typing can cause
+  // compiler errors on the method calls, so remove it for the hydrated document definition.
   type HydratedDocumentOverrides<T> = {
     [K in keyof T]: OmitThisParameterIfFunction<T[K]>;
   };
--- a/types/utility.d.ts
+++ b/types/utility.d.ts
@@ -161,6 +161,12 @@ declare module 'mongoose' {
       : T[K];
   };
 
+  type OmitThisParameterIfFunction<T> = T extends (...args: any[]) => any ? OmitThisParameter<T> : T;
+
+  type HydratedDocumentOverrides<T> = {
+    [K in keyof T]: OmitThisParameterIfFunction<T[K]>;
+  };
+
   /**
    * @summary Adds timestamp fields to a type
    * @description Adds createdAt and updatedAt fields of type Date, or custom timestamp fields if specified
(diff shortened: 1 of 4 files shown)


AbdelrahmanHafez (approved):

types/utility.d.ts
```diff
@@ -161,6 +161,12 @@ declare module 'mongoose' {
       : T[K];
   };
 
+  type OmitThisParameterIfFunction<T> = T extends (...args: any[]) => any ? OmitThisParameter<T> : T;
+
+  type HydratedDocumentOverrides<T> = {
```
hasezoey, line 169: This should likely have some documentation

hasezoey (approved):

PR #16345, opened 2026-06-28
feat(schema): add `strictRead` option to filter unknown fields on document hydration

### LINKED ISSUE #4279: Schema is not applied in reading
I have a collection on MongoDb with several fields and not all of them are defined in the model schema. For example on the db I have a document with fields `a, b, c` and `d`, and in  my mongoose schema I have only `a` and `b` defined. 
When I query the collection, the returned documents contains also these additional fields. 
Why? IMHO it shouldn't be like that, the returned document should contains only the declared fields.
Opinions?

thx

gourabsingha1:
## What does this PR do?

Closes #4279

Adds a new `strictRead` schema option that controls what happens when Mongoose hydrates a document from MongoDB that contains fields not defined in the schema.

### Why?

The existing `strict` option governs writes (construction of new documents), and `strictQuery` governs query filters. But there was no built-in mechanism to filter out extra fields when **reading** documents back from the database — e.g. when another application has written extra fields to the collection.

The maintainer (@vkarpov15) approved the `strictRead` naming in the issue thread.

---

### New option: `strictRead`

| Value | Behavior |
|---|---|
| `false` *(default)* | Existing behavior — unknown fields stored in `_doc` |
| `true`  | Unknown fields are silently stripped during `init` |
| `'throw'` | A `StrictModeError` is thrown when an unknown field is encountered |

### Example

```js
const schema = new Schema(
  { name: String, email: String },
  { strictRead: 'filter' } // or true, or 'throw'
);

// If the DB document has { name: 'Alice', email: 'a@b.com', legacyField: 'old' }
const doc = await User.findOne({ name: 'Alice' });
doc.name;          // 'Alice'
doc.email;         // 'a@b.com'
doc.legacyField;   // undefined — stripped!
doc._doc.legacyField; // undefined — not in _doc either
```

### Implementation details

- Modified `init()` in `lib/document.js` to read `schema.options.strictRead` and skip (or throw) for unknown fields before storing to `_doc`
- Added `strictRead: false` as the default in `Schema.prototype.defaultOptions()` in `lib/schema.js`
- Added `strictRead` to the list of valid global options in `lib/validOptions.js`
- **Does not** affect new document construction (that is governed by `strict`)
- **Does not** affect the `strict` or `strictQuery` semantics in any way

### Tests

8 new tests in `test/document.strictRead.test.js` covering:
- `strictRead: false` (default — existing behavior preserved)
- `strictRead: true` — strips unknown fields, preserves schema fields, doesn't affect new doc creation
- `strictRead: 'throw'` — throws on unknown fields, passes when all fields are known
- Nested object handling

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -72,6 +72,7 @@ const numberRE = /^\d+$/;
  * - [shardKey](https://mongoosejs.com/docs/guide.html#shardKey): object - defaults to `null`
  * - [strict](https://mongoosejs.com/docs/guide.html#strict): bool - defaults to true
  * - [strictQuery](https://mongoosejs.com/docs/guide.html#strictQuery): bool - defaults to false
+ * - [strictRead](https://mongoosejs.com/docs/guide.html#strictRead): bool or 'throw' - defaults to false. If set to `true` or `'filter'`, fields not in the schema will be stripped when hydrating documents from MongoDB. If set to `'throw'`, an error will be thrown when a field not in the schema is encountered during document hydration.
  * - [toJSON](https://mongoosejs.com/docs/guide.html#toJSON) - object - no default
  * - [toObject](https://mongoosejs.com/docs/guide.html#toObject) - object - no default
  * - [typeKey](https://mongoosejs.com/docs/guide.html#typeKey) - string - defaults to 'type'
@@ -610,6 +611,7 @@ Schema.prototype.defaultOptions = function(options) {
   options = {
     strict: defaultStrict,
     strictQuery: defaultStrictQuery,
+    strictRead: false,
     bufferCommands: true,
     capped: false, // { size, max, autoIndexId }
     versionKey: '__v',
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -607,11 +607,12 @@ Schema.prototype.defaultOptions = function(options) {
   const baseOptions = this.base?.options || {};
   const defaultStrict = baseOptions.strict ?? true;
   const defaultStrictQuery = baseOptions.strictQuery ?? false;
+  const defaultStrictRead = baseOptions.strictRead ?? false;
   const defaultId = baseOptions.id ?? true;
   options = {
     strict: defaultStrict,
     strictQuery: defaultStrictQuery,
-    strictRead: false,
+    strictRead: defaultStrictRead,
     bufferCommands: true,
     capped: false, // { size, max, autoIndexId }
     versionKey: '__v',
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -72,7 +72,7 @@ const numberRE = /^\d+$/;
  * - [shardKey](https://mongoosejs.com/docs/guide.html#shardKey): object - defaults to `null`
  * - [strict](https://mongoosejs.com/docs/guide.html#strict): bool - defaults to true
  * - [strictQuery](https://mongoosejs.com/docs/guide.html#strictQuery): bool - defaults to false
- * - [strictRead](https://mongoosejs.com/docs/guide.html#strictRead): bool or 'throw' - defaults to false. If set to `true` or `'filter'`, fields not in the schema will be stripped when hydrating documents from MongoDB. If set to `'throw'`, an error will be thrown when a field not in the schema is encountered during document hydration.
+ * - [strictRead](https://mongoosejs.com/docs/guide.html#strictRead): bool or 'throw' - defaults to false. If set to `true`, fields not in the schema will be stripped when hydrating documents from MongoDB. If set to `'throw'`, an error will be thrown when a field not in the schema is encountered during document hydration.
  * - [toJSON](https://mongoosejs.com/docs/guide.html#toJSON) - object - no default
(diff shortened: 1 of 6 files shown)


test/document.strictRead.test.js
```diff
@@ -0,0 +1,149 @@
...
+      const Model = db.model('StrictReadThrowOk', schema);
+      await Model.create({ name: 'valid', score: 99 });
+
+      const doc = await Model.findOne({ name: 'valid' });
+      assert.equal(doc.name, 'valid');
+      assert.equal(doc.score, 99);
+    });
+  });
+
+  describe('strictRead with nested objects', function() {
```
vkarpov15, line 120: Please also add tests covering strictRead with subdocuments and document arrays.

lib/schema.js
```diff
@@ -72,6 +72,7 @@ const numberRE = /^\d+$/;
  * - [shardKey](https://mongoosejs.com/docs/guide.html#shardKey): object - defaults to `null`
  * - [strict](https://mongoosejs.com/docs/guide.html#strict): bool - defaults to true
  * - [strictQuery](https://mongoosejs.com/docs/guide.html#strictQuery): bool - defaults to false
+ * - [strictRead](https://mongoosejs.com/docs/guide.html#strictRead): bool or 'throw' - defaults to false. If set to `true` or `'filter'`, fields not in the schema will be stripped when hydrating documents from MongoDB. If set to `'throw'`, an error will be thrown when a field not in the schema is encountered during document hydration.
```
vkarpov15: Is there a reason for the `'filter'` option? For one thing, these docs are inconsistent because they say "bool or 'throw'" but `'filter'` is also supported. Given that `'filter'` is the same as `true` I would err on the side of getting rid of it, for consistency with the other `strict*` options.

vkarpov15 (changes requested):
A couple of minor comments, also please fix lint. But general direction looks good, thank you!

lib/schema.js
gourabsingha1: yeah right. removed it!

test/document.strictRead.test.js
gourabsingha1: done. added in test/document.strictRead.test.js

test/document.strictRead.test.js
```diff
@@ -0,0 +1,166 @@
...
+
+  describe('strictRead with subdocuments and document arrays', function() {
+    it('strips unknown fields inside subdocuments with strictRead: true', async function() {
+      const childWriteSchema = new Schema({ heading: String }, { strict: false, _id: false });
+      const writeSchema = new Schema({ child: childWriteSchema }, { strict: false });
+      const WriteModel = db.model('StrictReadSubdocWrite', writeSchema);
+      await WriteModel.create({ child: { heading: 'intro', extra: 'nestedExtra' } });
+
+      const childReadSchema = new Schema({ heading: String }, { strictRead: true, _id: false });
+      const readSchema = new Schema({ child: childReadSchema }, { strictRead: true });
```
vkarpov15, line 156: Note to self to confirm behavior on subdocument with strictRead: false

vkarpov15 (approved):
Thank you :+1: merging into 9.8 branch so it can be released with our next minor release

test/document.strictRead.test.js
```diff
@@ -0,0 +1,179 @@
...
+
+      const readSchema = new Schema({ meta: { known: String } }, { strictRead: true });
+      const ReadModel = db.model('StrictReadNestedRead', readSchema, WriteModel.collection.name);
+      const doc = await ReadModel.findOne({});
+
+      assert.equal(doc.meta.known, 'yes');
+      assert.equal(doc._doc.topLevel, undefined, 'top-level extra field should be stripped');
+    });
+
+    it('strips unknown nested objects entirely with strictRead: true', async function() {
```
vkarpov15, line 134: @gourabsingha1 copilot found a legitimate issue here

lib/schema.js
```diff
@@ -610,6 +611,7 @@ Schema.prototype.defaultOptions = function(options) {
   options = {
     strict: defaultStrict,
     strictQuery: defaultStrictQuery,
+    strictRead: false,
```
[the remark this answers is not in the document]
vkarpov15: This is also legitimate given that `strictRead` was added to `validOptions`.

StrictRead should also be added to TypeScript `MongooseOptions`

vkarpov15 (changes requested):
Copilot found a couple of issues that I missed. Also we should add `strictRead` to the `QueryOptions` type in TypeScript

lib/schema.js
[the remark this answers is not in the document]
gourabsingha1: updated!

test/document.strictRead.test.js
gourabsingha1: added

vkarpov15 (approved):

vkarpov15:
Merging into 9.8 for further testing

test/document.strictRead.test.js
vkarpov15: Confirmed that subdoc strictRead false overrides parent strictRead true and vice versa. This is correct :+1:

PR #16347, opened 2026-06-28
feat(document): add `$revert()` method to undo unsaved changes

### LINKED ISSUE #8714: Feature to undo/revert changes on a document (which haven't been persisted yet)
Feature request

There is no easy way to revert changes to a document.

**What is the desired behavior?**
Something like:

```js
const foo = mongoose
  .model('Foo')
  .find({_id});

console.log(foo); //{x: 2, y: 4}

foo.x = 6;
foo.y = 8;

console.log(foo); //{x: 6, y: 8}

foo.revert();

console.log(foo); //{x: 2, y: 4}
```

**Use case**
We have a use case where we update a document in preparation for saving, but then when payment fails, we don't proceed with the save. However, that document is then passed to an email to notify the user of the failure, but it will contain the modified data. 

It'd be nice if there'd be an easy way to revert changes made on a document and revert back to the original values that were loaded from the database, thus preventing a re-fetch from the database or having to muck around with cloning objects.

gourabsingha1:
## What does this PR do?

Closes #8714

Adds a new public `$revert(paths?)` method to documents. This method allows discarding any unsaved modifications and restoring the document to its original state (either when it was last loaded from the database or after a successful save).

### Why?

Currently, there is no straightforward way to discard unsaved modifications to a document. Users who wanted to undo edits had to manually query the DB again, clone the object, or manually track and reset every property. This is especially useful in failure/retry/transaction rollback logic.

---

### API Signature: `doc.$revert(paths?)`

| Parameter | Type | Description |
| --- | --- | --- |
| `paths` *(optional)* | `string \| string[]` | If specified, only the specified path(s) (and their subpaths) are reverted. If omitted, the entire document is reverted. |

### Example

```js
const doc = await UserModel.findOne({ name: 'Alice', age: 30 });

// Edit some paths
doc.name = 'Bob';
doc.age = 45;
doc.$isModified('name'); // true

// Revert specific paths:
doc.$revert('name');
doc.name; // 'Alice' (reverted)
doc.age;  // 45 (not reverted)

// Revert all paths:
doc.name = 'Bob';
doc.$revert();
doc.name; // 'Alice' (reverted)
doc.age;  // 30 (reverted)
doc.$isModified(); // false (modified paths cleared)
```

### Implementation Details

1. **DB Hydration Snapshot**: Inside `$__init()` (the hydration path), we store a deep clone of the initial `_doc` onto `this.$__._originalDoc`.
2. **Post-Save Refresh**: Inside `$__reset()` (which runs after a successful save), we refresh the snapshot of `_originalDoc` using `clone(this._doc)`.
3. **`$revert()` Execution**:
   - If `paths` is omitted: We restore `this._doc` from the `_originalDoc` snapshot and clear all modified paths with `this.$clearModifiedPaths()`.
   - If `paths` is specified: We restore only those specific paths from `_originalDoc` onto `this._doc` and unmark those paths as modified.
4. **TypeScript Declarations**: Added type signature for `$revert()` in `types/document.d.ts`.

### Tests

Added a new test file `test/document.revert.test.js` containing 10 passing tests verifying:
- Full document revert
- Specific path revert (accepting both a single string and array of strings)
- Restores original value from DB and unmarks modification tracking
- Refreshing snapshot on save so subsequent reverts undo to the last-saved state
- Non-interference with other documents
- Graceful no-op handling for newly constructed (unsaved) documents
- Reverting nested object paths

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -725,10 +725,7 @@ Document.prototype.$__init = function(doc, opts) {
 
   applyDefaults(this, this.$__.selected, this.$__.exclude, hasIncludedChildren, false, this.$__.skipDefaults);
 
-  // Initialize savedState so $__saveInitialState() can lazily capture the
-  // pre-modification value of each top-level path the user later changes.
-  // This is O(1) — no clone — and enables $revert() at zero upfront cost.
-  this.$__.savedState = {};
+  this.$__._originalDoc = clone(doc);
 
   return this;
 };
@@ -5659,6 +5656,9 @@ Document.prototype.$clearModifiedPaths = function $clearModifiedPaths() {
     for (const child of subdocs) {
       child.$clearModifiedPaths();
     }
+    if (this.$__._originalDoc != null) {
+      this.$__._originalDoc = clone(this._doc);
+    }
   }
 
   return this;
@@ -5694,53 +5694,6 @@ Document.prototype._applyVersionIncrement = function _applyVersionIncrement() {
   }
 };
 
-/*!
- * Increment this document's version if necessary.
- */
-
-Document.prototype._applyVersionIncrement = function _applyVersionIncrement() {
-  if (!this.$__.version) return;
-  const doIncrement = VERSION_INC === (VERSION_INC & this.$__.version);
-
-  this.$__.version = undefined;
-  if (doIncrement) {
-    const key = this.$__schema.options.versionKey;
-    const version = this.$__getValue(key) || 0;
-    this.$__setValue(key, version + 1); // increment version if was successful
-  }
-};
-/**
- * Reverts unsaved changes to this document, restoring it to the state it was
- * in when it was last loaded from the database or last saved successfully.
- *
- * If `paths` is specified, only those paths (and their children) are reverted;
- * otherwise the entire document is reverted.
- *
- * #### Example:
- *
- *     const doc = await MyModel.findOne();
- *     doc.name = 'changed';
- *     doc.$isModified('name'); // true
- *
- *     doc.$revert();
- *     doc.name; // original value from DB
- *     doc.$isModified('name'); // false
- *
- *     // Revert only specific paths:
- *     doc.name = 'changed';
- *     doc.age = 99;
- *     doc.$revert(['name']);
- *     doc.name; // reverted to original
- *     doc.age;  // 99 — not reverted
- *
- * @param {string|string[]} [paths] specific path(s) to revert. If omitted, reverts all paths.
- * @return {Document} this
- * @api public
- * @method $revert
- * @memberOf Document
- * @instance
- */
-
 Document.prototype.$revert = function $revert(paths) {
   if (this.$isSubdocument) {
     const owner = this.ownerDocument();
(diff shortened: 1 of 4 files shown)


lib/document.js
```diff
@@ -5677,6 +5686,75 @@ Document.prototype._applyVersionIncrement = function _applyVersionIncrement() {
...
+    // Nothing to revert to.
+    return this;
+  }
+
+  if (paths == null) {
+    // Full revert: replace _doc entirely with the original snapshot
+    // Use init() to recast/rehydrate values properly from the snapshot
+    const keys = Object.keys(originalDoc);
+    for (const key of keys) {
+      utils.setValue(key, clone(originalDoc[key]), this._doc);
```
vkarpov15, line 5723: Why `setValue()` rather than simply `this._doc = clone(originalDoc)`? `setValue()` doesn't cast values anyway, so this set is redundant.

lib/document.js
```diff
@@ -724,6 +724,10 @@ Document.prototype.$__init = function(doc, opts) {
     null;
 
   applyDefaults(this, this.$__.selected, this.$__.exclude, hasIncludedChildren, false, this.$__.skipDefaults);
+
+  // Take a snapshot of the document as loaded from the DB so $revert() can restore it
+  this.$__._originalDoc = clone(this._doc);
```
vkarpov15: I'd strongly prefer this to be a method where you can take a snapshot of the current document and later restore the snapshot. `clone()` can be expensive on large documents, and `init()` is one of the most performance sensitive parts of Mongoose.

lib/document.js
gourabsingha1: you're right. it was unnecessary. I've replaced the entire $revert() body to use this.$set(key, clone(savedState[key])) instead, which additionally runs Mongoose's schema casting/rehydration correctly

lib/document.js
gourabsingha1: gott it! I've removed clone(this._doc) entirely from both `$__init() `and `$__reset()`. now it piggybacks on Mongoose's existing `$__savedState` lazy-capture mechanism. `$__init()` now just does a single O(1) `this.$__.savedState = {}` assignment.

the first time a user modifies any path, `$__saveInitialState()` clones only that top-level path, not the whole document. `$revert()` restores those lazily-captured values.

lib/document.js
```diff
@@ -724,6 +724,12 @@ Document.prototype.$__init = function(doc, opts) {
     null;
 
   applyDefaults(this, this.$__.selected, this.$__.exclude, hasIncludedChildren, false, this.$__.skipDefaults);
+
+  // Initialize savedState so $__saveInitialState() can lazily capture the
+  // pre-modification value of each top-level path the user later changes.
+  // This is O(1) — no clone — and enables $revert() at zero upfront cost.
+  this.$__.savedState = {};
```
vkarpov15: Cost isn't zero here, creating a new property on `$__` and creating a new object is fast but still comes with a cost. The other big tradeoff is that this triggers `savedState` storage even on newly created documents, which seems incorrect and unnecessary for this PR's stated goal of undoing unsaved changes - $revert() on a `new MyModel()` would leave current changes. Might be better to move this to init?

vkarpov15 (changes requested):
I have a few comments here and I did some digging into `savedState` - on my original review I didn't even remember we had `savedState` as a property. 

TBH, I don't want to turn on `savedState` for all documents, I worry that will be too much performance overhead for not enough benefit. If you want to store the original state of the document and later revert it, you could do the following:

```
const docSnapshot = doc.toObject({ getters: false });

// Later
doc.overwrite(docSnapshot);
doc.$clearModifiedPaths();
// `doc` now has the exact same properties as docSnapshot with nothing modified
```

WDYT about this @gourabsingha1 ?

lib/document.js
gourabsingha1: yeah right. moved it

gourabsingha1:
> I have a few comments here and I did some digging into `savedState` - on my original review I didn't even remember we had `savedState` as a property.
> 
> TBH, I don't want to turn on `savedState` for all documents, I worry that will be too much performance overhead for not enough benefit. If you want to store the original state of the document and later revert it, you could do the following:
> 
> ```
> const docSnapshot = doc.toObject({ getters: false });
> 
> // Later
> doc.overwrite(docSnapshot);
> doc.$clearModifiedPaths();
> // `doc` now has the exact same properties as docSnapshot with nothing modified
> ```
> 
> WDYT about this @gourabsingha1 ?

Hey Valeri! Thanks for the suggestion.

While using toObject({ getters: false }) + doc.overwrite() works as a workaround, a native $revert() has a few major advantages:

1. Performance (Eager vs. Lazy): toObject() is eager and copies the entire document tree, which is expensive for larger documents. The lazy savedState approach only captures the specific paths that are modified, which makes the upfront cost practically zero (especially now that it's moved out of the constructor).
2. Avoiding Side-Effects: overwrite() replaces all paths, which forces Mongoose to re-run setters, validation, and default generators on untouched fields. A targeted $revert() only restores the modified paths, avoiding any unintended side-effects on unmodified fields.
3. DX: $revert() is a standard API pattern in state tracking and much cleaner than having developers write three lines of boilerplate for a common query undo task.

vkarpov15:
This is something we should implement in a plugin so developers can opt-in if they need to. I don't want to enable tracking `savedState` for every single document - that's a non-trivial amount of performance overhead in a heavily used path for a new feature that I would rate as a 4 or 5 out of 10 in terms of importance. I'll keep #8714 open to implement this as a plugin.

PR #16348, opened 2026-06-29
Support `middleware` skip option for cursors and custom hooks

AbdelrahmanHafez:
re #8768, builds on top of https://github.com/mongoosejs/kareem/pull/46

We left some parts uncovered in #15883, namely statics, methods, and cursors (aggregation, and query).

This PR:

- Applies the existing `middleware` skip option to query cursors, including both pre `find` hooks and per-document post `find` hooks.
- Applies the skip option to aggregate cursors for pre `aggregate` hooks. Aggregate cursors still don't run post `aggregate` hooks, so the cursor-specific TS type only exposes `middleware: false` and `{ pre: false }`.
- Strips the Mongoose-only `middleware` option before passing cursor options to the MongoDB driver.
- Adds support for skipping middleware on custom statics and methods by reading `middleware` from the final plain options object.
- Documents the cursor behavior and the recommended custom static/method signature.

For now this pins `kareem` to the fork commit that adds wrapper `getOptions` support for custom statics/methods. Once the Kareem change lands and is released, we can switch this back to a normal version range.

A design decision that was made is how we pass the `middleware` option to statics/methods, I chose to pass them to the last arg if it's POJO, and the docs now have examples with a pattern where users define the last arg as options, following the same pattern as the other mongoose operations.

### DIFF
--- a/lib/helpers/buildMiddlewareFilter.js
+++ b/lib/helpers/buildMiddlewareFilter.js
@@ -15,8 +15,8 @@ const isBuiltInMiddleware = hook => hook.fn[symbols.builtInMiddleware];
  * @returns {Function|null} - null runs all middleware, isBuiltInMiddleware skips user middleware
  */
 function buildMiddlewareFilter(options, phase) {
-  const shouldRun = options?.middleware?.[phase] ?? options?.middleware ?? true;
-  return shouldRun ? null : isBuiltInMiddleware;
+  const shouldSkip = options?.middleware === false || options?.middleware?.[phase] === false;
+  return shouldSkip ? isBuiltInMiddleware : null;
 }
 
 module.exports = {
--- a/lib/query.js
+++ b/lib/query.js
@@ -5330,7 +5330,8 @@ Query.prototype._applyPaths = function applyPaths() {
  * Returns a wrapper around a [mongodb driver cursor](https://mongodb.github.io/node-mongodb-native/7.0/classes/FindCursor.html).
  * A QueryCursor exposes a Streams3 interface, as well as a `.next()` function.
  *
- * The `.cursor()` function triggers pre find hooks, but **not** post find hooks.
+ * The `.cursor()` function triggers pre find hooks before opening the cursor
+ * and post find hooks for each document.
  *
  * #### Example:
  *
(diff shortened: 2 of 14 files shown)


lib/helpers/buildMiddlewareFilter.js
```diff
@@ -15,8 +15,8 @@ const isBuiltInMiddleware = hook => hook.fn[symbols.builtInMiddleware];
  * @returns {Function|null} - null runs all middleware, isBuiltInMiddleware skips user middleware
  */
 function buildMiddlewareFilter(options, phase) {
-  const shouldRun = options?.middleware?.[phase] ?? options?.middleware ?? true;
-  return shouldRun ? null : isBuiltInMiddleware;
+  const shouldSkip = options?.middleware === false || options?.middleware?.[phase] === false;
+  return shouldSkip ? isBuiltInMiddleware : null;
```
AbdelrahmanHafez, line 19: Slightly changes things from checking against any false values to strict `false`.

lib/query.js
```diff
@@ -5330,7 +5330,8 @@ Query.prototype._applyPaths = function applyPaths() {
  * Returns a wrapper around a [mongodb driver cursor](https://mongodb.github.io/node-mongodb-native/7.0/classes/FindCursor.html).
  * A QueryCursor exposes a Streams3 interface, as well as a `.next()` function.
  *
- * The `.cursor()` function triggers pre find hooks, but **not** post find hooks.
+ * The `.cursor()` function triggers pre find hooks before opening the cursor
+ * and post find hooks for each document.
```
AbdelrahmanHafez, line 5334: This comment was outdated, we do now run `post` find hooks

types/index.d.ts
```diff
@@ -1152,7 +1152,7 @@ declare module 'mongoose' {
   /* for ts-mongoose */
   export class mquery { }
 
-  export function overwriteMiddlewareResult(val: any): Kareem.OverwriteMiddlewareResult;
+  export function overwriteMiddlewareResult(val: any): Kareem.OverwriteResult;
```
AbdelrahmanHafez, line 1155: Uses the updated correct type, old one referred a function that didn't exist

PR #16350, opened 2026-06-29
Support `middleware` skip option for cursors and custom hooks

AbdelrahmanHafez:
re #8768, builds on top of https://github.com/mongoosejs/kareem/pull/46

We left some parts uncovered in #15883, namely statics, methods, and cursors (aggregation, and query).

This PR:

- Applies the existing `middleware` skip option to query cursors, including both pre `find` hooks and per-document post `find` hooks.
- Applies the skip option to aggregate cursors for pre `aggregate` hooks. Aggregate cursors still don't run post `aggregate` hooks, so the cursor-specific TS type only exposes `middleware: false` and `{ pre: false }`.
- Strips the Mongoose-only `middleware` option before passing cursor options to the MongoDB driver.
- Adds support for skipping middleware on custom statics and methods by reading `middleware` from the final plain options object.
- Documents the cursor behavior and the recommended custom static/method signature.

For now this pins `kareem` to the fork commit that adds wrapper `getOptions` support for custom statics/methods. Once the Kareem change lands and is released, we can switch this back to a normal version range.

A design decision that was made is how we pass the `middleware` option to statics/methods, I chose to pass them to the last arg if it's POJO, and the docs now have examples with a pattern where users define the last arg as options, following the same pattern as the other mongoose operations.


test/model.skip.middleware.test.js
```diff
@@ -648,6 +655,415 @@ describe('middleware option to skip hooks (gh-8768)', function() {
...
+        assert.strictEqual(doc.name, 'John');
+      });
+
+      // Assert
+      assert.strictEqual(numDocs, 1);
+      assert.strictEqual(getPreCount('find'), 1);
+      assert.strictEqual(getPostCount('find'), 0);
+    });
+
+    it('skips pre/post find hooks when iterating with next()', async function() {
```
hasezoey:
```suggestion
    it('skips pre/post find hooks when iterating with next() when middleware: false', async function() {
```

Consistency

hasezoey (commented):
LGTM.
I dont really like how the options for custom methods is implicit, though aside from forcing users to *somehow* handle it (by transferring pre/post running to be manual calls), i dont see another way.

lib/cursor/aggregationCursor.js
```diff
@@ -49,7 +50,9 @@ function AggregationCursor(agg) {
   this._mongooseOptions = {};
 
   if (connection) {
-    this.cursor = connection.db.aggregate(agg._pipeline, agg.options || {});
+    const options = agg.options != null ? { ...agg.options } : {};
+    delete options.middleware;
```
vkarpov15: I recommend only deleting `options.middleware` if it is present, `delete` is relatively slow.

docs/middleware.md
```diff
@@ -652,6 +661,23 @@ await doc.save({ middleware: { pre: false } });
 await Model.find({}, null, { middleware: { post: false } });
 ```
 
+For custom statics and methods, reserve the last argument for Mongoose options and default it to `{}`.
+Mongoose reads `middleware` from that trailing options object:
+
+```javascript
+schema.statics.queueEmail = async function(to, emailOptions, options = {}) {
```
vkarpov15, line 670: I have some concerns that `function queueEmail(params)` will skip middleware if `params.middleware` is true, but `middleware` may be a parameter set for some other lib. Could potentially be backwards breaking if someone relies on middleware for statics while also passing `middleware` as a  parameter.

Potential alternatives:

1. Don't allow skipping middleware on methods and statics
2. Require a more explicit opt-in by only checking the last parameter if it is named `mongooseOptions`
3. Add new `mongoose.MiddlewareOptions` class and explicitly check for instances of `mongoose.MiddlewareOptions`

hasezoey: Another alternative, though i dont know how common it is, would be to assign some extra property to those functions, similar to:

```ts
schema.statics.someFunction = function(...) {}
schema.statics.someFunction.useMiddleware = true;
```

This could serve to enable / disable middleware options and could be enabled by default (or at least in a later version of mongoose)

AbdelrahmanHafez: @hasezoey I like that, with one tweak: if the flag toggled middleware itself it wouldn't solve the per-operation case, but as a gate for whether Mongoose reads the trailing options object it works well. I'd rename it to something like `supportsMiddlewareOption` though, `useMiddleware` sounds like "this static uses middleware" rather than "this static accepts the `middleware` option".

I think that gives us the best migration path: ship the flag as an explicit opt-in now, which is strictly non-breaking, then in v10 remove the flag entirely and always read the option. The flag becomes a harmless property assignment for anyone who set it, adopters' call sites never change, and they get the same `{ middleware: false }` shape as the built-in operations from day one. The cost is that adopters do a two-step setup until v10, which seems like a fair price for the transition.

The v10 break is also narrower than it looks. A user only breaks if all three are true:

1. They use statics.
2. They have middleware registered for that specific static (`schema.pre('queueEmail')`).
3. They pass a trailing options object with `middleware: false` in it for unrelated reasons. Something like `middleware: ['e-mail', 'sms']` is truthy, so hooks run exactly as before.

The check is currently falsy-based, I'll change it in this PR to only honor exactly `false` or the `{ pre, post }` object shape, which narrows that even further.

So my order of preference: `supportsMiddlewareOption` opt-in now and always-on in v10, then plain `middleware` directly accepting the narrow risk, then `mongooseOptions.middleware`. `mongoose.MiddlewareOptions` feels too verbose, and not supporting statics/methods at all is inconsistent with the rest of the feature.

WDYT?

vkarpov15: supportsMiddlewareOption works for me. Let's add that and avoid unnecessarily running `delete options.middleware` if middleware option not set, and this should be good to go

package.json
```diff
@@ -21,7 +21,7 @@
   "license": "MIT",
   "dependencies": {
     "@standard-schema/spec": "^1.1.0",
-    "kareem": "3.3.0",
+    "kareem": "git+https://github.com/AbdelrahmanHafez/kareem.git#f0ec8dcc94f616e7468a839fb122ebb6a14d5c6a",
     "mongodb": "~7.5",
```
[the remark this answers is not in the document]
AbdelrahmanHafez: The git pin is temporary, it points at the kareem change this PR depends on (mongoosejs/kareem#46). Once that lands and a new kareem version is released, this goes back to a normal version range.

docs/middleware.md
AbdelrahmanHafez: Addressed 👍

AbdelrahmanHafez:
Not sure if this is worth a minor version or can be shipped in a patch. I'm okay with either @vkarpov15

hasezoey (commented):
LGTM

docs/middleware.md
AbdelrahmanHafez: I've been thinking about this a bit, I think we should keep requiring setting `.supportsMiddlewareOption = true` in order for this to work even in v10. The discoverability isn't the best, but I think it mitigates the risk of implicitly skipping middlewares while still having the feature available behind a flag.

lib/aggregate.js
```diff
@@ -943,6 +947,11 @@ Aggregate.prototype.option = function(value) {
 
 Aggregate.prototype.cursor = function(options) {
   this._optionsForExec();
+  if (utils.hasUserDefinedProperty(options, 'middleware')) {
+    options = { ...options };
+    this.options.middleware = options.middleware;
+    delete options.middleware;
```
vkarpov15, line 953: Since `delete` is slow I would prefer to do `({ middleware, ...options } = options)` but this is fine for now.

vkarpov15 (approved):
LGTM :+1:

Issue #16362, opened 2026-07-03
`string` type assignment fails for `ObjectId` in nested subdocuments


E1ecti:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

9.7.3

### Node.js version

26.4.0

### MongoDB server version

Unrelated

### Typescript version (if applicable)

6.0.3

### Description

Related to #16316.

When using nested interfaces (for example, when using `class-validator` where subdocuments are defined by separate classes), the error `Index signature for type 'string' is missing in type 'N'` occurs, or `Type 'string' is not assignable to type 'ObjectId ...'` in the case of nested arrays.

This issue only affects cases where `string` is used instead of `mongoose.Types.ObjectId`.

It can also be solved by adding `[x: string]: {}` to the validation class, however, this is not a suitable solution.

### Steps to Reproduce

```ts
import * as mongoose from "mongoose"

interface User {
	dummy: string
}

interface Other {
	nested: {
		to: mongoose.PopulatedDoc<User>
	}
}

interface Nested {
	to: string
}
interface OtherNested {
	nested: Nested
}

const userSchema = new mongoose.Schema({ dummy: String })
const userModel = mongoose.model<User>("user", userSchema)

const otherSchema = new mongoose.Schema({
	nested: {
		to: { type: mongoose.Types.ObjectId, ref: "user" }
	}
})
const otherModel = mongoose.model<Other>("other", otherSchema)

async function main() {
	const user = await userModel.create({ dummy: "hello" })

	const userId = user._id.toString()
	const data: OtherNested = { nested: { to: userId } }

	// Index signature for type 'string' is missing in type 'Nested'.
	const other = await otherModel.create(data)

	console.log("Done", other)

	await mongoose.disconnect()
}

main()
```

### Expected Behavior

Using `string` should be acceptable as using `ObjectId` inside subdocuments during creation.

vkarpov15:
I put in a PR to fix, however you can also work around this by making `Nested` a type instead of an interface:

```ts
type Nested = {
        to: string
} 

interface OtherNested {
        nested: Nested
}
```

hasezoey:
FYI this is a issue (from the same user) that was also raised in typegoose's discord with reproduction in [`typegoose-testing`](https://github.com/typegoose/typegoose-testing/pull/2).

PR #16363, opened 2026-07-05
types(create): fix handling of nested objects typed as interfaces

vkarpov15:
**Summary**

#16362 is not an issue with strings vs objectids, it's an issue with how nested object types vs nested interfaces are handled by `create()` typing. In #16362, `Nested` is an interface not an object type, and interfaces are not assignable to `Record<string, unknown>` for TypeScript internal reasons. Probably because interfaces can be augmented (multiple definitions of the same interface add properties).

**Examples**

### DIFF
--- a/test/types/create.test.ts
+++ b/test/types/create.test.ts
@@ -222,9 +222,10 @@ async function createWithPopulatedDoc() {
     };
   }
   interface OtherNestedCreateInput {
-    nested: {
-      to: string;
-    };
+    nested: Nested;
+  }
+  interface Nested {
+    to: string;
   }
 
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):

AbdelrahmanHafez:
Probably worth addressing Copilot's comment, otherwise LGTM

hasezoey (approved):
Looks good to me and fixes the issue in [`typegoose-testing`](https://github.com/typegoose/typegoose-testing/pull/2).
I also agree that the Copilot comment should be addressed.

PR #16367, opened 2026-07-06
fix(document): don't skip optimistic concurrency when assigning a parent of an excluded nested path

AbdelrahmanHafez:
re #16054 / #16177.

#16177 uses the same path matching for include and exclude. Include needs to match in both directions (`profile.address.country` should be triggered by setting `profile.address`), but for exclude this means that excluding a nested path also excludes its parent. So if you exclude `profile.firstName` and then assign `profile`, the version check is skipped, even though other non-excluded fields under `profile` changed:

```js
  profile: { firstName: String, lastName: String },
  balance: Number
}, { optimisticConcurrency: { exclude: ['profile.firstName'] } });

const user = await User.create({ profile: { firstName: 'Alice', lastName: 'Smith' }, balance: 100 });

user.profile = { firstName: 'Alice', lastName: 'Johnson' };
user.$__delta();

// Before this PR: user.$__.version === undefined (no version check)
// After: user.$__.version === VERSION_ALL, profile.lastName is not excluded and it changed
```

This PR makes exclude only match the path itself or one of its parents, include keeps the current behavior. Copilot flagged this on https://github.com/Automattic/mongoose/pull/16177#discussion_r3023820287

Also removed a duplicate `_applyVersionIncrement` definition I noticed while working on this, the same function was defined twice back to back.

@vkarpov15 @hasezoey part of what I intended with #16054 is still not covered, there are two cases where include/exclude paths never match the runtime modified paths:

1. Map wildcards: `['settings.$*']` matches nothing, in both the include and exclude forms.
2. Arrays of subdocuments: `optimisticConcurrency: ['comments.text']` never triggers versioning, runtime paths look like `comments.0.text` and the matcher doesn't skip index segments. So the only way to protect a subdocument field right now is to include the whole array.

I'd handle these in a future PR by extracting the pathUtils helper we discussed on #16177, something that skips numeric segments and handles `.$*.`, and using it for both branches. WDYT? Should we not bear the maintenace cost until someone asks for it, or just go ahead and implement it?

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -5147,8 +5147,8 @@ Document.prototype.$__delta = function $__delta(pathsToSave, pathsToSaveSet) {
       const optimisticConcurrencyExcludeSet = this.$__schema.options._optimisticConcurrencyExcludeSet;
       const modPaths = this.directModifiedPaths();
       const hasRelevantModPaths = pathsToSave == null ?
-        modPaths.find(path => !_pathOverlapsSet(path, optimisticConcurrencyExcludeSet)) :
-        modPaths.find(path => !_pathOverlapsSet(path, optimisticConcurrencyExcludeSet) && isInPathsToSave(path, pathsToSaveSet, pathsToSave));
+        modPaths.find(path => !_pathOrAncestorInSet(path, optimisticConcurrencyExcludeSet)) :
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
LGTM

vkarpov15 (approved):

PR #16368, opened 2026-07-06
fix(document): validate child paths when `pathsToSave` includes their parent

AbdelrahmanHafez:
`save({ pathsToSave: ['profile'] })` saves a modified `profile.firstName` but never validates it. The save filter matches child paths of each entry, while the validation filter matches exact strings only, so invalid nested values are written with no `ValidationError`:

```js
  profile: {
    firstName: { type: String, minlength: 2 }
  }
});

user.profile.firstName = 'A'; // violates minlength
await user.save({ pathsToSave: ['profile'] }); // resolves and writes 'A'
await user.save(); // would throw ValidationError
```

This PR makes validation use the same child-path matching as the save filter, but only when the validated paths come from `pathsToSave`. Calling `validate(['profile'])` directly is unchanged.

Only plain nested paths were affected: subdocuments and document arrays already validate their children through the parent path.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2705,8 +2705,13 @@ Document.prototype.validate = async function validate(pathsToValidate, options) 
       return;
     }
 
+    // If `pathsToValidate` comes from `pathsToSave`, an entry like `profile` also
+    // saves child paths like `profile.firstName`, so validation must match child
+    // paths the same way the save filter does (gh-9583).
+    let pathsToValidateIncludesChildren = false;
     if (this.$__.saveOptions && this.$__.saveOptions.pathsToSave && !pathsToValidate) {
       pathsToValidate = [...this.$__.saveOptions.pathsToSave];
+      pathsToValidateIncludesChildren = true;
     }
 
     const hasValidateModifiedOnlyOption = options &&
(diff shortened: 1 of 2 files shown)


vkarpov15:
Another possibility would be to flatten out so `user.save({ pathsToSave: ['profile'] })` calls `validate()` with `['profile', 'profile.firstName']` to keep the current semantics of validate. Worth investigating to see which one is faster.

AbdelrahmanHafez:
I tested both approaches, a schema-only flatten handles plain nested paths and required children, but misses runtime paths such as Map keys because `metadata.views` isn't in `schema.paths`. I added `modifiedPaths({ includeChildren: true })` to make the flattened version equivalent to the current PR.

Both implementations pass all tests. Benchmark results (microseconds) per `validate()` call over 5 runs:

| Scenario | Current | Flattened |
|---|---:|---:|
| Small schema, one parent | 2.88 | 3.63 |
| 300 paths, one parent | 68.86 | 79.22 |
| 300 paths, ten parents | 316.69 | 419.92 |
| 300 paths, top-level only | 46.91 | 47.29 |

The current approach was faster in every case and handles runtime paths without separate expansion logic, so I think we should keep it and add a Map test. If you had a different flattening mechanism in mind, I'm happy to benchmark that too.

PR #16370, opened 2026-07-07
InsertMany perf improvements

vkarpov15:
**Summary**

`insertMany()` has a lot of avoidable overhead, particularly on the happy path with no errors and no subdocuments. 

- `insertMany()` doesn't need to create backup for newly created documents
- `$__.activePaths.clear()` would use `delete` to clean up paths, which would de-opt: use `clearAll()` instead for a full reset
- POJO `{}` much faster than `new Map()`
- Get version key path as array to avoid overhead of repeatedly checking the version key path for `'.'`
- Made `parallelLimit` much faster by avoiding using `new Set()`
- Added a new fast path for `$__setValue()`: if just setting a top-level path, avoid all the `mpath.set()` internal checks for `special` etc. and just set the property directly.
- Speed up iteration in `_getPathsToValidate()`, specifically avoid creating a `new Set()` until necessary.
- Add extra check to avoid running hooks if there aren't any hooks
- Check last char for `*` to avoid running full `endsWith()` on every path
- 

`insertManySimple` benchmark before:

image

```
{
  "Average mongoose insertMany time ms": 13.11,
  "Average driver insertMany time ms": 4.92
}
{
  "Average mongoose insertMany time ms": 13.62,
  "Average driver insertMany time ms": 4.96
}
{
  "Average mongoose insertMany time ms": 12.6,
  "Average driver insertMany time ms": 5.03
}
```

After:

image

```
{
  "Average mongoose insertMany time ms": 7.33,
  "Average driver insertMany time ms": 4.47
}
{
  "Average mongoose insertMany time ms": 7.39,
  "Average driver insertMany time ms": 3.94
}
{
  "Average mongoose insertMany time ms": 6.83,
  "Average driver insertMany time ms": 4.42
}

```

This change also improves saveSimple.js benchmark, before:

```
{
  "Average mongoose save time ms": 2.14,
  "Average driver insertOne time ms": 1.36
}
{
  "Average mongoose save time ms": 2.06,
  "Average driver insertOne time ms": 1.42
}
{
  "Average mongoose save time ms": 1.92,
  "Average driver insertOne time ms": 1.47
}
```

After:

```
{
  "Average mongoose save time ms": 1.7,
  "Average driver insertOne time ms": 1.15
}
{
  "Average mongoose save time ms": 1.29,
  "Average driver insertOne time ms": 1.28
}
{
  "Average mongoose save time ms": 1.45,
  "Average driver insertOne time ms": 1.21
}

```

**Examples**


lib/document.js
```diff
@@ -3633,27 +3693,32 @@ Document.prototype.$__reset = function reset() {
...
   // Clear 'dirty' cache
-  this.$__.activePaths.clear('modify');
-  this.$__.activePaths.clear('default');
-  this.$__.validationError = undefined;
-  this.$errors = undefined;
-  _this = this;
-  this.$__schema.requiredPaths().forEach(function(path) {
-    _this.$__.activePaths.require(path);
-  });
+  this.$__.activePaths.clearAll();
```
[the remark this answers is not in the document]
vkarpov15: I'm not convinced that the `init` concern is a problem. However the `ignore` part is a problem, that's something to consider. I won't merge this suggestion as is because of de-opt.

lib/helpers/parallelLimit.js
```diff
@@ -3,35 +3,61 @@
...
-  for (let index = 0; index < params.length; index++) {
-    const param = params[index];
-    const p = fn(param, index);
-    results.push(p);
-
-    executing.add(p);
+  // Fast path: there's nothing to bound, so kick everything off and wait once.
+  // Avoids the per-task `Set` + `Promise.race` bookkeeping of a sliding window.
+  if (limit >= length) {
+    const results = new Array(length);
```
hasezoey: Considering both paths create this value, maybe de-duplication would be good?

lib/document.js
```diff
@@ -3614,17 +3676,15 @@ Document.prototype.$isValid = function(path) {
  * @instance
  */
 
-Document.prototype.$__reset = function reset() {
-  let _this = this;
-
-  const onlyPrimitiveValues = this.$__hasOnlyPrimitiveValues();
+Document.prototype.$__reset = function reset(skipBackup, hasOnlyPrimitiveValues) {
```
hasezoey, line 3706: The new options should be added to the JSDoc

lib/model.js
```diff
@@ -3276,19 +3286,25 @@ async function _insertMany(arr, options) {
...
+  }
+  return docAttributes;
 }
 
 /*!
  * ignore
  */
 
-function _setIsNew(doc, val) {
+function _setIsNew(doc, val, hasOnlyPrimitiveValues) {
```
hasezoey, line 3303: This should now also deserve a JSDoc

hasezoey (commented):
LGTM, with some minor style nitpicks.

lib/document.js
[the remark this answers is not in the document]
AbdelrahmanHafez: I looked into the `init` part of this and it's a real behavior change, the failure needs two saves so no test catches it. `init` only gets marked during hydration and nothing re-marks it after a save, so the first `save()` on a hydrated doc permanently erases it. From the second save onwards, validators on hydrated-but-unmodified paths stop running (by default we validate those on every save, that's what `validateModifiedOnly` opts out of). `isInit()` also flips to false after the first save.

Repro that passes on master and fails on this branch:

```js
'use strict';

run().catch(err => {
  console.error(err);
  process.exit(1);
});

async function run() {
  const mongod = await MongoMemoryServer.create();

  try {
    await mongoose.connect(mongod.getUri(), { dbName: 'test' });

      name: String,
      requiresEmail: Boolean,
      email: {
        type: String,
        validate: {
          validator: function(v) {
            return !this.requiresEmail || (typeof v === 'string' && v.includes('@'));
          },
          message: 'invalid email'
        }
      }
    });

    // Valid at creation time: requiresEmail is false, so the junk email passes
    const { _id } = await User.create({ name: 'John', requiresEmail: false, email: 'not-an-email' });

    const user = await User.findById(_id);

    // Save #1: touch an unrelated field. Passes on both master and this branch.
    user.name = 'Johnny';
    await user.save();

    // Save #2: flip requiresEmail, which makes the hydrated `email` invalid.
    user.requiresEmail = true;
    await assert.rejects(() => user.save(), /invalid email/);

    console.log('PASS');
  } finally {
    await mongoose.disconnect();
    await mongod.stop();
  }
}
```

The second save rejects on master, on this branch it resolves and persists `requiresEmail: true` next to the invalid email. It also makes validation depend on save history rather than document state, meaning the same invalid combination is rejected if you set it on a freshly loaded doc, but accepted if any unrelated save happened in between.

The `ignore` part is benign as far as I can tell: `$ignore()` works by moving the path out of `modify` at call time, and nothing ever reads the `ignore` bucket itself (the only writer is `$ignore()`, and there's no `getStatePaths('ignore')` anywhere), so losing it on reset changes nothing observable.

One way to keep the perf win: only use `clearAll()` when the doc has no init'd paths, which is always true for the fresh docs `insertMany()` creates, and fall back to the targeted `clear('modify')` + `clear('default')` otherwise.

lib/helpers/parallelLimit.js
```diff
@@ -3,35 +3,60 @@
...
+  for (let i = 0; i < workers.length; ++i) {
+    workers[i] = worker();
+  }
+  await Promise.all(workers);
+
+  if (firstError !== null) {
+    throw firstError;
+  }
+  return results;
 }
```
AbdelrahmanHafez, line 62: `validateDoc` isn't `async` anymore, so a non-object entry throws `ObjectParameterError` synchronously mid-loop here, and the promises already in `results` never get handlers attached. If one of them rejects, that's an unhandled rejection, which crashes the process on modern Node. The old implementation attached `.catch(clean)` to every promise as it was created. The worker path below is fine.

Prints PASS on master, crashes on this branch:

```js
'use strict';

process.on('unhandledRejection', err => {
  console.error('unhandledRejection:', err.message);
  process.exit(1);
});

run().catch(err => {
  console.error(err);
  process.exit(1);
});

async function run() {
  const User = mongoose.model('User', new mongoose.Schema({
    email: { type: String, validate: () => false }
  }));

  // Index 0 fails validation async, index 1 throws ObjectParameterError sync.
  await User.insertMany([{ email: 'x' }, 42]).catch(() => {});

  await new Promise(resolve => setTimeout(resolve, 100));
  console.log('PASS');
}
```

lib/schemaType.js
```diff
@@ -1303,7 +1303,7 @@ SchemaType.prototype.getDefault = function getDefault(parentDoc, init, options)
   }
 
   if (ret != null) {
-    if (typeof ret === 'object' && !this.options?.shared) {
+    if (typeof ret === 'object' && ret != null && ret._bsontype !== 'ObjectId' && !this.options?.shared) {
```
AbdelrahmanHafez, line 1307: This skips the clone for every ObjectId default, including static ones (`default: someObjectId`) and function defaults that return a cached instance (`default: () => SHARED_ID`), both of which master clones. Every doc created without the field then shares the same ObjectId wrapper as the schema default, so identity comparisons, `.id` reassignment, and any properties set on the instance leak across documents. Master already aliases the underlying byte buffer (`clone()` does `new ObjectId(obj.id)`, which doesn't copy in bson 7), but it keeps a separate wrapper per doc.

```js
'use strict';

  orgId: { type: mongoose.Schema.Types.ObjectId, default: new mongoose.Types.ObjectId() }
});

// Passes on master (each doc gets its own wrapper), fails on this branch
assert.notStrictEqual(new User({}).orgId, new User({}).orgId);
console.log('PASS');
```

The perf win here is the auto `_id` default, which is a mongoose-owned function that always returns a fresh instance (`defaultId` in `lib/schema/objectId.js`). We already stamp flags on that function (`$runBeforeSetters`), so we could flag it and key the skip on provenance instead of type, which keeps the whole hot-path win with no user-facing behavior change:

```js
// lib/schema/objectId.js
defaultId.$fresh = true;

// here
if (typeof ret === 'object' && !this.options?.shared && !this.defaultValue?.$fresh) {
```

Side note, function-generated `Decimal128`/UUID/Buffer defaults have the same double allocation and still clone, but none of them are on the `_id` hot path, so probably not worth touching in this PR.

AbdelrahmanHafez (commented):
A couple of comments, otherwise great work!

lib/document.js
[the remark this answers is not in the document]
vkarpov15: You're right, `ignore` gets overridden by updates so ignore is a non-issue. Largely equivalent to `unmarkModified`. `doc.$ignore('name'); doc.name = 'test2';` would still save `test2`.

'init' is a problem here, you're right. I'm making this a `clearAllExcept('init')` instead

lib/schemaType.js
vkarpov15: I believe skipping clone for every ObjectId default is acceptable because ObjectIds do not support in-place updates. So even in the odd chance that someone uses `default: new mongoose.Types.ObjectId()`, which would be strange, that ObjectId instance can't be modified in any reasonable way. Yes you can attach properties but I don't think that's worth the extra perf overhead. Do you feel strongly about this @AbdelrahmanHafez ?

AbdelrahmanHafez: I missed this comment, you're right. 👍

PR #16371, opened 2026-07-07
docs(populate): document missing populate() options in Query and Document, add missing PopulateOptions TS fields

Wnayar:
**Summary**

`perDocumentLimit` is documented in the options list for `Model.populate()` but is missing from the `Query.prototype.populate()` reference, even though it's commonly used with query populate (`Model.find().populate({ path: '...', perDocumentLimit: 2 })`). The `Query.prototype.populate()` docs mention `limit` (via `options.options`) but not `perDocumentLimit`, which is the option that gives correct per-document limit behavior. This adds the same `perDocumentLimit` entry to the `Query.prototype.populate()` options list, using the existing wording from `Model.populate()`.

**Examples**
```js
// `limit` limits across the combined populate query
await Book.find().populate({ path: 'authors', options: { limit: 2 } });

// `perDocumentLimit` applies the limit per parent document
await Book.find().populate({ path: 'authors', perDocumentLimit: 2 });
```


AbdelrahmanHafez (approved):
Welcome @Wnayar, and thanks!

hasezoey (approved):

Wnayar:
Thanks for the welcome and for expanding this to cover the other missing options, @AbdelrahmanHafez!

vkarpov15 (approved):
Thanks :+1:

PR #16372, opened 2026-07-07
fix(schema): avoid mutating shared setters array in SchemaArray._castForQuery

### LINKED ISSUE #16364: SchemaArray._castForQuery mutates the shared setters array, corrupting setter order after an array query
### Description
`SchemaArray._castForQuery` calls `this.setters.reverse()`, which mutates the schema type's **persistent** `setters` array in place. A single array-valued query permanently flips the stored setter order for that path, corrupting setter application order for both subsequent queries **and** document creation. Every other reversal in the codebase iterates by index without mutating.

### Steps to reproduce
```js
const log = [];
const schema = new mongoose.Schema({ tags: { type: [String] } });
const p = schema.path('tags');
p.set(v => { log.push('A'); return v; }); // defined 1st
p.set(v => { log.push('B'); return v; }); // defined 2nd
const M = mongoose.model('A', schema);
const order = () => { log.length = 0; new M({ tags: ['x'] }); return log.join(''); };

console.log(order());              // "BABA"
M.find({ tags: ['x'] }).cast(M);   // one array-valued query
console.log(order());              // "ABAB"  <- setter order flipped
```
Observed: creating an identical document yields setter order `BABA`, then `ABAB` after a single unrelated array query. Expected: setter order is a stable property of the schema, independent of query history.

### Root cause
`lib/schema/array.js:513`:
```js
this.setters.reverse().forEach(setter => { val = setter.call(this, val, this); });
```
`this.setters` is the schema type's persistent instance array (`lib/schematype.js`, a one-time `.slice()`), and `Array.prototype.reverse()` mutates it in place. `_applySetters` then iterates `this.setters` by index relying on definition order, so once reversed, setters run in the wrong order.

### Suggested fix
Avoid mutating the shared array: `this.setters.slice().reverse()` (or iterate by index without reversing), matching the non-mutating pattern used elsewhere.

yakubka:
Fixes #16364.

SchemaArray._castForQuery calls this.setters.reverse() on line 513 of lib/schema/array.js. Array.prototype.reverse mutates the array in place, so the persistent setters array on the schema type is permanently reordered after the first array-valued query. All subsequent queries and document creations then apply setters in the reversed order, which is incorrect.

Every other place in the codebase that needs reversed iteration either iterates by index in reverse or uses a copy. This line is the only one that mutates.

The fix is a one-character change: this.setters.slice().reverse(). slice() creates a shallow copy, so the forEach iterates the reversed copy while the original array remains in definition order.

The reproduction case from the issue report confirms the fix:

Before the fix, creating a document after one array-valued query yields setter order ABAB instead of BABA. After the fix, setter order is BABA in both cases.

Yokubjon Nurullaev traced the mutation to lib/schema/array.js:513, confirmed the impact on setter ordering across queries and document creation, and implemented the slice fix.

IbrokhimN reviewed the fix against the other reverse usages in the codebase to confirm this was the only mutating call, and verified that the fix matches the non-mutating patterns used in schematype.js.

### DIFF
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -510,9 +510,9 @@ SchemaArray.prototype._castForQuery = function(val, context) {
   }
 
   if (Array.isArray(val)) {
-    this.setters.slice().reverse().forEach(setter => {
-      val = setter.call(this, val, this);
-    });
+    for (let i = this.setters.length - 1; i >= 0; i--) {
+      val = this.setters[i].call(this, val, this);
+    }
     val = val.map(function(v) {
       if (utils.isObject(v) && v.$elemMatch) {
         return v;
--- a/lib/schema/array.js
+++ b/lib/schema/array.js
@@ -510,7 +510,7 @@ SchemaArray.prototype._castForQuery = function(val, context) {
   }
 
   if (Array.isArray(val)) {
-    this.setters.reverse().forEach(setter => {
+    this.setters.slice().reverse().forEach(setter => {
       val = setter.call(this, val, this);
     });
     val = val.map(function(v) {
--- a/test/types.array.test.js
+++ b/test/types.array.test.js
@@ -1959,12 +1959,14 @@ describe('types array', function() {
     assert.equal(called, 2);
   });
 
-  it('does not mutate SchemaArray setters order when casting (gh-16372)', function() {
+  it('does not mutate SchemaArray setters order when casting an array-valued query filter (gh-16372)', function() {
     const schema = new Schema({ nums: { type: [Number], set: v => v } });
     schema.path('nums').set(v => v);
     const originalOrder = schema.path('nums').setters.slice();
     const M = db.model('Test', schema);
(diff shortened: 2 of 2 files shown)


AbdelrahmanHafez (changes requested):
Welcome @yakubka 
A couple of requests:
1. Could you add test(s) that demonstrate the bug and prevent future regressions?
2. Also use a normal backward for loop instead of creating a shallow copy of the array on each execution.

vkarpov15 (changes requested):
This is a good bug to fix, but I agree with @AbdelrahmanHafez 1) use for loop instead of copying the array, 2) add test case please

yakubka:
Updated the PR to address the review feedback.

Switched from slice().reverse().forEach to a backward for loop as requested. Also added a regression test, does not mutate SchemaArray setters order when casting (gh-16372), to cover the case of multiple setters on a path.

Let me know if anything else needs to change.

AbdelrahmanHafez (approved):
Thanks!

PR #16374, opened 2026-07-08
types(model): support extra fields type argument in hydrate() with strict: false

AbdelrahmanHafez:
re https://github.com/Automattic/mongoose/discussions/15940#discussioncomment-17566645

Follow-up to #15944, we agreed [in the review thread](https://github.com/Automattic/mongoose/pull/15944#discussion_r2665981444) to look into TS support for extra fields with `strict: false` later.

```ts
const user = User.hydrate<{ totalOrders: number }>(rawUser, null, { strict: false });
user.totalOrders; // number
```

The type param defaults to `unknown` (`T & unknown` is exactly `T`), and the plain signature stays last, so `ReturnType<(typeof Model)['hydrate']>` and existing calls are unaffected. Types only, no runtime changes.

Also widened `projection` to `ProjectionType<TRawDocType> | null | undefined` on both signatures, `hydrate(obj, null)` used to be a type error even though the runtime handles it, and the new overload needs the explicit union since its `projection` can't be optional.

### DIFF
--- a/test/types/models.test.ts
+++ b/test/types/models.test.ts
@@ -1351,3 +1351,46 @@ function hydrateWithStrictOption() {
 
   expect(doc4).type.toBe<ReturnType<(typeof TestModel)['hydrate']>>();
 }
+
+function gh15940() {
+  interface IUser {
+    name: string;
+    age: number;
+  }
+
+  const schema = new mongoose.Schema<IUser>({
+    name: String,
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
LGTM

vkarpov15 (approved):
Thanks :+1:

PR #16380, opened 2026-07-09
fix(cast): use current castForQuery() signature for object-shaped geo values

### LINKED ISSUE #16376: Geo query with an object-shaped value throws Cannot use [object Object] with Number (wrong castForQuery argument shape)
### Description
A geo query whose geometry is expressed as an object with a scalar leaf throws a confusing internal error instead of casting the value. `lib/cast.js` uses the pre-v7 single-object `castForQuery` signature in the object branch, while the sibling array branch (12 lines up) uses the current 3-arg form.

### Root cause
`lib/cast.js`:
```js
val[nkey] = numbertype.castForQuery({ val: item, context: context }); // legacy single-object arg
```
The current signature is `castForQuery($conditional, val, context)`, and every other call site (including the sibling array branch) passes `castForQuery(null, item, context)`. Here the object lands in the `$conditional` slot; being non-null, `SchemaNumber.castForQuery` does `this.$conditionalHandlers[$conditional]`, which stringifies the object to `"[object Object]"`, finds no handler, and throws.

### Steps to reproduce
```js
const schema = new mongoose.Schema({ name: String }, { strict: false, strictQuery: false });
const M = mongoose.model('B', schema);
M.find({ loc: { $geoIntersects: { foo: 5 } } }).cast(M);
// throws: MongooseError: Can't use [object Object] with Number.
```
The array-valued sibling case (`{ $geoIntersects: { coordinates: [1,2] } }`) works, proving the object branch alone is broken.

### Suggested fix
Change the object-branch call to `numbertype.castForQuery(null, item, context)`, matching the sibling array branch.

rajkumar0932:
## Summary

Fixes #16376

`_cast()` in `lib/cast.js` has two sibling branches for geo values: the array branch uses the current 3-arg `castForQuery(null, item, context)`, but the object branch still passed the legacy single-object argument (`castForQuery({ val: item, context })`). That object lands in the `$conditional` parameter, gets stringified to `"[object Object]"`, finds no conditional handler, and throws `Can't use [object Object] with Number` — instead of casting the value.

```js
const M = mongoose.model('B', schema);
M.find({ loc: { $geoIntersects: { foo: 5 } } }).cast(M);
// before: throws "Can't use [object Object] with Number"
// after:  casts the leaf like the array branch does
```

This changes the object-branch call to match the sibling array branch (and every other call site in the file).

## Examples

Before/after verified with a standalone script against `lib/cast` directly:

```
before: OBJECT branch THROWS: Can't use [object Object] with Number.
after:  OBJECT branch OK: {"loc":{"$geoIntersects":{"foo":5}}}
```

Array-shaped geo values (`{ coordinates: ['1','2'] }`) and `$near`/`$maxDistance` casting are unchanged.

Added two regression tests in `test/cast.test.js` (object-shaped leaf casts; array path stays intact).

### DIFF
--- a/lib/cast.js
+++ b/lib/cast.js
@@ -413,7 +413,7 @@ function _cast(val, numbertype, context) {
         _cast(item, numbertype, context);
         val[nkey] = item;
       } else {
-        val[nkey] = numbertype.castForQuery({ val: item, context: context });
+        val[nkey] = numbertype.castForQuery(null, item, context);
       }
     }
   }
--- a/test/cast.test.js
+++ b/test/cast.test.js
@@ -104,6 +104,26 @@ describe('cast: ', function() {
     });
   });
 
+  describe('geo queries', function() {
+    it('casts object-shaped geo values with scalar leaves (gh-16376)', function() {
+      const schema = new Schema({ name: String });
+
+      const res = cast(schema, {
+        loc: { $geoIntersects: { foo: '5' } }
+      }, { strictQuery: false });
+      assert.strictEqual(res.loc.$geoIntersects.foo, 5);
+    });
+
+    it('still casts array-shaped geo values (gh-16376)', function() {
+      const schema = new Schema({ name: String });
+
+      const res = cast(schema, {
+        loc: { $geoIntersects: { coordinates: ['1', '2'] } }
+      }, { strictQuery: false });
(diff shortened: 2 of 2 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #16385, opened 2026-07-13
perf(document): cache selected-path lookup to speed up isSelected()

### LINKED ISSUE #16373: toObject({getters:true}) on projected documents: applyGetters iterates all schema paths and each unselected path triggers an O(projectionKeys) isSelected scan — O(paths x projection) per doc
## toObject({getters:true}) on projected documents: applyGetters iterates all schema paths and each unselected path triggers an O(projectionKeys) isSelected scan — O(paths x projection) per doc

**Location:** lib/document.js:4394 (applyGetters per-path $__isSelected call) x lib/document.js:2572 (isSelected projection-key scan)
**Severity:** medium · **Confidence:** 0.75
**Type:** complexity-at-a-distance

### Description

applyGetters in lib/document.js (line ~4371) loops over every key of `schema.paths` and calls `self.$__isSelected(path)` for each. Document.prototype.isSelected (line ~2525) is O(S) for any path not literally present in the projection: after the O(1) `path in this.$__.selected` check fails, it loops all projection keys doing startsWith comparisons (lines ~2572-2584). So serializing one projected document with getters enabled (a very common toJSON config) costs O(P x S) where P = schema paths and S = projection keys — per document, per serialization. The chain crosses three modules: lib/query.js `_completeMany` builds projected docs (fields propagated into the Document constructor, `$__.selected` set at lib/document.js:158), then user/toJSON triggers `$toObject` -> applyGetters -> isSelected. Each function looks fine locally; the product only appears across the boundary.

### Benchmark

With schema width 2N and an inclusive projection of N keys, toObject({getters:true}) time should grow superlinearly, tracking the O(N^2) prediction (N=500 hundreds-to-thousands of times slower per pass than N=10) because each of the ~N unselected paths scans the N projection keys.

**Observed complexity:** O(N·M)

```
| N=10 | 2.75 | 1.00 | — |
| N=50 | 17.59 | 6.40 | — |
| N=100 | 68.20 | 24.79 | — |
| N=500 | 2199.44 | 799.43 | — |
```

benchmark confirmed

GiHoon1123:
**Summary**

Fixes #16373.

`toObject({ getters: true })` on a projected document (i.e. one loaded with `.select(...)`) gets slower than expected as the schema grows, because of how `applyGetters()` and `isSelected()` interact:

- `applyGetters()` (called from `$toObject()`) loops over **every schema path** and calls `isSelected(path)` for each one, to decide whether that path's value/getter should be included.
- `isSelected(path)`, when `path` isn't an exact key in the projection object, re-scans the **entire projection object** on every single call, comparing path prefixes to determine ancestor/descendant relationships (e.g. is `path` nested under a selected key, or does a selected key exist underneath `path`).

So one `toObject()` call costs roughly O(schema paths × projection size) - re-scanning the same, unchanging projection object once per schema path.

**Fix**

`$__.selected` (the projection) is set once at document construction and never changes for the life of the document, so the ancestor/descendant lookup it requires can be precomputed once and reused, instead of rebuilt on every `isSelected()` call.

Added `buildSelectedInfo()` (`lib/helpers/projection/buildSelectedInfo.js`), which builds a trie over the projection's dot-separated keys, cached on `$__.selectedInfo`. Each trie node records the insertion-order index of the projection key it represents (if any), plus the minimum such index among its descendants (precomputed bottom-up when the trie is built). `isSelected()` then walks the trie by `path`'s own segments - O(path depth) instead of O(projection size) - to determine the same ancestor/descendant relationship the original per-key loop computed.

One subtlety: a projection can (rarely) select both a path and a deeper path underneath it, e.g. `{ a: 1, 'a.b.c': 1 }`. The original code resolves this by returning on whichever of the two relationships (path is an ancestor of a selected key, or path is a descendant of one) is found first while looping over the projection object in insertion order. The trie preserves this by tracking each key's insertion index and comparing the ancestor-side and descendant-side candidates' indices, so the result matches the original implementation exactly rather than picking an arbitrary tie-break.

**Verification**

- Compared the original per-key-loop algorithm against the new trie-based one across 2 million randomly generated `(projection, path)` combinations (5 seeds x 400k cases each), including projections that select both a path and a deeper path underneath it - 0 mismatches.
- Added a regression test (`test/document.isselected.test.js`) pinning down that specific tie-break behavior, passes against both the original and the new implementation (i.e. it documents preserved behavior, not new behavior).
- Benchmark (`benchmarks/toObjectWithProjectionGetters.js`, no live MongoDB connection needed - constructs documents directly the same way query hydration does): a document with a 1000-path schema and a 500-path projection serializes about 8x faster (4.27ms -> 0.49ms per `toObject({ getters: true })` call), and scales much closer to linearly with schema size (before: ~630x slower at 50x the schema width; after: ~120x).
- Full existing test suite: `document.isselected.test.js`, `document.test.js`, `model.field.selection.test.js`, `model.hydrate.test.js`, `model.querying.test.js`, `model.populate.test.js` all pass (1000+ tests).
- `eslint .` clean on all changed/added files.

**Examples**

See `benchmarks/toObjectWithProjectionGetters.js` for a runnable before/after comparison (checkout the parent commit to compare against the original implementation).


benchmarks/toObjectWithProjectionGetters.js
```diff
@@ -0,0 +1,59 @@
...
+
+function bench(numPaths, projectedCount, iterations) {
+  const doc = buildModelAndDoc(numPaths, projectedCount);
+  // warm up the JIT before measuring
+  for (let i = 0; i < 200; i++) {
+    doc.toObject({ getters: true });
+  }
+  const start = process.hrtime.bigint();
+  for (let i = 0; i < iterations; i++) {
+    doc.toObject({ getters: true });
```
vkarpov15, line 48: I don't think this benchmark is very realistic - executing `toObject()` repeatedly on one document is not a common use case. However, it _does_ look like `isSelected()` is called in `_getPathsToValidate()` for required documents, which is a much more likely cause of overhead. I'm looking into that.

lib/helpers/projection/buildSelectedInfo.js
```diff
@@ -0,0 +1,86 @@
...
+  const root = { children: null, index: undefined, key: undefined, minDescendant: null };
+  keys.forEach((key, index) => {
+    if (key === '_id') {
+      return;
+    }
+    const parts = key.split('.');
+    let node = root;
+    for (const part of parts) {
+      if (node.children == null) {
+        node.children = new Map();
```
vkarpov15: `new Map()` is significantly slower than creating a POJO `{}`, so performance would likely be faster if this were `node.children = {}`. 

Image

GiHoon1123:
Switched this from `Map` to a null-prototype object.

I also added a `doc.validate()` benchmark that exercises `_getPathsToValidate()`
with projected docs and many required paths. On my machine, the largest case in
that benchmark (1000 required paths, 500 projected) goes from ~9.36ms/call on
master to ~6.12ms/call with this PR.

vkarpov15:
I ran a more focused benchmark on `isSelected()`:

```
'use strict';

// Measures only Document#isSelected(). Document/model construction is kept
// outside the timed sections. The first-call measurement includes any lazy
// projection cache construction, while the repeated-call measurement shows
// the steady-state lookup cost.

  user: {
    profile: {
      contact: {
        email: String,
        phone: String,
        address: {
          city: String,
          country: String
        }
      },
      name: String
    },
    settings: {
      theme: String,
      locale: String
    }
  },
  createdAt: Date,
  status: String
}, { versionKey: false });

const Model = mongoose.model('IsSelectedBenchmark', schema);
const projection = {
  'user.profile.contact.email': 1,
  'user.profile.contact.phone': 1,
  'user.profile.name': 1,
  'user.settings.theme': 1,
  status: 1,
  _id: 0
};
const lookupPaths = [
  'user.profile.contact.email',
  'user.profile.contact.phone',
  'user.profile.contact.address.city',
  'user.profile.name',
  'user.settings.locale',
  'status'
];

const FIRST_CALL_DOCS = 10000;
const STEADY_STATE_CALLS = 1000000;

function makeDoc() {
  return new Model({}, { ...projection });
}

function nsPerFirstCall() {
  const docs = Array.from({ length: FIRST_CALL_DOCS }, makeDoc);
  let selected = 0;
  for (const doc of docs) {
    selected += doc.isSelected(lookupPaths[0]) ? 1 : 0;
  }
  const elapsed = process.hrtime.bigint() - start;
  if (selected !== FIRST_CALL_DOCS) {
    throw new Error('Unexpected isSelected() result');
  }
  return Number(elapsed) / FIRST_CALL_DOCS;
}

function nsPerSteadyStateCall() {
  const doc = makeDoc();
  // Build the lazy cache and warm up the lookup before measuring.
  doc.isSelected(lookupPaths[0]);
  for (let i = 0; i < 10000; ++i) {
    doc.isSelected(lookupPaths[i % lookupPaths.length]);
  }

  let selected = 0;
  for (let i = 0; i < STEADY_STATE_CALLS; ++i) {
    selected += doc.isSelected(lookupPaths[i % lookupPaths.length]) ? 1 : 0;
  }
  const elapsed = process.hrtime.bigint() - start;
  if (selected === 0) {
    throw new Error('Unexpected isSelected() result');
  }
  return Number(elapsed) / STEADY_STATE_CALLS;
}

const first = nsPerFirstCall();
const steady = nsPerSteadyStateCall();
const result = {
  projectionKeys: Object.keys(projection).length,
  lookupDepth: 4,
  'first call (ns/call)': Number(first.toFixed(2)),
  'steady state (ns/call)': Number(steady.toFixed(2)),
  'first / steady ratio': Number((first / steady).toFixed(2)),
  'estimated total for 1 call (ns)': Number(first.toFixed(2)),
  'estimated total for 10 calls (ns)': Number((first + steady * 9).toFixed(2)),
  'estimated total for 100 calls (ns)': Number((first + steady * 99).toFixed(2))
};

console.log(JSON.stringify(result, null, 2));
```

The output shows that there's a non-negligible cost to building up the trie - first call is slow, subsequent calls are fast. With this patch:

```
{
  "projectionKeys": 6,
  "lookupDepth": 4,
  "first call (ns/call)": 4232.28,
  "steady state (ns/call)": 79.06,
  "first / steady ratio": 53.53,
  "estimated total for 1 call (ns)": 4232.28,
  "estimated total for 10 calls (ns)": 4943.81,
  "estimated total for 100 calls (ns)": 12059.17
}
```

Without this patch:

```
{
  "projectionKeys": 6,
  "lookupDepth": 4,
  "first call (ns/call)": 603.51,
  "steady state (ns/call)": 304.05,
  "first / steady ratio": 1.98,
  "estimated total for 1 call (ns)": 603.51,
  "estimated total for 10 calls (ns)": 3340,
  "estimated total for 100 calls (ns)": 30704.92
}
```

So looks like the break-even would be somewhere around 10-20 calls to isSelected. There's definitely a performance tradeoff here: significantly slower for smaller schemas with fewer required fields. If `toObject({ getters: true })` is not called then this change may actually be a net negative. WDYT @GiHoon1123 ?

GiHoon1123:
I pushed an update to this branch.

The trie is no longer built by an isolated `isSelected()` call. `applyGetters()` and validation build it only when they are about to make more than 10 lookups. Documents using the same projection object share the trie through a `WeakMap`.

The update adds cold-path benchmarks, threshold and validation tests, and a memory benchmark comparing shared and per-document tries. The cache assumes the projection object is unchanged after hydration, which matches the current hydration path.

The full test suite passes: 4423 passing, 44 pending.

vkarpov15:
Closing in favor of #16407: creating a trie has substantial performance and maintenance overhead, and isn't going to be faster than the changes in #16407 unless we're looking at 500+ projected fields with custom getters. We can investigate using tries for lookups in large projections sometime in the future, but for now there's a lot of lower hanging fruit we can get for improving performance.

PR #15809, opened 2025-11-28
fix(model): bump version if necessary after successful bulkSave()

vkarpov15:
Fix #15800

**Summary**

Make `bulkSave()` increment the document's version on successful save.

Need to backport this to 7.x.

**Examples**

Issue #15800, opened 2025-11-27
mongo version bug,lose data


spiritwindy:
### Prerequisites

- [x] I have written a descriptive issue title
- [ ] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

7.5.0

### Node.js version

22

### MongoDB server version

5.0

### Description

```ts
import mongoose from 'mongoose';

async function test() {
  // Schema
  const userSchema = new mongoose.Schema({
    name: {
      type: Array,
      required: true
    },
    email: {
      type: String,
      required: true
    }
  });
  // {versionKey: false}

  await mongoose.connect('mongodb://10.21.210.79:27017/test22?authSource=admin&directConnection=true');

  const User = mongoose.model('User', userSchema);
  const user1 = new User({ name: ['123'], email: '12314' });
  await user1.save();

  const user = await User.findOne({ _id: user1._id });
  if (!user) {
    console.log('no user');
    return;
  }
  //  must add this
  // db __v +1 but model don't add 1
  user.markModified('name');
  const n = await User.bulkSave([user]);
  //  must

  // user.name.push("123")
  // user.markModified("name")
  // let b = await User.bulkSave([user])
  //
  // console.log('db save b:', b);
  user.email = '1375';
  user.markModified('email');
  const c = await User.bulkSave([user]);
  console.log('db save c:', c);
  // no change
}
```

### Steps to Reproduce

db save c: BulkWriteResult {
  insertedCount: 0,
  matchedCount: 0,
  modifiedCount: 0,
  deletedCount: 0,
  upsertedCount: 0,
  upsertedIds: {},
  insertedIds: {}
}

### Expected Behavior

  user.email = "1375" should save

PR #16386, opened 2026-07-13
fix(model): make bulkSave() version handling of new documents consistent with save()

AbdelrahmanHafez:
re #15800, follow-up to #15809.

`bulkSave()` treated new documents differently from `save()` in two ways, and together they caused data loss when `increment()` was called on a new document:

- inserts didn't write the version key at all, while `save()` initializes `__v` to 0
- after the insert, `handleSuccessfulWrite()` applied the version bump anyway, so the in-memory `__v` became 1 while the database had no `__v`. The next update that filters on the version key matched nothing and threw:

```js
const user = new User({ name: 'Test User', items: [{ name: 'item1' }] });
user.increment();
await User.bulkSave([user]);
// in-memory __v is 1, DB has no __v

user.items[0].name = 'updated-item';
await User.bulkSave([user]);
// throws "was not able to update 1 of the given documents", update lost
```

This PR makes both match `save()`: inserts initialize the version key to 0, and the version bump is skipped for inserted documents so a pending `increment()` applies on the next save.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3662,9 +3662,7 @@ async function handleSuccessfulWrite(document, options) {
   }
 
   document.$__reset();
-  // Like `$__save()`, inserts skip the version bump: the version key was
-  // initialized to 0 on insert, so a pending `increment()` applies on the
-  // next save instead (gh-15800)
+  // version key gets initialized to 0 when inserting a new document so avoid incrementing version key here
   if (!wasNew) {
     document._applyVersionIncrement();
   }
--- a/lib/model.js
+++ b/lib/model.js
@@ -3662,9 +3662,9 @@ async function handleSuccessfulWrite(document, options) {
   }
 
   document.$__reset();
-  // Like `$__save()`, inserts skip version handling: the insert does not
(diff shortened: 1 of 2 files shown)


hasezoey (commented):
LGTM

lib/model.js
```diff
@@ -3656,12 +3656,18 @@ async function buildPreSavePromise(document, options) {
...
-  if (document.$isNew) {
+  const wasNew = document.$isNew;
+  if (wasNew) {
     _setIsNew(document, false);
   }
 
   document.$__reset();
-  document._applyVersionIncrement();
+  // Like `$__save()`, inserts skip the version bump: the version key was
+  // initialized to 0 on insert, so a pending `increment()` applies on the
```
vkarpov15: Minor nitpick, "pending `increment()`" is a strange term here because the `increment()` state gets wiped out after save, so there's no pending increment left.

It is a little weird that `increment()` doesn't increase the version to `1` for new documents, but we can justify that by saying that new documents always have version `0`.

PR #15648, opened 2025-09-27
fix(update): avoid throwing error if update has a top-level $addToSet

vkarpov15:
Fix #15642

**Summary**

#15642 is caused by `pref` being undefined, which is expected because logic should be using `prefix` instead. I removed the `pref` vs `prefix` distinction to prevent future confusion.

**Examples**

Issue #15642, opened 2025-09-23
Unhandled exception occur when trying to update document


KhaledMosaad:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

8.18.2

### Node.js version

22.12

### MongoDB server version

8.0

### Description

I have an issue when try to update a document with reach embedded documents, the stack trace of the error is: 
```
TypeError: Cannot read properties of undefined (reading 'split')
    at Schema.hasMixedParent (/app/node_modules/mongoose/lib/schema.js:1934:25)
    at Schema.path (/app/node_modules/mongoose/lib/schema.js:1233:23)
    at Schema._getSchema (/app/node_modules/mongoose/lib/schema.js:2876:28)
    at walkUpdatePath (/app/node_modules/mongoose/lib/helpers/query/castUpdate.js:384:27)
    at castUpdate (/app/node_modules/mongoose/lib/helpers/query/castUpdate.js:144:7)
    at model.Query._castUpdate (/app/node_modules/mongoose/lib/query.js:4874:10)
    at model.Query._updateThunk (/app/node_modules/mongoose/lib/query.js:4062:25)
    at model.Query._updateOne (/app/node_modules/mongoose/lib/query.js:4150:23)
    at model.Query.exec (/app/node_modules/mongoose/lib/query.js:4627:80)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at Object.updateSettings (file:///app/src/modules/assessments/assessments/services/assessments.js:843:9)
    at updateSettings (file:///app/src/modules/assessments/assessments/controller/index.js:73:7)
```

### Steps to Reproduce

I can't reproduce it systematically

vkarpov15:
The most likely explanation is that `updateSettings` has a update call like this: `Model.updateOne({}, { $addToSet: { $each: ['test'] } });` where `$addToSet` is in the top-level. This update is incorrect and would be a no-op if it didn't throw an error, but that being said the error you're seeing is a bug that we will fix with #15648.

KhaledMosaad:
Hey @vkarpov15 
[...]

PR #16387, opened 2026-07-14
fix(query): reject all pathless update modifiers

AbdelrahmanHafez:
re #15642 #15648 #15670

`walkUpdatePath()` currently rejects `$each`, `$or`, `$and`, and `$in` when they appear without a field path, but the check only runs when the modifier value is not a plain object. It also only covers those four modifiers.

That leaves malformed updates such as these untreated:

```js
{ $push: { $each: { name: 'admin' } } }
{ $pull: { $nin: ['admin'] } }
```

With the default strict mode, these can be silently stripped into an empty update. With `strict: false`, they can be passed through to MongoDB.

This PR:

- moves pathless modifier validation to the start of `walkUpdatePath()`, before `$pull` handling and before branching on the value shape
- performs the check within the existing traversal, without adding a second pass over the update
- uses the existing operator classifier instead of maintaining a partial modifier allowlist
- resolves operator-like keys against the schema once, then reuses that result during casting, so actual dollar-prefixed paths including `$each` continue to work without another lookup, preserving the behavior from #13786
- includes a concrete correction in the error using the actual operator and misplaced modifier, for example `{ $addToSet: { fieldName: { $each: [...] } } }`

The tests cover array-valued and object-valued `$each`, a pathless `$nin`, a correctly nested `$nin`, a modifier-named `$each` schema path, and the actionable error message.

### DIFF
--- a/lib/helpers/query/castUpdate.js
+++ b/lib/helpers/query/castUpdate.js
@@ -238,6 +238,7 @@ function walkUpdatePath(schema, obj, op, options, context, filter, prefix) {
 
     if (!prefix && isOperator(key) && schema._getSchema(key) == null) {
       throw new MongooseError('Invalid update: Unexpected modifier "' + key + '" as a key in operator "' + op + '". '
+        + 'Did you mean something like { ' + op + ': { fieldName: { ' + key + ': [...] } } }? '
         + 'Modifiers must appear under a valid field path.');
     }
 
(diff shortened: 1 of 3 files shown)


hasezoey (approved):

hasezoey (commented):
LGTM

lib/helpers/query/castUpdate.js
vkarpov15, line 271: This is the same thing as `fullPath` so no longer necessary. Also it looks like `prefix + key` is faster than `prefix ? prefix + key : key` so would be doubly good to remove this.

Image

vkarpov15 (approved):
One minor comment but not a blocker, LGTM overall

PR #16388, opened 2026-07-16
perf(document): avoid rebuilding modifiedPaths per required path during validation

### LINKED ISSUE #16379: Document.isModified(path) rebuilds the full modifiedPaths() set on every call — O(N^2) when checked per-path in loops (including mongoose's own validation)
## Document.isModified(path) rebuilds the full modifiedPaths() set on every call — O(N^2) when checked per-path in loops (including mongoose's own validation)

**Location:** lib/document.js:2339 (isModified -> modifiedPaths rebuild) called per-path from lib/document.js:2965 (_getPathsToValidate) and user hooks
**Severity:** medium · **Confidence:** 0.8
**Type:** complexity-at-a-distance

### Description

Document.prototype.isModified in lib/document.js (line ~2321): when `path` is not directly modified, it falls through to `this[documentModifiedPaths]()` (line ~2339), which rebuilds the complete modified-path set — Object.keys of activePaths plus parentPaths expansion per entry, O(M x depth) — then does a linear `modified.indexOf(path)` and a nested `paths.some(...directModifiedPaths.some(startsWith))` scan. Nothing is memoized on the document, so K isModified() calls cost O(K x M). Mongoose's own hot path hits this: `_getPathsToValidate` (lib/document.js line ~2965, called from Model.prototype.save in lib/model.js via $__validate) runs `!doc.$__isSelected(path) && !doc.$isModified(path)` in a filter over all required paths without passing a precomputed modifiedPaths, so projected documents with many required paths and many modifications pay O(required x modified) per save. User code that loops isModified over paths in pre-save hooks (a very common pattern) is quadratic the same way.

### Benchmark

With N fields modified, making N isModified() calls on unmodified paths will grow quadratically: N=500 should be roughly 2500x slower than N=10 per pass (each call rebuilds an O(N) modified-path set), clearly matching the O(N^2) prediction rather than O(N).

**Observed complexity:** O(N²)

```
| N=10 | 0.22 | 1.00 | — |
| N=50 | 3.17 | 14.69 | — |
| N=100 | 10.88 | 50.49 | — |
| N=500 | 364.97 | 1693.25 | — |

O(N) predicts 1.0x, O(N²) predicts 1.0x
O(N) predicts 5.0x, O(N²) predicts 25.0x
O(N) predicts 10.0x, O(N²) predicts 100.0x
O(N) predicts 50.0x, O(N²) predicts 2500.0x
```

benchmark confirmed

xianjianlf2:
**Summary**

Fixes #16379.

For a projected document that has modifications, validation of required paths was rebuilding the full modified-paths set once per required path, making required-path validation O(required x modified).

In `_getPathsToValidate()` (`lib/document.js`), the required-path filter calls `doc.$isModified(path)` for every required path. `$isModified` accepts an optional third argument, a precomputed modified-paths set, but the call did not pass one. Without it, each `$isModified` call falls through to `doc.modifiedPaths()` and rebuilds the whole set. For a document with many deselected required paths this repeats the rebuild once per path.

The fix computes the modified-path set at most once and threads it through the existing `$isModified` argument:

```js
let _modifiedPaths;
const getModifiedPaths = () => (_modifiedPaths ??= doc.modifiedPaths());
...
if (!doc.$__isSelected(path) && !doc.$isModified(path, null, getModifiedPaths())) {
  return false;
}
```

The set is built lazily, only when a deselected required path is actually checked, and the same precomputed set is reused for the later subdocument pass (which already called `doc.modifiedPaths()` directly a few lines below — that call now reuses `getModifiedPaths()` too). This mirrors the third-argument usage already present elsewhere in the same function. Behavior is unchanged: the precomputed set is identical to what `$isModified` would otherwise compute internally.

**Examples**

Added a regression test in `test/document.test.js` (gh-16379). It hydrates a projected document with 40 required paths (all but one deselected), makes one modification so the modified-paths set is actually built, then runs `validateSync()` while spying on the internal `modifiedPaths` computation. Before the fix the spy was called once per deselected required path; after the fix it is called at most once.

The test is a targeted, spy-based check that runs under `validateSync()` without a live database. I verified the fix and the regression test locally; the full mocha suite needs a local MongoDB, which was not run here.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2959,9 +2959,13 @@ function _getPathsToValidate(doc, pathsToValidate, pathsToSkip, isNestedValidate
   const doValidateOptions = {};
 
   _evaluateRequiredFunctions(doc);
+  // gh-16379: compute the modified-path set at most once and reuse it below, so
+  // checking many required paths doesn't rebuild it per-path (O(required x modified)).
+  let _modifiedPaths;
+  const getModifiedPaths = () => (_modifiedPaths ??= doc.modifiedPaths());
   // only validate required fields when necessary
   let paths = new Set(Object.keys(doc.$__.activePaths.getStatePaths('require')).filter(function(path) {
-    if (!doc.$__isSelected(path) && !doc.$isModified(path)) {
+    if (!doc.$__isSelected(path) && !doc.$isModified(path, null, getModifiedPaths())) {
       return false;
     }
     if (path.endsWith('.$*')) {
@@ -3040,7 +3044,7 @@ function _getPathsToValidate(doc, pathsToValidate, pathsToSkip, isNestedValidate
         }
       }
     }
-    const modifiedPaths = doc.modifiedPaths();
+    const modifiedPaths = getModifiedPaths();
     for (const subdoc of topLevelSubdocs) {
       if (subdoc.$basePath) {
         const fullPathToSubdoc = subdoc.$__pathRelativeToParent();
--- a/test/document.test.js
+++ b/test/document.test.js
@@ -2466,6 +2466,43 @@ describe('document', function() {
       assert.equal(called, 1);
     });
 
+    it('does not rebuild the modified paths set once per required path when validating (gh-16379)', function() {
+      const symbols = require('../lib/helpers/symbols');
+
+      const N = 40;
+      const def = {};
+      for (let i = 0; i < N; ++i) {
+        def['f' + i] = { type: String, required: true };
+      }
+      const Test = db.model('Test', new Schema(def));
+
+      const obj = { _id: new mongoose.Types.ObjectId() };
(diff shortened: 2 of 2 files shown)


xianjianlf2:
Addressed Copilot's test coverage concern in 0d89691d3 by spying on the public `modifiedPaths()` path and temporarily wiring the internal symbol to the same spy, so both production call sites are counted.\n\nLocal check:\n- `npx mocha --exit ./test/document.test.js --grep "does not rebuild the modified paths set once per required path"`

vkarpov15:
I removed the test case because I think the test case is not sufficiently valuable to keep, but the rest of the code looks good - I tested myself and yes having 500 modified paths causes some meaningful overhead (100ms or so) which we should avoid.

vkarpov15 (approved):

PR #16390, opened 2026-07-16
Optimize array.unique by caching toString() result

vivek180905:
**Summary**

This PR optimizes the `array.unique` method by caching the result of `.toString()`. 

Previously, `toString()` was called twice per unique ObjectId: once during the `Set.has()` check, and again during `Set.add()`. By caching the result, this change eliminates a redundant string allocation per item, reducing memory overhead and improving execution time, especially for large arrays of ObjectIds.

**Examples**

This is an internal performance and memory optimization, so there are no changes to the public API or user-facing behavior. All existing tests pass successfully.


vkarpov15 (approved):
Thanks

PR #16392, opened 2026-07-16
9.8

vkarpov15:
9.8 ready for release minus #16370, which I may push back to 9.9.

### DIFF
--- a/test/document.strictRead.test.js
+++ b/test/document.strictRead.test.js
@@ -1,0 +1,149 @@
+'use strict';
+
+/**
+ * Tests for the `strictRead` schema option.
+ * Closes: https://github.com/Automattic/mongoose/issues/4279
+ *
+ * `strictRead` controls what happens when a document hydrated from MongoDB
+ * contains fields not defined in the schema:
+ *   - false (default): fields are stored in _doc (existing behavior)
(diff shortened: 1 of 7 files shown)


AbdelrahmanHafez (approved):

test/document.strictRead.test.js
```diff
@@ -0,0 +1,216 @@
...
+
+        const readSchema = new Schema({ name: String });
+        const ReadModel = db.model('GlobalStrictReadRead', readSchema, WriteModel.collection.name);
+        const doc = await ReadModel.findOne({ name: 'test' });
+
+        assert.equal(doc.name, 'test');
+        assert.equal(doc._doc.extraField, undefined, 'extraField should be stripped via global strictRead settings');
+      } finally {
+        mongoose.set('strictRead', originalStrictRead);
+      }
```
hasezoey, line 213: I dont think cleanup should be done in the test itself (for example what if it times-out?)

hasezoey (approved):
Looks good, aside from a single test.

test/document.strictRead.test.js
vkarpov15: Will fix in master

PR #16395, opened 2026-07-17
docs: add Atlas Vector Search and Atlas Search documentation

RaschidJFR:
## Summary

Adds comprehensive documentation for **Atlas Vector Search** and **Atlas Search** in Mongoose. These features are fully implemented in the codebase but were previously undocumented.

This PR focuses on Mongoose-specific usage patterns, with references to MongoDB's official documentation for implementation details.

### Atlas Vector Search (`docs/atlas-vector-search.md`)
- Creating vector search indexes with `schema.searchIndex()` and `type: 'vectorSearch'`
- Querying with `$vectorSearch` aggregation stage
- Atlas Automated Embeddings 
- Third-party embedding providers 
- Pre-filtering and search options

### Atlas Search (`docs/atlas-search.md`)
- Creating text search indexes with Lucene analyzers
- Managing indexes (create, list, update, delete)
- Text search queries with `$search` aggregation stage
- Multi-field and fuzzy search
- Compound queries with scoring
- Hybrid search patterns (sequential re-ranking)
- Production best practices

### New Files (2)
- `docs/atlas-vector-search.md` - Complete vector search guide (~260 lines)
- `docs/atlas-search.md` - Complete text search guide (~390 lines)

### Modified Files (3)
- `docs/guide.md` - Enhanced autoSearchIndex section with vector search example
- `docs/layout.pug` - Added sidebar navigation entries
- `docs/source/index.js` - Registered new doc routes

### Navigation
- Added to sidebar: Atlas Search → Vector Search
- Updated `docs/guide.md` autoSearchIndex section


docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+    mappings: {
+      dynamic: true  // Automatically index all fields
+    }
+  }
+});
+
+const Article = mongoose.model('Article', articleSchema);
+
+// Index will be created when model initializes
+await Article.init();
```
vkarpov15: Mongoose runs `init` automatically when `mongoose.model()` is called so it isn't necessary to call `Article.init()` unless you want to wait for `init` to complete.

It is also worthwhile to call out the `Model.createSearchIndexes()` function, which users can use to explicitly create all search indexes on a particular model without having to enable `autoSearchIndex`.

docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+      }
+    }
+  }
+});
+```
+
+### Vector Search Index
+
+For semantic search using vector embeddings, create a vector search index.
+See the dedicated [Vector Search](atlas-vector-search.html) guide for detailed information.
```
vkarpov15: We prefer to use absolute links e.g. `https://mongoosejs.com/docs/atlas-vector-search.html` instead of relative links to make the links usable when reading the markdown file directly.

docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+      }
+    }
+  },
+  { $limit: 10 }
+]);
+```
+
+### Compound Queries
+
+Combine multiple search criteria with `must`, `should`, and `filter` clauses. Include relevance scores using the `$meta` operator:
```
vkarpov15: It would be helpful to explain what this query does, e.g. "find 5 most relevant articles with 'mongodb' in the content since 2023, prioritizing articles where the title includes 'mongodb'"

docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+```
+
+For production use, see the [Atlas Hybrid Search documentation](https://www.mongodb.com/docs/atlas/atlas-search/tutorial/hybrid-search/) for optimized patterns and best practices.
+
+## Best Practices
+
+### Index Management
+
+* **Use `autoSearchIndex: true` in development**: Automatically sync indexes with your schema
+* **Manage indexes manually in production**: Create and update indexes through Atlas UI, MongoDB CLI, or deployment scripts to avoid unintended changes during application deployments
```
vkarpov15: Good to add a note about `Model.createSearchIndexes()` here

docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+      title: 1,
+      content: 1,
+      // The $meta: 'searchScore' operator returns the relevance score
+      // calculated by Atlas Search based on how well the document matches the query
+      score: { $meta: 'searchScore' }
+    }
+  },
+  {
+    $match: {
+      score: { $gte: 5 }  // Only return high-quality matches
```
vkarpov15: It would be helpful to give some rationale for this number with a link to details on how scoring is calculated. If the user creates some basic placeholder data it is likely that no documents will match `score: { $gte: 5 }`.

docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+```javascript
+// Good: Define indexes in schema for version control
+articleSchema.searchIndex({
+  name: 'article_search',
+  definition: { mappings: { dynamic: false, fields: { /* ... */ } } }
+});
+
+// Also good: Separate index management for production
+const createProductionIndexes = async () => {
+  const indexes = await Article.listSearchIndexes();
```
vkarpov15: Is it ok to just use `Article.createSearchIndexes()` here? It seems like `createSearchIndexes()` doesn't throw an error if creating search indexes that already exist.

docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+      type: 'vector',
+      path: 'plot_embedding',
+      numDimensions: 1536,
+      similarity: 'cosine'
+    }]
+  }
+});
+```
+
+## Managing Search Indexes {#managing-indexes}
```
vkarpov15: I'd prefer to use the default generated slug where possible instead of explicitly defining the header as `{#managing-indexes}`. We only explicitly define slugs in cases where backwards compatibility is a concern.

docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+    $match: {
+      score: { $gte: 5 }  // Only return high-quality matches
+    }
+  }
+]);
+```
+
+## Vector Search {#vector-search}
+
+For semantic search using vector embeddings, use the `$vectorSearch` stage.
```
vkarpov15, line 231: This is the second call-out to visit the separate vector search page - I don't think this section adds much over the other section on "Vector Search Index" and breaks the flow a bit. We could simply have the subsequent hybrid search section link out to vector search.

docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+});
+```
+
+### Vector Search Index
+
+For semantic search using vector embeddings, create a vector search index.
+See the dedicated [Vector Search](atlas-vector-search.html) guide for detailed information.
+
+```javascript
+const movieSchema = new mongoose.Schema({
```
vkarpov15: It would be better if this example used the `article` model used everywhere else in this file, rather than introducing a new model.

docs/atlas-search.md
```diff
@@ -0,0 +1,388 @@
...
+    }
+  },
+  {
+    $addFields: {
+      vectorScore: { $meta: 'vectorSearchScore' }
+    }
+  },
+  // 2. Re-rank using text search
+  {
+    $search: {
```
vkarpov15: Running this aggregation gives me the following error, am I missing something here? Looks like should use `$rankFusion`

```
/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/cmap/connection.js:320
                    throw new error_1.MongoServerError((object ??= document.toObject(bsonOptions)));
                          ^

MongoServerError: $search is only valid as the first stage in a pipeline
    at Connection.sendCommand (/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/cmap/connection.js:320:27)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Connection.command (/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/cmap/connection.js:347:26)
    at async Server.command (/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/sdam/server.js:208:29)
    at async executeOperationWithRetries (/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/operations/execute_operation.js:193:32)
    at async executeOperation (/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/operations/execute_operation.js:83:16)
    at async AggregationCursor._initialize (/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/cursor/aggregation_cursor.js:57:26)
    at async AggregationCursor.cursorInit (/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/cursor/abstract_cursor.js:620:27)
    at async AggregationCursor.fetchBatch (/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/cursor/abstract_cursor.js:654:13)
    at async AggregationCursor.next (/home/v/Desktop/MongoDB/mongoose/node_modules/mongodb/lib/cursor/abstract_cursor.js:326:17) {
  errorLabelSet: Set(0) {},
  errorResponse: {
    ok: 0,
    errmsg: '$search is only valid as the first stage in a pipeline',
    code: 40602,
    codeName: 'Location40602',
    '$clusterTime': {
      clusterTime: Timestamp { high: 1784573406, low: 55, unsigned: true },
      signature: {
        hash: Binary {
          buffer: Buffer(20) [Uint8Array] [
             16,   3, 193, 64, 244,  54,
            194,  56,  21, 79,  62, 193,
            104, 227, 226, 86, 226, 242,
             16,  66
          ],
          sub_type: 0,
          position: 20
        },
        keyId: Long { high: 1769626381, low: 88, unsigned: false }
      }
    },
    operationTime: Timestamp { high: 1784573406, low: 55, unsigned: true }
  },
  ok: 0,
  code: 40602,
  codeName: 'Location40602',
  '$clusterTime': {
    clusterTime: Timestamp { high: 1784573406, low: 55, unsigned: true },
    signature: {
      hash: Binary {
        buffer: Buffer(20) [Uint8Array] [
           16,   3, 193, 64, 244,  54,
          194,  56,  21, 79,  62, 193,
          104, 227, 226, 86, 226, 242,
           16,  66
        ],
        sub_type: 0,
        position: 20
      },
      keyId: Long { high: 1769626381, low: 88, unsigned: false }
    }
  },
  operationTime: Timestamp { high: 1784573406, low: 55, unsigned: true }
}

```

docs/atlas-vector-search.md
```diff
@@ -0,0 +1,278 @@
...
+You can do this by calling `schema.searchIndex()` with `type: 'vectorSearch'` and defining the vector fields you want to index.
+
+```javascript
+const movieSchema = new mongoose.Schema({
+  title: String,
+  plot: String,
+  // Vector embeddings for the plot
+  // Typically 1536 dimensions for OpenAI ada-002, 768 for sentence-transformers,
+  // or 1024 for Voyage AI models
+  plot_embedding: {
```
vkarpov15: Mongoose property names are typically camelCase, but this is fine since I presume this guide is using MongoDB's sample_mflix sample data set. It would be good to link out to that so users know they can get this data set.

docs/atlas-vector-search.md
```diff
@@ -0,0 +1,278 @@
...
+    type: [Number],
+    required: true
```
vkarpov15: For embeddings we also recommend

1: Add `default: undefined` because otherwise Mongoose will store an empty array by default, with embeddings you typically either want a full embedding or nothing as opposed to an empty array
2. Add a validator to validate the length of the array. Since the search index below uses 1536, add a validator to validate that the array is nullish or length 1536

docs/atlas-vector-search.md
```diff
@@ -0,0 +1,278 @@
...
+If you need to generate embeddings in your application (for example, when using embedding models not supported by Atlas Automated Embeddings), you can use any embedding provider. Popular options include:
+
+* **[Voyage AI](https://www.mongodb.com/products/platform/ai-search-and-retrieval)** - High-quality embeddings optimized for retrieval, with models like `voyage-2` and `voyage-code-2`
+* **[OpenAI Embeddings](https://platform.openai.com/docs/guides/embeddings)** - Models like `text-embedding-ada-002` (1536 dimensions) and `text-embedding-3-small`
+* **[Anthropic Embeddings](https://platform.claude.com/docs/en/build-with-claude/embeddings)** - Multiple models, including Voyage AI family.
+
+Example workflow for application-generated embeddings:
+
+```javascript
+import OpenAI from "openai";
```
vkarpov15: 1. Mongoose docs currently use CJS not ESM
2. Let's avoid using the OpenAI SDK for compatibility reasons - better to use `fetch` to make a request to the OpenAI API or some other embedding API directly here. I typically use [Nomic](https://www.nomic.ai/) for embeddings.

docs/atlas-vector-search.md
```diff
@@ -0,0 +1,278 @@
...
+
+```javascript
+const movieSchema = new mongoose.Schema({
+  title: String,
+  plot: String,
+  // Vector embeddings for the plot
+  // Typically 1536 dimensions for OpenAI ada-002, 768 for sentence-transformers,
+  // or 1024 for Voyage AI models
+  plot_embedding: {
+    type: [Number],
```
vkarpov15: In the sample_mflix data set it looks like `plot_embedding` is a Buffer not an array of numbers, which causes Mongoose to record a CastError when loading this property

docs/atlas-vector-search.md
```diff
@@ -0,0 +1,278 @@
...
+## Generating Embeddings {#embeddings}
+
+Vector embeddings are numerical representations of your data generated by embedding models.
+Mongoose doesn't generate embeddings for you - you need to generate them before saving documents or performing queries.
+
+### Using Atlas Automated Embeddings (Recommended)
+
+**MongoDB Atlas can automatically generate embeddings for your documents** using [Atlas Vector Search Automated Embeddings](https://www.mongodb.com/docs/vector-search/crud-embeddings/automated-embedding/overview/). Available in MongoDB Atlas and self-managed deployments (requires MongoDB 8.2+).
+
+With Automated Embeddings, you define an `autoEmbed` index on a text field — no embedding field in your schema required. Atlas generates and manages the embeddings internally.
```
vkarpov15, line 113: How would the `generateEmbeddings()` function look if someone were using Atlas Automated Embeddings - is there a way to calculate the embedding so a user can search by auto embedded documents?

vkarpov15 (changes requested):
I walked through these tutorials and have a lot of comments but mostly minor. Mostly I just want to make sure I can fully test the code samples end to end in both cases, and in both cases I wasn't quite able to make it all the way through. The `$rankFusion` issue for atlas-search.md and not having a way to calculate auto embeddings for atlas-vector-search.md are the blockers right now.

docs/atlas-search.md
RaschidJFR: Removed `Article.init()`.
Updated the prose to mention `createSearchIndexes`.

docs/atlas-search.md
RaschidJFR: Updated all links in both .md files to absolute path

docs/atlas-search.md
RaschidJFR: updated

docs/atlas-search.md
RaschidJFR: updated in both files

docs/atlas-search.md
RaschidJFR: Added inline comments

docs/atlas-search.md
RaschidJFR: Added more context in the prose before the example

docs/atlas-search.md
RaschidJFR: Agree. I removed the Vector Search Indexes section since this page focuses on Atlas Search; and here I just left the link to the Vector Search page rather than duplicating another example.

docs/atlas-search.md
RaschidJFR: Included

docs/atlas-search.md
RaschidJFR: According to the [docs](https://www.mongodb.com/docs/manual/reference/method/db.collection.createSearchIndex/#command-fields), it does throw

docs/atlas-vector-search.md
RaschidJFR: ~added in the prose~
UPDATE: Given m_flix's embeddings are stored as binary data rather than an array of numbers, a working example for it would be too complex for the purpose of this guide. There is no suitable sample dataset available from mongodb for a simple, working example.
Will omit the reference/link for now.

docs/atlas-vector-search.md
RaschidJFR: You are right. For now, I will add a note about this to avoid changing all the examples in this page.

docs/atlas-vector-search.md
RaschidJFR: added

docs/atlas-vector-search.md
RaschidJFR: Expanded the example to show how it is queried

docs/atlas-vector-search.md
RaschidJFR: updated example to use Voyage's REST endpoint

docs/atlas-search.md
RaschidJFR: fixed and verified

docs/atlas-search.md
```diff
@@ -0,0 +1,322 @@
...
+
+```javascript
+const articleSchema = new mongoose.Schema({
+  title: String,
+  content: String,
+  author: String,
+  tags: [String],
+  publishedAt: Date
+}, {
+  autoSearchIndex: true  // Automatically create search indexes
```
vkarpov15: Given that this document references `createSearchIndexes()` as the preferred way to create search indexes, I would advise not setting `autoSearchIndex: true` here - just omit `autoSearchIndex` or put it in a separate section. People will copy/paste this code snippet along with `createSearchIndexes()` and that will lead to them double-creating search indexes. `createSearchIndexes()` doesn't require `autoSearchIndex` option to be set.

docs/atlas-search.md
```diff
@@ -0,0 +1,322 @@
...
+  {
+    $project: {
+      title: 1,
+      content: 1,
+      score: { $meta: 'searchScore' }  // Include the relevance score in the results
+    }
+  },
+  {
+    $match: {
+      score: { $gte: 5 }  // Adjust this threshold based on your data
```
vkarpov15: In my brief experiments `5` is very high. For example, with query 'mongodb' still gives a score of 1.97 for an article with title 'MongoDB Tutorial':

```
  {
    _id: new ObjectId('000000000000000000000001'),
    content: 'MongoDB is a document database designed for developer productivity.',
    title: 'MongoDB Tutorial',
    score: 1.973141074180603
  }
```

I'd recommend making this something like `$gte: 0.1` or some other low value to make it reasonably likely the user will get a result when executing a query.

It would also be helpful to provide some sample data too

docs/atlas-search.md
```diff
@@ -0,0 +1,322 @@
...
+]);
+```
+
+For more details, see the [Atlas Hybrid Search documentation](https://www.mongodb.com/docs/vector-search/hybrid-search/hybrid-search/).
+
+## Best Practices
+
+### Index Management
+
+* **Use `autoSearchIndex: true` in development**: Automatically sync indexes with your schema
```
vkarpov15: Minor pedantic point - "Automatically sync indexes" -> "Automatically create indexes from your schema"

In Mongoose, syncing indexes e.g. `syncIndexes()` drops indexes which are not in the schema. Creating indexes just creates new indexes without dropping existing indexes.

docs/atlas-search.md
vkarpov15: I don't see anything in the referenced docs about throwing an error if the index already exists. Also in my experiments, I'm calling `Article.createSearchIndexes()` every time I run a script - no errors.

docs/atlas-search.md
```diff
@@ -0,0 +1,322 @@
...
+See the complete [Vector Search](https://mongoosejs.com/docs/atlas-vector-search.html) guide for detailed examples.
+
+## Hybrid Search
+
+Combine text search and vector search to leverage both keyword relevance and semantic similarity.
+
+Use `$rankFusion` to run `$vectorSearch` and `$search` as separate subpipelines and merge their results using [Reciprocal Rank Fusion (RRF)](https://www.mongodb.com/docs/vector-search/hybrid-search/hybrid-search/). Note that `$search` must be the first stage in its subpipeline, which is why it cannot be used directly after `$vectorSearch` in the same pipeline.
+
+```javascript
+const queryEmbedding = await generateEmbedding('machine learning tutorial');
```
vkarpov15: Given that other examples in this document use "mongodb tutorial" as an example, I'd prefer to use those semantics for this example so the user doesn't need to create a new data set specifically for this example.

docs/atlas-search.md
```diff
@@ -0,0 +1,322 @@
...
+  if (!indexes.find(idx => idx.name === 'article_search')) {
+    await Article.createSearchIndex({ /* definition */ });
+  }
+};
+```
+
+### Query Optimization
+
+* **Use `$limit` early**: Reduce the number of documents passed to subsequent pipeline stages
+* **Filter before searching**: Use `$match` before `$search` when possible to reduce search scope
```
vkarpov15: This suggestion is incorrect AFAIK, would throw a `$search is only valid as the first stage in a pipeline` error. That's why `filter` is a property in `$search`. I recommend instead putting that `$search` should be the first stage in the pipeline so users don't get errors by surprise.

docs/atlas-search.md
```diff
@@ -0,0 +1,322 @@
...
+* **Project only needed fields**: Use `$project` to return only necessary data
+* **Index the right fields**: Don't use `dynamic: true` in production; explicitly index only the fields you search
```
vkarpov15: Why is `dynamic: true` bad in production - slows down writes?

docs/atlas-vector-search.md
vkarpov15: I agree, using `plot_embedding` is fine since we're using sample_mflix. However, in that case we should have a link to documentation or some information about sample_mflix so users know this is a data set they can import into Atlas directly. I think this page: https://www.mongodb.com/docs/manual/sample-data/sample-mflix/#std-label-mflix-embedded_movies would be a good one to link to.

docs/atlas-vector-search.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+## Querying with $vectorSearch
+
+Once your vector search index is created, and your embeddings are generated (see [Generating Embeddings](#generating-embeddings)), you can perform vector similarity searches using the `$vectorSearch` aggregation stage.
+
+```javascript
+// Your query embedding (example function - see "Generating Embeddings" section below)
+// MongoDB Atlas can generate embeddings automatically, or you can use third-party providers
+const queryEmbedding = await generateEmbedding('romantic comedy about friendship');
```
vkarpov15: I would prefer we include an implementation of `generateEmbedding()` that works with sample_mflix to make the code in this document immediately usable, rather than leaving significant implementation details as an exercise to the reader.

I was able to get the existing sample_mflix data searchable using the following:

```javascript
const movieSchema = new mongoose.Schema({
  title: String,
  plot: String,
  // Vector embeddings for the plot
  // The stored vectors are 2048-dimensional float32 BSON vectors.
  plot_embedding_voyage_3_large: {
    type: Buffer,  // use Buffer if your embeddings are stored as binData
    validate: {
      validator: v => v == null || v.length === 8192,  // 2048 float32 values, 4 bytes each
      message: 'plot_embedding_voyage_3_large must be nullish or exactly 8192 bytes'
    }
  }
}, { });

// Define a vector search index
movieSchema.searchIndex({
  name: 'vector_index',
  type: 'vectorSearch',
  definition: {
    fields: [{
      type: 'vector',
      path: 'plot_embedding_voyage_3_large',
      numDimensions: 2048,  // Must match your embedding model's dimensions
      similarity: 'cosine'  // or 'euclidean' or 'dotProduct'
    }]
  }
});

const Movie = mongoose.model('Movie', movieSchema, 'embedded_movies');

async function generateEmbedding(input) {
  const voyageAPIKey = process.env.VOYAGE_API_KEY;
  if (!voyageAPIKey) {
    throw new Error('VOYAGE_API_KEY environment variable is required');
  }

  const response = await fetch('https://ai.mongodb.com/v1/embeddings', {
    method: 'POST',
    headers: {
      Authorization: `Bearer ${voyageAPIKey}`,
      'Content-Type': 'application/json'
    },
    body: JSON.stringify({
      input,
      model: 'voyage-3-large',
      output_dimension: 2048
    })
  });

  if (!response.ok) {
    throw new Error(`Voyage AI request failed: ${response.status} ${await response.text()}`);
  }

  const result = await response.json();
  const embeddings = result.data.map(({ embedding }) => embedding);
  return Array.isArray(input) ? embeddings : embeddings[0];
}

void async function main() {
  await Movie.createSearchIndexes();

  // Your query embedding (example function - see "Generating Embeddings" section below)
  // MongoDB Atlas can generate embeddings automatically, or you can use third-party providers
  const queryEmbedding = await generateEmbedding('vacation to mars');
  const queryVector = Binary.fromFloat32Array(Float32Array.from(queryEmbedding));

  // Perform vector search
  const results = await Movie.aggregate([
    {
      $vectorSearch: {
        index: 'vector_index',                  // Name of your vector search index
        path: 'plot_embedding_voyage_3_large',  // Field containing the vectors
        queryVector,                            // Match the stored float32 BSON vector representation
        numCandidates: 100,                     // Number of candidates to consider (should be >= limit)
        limit: 10                               // Number of results to return
      }
    },
    {
      $project: {
        title: 1,
        plot: 1,
        score: { $meta: 'vectorSearchScore' }   // Include similarity score
      }
    }
  ]);
  console.log(results[0].title); // "Total Recall"
  await mongoose.disconnect();
}();

```

Using [Voyage AI API as described here](https://www.mongodb.com/docs/voyageai/api-and-clients/?client-curl-default=curl) and [Voyage AI API key setup as described here](https://www.mongodb.com/docs/voyageai/management/api-keys/#std-label-voyage-api-keys)

vkarpov15 (changes requested):
Some minor comments on the search indexes docs, those look mostly good.

For vector search, I would like that document to be more immediately usable. If we make it clear that document is meant to integrate with the `embedded_movies` collection in `sample_mflix` (https://www.mongodb.com/docs/manual/sample-data/sample-mflix/#std-label-mflix-embedded_movies), we can make it so that users can immediately use semantic search on the movies database with only one additional step of setting up a Voyage AI API key. 

An alternative path would be to create a new schema and minimal data set from scratch - that approach has the benefit of not requiring working with buffers, which may be a bit confusing (although absolutely makes sense - buffers are much more space efficient than arrays of numbers for storing vectors and vectors are _huge_ relative to your average name, age, etc. properties). However, I think there's a benefit to being immediately able to search a non-trivial data set as a way to demonstrate the power of semantic search.

docs/atlas-search.md
RaschidJFR: updated

docs/atlas-search.md
RaschidJFR: Removed from code example.
Added a link to https://mongoosejs.com/docs/guide.html#autoSearchIndex for more info.

docs/atlas-search.md
RaschidJFR: Thanks for catching this. Fixed

docs/atlas-search.md
RaschidJFR: Because with `dynamic:true` every field gets indexed, making the search index potentially larger than the raw data if not used carefully. 

Added a not about this

Source: https://www.mongodb.com/docs/search/performance/index-performance/?deployment-type=atlas#dynamic-vs-static-mappings

docs/atlas-vector-search.md
RaschidJFR: Added a reference to mflix and updated all examples to be compatible with its Voyage version (using `plot_embedding_voyage_3_large` instead of `plot_embedding`)

docs/atlas-search.md
RaschidJFR: Reworked the guide to use mflix examples

docs/atlas-search.md
RaschidJFR: With this example agains the mflix dataset, the minimum score is `1.27`. Nontheless, I lowered to `0.1` to back the idea of starting low.

docs/atlas-vector-search.md
RaschidJFR: Added function definition

RaschidJFR (commented):
Reworked both guides to share the same semantics and be compatible with mflix dataset

docs/atlas-search.md
RaschidJFR: You are right, it doesn't throw as long as the definition is the same. Updated!

vkarpov15 (approved):
Reviewed and tested thoroughly, I made a couple of minor tweaks but looks good :+1:

Issue #11861, opened 2022-05-31
Make `strictQuery` `false` by default again


vkarpov15:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

6.x

### Node.js version

Any

### MongoDB server version

Any

### Description

The `strictQuery` change in Mongoose 6 (https://mongoosejs.com/docs/migrating_to_6.html#strictquery-is-removed-and-replaced-by-strict) has caused a lot of confusion. Here's a couple:

https://github.com/Automattic/mongoose/issues/11609
https://github.com/Automattic/mongoose/issues/11807
https://github.com/Automattic/mongoose/issues/11514

Etc.

### Steps to Reproduce

From #11609: 

```javascript
const mongoose = require('mongoose');

const testSchema = new mongoose.Schema({
    title: [{
        value: String,
        language: String
    }]
});

const Test = mongoose.model('Test', testSchema);

async function run() {
    await mongoose.connect('mongodb://localhost:27017');
    await mongoose.connection.dropDatabase();

    await Test.create({
        title: [{
            value: 'El Dave valencia',
            language: 'ca'
        }, {
            value: 'El Dave castellano',
            language: 'es'
        }]
    });

    await Test.create({
        title: [{
            value: 'test',
            language: 'ca'
        }, {
            value: 'test',
            language: 'es'
        }]
    });

    const query = await Test.find({$or: [{"title.value":/Dave/i},{"name.value":/Dave/i} ]});
    console.log('done', query, query[0].title, query[1].title);
    
}

run();
```

### Expected Behavior

Make `strictQuery` false by default in Mongoose 7

hasezoey:
as mentioned in https://github.com/Automattic/mongoose/issues/12389#issuecomment-1252553337, what about setting it to `throw` (or return some kind of validation errors) instead of reverting to `false` instead of the current "silently remove things" or going back to include "as is"?

vkarpov15:
@hasezoey 'strictQuery: throw' by default would be a bit too strict to be a default for Mongoose. Wouldn't fit in well with the rest of our defaults.

vkarpov15:
We should also decouple `strict` and `strictQuery` again. Setting `strictQuery` to `strict` by default was a workaround, setting `strict: false` in your schema shouldn't disable `strictQuery`.

ilovepixelart:
[...]

Issue #6032, opened 2018-01-23
inconsistency in count()


TrejGun:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
`count()` is not removing field which are not present in model

```js
const model = Schecma({
field: String
})

model.count({field:"abc", imnotpresentinthemodel:1})
// -> collection.count({field:"abc"}, {})


```

**What is the expected behavior?**
`count()` should remove fields which are not present in model

**Please mention your node.js, mongoose and MongoDB version.**
node 8.9
mongoose 5.0.0

varunjayaraman:
I think what @TrejGun is getting at here is that mongoose should strip fields from a query that dont exist in the schema. I'm not sure if this is by design or not, but right now mongoose will send extra fields to the query to mongodb:

```javascript
const mongoose = require('mongoose');
const assert = require('assert');
mongoose.Promise = global.Promise;

mongoose.set('debug', true);

const GITHUB_ISSUE = `gh-6032`;

run().catch(error => console.error(error.stack));

async function run() {
  await mongoose.connect(`mongodb://localhost:27017/${ GITHUB_ISSUE }`);

  const schema = new mongoose.Schema({ name: String });

  const Model = mongoose.model('Model', schema);

  await Model.remove({});

  await Model.create({ name: 'test' });
  const count = await Model.count({ name: 'test', nonExistentField: 'asdf' });
  assert.equal(count, 1);
}
```

vkarpov15:
This has been an issue for quite some time. We added a `strictQuery` option to tune strict mode for query filters in 4.10.0 with #4136 but didn't do a good job documenting it. In 5.0.4 you'll be able to set `strictQuery: true` on individual queries or on your schema to make strict mode apply to query filters. Thanks for reporting :+1: :palm_tree:

PR #16397, opened 2026-07-18
docs(guide): clarify what strictQuery does to filter paths not in the schema

MuhammadFarhantahir:
With `strictQuery: false` (the default since 7.0), a filter path that isn't in the schema is passed through uncasted, so the query usually matches nothing. With `strictQuery: true` the path is silently stripped, so the query can match more than intended. The guide shows both examples but never spells out this difference — it's an easy silent bug to hit (we did, via an `updateMany` filter that quietly matched 0 docs).

Also fixes the first example in the section, which still says `strict: true` filters unknown paths out of query filters — that hasn't been true since strictQuery stopped inheriting from strict in 7.0 (#11861).

Behavior verified against 8.24.1 and 9.7.4, and matches the existing gh-11861 / gh-6032 tests in test/query.test.js. `npm run lint-md` passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -949,9 +949,9 @@ Mongoose supports a separate `strictQuery` option to avoid strict mode for query
 This is because empty query filters cause Mongoose to return all documents in the model, which can cause issues.
 
 ```javascript
-const mySchema = new Schema({ field: Number }, { strict: true });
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thank you :+1:

PR #16398, opened 2026-07-20
Fix/16383 optimistic concurrency wildcards

### LINKED ISSUE #16383: `optimisticConcurrency` paths don't match map wildcards or subdocument array paths
re #16054 / #16177 / #16367.

Splitting this out of #16367 as discussed with @vkarpov15. Part of what I intended with #16054 is still not covered, there are two cases where `optimisticConcurrency` include/exclude paths never match the runtime modified paths:

1. Map wildcards: `['settings.$*']` matches nothing, in both the include and exclude forms.
2. Arrays of subdocuments: `optimisticConcurrency: ['comments.text']` never triggers versioning, runtime paths look like `comments.0.text` and the matcher doesn't skip index segments. So the only way to protect a subdocument field right now is to include the whole array.

```js
  settings: { type: Map, of: String },
  balance: Number
}, { optimisticConcurrency: ['settings.$*'] });

const user = await User.create({ settings: { theme: 'dark' }, balance: 100 });

user.settings.set('theme', 'light');
user.$__delta();

// user.$__.version === undefined, versioning never triggers
// the modified path is `settings.theme`, which never matches `settings.$*`
```

```js
const postSchema = new Schema({
  comments: [{ text: String }],
  title: String
}, { optimisticConcurrency: ['comments.text'] });

const Post = mongoose.model('Post', postSchema);
const post = await Post.create({ title: 'Hello', comments: [{ text: 'First' }] });

post.comments[0].text = 'Edited';
post.$__delta();

// post.$__.version === undefined, versioning never triggers
// the modified path is `comments.0.text`, which never matches `comments.text`
```

The matching in `$__delta()` (`_pathOverlapsSet()` / `_pathOrAncestorInSet()`) compares dot segments literally, so `$*` and numeric array indexes never line up with what `directModifiedPaths()` returns.

I'd handle this by extracting the `pathUtils` helper we discussed on https://github.com/Automattic/mongoose/pull/16177#discussion_r2994719194, something that skips numeric segments and handles `.$*.`, and using it for both the include and exclude branches.

The open question is whether we should bear the maintenance cost now, or not until someone actually asks for it. What do you think? @vkarpov15 @hasezoey

vivek180905:
fix(document): support map wildcards and array paths in optimisticConcurrency

Fixes #16383 

This PR addresses the issue where `optimisticConcurrency` fails to match paths with map wildcards (`$*`) or subdocument arrays (e.g., `comments.text` vs modified `comments.0.text`).

**Changes made:**
- Introduced a recursive `_checkOverlap` helper in `lib/document.js`.
- It dynamically skips numeric array indices in the modified path unless the schema explicitly expects a numeric segment, and it handles `$*` map wildcards effectively.
- Refactored both `_pathOrAncestorInSet` and `_pathOverlapsSet` to use this new helper.
- Added tests for both map wildcards and array of subdocuments in `test/versioning.test.js`.

All 4,400+ Mongoose tests continue to pass seamlessly with these updates.

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -5673,21 +5673,47 @@ Document.prototype._applyVersionIncrement = function _applyVersionIncrement() {
  */
 
 /*!
+ * Helper for _pathOrAncestorInSet and _pathOverlapsSet
+ */
+function _checkOverlap(modParts, m, schemaParts, s) {
+  if (s === schemaParts.length) {
+    return { ancestor: true, overlap: true };
+  }
+  if (m === modParts.length) {
+    return { ancestor: false, overlap: true };
+  }
+
+  if (modParts[m] === schemaParts[s] || schemaParts[s] === '$*') {
+    const res = _checkOverlap(modParts, m + 1, schemaParts, s + 1);
+    if (res.overlap) {
+      return res;
+    }
+  }
+
+  if (/^\d+$/.test(modParts[m]) && !/^\d+$/.test(schemaParts[s])) {
+    const res = _checkOverlap(modParts, m + 1, schemaParts, s);
+    if (res.overlap) {
+      return res;
+    }
+  }
+
+  return { ancestor: false, overlap: false };
+}
+
+/*!
  * Check if `path` or any of its ancestor paths exist in `pathSet`.
  * For example:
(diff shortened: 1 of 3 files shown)


vkarpov15:
Closing in favor of #16436

Issue #3199, opened 2015-07-24
Required validation through API not working


andresmatasuarez:
Hi guys!

I have the following schema with its validations (reduced for more clarity):

```
var MySchema = new Schema({
  ...
  evaluations: [{
    choices: [{
      claim: String
      isCorrect: Boolean
    }]
  }]
});

MySchema.path('evaluations').schema.path('choices').schema.path('claim').required(true, errorMessageRequired);

MySchema.path('evaluations').schema.path('choices').schema.path('claim').validate(maxLengthValidator, errorMessageMaxLength);
```

The max length validator works perfectly, but the required validator is not taken into account while creating a new MySchema document without evaluations.choices.claim.

HOWEVER, if I set the `required` options in the schema definition, it IS taken into account and the validation goes smoothly:
```
[{
  claim: { type: String, required: true }
}]
```

What do you think? I can give a look at the code, and maybe fix it in a PR, but I wanted to find out what your opinion was before doing anything.

Anyways, I really don't think this is expected behaviour.

vkarpov15:
Brief aside, why are you doing `MySchema.path('evaluations').schema.path('choices').schema.path('claim').required(true, errorMessageRequired);` instead of `MySchema.path('evaluations.choices.claim').required()`?

Also, please clarify what you mean by 'the required validator is not taken into account'.

andresmatasuarez:
First of all, if I specify schema path with dot notation, I get "TypeError: Cannot call method 'required' of undefined" OR "TypeError: Cannot call method 'validate' of undefined". As I was struggling with this very issue, I came across this article dealing with the same problem, so at least I am not the only one having to access schema paths for embedded arrays that way.

Secondly, by "The required validator is not taken into account" I mean that it is not working. If I create a MySchema document with some "evaluation"s and one of them lacks a "claim", it will not be prevented from saving. However, if a move the .required() call to required: true in schema definition, document creation will fail with the expected ValidationError.

andresmatasuarez:
Great!

PR #16399, opened 2026-07-21
perf(document): reuse cached required paths on construction

### LINKED ISSUE #16377: Document constructor rescans every schema path via requiredPaths(true) on each instantiation — O(schemaPaths) per doc, O(docs x paths) per query
## Document constructor rescans every schema path via requiredPaths(true) on each instantiation — O(schemaPaths) per doc, O(docs x paths) per query

**Location:** lib/document.js:147 (Document constructor) -> lib/schema.js Schema.prototype.requiredPaths (cache bypassed via invalidate=true)
**Severity:** high · **Confidence:** 0.9
**Type:** redundant-work

### Description

The Document constructor in lib/document.js (line ~147) calls `schema.requiredPaths(true)`. Schema.prototype.requiredPaths in lib/schema.js caches its result in `this._requiredpaths`, but the `invalidate=true` argument forces it to bypass the cache and re-iterate Object.keys(schema.paths), checking `isRequired` on every path — on EVERY document construction. This is cross-module redundant work: query hydration in lib/query.js `Query.prototype._completeMany` calls `helpers.createModel` per result row, which runs this constructor, so a find() returning D documents against a schema with P paths does O(D x P) work recomputing a list that is immutable once the model is compiled. The same cost hits `new Model()`, Model.hydrate, insertMany's `new Model(toSave)` in lib/model.js, and every embedded subdocument instantiation.

### Benchmark

Constructing a fixed number of documents with a constant 1-field payload will scale roughly linearly with the number of schema paths (N=500 should be ~10-40x slower than N=10 despite identical per-document data), because requiredPaths(true) plus other per-construction schema walks are O(schemaPaths) per doc.

**Observed complexity:** O(N)

```
| N=10 | 2.83 | 1.00 | — |
| N=50 | 13.88 | 4.90 | — |
| N=100 | 24.74 | 8.74 | — |
| N=500 | 127.17 | 44.92 | — |
```

benchmark confirmed

BIGSUS24:
Fixes #16377.

The `Document` constructor calls `schema.requiredPaths(true)` for every document, and the `true` makes `requiredPaths` skip its own `_requiredpaths` cache and rescan all paths. So constructing N documents means N full scans of the schema's path list.

The `invalidate` flag goes back to #3199: `.required()` can set `isRequired` after the cache is already warm, and nothing cleared `_requiredpaths` at that point, so forcing a recompute on construction was the workaround.

This clears the cache where the required paths can actually change instead, and lets the constructor read the cache:

- `SchemaType#required()` when it flips `isRequired`
- `Schema#remove()`

Discriminators already `delete schema._requiredpaths` on registration, and `Schema#add`/`#path` go through `required()`, so those cases stay covered.

Added a test for the ordering the suite didn't have: construct a doc (this warms the cache) → `path.required(true)` → construct another doc → the newly required path is still enforced. Reverting only the `required()` invalidation makes that test fail, so it guards the #3199 behavior.

Quick benchmark, 100k docs of a 40-path schema (20 required):

- before: ~22.1 µs/doc
- after: ~19.3 µs/doc (~13% of construction time)

The saving grows with the number of paths.

One thing worth a look: the invalidation uses `this.parentSchema`, set at schematype creation. If a schematype were shared across schemas, nulling could touch the wrong schema's cache — but that's only a wasted recompute, never a stale result. I didn't hit any such case across the schema/document/validation suites (all green).

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -143,7 +143,7 @@ function Document(obj, fields, options) {
     this.$__.strictMode = schema.options.strict;
   }
 
-  const requiredPaths = schema.requiredPaths(true);
+  const requiredPaths = schema.requiredPaths();
   for (const path of requiredPaths) {
     this.$__.activePaths.require(path);
   }


lib/schemaType.js
```diff
@@ -1122,6 +1122,9 @@ SchemaType.prototype.required = function(required, message) {
 
     this.isRequired = false;
     delete this.originalRequiredValue;
+    if (this.parentSchema != null) {
```
vkarpov15: parentSchema unfortunately isn't 100% reliable. For example, consider the following case:

```javascript
'use strict';

const firstSchema = new mongoose.Schema({ name: String });
const reusedSchemaType = firstSchema.path('name');

// Reuse the same SchemaType instance at the same path in another schema.
const secondSchema = new mongoose.Schema({ name: reusedSchemaType });

assert.strictEqual(secondSchema.path('name'), reusedSchemaType);
assert.strictEqual(reusedSchemaType.parentSchema, firstSchema);
assert.notStrictEqual(reusedSchemaType.parentSchema, secondSchema);

console.log('SchemaType was reused:', secondSchema.path('name') === reusedSchemaType);
console.log('actual parentSchema:', reusedSchemaType.parentSchema === firstSchema ? 'firstSchema' : 'secondSchema');
console.log('expected parentSchema:', 'secondSchema');
```

If you reuse a schematype from one schema in another schema, Mongoose will keep the original parentSchema. This is an issue with `parentSchema` that we should look into fixing, but it's also problematic here because you can imagine `firstSchema.path('name').required(true)` won't bust `secondSchema`'s required cache.

However, given that Mongoose also does not support modifying a schema after it has been compiled into a model, this might not even be necessary - creating a document with that schema presupposes that the schema was already compiled into a model. Worth looking into the original #3199 issue to see if there's an alternative fix.

BIGSUS24:
Thanks, good catch — I reproduced the reused-schematype case, `parentSchema` does stay pointing at the first schema.

Taking your pointer back to #3199, I think the cleaner answer is to drop the invalidation entirely rather than try to make it reliable. The constructor now just reads `requiredPaths()` (cache), no `invalidate` flag, and nothing else changed — it's a one-line diff now.

The reasoning: the only thing that warms `_requiredpaths` is the Document code path itself (discriminators already reset it explicitly), and the required paths are fixed once the schema is compiled into a model. So the first construction computes the list fresh — after the schema is fully defined — and every later construction reuses it. The existing `required() with doc arrays (gh-3199)` test still passes, because that `.required(true)` runs before the first document is built.

That also makes the `parentSchema` issue moot here — there's no invalidation code left to be unreliable. I checked the reused-schematype case with `required(true)` set before the first construction and validation is still enforced, since the list is computed from the actual paths at that point.

The one behavior change: flipping a path's `required` after the cache is already warm (i.e. after a document has been constructed) no longer takes effect. That lines up with schemas being immutable once compiled/used, but if you'd rather keep that working I can invalidate at compile time instead — reliable, schema-level, no `parentSchema`.

Perf is the same win as before: ~15% faster construction on a 200–500 path schema (best of 5 runs), negligible on small ones.

vkarpov15:
Closing in favor of #16404

PR #16401, opened 2026-07-23
docs(typescript): note type predicate helper for PopulatedDoc checks

### LINKED ISSUE #12664: Add Typeguard for Populated documents, or at least note that it is possible in the populate guide
### Prerequisites

### Issue

This issue is a proposal to add a official `isPopulated(doc.path)` function to act as a typeguard both for types and in runtime.

A typeguard in typescript asserts that a input type is a specific type, and the same (or at least close to it) at runtime, for example (how typegoose does it):

```ts
function isPopulated<T, S extends mongoose.RefType>(
  doc: mongoose.PopulatedDoc<T, S>
): doc is T extends mongoose.HydratedDocument<infer T1, infer T2, infer T3> ? mongoose.HydratedDocument<T1, T2, T3> : T extends object ? DocumentType<T> : never  {
  return doc instanceof mongoose.Model;
}
```
Note: the typeguard above is a modified version from typegoose and the types have been changed to similar types in mongoose, but has not been tested, so it is just pseudo-code.

Also read [Typescript Narrowing: Using type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

Maybe a opposite function to check if the input is the ref type (ie unpopulated) would be a good idea; but from what i know it would not be just a simple invert of `isPopulated`.

If such a function is not added to mongoose itself, it should be noted in `docs/typescript/populate.html` that it is possible to outsource the check to a function while having the types still work.

Hashim1999164:
## Summary
- Documents using a TypeScript type predicate helper to narrow `PopulatedDoc` paths after a runtime ObjectId check.
- Addresses the docs half of #12664 so users do not need to repeat the check inline everywhere.

## Test plan
- [ ] Review the new example in `docs/typescript/populate.md`
- [ ] Confirm the example matches existing populate TypeScript guidance

Fixes #12664

### DIFF
--- a/docs/typescript/populate.md
+++ b/docs/typescript/populate.md
@@ -97,3 +97,21 @@ Here's two reasons why:
 
 1. You still need to add an extra check to check if `child instanceof ObjectId`. Otherwise, the TypeScript compiler will fail with `Property name does not exist on type ObjectId`. So using `PopulatedDoc<>` means you need an extra check everywhere you use `doc.child`.
 2. In the `Parent` interface, `child` is a hydrated document, which makes it difficult for Mongoose to infer the type of `child` when you use `lean()` or `toObject()`.
+
+If you do use `PopulatedDoc`, you can move that runtime check into a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) helper. TypeScript will then narrow the populated type after the helper returns true:
+
+```ts
+import { Types } from 'mongoose';
+
+function isPopulated<T>(doc: T | Types.ObjectId | null | undefined): doc is T {
+  return doc != null && !(doc instanceof Types.ObjectId);
+}
+
+ParentModel.findOne({}).populate('child').orFail().then((doc: Parent) => {
+  if (!isPopulated(doc.child)) {
+    throw new Error('should be populated');
+  }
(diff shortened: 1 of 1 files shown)


docs/typescript/populate.md
```diff
@@ -97,3 +97,21 @@ Here's two reasons why:
...
+
+```ts
+import { Types } from 'mongoose';
+
+function isPopulated<T>(doc: T | Types.ObjectId | null | undefined): doc is T {
+  return doc != null && !(doc instanceof Types.ObjectId);
+}
+
+ParentModel.findOne({}).populate('child').orFail().then((doc: Parent) => {
+  if (!isPopulated(doc.child)) {
```
vkarpov15, line 111: I suppose you could do this, but I'm not sure why you wouldn't use `if (doc.child == null || doc.child instanceof Types.ObjectId)` here - same end result with no additional helper.

Another alternative would be `if (!(doc.child instanceof mongoose.Document))` - that wouldn't catch the case where `doc.child` is populated using `lean()` but that case is uncommon.

Hashim1999164:
That makes sense. The existing example already demonstrates the direct `null`/`ObjectId` check, and the helper does not improve the runtime behavior. I’m closing this rather than adding redundant guidance.

PR #16405, opened 2026-07-25
perf(document): cache which schema paths have a transform (gh-16378)

### LINKED ISSUE #16378: toObject()/toJSON() walks the entire schema path list per call via applySchemaTypeTransforms, even when no path has a transform
## toObject()/toJSON() walks the entire schema path list per call via applySchemaTypeTransforms, even when no path has a transform

**Location:** lib/document.js:$toObject (~4083) -> applySchemaTypeTransforms (~4443) — Object.keys(schema.paths) loop per document per serialization
**Severity:** medium · **Confidence:** 0.85
**Type:** hot-path-by-call-frequency

### Description

Document.prototype.$toObject in lib/document.js defaults `transform` to true (line ~4070-4085) and unconditionally calls `applySchemaTypeTransforms(this, ret)` (line ~4443). That helper recomputes `Object.keys(schema.paths)` and, for every schema path, performs two option lookups (`schematype.options.transform`, `embeddedSchemaType.options.transform`) — even when zero paths define a transform. The set of transform-bearing paths is a static property of the schema and could be precomputed once at model compile time in lib/schema.js; instead it is re-derived per document per serialization. Cross-module: query results from lib/query.js are hydrated by lib/model.js, and serializing D docs (toJSON in an API response is the classic hot path) costs O(D x P) pure schema-walk overhead independent of how many fields are actually set on each document.

### Benchmark

Calling toObject() on a document with a fixed 5 set fields will get slower roughly linearly as schema path count grows (N=500 several times slower than N=10), proving serialization cost scales with schema width rather than document content; O(N) growth expected.

**Observed complexity:** O(N)

```
| N=10 | 1.69 | 1.00 | — |
| N=50 | 6.49 | 3.84 | — |
| N=100 | 11.52 | 6.81 | — |
| N=500 | 78.35 | 46.32 | — |
```

benchmark confirmed

BIGSUS24:
Fixes #16378.

`applySchemaTypeTransforms` walks `Object.keys(schema.paths)` and re-resolves each path's `transform` option on every `toObject()`/`toJSON()` call, even when the schema has zero transforms (the common case). That makes serialization scale with schema width instead of document size.

## Fix

Cache the list of transform-carrying paths on the schema (`Schema#$transformPaths`, mirrors the existing `requiredPaths` cache), and invalidate from the three schema-local mutation points: `Schema#path`, `Schema#remove`, and discriminator merge.

Two mutation vectors aren't schema-local: `SchemaType.set('transform', fn)` and `SchemaType#transform(fn)` can register a transform on a type after a schema using it has already compiled and serialized once — see the existing `gh-15084` test. Neither can walk back to every schema that uses that SchemaType, so they bump a `SchemaType.optionsVersion` counter instead, and `$transformPaths()` recomputes when it sees a newer version than what it cached. I called this out because a compile-time-only precompute looks like the obvious fix and silently breaks both of those cases (confirmed by writing repro scripts for each before settling on this approach).

## Benchmark

Best-of-5, 20k `toObject()` calls on a document with 5 set fields, at varying schema width:

| schema paths | before | after |
|---|---|---|
| 10 | 22.9ms | 10.7ms |
| 100 | 246.6ms | 11.1ms |
| 500 | 986.8ms | 10.2ms |

`toObject()` cost is now flat in schema size instead of scaling with it.

## Tests

Added three tests in `test/document.test.js` next to the existing `gh-15084` test, covering: `SchemaType.set()` after first serialization, `SchemaType#transform()` after first serialization, and `schema.add()` after first serialization. All three fail without the version-counter/invalidation logic. Also updated a schema mock in `test/document.unit.test.js` that didn't implement the new `$transformPaths()` method.

Full suite passes locally (`npm test`), no live MongoDB connection required for the new tests.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1299,6 +1299,8 @@ Schema.prototype.path = function(path, obj) {
       : undefined;
   }
 
+  this._transformPaths = null;
+
   const subpaths = path.indexOf('.') === -1 ? [path] : path.split('.');
   const last = subpaths.pop();
   if (utils.specialProperties.has(last)) {
@@ -1869,6 +1871,39 @@ Schema.prototype.requiredPaths = function requiredPaths(invalidate) {
   this._requiredpaths = ret;
   return this._requiredpaths;
 };
+
+/**
+ * Returns an Array of path strings that have a `transform` function, either
+ * on the path itself or as a SchemaType-level default (cached).
+ *
+ * @api private
+ * @return {Array}
+ */
+
+Schema.prototype.$transformPaths = function $transformPaths() {
+  if (this._transformPaths != null && this._transformPathsVersion === SchemaType.optionsVersion) {
+    return this._transformPaths;
+  }
+
+  const ret = [];
+  for (const path of Object.keys(this.paths || {})) {
+    const schematype = this.paths[path];
+    if (_hasTransform(schematype) || _hasTransform(schematype.embeddedSchemaType)) {
+      ret.push(path);
+    }
+  }
+
+  this._transformPaths = ret;
+  this._transformPathsVersion = SchemaType.optionsVersion;
+  return this._transformPaths;
+};
+
+function _hasTransform(schematype) {
+  if (schematype == null) {
+    return false;
+  }
+  return typeof (schematype.options?.transform ?? schematype.constructor?.defaultOptions?.transform) === 'function';
+}
 
 /**
  * Returns indexes from fields and schema-level indexes (cached).
@@ -2704,6 +2739,8 @@ Schema.prototype.virtualpath = function(name) {
  * @api public
  */
 Schema.prototype.remove = function(path) {
+  this._transformPaths = null;
+
   if (typeof path === 'string') {
     path = [path];
   }
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1881,7 +1881,7 @@ Schema.prototype.requiredPaths = function requiredPaths(invalidate) {
  */
 
 Schema.prototype.$transformPaths = function $transformPaths() {
-  if (this._transformPaths != null && this._transformPathsVersion === SchemaType.optionsVersion) {
+  if (this._transformPaths != null && this._transformPathsVersion === SchemaType.$transformVersion) {
     return this._transformPaths;
   }
 
@@ -1894,7 +1894,7 @@ Schema.prototype.$transformPaths = function $transformPaths() {
   }
 
   this._transformPaths = ret;
-  this._transformPathsVersion = SchemaType.optionsVersion;
+  this._transformPathsVersion = SchemaType.$transformVersion;
   return this._transformPaths;
 };
 
@@ -3134,6 +3134,7 @@ function isArrayFilter(piece) {
 Schema.prototype._preCompile = function _preCompile() {
   this.plugin(idGetter, { deduplicate: true });
   _precomputeOptimisticConcurrency(this);
+  this.$transformPaths();
 };
 
 /*!
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1881,7 +1881,7 @@ Schema.prototype.requiredPaths = function requiredPaths(invalidate) {
  */
 
 Schema.prototype.$transformPaths = function $transformPaths() {
-  if (this._transformPaths != null && this._transformPathsVersion === SchemaType.$transformVersion) {
+  if (this._transformPaths != null) {
     return this._transformPaths;
   }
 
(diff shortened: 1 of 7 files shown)


BIGSUS24:
Fixed the `_doc` ordering in 08751f113 — check `self._doc` before calling `$transformPaths()` now, so the doc-less early-return path doesn't force a cache-fill scan for nothing.

On the `optionsVersion` granularity point (schemaType.js:389): correct that it's coarser than strictly necessary — it bumps on any `SchemaType.set()` option, not just `transform`. Leaving it as-is since narrowing it means threading the option name through `SchemaType.set()`'s call sites for a cache miss that only matters if you're calling `.set()` on unrelated options in a hot path post-warm, which isn't a pattern in the existing test suite or benchmarks. Happy to narrow it if a maintainer wants it tighter.

vkarpov15 (changes requested):
I think it is better to cache the transformed paths when compiling the model so we don't need to track additional versioning info.

BIGSUS24:
Good call, pushed a change in 2298ca040.

I precompute `Schema#\$transformPaths()` in \`_preCompile()\` now, so the common case (no post-compile mutation) never touches the cache-miss path at all — it's populated by the time the model finishes compiling, no `toObject()`/`toJSON()` call ever recomputes it.

I did keep a narrow invalidation path, because a strictly compile-once-forever cache with zero invalidation breaks tests that predate this PR — specifically `gh-15084` (`test/document.test.js`), which registers a transform via `SchemaType.set('transform', fn)` *after* the model is already compiled and a document has already been serialized once, and expects the next `toObject()` call to pick it up. The 3 tests I added for this PR cover the same shape for `schema.add()` and `SchemaType#transform()` called post-compile.

To keep that narrow: renamed `SchemaType.optionsVersion` to `SchemaType.\$transformVersion` and scoped it to bump only when the `transform` default itself changes — not on every `SchemaType.set()` call for unrelated options like `trim`/`lowercase` (Copilot's review flagged the same thing independently). So it's no longer "versioning info" in the broad sense, just a single counter tied to the one thing that can't be observed from the schema itself.

Let me know if you'd rather drop support for the post-compile-mutation cases outright (i.e. treat `gh-15084`-style usage as no longer guaranteed) — happy to rip out the counter entirely if so, just didn't want to silently regress a test that's presumably there for a reason.

lib/schema.js
```diff
@@ -1870,6 +1872,39 @@ Schema.prototype.requiredPaths = function requiredPaths(invalidate) {
...
+/**
+ * Returns an Array of path strings that have a `transform` function, either
+ * on the path itself or as a SchemaType-level default (cached).
+ *
+ * @api private
+ * @return {Array}
+ */
+
+Schema.prototype.$transformPaths = function $transformPaths() {
+  if (this._transformPaths != null && this._transformPathsVersion === SchemaType.$transformVersion) {
```
vkarpov15: Is transformPathsVersion still useful?

vkarpov15:
I think the better approach would be to fix the test for #15084 - the fact that `mongoose.Schema.Types.CustomType.set('transform', v => v == null ? v : v.value)` is after `db.model()` is incorrect

vkarpov15:
Also could you please commit the benchmark you used?

BIGSUS24:
Okay ,all three, pushed in 024ecdd7e:

- Fixed the gh-15084 test itself — moved the `mongoose.Schema.Types.CustomType.set('transform', ...)` call to before `db.model()`. Agreed that registering a global schematype default after a model's already compiled and serving traffic isn't a pattern we should be bending the cache design around for.
- That made `transformPathsVersion` dead weight, so it's gone — `$transformPaths()` is now a plain "compute once, cache forever" function, no version check. Also removed `SchemaType.$transformVersion` and the bump in `SchemaType.set()`/`SchemaType#transform()` entirely.
- I also fixed up the other two tests I'd added earlier in this PR that leaned on the same post-compile-mutation shape (one for `SchemaType#transform()`, one that was really just a duplicate of the gh-15084 case once `.set('transform', ...)` was involved — dropped that one instead of rewriting it). The `schema.add()`-after-serialization test is untouched, since that one invalidates through `Schema#path()`'s existing local reset, not the static counter.
- Benchmark's in `benchmarks/transformPaths.js` now (500 paths, 20k `toObject()` calls, best of 5 runs after 2 warmup runs).

Full suite: 4436 passing, 0 failing.

vkarpov15:
I'll close this PR since mostly equivalent changes are in #16408. Thanks for the PR - I will give you credit for this work in the changelog.

BIGSUS24:
Appreciate the transparency on this, and glad the caching approach ended up being useful for the bigger perf pass.

Two things:
- For the changelog credit — could you link back to #16405 specifically (or my handle) rather than just a general mention? Helps have something concrete to point to.
- If there's another open perf/bug issue in the repo you'd rather see picked up (labeled `help wanted` or otherwise), point me at it — happy to keep contributing.

PR #16408, opened 2026-07-28
perf: improve toObject() performance re: #14394

vkarpov15:
**Summary**

Re: #14394, there's a few more optimizations we can make:

1. idGetter gets called on every document by default and using `this._id` fires internal getters, which causes perf overhead
2. schema `defaultOptions` should be `null` instead of empty object to make checking for default options faster
3. Allow running shallow clone on populated subdocs to avoid full `clone()` overhead

Before:

```
$ node gh-14394.js 
toObject ms: 291
Done
^C

```

After:

```
$ node gh-14394.js 
toObject ms: 162
Done
^C

```

I also adapted the changes from #16405.

**Examples**

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -727,7 +727,7 @@ Schema.prototype._getDocumentMiddleware = function _getDocumentMiddleware() {
 
 Schema.prototype._defaultToObjectOptions = function(json) {
   const path = json ? 'toJSON' : 'toObject';
-  if (this._defaultToObjectOptionsMap && this._defaultToObjectOptionsMap[path]) {
+  if (this._defaultToObjectOptionsMap && path in this._defaultToObjectOptionsMap) {
     return this._defaultToObjectOptionsMap[path];
   }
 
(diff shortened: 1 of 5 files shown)


lib/schema.js
```diff
@@ -738,7 +738,7 @@ Schema.prototype._defaultToObjectOptions = function(json) {
   const defaultOptions = Object.assign({}, baseOptions, schemaOptions);
 
   this._defaultToObjectOptionsMap = this._defaultToObjectOptionsMap || {};
-  this._defaultToObjectOptionsMap[path] = defaultOptions;
+  this._defaultToObjectOptionsMap[path] = Object.keys(defaultOptions).length > 0 ? defaultOptions : null;
```
AbdelrahmanHafez: If that's on a hot path, consider:
```suggestion
  this._defaultToObjectOptionsMap[path] = defaultOptions && utils.hasOwnKeys(defaultOptions) ? defaultOptions : null;
```

AbdelrahmanHafez (approved):
LGTM!

lib/schema.js
vkarpov15: Not on a hot path (most of the time), but I'll use `utils.hasOwnKeys()` instead just in case.

hasezoey (approved):

Issue #5603, opened 2017-08-31
`select: false` does not work as part of a Child Schema if Child Schema is used more than once


wlingke:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**

In the script below, when we pull the parent document, we see that `child` comes back as `{}`, but `child2` comes back as `{ field: 'test'}` despite the fact that `ChildSchema.field` has `select: false`.

When looking at the debugger, only `child.field` is being selected out and `child2.field` is not.

```


const mongoose = require('mongoose');

mongoose.connect('mongodb://127.0.0.1/test');
const db = mongoose.connection;

db.once('open', async () => {
  try {
    const ChildSchema = new mongoose.Schema({
      field: {
        type: String,
        select: false,
      },
      _id: false,
    }, { id: false });

    const ParentSchema = new mongoose.Schema({
      child: ChildSchema,
      child2: ChildSchema,
    });
    const Parent = mongoose.model('Parent', ParentSchema);
    const ogParent = new Parent();
    ogParent.child = { field: 'test' };
    ogParent.child2 = { field: 'test' };
    await ogParent.save();

    const foundParent = await Parent.findById(ogParent._id).exec();
    console.log(foundParent);

    db.close();
  } catch (e) {
    console.log(e);
  }
});


```

**Please mention your node.js, mongoose and MongoDB version.**
4.11.8

varunjayaraman:
thanks for the full repro script, this looks like a bug

heisian:
Giving @wlingke an imaginary badge for Best Mongoose Bug Finder of the Year, 2017™ 🐜

vkarpov15:
@wlingke , finding all the most subtle bugs in mongoose since 2014 :rocket: :1st_place_medal: 

![image](https://user-images.githubusercontent.com/1620265/30464302-ac53f430-9985-11e7-95a2-ef80e203a45a.png)

PR #16409, opened 2026-07-28
fix(query): avoid path collision when excluding subdocument with select: false child (gh-12798)

### LINKED ISSUE #12798: Merge select/projection of subdocument tree

BIGSUS24:
Fixes #12798.

`applyPaths()` was unconditionally projecting every schema-level `select: false` path onto `fields`, even when an ancestor of that path was already excluded by the user's own `.select()` call. So `.select('-subd')` on a schema where `subd.raw` has `select: false` produced `{ subd: 0, 'subd.raw': 0 }`, which MongoDB rejects with a path collision error, even though `subd: 0` already covers `subd.raw`.

Fix: in the exclusive-projection branch of `applyPaths` (`lib/queryHelpers.js`), skip adding a schema-excluded path if one of its dot-path ancestors is already excluded in `fields`. The check walks the path segments directly (`O(depth)`, depth is realistically 1-3), no scan over the whole projection object, so it doesn't reintroduce the O(n²) concern from the previous attempt at this (#15807).

This is the missing half of an existing pattern rather than a new mechanism: `analyzePath`'s existing "check for parent exclusions" loop already skips a schema-excluded path when an *ancestor is also schema-excluded* (`select: false` on the parent). That loop only checks the `excluded` array, so it never covered the case here, where the ancestor is excluded by the *user's own* `.select()` call instead of the schema. This fix adds that other half.

```js
  name: String,
  subd: {
    raw: { type: String, select: false },
    clean: String
  }
});
const Bar = mongoose.model('Bar', BarSchema);

Bar.find().select('-subd'); // no longer throws "Path collision"
```

The ancestor check treats any falsy defining value (`0`, `false`, `''`, `NaN`) as exclusion, matching how `exclude` itself is derived a few lines above (`exclude = !field`) -- e.g. `.select({ subd: false })` is equally valid and isn't normalized to `0` by the time this check runs.

Added a regression test in `test/query.test.js` under the existing `bug fixes` block (same spot as the neighboring gh-5603 test) covering: the issue's exact schema shape (asserts the resolved projection collapses to `{ subd: 0 }` and a live query resolves with the excluded fields undefined instead of throwing), the boolean- and empty-string-exclusion forms, and 3-4 level deep nesting plus a document array, each asserting the ancestor-covered descendant is dropped while an unrelated sibling exclusion still comes through (so the guard isn't over-eager).

Known, unrelated, pre-existing gap left untouched: `Model.find().slice('arr', 5)` on a schema where `arr`'s children have `select: false` can still produce a similar collision (`{arr: {$slice: 5}, 'arr.child': 0}`). That path never reaches this fix at all -- `$slice`/`$meta` values don't count as a defining projection, so the query has no exclusive/inclusive determination (`exclude` stays `undefined`) and falls into a separate branch this PR doesn't touch. Widening that branch risks silently dropping `select: false` fields that should still apply alongside `$slice`/`$meta`, so didn't fold it into this PR -- flagging in case it's worth its own issue.

Two files touched, no reformatting. `npm run lint` and the full suite pass locally (4461 passing; the one unrelated failure is an encryption integration test that requires `npm run setup-test-encryption`, pre-existing and unrelated to this change).

### DIFF
--- a/lib/queryHelpers.js
+++ b/lib/queryHelpers.js
@@ -384,9 +384,11 @@ exports.applyPaths = function applyPaths(fields, schema, sanitizeProjection) {
     let i = -1;
     while ((i = path.indexOf('.', i + 1)) !== -1) {
       const ancestor = fields[path.slice(0, i)];
-      // `false` is also a valid exclusion value alongside `0`, e.g.
-      // `.select({ subd: false })`, so check both. See gh-12798
-      if (ancestor === 0 || ancestor === false) {
+      // Any falsy defining value (`0`, `false`, `''`, `NaN`) excludes a
+      // path, matching how `exclude` itself is derived above (`!field`).
+      // `ancestor == null` means the key isn't present at all, which is
+      // not an exclusion. See gh-12798
+      if (ancestor != null && !ancestor) {
         return true;
       }
     }
--- a/lib/queryHelpers.js
+++ b/lib/queryHelpers.js
@@ -383,7 +383,10 @@ exports.applyPaths = function applyPaths(fields, schema, sanitizeProjection) {
   function hasExcludedAncestor(fields, path) {
     let i = -1;
     while ((i = path.indexOf('.', i + 1)) !== -1) {
-      if (fields[path.slice(0, i)] === 0) {
+      const ancestor = fields[path.slice(0, i)];
+      // `false` is also a valid exclusion value alongside `0`, e.g.
+      // `.select({ subd: false })`, so check both. See gh-12798
+      if (ancestor === 0 || ancestor === false) {
         return true;
       }
     }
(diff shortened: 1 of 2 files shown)


BIGSUS24:
Happy to adjust anything here — different approach, more/different test cases, whatever. Just let me know what you'd want changed.

vkarpov15 (approved):
Neat, thank you :+1:

PR #16411, opened 2026-07-29
perf(timestamps): avoid adding $setOnInsert for createdAt unless upsert set

vkarpov15:
**Summary**

I spotted a minor issue when working on [this blog post](https://thecodebarbarian.com/mongoose-studio-cassandra-data-api.html): `applyTimestampsToUpdate` always adds `$setOnInsert: { createdAt }` for _all_ updates, even ones that don't have `upsert: true`. `$setOnInsert` is guaranteed to be a no-op unless `upsert` is set, so `applyTimestampsToUpdate` is constructing useless objects and wasting bandwidth.

I also removed the `get()` calls in `applyTimestampsToUpdate()`, which should slightly help performance.

**Examples**

### DIFF
--- a/lib/helpers/update/applyTimestampsToUpdate.js
+++ b/lib/helpers/update/applyTimestampsToUpdate.js
@@ -13,10 +13,10 @@ module.exports = applyTimestampsToUpdate;
  * ignore
  */
 
-function applyTimestampsToUpdate(now, createdAt, updatedAt, currentUpdate, options, isReplace) {
+function applyTimestampsToUpdate(now, createdAt, updatedAt, currentUpdate, options, upsert, isReplace) {
   const updates = currentUpdate;
   let _updates = updates;
-  const timestamps = get(options, 'timestamps', true);
+  const timestamps = options?.timestamps ?? true;
 
   // Support skipping timestamps at the query level, see gh-6980
   if (!timestamps || updates == null) {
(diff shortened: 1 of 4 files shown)


hasezoey (approved):

lib/helpers/update/applyTimestampsToUpdate.js
```diff
@@ -81,7 +80,7 @@ function applyTimestampsToUpdate(now, createdAt, updatedAt, currentUpdate, optio
   }
 
   if (!skipCreatedAt && createdAt) {
-    const overwriteImmutable = get(options, 'overwriteImmutable', false);
+    const overwriteImmutable = options?.overwriteImmutable;
```
AbdelrahmanHafez, line 83: Not sure if it matters here, but this _could_ change behavior.
```suggestion
    const overwriteImmutable = options?.overwriteImmutable ?? false;
```

On another note, do you think we shuold get rid of all `get` calls in favor of optional chaining? Happy to create a cleanup PR with this.

AbdelrahmanHafez (commented):
One minor comment, otherwise LGTM. Good catch!

lib/helpers/update/applyTimestampsToUpdate.js
vkarpov15: Could change behavior sometime in the future, but as written `?? false` makes no difference - local scoped variable only used with a truthy check on L89.

In terms of removing `get()` calls - yes but only in cases where the `path` param is a string literal. There's several cases where we use `get()` with `path` as a variable.

PR #16436, opened 2026-08-04
fix(document): support map wildcards in optimistic concurrency paths option

vkarpov15:
Re: #16383

**Summary**

I created a new `PathTrie` class that we can use to create a trie for faster lookups for the `optimisticConcurrency` path. Since `optimisticConcurrency` is set on the schema level, building the trie should happen when the model is compiled, so no need to worry about the overhead of building a trie in a `save()` call. This trie class should also help us in other cases where we want to support wildcards.

I'm going to work on the array path next.

**Examples**


lib/helpers/pathTrie.js
hasezoey, line 1: It would be good to have all the functions here to have a `@param` type documentation.

hasezoey (approved):

AbdelrahmanHafez (approved):

lib/helpers/pathTrie.js
vkarpov15: Fixed in e0901ed

PR #16438, opened 2026-08-05
fix(model): set document session in bulkSave() so transaction retries restore document state

### LINKED ISSUE #16432: Connection#transaction() does not restore document state for bulkSave({ session }) before a callback retry
### Prerequisites

### Mongoose version

9.9.1

### Node.js version

24.11.0

### MongoDB server version

7.0.24

N/A

### Description

When the same hydrated documents are passed to
`Model.bulkSave(documents, { session })` from a
`Connection#transaction()` callback, a full callback retry can omit the
intended insert or update.

The reproduction performs a successful write and then throws a synthetic
`MongoServerError` with the `TransientTransactionError` label. This is only a
deterministic way to make the driver retry the complete callback; it is not
presented as a server-generated `WriteConflict`.

On the first attempt, `bulkSave()` changes the in-memory document state after
the write. MongoDB then aborts that attempt. The second callback starts with
the reused documents still reporting `isNew === false` and no modified paths,
so `bulkSave()` sends no second write command. The transaction resolves, but
the first attempt's database write was rolled back.

The same sequence using `document.save({ session })` is included as a control.
Its document state is restored before attempt 2, two insert commands are sent,
and the document is persisted.

This report is scoped to Mongoose's `Connection#transaction()` wrapper. It
does not claim that a direct `session.withTransaction()` call should restore
Mongoose document state.

An auxiliary `repro-write-conflict.js` in the reproduction bundle repeats the
same three cases without throwing a synthetic error or altering server
behavior. A second MongoDB client updates a gate document after the subject
transaction has established its snapshot. The subject transaction then
receives server error 112 (`WriteConflict`) with the
`TransientTransactionError` label. All callbacks run twice and produce the
same command counts and persisted results shown below.

### Related issues

- [#13698](https://github.com/Automattic/mongoose/issues/13698) and
  [#13973](https://github.com/Automattic/mongoose/issues/13973) fixed `isNew`
  and modified-state restoration for `save()`. This report exercises
  `bulkSave()` and remains reproducible on 9.9.1.
- [#14763](https://github.com/Automattic/mongoose/issues/14763), fixed by
  [#14884](https://github.com/Automattic/mongoose/pull/14884), covers generated
  bulk operations that do not all match. Here attempt 2 generates zero write
  operations, so its incomplete-write count check is not reached.

### Source observations and suspected mechanism

The following is an implementation hypothesis based on the 9.9.1 source and
the controls below, not a prescribed fix:

- `Document#save()` associates an explicitly supplied session with the
  document ([source](https://github.com/Automattic/mongoose/blob/9.9.1/lib/model.js#L651-L654)).
- Transaction tracking reads the session from `this.$session()`
  ([source](https://github.com/Automattic/mongoose/blob/9.9.1/lib/plugins/trackTransaction.js#L12-L33)).
- `bulkSave()` forwards its options to `bulkWrite()`
  ([source](https://github.com/Automattic/mongoose/blob/9.9.1/lib/model.js#L3608-L3629)),
  then its successful-write handler updates `isNew` and modified-path state
  ([source](https://github.com/Automattic/mongoose/blob/9.9.1/lib/model.js#L3677-L3686)).
- `Connection#transaction()` can reset only documents present in its tracked
  session map
  ([source](https://github.com/Automattic/mongoose/blob/9.9.1/lib/connection.js#L748-L786)).

As a diagnostic control, explicitly calling `document.$session(session)` before
`bulkSave(documents, { session })` makes both failing cases pass. This is
consistent with the documents not being registered for transaction-state
restoration, but the maintainers may identify a different root cause or prefer
a different correction.

### Steps to Reproduce

Create a clean project, generate a lockfile with the exact direct dependencies,
and install from that lockfile:

```bash
mkdir mongoose-bulk-save-retry-repro
cd mongoose-bulk-save-retry-repro
npm init -y
npm install --package-lock-only --save-exact \
  mongoose@9.9.1 \
  mongodb@7.5.0 \
  mongodb-memory-server@11.2.0
npm ci
```

Save the following as `repro.js`:

<details>
<summary>repro.js</summary>

```javascript
'use strict';

const assert = require('node:assert/strict');
const { MongoServerError } = require('mongodb');
const { version: mongodbDriverVersion } = require('mongodb/package.json');
const { MongoMemoryReplSet } = require('mongodb-memory-server');
const { version: mongodbMemoryServerVersion } = require('mongodb-memory-server/package.json');

const scenarioNames = ['saveControl', 'bulkNew', 'bulkExisting'];

function makeTransientTransactionError() {
  const error = new MongoServerError({
    message: 'Synthetic error to force one full transaction callback retry',
    code: 112,
  });
  error.addErrorLabel('TransientTransactionError');
  return error;
}

function documentState(document) {
  return {
    isNew: document.isNew,
    modifiedPaths: document.modifiedPaths().sort(),
  };
}

async function runWithOneRetry(connection, operation, getState) {
  const attempts = [];

  await connection.transaction(async (session) => {
    const attempt = attempts.length + 1;
    const before = getState();

    await operation(session);

    attempts.push({
      attempt,
      before,
      after: getState(),
    });

    if (attempt === 1) {
      throw makeTransientTransactionError();
    }
  });

  return attempts;
}

async function main() {
  let replSet;
  let connection;

  try {
    replSet = await MongoMemoryReplSet.create({
      binary: { version: '7.0.24' },
      replSet: {
        count: 1,
        storageEngine: 'wiredTiger',
      },
    });

    connection = await mongoose.createConnection(replSet.getUri(), {
      dbName: 'mongoose_bulk_save_retry_repro',
      monitorCommands: true,
    }).asPromise();

    let activeScenario = null;
    const writeCommands = [];
    connection.getClient().on('commandStarted', (event) => {
      if (activeScenario == null || !['insert', 'update'].includes(event.commandName)) {
        return;
      }

      const operations = event.command.documents ?? event.command.updates ?? [];
      writeCommands.push({
        scenario: activeScenario,
        command: event.commandName,
        operationCount: operations.length,
      });
    });

    const Item = connection.model('Item', new mongoose.Schema({
      name: {
        type: String,
        required: true,
      },
    }));

    // Avoid implicit collection creation during a transaction.
    await Item.init();

    const runScenario = async (name, operation, getState) => {
      activeScenario = name;
      try {
        return await runWithOneRetry(connection, operation, getState);
      } finally {
        activeScenario = null;
      }
    };

    // Control: save({ session }) restores the reused document before retrying.
    const saveDocument = new Item({ name: 'save control' });
    const saveAttempts = await runScenario(
      'saveControl',
      (session) => saveDocument.save({ session }),
      () => documentState(saveDocument),
    );

    // Failing case 1: a reused new document is not inserted on the retry.
    const bulkNewDocument = new Item({ name: 'bulk new' });
    const bulkNewAttempts = await runScenario(
      'bulkNew',
      (session) => Item.bulkSave([bulkNewDocument], { session }),
      () => documentState(bulkNewDocument),
    );

    // Failing case 2: a reused existing document is not updated on the retry.
    const bulkExistingDocument = await Item.create({ name: 'before' });
    bulkExistingDocument.name = 'after';

    const bulkExistingAttempts = await runScenario(
      'bulkExisting',
      (session) => Item.bulkSave([bulkExistingDocument], { session }),
      () => ({
        ...documentState(bulkExistingDocument),
        name: bulkExistingDocument.name,
      }),
    );

    const serverInfo = await connection.db.admin().serverInfo();
    const persistedExisting = await Item.findById(bulkExistingDocument._id).lean();
    const writeCommandCounts = Object.fromEntries(
      scenarioNames.map(name => [
        name,
        writeCommands.filter(command => command.scenario === name).length,
      ]),
    );

    const result = {
      versions: {
        node: process.version,
        mongoose: mongoose.version,
        mongodbDriver: mongodbDriverVersion,
        mongodbMemoryServer: mongodbMemoryServerVersion,
        mongodbServer: serverInfo.version,
      },
      attempts: {
        saveControl: saveAttempts,
        bulkNew: bulkNewAttempts,
        bulkExisting: bulkExistingAttempts,
      },
      writeCommands,
      writeCommandCounts,
      persisted: {
        saveControl: await Item.exists({ _id: saveDocument._id }) !== null,
        bulkNew: await Item.exists({ _id: bulkNewDocument._id }) !== null,
        bulkExistingName: persistedExisting.name,
      },
    };

    console.log(JSON.stringify(result, null, 2));

    assert.deepEqual(
      result.persisted,
      {
        saveControl: true,
        bulkNew: true,
        bulkExistingName: 'after',
      },
      'transaction resolved without the intended bulkSave() writes',
    );
  } finally {
    if (connection != null) {
      await connection.close();
    }
    if (replSet != null) {
      await replSet.stop();
    }
  }
}

main().catch((error) => {
  console.error(error);
  process.exitCode = 1;
});
```

</details>

Run:

```bash
node repro.js
```

### Observed output

All three callbacks run twice. The state trace shows that the `save()` control
starts attempt 2 as a new, modified document, whereas both `bulkSave()` cases
start attempt 2 with no modified paths. Command monitoring records two writes
for the control and only the rolled-back first-attempt write for each
`bulkSave()` case.

Relevant output:

```json
{
  "versions": {
    "node": "v24.11.0",
    "mongoose": "9.9.1",
    "mongodbDriver": "7.5.0",
    "mongodbMemoryServer": "11.2.0",
    "mongodbServer": "7.0.24"
  },
  "writeCommandCounts": {
    "saveControl": 2,
    "bulkNew": 1,
    "bulkExisting": 1
  },
  "persisted": {
    "saveControl": true,
    "bulkNew": false,
    "bulkExistingName": "before"
  }
}
```

The final assertion intentionally fails, because the expected new document is
absent and the existing document still contains `name: "before"`.

### Expected Behavior

Before `Connection#transaction()` retries the callback, documents successfully
processed by `bulkSave({ session })` should be in a state that allows the
intended writes to be generated again. In this reproduction that means:

- the new document is inserted on attempt 2;
- the existing document's `name` update is sent again on attempt 2;
- the committed database state contains both intended writes.

If `bulkSave()` cannot participate in `Connection#transaction()` document-state
restoration, the operation should reject or the limitation should be explicit,
rather than allowing the wrapper to resolve after omitting those writes.

yogesh968:
**Summary**

`Connection#transaction()` snapshots document state in the built-in `trackTransaction` `pre('save')` middleware and restores it in `_resetSessionDocuments()` when the callback is retried. That middleware reads `this.$session()`, but `bulkSave(docs, { session })` never associated the session with the documents the way `save({ session })` does, so the snapshot was skipped entirely.

The consequence is silent data loss. On a retry the documents still report `isNew === false` with no modified paths, so `bulkSave()` issues no second write. The transaction then commits successfully even though the first attempt's writes were rolled back.

This associates the documents with `options.session` before running `pre('save')` hooks, mirroring what `Model.prototype.$save()` already does.

Fixes #16432

**Examples**

Both tests fail on `master` and pass with this change:

```js
const doc = new Test({ name: 'test' });
let retryCount = 0;
await db.transaction(async(session) => {
  assert.ok(doc.$isNew);              // fails on attempt 2: $isNew is already false
  await Test.bulkSave([doc], { session });
  if (++retryCount < 3) {
    throw new mongoose.mongo.MongoServerError({ errorLabels: ['TransientTransactionError'] });
  }
});
assert.equal((await Test.find()).length, 1);
```

The update path loses data rather than just state:

```js
const doc = await Test.findById(_id).orFail();
await db.transaction(async(session) => {
  doc.name = 'updated';
  await Test.bulkSave([doc], { session });
  if (++retryCount < 3) { /* transient error */ }
});
// on master the document is still 'initial'
```

Note that #16435 fixes the `deleteOne()` case in the same family but does not touch `bulkSave()`.

**Testing**

Two tests added to `test/docs/transactions.test.js`. Full transactions suite (25) and all `bulkSave` tests in `model.test.js` (29) pass, along with `document.test.js` / `model.skip.middleware.test.js` (1153 total).


vkarpov15 (approved):
Thanks :+1:

PR #16439, opened 2026-08-05
perf(document): cache projection metadata and avoid scanning all projected paths for flat projections

yogesh968:
**Summary**

`isSelected()` recomputed everything from `$__.selected` on every call: it allocated a fresh `Object.keys()` array, re-derived whether the projection was inclusive or exclusive, and then scanned every projected key looking for a prefix relationship with the requested path.

`toObject({ getters: true })` calls `isSelected()` once per schema path, so serializing a projected document cost O(schema paths x projected paths). #16407 reduced how often it is called; this reduces what each call costs.

Two changes:

- `$__.selected` is assigned once in the Document constructor (`document.js:159`) and never mutated afterwards, so the derived values are computed once per document and cached on `$__`. `isDirectSelected()` reuses the same cache.
- When no projected key is nested, no key can start with `path + '.'`, so only the "is an ancestor of `path` projected?" check can match. Every matching ancestor yields the same answer, so walking the path's own ancestors is equivalent to scanning every projected key, and is O(depth of path) instead of O(number of projected paths). Projections containing a nested key still take the original scan, unchanged.

Re: #16373

**Examples**

Schema width 2N, inclusive projection of N keys, every path with a getter, 200 `toObject({ getters: true })` calls:

| N | master | this PR | speedup |
|---|---|---|---|
| 10 | 1.92ms | 1.00ms | 1.9x |
| 50 | 10.41ms | 5.31ms | 2.0x |
| 100 | 37.06ms | 17.14ms | 2.2x |
| 500 | 582.60ms | 120.58ms | **4.8x** |

Growth relative to N=10 drops from 304x to 126x.

**Testing**

Four tests added to `test/document.isselected.test.js` covering inclusive projections, exclusive projections, parent-selects-children, and agreement between the optimized flat path and the general nested path.

To check equivalence rather than assume it, I also ran a differential fuzz comparing this implementation against the previous algorithm across randomly generated flat and nested projections: **68,000 checks, 0 mismatches**.

`document.test.js`, `document.isselected.test.js`, `model.field.selection.test.js`, `query.test.js`, `model.populate.test.js`, `document.populate.test.js`, `document.modified.test.js` and `document.strict.test.js` all pass (1420 tests).

### DIFF
--- a/lib/document.js
+++ b/lib/document.js
@@ -2570,23 +2570,13 @@ Document.prototype.isSelected = function isSelected(path) {
     return path.some(p => this.$__isSelected(p));
   }
 
-  const paths = Object.keys(this.$__.selected);
-  let inclusive = null;
+  const index = _getProjectionIndex(this);
+  const paths = index.paths;
+  const inclusive = index.inclusive;
 
-  if (paths.length === 1 && paths[0] === '_id') {
+  if (index.onlyId) {
     // only _id was selected.
     return this.$__.selected._id === 0;
-  }
-
-  for (const cur of paths) {
-    if (cur === '_id') {
-      continue;
-    }
-    if (!isDefiningProjection(this.$__.selected[cur])) {
-      continue;
-    }
-    inclusive = !!this.$__.selected[cur];
-    break;
   }
 
   if (inclusive === null) {
@@ -2598,6 +2588,25 @@ Document.prototype.isSelected = function isSelected(path) {
   }
 
   const pathHasDot = path.indexOf('.') !== -1;
+
+  if (!index.hasNestedKey) {
+    // No projection key is nested, so no key can start with `path + '.'` and
+    // only the "is an ancestor of `path` projected?" check below can match.
+    // Every ancestor that matches returns the same value, so we can walk
+    // `path`'s ancestors instead of scanning every projection key. This keeps
+    // `isSelected()` O(depth of path) rather than O(number of projected paths),
+    // which matters because `toObject({ getters: true })` calls this once per
+    // schema path. Re: gh-16373
+    if (pathHasDot) {
+      for (let dot = path.indexOf('.'); dot !== -1; dot = path.indexOf('.', dot + 1)) {
+        const ancestor = path.slice(0, dot);
+        if (ancestor !== '_id' && Object.hasOwn(this.$__.selected, ancestor)) {
+          return inclusive;
+        }
+      }
+    }
+    return !inclusive;
+  }
 
   for (const cur of paths) {
     if (cur === '_id') {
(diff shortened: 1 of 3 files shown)


vkarpov15 (approved):
I did some testing and you're right, this is a good performance improvement. The following benchmark highlights how much faster this is in the case where there's a large projection with no nested paths:

```
'use strict';

// Measures isSelected() for projections with many flat keys. This is the case
// where the fast path avoids scanning every projection key for every schema
// path (for example, toObject({ getters: true })).

const iterations = Number(process.env.MONGOOSE_BENCHMARK_ITERATIONS) || 100000;
const samples = Number(process.env.MONGOOSE_BENCHMARK_SAMPLES) || 5;
const projectionSizes = [10, 100, 500];

const schemaDefinition = {
  nested: { deep: { value: String } }
};
for (let i = 0; i < 1000; ++i) {
  schemaDefinition[`field${i}`] = String;
}

const Model = mongoose.model('IsSelectedBenchmark', new mongoose.Schema(schemaDefinition));

function makeCase(projectionSize) {
  const projection = {};
  for (let i = 0; i < projectionSize; ++i) {
    projection[`field${i}`] = 1;
  }

  return { projection, path: 'nested.deep.value' };
}

function makeDoc(testCase) {
  return new Model({}, testCase.projection);
}

function median(values) {
  values.sort((a, b) => a - b);
  return values[Math.floor(values.length / 2)];
}

function measure(testCase, count) {
  const doc = makeDoc(testCase);

  // Include lazy-index construction in the first-call measurement.
  const firstStart = process.hrtime.bigint();
  const first = doc.isSelected(testCase.path);
  const firstNs = Number(process.hrtime.bigint() - firstStart);

  for (let i = 0; i < 10000; ++i) {
    doc.isSelected(testCase.path);
  }

  let selected = first ? 1 : 0;
  for (let i = 0; i < count; ++i) {
    selected += doc.isSelected(testCase.path) ? 1 : 0;
  }
  const steadyNs = Number(process.hrtime.bigint() - start) / count;

  if (selected !== 0) {
    throw new Error('Unexpected isSelected() result');
  }
  return { firstNs, steadyNs };
}

console.log(`iterations per sample: ${iterations}, samples: ${samples}`);
console.log('projection keys\tfirst call ns\tsteady state ns\testimated 100 calls ns');

for (const projectionSize of projectionSizes) {
  const testCase = makeCase(projectionSize);
  const results = [];

  // Warm up the implementation before collecting samples.
  measure(testCase, Math.min(iterations, 1000));
  for (let i = 0; i < samples; ++i) {
    results.push(measure(testCase, iterations));
  }

  const first = median(results.map(result => result.firstNs));
  const steady = median(results.map(result => result.steadyNs));
  console.log(`${projectionSize}\t\t${first.toFixed(2)}\t\t${steady.toFixed(2)}\t\t${(first + steady * 99).toFixed(2)}`);
}
```

Before:

```
Iterations per sample: 100000, samples: 5
projection keys	first call ns	steady state ns	estimated 100 calls ns
10		1386.00		199.64		21149.99
100		4885.00		2429.77		245432.46
500		32405.00		11762.67		1196909.53

```

After:

```
iterations per sample: 100000, samples: 5
projection keys	first call ns	steady state ns	estimated 100 calls ns
10		2099.00		189.92		20900.98
100		8494.00		198.98		28192.66
500		56618.00		209.70		77377.91
```

The below jsbenchmark in particular shows that `Object.keys()` is non-negligible overhead

PR #16442, opened 2026-08-05
docs(populate): document that `skip` needs `sort` to paginate reliably

### LINKED ISSUE #7579: Populate with options, `skip` does not work when `sort` not defined
Bug

Populate with options, `skip` does not work when `sort` not defined
`.populate({ options: {skip, limit} })`

https://github.com/199911/mongoose-populate-bug/blob/master/index.js#L98

Story list have stories [A, B, C]
expect `.populate({ options: {skip: 0, limit: 1} })` return A
expect `.populate({ options: {skip: 1, limit: 1} })` return B
but `.populate({ options: {skip: 1, limit: 1} })` still return A

Node,js 8, mongoose 5, MongoDB 4

yogesh968:
**Summary**

MongoDB does not guarantee result order without a sort, and `skip` counts against that order, so paginating a populated path with `skip` alone can return a document twice or miss one. Mongoose sends the query correctly, which is why this reads as a Mongoose bug when it is really a missing `sort` — the conclusion reached in #7579.

Adds a short section next to the existing `limit` vs. `perDocumentLimit` discussion showing the unreliable and reliable forms, plus a ToC entry, and notes that the same applies to `skip` on ordinary queries.

Fixes #7579

**Examples**

```js
// Not reliable: the server is free to return the fans in any order, so
// there is no guarantee that page 2 continues where page 1 stopped.
await Story.findById(id).populate({ path: 'fans', options: { skip: 10, limit: 10 } });

// Reliable: `sort` gives `skip` a well-defined order to count from.
await Story.findById(id).populate({ path: 'fans', options: { sort: { _id: 1 }, skip: 10, limit: 10 } });
```

Only `docs/populate.md` is touched; no generated `.html`. `markdownlint` passes.

### DIFF
--- a/docs/populate.md
+++ b/docs/populate.md
@@ -41,6 +41,7 @@ the `Story` model.
   <li><a href="#populating-multiple-paths">Populating Multiple Paths</a></li>
   <li><a href="#query-conditions">Query conditions and other options</a></li>
   <li><a href="#limit-vs-perDocumentLimit"><code>limit</code> vs. <code>perDocumentLimit</code></a></li>
+  <li><a href="#populate-skip-sort">Always pair <code>skip</code> with <code>sort</code></a></li>
   <li><a href="#refs-to-children">Refs to children</a></li>
   <li><a href="#populate_an_existing_mongoose_document">Populating an existing document</a></li>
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #16444, opened 2026-08-06
fix(schema): put enum on array elements and support the object enum form in toJSONSchema

### LINKED ISSUE #16443: toJSONSchema() puts `enum` on the array instead of its elements, and throws on the `{ values, message }` enum form
### Prerequisites

### Mongoose version

9.9.1 (also reproduces on 8.x)

### Node.js version

22.12.0

### MongoDB server version

8.2.6

### Description

`Schema.prototype.toJSONSchema()` reads enum values straight off `schemaType.options.enum`. That
raw option is not the normalized enum, which causes three separate problems:

1. **Arrays of enums get the `enum` on the array itself instead of on its items.** For
   `{ tags: { type: [String], enum: ['a', 'b'] } }` Mongoose validates each *element* against the
   enum, but the generated JSON schema requires the whole `tags` array to be equal to `'a'`,
   `'b'` or `null`. Every document Mongoose considers valid is rejected by the `$jsonSchema`
   validator Mongoose generated for it, so `Model.create()` fails against a collection created
   from `schema.toJSONSchema({ useBsonType: true })`.

2. **The documented `enum: { values, message }` form throws.** `[...schemaType.options.enum, null]`
   spreads a plain object, so `toJSONSchema()` exits with
   `TypeError: schemaType.options.enum is not iterable`. When the path is `required`, it does not
   throw but emits `enum: { values: [...], message: '...' }`, which is not a valid JSON schema —
   Ajv rejects it and MongoDB errors with `$jsonSchema keyword 'enum' must be an array`.

3. **Enums declared on the array element, or via `.enum()`, are dropped.** `[{ type: String, enum: [...] }]`
   and `schema.path('status').enum('a', 'b')` never set `options.enum` on the path
   `toJSONSchema()` looks at, so the constraint silently disappears from the generated schema.

### Steps to Reproduce

```js
const Ajv = require('ajv');

// 1. array of enums: valid document rejected by its own generated schema
const arraySchema = new mongoose.Schema({ tags: { type: [String], enum: ['a', 'b'] } });
const Test = mongoose.model('Test', arraySchema);

console.log(JSON.stringify(arraySchema.toJSONSchema().properties.tags));
// { "type": ["array","null"], "items": { "type": ["string","null"] }, "enum": ["a","b",null] }
//                                                                     ^^^^ on the array, not the items

const doc = new Test({ tags: ['a', 'b'] });
console.log(doc.validateSync()); // undefined -- Mongoose says this document is valid

const validate = new Ajv().compile(arraySchema.toJSONSchema());
console.log(validate(JSON.parse(JSON.stringify(doc.toObject())))); // false
console.log(validate.errors[0].message); // 'must be equal to one of the allowed values'

// 2. `{ values, message }` enum form
const objectFormSchema = new mongoose.Schema({
  status: { type: String, enum: { values: ['a', 'b'], message: '{VALUE} is not supported' } }
});
objectFormSchema.toJSONSchema(); // TypeError: schemaType.options.enum is not iterable

// 3. enum on the array element is dropped
const casterSchema = new mongoose.Schema({ tags: [{ type: String, enum: ['a', 'b'] }] });
console.log(JSON.stringify(casterSchema.toJSONSchema().properties.tags));
// { "type": ["array","null"], "items": { "type": ["string","null"] } } -- no enum at all
```

Case 1 end to end against a real server: creating the collection with
`validator: { $jsonSchema: arraySchema.toJSONSchema({ useBsonType: true }) }` and then calling
`Test.create({ tags: ['a'] })` fails with `MongoServerError: Document failed validation`.

### Expected Behavior

`toJSONSchema()` should describe the same values the enum validator accepts:

- arrays put the `enum` on `items`, since that is what Mongoose validates
- all documented enum declarations (`['a','b']`, `{ values, message }`, an object of values,
  and the `.enum()` method) produce the same `enum` array

The normalized values are already available as `schemaType.enumValues`, which covers every form.

Jaybhade:
**Summary**

Fix #16443.

`Schema.prototype.toJSONSchema()` read enum values straight off `schemaType.options.enum` — the
raw user input rather than the normalized enum — which broke in three ways:

1. **Arrays of enums got the `enum` on the array instead of on its items.** Mongoose validates
   each element against the enum, but the generated schema required the whole array to equal one
   of the enum values, so the `$jsonSchema` validator rejected every document Mongoose considered
   valid.
2. **The documented `enum: { values, message }` form threw.** `[...schemaType.options.enum, null]`
   spread a plain object: `TypeError: schemaType.options.enum is not iterable`. On a `required`
   path it did not throw but emitted `enum: { values, message }`, which MongoDB rejects with
   `$jsonSchema keyword 'enum' must be an array`.
3. **Enums declared on the array element or set with `.enum()` were dropped** — neither sets
   `options.enum` on the path `toJSONSchema()` inspects.

The fix emits the enum from the schema type that owns it — `SchemaString` and `SchemaNumber`,
the two types the API documents enum support for — using the normalized `enumValues`. Arrays then
describe their items for free, since `SchemaArray.toJSONSchema()` already delegates to the
embedded type, and every declaration form is covered because `enumValues` is what the enum
validator itself uses. `null` is appended exactly where the type definition is nullable, which
also makes `allowNull: false` consistent between the type and the enum.

**Examples**

Before, for `{ tags: { type: [String], enum: ['funny', 'sad'] } }`:

```js
{ type: ['array', 'null'], items: { type: ['string', 'null'] }, enum: ['funny', 'sad', null] }
```

After:

```js
{
  type: ['array', 'null'],
  items: { type: ['string', 'null'], enum: ['funny', 'sad', null] }
}
```

**Tests**

Two tests in `test/schema.test.js`, both failing before the change:

- `puts enums on array elements rather than on the array (gh-16443)` — checks both `useBsonType`
  modes, creates a real collection with the generated `$jsonSchema` validator and asserts the
  server now accepts `{ tags: ['funny'] }` and still rejects an out-of-enum element, and runs the
  same cases through Ajv.
- `supports enums declared as an object or set with enum() (gh-16443)` — covers the
  `{ values, message }` form on both an optional and a `required` path, and `schema.path(...).enum()`.

### DIFF
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -4121,6 +4121,97 @@ describe('schema', function() {
       assert.ok(!validate({ _id: 'not-an-objectid' }));
       assert.ok(!validate({ _id: '0'.repeat(24), author: 'not-an-objectid' }));
     });
+
+    it('puts enums on array elements rather than on the array (gh-16443)', async function() {
+      const schema = new Schema({
+        tags: { type: [String], enum: ['funny', 'sad'] },
+        scores: [{ type: Number, enum: [1, 2] }]
+      }, { autoCreate: false, autoIndex: false });
+
+      assert.deepStrictEqual(schema.toJSONSchema({ useBsonType: true }), {
+        required: ['_id'],
+        properties: {
+          tags: {
+            bsonType: ['array', 'null'],
+            items: {
+              bsonType: ['string', 'null'],
+              enum: ['funny', 'sad', null]
+            }
+          },
+          scores: {
+            bsonType: ['array', 'null'],
+            items: {
+              bsonType: ['number', 'null'],
+              enum: [1, 2, null]
+            }
+          },
+          _id: {
+            bsonType: 'objectId'
+          }
+        }
+      });
+
+      await db.createCollection(collectionName, {
+        validator: {
+          $jsonSchema: schema.toJSONSchema({ useBsonType: true })
+        }
+      });
+      const Test = db.model('Test', schema, collectionName);
+
+      const doc = await Test.create({ tags: ['funny'], scores: [1, 2] });
+      assert.deepStrictEqual(doc.toObject().tags, ['funny']);
+
+      await assert.rejects(
+        Test.create([{ tags: ['funny', 'something else'] }], { validateBeforeSave: false }),
+        /MongoServerError: Document failed validation/
+      );
+
+      const ajv = new Ajv();
+      const validate = ajv.compile(schema.toJSONSchema());
+
+      assert.ok(validate({ _id: '0'.repeat(24), tags: ['funny', 'sad'], scores: [1] }));
+      assert.ok(!validate({ _id: '0'.repeat(24), tags: ['funny', 'something else'] }));
+      assert.ok(!validate({ _id: '0'.repeat(24), scores: [3] }));
+    });
+
+    it('supports enums declared as an object or set with enum() (gh-16443)', function() {
+      const schema = new Schema({
+        status: { type: String, enum: { values: ['on', 'off'], message: '{VALUE} is not supported' } },
+        level: { type: Number, required: true, enum: { values: [1, 2], message: 'invalid' } },
+        color: String
+      }, { autoCreate: false, autoIndex: false });
+      schema.path('color').enum('red', 'green');
+
+      assert.deepStrictEqual(schema.toJSONSchema(), {
+        type: 'object',
+        required: ['level', '_id'],
+        properties: {
+          status: {
+            type: ['string', 'null'],
+            enum: ['on', 'off', null]
+          },
+          level: {
+            type: 'number',
+            enum: [1, 2]
+          },
+          color: {
+            type: ['string', 'null'],
+            enum: ['red', 'green', null]
+          },
+          _id: {
+            type: 'string',
+            pattern: '^[A-Fa-f0-9]{24}$'
+          }
+        }
+      });
+
+      const ajv = new Ajv();
+      const validate = ajv.compile(schema.toJSONSchema());
+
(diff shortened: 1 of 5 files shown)


test/schema.test.js
```diff
@@ -4121,6 +4121,97 @@ describe('schema', function() {
...
+      });
+
+      await db.createCollection(collectionName, {
+        validator: {
+          $jsonSchema: schema.toJSONSchema({ useBsonType: true })
+        }
+      });
+      const Test = db.model('Test', schema, collectionName);
+
+      const doc = await Test.create({ tags: ['funny'], scores: [1, 2] });
```
vkarpov15, line 4161: Let's also update this test to confirm null is a valid array entry

vkarpov15 (commented):
One minor comment but looks good overall

Jaybhade:
Done in 566cab41. Covered on both sides of the test:

```js
const withNulls = await Test.create({ tags: [null], scores: [null] });
assert.deepStrictEqual(withNulls.toObject().tags, [null]);
assert.deepStrictEqual(withNulls.toObject().scores, [null]);
```

and for the plain-JSON-Schema form:

```js
assert.ok(validate({ _id: '0'.repeat(24), tags: [null], scores: [null] }));
```

Both are load-bearing — I stripped `null` back out of `items.enum` on the generated schema and re-ran: the server rejects the insert with `Document failed validation` and Ajv returns `false`. Which is really the point of putting the enum on `items`: on the array itself, `enum` never had a chance to describe elements at all.

`test/schema.test.js` is green (222 passing) and `eslint` is clean on the changed files.

vkarpov15 (approved):
Thank you :+1:

Issue #11836, opened 2022-05-26
Global `strictQuery` flag should apply to arrayFilters


postama:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
An error is thrown when using `arrayFilters` that do not match the schema, even if `strictQuery` is set to `false`.
This is similar to https://github.com/Automattic/mongoose/issues/11062 which fixed it on the query-by-query basis, but not with the global flag.

**If the current behavior is a bug, please provide the steps to reproduce.**
Error: Could not find path "items.0.oldProp" in schema when `strictQuery` is set to `false`

```js
const mongoose = require('mongoose')
mongoose.set('strictQuery', false)
const { Schema } = mongoose
const uri = `mongodb://localhost:27017/test`
console.log(`Mongoose version: ${mongoose.version}`)

const ItemSchema = new Schema({
  newProp: {
    type: String
  }
})

//Schema
const FooSchema = new Schema({
  foo: {
    type: String
  },
  items: [ItemSchema]
})

//Model
const Foo = mongoose.model('Foo', FooSchema)

//Test function
async function test() {
  //Connect to database
  await mongoose.connect(uri)
  await mongoose.connection.dropDatabase()

  //Create new item
  const foo = {
    items: [
      {
        oldProp: false
      },
      {
        oldProp: false
      }
    ]
  }
  await Foo.create(foo, { strict: false })

  await Foo.updateMany(
    {},
    {
      $set: { 'items.$[m].newProp': 'test' }
    },
    {
      arrayFilters: [
        {
          'm.oldProp': false
        }
      ]
    }
  )
}

test()
```

**What is the expected behavior?**
This update works if `strict` is set to false on the individual query, but it should respect the global `strictQuery` flag as well, and not throw an error.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
Node: 14.x
Mongoose: 6.3.4
MongoDB: 4.2.6

Issue #11062, opened 2021-12-08
Strict flag possibly ignored when using arrayFilters


adamreisnz:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
We’re running a query to migrate a data structure, which generally works fine as long as we use `strict: false` to have Mongoose ignore unknown (old) schema data.

However, in this case we’re using arrayFilters and it seems Mongoose doesn’t like it when we use paths that are no longer defined on the schema, even with `strict: false`.

This is the query in question:

```
await mongoose
    .model('Foo')
    .updateMany({}, {
      $set: {'items.$[m].newProp': 'bar'},
    }, {
      strict: false,
      arrayFilters: [{
        'm.oldProp': false,
      }],
    })
```

So we’re basically changing any sub doc in the `items` array with `oldProp: false` to `newProp: 'bar'`

However, the query throws the following error:

```
Error: Could not find path "items.0.oldProp" in schema
  at _castArrayFilters (/node_modules/mongoose/lib/helpers/update/castArrayFilters.js:61:15)
  at castArrayFilters (/node_modules/mongoose/lib/helpers/update/castArrayFilters.js:15:3)
  at _castArrayFilters (/node_modules/mongoose/lib/query.js:2069:5)
  at model.Query._updateThunk (/node_modules/mongoose/lib/query.js:3917:3)
  at model.Query.<anonymous> (/node_modules/mongoose/lib/query.js:4030:23)
  at model.Query._wrappedThunk [as _updateMany] (/node_modules/mongoose/lib/helpers/query/wrapThunk.js:27:8)
  at /node_modules/kareem/index.js:279:20
  at _next (/node_modules/kareem/index.js:103:16)
  at /node_modules/kareem/index.js:508:38
  at processTicksAndRejections (node:internal/process/task_queues:78:11)
From previous event:
  at promiseOrCallback (/node_modules/mongoose/lib/helpers/promiseOrCallback.js:31:10)
  at model.Query.exec (/node_modules/mongoose/lib/query.js:4562:10)
  at model.Query.Query.then (/node_modules/mongoose/lib/query.js:4645:15
```

It seems that the strict flag is perhaps not working properly with arrayFilters?
Possibly related to #10696?

**If the current behavior is a bug, please provide the steps to reproduce.**
```js
const mongoose = require('mongoose')
const {Schema} = mongoose
const uri = `mongodb://localhost:27017/test`
[...]

PR #16448, opened 2026-08-08
fix(query): respect `strictQuery` query option when casting array filters

### LINKED ISSUE #16447: strictQuery` passed as a query option is ignored when casting `arrayFilters
### Prerequisites

### Mongoose version

9.9.1

### Node.js version

26.7.0

### MongoDB server version

 7.0.18

_No response_

### Description

strictQuery is respected for array filters when set on the schema or globally, and query-level strict is respected too, but query-level strictQuery is silently ignored. Passing it to updateOne() or setOptions() still throws Could not find path ... in schema.

lib/helpers/update/castArrayFilters.js reads query._mongooseOptions.strictQuery, but nothing assigns that key for a normal query — setOptions() copies options into _mongooseOptions by explicit name and strictQuery isn't among them, while strict is (lib/query.js:4559). Filter casting reads query.options.strictQuery instead (lib/cast.js:423, fed by lib/query.js:5224), which is why the same option works there.

Related: #11062 fixed query-level strict for array filters, #11836 fixed the global strictQuery flag. Query-level strictQuery looks like the remaining case.

### Steps to Reproduce

```
'use strict';

const itemSchema = new mongoose.Schema({ name: String }, { _id: false });
const update = { $set: { 'items.$[item].name': 'updated' } };
const arrayFilters = [{ 'item._id': 'item-1' }];

async function seed() {
  const coll = mongoose.connection.collection('things');
  await coll.deleteMany({});
  // Inserted with the driver: `_id: false` means Mongoose would strip these _ids.
  await coll.insertOne({
    tag: 't',
    items: [{ _id: 'item-1', name: 'original' }, { _id: 'item-2', name: 'keep' }]
  });
}

async function attempt(label, model, options) {
  await seed();
  try {
    const res = await model.updateOne({ tag: 't' }, update, options);
    console.log(label.padEnd(32), 'OK    modifiedCount=' + res.modifiedCount);
  } catch (err) {
    console.log(label.padEnd(32), 'THROW ' + err.message);
  }
}

async function main() {

  const fields = { tag: String, items: [itemSchema] };
  const Loose = mongoose.model('Loose', new mongoose.Schema(fields, { strictQuery: false }), 'things');
  const Default = mongoose.model('Default', new mongoose.Schema(fields), 'things');

  await attempt('schema { strictQuery: false }', Loose, { arrayFilters });
  await attempt('query { strict: false }', Default, { arrayFilters, strict: false });
  await attempt('query { strictQuery: false }', Default, { arrayFilters, strictQuery: false });

  await seed();
  try {
    const res = await Default.updateOne({ tag: 't' }, update, { arrayFilters })
      .setOptions({ strictQuery: false });
    console.log('setOptions({ strictQuery: false })', 'OK    modifiedCount=' + res.modifiedCount);
  } catch (err) {
    console.log('setOptions({ strictQuery: false })', 'THROW ' + err.message);
  }

  await mongoose.disconnect();
}

main();
```

Output on 9.9.1:

schema { strictQuery: false }    OK    modifiedCount=1
query { strict: false }          OK    modifiedCount=1
query { strictQuery: false }     THROW Could not find path "items.0._id" in schema
setOptions({ strictQuery: false }) THROW Could not find path "items.0._id" in schema

### Expected Behavior

The last two lines should behave like the first two — OK modifiedCount=1. Query-level strictQuery should override the schema and global settings for array filters, the same way query-level strict already does and the same way strictQuery already does for query filters.

vrana055:
**Summary**

Fixes #16447.

`castArrayFilters()` resolves `strictQuery` from `query._mongooseOptions.strictQuery`, but nothing ever assigns that key for a normal query: `Query.prototype.setOptions()` copies options into `_mongooseOptions` by explicit name and `strictQuery` isn't among them (`strict` is). A `strictQuery` passed to `updateOne()` or `setOptions()` lands on `query.options` instead, so that branch never matched and query-level `strictQuery` was silently ignored for array filters.

That left an inconsistency: schema-level `strictQuery`, global `strictQuery`, and query-level `strict` were all respected for array filters, but query-level `strictQuery` was not — even though query filter casting already reads `query.options.strictQuery` (`lib/cast.js`, fed by `lib/query.js`).

The fix reads `query.options.strictQuery` as well, matching how `_castUpdate()` already resolves the option. `query.options.strictQuery` is `undefined` unless the caller passes it explicitly, so the default stays unchanged and array filters remain strict by default.

This is the remaining case after #11062 (query-level `strict` for array filters) and #11836 (global `strictQuery`).

**Examples**

Before, both of these threw `Could not find path "items.0._id" in schema`:

```javascript
const itemSchema = new Schema({ name: String }, { _id: false });
const Test = mongoose.model('Test', new Schema({ tag: String, items: [itemSchema] }));

const update = { $set: { 'items.$[item].name': 'updated' } };
const arrayFilters = [{ 'item._id': 'item-1' }];

// Now respects the option and sends the filter uncast, like `strict: false` already did
await Test.updateOne({ tag: 't' }, update, { arrayFilters, strictQuery: false });

// Same via setOptions()
await Test.updateOne({ tag: 't' }, update, { arrayFilters }).setOptions({ strictQuery: false });
```

**Also in this PR**

`docs/guide.md` claimed `strictQuery` is "**just** for query filters". Added a `strictQuery and arrayFilters` section documenting that `strictQuery` applies to array filters too, and that array filters are strict by default even though `strictQuery` defaults to `false` — Mongoose throws rather than silently stripping an array filter, because dropping one would change which array elements the update modifies. That behaviour was previously undocumented.

**Tests**

Added `respects 'strictQuery' passed as a query option (gh-16447)` to `test/helpers/update.castArrayFilters.test.js`, covering the default throw, `strictQuery: false` as a query option, and the same via `setOptions()`.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -957,7 +957,9 @@ MyModel.find({ notInSchema: 1 });
 ```
 
 The `strict` option does apply to updates.
-The `strictQuery` option is **just** for query filters.
+<!-- markdownlint-disable MD051 -->
+The `strictQuery` option is for query filters and for [`arrayFilters`](#strictQuery-arrayFilters).
+<!-- markdownlint-enable MD051 -->
 
 ```javascript
 // Mongoose will strip out `notInSchema` from the update if `strict` is
@@ -993,6 +995,37 @@ await MyModel.find({ notInSchema: 1 });
 // Matches _all_ documents with `strictQuery: true`: Mongoose strips out
 // `notInSchema: 1`, leaving an empty filter `{}`.
 await MyModel.find({ notInSchema: 1 }).setOptions({ strictQuery: true });
+```
+
+### strictQuery and arrayFilters {#strictQuery-arrayFilters}
+
+Mongoose also checks `arrayFilters` paths against your schema.
+Array filters are strict by default, even though `strictQuery` is `false` by default.
+Mongoose never silently strips an array filter, because dropping one would change which array elements the update modifies, so an array filter path that isn't in the schema throws instead.
+
+```javascript
+const itemSchema = new Schema({ name: String }, { _id: false });
+const MyModel = mongoose.model('Test', new Schema({ items: [itemSchema] }));
+
+// Throws 'Could not find path "items.0._id" in schema', because `itemSchema` is
+// declared with `_id: false`, so `items._id` is not in the schema.
+await MyModel.updateOne(
+  {},
+  { $set: { 'items.$[item].name': 'test' } },
+  { arrayFilters: [{ 'item._id': 42 }] }
+);
+```
+
+Set `strictQuery: false` to leave the path as-is and send it to the server uncast.
+That is useful when your stored documents contain properties that your schema doesn't declare.
+Mongoose throws for array filters with both `strictQuery: true` and `strictQuery: 'throw'`.
+
+```javascript
+await MyModel.updateOne(
+  {},
+  { $set: { 'items.$[item].name': 'test' } },
+  { arrayFilters: [{ 'item._id': 42 }], strictQuery: false }
+);
 ```
 
 In general, we do **not** recommend passing user-defined objects as query filters:
--- a/lib/helpers/update/castArrayFilters.js
+++ b/lib/helpers/update/castArrayFilters.js
@@ -28,6 +28,13 @@ module.exports = function castArrayFilters(query) {
   if (query._mongooseOptions.strictQuery != null) {
     strictQuery = query._mongooseOptions.strictQuery;
   }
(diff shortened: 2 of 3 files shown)


vkarpov15:
@vrana055 I put in an alternative fix in #16451 WDYT?

vrana055:
> @vrana055 I put in an alternative fix in #16451 WDYT?

@vkarpov15  Agreed, #16451 is the better fix — it's root cause rather than a patch on top of castArrayFilters.js.
Let me close #16448 in favor of #16451
Thanks.

PR #16451, opened 2026-08-14
Set strictQuery and strict on _mongooseOptions consistently

vkarpov15:
**Summary**

#16447 pointed out that, because `strictQuery` isn't applied onto `_mongooseOptions`, it doesn't apply to array filters. We should make sure `strict` and `strictQuery` consistently end up on `_mongooseOptions` rather than `options` in order to be consistent with our documentation. The `mongooseOptions()` function explicitly lists `strict` and `strictQuery` as Mongoose internal options.

**Examples**


hasezoey (commented):
LGTM

PR #16452, opened 2026-08-17
fix: apply defaults to undefined array elements

vkarpov15:
Re: #9232

**Summary**

Currently defaults are not applied to array elements: `tags: [{ type: String, default: 'Unknown' }]` the default is silently ignored. This PR applies defaults to arrays.

Putting this into 9.10 for caution.

**Examples**


hasezoey (commented):
LGTM

AbdelrahmanHafez (approved):
Nice! Other than Copilot's comment, LGTM.

PR #16453, opened 2026-08-17
fix(hydration): use the doc model for looking up a ref

rawmind:
**Summary**

The hydration doesn't work for embedded docs
Reproduced on `9.8.0` but I think it affects all the existing versions, including 7.x, 8.x, and 9.x


rawmind:
@vkarpov15 could I get pipeline approval for this?

rawmind:
@vkarpov15 ty, all tests are green. Please provide a review when you have a chance

vkarpov15 (approved):
Good find! The issue is that `doc` may be a subdocument here, and subdocuments don't have a `db` property - there's no reason why a subdocument would ever have a different `db` than its parent. So we need to get the owner document's `db` handle. Thankfully we already calculate the `ownerDocument` so there's no additional performance hit.

Issue #8235, opened 2019-10-10
eachAsync: Cursor already in use


simllll:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
Since the last update of mongoose I experience a lot of "Cursor already in use" in long running queries where a cursor is used in combination with eachAsync().

**If the current behavior is a bug, please provide the steps to reproduce.**
Run a long runnign query and use eachASync to iterate over it. It's not 100% reproducable, but it happens significantely often.

**What is the expected behavior?**
No cursor already in use error.

**What are the versions of Node.js, Mongoose and MongoDB you are using? Note that "latest" is not a version.**
* tested with v12.6.0 and v10.16.3 (LTS).
* mongoose: v5.7.3
* mongodb: 4.0.10

simllll:
```
const stream = DbModel.find({
				createdAt: { $lt: new Date() },
			})
				.lean()
				.cursor({ batchSize: 1500 });

await stream.eachAsync(myObject => asyncDataProcessing());
```

simllll:
One more thing: It seems sometimes if there is no error, the script just finishes earlier without all processed data sets. Sometimes a lot (e.g. supposed to iterate over 3500 sets, finished only around 700).

simllll:
Backtrace:

```
MongoError: Cursor already in use (namespace: 'aDatabase.aCollection', id: 5482770079154774044).
  File "/var/app/current/node_modules/mongoose/node_modules/mongodb/lib/core/connection/pool.js", line 466, col 61, in Connection.<anonymous>
    return handleOperationCallback(self, workItem.cb, new MongoError(responseDoc));
  File "events.js", line 198, col 13, in Connection.emit
  File "domain.js", line 448, col 20, in Connection.EventEmitter.emit
  File "/var/app/current/node_modules/mongoose/node_modules/mongodb/lib/core/connection/connection.js", line 364, col 10, in processMessage
    conn.emit(
  File "/var/app/current/node_modules/mongoose/node_modules/mongodb/lib/core/connection/connection.js", line 566, col 15, in Socket.<anonymous>
    processMessage(conn, emitBuffer);
  File "events.js", line 198, col 13, in Socket.emit
  File "domain.js", line 448, col 20, in Socket.EventEmitter.emit
  File "_stream_readable.js", line 288, col 12, in addChunk
  File "_stream_readable.js", line 269, col 11, in readableAddChunk
[...]

PR #16456, opened 2026-08-19
fix(schema): do not repeat null in a toJSONSchema enum that already lists null

luantaraschi:
**Summary**

`toJSONSchema()` appends `null` to a path's `enum` whenever the path allows null, without checking what the enum already holds. A schema that lists `null` itself ends up with the value twice, and both consumers of the output refuse the result:

- MongoDB: `$jsonSchema keyword 'enum' array cannot contain duplicate values.`
- Ajv: `enum must NOT have duplicate items (items ## 2 and 3 are identical)`

So the collection cannot be created at all, and the schema cannot be compiled either. It happens on String and Number paths, on both `enum` forms (array and object), and on the array element enums from #16443, with and without `useBsonType`. Paths declared `required: true` are not affected, because those never receive the extra `null`.

The fix adds `null` only when the enum does not carry it already.

**Examples**

```js

  status: { type: String, enum: ['active', 'inactive', null] }
});

console.log(schema.toJSONSchema({ useBsonType: true }).properties.status);

await mongoose.connect('mongodb://127.0.0.1:27017/reprodb');
await mongoose.connection.db.createCollection('tests', {
  validator: { $jsonSchema: schema.toJSONSchema({ useBsonType: true }) }
});
```

On master:

```
{
  bsonType: [ 'string', 'null' ],
  enum: [ 'active', 'inactive', null, null ]
}
MongoServerError: $jsonSchema keyword 'enum' array cannot contain duplicate values.
```

With this change:

```
{
  bsonType: [ 'string', 'null' ],
  enum: [ 'active', 'inactive', null ]
}
createCollection: ok
```

The test goes next to the other `jsonSchema()` cases in `test/schema.test.js` and follows the same shape: it compares both outputs, creates the collection with the generated validator, writes a document through it, and compiles the plain JSON Schema with Ajv. It fails on master and passes here.

`npm test` on this branch: 4502 passing, 47 pending, 1 failing. The failure is `eachAsync() exhausts large cursor without parallel calls (gh-8235)`, a 10s timeout that reproduces the same way on master on this machine.

### DIFF
--- a/lib/schemaType.js
+++ b/lib/schemaType.js
@@ -230,8 +230,12 @@ SchemaType.prototype._addJSONSchemaEnum = function _addJSONSchemaEnum(definition
   }
 
   // The enum validator allows nullish values, so allow `null` wherever the type does.
+  // Skip that when the user already listed `null`: MongoDB refuses a `$jsonSchema`
+  // whose `enum` repeats a value, and so does Ajv.
   const allowsNull = Array.isArray(definition.type ?? definition.bsonType);
-  definition.enum = allowsNull ? [...this.enumValues, null] : [...this.enumValues];
+  definition.enum = allowsNull && !this.enumValues.includes(null)
+    ? [...this.enumValues, null]
+    : [...this.enumValues];
 
   return definition;
 };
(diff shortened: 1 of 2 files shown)


AbdelrahmanHafez (approved):
LGTM, thanks!

PR #16458, opened 2026-08-19
docs: fix dead MongoDB driver Collection link

rajanpanth:
The README's "native MongoDB driver collection" link points at `http://mongodb.github.io/node-mongodb-native/2.1/api/Collection.html`, which 404s — the driver docs moved off the `/<version>/api/` layout years ago, and 2.1 is long gone.

Updated to [`7.5/classes/Collection.html`](https://mongodb.github.io/node-mongodb-native/7.5/classes/Collection.html), which matches the `"mongodb": "~7.5"` dependency in `package.json`, so the docs a reader lands on describe the driver Mongoose actually bundles. Also switched to https.

Happy to use `/Next/` instead if you'd rather the link float with the newest driver docs rather than track the pinned version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,7 +340,7 @@ new Schema({
 
 ### Driver Access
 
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Thanks :+1:

PR #16459, opened 2026-08-19
fix(cast): reject arrays when casting to Date

official-burak:
**Summary**

`castDate` falls through to `new Date(value.valueOf())` for non-number, non-string inputs. Arrays implement `valueOf()` by returning themselves, and `new Date(['2024-01-15'])` stringifies the array then parses that string, so a single date-string (or other Date-parseable) element silently becomes a Date instead of a CastError:

```js
const schema = new Schema({ x: Date });
const Model = mongoose.model('T', schema);

new Model({ x: ['2024-01-15'] }); // x is Mon Jan 15 2024 -- no CastError
```

This is the same class of bug recently fixed for Int32 and Double in #16446. `castNumber()` already guards with `Array.isArray()`, and empty/multi-element arrays already failed for Date (`[]` and `[2024, 0, 15]` produce Invalid Date). Only the Date-parseable single-element case was silently wrong. Objects with a custom `valueOf()` (moment.js, etc.) are unchanged.

**Examples**

Covered by a new test in `test/schema.date.test.js` (`['2024-01-15']`, `['January 1, 2024']`, `[5]`, `[]`, `[2024, 0, 15]`). The test fails on current `master` (`actual` is `2024-01-15T00:00:00.000Z`, `expected` is `undefined`) and passes with this change. `npx mocha --exit ./test/schema.date.test.js` is 7/7; eslint is clean on the touched files.

### DIFF
--- a/lib/cast/date.js
+++ b/lib/cast/date.js
@@ -24,6 +24,12 @@ module.exports = function castDate(value) {
   } else if (typeof value === 'string' && !isNaN(Number(value)) && (Number(value) >= 275761 || Number(value) < -271820)) {
     // string representation of milliseconds take this path
     date = new Date(Number(value));
+  } else if (Array.isArray(value)) {
+    // `new Date(['2024-01-15'])` stringifies the array and Date parses that
+    // string, so a single date-string (or other Date-parseable) element would
+    // silently coerce instead of throwing. Same class of bug as the Int32 and
+    // Double array guards.
+    assert.ok(false);
   } else if (typeof value.valueOf === 'function') {
     // support for moment.js. This is also the path strings will take because
     // strings have a `valueOf()`
--- a/test/schema.date.test.js
+++ b/test/schema.date.test.js
@@ -44,6 +44,18 @@ describe('SchemaDate', function() {
     assert.equal(doc.x.valueOf(), mockDate);
   });
 
+  it('throws a CastError when an array is provided to a Date field', async function() {
+    for (const value of [['2024-01-15'], ['January 1, 2024'], [5], [], [2024, 0, 15]]) {
+      const doc = new M({ x: value });
+
+      assert.strictEqual(doc.x, undefined);
+      const err = await doc.validate().catch(e => e);
+      assert.ok(err);
+      assert.ok(err.errors['x']);
+      assert.equal(err.errors['x'].name, 'CastError');
+    }
+  });
+
   it('casts string representation of unix timestamps (gh-6443)', function() {
     const timestamp = Date.now();
     const stringTimestamp = timestamp.toString();


AbdelrahmanHafez (approved):

vkarpov15 (commented):
I'm less concerned about this one, and honestly I'm rethinking the Int32 and Double change because of this.

Automatically unboxing a single element array `['2024-01-15']` so it casts the same as `'2024-01-15'` seems like a reasonable convenience consistent with the rest of Mongoose casting logic. My concern would be casting `[]` or `['2024-01-15', '2026-01-15']` to valid dates. However both of those result in an "Invalid Date" object whose `date.valueOf()` is NaN, so they would fail the check on L36. Single element array though? There's no risk of losing data from additional array entries because there's only one.

@AbdelrahmanHafez @hasezoey WDYT?

AbdelrahmanHafez:
You're right, if it applies to single element arrays only, should be a nice convenience and should be kept as is.
Maybe this PR should be about writing tests that assert the behavior as an explicit design choice instead. WDYT?

vkarpov15:
I'll close this PR for now and we'll allow unboxing single array elements for remaining schema paths in 10.0. The following script shows which scalar types cast array elements - at present only Date, ObjectId, Buffer, and UUID.

```javascript
'use strict';

const { Schema } = mongoose;
const objectId = new mongoose.Types.ObjectId().toString();
const uuid = '6a887010-fdd9-4760-b9e4-4227a887010f';

const schema = new Schema({
  string: String,
  number: Number,
  date: Date,
  boolean: Boolean,
  objectId: Schema.Types.ObjectId,
  decimal128: Schema.Types.Decimal128,
  int32: Schema.Types.Int32,
  double: Schema.Types.Double,
  bigInt: Schema.Types.BigInt,
  buffer: Buffer,
  uuid: Schema.Types.UUID,
  mixed: Schema.Types.Mixed
}, { _id: false });

const values = {
  string: 'foo',
  number: '42',
  date: '2020-01-01',
  boolean: 'true',
  objectId,
  decimal128: '1.23',
  int32: '42',
  double: '1.23',
  bigInt: '42',
  buffer: 'foo',
  uuid,
  mixed: 'foo'
};

function cast(path, value) {
  try {
    return path.cast(value);
  } catch (error) {
    return `ERROR: ${error.message.split('\n')[0]}`;
  }
}

function describe(value) {
  if (value instanceof Date) {
    return `Date(${value.toISOString()})`;
  }
  if (Buffer.isBuffer(value)) {
    return `Buffer(${value.toString('hex')})`;
  }
  if (value instanceof mongoose.Types.ObjectId) {
    return `ObjectId(${value})`;
  }
  if (typeof value === 'bigint') {
    return `${value}n`;
  }
  return JSON.stringify(value);
}

console.log(`Mongoose ${mongoose.version}`);

schema.eachPath((name, path) => {
  const value = values[name];
  console.log(`\n${name} (${path.instance})`);
  for (const input of [[], [value], [value, value]]) {
    const result = cast(path, input);
    const output = typeof result === 'string' && result.startsWith('ERROR:') ? result : describe(result);
    console.log(`${JSON.stringify(input)} => ${output}`);
  }
});

```

PR #16460, opened 2026-08-19
feat(model): add Model.findAndCount() wrapper for limit+skip based pagination with countDocuments

### LINKED ISSUE #16454: Add Model.findAndCount() for paginated queries
### Prerequisites

### 🚀 Feature Proposal

Add a `Model.findAndCount()` method for offset-based pagination. The method would run a standard `find()` and a `countDocuments()` with the same filter and return both results as a tuple:

```js
const [documents, total] = await Model.findAndCount(filter, projection, options);
```

The proposed behavior is:

- `documents` follows the normal `Model.find()` semantics, including projection, sort, skip, limit, populate, and lean options.
- `total` is the number of documents matching the filter before applying `skip` and `limit`.
- The count ignores result-shaping options that do not affect which documents match, such as projection, sort, populate, and lean.
- Compatible execution options that affect both operations, such as session, collation, hint, read preference, and `maxTimeMS`, are forwarded where supported.
- Normal `find` and `countDocuments` middleware continues to run.
- TypeScript infers the document result in the same way as `Model.find()`, including the effect of `lean` and projection where possible.

This would be an ergonomic wrapper around the existing operations, not a new atomic MongoDB operation. The returned documents and count would therefore have the same consistency guarantees as manually running `find()` and `countDocuments()`. Applications that require snapshot consistency can continue to use the appropriate session or transaction semantics.

This proposal is related to, but not a duplicate of, [#3283](https://github.com/Automattic/mongoose/issues/3283), where the recommended solution for pagination was to execute separate `find()` and `countDocuments()` queries. It is also complementary to the cursor-pagination proposal in [#16410](https://github.com/Automattic/mongoose/issues/16410): `findAndCount()` targets offset-based pagination where an exact total is required.

### Motivation

Returning a paginated list together with the total number of matching documents is a common API requirement. Today, every application needs to repeat the same pattern:

```js
const filter = { status: 'active' };

const [documents, total] = await Promise.all([
  User.find(filter)
    .sort({ createdAt: -1 })
    .skip(40)
    .limit(20)
    .lean(),
  User.countDocuments(filter),
]);
```

When a reusable filter helper builds a Mongoose query incrementally, applications often need to clone that query before executing the paginated find and the count separately:

```js
const query = User
  .find({ status: 'active' })
  .sort({ createdAt: -1 });

const dataQuery = query.clone();
const countQuery = query.clone();

const documents = await dataQuery
  .skip(40)
  .limit(20)
  .lean();
const total = await countQuery.countDocuments();
```

This is straightforward, but it has several recurring problems:

- The filter and compatible execution options must be kept in sync between two queries.
- Reusable pagination helpers need to reproduce Mongoose's TypeScript inference and query-option behavior.
- Sessions, collation, hints, middleware, discriminators, and future query features can be forwarded inconsistently by userland helpers.
- Projects often add a pagination plugin solely to provide this small abstraction.

MikroORM exposes the same use case as `findAndCount()`: the list respects `limit` and `offset`, while the returned count represents all matching entities before pagination. Providing this primitive in Mongoose would remove repeated boilerplate while preserving the existing `find()` and `countDocuments()` behavior.

### Example

```js
const page = 3;
const pageSize = 20;

const [users, total] = await User.findAndCount(
  { status: 'active' },
  { name: 1, email: 1, createdAt: 1 },
  {
    sort: { createdAt: -1 },
    skip: (page - 1) * pageSize,
    limit: pageSize,
    lean: true,
  }
);

res.json({
  users,
  total,
  page,
  pageSize,
  pageCount: Math.ceil(total / pageSize),
});
```

For example, if 135 users match the filter, this returns at most 20 users for page 3 and `total === 135`.

vkarpov15:
Fix #16454

**Summary**

Add `await Model.findAndCount()` as a convenience wrapper around calling `find()` and `countDocuments()` in parallel with the same filter for pagination.

**Examples**

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -2134,6 +2134,48 @@ Model.find = function find(conditions, projection, options) {
 };
 
 /**
+ * Finds documents and counts the number of documents matching the filter.
+ *
+ * The count ignores options that only affect the shape of the find results,
+ * including projection, sort, skip, limit, populate, and lean.
+ *
+ * #### Example:
+ *
+ *     const [adventures, total] = await Adventure.findAndCount(
+ *       { type: 'jungle' },
+ *       null,
+ *       { sort: { name: 1 }, skip: 10, limit: 10 }
+ *     );
+ *
+ * @param {object|ObjectId} filter
+ * @param {object|string|string[]} [projection] optional fields to return
+ * @param {object} [options] optional query options
+ * @return {Promise<Array>} a promise that resolves to `[documents, total]`
+ * @api public
+ */
+
+Model.findAndCount = async function findAndCount(conditions, projection, options) {
+  _checkContext(this, 'findAndCount');
+
+  const findQuery = this.find(conditions, projection, options);
+  const countQuery = findQuery.clone().countDocuments();
+
+  // countDocuments() only uses the filter and execution options. Clear the
+  // find query's result-shaping state from the clone before executing it.
+  countQuery._fields = null;
+  countQuery._mongooseOptions = { ...countQuery._mongooseOptions };
+  delete countQuery._mongooseOptions.populate;
+  delete countQuery._mongooseOptions.lean;
+  delete countQuery.options.projection;
+  delete countQuery.options.fields;
+  delete countQuery.options.sort;
+  delete countQuery.options.skip;
+  delete countQuery.options.limit;
+
+  return Promise.all([findQuery.exec(), countQuery.exec()]);
+};
+
+/**
  * Finds a single document by its _id field. `findById(id)` is equivalent to `findOne({ _id: id })`.
  *
  * The `id` is cast based on the Schema before sending the command.
--- a/lib/model.js
+++ b/lib/model.js
@@ -2146,6 +2146,9 @@ Model.find = function find(conditions, projection, options) {
  *       { sort: { name: 1 }, skip: 10, limit: 10 }
  *     );
  *
+ *     adventures; // Array of at most 10 documents matching type 'jungle'
+ *     total; // Number of documents matching type 'jungle'
+ *
  * @param {object|ObjectId} filter
  * @param {object|string|string[]} [projection] optional fields to return
  * @param {object} [options] optional query options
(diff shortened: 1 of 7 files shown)


lib/model.js
```diff
@@ -2133,6 +2133,48 @@ Model.find = function find(conditions, projection, options) {
...
+  countQuery._mongooseOptions = { ...countQuery._mongooseOptions };
+  delete countQuery._mongooseOptions.populate;
+  delete countQuery._mongooseOptions.lean;
+  delete countQuery.options.projection;
+  delete countQuery.options.fields;
+  delete countQuery.options.sort;
+  delete countQuery.options.skip;
+  delete countQuery.options.limit;
+
+  return Promise.all([findQuery.exec(), countQuery.exec()]);
```
AbdelrahmanHafez: Instead of cleaning up the `findQuery`'s clone, WDYT of explicitly picking the filter and the execution options from the cloned query?

test/types/queries.test.ts
```diff
@@ -79,6 +79,15 @@ Test.find().byName('test').byName('test2').orFail().exec().then(console.log);
 Test.countDocuments({ name: /Test/ }).exec().then((res: number) => console.log(res));
 Test.findOne({ 'docs.id': 42 }).exec().then(console.log);
 
+Test.findAndCount({ name: /Test/ }).then(([docs, count]) => {
```
AbdelrahmanHafez: We should also add some runtime tests for this PR

AbdelrahmanHafez (commented):
A couple of comments, otherwise LGTM.

lib/model.js
```diff
@@ -2133,6 +2133,48 @@ Model.find = function find(conditions, projection, options) {
...
+ * The count ignores options that only affect the shape of the find results,
+ * including projection, sort, skip, limit, populate, and lean.
+ *
+ * #### Example:
+ *
+ *     const [adventures, total] = await Adventure.findAndCount(
+ *       { type: 'jungle' },
+ *       null,
+ *       { sort: { name: 1 }, skip: 10, limit: 10 }
+ *     );
```
hasezoey, line 2147: Like some other examples, this should likely include what `adventures`'s, or at the very least `total`'s value would likely be.

hasezoey (approved):

Issue #4049, opened 2016-04-06
timestamps schema option doesn't work on embedded subdocuments (tested in array)


jeremyml:
The mongoose timestamps schema option feature does not create a createdAt or updatedAt field for subdocuments.  I tested for subdocuments in an array only.  This is the schema feature I am referring to: http://mongoosejs.com/docs/guide.html#timestamps

hymloth:
Yes I can verify that too

vkarpov15:
Will take a look. I'm pretty sure we didn't think of this use case when we implemented timestamps :)

vkarpov15:
@jeremyml @hymloth can you guys provide a code sample that reproduces the issue you're seeing? I added some test cases above to cover this case, because we didn't have coverage for it, and everything seems to work as expected...

hymloth:
```
var childSchema = new mongoose.Schema({
  senderId: {type: String}
},{
    timestamps:true,
  toObject: {
    virtuals: true
  },
  toJSON: {
    virtuals: true
  }
});

var parentSchema = new mongoose.Schema({
  children: [childSchema],
},{
  timestamps:true,
  toObject: {
    virtuals: true
  },
  toJSON: {
    virtuals: true
  }
});

var Parent = mongoose.model('Parent', parentSchema);
var Child = mongoose.model('Child', childSchema);

Parent.find({}).remove()
Child.find({}).remove()

var b1 = new Parent({children:[{senderId:"1"}]})
b1.save(function(err, d){
  Parent.findOne({_id:d._id}).then(function(res){
    console.log(1, res) // contains timestamps
  })
})

var b2 = new Parent()
b2.save(function(err, d){
  Parent.findOneAndUpdate({_id:d._id}, {$push:{children: {senderId:"234"}}}, {new:true, fields:{children: {$slice: -1}}}).exec().then(function(res){
    console.log(2, res) // doesn't contain timestamps (shouldn't?)
  })
})
```

vkarpov15:
Ah ok its an issue with findOneAndUpdate and timestamps, got it.

Jeff-Lewis:
Is this relevant? [drudge/mongoose-timestamp Adding support for mongoose 4.0.1 findOneAndUpdate hooks.](https://github.com/drudge/mongoose-timestamp/commit/aad9a7fe65962a2f11f162083c5a666fdd9ad899)

jeremyml:
[...]

PR #16461, opened 2026-08-20
fix(schema): remove a map that lives under a nested path

luantaraschi:
Follow up to #16457, which covered the flat case. The nested branch of `Schema#remove()` has the same two problems, and one of them throws:

```js
const schema = new Schema({ n: { m: { type: Map, of: Number } } });
schema.remove('n');
// TypeError: Cannot convert undefined or null to object
//     at _deletePath (lib/schema.js:2806:17)
```

A map registers its values under `<path>.$*` in `paths`, so the nested branch walks `['n.m', 'n.m.$*']` and calls `_deletePath` on each. `_deletePath` walks `schema.tree` down the pieces, and by the time it gets to `n.m.$*` the parent `tree.n.m` has already been deleted by the previous iteration, so `branch` is `undefined` and the `delete` throws. `$*` has no branch of its own in the tree in any case.

So `_deletePath` now stops when a piece along the way is missing. Nothing to delete is not an error there.

With that out of the way the second problem shows: the nested branch never filtered `mapPaths`, which is what the flat branch does. Measured after `schema.remove('n')`, before and after:

| | `paths` | `mapPaths` | `pathType('n.m')` |
| --- | --- | --- | --- |
| before | `['_id']` | `['n.m.$*']` | `real` |
| after | `['_id']` | `[]` | `adhocOrUndefined` |

So the schema kept answering for a path it had just removed, which is the same thing #16457 described, only reached through a nested parent. Both branches now do the same cleanup.

Test in the `remove()` block. It fails on `master` on the first assertion, since `remove` throws before the rest runs.

`test/schema.test.js` passes 225. Full `npm test` here matches master, the only failure being an `eachAsync` timeout that reproduces on a clean checkout on this machine. `eslint` is clean.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -2764,9 +2764,6 @@ Schema.prototype.remove = function(path) {
           }
         }
 
-        // Maps under this branch register their values in `mapPaths` as well as in
-        // `paths`. Left behind, they keep answering `pathType()` and casting queries
-        // for a path the schema no longer has.
         this.mapPaths = this.mapPaths.filter(
           schemaType => !schemaType.path.startsWith(name + '.')
         );
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -2764,6 +2764,13 @@ Schema.prototype.remove = function(path) {
           }
         }
 
+        // Maps under this branch register their values in `mapPaths` as well as in
+        // `paths`. Left behind, they keep answering `pathType()` and casting queries
+        // for a path the schema no longer has.
+        this.mapPaths = this.mapPaths.filter(
+          schemaType => !schemaType.path.startsWith(name + '.')
+        );
+
         delete this.nested[name];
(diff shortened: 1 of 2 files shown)


luantaraschi:
The failing leg is `timestamps > nested docs with push (gh-4049)`. It records `Date.now()`, creates a document, pushes a second subdocument, saves, reloads, and then asserts `createdAt.getTime() > now`, strictly greater, so it depends on the clock moving between those two points. The sibling test right above it does the same thing with `>=`.

This branch only touches `Schema.prototype.remove` and the `_deletePath` helper, which is called from nowhere else, so it does not reach the timestamps path. I ran that test eight times on the branch against a local mongod and it passed every time.

vkarpov15 (approved):
Thanks :+1:

PR #16463, opened 2026-08-20
fix(schema): clone a document array element with its own constructor signature

### LINKED ISSUE #16462: schema.clone() throws for an array of document arrays
### Prerequisites

### Mongoose version

9.9.3 (also on 8.x, the code is the same)

### Node.js version

22.14.0

### MongoDB server version

Not needed, this happens during schema construction

### Description

`schema.clone()` throws for a schema that holds an array of document arrays. The schema itself builds fine, casts, applies defaults and validates, so the shape is supported everywhere except in `clone()`.

```
MongooseError: Cannot create DocumentArrayElement schematype without a parent
    at new SchemaDocumentArrayElement (lib/schema/documentArrayElement.js:30:11)
    at SchemaType.clone (lib/schemaType.js:1875:22)
    at SchemaDocumentArrayElement.clone (lib/schema/documentArrayElement.js:123:42)
    at clone (lib/helpers/clone.js:138:16)
    at cloneObject (lib/helpers/clone.js:186:17)
    at Schema._clone (lib/schema.js:453:16)
```

### Steps to Reproduce

```js
const { Schema } = mongoose;

const schema = new Schema({ a: [[{ x: Number }]] });

// the shape works
const M = mongoose.model('T', schema);
const doc = new M({ a: [[{ x: 1 }, { x: 2 }], [{ x: 3 }]] });
console.log(doc.toObject().a);   // both rows cast, each item gets an _id
console.log(doc.validateSync()); // undefined

schema.clone();                  // throws
```

Affected shapes, all of which put a `SchemaDocumentArrayElement` in `schema.subpaths`:

```js
new Schema({ a: [[{ x: Number }]] }).clone();            // throws, subpath a.$.$
new Schema({ o: [{ a: [[{ x: Number }]] }] }).clone();   // throws, subpath o.a.$.$
new Schema({ a: [[[{ x: Number }]]] }).clone();          // throws, subpath a.$.$.$
```

An array of a primitive array is fine (`{ g: [[Number]] }`), and so is a plain document array (`{ b: [{ x: Number }] }`), because neither leaves a `SchemaDocumentArrayElement` in `subpaths`.

### Expected Behavior

`clone()` returns a copy, as it does for every other schema shape.

### What is actually going on

`SchemaDocumentArrayElement.prototype.clone()` delegates to `SchemaType.prototype.clone()`, which constructs with

```js
new this.constructor(this.path, options, this.instance, this.parentSchema);
```

That signature is `(path, options, instance, parentSchema)`, but this schematype's is `(path, schema, options, parentSchema)`. So `options` lands in the `schema` parameter and the string `'DocumentArrayElement'` lands in `options`, and the constructor's `options?.$parentSchemaType` reads it off a string and finds nothing.

It is broken for every document array, not just the nested ones:

```js
new Schema({ b: [{ x: Number }] }).path('b').embeddedSchemaType.clone();
// MongooseError: Cannot create DocumentArrayElement schematype without a parent
```

A plain document array simply never reaches it, since `Schema.prototype._clone` clones the `b` path through `SchemaDocumentArray.prototype.clone()` and the element goes along with the subdocument schema. It is only when the element sits in `subpaths` that `lib/helpers/clone.js` calls `.clone()` on it directly.

The two sibling schematypes with an unusual constructor signature, `SchemaSubdocument` and `SchemaDocumentArray`, both construct directly in their `clone()` instead of delegating. I will send a PR that does the same here.

luantaraschi:
Fixes #16462.

`SchemaDocumentArrayElement.prototype.clone()` delegated the construction to `SchemaType.prototype.clone()`:

```js
new this.constructor(this.path, options, this.instance, this.parentSchema);
```

That signature is `(path, options, instance, parentSchema)`, but this schematype's is `(path, schema, options, parentSchema)`. So `options` went into the `schema` parameter and the string `'DocumentArrayElement'` into `options`, and the constructor read `options?.$parentSchemaType` off a string and threw. Putting `$parentSchemaType` back on `this.options` first, as the old code did, could not help: the argument it was riding on never reached the parameter that gets read.

It was broken for every document array, not only the nested case:

```js
new Schema({ b: [{ x: Number }] }).path('b').embeddedSchemaType.clone();
// MongooseError: Cannot create DocumentArrayElement schematype without a parent
```

A plain document array never reaches it. `Schema.prototype._clone` clones the `b` path through `SchemaDocumentArray.prototype.clone()` and the element rides along with the subdocument schema. The element is only cloned on its own when it sits in `subpaths`, which happens for an array of document arrays:

```js
new Schema({ a: [[{ x: Number }]] }).subpaths;
// { 'a.$': SchemaDocumentArray, 'a.$.$': SchemaDocumentArrayElement }
```

so `s.subpaths = clone(this.subpaths)` in `_clone` reaches `lib/helpers/clone.js`, which calls `.clone()` on it directly, and `schema.clone()` throws. `{ o: [{ a: [[{ x: Number }]] }] }` and `{ a: [[[{ x: Number }]]] }` throw for the same reason. An array of a primitive array is fine, since `g.$.$` is a `SchemaNumber`.

The shape itself is supported. It casts, gives each item an `_id`, applies defaults and validates. Only `clone()` was out of step.

## The change

Construct directly, which is what the two sibling schematypes with an unusual signature already do: `SchemaSubdocument.prototype.clone()` and `SchemaDocumentArray.prototype.clone()`. `$parentSchemaType` and `Constructor` go in through `options`, where the constructor expects them, and `schema` goes into its own parameter, so the copy no longer needs the two assignments that were patching up the result afterwards.

Nothing in the old path is lost. It never produced a working copy to preserve behavior from.

## Tests

Two in the `clone()` group of `test/schema.test.js`. The first is the throw itself plus the subpaths of the copy, the second checks the copy is a distinct object that kept the subdocument schema, the `Constructor` and the parent link, and that a model built from the copy still applies the defaults declared inside. Both fail on `master` with the error above.

Ran locally: `schema*.test.js`, `types.array.test.js`, `types.documentarray.test.js` and `document.test.js`, 1239 passing, plus `model.test.js`, 396 passing and 15 pending.

### DIFF
--- a/lib/schema/documentArrayElement.js
+++ b/lib/schema/documentArrayElement.js
@@ -119,14 +119,26 @@ SchemaDocumentArrayElement.prototype.doValidate = async function doValidate(valu
  */
 
 SchemaDocumentArrayElement.prototype.clone = function() {
-  this.options.$parentSchemaType = this.$parentSchemaType;
-  const ret = SchemaType.prototype.clone.apply(this, arguments);
-  delete this.options.$parentSchemaType;
+  // This schematype takes the subdocument schema where `SchemaType` takes the
+  // options, so it cannot go through `SchemaType.prototype.clone()`: the
+  // arguments would land in the wrong parameters and `$parentSchemaType` would
+  // never reach the constructor.
+  const options = Object.assign({}, this.options, {
+    $parentSchemaType: this.$parentSchemaType,
+    Constructor: this.Constructor
+  });
+  const schematype = new this.constructor(
+    this.path,
+    this.schema,
+    options,
+    this.parentSchema
+  );
+  schematype.validators = this.validators.slice();
+  if (this.requiredValidator !== undefined) {
+    schematype.requiredValidator = this.requiredValidator;
+  }
 
-  ret.Constructor = this.Constructor;
-  ret.schema = this.schema;
-
-  return ret;
+  return schematype;
 };
 
 /*!
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -1832,6 +1832,33 @@ describe('schema', function() {
     });
 
     describe('clone()', function() {
+      it('works with an array of document arrays (gh-16462)', function() {
+        const schema = new Schema({ a: [[{ x: Number }]] });
+
+        const clone = schema.clone();
+        assert.deepStrictEqual(Object.keys(clone.paths).sort(), ['_id', 'a']);
+        assert.deepStrictEqual(Object.keys(clone.subpaths).sort(), ['a.$', 'a.$.$']);
+        assert.equal(clone.subpaths['a.$.$'].instance, 'DocumentArrayElement');
+        assert.ok(clone.subpaths['a.$.$'].$parentSchemaType);
+      });
+
+      it('carries the document array element over to the copy (gh-16462)', function() {
+        const schema = new Schema({ a: [[{ x: { type: Number, default: 3 } }]] });
+
+        const clone = schema.clone();
+        const original = schema.subpaths['a.$.$'];
+        const copy = clone.subpaths['a.$.$'];
+        assert.notStrictEqual(copy, original);
+        assert.strictEqual(copy.schema, original.schema);
+        assert.strictEqual(copy.Constructor, original.Constructor);
+        assert.ok(copy.$parentSchemaType);
+
+        const M = mongoose.model('gh16462', clone);
+        const doc = new M({ a: [[{}, { x: 7 }]] });
+        assert.equal(doc.a[0][0].x, 3);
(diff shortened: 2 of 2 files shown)


vkarpov15 (approved):
Correct, the issue is that the SchemaDocumentArrayElement constructor pulls $parentSchemaType from options.$parentSchemaType, and `options` is 3rd argument to SchemaDocumentArrayElement constructor. However, lib/schemaType.js `clone()` relies on SchemaType's constructor signature, where `options` is the 2nd argument, not the 3rd. So for SchemaDocumentArrayElement we need to skip the base class's `clone()` entirely.

Issue #15966, opened 2026-01-15
Discriminator indexes duplicated when schema has nested paths


Ju-NXH:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Mongoose version

9.1.3

### Node.js version

24.13.0

### MongoDB server version

7.0.28

### Description

When a discriminator schema is created using `Base.discriminator()` and both base and discriminator schemas share the same specs structure containing nested schemas, the discriminator's `_indexes` array gets populated with duplicate copies of the base schema's indexes. The number of duplicates equals the number of nested schema paths in the specs.

### Steps to Reproduce

In this example: 3 base indexes × 3 nested schema paths = 9 indexes in the discriminator schema.

```javascript
const mongoose = require('mongoose');

// Shared nested schema
const NESTED_SCHEMA = new mongoose.Schema({ 
  itemId: mongoose.Schema.Types.ObjectId 
}, { _id: false });

// Specs used by BOTH base and discriminator
const GENERIC_SPECS = {
  parentRef: { type: mongoose.Schema.Types.ObjectId },
  createdAt: Date,
  externalId: String,
  source: NESTED_SCHEMA,
  reference: NESTED_SCHEMA,
  comments: [{ value: String }]  // 3 nested schema paths total
};

function createSchema(collection, specs) {
  return new mongoose.Schema(specs, { collection });
}

async function main() {
  await mongoose.connect('mongodb://localhost:27017/test');

  const baseSchema = createSchema('Item', GENERIC_SPECS);
  baseSchema.index({ parentRef: 1, createdAt: -1 });
  baseSchema.index({ externalId: 1, parentRef: 1 });
  baseSchema.index({ 'source.itemId': 1, parentRef: 1 });

  const Base = mongoose.model('Item', baseSchema);

  const discSchema = createSchema('Item.typeA', GENERIC_SPECS);
  const Child = Base.discriminator('Item.typeA', discSchema, { clone: false });

  console.log('Base _indexes:', Base.schema._indexes.length);        // 3
  console.log('Discriminator _indexes:', discSchema._indexes.length); // 9 (expected: 3)

  // Indexes are tripled: [1,2,3, 1,2,3, 1,2,3]
  discSchema._indexes.forEach((idx, i) => console.log(i, JSON.stringify(idx[0])));

  // This fails because diffIndexes returns 9 duplicate indexes
[...]

Issue #7379, opened 2018-12-28
ensureIndex is called for each discriminator model


arkihillel:
**Bug**

Whenever using discriminators, the index creation is called for the main schema and for every discriminator schemas

Minimal reproduction:
```js
const mongoose = require('mongoose');

mongoose.connect('mongodb://localhost:27017/discriminator', {
	useNewUrlParser: true, 
	useCreateIndex: true
});

mongoose.set('debug', true);

var options = {discriminatorKey: 'kind'};

var eventSchema = new mongoose.Schema({time: {type: Date, index: true}}, options);
var Event = mongoose.model('Event', eventSchema);

var ClickedLinkEvent = Event.discriminator('ClickedLink',
  new mongoose.Schema({url: String}, options));
```

Output
```sh
$ node index.js 
Mongoose: events.createIndex({ time: 1 }, { background: true })
Mongoose: events.createIndex({ time: 1 }, { background: true })
```

**Expected behavior**
`createIndex` is called only from the main schema

This is a pretty minor issue but might be worth taking a look at just in case it hides something else.

Mongoose 5.4.1
NodeJS 10
MongoDB 4

vkarpov15:
It is expected behavior for now because we don't want to diff indexes between the discriminator schema and the base schema, and because calling `createIndex()` multiple times shouldn't cause any errors AFAIK. Will keep this open for the future.

PR #16467, opened 2026-08-21
fix(discriminator): prevent indexes and callQueue duplication with shared nested schemas #15966 (8.x backport)

### LINKED ISSUE #16466: 8.x: base schema indexes duplicated into discriminator schemas and built as partial indexes — hint() silently returns partial results (gh-15966 regression, not backported)
### Prerequisites

### Mongoose version

8.24.1 (regression present in all 8.x versions >= 8.20.0, verified up to 8.24.3)

### Node.js version

22.15.1

### MongoDB server version

8.0.23

### Description

This is the same root cause as #15966, which was fixed on `master` in 9.1.5 via #15974 — but the fix was never backported to the 8.x line, and **8.x is still affected**. We hit this in production after upgrading from 8.14.3 to 8.24.1. Fix PR with the backport and tests: https://github.com/Automattic/mongoose/pull/16467

**What happens on 8.x >= 8.20.0:**

8.20.0 added `SchemaType` cloning to `mergeDiscriminatorSchema()` (lib/helpers/discriminator/mergeDiscriminatorSchema.js). When a discriminator schema is created and its definition shares nested schema objects with the base schema, the merge recurses into document-array/single-nested `SchemaType` instances and follows their `parentSchema` back-references, causing the base schema to be re-merged once per nested path. Each re-merge concatenates the base schema's `_indexes` (and `callQueue`) into the discriminator schema again.

The duplication then cascades into a second, more serious bug specific to index creation:

1. The discriminator schema ends up holding N duplicate copies of every base schema-level index (N = number of nested schema paths).
2. `getIndexes()` decorates `schema._indexes` entries in place via `decorateDiscriminatorIndexOptions()`, adding `partialFilterExpression: { <discriminatorKey>: <value> }`.
3. Because of the added `partialFilterExpression`, the `utils.deepEqual()` skip in `_ensureIndexes()` ("don't create indexes already defined on the base schema", gh-7379) no longer matches, so **every discriminator creates the base schema's indexes as partial indexes**.
4. All of these share the same auto-generated index name (e.g. `customerId_1_orderNumber_1`), so whichever `createIndex` lands first wins the name and the rest fail with `IndexKeySpecsConflict`. In the default autoIndex flow this is a background error that is easy to miss.
5. If the **partial** index wins the race, the full index never exists. Queries that use `.hint()` with that key pattern then **silently return only the documents matching the partial filter** — MongoDB force-uses a hinted partial index even when the query predicate does not imply the `partialFilterExpression`. No error is raised.

In our application (a base model with ~40 discriminators and schema-level indexes used via `hint()` for sorted listing queries), listing endpoints started returning empty/subset results after the upgrade — with the winning `partialFilterExpression` varying between environments because the index-name claim is a race.

Verified by bisection: 8.19.1 builds full indexes and behaves correctly; 8.20.0 through 8.24.3 (latest 8.x) exhibit the bug. 9.1.5+ is fixed.

### Steps to Reproduce

```javascript
'use strict';
const { Schema } = mongoose;

const run = async () => {
  await mongoose.connect('mongodb://localhost:27017/gh_8x_discriminator_indexes');
  await mongoose.connection.db.dropDatabase();

  // Shared schema definition with nested schemas (the trigger, like gh-15966)
  const addressSchema = new Schema({ city: String, street: String }, { _id: false });
  const orderSchemaDefinition = {
    customerId: { type: Schema.Types.ObjectId },
    createdAt: Date,
    orderNumber: String,
    billingAddress: addressSchema,
    shippingAddress: addressSchema,
    notes: [{ message: String }]
  };

  const orderSchema = new Schema(orderSchemaDefinition, { autoIndex: false });
  orderSchema.index({ customerId: 1, orderNumber: 1 });

  const Order = mongoose.model('Order', orderSchema);
  Order.discriminator('WholesaleOrder', new Schema(orderSchemaDefinition, { autoIndex: false }), { clone: false });
  Order.discriminator('RetailOrder', new Schema(orderSchemaDefinition, { autoIndex: false }), { clone: false });

  // Base schema has 1 schema-level index...
  console.log('base _indexes:', JSON.stringify(orderSchema._indexes));
  // ...but the discriminator schema holds 3 duplicated copies (one per nested path)
  console.log('wholesale _indexes:', JSON.stringify(Order.discriminators.WholesaleOrder.schema._indexes));

  try {
    // discriminator first: its partial copy of the base index claims the
    // auto-generated index name; the base model's full index then conflicts
    await Order.discriminators.WholesaleOrder.createIndexes();
    await Order.createIndexes();
  } catch (err) {
    console.log('index build error:', err.message.split('. Requested')[0]);
  }

  const dbIndexes = await Order.collection.indexes();
  console.log('DB indexes:', JSON.stringify(dbIndexes.map(i => ({ name: i.name, key: i.key, partial: i.partialFilterExpression }))));

  await Order.create([
    { customerId: new mongoose.Types.ObjectId(), orderNumber: 'W1' },
    { customerId: new mongoose.Types.ObjectId(), orderNumber: 'R1' }
  ]);
  await Order.collection.updateOne({ orderNumber: 'W1' }, { $set: { __t: 'WholesaleOrder' } });
  await Order.collection.updateOne({ orderNumber: 'R1' }, { $set: { __t: 'RetailOrder' } });

  const hinted = await Order.find({}).sort({ orderNumber: 1 }).hint({ customerId: 1, orderNumber: 1 }).lean();
  console.log('hinted find:', hinted.length, 'docs', hinted.map(d => d.__t));

  await mongoose.connection.db.dropDatabase();
  await mongoose.disconnect();
};

run().catch(err => { console.error(err); process.exit(1); });
```

Actual output on 8.24.1:

```
base _indexes: [[{"customerId":1,"orderNumber":1},{}]]
wholesale _indexes: [[{"customerId":1,"orderNumber":1},{}],[{"customerId":1,"orderNumber":1},{}],[{"customerId":1,"orderNumber":1},{}]]
index build error: An existing index has the same name as the requested index. ...
DB indexes: [{"name":"_id_","key":{"_id":1}},{"name":"customerId_1_orderNumber_1","key":{"customerId":1,"orderNumber":1},"partial":{"__t":"WholesaleOrder"}}]
hinted find: 1 docs [ 'WholesaleOrder' ]   <-- silent partial results
```

### Expected results

- Discriminator schemas should not accumulate copies of the base schema's `_indexes` (or `callQueue`) — `wholesale _indexes` should be `[]`.
- Base schema indexes should be created once, as full (non-partial) indexes.
- The hinted query should return both documents.

Applying the #15974 patch (skip merging `parentSchema`) to 8.x fixes all of the above — verified with the repro above and with the gh-15966 tests from #15974 ported to the 8.x branch (they fail without the fix, pass with it; full `model.discriminator`, `model.indexes`, `schema` suites pass).

Backport PR: https://github.com/Automattic/mongoose/pull/16467

AbinMadathil-Celigo:
Fixes #16466

Backport of #15974 (fix for #15966, shipped in 9.1.5) to the `8.x` branch.

**Summary**

Skip merging `parentSchema` back-references in `mergeDiscriminatorSchema()`. On 8.x (>= 8.20.0), without this skip, creating a discriminator whose schema shares nested schema objects with the base re-merges the base schema once per nested path, duplicating the base schema's `_indexes` and `callQueue` into the discriminator schema.

The `_indexes` duplication is more severe on 8.x than a bookkeeping issue: `getIndexes()` decorates those duplicated entries with `partialFilterExpression` for the discriminator, which defeats the `deepEqual` base-index skip in `_ensureIndexes()` (gh-7379). Every discriminator then creates the base schema's indexes as **partial indexes**, and the auto-generated index-name collision means the full base index may never be created. Queries using `.hint()` with such a key pattern then silently return only the documents of one discriminator value (MongoDB force-uses hinted partial indexes even when the query doesn't match the partial filter). Detailed analysis and repro in #16466.

**Changes**

- `lib/helpers/discriminator/mergeDiscriminatorSchema.js`: skip `parentSchema` key during merge (identical to the master fix).
- `test/model.discriminator.test.js`: ported the three gh-15966 tests from #15974, with expectations adjusted for 8.x (`getIndexes()` still adds `background: true` to stored index options on 8.x).

**Testing**

- The 3 ported tests fail without the fix and pass with it.
- Full suites pass on the 8.x branch with the fix: `model.discriminator` (99), `model.discriminator.querying` + `index` (123), `model.indexes` + `schema` (238).
- Verified end-to-end with the #16466 repro script: unpatched 8.24.1 builds a partial index and a hinted find returns a subset; patched 8.x builds the full index and returns all documents.

### DIFF
--- a/lib/helpers/discriminator/mergeDiscriminatorSchema.js
+++ b/lib/helpers/discriminator/mergeDiscriminatorSchema.js
@@ -46,6 +46,11 @@ module.exports = function mergeDiscriminatorSchema(to, from, path, seen) {
     if (specialProperties.has(key)) {
       continue;
     }
+    if (key === 'parentSchema') {
+      // Skip parentSchema to avoid merging parent schema multiple times
+      // through child paths' back-references (gh-15966)
+      continue;
+    }
     if (to[key] == null) {
       to[key] = from[key];
     } else if (isObject(from[key])) {
--- a/test/model.discriminator.test.js
+++ b/test/model.discriminator.test.js
@@ -2400,4 +2400,123 @@ describe('model', function() {
     assert.equal(documentPreSaveHooks.length, 1);
     assert.equal(documentPreSaveHooks[0], discriminator);
   });
+
+  it('does not duplicate _indexes when base and discriminator schemas share nested schema (gh-15966)', async function() {
+    // Arrange
+    const addressSchema = new Schema({ city: String, street: String }, { _id: false });
+    const orderSchemaDefinition = {
+      customerId: { type: Schema.Types.ObjectId },
+      createdAt: Date,
+      orderNumber: String,
+      billingAddress: addressSchema,
+      shippingAddress: addressSchema,
+      notes: [{ message: String }]
+    };
+
+    const orderSchema = new Schema(orderSchemaDefinition);
+    orderSchema.index({ customerId: 1, createdAt: -1 });
+    orderSchema.index({ orderNumber: 1, customerId: 1 });
+    orderSchema.index({ 'billingAddress.city': 1, customerId: 1 });
+
+    const Order = db.model('Order', orderSchema);
+    const wholesaleOrderSchema = new Schema(orderSchemaDefinition);
+
+    // Act
+    const WholesaleOrder = Order.discriminator('WholesaleOrder', wholesaleOrderSchema, { clone: false });
+    const wholesaleDiff = await WholesaleOrder.diffIndexes();
+    const orderDiff = await Order.diffIndexes();
+
+    // Assert
+    // Note: on 8.x, getIndexes() adds `background: true` to stored index options
+    assert.deepStrictEqual(orderSchema._indexes, [
+      [{ customerId: 1, createdAt: -1 }, { background: true }],
+      [{ orderNumber: 1, customerId: 1 }, { background: true }],
+      [{ 'billingAddress.city': 1, customerId: 1 }, { background: true }]
+    ]);
+    assert.deepStrictEqual(wholesaleOrderSchema._indexes, []);
+
+    assert.deepStrictEqual(orderDiff, {
+      toDrop: [],
+      toCreate: [
+        { customerId: 1, createdAt: -1 },
+        { orderNumber: 1, customerId: 1 },
+        { 'billingAddress.city': 1, customerId: 1 }
+      ]
+    });
+    assert.deepStrictEqual(wholesaleDiff, {
+      toDrop: [],
+      toCreate: []
+    });
+  });
+
+  it('preserves discriminator-specific indexes (gh-15966)', async function() {
+    // Arrange
+    const addressSchema = new Schema({ city: String, street: String }, { _id: false });
+    const orderSchemaDefinition = {
+      customerId: { type: Schema.Types.ObjectId },
+      billingAddress: addressSchema,
+      shippingAddress: addressSchema
+    };
+
+    const orderSchema = new Schema(orderSchemaDefinition);
+    orderSchema.index({ customerId: 1 });
+
+    const Order = db.model('Order', orderSchema);
+
+    const wholesaleOrderSchema = new Schema({ ...orderSchemaDefinition, vendorCode: String });
+    wholesaleOrderSchema.index({ vendorCode: 1 });
+
+    // Act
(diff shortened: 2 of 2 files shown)


vkarpov15 (approved):
Works for me, thank you. Will release 8.24.4 with this :+1:

PR #16469, opened 2026-08-21
feat(schema): unbox single element arrays for scalar paths

vkarpov15:
**Summary**

Re: discussion in #16459: for 10.0 release, add auto unboxing when casting single element arrays to scalars consistently. Currently Mongoose casts `[x]` as `x` for some types (Date, ObjectId, UUID) but not others (String, Number; recently changed Int32 and Double). I think unboxing single element arrays is a reasonable convenience that can be handy in cases like casting query strings where multiple keys are allowed, and there's no risk of data loss when allowing `['2026-06-01']` to be casted as `'2026-06-01'` - empty arrays and multi-element arrays explicitly error out but single element arrays are coerced.

@AbdelrahmanHafez @hasezoey @official-burak WDYT?

**Examples**


AbdelrahmanHafez (commented):
Looks good overall, found a few coverage gaps and edge-cases, addressed them in #16471

hasezoey (approved):

vkarpov15:
@AbdelrahmanHafez @hasezoey @IslandRhythms  Claude had a potential concern about converting empty string to `null` - currently query string `?myId=` gets parsed as `myId: ''`, queries `MyModel.find({ myId: req.query.myId })` error out with a CastError. However, with this PR's changes to cast `''` -> `null` for ObjectIds, that query would instead execute as `MyModel.find({ myId: null })` which would match any document where `myId` is not set. Or, more risky: `MyModel.deleteMany({ myId: req.query.myId })`.

Overall a relatively minor concern - only affecting query strings, and not likely to escalate to a security concern because `req.query.myId` would not be used for authentication and authorization in a reasonable application. But I would still like to hear your thoughts on this.

AbdelrahmanHafez:
Good catch. I confirmed this affects every non-String scalar type because `''` becomes `null` and matches missing fields. Since we're shipping this in 10.0, I think we should make every non-String scalar type reject empty strings. WDYT?

If you agree, happy to make this change.

hasezoey (approved):

vkarpov15:
@AbdelrahmanHafez done :+1:

AbdelrahmanHafez (approved):

lib/cast/bigint.js
```diff
@@ -30,7 +30,7 @@ module.exports = function castBigInt(val) {
     return val;
   }
   if (val === '') {
-    return null;
+    throw new Error('Cannot convert value to BigInt: ""');
```
hasezoey, line 33: Shouldnt this (and anywhere else) also be a `CastError`?

lib/cast/decimal128.js
```diff
@@ -18,7 +18,7 @@ module.exports = function castDecimal128(value) {
     return value;
   }
   if (value === '') {
-    return null;
+    assert.ok(false);
```
hasezoey, line 21: Shouldnt this be a `CastError`? (otherwise maybe add documentation?)

hasezoey (changes requested):
The casting errors are looking like a mess currently (even before this PR)... Or i am misunderstanding something.

lib/cast/decimal128.js
vkarpov15: Not necessarily because `SchemaType.prototype.cast` is responsible for creating a `CastError`. Although the error thrown here is stored as the CastError's `reason` so we should store a more reasonable message here - in 616fbd4 I made it so that `lib/cast` functions all throw vanilla `Error` instead of `CastError` or `AssertionError`.

PR #16470, opened 2026-08-23
perf: omit cloned properties with object rest

AbdelrahmanHafez:
re https://github.com/Automattic/mongoose/pull/16350#discussion_r3833901281

I also verified this with a microbenchmark on Node 24.15.0: I was surprised to learn that object rest was 1.4x to 2.4x faster than clone plus `delete`. I applied this change to multiple other places where we were also using `delete`s while keeping things behaviorally identical.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -43,6 +43,7 @@ export default defineConfig([
       'no-constant-condition': 'off',
       'no-multi-spaces': 'error',
       'func-call-spacing': 'error',
+      'no-unused-vars': ['error', { ignoreRestSiblings: true }],
       'no-trailing-spaces': 'error',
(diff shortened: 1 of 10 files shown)


vkarpov15 (approved):
Thanks! Yeah `delete` is a common cause of JIT de-opt: on `delete`, V8 automatically changes the object to "dictionary mode". Something for us to avoid.

AbdelrahmanHafez:
I've been thinking, we should create a file with all the perf optimizations we usually do (includes vs indexOf, delete vs object rest). Persisting this information in the repo helps agents find them out. We can also add these instructions to GH Copilot's reviewer.
These are the ones I remember now, I'll create a PR with these changes and then you can extend the list later if needed.

PR #16471, opened 2026-08-23
fix(schema): preserve scalar semantics when unboxing arrays

AbdelrahmanHafez:
re #16469

The main goal of this PR was to add coverage for unboxing single-element arrays across scalar types.

The tests found a few inconsistent cases. `[null]`, `[undefined]`, and `['']` didn't always behave like their scalar values. Some casters also accepted nested arrays through JavaScript coercion.

This PR fixes those cases by unboxing before the existing nullish and empty-string checks. JavaScript coercion caused some casters to accept nested arrays such as [['42']], so this PR rejects them consistently. The tests also cover empty and multi-element arrays, required validation, updates, implicit `$in`, and `sanitizeFilter`.
WDYT? @vkarpov15

### DIFF
--- a/test/cast.scalar.test.js
+++ b/test/cast.scalar.test.js
@@ -127,7 +127,8 @@ describe('scalar casting from arrays', function() {
       viewCount: { type: BigInt, required: true }
     });
     const RequiredScalarValues = mongoose.model('RequiredScalarValues', requiredScalarValuesSchema);
-    const requiredPaths = Object.keys(requiredScalarValuesSchema.paths).filter(path => path !== '_id');
+    const requiredPaths = Object.keys(requiredScalarValuesSchema.paths).
+      filter(path => path !== '_id' && path !== '__v');
 
     return { casterCases, RequiredScalarValues, requiredPaths };
   }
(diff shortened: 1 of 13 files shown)


test/cast.scalar.test.js
```diff
@@ -0,0 +1,121 @@
...
+
+const mongoose = start.mongoose;
+const Schema = mongoose.Schema;
+
+describe('scalar casting from arrays', function() {
+  const uuid = '09190f70-3d30-11e5-8814-0f4df9a59c41';
+  const objectId = '0123456789abcdef01234567';
+  const casterCases = [
+    { name: 'BigInt', cast: castBigInt, value: '42', emptyStringIsNull: true },
+    { name: 'Boolean', cast: castBoolean, value: 'true' },
```
vkarpov15, line 25: On a related note, we should have `emptyStringIsNull: true` for all schema types except `string`.

The rationale for this behavior is that HTML text inputs send empty string - a form submitting an `<input type="number">` with no input would submit an empty string,

vkarpov15 (approved):
Thanks :+1:

vkarpov15:
I merged this in a separate commit since I'm a little confused about how github stacked PRs work, it doesn't seem like I can just merge this PR without merging the parent PR as well. So I'll close this since the changes in this PR are already merged

PR #16474, opened 2026-08-23
perf(model): index bulkSave write errors by document id

yoominho91:
## Summary

- index MongoBulkWriteError document ids once before updating bulkSave document state
- replace the per-document linear writeErrors scan with Set membership
- add coverage for multiple write errors with ordered: false, ensuring only successful documents transition out of isNew

## Why

bulkSave is intended for large document batches, including 10K+ documents. When an unordered bulk write returns multiple errors, the current state-update loop scans every write error for every document, making this local bookkeeping O(documents × write errors). The Set keeps the same id-string comparison while making the lookup O(documents + write errors).

This only affects the MongoBulkWriteError path; successful writes and the default single-error ordered path keep the same behavior.

## Local lookup benchmark

Node 22.14.0, 10,000 synthetic documents, median of 5 runs. Times include building the Set.

| Write errors | Current | Set | Speedup |
|---:|---:|---:|---:|
| 1 | 0.51 ms | 0.26 ms | 1.9x |
| 100 | 16.41 ms | 0.29 ms | 56.6x |
| 1,000 | 179.10 ms | 0.32 ms | 551.5x |
| 5,000 | 701.56 ms | 0.48 ms | 1470.5x |

This benchmark isolates the in-memory document/error matching rather than MongoDB network time.

## Verification

- npm test -- -g bulkSave — 38 passing, 2 pending, 0 failing
- npm run lint -- --quiet
- fixearly pre-PR repository convention check
- fixearly O(n²) candidate count: 13 before, 12 after

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -3654,14 +3654,19 @@ Model.bulkSave = async function bulkSave(documents, options) {
     );
   }
 
-  const failedDocumentIds = new Set(bulkWriteError?.writeErrors?.map(writeError => {
+  const failedDocumentIds = documents.length >= 25 ? new Set(bulkWriteError?.writeErrors?.map(writeError => {
     const writeErrorDocumentId = writeError.err.op._id || writeError.err.op.q._id;
     return writeErrorDocumentId.toString();
-  }));
+  })) : null;
   const successfulDocuments = [];
   for (let i = 0; i < documents.length; i++) {
     const document = documents[i];
-    if (!failedDocumentIds.has(document._doc._id.toString())) {
+    const documentId = document._doc._id.toString();
+    const failed = failedDocumentIds == null ? bulkWriteError?.writeErrors.find(writeError => {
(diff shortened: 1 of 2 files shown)


lib/model.js
```diff
@@ -3654,15 +3654,14 @@ Model.bulkSave = async function bulkSave(documents, options) {
     );
   }
 
+  const failedDocumentIds = new Set(bulkWriteError?.writeErrors?.map(writeError => {
```
vkarpov15: For performance reasons I think we should avoid creating a `new Set()` unless `documents.length >= 25`. Reason being that creating a `new Set()` is expensive, so we don't want to create a Set unless there's enough `find()` calls to offset the cost of the Set. And somewhere in the 10-25 range is where Set will be faster than find()

Image

vkarpov15 (changes requested):

vkarpov15 (approved):
Thanks :+1:

PR #16478, opened 2026-08-25
docs(schema): clarify duplicate index warning to note index is not created (gh-16476)

### LINKED ISSUE #16476: Duplicate index warning does not say the duplicate is never created
### Prerequisites

### Mongoose version

9.9.3

### Node.js version

22.23.1

### MongoDB server version

8.2.6

_No response_

### Description

Declaring an index with both `index: true` on the path and `schema.index()` on the same key produces two index definitions. MongoDB rejects the second one with IndexOptionsConflict, so options that exist only on the second definition (expireAfterSeconds, unique, partialFilterExpression) are never applied.

Mongoose does warn about the duplicate, but the warning only names the cause:

> Duplicate schema index on {"at":1} for model "Census". This is often due to declaring an index using both "index: true" and "schema.index()". Please remove the duplicate index definition.

That reads like a schema hygiene note. Nothing in it says one of the two indexes was not created, which is the part that matters. For a TTL index it means retention silently never happens and the collection grows forever.

The behavior is already known internally. test/schema.test.js carries the comment "createIndexes() will error because MongoDB refuses to create duplicate indexes, but the warning should still be emitted before the error". It just never reached the user-facing message.

### Steps to Reproduce

```js
'use strict';

async function run() {
  console.log('mongoose', mongoose.version);
  await mongoose.connect('mongodb://127.0.0.1:27017/gh_repro');

    at: { type: Date, required: true, index: true }
  }, { autoIndex: false });

  schema.index({ at: 1 }, { expireAfterSeconds: 60 });

  const Test = mongoose.model('Census', schema);
  await Test.collection.drop().catch(() => {});
  await Test.createCollection();

  const err = await Test.createIndexes().then(() => null, e => e);
  console.log('createIndexes error:', err && err.codeName);

  const indexes = await Test.collection.indexes();
  console.log('indexes:', JSON.stringify(
    indexes.map(i => ({ name: i.name, expireAfterSeconds: i.expireAfterSeconds }))
  ));

  await mongoose.disconnect();
}
run();
```

Output:

```
mongoose 9.9.3
[MONGOOSE] Warning: mongoose: Duplicate schema index on {"at":1} for model "Census". This is often due to declaring an index using both "index: true" and "schema.index()". Please remove the duplicate index definition.
createIndexes error: IndexOptionsConflict
indexes: [{"name":"_id_"},{"name":"at_1"}]
```

Note `at_1` has no `expireAfterSeconds`.

### Expected Behavior

The warning should say the duplicate is not created and that options on the later definition are not applied, so a reader can tell a hygiene note apart from a TTL that silently does not exist. Something like:

> Duplicate schema index on {"at":1} for model "Census". This is often due to declaring an index using both "index: true" and "schema.index()". MongoDB rejects the duplicate with an IndexOptionsConflict error, so options on the later definition (for example expireAfterSeconds or unique) are not applied. Please remove the duplicate index definition.

I am not proposing a behavior change. I saw #15093 and #15135 and understand the preference for warning over failing, so this is only about the message text.

I have a PR ready for this.

vjymisal0:
**Summary**

Clarifies the duplicate index warning in `lib/model.js` to explicitly state that MongoDB does not create duplicate indexes, and options on the duplicate definition (such as `expireAfterSeconds` or `unique`) will not be applied (gh-16476).

Fixes #16476

**Examples**

When declaring a duplicate index on the same path using both `{ index: true }` and `schema.index()`:

Before:
> `mongoose: Duplicate schema index on {"at":1} for model "Census". This is often due to declaring an index using both "index: true" and "schema.index()". Please remove the duplicate index definition.`

After:
> `mongoose: Duplicate schema index on {"at":1} for model "Census". This is often due to declaring an index using both "index: true" and "schema.index()". MongoDB will not create the duplicate index and options on the duplicate definition (such as expireAfterSeconds or unique) will not be applied. Please remove the duplicate index definition.`

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -1727,6 +1727,7 @@ function _ensureIndexes(model, options, callback) {
           utils.warn('mongoose: Duplicate schema index on ' + JSON.stringify(fields) +
             ' for model "' + model.modelName + '". ' +
             'This is often due to declaring an index using both "index: true" and "schema.index()". ' +
+            'MongoDB will not create the duplicate index and options on the duplicate definition (such as expireAfterSeconds or unique) will not be applied. ' +
             'Please remove the duplicate index definition.');
           break;
         }
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -3562,6 +3562,10 @@ describe('schema', function() {
         message.includes('for model "gh15056"'),
         'Warning should include model name'
       );
+      assert.ok(
+        message.includes('MongoDB will not create the duplicate index and options on the duplicate definition'),
+        'Warning should mention duplicate index is not created and options not applied (gh-16476)'
+      );
     } finally {
       sinon.restore();
     }


vkarpov15 (approved):
Thanks :+1:

PR #16479, opened 2026-08-25
BREAKING CHANGE: remove adhocPaths and on-the-fly casting for get() and set()

vkarpov15:
**Summary**

"On-the-fly" casting and adhoc paths is an old pattern that allowed defining custom paths with types on an individual document, disconnected from the schema. For example, `doc.set('myNewPath', 42, Number)` would add `myNewPath` as a path on the schema with type `Number`. After that, you could use `doc.myNewPath = 43` like any other schema path.

However, it doesn't look like on-the-fly casting is used much. Most content recommends using `strict: false`. Also on-the-fly casting isn't as well tested - there aren't tests which cover things like populating adhoc paths or using subdocuments as adhoc paths. Removing adhocs lets us simplify the API a bit and removes a whole class of corner cases that we would need to test if we want to button up on-the-fly casting.

Similarly, there's a `doc.get('age', String)` code path that casts `age` to a String before returning. IMO this code path isn't especially valuable. 

**Examples**


AbdelrahmanHafez (approved):

lib/document.js
```diff
@@ -1090,16 +1090,9 @@ Document.prototype.overwrite = function overwrite(obj) {
  * @api public
  */
 
-Document.prototype.$set = function $set(path, val, type, options) {
-  if (utils.isPOJO(type)) {
-    options = type;
-    type = undefined;
-  }
-
+Document.prototype.$set = function $set(path, val, options, constructing) {
```
hasezoey, line 1094: Just for those who missed the switch of the arguments, maybe add a little check again `options` being a boolean.

hasezoey (commented):
Looks good to me, though the typescript types are not updated yet.
Also, i dont know if it is planned, but a deprecation warning before 10.0 would likely be good to have.

lib/document.js
vkarpov15: I added a check for boolean, function, or string - that should cover all cases. And helped me catch a leftover test case from #5800 that mistakenly used the adhoc `Number` cast

docs/migrating_to_10.md
```diff
@@ -0,0 +1,30 @@
...
+Mongoose 10 no longer supports on-the-fly (also called ad hoc) casting for document paths. In Mongoose 9, you could pass a type as the third argument to `Document#set()` or the second argument to `Document#get()` to cast a value for a path that is not in the schema. Mongoose 10 ignores these type arguments and no longer stores ad hoc schema types on individual documents.
+
+```javascript
+const schema = new mongoose.Schema({}, { strict: false });
+const Test = mongoose.model('Test', schema);
+const doc = new Test();
+
+// Worked in Mongoose 9, but no longer casts in Mongoose 10
+doc.set('count', '42', Number);
+doc.get('count', Number); // string '42' in Mongoose 10
```
hasezoey: Isnt this wrong now, as it throws a error?

hasezoey (approved):

docs/migrating_to_10.md
vkarpov15: Not as implemented, but this logic should throw an error you're right, I made that change in 1cb343a

PR #16480, opened 2026-08-26
fix(document): replace {MODEL} in custom cast error messages from document validation

### LINKED ISSUE #8300: Ability to change CastError error message

I am reporting a bug.

A schema property of type `ObjectId` cannot be validated with a custom validator because the casting to `ObjectId` fails before the validator is called.

```js

const schema = new Schema({
  user: {
    type: Schema.Types.ObjectId,
    validate: {
      message: 'User ID is invalid',
      validator: function(value) {
        return mongoose.Types.ObjectId.isValid(value);
      }
    }
  }
});

const Thing = mongoose.model('Thing', schema);

const thing = new Thing({
  user: 'foo'
});

thing.save()
  .then(() => console.log('Thing saved'))
  .catch(err => console.error(err.stack))
  .finally(() => mongoose.disconnect());
```

The error is `ValidationError: Thing validation failed: user: Cast to ObjectID failed for value "foo" at path "user"`.

I should be able to validate the format of the property before casting. With the code above, I am expecting to see a validation error with my custom error message.

Note that this seems to have already happened in the pase: https://github.com/Automattic/mongoose/issues/2237 (Aug-Dec 2014)

* Node.js: 12.10.0
* MongoDB: 4.0.3
* Mongoose: 5.7.7

AbinMadathil-Celigo:
**Summary**

Fixes #8300

`CastError` supports the `{MODEL}` placeholder in custom cast messages (set via the schema path's `cast` option), and the maintainer noted in the issue that `{MODEL}` support would be useful. However, `{MODEL}` is currently only replaced for **query** cast errors, because `CastError#setModel()` is only called from `Query#cast()` (lib/query.js). Cast errors raised during **document validation** (`$set` → `invalidate()`) never get a model, so the literal text `{MODEL}` leaks into the error message:

```javascript
const schema = new Schema({
  age: { type: Number, cast: '{VALUE} is not a valid number for model {MODEL}' }
});

const err = new User({ age: 'twenty' }).validateSync();
// Before: '"twenty" is not a valid number for model {MODEL}'
// After:  '"twenty" is not a valid number for model User'
```

The fix calls `err.setModel()` in `Document.prototype.invalidate()` for cast errors that have a custom message format, using the owner document's constructor (so single nested subdocument paths get the top-level model). Function-style cast messages `(value, path, model, kind) => ...` now also receive the model for document validation.

Default cast error messages are intentionally unchanged: `setModel()` is only called when a custom `messageFormat` is set, so `Cast to Number failed for value "twenty" (type string) at path "age"` stays as-is and there is no change for users who do not customize cast messages.

Also documents `{MODEL}` in the cast error templating section of `docs/validation.md`, which previously only listed `{PATH}`, `{VALUE}`, and `{KIND}`.

**Examples**

Added tests to the existing `path-level custom cast (gh-8300)` block in `test/schema.test.js` covering:

1. `{MODEL}` replaced with the model name on top-level document validation
2. `{MODEL}` replaced with the top-level model name for single nested subdocument paths
3. Function-style cast messages receiving the model on document validation

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -379,6 +379,7 @@ Mongoose's cast error message templating supports the following parameters:
 * `{PATH}`: the path that failed to cast
 * `{VALUE}`: a string representation of the value that failed to cast
 * `{KIND}`: the type that Mongoose attempted to cast to, like `'String'` or `'Number'`
+* `{MODEL}`: the name of the model that the value was being cast for, if the error is associated with a model
 
 You can also define a function that Mongoose will call to get the cast error message as follows.
 
--- a/lib/document.js
+++ b/lib/document.js
@@ -28,6 +28,7 @@ const compile = require('./helpers/document/compile').compile;
 const defineKey = require('./helpers/document/compile').defineKey;
 const firstKey = require('./helpers/firstKey');
 const flatten = require('./helpers/common').flatten;
+const getConstructorName = require('./helpers/getConstructorName');
 const getEmbeddedDiscriminatorPath = require('./helpers/document/getEmbeddedDiscriminatorPath');
 const getKeysInSchemaOrder = require('./helpers/schema/getKeysInSchemaOrder');
 const getSubdocumentStrictValue = require('./helpers/schema/getSubdocumentStrictValue');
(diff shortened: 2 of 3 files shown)


AbinMadathil-Celigo:
@vkarpov15 please do review this PR

vkarpov15 (approved):
Thanks :+1: there's some test failures to patch up but I'll fix those up in master

Issue #15863, opened 2025-12-04
[Suggestion] make `limit` with `populate` more reliabe.


AbdelrahmanHafez:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I'm opening this issue as a place to discuss potential improvements for `limit` and `perDocumentLimit` with `populate`, thoughts may not be very organized since we have so many options.

## Mongoose current behavior

Given the following `posts` when populating `comments` with `limit` 2
```js
[
  {
    title: 'Post 1',
    commentsIds: ['1', '2', '3', '4', '5']
  },
  {
    title: 'Post 2',
    commentsIds: ['6', '7', '8', '9', '10']
  },
]
```

Currently if we use `limit` mongoose sends a query on `comments` with `limit` of `4` (posts.length * limit) with the following ids: `1,2,3,4,5,6,7,8,9,10`, which results in getting back 4 ids relying on [natural order](https://www.mongodb.com/docs/manual/reference/glossary/#std-term-natural-order), meaning that even though both posts have comments, there's no guarantee they'll both get 2 comments populated, because Mongo sends back `1` to `4` when tested (we shouldn't rely on that fact since natural order can change).

This is explained in the docs: [`limit` vs. `perDocumentLimit`](https://mongoosejs.com/docs/8.x/docs/populate.html#limit-vs-perDocumentLimit). 

### Considerations for potential improvements
* When populating documents like `commentsIds` above, there's no guarantee that they refer to documents that exist in the `comments` collection. From my experience, most people never hard-delete documents (not without cleaning up their references).
* We don't necessarily need a single limit/perDocumentLimit algorithm to handle all scenarios, we could find optimizations that are reliable when certain conditions are met, this would be better for performance, but adds to the maintenance cost of the project.
* For most of these optimizations, when users pass `sort`, or `match` inside the `populate` object, those optimizations would be invalidated and we'd fallback to default/current behavior, at least in the initial phase.
* If we decide to proceed with any of those 

## Optimization #1 for `perDocumentLimit`
Instead of always sending a query per document in `perDocumentLimit`, there is a scenario when `allCommentsIds.length <= posts.length * perDocumentLimit`. In that case, sending a single find query would guarantee we'd get back all the comments. The implementation for this change can be as simple as adding those two lines in a condition somewhere
```
opts.limit = opts.perDocumentLimit;
delete opts.perDocumentLimit;
[...]

PR #16481, opened 2026-08-26
perf(populate): skip per-document splitting when all docs have IDs within perDocumentLimit

lokeshllkumar:
Addresses #15863 

# Summary

When `perDocumentLimit` is set in a `populate()` operation, Mongoose unconditionally fires one query per parent document. This is unnecessary when every document's stored reference count is already at or below the limit i.e a single combined query is more efficient and produces the same result. This approach builds on Optimization 1 mentioned in the linked issue.

# Changes

- `lib/helpers/populate/splitPopulateQuery.js`: skip the per-document split for non-virtual populates when every document's ID count <= `perDocumentLimit`. Also clears `perDocumentLimit` from a soft-copy of `options` so the downstream query builder doesn't apply an incorrect limit to the combined query.

> In virtual populate, `mod.ids` holds the parent document's local field values (one per doc), not the child reference counts, so checking `mod.ids.length <= perDocumentLimit` would compare the wrong thing and skip splitting incorrectly. A few populate tests failed without this guard.

# Testing

- All existing populate tests pass
- All `perDocumentLimit` tests pass

### DIFF
--- a/lib/helpers/populate/splitPopulateQuery.js
+++ b/lib/helpers/populate/splitPopulateQuery.js
@@ -52,6 +52,17 @@ function splitPopulateQuery(mod, ids, select, assignmentOpts) {
     return null;
   }
 
+  if (perDocumentLimit != null && !mod.isVirtual &&
+      mod.ids.every(docIds => utils.array.flatten(Array.isArray(docIds) ? docIds : [docIds], flatten).length <= perDocumentLimit)) {
+    mod.options = { ...mod.options };
+    delete mod.options.perDocumentLimit;
+    if (mod.options.options != null) {
+      mod.options.options = { ...mod.options.options };
+      delete mod.options.options.perDocumentLimit;
+    }
+    return null;
+  }
+
   return mod.docs.map((doc, i) => {
     const subMod = {
       ...mod,


lokeshllkumar:
@abdelrahman-elkady @vkarpov15 

I ran a local benchmark test by populating MongoDB with upto 1000 parent docs and 4 child docs for each, comparing the time taken before and after the query split optimization. 

| Parent Docs | BEFORE(ms) | AFTER (ms) | Speedup |
|-------|-----------------|---------------|---------|
|    50 |           13.27 |          6.17 |   ~2x |
|   200 |           50.10 |         14.40 |   ~3.5x |
|   500 |          116.11 |         29.47 |   ~4x |
|  1000 |          264.38 |         53.33 |   ~5x |

lib/helpers/populate/splitPopulateQuery.js
```diff
@@ -52,6 +52,17 @@ function splitPopulateQuery(mod, ids, select, assignmentOpts) {
     return null;
   }
 
+  if (perDocumentLimit != null && !mod.isVirtual &&
+      mod.ids.every(docIds => utils.array.flatten(Array.isArray(docIds) ? docIds : [docIds], flatten).length <= perDocumentLimit)) {
+    mod.options = { ...mod.options };
+    delete mod.options.perDocumentLimit;
```
vkarpov15, line 58: Avoid using `delete` for performance implications - better to set `mod.options.perDocumentLimit = undefined` or exclude it from the spread like `const { perDocumentLimit, ...remainingOptions } = mod.options`.

vkarpov15 (commented):
There's a test failure and `delete` usage in this PR, so I'll close this and create a separate PR with the fixes

vkarpov15:
Actually upon further review I'm not entirely convinced this performance improvement is worthwhile at all. Yes it does help in the case where `perDocumentLimit` is extremely loose and every document has `< perDocumentLimit` foreign ids. But in that case the user would likely not use `perDocumentLimit` at all - `perDocumentLimit` is only useful when it cuts off some of the result set, in which case this additional loop is just extra wasted work.

PR #16482, opened 2026-08-27
types: make Schema.prototype.static() return type with new static

vkarpov15:
**Summary**

Analog to #16464 but for `Schema.prototype.static()` - make `static()` return a schema with the static added. This way when writing plugins you can more easily add statics to schemas.

**Examples**


AbdelrahmanHafez (approved):

AbdelrahmanHafez:
> Copilot reviewed 1 out of 2 changed files

Why though?

PR #16486, opened 2026-08-31
types(model): add missing properties to listSearchIndexes() return type

### LINKED ISSUE #16485: `listSearchIndexes()` return type does not match reality
### Prerequisites

### Mongoose version

9.9.2

### Node.js version

any

### MongoDB server version

8.3

6.0.3

### Description

The return type of `listSearchIndexes()` is listed as:

```ts
Promise<Array<{ name: string }>>
```

However the actually returned object has the following shape:

```ts
{
  id: ObjectId,
  name: string,
  type: string,
  status: string,
  queryable: boolean,
  latestVersion: number,
  latestDefinition: { mappings: { dynamic: boolean, fields: Record<string, unknown>] }
}
```

The `type` key is probably an enum and the fields definition could potentially be supplied as a type argument if it can't be inferred from the model's types.

### Steps to Reproduce

```ts
const schema = new Schema({ name: { type: String, unique: true } });
const Customer = mongoose.model('Customer', schema);

await Customer.createSearchIndex({ name: 'test', definition: { mappings: { dynamic: true } } });
const res = await Customer.listSearchIndexes();
console.log(res.status) // Property 'status' does not exist on type '{ name: string; }'.ts(2339)
```

### Expected Behavior

All exposed attributes are defined as types

lazerg:
**Summary**

`listSearchIndexes()` is typed as returning `Array<{ name: string }>`, but `$listSearchIndexes` returns quite a bit more than the name: `id`, `status`, `queryable`, `latestDefinition` and friends. So reading any of those off the result is a compile error even though the value is there at runtime.

This adds a `SearchIndexInfo` interface with the fields the server documents, and uses it as the return type. The optional ones (`type`, `latestVersion`, `statusDetail`, the synonym mapping fields) aren't returned by every server version, so they're marked optional.

Fixes #16485

**Examples**

```ts
const indexes = await Customer.listSearchIndexes();
indexes[0].status; // 'READY'
indexes[0].queryable; // true
```


types/indexes.d.ts
```diff
@@ -94,4 +94,33 @@ declare module 'mongoose' {
...
+   */
+  interface SearchIndexInfo {
+    /** The index's unique identifier, a hex string rather than an ObjectId. */
+    id: string;
+    /** The name of the index. */
+    name: string;
+    /** The type of the index, `search` or `vectorSearch`. Only returned for Atlas deployments. */
+    type?: string;
+    /** The build state of the index, for example `BUILDING`, `READY` or `FAILED`. */
+    status: string;
```
AbdelrahmanHafez: Can we add a named union for the [seven documented values](https://www.mongodb.com/docs/manual/reference/operator/aggregation/listSearchIndexes/#mongodb-search-index-statuses) and reuse it for nested statuses?

types/indexes.d.ts
```diff
@@ -94,4 +94,33 @@ declare module 'mongoose' {
   };
 
   type SearchIndexDescription = mongodb.SearchIndexDescription;
+
+  /**
+   * A single entry of the [`$listSearchIndexes`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/listSearchIndexes/)
+   * output, as returned by `Model.prototype.listSearchIndexes()`.
```
AbdelrahmanHafez: `listSearchIndexes()` is static, so this should be `Model.listSearchIndexes()`.

types/indexes.d.ts
```diff
@@ -94,4 +94,33 @@ declare module 'mongoose' {
...
+   */
+  interface SearchIndexInfo {
+    /** The index's unique identifier, a hex string rather than an ObjectId. */
+    id: string;
+    /** The name of the index. */
+    name: string;
+    /** The type of the index, `search` or `vectorSearch`. Only returned for Atlas deployments. */
+    type?: string;
```
AbdelrahmanHafez: Can we use the documented `search` and `vectorSearch` literals here?

types/indexes.d.ts
```diff
@@ -94,4 +94,33 @@ declare module 'mongoose' {
...
+    /** Whether the index is ready to serve queries. */
+    queryable: boolean;
+    /** The version number of the most recent definition. */
+    latestVersion?: number;
+    /** When the most recent definition was created, and its version number. */
+    latestDefinitionVersion?: { version: number, createdAt: Date };
+    /** The most recent definition of the index. */
+    latestDefinition: AnyObject;
+    /** The status of the index on each individual search node. */
+    statusDetail?: AnyObject[];
```
AbdelrahmanHafez: MongoDB documents these objects. Can we type `statusDetail` and `synonymMappingStatusDetail` instead of using `AnyObject[]`?

types/indexes.d.ts
```diff
@@ -94,4 +94,33 @@ declare module 'mongoose' {
...
+    /** The status of the index's synonym mappings, only returned when the index defines any. */
+    synonymMappingStatus?: string;
```
AbdelrahmanHafez: Can we use a separate union for `BUILDING`, `FAILED`, and `READY` here?

test/types/models.test.ts
```diff
@@ -1512,3 +1513,12 @@ function gh15940() {
...
   expect(User.hydrate<{ totalOrders: number }>).type.not.toBeCallableWith(rawUser);
 }
+
+async function gh16485() {
+  const schema = new Schema({ name: String });
+  const Customer = model('Customer', schema);
+
+  const indexes = await Customer.listSearchIndexes();
+  expect(indexes).type.toBe<Array<SearchIndexInfo>>();
+  expect(indexes[0].status).type.toBe<string>();
```
AbdelrahmanHafez: This assertion blocks the status union. Can we assert the named type and keep field-level checks for `SearchIndexInfo`?

AbdelrahmanHafez (changes requested):
Welcome @lazerg 

A few improvements, otherwise LGTM

lazerg:
All six points are in 38dbcd93b.

`SearchIndexStatus` covers the seven documented values, and `statusDetail[n].status` plus the nested `mainIndex`/`stagedIndex` reuse it. Synonym mappings only report `BUILDING`, `FAILED` or `READY`, so they get their own `SearchIndexSynonymMappingStatus`.

`type` is now `'search' | 'vectorSearch'`, and the doc comment says `Model.listSearchIndexes()`.

`statusDetail` is `SearchIndexStatusDetail[]` with `hostname`, `status`, `queryable`, `mainIndex` and an optional `stagedIndex`. Both index generations use `SearchIndexGenerationDetail`.

For `synonymMappingStatusDetail` the docs contradict themselves. The field table puts `status` and `queryable` directly on each array element. The sample output keys each element by the mapping name, like `[{ synonym_mapping: { status: 'READY', queryable: true } }]`. I followed the sample output, so the type is `Array<Record<string, SearchIndexSynonymMappingDetail>>`. Tell me if you read it the other way and I will flip it.

The test asserts `SearchIndexInfo` on an element, then checks `id`, `queryable`, `status`, `type` and `statusDetail` field by field.

One thing this turned up: inside `declare module 'mongoose'`, `Date` resolves to the `Schema.Types.Date` alias in `schematypes.d.ts`, not to the global. So `createdAt` was wrong in my earlier commit. It uses `NativeDate` now. I checked it by type-checking the sample output from the docs page against the new interface.

AbdelrahmanHafez (approved):

vkarpov15 (approved):
Thanks :+1:

PR #16501, opened 2026-09-08
BREAKING CHANGE: remove automatic query merging in `find()` etc.

vkarpov15:
**Summary**

Removes support for calling `const q = Model.find(); await Model.find(q);` - this syntax adds a lot of overrides in `models.d.ts` and `queries.d.ts` and these overrides add to a meaningful number of instantiations in TypeScript. The models.d.ts overrides alone add about 13k instantiations.

I made `Model.find(q)` throw an error, but `Model.find().merge(q)` is still supported - that is a better way to more explicitly merge one query into another.

**Examples**

### DIFF
--- a/docs/migrating_to_10.md
+++ b/docs/migrating_to_10.md
@@ -28,3 +28,18 @@ Cast values explicitly instead of relying on adhoc type casting.
 
 The function signature for `Document#set()` and its alias `Document#$set()` is now `function set(path, val, options?)` - the 3rd argument is now `options`.
 The `type` argument has been removed.
+
+## Removed support for passing a query to model query methods
+
+Mongoose 10 no longer supports passing a query instance as the filter to model query methods like `find()` and `findOne()`.
+
+If you need to copy a query into another query, use `Query.prototype.merge()`:
+
+```javascript
+const query = User.find({ status: 'active' }).select('name');
+const queryToRun = User.find().merge(query);
+
+await queryToRun.exec();
+```
+
+`Query.prototype.merge()` copies the query's conditions, field selection, and options to the query it is called on.
--- a/lib/query.js
+++ b/lib/query.js
@@ -5760,8 +5760,7 @@ Query.prototype.model;
  */
 
 function canMerge(value) {
-  return value instanceof Query || utils.isObject(value) || value === null;
-
+  return utils.isObject(value) || value === null;
 }
 
 /*!
--- a/lib/query.js
+++ b/lib/query.js
@@ -5760,7 +5760,7 @@ Query.prototype.model;
  */
 
 function canMerge(value) {
-  return utils.isObject(value) || value === null;
+  return (!(value instanceof Query) && utils.isObject(value)) || value === null;
 }
 
 /*!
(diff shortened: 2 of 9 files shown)


AbdelrahmanHafez (approved):
Didn't know that this was a thing. Glad it's gone.

Also, feels good:
Image

hasezoey (approved):

vkarpov15:
@AbdelrahmanHafez could you please take another look? I made some substantial changes since you last reviewed - I consolidated more overrides leading to about 8% fewer instantiations. Huge TypeScript perf win.

AbdelrahmanHafez (commented):
> about 8% fewer instantiations. Huge TypeScript perf win

Nice!

I used Astra to take a look on the PR now, found a couple of findings that I think, at least the first, is worth fixing and adding regressions for:

---AI generated 👇---

1. Explicit result types lose option-based return types

Could we preserve the option-based return type when callers provide an explicit result type?

```ts
const result = await User.findOneAndUpdate<UserDoc>(
  { name: 'Alice' },
  { name: 'Bob' },
  { includeResultMetadata: true }
);

result.value;
```

Before this change, TypeScript returns `ModifyResult<UserDoc>`. After this change, it returns `UserDoc | null`, so accessing `result.value` fails type checking.

When we supply `UserDoc`, TypeScript uses the default type for `Options` instead of inferring it from the argument. The return type therefore loses `includeResultMetadata: true`.

The same issue makes explicit result types nullable with `{ upsert: true, new: true }`.

2. Delete results become non-null with unsupported upsert options

Could we keep the upsert-specific return type limited to update and replace operations?

```ts
const result = await User.findOneAndDelete(
  { name: 'Alice' },
  { upsert: true, new: true }
);

result.name; // TypeScript accepts this, but result can be null.
```

`findOneAndDelete()` does not support upsert. However, its shared options type accepts these properties, and the new result helper uses them to remove `null`.

The operation still returns `null` when nothing matches. Both model and chained `findByIdAndDelete()` have the same issue.

This is a narrower concern because it requires options that do not apply to deletion.

vkarpov15:
1. Explicit result types lose option-based return types

This is a good call and a blocking issue. I will undo the `Options` generic refactor because there's no good way to work around explicit result types losing option-based return types. Once the first generic parameter is specified, subsequent generics get their default value instead of inferring - just how TypeScript works.

2. Delete results become non-null with unsupported upsert options

This is no longer an issue if we revert the `Options` generic as per (1). Another good reason to revert the `Options` refactor because yes `upsert` should not remove `null` from the result type for `findOneAndDelete`... but on the other hand, `findOneAndDelete(filter, options).findOneAndUpdate()` _should_ not have `null` in its result type. 

Without the Options refactor, the instantiations reduction is closer to 4%: less good, but better than nothing.

vkarpov15:
@AbdelrahmanHafez I found a better way to reduce instantiations - refactor out the document and query functions on `Model` into separate interfaces. Reduces instantiations from 395k to 360k while retaining the exact same overrides :rocket:

PR #16505, opened 2026-09-11
types: relax populate Model generics to allow typed virtuals to be passed to populate in TypeScript 7 with skipLibCheck

### LINKED ISSUE #16503: 9.10.0: Model with typed virtuals is no longer assignable to `Model<any>`
### Prerequisites

### Mongoose version

`9.10.0` (regression vs `9.9.5`)

### Node.js version

`24.21.0`

### MongoDB server version

n/a — compile-time TypeScript error

`7.0.2`

### Description

In 9.10.0, a `Model` with a non-empty `TVirtuals` generic is no longer assignable to `Model<any>` (non-empty `TInstanceMethods`/`TQueryHelpers` alone still work). This breaks every API typed as `string | Model<any>`: [`PopulateOptions.model`](https://github.com/Automattic/mongoose/blob/9.10.0/types/populate.d.ts#L63), [`SchemaTypeOptions.ref`](https://github.com/Automattic/mongoose/blob/9.10.0/types/schematypes.d.ts#L123), [`Document#populate()`](https://github.com/Automattic/mongoose/blob/9.10.0/types/document.d.ts#L255), `Model.populate()`, and `as Model<unknown>` casts.

Likely introduced by the `Schema.prototype.queryHelper()` / `static()`/`statics` typing changes (#16464, #16350).

Notably, this release already relaxed `doc.$model()`/`doc.model()` to `Model<DocType, any, any, any>` for the same reason — see [document.d.ts#L99](https://github.com/Automattic/mongoose/blob/9.10.0/types/document.d.ts#L99) and [document.d.ts#L235](https://github.com/Automattic/mongoose/blob/9.10.0/types/document.d.ts#L235). `PopulateOptions.model` etc. were presumably missed.

This affects every `@typegoose/typegoose` project: its `ModelType` always sets `TVirtuals = { id: string }`, so all typegoose models now fail `populate({ model })`.

### Steps to Reproduce

```ts
import type { Model, PopulateOptions } from 'mongoose';

// any Model with non-empty TVirtuals
declare const m: Model<{ a: string }, {}, {}, { id: string }>;

const a: Model<any> = m;                            // TS2322 on 9.10.0, OK on 9.9.5
const p: PopulateOptions = { path: 'x', model: m }; // TS2322 on 9.10.0
```

`tsc --noEmit --strict --skipLibCheck` prints:

```
error TS2322: Type 'Model<{ a: string; }, {}, {}, { id: string; }, ...>' is not assignable to type 'Model<any, {}, {}, {}, any, any, any>'.
  The types returned by 'discriminator(...).schema.queryHelper(...).statics' are incompatible between these types.
    ...
      The 'this' types of each signature are incompatible.
        Type 'Model<any, {}, {}, {}, any, any, any>' is not assignable to type 'Model<any, {}, {}, { id: string; }, any, any, any>'.
          The types of 'schema.virtuals' are incompatible between these types.
            Property 'id' is missing in type '{}' but required in type '{ id: string; }'.
```

### Expected Behavior

`Model<Doc, Q, M, V>` should remain assignable to `Model<any>` as in 9.9.5, so typed models can be passed to `populate({ model })`, `ref`, etc. Suggested fix: type these positions as `Model<any, any, any, any>` — the same relaxation already applied to `doc.$model()`/`doc.model()` in this release.

vkarpov15:
Fix #16503

**Summary**

#16503 is a strange bug that only occurs with TypeScript 7 and `skipLibCheck: true` - relaxing some types and adding some more `any` generics fixes the issue.

**Examples**

### DIFF
--- a/test/types/populate.test.ts
+++ b/test/types/populate.test.ts
@@ -739,3 +739,13 @@ async function gh16101() {
     expect(doc.owner.name).type.toBe<string>();
   }
 }
+
+function gh16503() {
+  const m: Model<{ a: string }, {}, {}, { id: string }> = mongoose.model(
+    'Test',
+    new mongoose.Schema({ a: { type: String, required: true } })
+  );
+
+  const a: Model<any, any, any, any> = m;
+  const p: mongoose.PopulateOptions = { path: 'x', model: m };
+}
--- a/types/aggregate.d.ts
+++ b/types/aggregate.d.ts
@@ -125,7 +125,7 @@ declare module 'mongoose' {
      * Binds this aggregate to a model.
      * @param model the model to which the aggregate is to be bound
      */
-    model(model: Model<any>): this;
+    model(model: Model<any, any, any, any>): this;
 
     /**
      * Returns the current model bound to this aggregate object
--- a/types/document.d.ts
+++ b/types/document.d.ts
@@ -252,7 +252,7 @@ declare module 'mongoose' {
 
     /** Populates document references. */
     populate<Paths = {}>(path: string | PopulateOptions | (string | PopulateOptions)[]): Promise<PopulateDocumentResult<this, Paths, PopulatedPathsDocumentType<DocType, Paths>, DocType>>;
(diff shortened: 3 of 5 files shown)


hasezoey (commented):
LGTM

PR #16507, opened 2026-09-13
fix(documentarray): reindex subdocs after addToSet() skips a duplicate

giaBaoJS:
`MongooseDocumentArray` wraps `push()`, `pull()`, `shift()`, `$shift()`, `splice()`, `reverse()`, `sort()` and `unshift()` so surviving subdocs get their `__index` restamped, but `addToSet()` has no wrapper and falls through to the plain-array implementation.

That implementation casts every argument up front with a speculative index (`_mapCast` uses `this.length + index`) and only afterwards skips the arguments that are already in the array. Each value added after a skipped duplicate therefore keeps an `__index` one higher than the position it actually lands in. A later nested edit marks the wrong path, and because the delta serializes the value from whatever subdoc currently sits at that stale index, the update writes a no-op onto the wrong element and the user's change never reaches MongoDB:

```js
const user = await User.create({
  addresses: [{ city: 'Boston' }, { city: 'Chicago' }, { city: 'Denver' }]
});

user.addresses.addToSet(
  { _id: user.addresses[0]._id, city: 'Boston' }, // duplicate, skipped
  { city: 'Amsterdam' },
  { city: 'Rotterdam' }
);
await user.save();

user.addresses.map(a => a.__index); // [0, 1, 2, 4, 5] on a length-5 array

user.addresses[3].city = 'New York';
user.$getChanges(); // { $set: { 'addresses.4.city': 'Rotterdam' } }
```

After `save()` the document still reads `['Boston', 'Chicago', 'Denver', 'Amsterdam', 'Rotterdam']`. Validation error paths are wrong for the same reason: a missing required field on the subdoc at position 3 is reported as `addresses.4.street`.

This PR adds the missing `addToSet()` wrapper, restamping subdoc indexes after the underlying call in the same shape as the siblings. With it the example gives `__index` `[0, 1, 2, 3, 4]`, `{ $set: { 'addresses.3.city': 'New York' } }`, and `['Boston', 'Chicago', 'Denver', 'New York', 'Rotterdam']` in the database.

The wrapper deliberately restamps indexes only and does not call `_updateParentPopulated()`. `addToSet()` appends and never reorders, so the populated cache order cannot change, and rebuilding the cache here would extend it with an `undefined` entry for the appended subdoc, which makes a later `depopulate()` emit `undefined` in place of that subdoc's `ObjectId`. `push()` already behaves that way today, so this looked worth leaving to a separate change rather than propagating.

Follow-up to #16282, which added the sweep to the other mutators. Its note that append-only `push()` can stay on the fast path because "the new subdoc is already cast with the correct appended index" holds for `push()` but not for `addToSet()`, which casts first and skips afterwards.

### Tests

Two regression tests in `test/types.documentarray.test.js` under a new `document array indexes after addToSet` block, asserting the `addToSet()` return value, `__index`, `$__fullPath()`, `modifiedPaths()`, `$getChanges()`, the persisted document, and the validation error path. Both go red on master and green with the fix. The fixture uses five distinct cities and places the duplicate first so the appended entries actually shift; the bug is invisible on a single-element array or one where every entry is identical.

`npm run test:ci` goes from 4557 to 4559 passing, 47 pending, no failures.


vkarpov15 (approved):
Thanks :+1:

vkarpov15:
I also backported to 8.x.

Issue #16504, opened 2026-09-11
Add `Query.prototype.findAndCount()` for composable paginated queries


nikzanda:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Hi @vkarpov15, thank you for considering the `Model.findAndCount()` proposal I opened in [#16454](https://github.com/Automattic/mongoose/issues/16454), and for implementing it in Mongoose `9.10.0` through [#16460](https://github.com/Automattic/mongoose/pull/16460).

It would also be useful to expose the same operation on an existing `Query` instance, especially when queries are built incrementally and passed between functions.

Would you consider adding `Query.prototype.findAndCount()` so that a previously constructed find query can be executed as follows?

```javascript
const query = buildQuery(model, filters);

const [documents, total] = await query
  .skip(40)
  .limit(20)
  .findAndCount();
```

The proposed method would be a terminal operation returning a promise that resolves to `[documents, total]`, consistent with `Model.findAndCount()`.

### Motivation

Reusable query-building functions can apply filters and sorting, then return a query for another function to paginate and execute.

Adopting `Model.findAndCount()` in this situation means extracting the existing query's filter, projection and options and constructing another model-level call. A method on `Query` would allow the caller to execute the query it already has.

The following example illustrates this pattern using a generic model and placeholder fields.

### Build a query

Here, `model` is any Mongoose model with `status` and `createdAt` fields. This synchronous helper returns an unexecuted query.

```javascript
function buildQuery(model, filters = {}) {
  const query = model.find();

  if (filters.status) {
    query.where('status').equals(filters.status);
  }

  query.sort({ createdAt: -1, _id: -1 });

  return query;
}
```

### Current approach: clone the query for data and count

```javascript
async function getPage(model, filters, offset, limit) {
  const query = buildQuery(model, filters);

  const dataQuery = query.clone();
  const countQuery = query.clone();

  return {
    documents: await dataQuery.skip(offset).limit(limit),
    total: await countQuery.countDocuments(),
  };
}
```

The count must use the same filtering conditions without inheriting the pagination applied to the data query. This requires manually coordinating the two queries at each call site or writing a custom wrapper.

### Proposed approach: execute findAndCount() on the existing query

The query builder above would remain unchanged. Only the execution step would change:

```javascript
async function getPage(model, filters, offset, limit) {
  const query = buildQuery(model, filters);

[...]

PR #16508, opened 2026-09-14
feat: add Query.prototype.findAndCount() for composable paginated queries

lokeshllkumar:
Addresses #16504 

Base branch set to [9.11](https://github.com/Automattic/mongoose/tree/9.11), respecting the milestone.

# Summary

This PR adds `Query.prototype.findAndCoutn()`, allowing a previously constructed query to be executed directly instead of requiring filter/projection/options to be passed all at once via `Model.findAndCount()`. Query builders that compose filters and sorting incrementally across functions can now paginate and execute the query where it's used, without extracting its filter/options back out of the query object.

# Changes

- `lib.query.js`: Add `Query.prototype.findAndCount()` as a terminal operation that validates `sort` and `limit` are set, clones the query into a `countDocuments` query with skip/limit omitted, and runs both in parallel.
- `types/query.d.ts`: Add TypeScript type signature for `findAndCount()`.

# Testing

- All 4 existing tests for `Model.findAndCount()` pass.
- 4 new tests are added to cover basic usage, incrementaly built queries, and sort/limit validation errors.


vkarpov15 (approved):
Thanks :+1:

PR #16510, opened 2026-09-15
test: run TypeScript tests on typescript 5+7 with combinations of exactOptionalPropertyTypes, skipLibCheck, strict

### LINKED ISSUE #16506: Run tests against TypeScript 7
### Prerequisites

### Issue

In #16503 we found our first TypeScript-7-specific bug, so it's about time we ran tests against both TypeScript 5 and TypeScript 7. This bug was also `skipLibCheck` specific, so we likely need a testing matrix of TypeScript versions, `exactOptionalProperties` true/false, `skipLibCheck` true/false, and `strict` true/false.

vkarpov15:
Fix #16506

**Summary**

Run TypeScript tests on all permutations of TypeScript 5/7, exactOptionalPropertyTypes true/false, skipLibCheck true/false, strict true/false. In an abundance of caution to minimize risk of surprise errors under different configurations.

**Examples**


hasezoey (approved):
Looks good, though that makes test jobs quite messy.

AbdelrahmanHafez (commented):
Other than [Copilot's comment](https://github.com/Automattic/mongoose/pull/16510/changes#r4017386061), LGTM 👍

.github/workflows/types.yml
```diff
@@ -48,4 +57,21 @@ jobs:
       - run: npm install
 
       - name: Typings
-        run: npm run test:types
+        # TSTyche 7.2.5 does not support the TypeScript 7 compiler API. The
+        # TypeScript 7 jobs compile the complete suite with tsc below.
+        if: matrix.typescript == '5.9.3'
+        # TSTyche assertions and Mongoose's conditional types require these
+        # individual checks even when the strict umbrella option is disabled.
```
AbdelrahmanHafez: The `strict=false` rows still keep four strict checks enabled (`strictNullChecks`, `noImplicitAny`, `noImplicitThis`, and `strictFunctionTypes`) because the existing tests require them. Other strict checks change, so these rows still add useful coverage. However, they don't cover fully non-strict projects. I'd clarify that here, but I don't think this needs to block merging.

```suggestion
        # Both steps keep strictNullChecks, noImplicitAny, noImplicitThis,
        # and strictFunctionTypes enabled because the existing tests require them.
        # The strict matrix option toggles the remaining strict checks.
        # This matrix does not cover fully non-strict configurations.
```

vkarpov15:
I removed the strict true/false dimension because tests fail with `strict: false` and there doesn't seem to be a maintainable way to make the tests work properly with strict: true and strict: false. So for now I think we focus on the task at hand: run tests against both TypeScript 5 and TypeScript 7. Keep the skipLibCheck true/false dimension as an added bonus, but strict true/false out of scope for now.

PR #16514, opened 2026-09-16
fix: preserve circular arrays when cloning with a seen map

Shubham-Padkonde:
**Summary**

Calling `toObject()` on a document whose Mixed path contains a self-referencing array throws `RangeError: Maximum call stack size exceeded`. Object cloning already tracks visited values through `_seen`, but array cloning does not.

Register each cloned array before visiting its elements and reuse previously cloned arrays. This preserves circular and shared array references consistently with object cloning.

**Examples**

Added regression tests for a document with a self-referencing Mixed array, mutually referencing arrays, and a shared array. All three fail on the base revision and pass with the fix.

Validation: the standalone regression suite and ESLint for both changed files pass. The database-backed suite could not run because MongoDB startup is denied in the test environment; it still needs upstream CI.

Prepared with Codex assistance.


AbdelrahmanHafez (approved):
Welcome @Shubham-Padkonde 
LGTM, thanks!

vkarpov15 (approved):
Thanks :+1:

PR #16518, opened 2026-09-21
fix(schema): register path-level setters on Map paths

giaBaoJS:
**Summary**

A `set` function declared on a Map path never runs. A `get` on the same path does.

`set` is listed under "All Schema Types" in the [SchemaType options docs](https://mongoosejs.com/docs/schematypes.html#schematype-options): "function, defines a custom setter for this property".

The cause is a name collision. Sixteen schema types expose the "set a default option for all X instances" helper as a static, `SchemaString.set = SchemaType.set;`. `SchemaMap` declares it inside the class body instead, so `SchemaMap.prototype.set` shadows `SchemaType.prototype.set`, which is the method the `SchemaType` constructor calls to register `options.set`. Declaring that path also writes a key named after the stringified setter into `SchemaType.defaultOptions`, process-wide.

This moves the helper to `SchemaMap.set = SchemaType.set` so it matches the other sixteen types. The static keeps working, since it already resolved through class inheritance; `test/schematype.test.js` asserts it for every schema type.

Union paths have the same symptom from a different cause. `new Schema({ u: { type: 'Union', of: ['String', 'Number'], set: v => v.trim() } })` ignores the setter because `Union.prototype.applySetters` overrides the base method and only applies the setters of the types in `of`. I left it out because it needs a decision on whether the union-level setter runs before or after the per-type ones. Can send that separately.

**Examples**

```js
const schema = new Schema({
  books: {
    type: Map,
    of: String,
    set: v => Object.fromEntries(Object.entries(v).map(([k, val]) => [k.toLowerCase(), val]))
  }
});

const doc = new Model({ books: { 'Casino-Royale': 'Casino Royale' } });
Array.from(doc.books.keys()); // ['Casino-Royale'] on master, ['casino-royale'] with this change
```

One regression test in `test/types.map.test.js` covers construction and reassignment. It goes red on master with `'Casino-Royale' !== 'casino-royale'`.


vkarpov15 (approved):
Good find, thank you :+1:

PR #16521, opened 2026-09-21
fix(schema): drop subpaths and singleNestedPaths when a path is removed

### LINKED ISSUE #16520: Schema#remove() leaves subpaths and singleNestedPaths behind, so a removed path keeps casting
### Prerequisites

### Mongoose version

9.10.1 (master at 38e59009)

### Node.js version

22.20.0

### MongoDB server version

Not needed, everything below happens in the schema and in query casting

### Description

`schema.remove()` deletes the path from `paths`, `nested`, `mapPaths` and the tree, but it does not touch `subpaths` (where a document array keeps its element paths) or `singleNestedPaths` (where a subdocument keeps its paths). Those entries survive the removal, so `path()` still returns a schematype for the removed path, `pathType()` still says `real`, and a query on the removed path is still cast as if the path existed.

`omit()` calls `remove()` and has the same leftovers. It only deletes a `singleNestedPaths` entry when the omitted path is that exact key, so omitting the subdocument itself leaves all of its child paths behind.

### Steps to Reproduce

```js
const { Schema } = mongoose;

const child = new Schema({ name: String, nested: { x: Number } }, { _id: false });
const schema = new Schema({ child, arr: [{ v: Number }], tags: [String] });
schema.path('arr.0.v'); // positional lookups register the element paths lazily
schema.path('tags.$');

schema.remove(['child', 'arr', 'tags']);

console.log(Object.keys(schema.paths));             // [ '_id' ]
console.log(Object.keys(schema.singleNestedPaths)); // [ 'child.name', 'child.nested.x', 'child.nested' ]
console.log(Object.keys(schema.subpaths));          // [ 'arr.v', 'arr._id', 'tags.$', 'arr.$.v' ]

console.log(schema.pathType('child.name'), !!schema.path('child.name')); // real true
console.log(schema.pathType('arr.0.v'), !!schema.path('arr.0.v'));       // real true

const M = mongoose.model('T', schema);
const q = M.find({ 'child.name': 123, 'arr.0.v': '5' });
q.cast(M);
console.log(q.getFilter()); // { 'child.name': '123', 'arr.0.v': 5 }, cast through paths the schema no longer has

const omitted = new Schema({ child, n: Number }).omit(['child']);
console.log(Object.keys(omitted.singleNestedPaths)); // [ 'child.name', 'child.nested.x', 'child.nested' ]
```

### Expected Behavior

After `remove()` or `omit()` the removed path and everything under it is gone from `subpaths` and `singleNestedPaths` as well, the same way `mapPaths` is cleaned up today: `path('child.name')` is `undefined`, `pathType('arr.0.v')` is `adhocOrUndefined`, and the filter above is left as `{ 'child.name': 123, 'arr.0.v': '5' }`.

luantaraschi:
**Summary**

Fix #16520

`remove()` cleans `paths`, `nested`, `mapPaths` and the tree, and stops there. A document array keeps its element paths in `subpaths` and a subdocument keeps its paths in `singleNestedPaths`, and both survived the removal. The visible side of that: `schema.path('child.name')` still returned a schematype after `remove('child')`, `pathType('arr.0.v')` still said `real` after `remove('arr')`, and `Model.find({ 'child.name': 123 })` was still cast through a path the schema no longer had.

This adds a `_deleteSubpaths` helper that drops the exact key and everything under `name + '.'` from both maps, called from the two branches of `remove()` next to the existing `mapPaths` filter. `omit()` goes through `remove()`, so its own loop over `singleNestedPaths`, which only handled the exact key, is no longer needed and is gone.

**Examples**

Two tests, `removes the subpaths of a document array and of a subdocument (gh-16520)` under `remove()` and `removes the paths of an omitted subdocument (gh-16520)` under `omit()`. Both fail on master (`Expected values to be strictly deep-equal` on the leftover keys) and pass with the change. `test/schema.test.js` is 239 passing, and `document`, `query` and `model` are 1379 passing against MongoDB 8.2.12, with no change to the pending count.


vkarpov15 (approved):
Thanks :+1:

